PR #4, opened 2025-05-21
Enable ci

kwryankrattiger:
Enable Gitlab CI


.ci/gitlab/.gitlab-ci.yml
```diff
@@ -159,7 +174,7 @@ default:
   interruptible: true
   timeout: 60 minutes
   retry:
-    max: 2
+    max: 1
```
zackgalbreath: why retry generate jobs less?

kwryankrattiger: to avoid spamming failures during testing. Just pulled out testing hacks since I got green CI.

PR #16, opened 2025-06-04
intel-oneapi-mkl: Make `threads!=none` work after compilers-as-nodes

msimberg:
> [!NOTE]
> This PR was migrated from **https://github.com/spack/spack/pull/50233**.

Previously the `intel-oneapi-mkl` package (or the `oneapi` build system) had an implicit dependency on a compiler, from which openmp library paths were extracted. After compilers-as-nodes, `intel-oneapi-mkl` no longer has an implicit compiler, so this detection broke.

This PR adds an explicit `depends_on("c")` to the `intel-oneapi-mkl` package (could be `cxx` as well, but I figured `c` is the most basic compiler from which the openmp paths can be extracted). This means that the `oneapi` build system can query the `"c"` package for compiler paths and call the compiler. I also had to change the way the compiler is used in `oneapi.py` (@alalazo, does that look sane to you?).

A bit of an unrelated change, but this also changes the `tbb` dependency to be optional.

As before, using `intel-oneapi-mkl` remains broken as an external when `threads!=none`, since externals don't have dependencies (this was also implicit before compilers-as-nodes), and I'm not attempting to fix that here.

### DIFF
--- a/repos/spack_repo/builtin/build_systems/oneapi.py
+++ b/repos/spack_repo/builtin/build_systems/oneapi.py
@@ -11,6 +11,7 @@ from llnl.util.link_tree import LinkTree
 
 import spack.util.path
 from spack.build_environment import dso_suffix
+from spack.compilers.libraries import CompilerPropertyDetector
 from spack.package import (
     EnvironmentModifications,
     Executable,
@@ -219,8 +220,8 @@ class IntelOneApiLibraryPackage(IntelOneApiPackage):
             )
 
         # query the compiler for the library path
-        with self.compiler.compiler_environment():
-            omp_lib_path = Executable(self.compiler.cc)(
+        with CompilerPropertyDetector(self["c"].spec).compiler_environment():
+            omp_lib_path = Executable(self["c"].cc)(
                 "--print-file-name", f"{libname}.{dso_suffix}", output=str
             ).strip()
 
--- a/repos/spack_repo/builtin/packages/intel_oneapi_mkl/package.py
+++ b/repos/spack_repo/builtin/packages/intel_oneapi_mkl/package.py
@@ -191,7 +191,10 @@ class IntelOneapiMkl(IntelOneApiLibraryPackage):
         msg="MKL with OpenMP threading requires GCC, clang, or Intel compilers",
     )
 
-    depends_on("tbb")
+    # Nothing is compiled, but openmp_libs in the oneapi build system queries
+    # the compiler for openmp library paths.
+    depends_on("c", when="threads=openmp")
+    depends_on("tbb", when="threads=tbb")
     # cluster libraries need mpi
     depends_on("mpi", when="+cluster")
 


tldahlgren:
Please resolve the merge conflict markings.

repos/spack_repo/builtin/build_systems/oneapi.py
```diff
@@ -219,10 +220,36 @@ def openmp_libs(self):
             )
 
         # query the compiler for the library path
+<<<<<<< HEAD:var/spack/repos/spack_repo/builtin/build_systems/oneapi.py
+<<<<<<< HEAD:var/spack/repos/spack_repo/builtin/build_systems/oneapi.py
+        with CompilerPropertyDetector(self["c"].spec).compiler_environment():
+            omp_lib_path = Executable(self["c"].cc)(
+                "--print-file-name", f"{libname}.{dso_suffix}", output=str
+            ).strip()
+||||||| parent of 15700f5e99 (Workaround for broken oneapi library search):lib/spack/spack/build_systems/oneapi.py
```
tldahlgren: The conflicts need to be resolved.

repos/spack_repo/builtin/build_systems/oneapi.py
```diff
@@ -219,10 +220,36 @@ def openmp_libs(self):
...
                 "--print-file-name", f"{libname}.{dso_suffix}", output=str
             ).strip()
+=======
+        # TODO: Why does the following line find no libraries?
+        # with self.compiler.compiler_environment():
+        omp_lib_path = Executable(self.compiler.cc)(
+            "--print-file-name", f"{libname}.{dso_suffix}", output=str
+        ).strip()
+>>>>>>> 15700f5e99 (Workaround for broken oneapi library search):lib/spack/spack/build_systems/oneapi.py
+||||||| parent of b6b6acdd76 (Update compiler_environment in oneapi build system):lib/spack/spack/build_systems/oneapi.py
```
tldahlgren: Same here

tldahlgren (changes requested):
Please cleanup the merged content.

repos/spack_repo/builtin/build_systems/oneapi.py
msimberg: Oof, sorry and thanks for spotting that! Fixed.

repos/spack_repo/builtin/build_systems/oneapi.py
msimberg: Fixed here as well.

tldahlgren (approved):
Let's see if CI passes.  Would also like to hear from @alalazo

tldahlgren:
CI is failing. Let's try again after updating the branch.

hfp:
At least the changes in `oneapi.py` can cause:
> OneAPI package requires OpenMP library: /lib/x86_64-linux-gnu/libomp.so

which appears to be wrong given I wanted to build using the GNU compiler (`%gcc`) but relying for instance on Intel MKL aka oneMKL (it shall use libgomp.so rather than libomp.so). Specifically, I am working on something like:

> spack install cp2k@master %gcc ^intel-oneapi-mpi ^intel-oneapi-mkl@2025 +cluster +gfortran threads=openmp ^dbcsr@develop

msimberg:
> At least the changes in `oneapi.py` can cause:

Are you saying it behaved differently before, or that with this branch this is simply what you get (without regard for what it did before)? This PR isn't trying to change any behaviour, just fix the build in the first place after compilers as nodes broke it. I suspect if it gives you the wrong OMP library now, it probably gave you the wrong one before as well. It's somewhat strange that it would give you `libomp` with gcc though, given that it tries to look for `libgomp`: https://github.com/spack/spack-packages/blob/808259eb938e5607595247f3e047d7171da76c66/repos/spack_repo/builtin/build_systems/oneapi.py#L211-L212.

This PR focuses solely on getting `intel-oneapi-mkl` to a working state again after compilers as nodes.

Let me be clear: I really dislike the way the package handles openmp threading and welcome simplification from intel, but I'd like that possible simplification to not block getting this PR merged unless you think this makes things worse than before.

hfp:
> Are you saying it behaved differently before, or that with this branch this is simply what you get (without regard for what it did before)?

I started to try some own fixes for https://github.com/spack/spack-packages/issues/1213, basically just `depends_on("c", type="build")`. Then colleagues pointed me to your PR, and I merged it into my [fork](https://github.com/hfp/spack-packages) for testing it. I got said message with the changes in `oneapi.py`. However, I did not notice any issue wrt lib[g]omp when only taking the changes in `repos/spack_repo/builtin/packages/intel_oneapi_mkl/package.py` (I will double-check as you suggested). To say something more helpful: perhaps it's useful to have two separate PRs.

msimberg:
> > Are you saying it behaved differently before, or that with this branch this is simply what you get (without regard for what it did before)?
> 
> I started to try some own fixes for [spack/spack-packages#1213](https://github.com/spack/spack-packages/issues/1213), basically just `depends_on("c", type="build")`. Then colleagues pointed me to your PR, and I merged it into my [fork](https://github.com/hfp/spack-packages) for testing it. I got said message with the changes in `oneapi.py`. However, I did not notice any issue wrt lib[g]omp when only taking the changes in `repos/spack_repo/builtin/packages/intel_oneapi_mkl/package.py` (I will double-check as you suggested). To say something more helpful: perhaps it's useful to have two separate PRs.

Interesting... thanks for the details! I had more or less the opposite experience (though it's a while ago already): without the changes to openmp library detection I wasn't able to install intel-oneapi-mkl at all. I'm happy to make a separate PR for the compiler dependency, since that's a must anyway. This PR can then act as a follow-up to have further fixes in.

msimberg:
I opened https://github.com/spack/spack-packages/pull/401 with only the `depends_on("c")` change.

alalazo:
@msimberg I started looking into this. The main issue is that we look at `intel-oneapi-mkl` compiler to decide which library to return for the `.libs` attribute. This is inherently wrong, since we should look at the compiler the caller is using instead.

I think the:
```
depends_on("c", type="build")
```
change for `intel-mkl` is not the right workaround for this issue, because:
1. It introduces a dependency that should not be there (i.e. `intel-oneapi-mkl` does not really depend on a C compiler)
2. It prevent external `intel-oneapi-mkl` to be working

The best solution is imho to implement https://github.com/spack/spack/issues/49202, but we don't have the time before v1.0 to do that. 

A possible alternative in the meanwhile could be to use the query arguments, similarly to what the `hdf5` package is doing, or find an alternative way to let the `intel-oneapi-mkl` package know which compiler is being used by its parent node. What do you think about that?

teabagk7:
> @msimberg I started looking into this. The main issue is that we look at `intel-oneapi-mkl` compiler to decide which library to return for the `.libs` attribute. This is inherently wrong, since we should look at the compiler the caller is using instead.
> 
> I think the:
> 
> ```
> depends_on("c", type="build")
> ```
> 
> change for `intel-mkl` is not the right workaround for this issue, because:
> 
> 1. It introduces a dependency that should not be there (i.e. `intel-oneapi-mkl` does not really depend on a C compiler)
> 2. It prevent external `intel-oneapi-mkl` to be working
> 
> The best solution is imho to implement [spack/spack#49202](https://github.com/spack/spack/issues/49202), but we don't have the time before v1.0 to do that.
> 
> A possible alternative in the meanwhile could be to use the query arguments, similarly to what the `hdf5` package is doing, or find an alternative way to let the `intel-oneapi-mkl` package know which compiler is being used by its parent node. What do you think about that?

Taking compiler from parent node is good option. Seems the most logical to me.

hfp:
As a side-note, the `oneapi.py` seems to have more issues. For instance (develop branch), https://github.com/spack/spack-packages/blob/develop/repos/spack_repo/builtin/build_systems/oneapi.py#L288 seems to miss inheritance `class IntelOneApiStaticLibraryList` is better written `class IntelOneApiStaticLibraryList(LibraryList)` (otherwise none of the assumed attributes are present since the type is not a `LibraryList`).

msimberg:
@alalazo yeah, I agree that sounds like the better option.

Unfortunately it seems like CI is failing both here and in spack/spack#401, and I don't have time to look at this further at the moment. I was hoping to simply restore the previous behaviour, but that seems to be trickier than I hoped.

I may return to this at some point, but have no plans to do so in the near future, so I'll close this in case someone else wants to pick things up from here.

hfp:
Apparently, removing https://github.com/spack/spack-packages/blob/develop/repos/spack_repo/builtin/packages/intel_oneapi_mkl/package.py#L184-L192 solves the problem too. I think these lines are close to "micro-management" and are practically not missing. I will prepare a PR if there are no objections.

PR #29, opened 2025-06-06
README.md: document repository structure

tgamblin:
So that I can point to this when people ask why the repository is structured this way.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,6 +4,34 @@ This is the default [Spack](https://github.com/spack/spack) packages repository,
 contains the set of packages maintained by the Spack community. In Spack v1.0 and later,
 the repository here is automatically added to the Spack configuration.
 
+## Structure of this repo
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -4,6 +4,34 @@ This is the default [Spack](https://github.com/spack/spack) packages repository,
 contains the set of packages maintained by the Spack community. In Spack v1.0 and later,
 the repository here is automatically added to the Spack configuration.
 
+## Structure of this repo
+
+This repository does not look like the original Spack package repositories. Its structure
```
haampie, line 9: For newcomers in Spack 1.0 (hopefully) I don't think we need to talk about Spack 0.x structure?

bernhardkaindl (approved):
On the review comment by haampie: As most all users will be existing users/developers, I don't think that's valid, and we should merge this.

PR #31, opened 2025-06-06
dpdk: enable building on aarch64

mdorier:
Enable building dpdk on an aarch64 machine.


mdorier:
Note: I think this package should also have `depends_on("libmd")`. I've had it fail to build on some platforms and I think the reason it builds on some platforms and not others depends on whether python is external or not. A spack-built python will bring libmd into the environment. An external python may not have it.

hyoklee (approved):

tldahlgren (approved):

PR #39, opened 2025-06-06
ML CI: update image, build TF/JAX with Clang

adamjstewart:
Updates to new image with LLVM installed.

Builds Google stuff (TF/JAX) with Clang and Meta stuff (PyTorch) with GCC.

Except TF + aarch64, which seems to build better with GCC for some reason 🤷


adamjstewart:
No specs to rebuild surprises me, but let's see if I can get things building with Clang instead of GCC.

.ci/gitlab/stacks/ml-linux-x86_64-rocm/spack.yaml
```diff
@@ -13,17 +13,22 @@ spack:
       - amdgpu_target=gfx90a
     gl:
       require: "osmesa"
+    miopen-hip:
+      require: ~ck
     mpi:
       require: openmpi
+    py-jax:
```
rbberger: here and in all other changed files:
```suggestion
    py-jaxlib:
```

.ci/gitlab/stacks/e4s/spack.yaml
```diff
@@ -100,7 +106,7 @@ spack:
   - dxt-explorer
   - dyninst
   - e4s-cl
-  - ecp-data-vis-sdk ~cuda ~rocm +adios2 +ascent +cinema +darshan +faodel +hdf5 +paraview +pnetcdf +sz +unifyfs +veloc +visit +vtkm +zfp # 
+  - ecp-data-vis-sdk ~cuda ~rocm +adios2 +ascent +cinema +darshan +faodel +hdf5 +paraview +pnetcdf +sz +unifyfs +veloc +visit +vtkm +zfp #
```
rbberger, line 105: this file still needs the container change to `ghcr.io/spack/ubuntu-24.04:v2025-06-05`

.ci/gitlab/stacks/ml-linux-x86_64-cuda/spack.yaml
```diff
@@ -13,18 +13,23 @@ spack:
...
-      require: 
+      require:
       - ~cuda
       - target=x86_64_v3
     mpi:
       require: openmpi
+    py-jaxlib:
+      require: "%c,cxx=clang"
+    py-tensorflow:
+      require: "%c,cxx=clang"
```
rbberger, line 24:
```suggestion
      require: "%c,cxx=clang"
    openblas:
      require: "%c,cxx,fortran=gcc"
```

adamjstewart: Would rather add a conflict for flang

.ci/gitlab/stacks/e4s/spack.yaml
adamjstewart: I think I'll just ignore e4s and let them manage their own stack

repos/spack_repo/builtin/packages/openblas/package.py
```diff
@@ -242,6 +242,8 @@ class Openblas(CMakePackage, MakefilePackage):
         when="@0.3.27 %oneapi",
     )
 
+    conflicts("%fortran=clang")
```
rbberger: This might be too restrictive. `-mtune` seems to exists in `flang-new` in LLVM 19+.

adamjstewart: Seems to first exist in 19.1: https://github.com/llvm/llvm-project/pull/98517

adamjstewart: I guess they never had a 19.0, so we are both right.

adamjstewart: I actually wonder which version of OpenBLAS added `-mtune`, we could be even less restrictive.

repos/spack_repo/builtin/packages/openmpi/package.py
```diff
@@ -780,6 +780,8 @@ def patch(self):
     conflicts("~internal-pmix", "+internal-hwloc")
     conflicts("~internal-pmix", "+internal-libevent")
 
+    conflicts("%fortran=clang")
```
rbberger: Similar story here: confirmed to get passed the configure step with LLVM 20 (`flang` == `flang-new`). For LLVM 19 and 18 you need to have `FC=flang-new`.

adamjstewart: I'm not entirely sure how the compiler wrappers work these days, but maybe @haampie can look into this. For now, I'll restrict the conflict to llvm 19 and older.

adamjstewart:
Only remaining thing holding up this PR (and updating TF) is the libclang build failure. Two potential solutions:

* Add `llvm-config` to the image: https://github.com/spack/gitlab-runners/pull/67
* Convince Spack to build LLVM

I would honestly prefer the latter solution. We have to register an external LLVM in order to use the system clang/flang. But this external LLVM is missing many core tools and libraries. If we have a variant that controls this, we can detect that `~dev` is found but `+dev` is needed. Any suggestions @haampie  @skosukhin  @trws?

adamjstewart:
@spackbot run pipeline

adamjstewart:
For the sake of debugging, the current specs that fail to concretize within the time limit:

* ml-linux-aarch64-cuda-generate: `py-jaxlib`, `py-keras backend=jax`
* ml-linux-x86_64-cpu-generate: `py-jaxlib`
* ml-linux-x86_64-cuda-generate: `py-jax`, `py-jaxlib`
* ml-linux-x86_64-rocm-generate: `py-jax`

So it's always jax, but not always the same jax. Note that `py-keras backend=jax` depends on `py-jax`, and `py-jax` depends on `py-jaxlib`, so there's no reason it should be able to concretize one and not the other.

.ci/gitlab/.gitlab-ci.yml
```diff
@@ -105,7 +105,7 @@ default:
     SPACK_CONCRETIZER_REQUIRE_CHECKSUM: 1
     SPACK_BACKTRACE: 1
   interruptible: true
-  timeout: 60 minutes
+  timeout: 120 minutes
```
alalazo:
```suggestion
  timeout: 60 minutes
```

adamjstewart:
Now stuck on https://github.com/tensorflow/tensorflow/issues/67251, https://github.com/tensorflow/tensorflow/issues/62357

zackgalbreath (approved):

PR #57, opened 2025-06-06
gfe: protections for CMake 4.0

mathomp4:
> [!NOTE]
> This PR was migrated from **https://github.com/spack/spack/pull/50084**.

Users recently found that the GFE libraries:

- gFTL
- gFTL-shared
- fArgParse
- pFUnit
- yaFyaml
- pFlogger

have some `cmake_minimum_required(VERSION 3.0)` calls in the CMake. With CMake 4, [this is now an error](https://cmake.org/cmake/help/latest/release/4.0.html#deprecated-and-removed-features).

Efforts are underway to update all the GFE libraries to use CMake 3.24 as a minimum (which is a version we use in codes that *use* GFE, so we are comfortable with it).

But until those versions can be tested and released, we need to protect spack users from having failing builds.

You'll note below that `pfunit` is a "special" case as a user already found this out and a version, 4.12, was released that can work with CMake 4:

```python
    depends_on("cmake@3.12:3", type="build", when="@:4.11")
    depends_on("cmake@3.12:", type="build", when="@4.12")
    depends_on("cmake@3.24:", type="build", when="@4.13:")
```

but with the upcoming 4.13 we'll move to CMake 3.24 as well.

### DIFF
--- a/repos/spack_repo/builtin/packages/fargparse/package.py
+++ b/repos/spack_repo/builtin/packages/fargparse/package.py
@@ -37,7 +37,9 @@ class Fargparse(CMakePackage):
 
     depends_on("gftl-shared")
     depends_on("gftl")
-    depends_on("cmake@3.12:", type="build")
+
+    depends_on("cmake@3.12:3", type="build", when="@:1.9")
+    depends_on("cmake@3.24:", type="build", when="@1.10:")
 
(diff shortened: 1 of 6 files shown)


tldahlgren:
Added @haampie based on  prior PR.

tldahlgren (approved):

tldahlgren:
It's been weeks so merging.

PR #59, opened 2025-06-06
aocc: help flang find gcc

nolta:
> [!NOTE]
> This PR was migrated from **https://github.com/spack/spack/pull/50227**.

AOCC `flang` can't find `gcc` if `gcc` is installed in a non-standard location.

This commit:

- Patches `flang1` & `flang2` with the rpath for libquadmath.
- Creates a wrapper script to pass options to `flang`, since `flang.cfg` is currently ignored.
- Passes `--gcc-toolchain`.
- Passes `-Wl,-rpath` so executables can find libquadmath.
- Passes `-Wno-unused-command-line-argument` to silence warnings generated by `-Wl,-rpath`, since they can cause problems.

### DIFF
--- a/repos/spack_repo/builtin/packages/aocc/package.py
+++ b/repos/spack_repo/builtin/packages/aocc/package.py
@@ -1,7 +1,8 @@
 # Copyright Spack Project Developers. See COPYRIGHT file for details.
 #
 # SPDX-License-Identifier: (Apache-2.0 OR MIT)
-import os.path
+import glob
+import os
 
 from spack_repo.builtin.build_systems.compiler import CompilerPackage
 from spack_repo.builtin.build_systems.generic import Package
@@ -69,6 +70,8 @@ class Aocc(Package, LlvmDetection, CompilerPackage):
 
     depends_on("c", type="build")
     depends_on("cxx", type="build")
+    depends_on("fortran", type="build")
+    depends_on("patchelf@:0.17", when="@:5 %gcc", type="build")
 
     depends_on("libxml2")
     depends_on("zlib-api")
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/aocc/package.py
```diff
@@ -119,6 +122,33 @@ def cfg_files(self):
...
+            # help flang{1,2} find libquadmath
+            libdir = self._libquadmath_dir()
+            patchelf = which("patchelf")
+            patchelf.add_default_arg("--set-rpath", libdir)
+            patchelf(join_path(self.prefix.bin, "flang1"))
+            patchelf(join_path(self.prefix.bin, "flang2"))
+
+            # pass --gcc-toolchain & -Wl,-rpath to flang
+            # flang.cfg is ignored, so replace the flang symlink with a wrapper script
+            compiler_options += ' -Wno-unused-command-line-argument "-Wl,-rpath,{libdir}"'
```
bernhardkaindl, line 136: See the previous discussion about some concerns with this: https://github.com/spack/spack/pull/50227#discussion_r2068640916

@nolta, as you were concerned about injecting it as well, would it be an option to postpone injecting this to a later PR?

Could the change also be restricted to only the needed path like in the comment above the location that you quoted in the linked discussion?
```py
 # TODO: it is unclear whether we should really use all elements of 
 #  _ld_library_path because it looks like the only rpath that needs to be 
 #  injected is self.component_prefix.linux.compiler.lib.intel64_lin. 
 if self.v2_layout: 
     common_flags = ["-Wl,-rpath,{}".format(self.component_prefix.lib)] 
 else: 
     common_flags = ["-Wl,-rpath,{}".format(d) for d in self._ld_library_path()] 
```

nolta: 1) No, spack packages will fail to compile because of the unexpected output.

2) It already is.

My assumption is that this is the best possible solution, despite the flaws, otherwise the intel-oneapi-compilers package would have done something different.

bernhardkaindl (approved):
@amd-toolchain-support, see the comment by @nolta, in the resolved discussion thread.

As you're maintaining `aocc`, is this good to merge by you?

amd-toolchain-support (approved):
These changes looks good,

amd-toolchain-support:
> @amd-toolchain-support, see the comment by @nolta, in the resolved discussion thread.
> 
> As you're maintaining `aocc`, is this good to merge by you?

Apologies for delay, 
In my testing, did not encountered any issue with these changes as such. Approving it for merge.

bernhardkaindl (approved):
Thanks @amd-toolchain-support, will merge.

PR #61, opened 2025-06-06
gromacs: add v2024.4, apple-clang openmp support for 2025+

mathomp4:
This is a "by-hand" transfer of https://github.com/spack/spack/pull/47931 to spack-packages. The `migrate-package-prs` script by @haampie didn't work for this because of a conflict.

---

First up, I mention the maintainers of Gromacs:
* @eirrgang
* @junghans 
* @mabraham

I was trying to build Gromacs on my Mac today and when I tried I got the:
```cmake
        if(CMAKE_CXX_COMPILER_ID MATCHES "AppleClang")
            message(FATAL_ERROR "The compiler you are using does not support OpenMP parallelism, "
                "Apple has unfortunately explicitly disabled OpenMP in their clang-derived compiler. "
                "You can disable OpenMP in Gromacs with -DGMX_OPENMP=OFF, but instead "
                "we recommend installing the unsupported library distributed by the R "
                "project from https://mac.r-project.org/openmp/ - or switch to gcc.")
```
error. This is because `+openmp` is the default variant.

Now, I've seen similar in the past on projects I work on and the solution has been to depend on `llvm-openmp`. So I added that.

Further build testing showed, though, that the latest version here, 2024.3, was not yet happy as it failed in build with:
```
  >> 397    clang++: error: unknown argument: '-Xclang -fopenmp'
```

The issue was that in the verbose output it had in the flags:
```
...
-mmacosx-version-min=15.0 -fPIC "-Xclang -fopenmp" -Xclang -fopenmp -MD -MT
...
```

So, I did a bit more testing and did some `spack checksum` as well and found that `gromacs@2024.4` did not work, but `gromacs@2025.0-beta` did and so did `gromacs@main`. As there isn't yet an official 2025 version, this is sort of a...PR in waiting.

Now, the one issue is, I don't really know how to *run* Gromacs[^1] let alone test if OpenMP is working. So I'm hoping one of the maintainers can try out this change and see if it lets the macOS version of Gromacs use OpenMP.

Echoing what @mabraham did in #47935, I've added a conflict if the version is 2024 or older.

I've also added 2024.4 as...it wasn't there. 

[^1]: I know, sounds odd, but I saw [this Macinchem post](https://macinchem.org/2024/12/04/using-vmd-and-gromacs-on-m4-macs) in my RSS feeds asking about Gromacs and Apple Silicon and, well, spack was right there on my laptop so... 😄


repos/spack_repo/builtin/packages/gromacs/package.py
```diff
@@ -226,9 +226,19 @@ class Gromacs(CMakePackage, CudaPackage):
     )
 
     variant("openmp", default=True, description="Enables OpenMP at configure time")
+
+    # When using apple-clang version 15.x or newer, need to use the llvm-openmp library
+    # We also protect with version 2025+ as there seems to be a CMake bug with
+    # Apple Clang and OpenMP that is fixed in 2025
+    depends_on("llvm-openmp", when="@2025: %apple-clang@15: +openmp", type=("build", "run"))
```
tldahlgren: Syntax changed so you want the variant to be associated with this package and not the `apple-clang` compiler:

```suggestion
    depends_on("llvm-openmp", when="@2025: +openmp %apple-clang@15:", type=("build", "run"))
```

tldahlgren: `spack spec gromacs` succeeds with this change.

tldahlgren (changes requested):
The syntax has changed. See the suggested fix.

repos/spack_repo/builtin/packages/gromacs/package.py
```diff
@@ -226,9 +226,19 @@ class Gromacs(CMakePackage, CudaPackage):
     )
 
     variant("openmp", default=True, description="Enables OpenMP at configure time")
+
+    # When using apple-clang version 15.x or newer, need to use the llvm-openmp library
+    # We also protect with version 2025+ as there seems to be a CMake bug with
+    # Apple Clang and OpenMP that is fixed in 2025
+    depends_on("llvm-openmp", when="@2025: +openmp %apple-clang@15:", type=("build", "run"))
+
+    # But we need to block +openmp for Gromacs older than 2025
```
mabraham:
```suggestion
    # But we need to block +openmp %apple-clang for GROMACS older than 2025
```

mabraham: clarifying wording of comment

alalazo (dismissed):

mabraham (approved):

mabraham:
I'm a listed maintainer, why can't I approve this?

mabraham:
@alalazo Thanks for the tip. If openmp and apple-clang causes trouble, I'll revert it. But since we're only supporting the latest of everything maybe it will be ok...

mathomp4:
@alalazo I can close this if you like, in re openmp on apple-clang.

tldahlgren (approved):
Will defer to @alalazo for final decision on this PR being merged.

tldahlgren:
> Will defer to @alalazo for final decision on this PR being merged.

ping @alalazo

alalazo:
Apologies. I missed this ping, but feel free to merge if this seems ready. I really entered this review by accident, thinking I was approving another PR :slightly_smiling_face: 

The comment above https://github.com/spack/spack-packages/pull/61#event-18572972840 was really just an "fyi". I didn't meant to block (otherwise I would have used "require changes")

PR #63, opened 2025-06-06
Kokkos: add conflict between shared and cuda_relocatable_device_code

tpadioleau:
The PR adds a conflict between `+shared` and `+cuda_relocatable_device_code`. It follows the behavior found in cmake
https://github.com/kokkos/kokkos/blob/64edd7263e2d8ece5cdf140e674761dc29f7ad5a/cmake/kokkos_enable_options.cmake#L251-L253


cedricchevalier19 (approved):
Good catch.

tldahlgren (approved):

PR #67, opened 2025-06-06
audacious: add v4.3.1, v4.4 (support Qt6)

wdconinc:
> [!NOTE]
> This PR was migrated from **https://github.com/spack/spack/pull/49667**.

Note: orphaned branch by me. Created PR to make sure this doesn't drop off my radar.

As of v4.4, `audiacious` uses Qt6 as the default Qt interface (we never supported GTK, but now explicitly disable it).


tldahlgren (approved):
Confirmed the sha256s and 'spack  spec' succeeds.

PR #77, opened 2025-06-06
meson: define and define_from_variant

wdconinc:
> [!NOTE]
> This PR was migrated from **https://github.com/spack/spack/pull/40269**.

Based on user code in https://github.com/spack/spack/pull/39854/files#diff-5376f65afcea4f0f272594553d063ebb89de44fd25747f6931bef681ae755edfR69-R74 it seems we may benefit from an approach to defining build options in Meson packages that is similar to what is already in common use for Autotools and CMake packages. There are currently 23 MesonPackage packages that implement `meson_args` and may benefit from this.

Based on CMakeBuilder, this introduces:
- `MesonBuilder.define(meson_var, value)`
- `MesonBuilder.define_from_variant(self, meson_var, variant=None)`

There are the following differences:
- Meson developers don't have caps lock on, so examples use lowercase,
- Meson does not ignore options with empty values, so for conditional variant values this needs caution.

Out of scope: actually changing the packages to use this.

### DIFF
--- a/repos/spack_repo/builtin/build_systems/meson.py
+++ b/repos/spack_repo/builtin/build_systems/meson.py
@@ -2,7 +2,6 @@
 #
 # SPDX-License-Identifier: (Apache-2.0 OR MIT)
 import collections.abc
-import inspect
 import os
 from typing import List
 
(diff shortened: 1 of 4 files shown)


tldahlgren:
Added missing package maintainers.

PR #78, opened 2025-06-06
Add kokkos-fft

tpadioleau:
The PR introduces kokkos-fft.

I can see three ways to implement the recipe, for example with the `cufft` backend:

1. It is a boolean option, default is off, the `cufft` backend is enabled by explicitly requesting `kokkos-fft +cufft` (`+cufft` requires `kokkos +cuda`).
2. It is a boolean option, default is on, the `cufft` backend is enabled if we have `kokkos +cuda` (`+cufft` does not require `kokkos +cuda`)
3. Three values: auto, on, off:
   - "on"/"off" correspond to behavior 1.
   - "auto" is kind of behavior 2.

For now I have implemented the **behavior 1**.

Does it make sense ?

Note that the handling of FFTW is a bit different, we don't control whether we want the OpenMP or serial version. Should we fix this behavior ?

### DIFF
--- a/repos/spack_repo/builtin/packages/kokkos_fft/package.py
+++ b/repos/spack_repo/builtin/packages/kokkos_fft/package.py
@@ -25,7 +25,9 @@ class KokkosFft(CMakePackage):
     variant("tests", default=False, description="Enable tests")
 
     depends_on("kokkos@4.4:4 +complex_align") # not supported in Spack 0.23.1
-    requires("^kokkos +wrapper", when="^kokkos +cuda")
+    requires("^kokkos +cuda +wrapper", when="+cufft")
+    requires("^kokkos +rocm", when="+hipfft")
+    requires("^kokkos +sycl", when="+onemkl")
     depends_on("googletest@1.15:1", when="+tests")
 
     depends_on("fftw@3.3:3 ~mpi precision=float,double", when="+fftw")
@@ -34,10 +36,6 @@ class KokkosFft(CMakePackage):
     depends_on("cuda@11:12", when="+cufft")
     depends_on("hipfft@5.3:6", when="+hipfft")
     depends_on("intel-oneapi-mkl@2023:2025", when="+onemkl")
-
-    requires("^kokkos +cuda", when="+cufft")
-    requires("^kokkos +rocm", when="+hipfft")
-    requires("^kokkos +sycl", when="+onemkl")
 
     def cmake_args(self):
         args = [
--- a/repos/spack_repo/builtin/packages/kokkos_fft/package.py
+++ b/repos/spack_repo/builtin/packages/kokkos_fft/package.py
@@ -64,8 +64,8 @@ class KokkosFft(CMakePackage):
             args.append(self.define("KokkosFFT_ENABLE_ONEMKL", "ON"))
 
         if self.spec.satisfies("^kokkos+rocm"):
-            args.append(self.define("CMAKE_CXX_COMPILER", self.spec["hip"].hipcc))
+            args.append(self.define("CMAKE_CXX_COMPILER", self["hip"].hipcc))
         else:
-            args.append(self.define("CMAKE_CXX_COMPILER", self.spec["kokkos"].kokkos_cxx))
+            args.append(self.define("CMAKE_CXX_COMPILER", self["kokkos"].kokkos_cxx))
 
         return args
--- a/repos/spack_repo/builtin/packages/kokkos_fft/package.py
+++ b/repos/spack_repo/builtin/packages/kokkos_fft/package.py
@@ -29,6 +29,7 @@ class KokkosFft(CMakePackage):
     variant("tests", default=False, description="Enable tests")
 
     depends_on("kokkos@4.4:4 +complex_align") # not supported in Spack 0.23.1
+    # kokkos-fft currently only supports compilation with the Kokkos nvcc wrapper
     requires("^kokkos +cuda +wrapper", when="device_backend=cufft")
     requires("^kokkos +rocm", when="device_backend=hipfft")
     requires("^kokkos +sycl", when="device_backend=onemkl")
--- a/repos/spack_repo/builtin/packages/kokkos_fft/package.py
+++ b/repos/spack_repo/builtin/packages/kokkos_fft/package.py
@@ -28,6 +28,8 @@ class KokkosFft(CMakePackage):
             description="Enable device backend")
     variant("tests", default=False, description="Enable tests")
 
+    depends_on("cxx", type="build")
+
     depends_on("kokkos@4.4:4 +complex_align")
     # kokkos-fft currently only supports compilation with the Kokkos nvcc wrapper
     requires("^kokkos +cuda +wrapper", when="device_backend=cufft")
--- a/repos/spack_repo/builtin/packages/kokkos_fft/package.py
+++ b/repos/spack_repo/builtin/packages/kokkos_fft/package.py
@@ -38,7 +38,9 @@ class KokkosFft(CMakePackage):
     depends_on("googletest@1.15:1", when="+tests")
 
     depends_on("fftw@3.3:3 ~mpi ~openmp precision=float,double", when="host_backend=fftw-serial")
+    requires("^kokkos +serial", when="host_backend=fftw-serial")
     depends_on("fftw@3.3:3 ~mpi +openmp precision=float,double", when="host_backend=fftw-openmp")
+    requires("^kokkos +openmp", when="host_backend=fftw-openmp")
     depends_on("cuda@11:12", when="device_backend=cufft")
     depends_on("hipfft@5.3:6", when="device_backend=hipfft")
     depends_on("intel-oneapi-mkl@2023:2025", when="device_backend=onemkl")
--- a/repos/spack_repo/builtin/packages/kokkos_fft/package.py
+++ b/repos/spack_repo/builtin/packages/kokkos_fft/package.py
@@ -32,15 +32,15 @@ class KokkosFft(CMakePackage):
 
     depends_on("kokkos@4.4:4 +complex_align")
     # kokkos-fft currently only supports compilation with the Kokkos nvcc wrapper
+    requires("^kokkos +serial", when="host_backend=fftw-serial")
+    requires("^kokkos +openmp", when="host_backend=fftw-openmp")
     requires("^kokkos +cuda +wrapper", when="device_backend=cufft")
     requires("^kokkos +rocm", when="device_backend=hipfft")
     requires("^kokkos +sycl", when="device_backend=onemkl")
     depends_on("googletest@1.15:1", when="+tests")
 
     depends_on("fftw@3.3:3 ~mpi ~openmp precision=float,double", when="host_backend=fftw-serial")
-    requires("^kokkos +serial", when="host_backend=fftw-serial")
     depends_on("fftw@3.3:3 ~mpi +openmp precision=float,double", when="host_backend=fftw-openmp")
-    requires("^kokkos +openmp", when="host_backend=fftw-openmp")
     depends_on("cuda@11:12", when="device_backend=cufft")
     depends_on("hipfft@5.3:6", when="device_backend=hipfft")
     depends_on("intel-oneapi-mkl@2023:2025", when="device_backend=onemkl")
--- a/repos/spack_repo/builtin/packages/kokkos_fft/package.py
+++ b/repos/spack_repo/builtin/packages/kokkos_fft/package.py
@@ -39,8 +39,8 @@ class KokkosFft(CMakePackage):
     requires("^kokkos +sycl", when="device_backend=onemkl")
     depends_on("googletest@1.15:1", when="+tests")
 
-    depends_on("fftw@3.3:3 ~mpi ~openmp precision=float,double", when="host_backend==fftw-serial")
-    depends_on("fftw@3.3:3 ~mpi +openmp precision=float,double", when="host_backend=fftw-openmp")
+    depends_on("fftw@3.3:3 ~mpi precision=float,double")
+    requires("^fftw +openmp", when="host_backend=fftw-openmp")
     depends_on("cuda@11:12", when="device_backend=cufft")
     depends_on("hipfft@5.3:6", when="device_backend=hipfft")
     depends_on("intel-oneapi-mkl@2023:2025", when="device_backend=onemkl")
--- a/repos/spack_repo/builtin/packages/kokkos_fft/package.py
+++ b/repos/spack_repo/builtin/packages/kokkos_fft/package.py
@@ -37,9 +37,8 @@ class KokkosFft(CMakePackage):
     requires("^kokkos +sycl", when="device_backend=onemkl")
     depends_on("googletest@1.15:1", when="+tests")
 
-    depends_on("fftw@3.3:3 ~mpi precision=float,double", when="host_backend=fftw-serial")
-    depends_on("fftw@3.3:3 ~mpi precision=float,double", when="host_backend=fftw-openmp")
-    requires("^fftw +openmp", when="host_backend=fftw-openmp")
+    depends_on("fftw@3.3:3 ~mpi ~openmp precision=float,double", when="host_backend=fftw-serial")
+    depends_on("fftw@3.3:3 ~mpi +openmp precision=float,double", when="host_backend=fftw-openmp")
     depends_on("cuda@11:12", when="device_backend=cufft")
     depends_on("hipfft@5.3:6", when="device_backend=hipfft")
     depends_on("intel-oneapi-mkl@2023:2025", when="device_backend=onemkl")
--- a/repos/spack_repo/builtin/packages/kokkos_fft/package.py
+++ b/repos/spack_repo/builtin/packages/kokkos_fft/package.py
@@ -39,7 +39,7 @@ class KokkosFft(CMakePackage):
     requires("^kokkos +sycl", when="device_backend=onemkl")
     depends_on("googletest@1.15:1", when="+tests")
 
-    depends_on("fftw@3.3:3 ~mpi ~openmp precision=float,double", when="host_backend=fftw-serial")
+    depends_on("fftw@3.3:3 ~mpi ~openmp precision=float,double", when="host_backend==fftw-serial")
     depends_on("fftw@3.3:3 ~mpi +openmp precision=float,double", when="host_backend=fftw-openmp")
     depends_on("cuda@11:12", when="device_backend=cufft")
     depends_on("hipfft@5.3:6", when="device_backend=hipfft")
--- a/repos/spack_repo/builtin/packages/kokkos_fft/package.py
+++ b/repos/spack_repo/builtin/packages/kokkos_fft/package.py
@@ -55,13 +55,23 @@ class KokkosFft(CMakePackage):
 
         if self.spec.satisfies("host_backend=fftw-serial") or self.spec.satisfies("host_backend=fftw-openmp"):
             args.append(self.define("KokkosFFT_ENABLE_FFTW", "ON"))
+        else:
+            args.append(self.define("KokkosFFT_ENABLE_FFTW", "OFF"))
 
         if self.spec.satisfies("device_backend=cufft"):
             args.append(self.define("KokkosFFT_ENABLE_CUFFT", "ON"))
+        else:
+            args.append(self.define("KokkosFFT_ENABLE_CUFFT", "OFF"))
+
         if self.spec.satisfies("device_backend=hipfft"):
             args.append(self.define("KokkosFFT_ENABLE_HIPFFT", "ON"))
+        else:
+            args.append(self.define("KokkosFFT_ENABLE_HIPFFT", "OFF"))
+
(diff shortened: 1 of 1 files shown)


tpadioleau:
@yasahi-hpc if you want to have a look

tpadioleau:
Should we add a `rocfft` option ?

repos/spack_repo/builtin/packages/kokkos_fft/package.py
```diff
@@ -0,0 +1,58 @@
...
+    maintainers("cedricchevalier19", "tpadioleau", "yasahi-hpc")
+
+    license("Apache-2.0 WITH LLVM-exception OR MIT", checked_by="cedricchevalier19")
+
+    version("0.3.0", sha256="a13c423775afec5f9f79fa9a23dd6001d3d63bae9f4786b1e0cd3ed65b3993a3")
+
+    variant("fftw", default=True, description="Enable FFTW")
+    variant("cufft", default=False, description="Enable cufft")
+    variant("hipfft", default=False, description="Enable hipfft")
+    variant("onemkl", default=False, description="Enable oneMKL fft")
```
cedricchevalier19: Do we allow different device backends at once? (I do not think so)
If not, a `device_backend` variant that lists the available option can be a cleaner API. 

It is farther away from CMake options, but I think it makes more sense to explain that you can select only one host and only one device FFT backends.

repos/spack_repo/builtin/packages/kokkos_fft/package.py
```diff
@@ -0,0 +1,58 @@
...
+    variant("tests", default=False, description="Enable tests")
+
+    depends_on("kokkos@4.4:4 +complex_align") # not supported in Spack 0.23.1
+    requires("^kokkos +cuda +wrapper", when="+cufft")
```
cedricchevalier19: Comment that it is a limitation in kokkos-fft.

repos/spack_repo/builtin/packages/kokkos_fft/package.py
```diff
@@ -0,0 +1,58 @@
...
+
+    version("0.3.0", sha256="a13c423775afec5f9f79fa9a23dd6001d3d63bae9f4786b1e0cd3ed65b3993a3")
+
+    variant("fftw", default=True, description="Enable FFTW")
+    variant("cufft", default=False, description="Enable cufft")
+    variant("hipfft", default=False, description="Enable hipfft")
+    variant("onemkl", default=False, description="Enable oneMKL fft")
+    variant("tests", default=False, description="Enable tests")
+
+    depends_on("kokkos@4.4:4 +complex_align") # not supported in Spack 0.23.1
```
cedricchevalier19:
```suggestion
    depends_on("kokkos@4.4:4 +complex_align")
```

I do not think we'll have to deal with spack < 1.0

repos/spack_repo/builtin/packages/kokkos_fft/package.py
tpadioleau: I agree it is worth trying!

repos/spack_repo/builtin/packages/kokkos_fft/package.py
tpadioleau: Adastra is using Spack 0.23 but I guess we will need to backport this recipe elsewhere anyway.

repos/spack_repo/builtin/packages/kokkos_fft/package.py
tpadioleau: Btw, should we support this option ?

tpadioleau: done in 388e97b0077f54fc37666c7b10f9b0b853f3f242

repos/spack_repo/builtin/packages/kokkos_fft/package.py
tpadioleau: done in 3c3a83a9f7701bb7a0cc6e6d111ea23cf6a0adad

repos/spack_repo/builtin/packages/kokkos_fft/package.py
tpadioleau: done in be968b0ee4cb3f0d632b818774310943e51eb991, what do you think ? I kind of like it!

tpadioleau:
I have edited the recipe, there are some conditions that I don't know yet how to express better. Please improve it if you know how.

Overall I think the version with multi values variants expresses better the way kokkos-fft works (maybe Kokkos could also be expressed in this way ?)

repos/spack_repo/builtin/packages/kokkos_fft/package.py
```diff
@@ -0,0 +1,69 @@
...
+            args.append(self.define("KokkosFFT_ENABLE_CUFFT", "ON"))
+        if self.spec.satisfies("device_backend=hipfft"):
+            args.append(self.define("KokkosFFT_ENABLE_HIPFFT", "ON"))
+        if self.spec.satisfies("device_backend=onemkl"):
+            args.append(self.define("KokkosFFT_ENABLE_ONEMKL", "ON"))
+
+        if self.spec.satisfies("^kokkos+rocm"):
+            args.append(self.define("CMAKE_CXX_COMPILER", self.spec["hip"].hipcc))
+        else:
+            args.append(self.define("CMAKE_CXX_COMPILER", self.spec["kokkos"].kokkos_cxx))
```
tretre91:
```suggestion
            args.append(self.define("CMAKE_CXX_COMPILER", self["kokkos"].kokkos_cxx))
```
`self.spec["kokkos"].kokkos_cxx` doesn't work with 1.0, it should be `self["kokkos"].kokkos_cxx`
what kokkos-kernels does: https://github.com/spack/spack-packages/blob/98e5bd13ea6d53481df3cfc0090e5999c052fafa/repos/spack_repo/builtin/packages/kokkos_kernels/package.py#L281-L284

repos/spack_repo/builtin/packages/kokkos_fft/package.py
```diff
@@ -0,0 +1,69 @@
...
+            values=("fftw-serial", "fftw-openmp"),
+            multi=True,
+            description="Enable host backend")
+    variant("device_backend", default="none",
+            values=("none", "cufft", "hipfft", "onemkl"),
+            multi=False,
+            description="Enable device backend")
+    variant("tests", default=False, description="Enable tests")
+
+    depends_on("kokkos@4.4:4 +complex_align")
```
tretre91, line 40: You should also add a dependency to `cxx`
```suggestion
    depends_on("cxx", type="build")

    depends_on("kokkos@4.4:4 +complex_align")
```

repos/spack_repo/builtin/packages/kokkos_fft/package.py
tpadioleau: Thanks! I did not test it with Spack 1 actually :D

repos/spack_repo/builtin/packages/kokkos_fft/package.py
tpadioleau: yes very true

repos/spack_repo/builtin/packages/kokkos_fft/package.py
tpadioleau: done in 2a837689dd4692bc6f60b68e75a8d1d4da1d3dc7

repos/spack_repo/builtin/packages/kokkos_fft/package.py
tpadioleau: done in 285e88b4d3b288d4f4b9800d2808ed1aaabe5413

repos/spack_repo/builtin/packages/kokkos_fft/package.py
```diff
@@ -0,0 +1,71 @@
...
+    # kokkos-fft currently only supports compilation with the Kokkos nvcc wrapper
+    requires("^kokkos +cuda +wrapper", when="device_backend=cufft")
+    requires("^kokkos +rocm", when="device_backend=hipfft")
+    requires("^kokkos +sycl", when="device_backend=onemkl")
+    depends_on("googletest@1.15:1", when="+tests")
+
+    depends_on("fftw@3.3:3 ~mpi precision=float,double", when="host_backend=fftw-serial")
+    depends_on("fftw@3.3:3 ~mpi precision=float,double", when="host_backend=fftw-openmp")
+    requires("^fftw +openmp", when="host_backend=fftw-openmp")
```
tpadioleau, line 50: Can we use `fftw +openmp` when `host_backend=fftw-serial` ?

yasahi-hpc: Probably not. Unless `host_parallel_backend` of Kokkos is `OpenMP`, it may not compile

tpadioleau: Ok so you think I should reformulate:
```py
    depends_on("fftw@3.3:3 ~mpi ~openmp precision=float,double", when="host_backend=fftw-serial")
    depends_on("fftw@3.3:3 ~mpi +openmp precision=float,double", when="host_backend=fftw-openmp")
```

yasahi-hpc: Looks good

repos/spack_repo/builtin/packages/kokkos_fft/package.py
```diff
@@ -0,0 +1,71 @@
...
+            self.define("KokkosFFT_ENABLE_DOCS", False),
+            self.define("KokkosFFT_ENABLE_BENCHMARK", False),
+            self.define("KokkosFFT_ENABLE_EXAMPLES", False),
+        ]
+
+        if self.spec.satisfies("host_backend=fftw-serial") or self.spec.satisfies("host_backend=fftw-openmp"):
+            args.append(self.define("KokkosFFT_ENABLE_FFTW", "ON"))
+
+        if self.spec.satisfies("device_backend=cufft"):
+            args.append(self.define("KokkosFFT_ENABLE_CUFFT", "ON"))
```
tpadioleau: I should maybe explicitly disable the option when it does not satisfy `device_backend=cufft` ?

yasahi-hpc: Even if `cufft` is unavailable (say HIP backend), it is valid to set `KokkosFFT_ENABLE_CUFFT=ON`. As long as you have at least one enabled TPL, it would compile.

tpadioleau: I am surprised, wouldn't find_package(cuda) fail ?

yasahi-hpc: kokkos-fft tries to find package only if it makes sense (say `Kokkos_ENABLE_CUDA=ON` && `KokkosFFT_ENABLE_CUFFT=ON`). It does not try to find `cufft` if CUDA backend is not enabled.

tpadioleau: I am not sure we should try to make it work in this kind of setup, how likely this would be the actual intent of the user ?

yasahi-hpc: I think we should disallow that. 
People do not want `cufft` for non-cuda backend.

repos/spack_repo/builtin/packages/kokkos_fft/package.py
tpadioleau: done in e9bc778f153c0065dc7b3acd524d99bb54fef127

repos/spack_repo/builtin/packages/kokkos_fft/package.py
tpadioleau: done in aa671a2a8218f057e0a246b4403a1b1fa134ab20

repos/spack_repo/builtin/packages/kokkos_fft/package.py
```diff
@@ -0,0 +1,82 @@
...
+    maintainers("cedricchevalier19", "tpadioleau", "yasahi-hpc")
+
+    license("Apache-2.0 WITH LLVM-exception OR MIT", checked_by="cedricchevalier19")
+
+    version("0.3.0", sha256="a13c423775afec5f9f79fa9a23dd6001d3d63bae9f4786b1e0cd3ed65b3993a3")
+
+    variant("host_backend", default="fftw-serial",
+            values=("fftw-serial", "fftw-openmp"),
+            multi=True,
+            description="Enable host backend")
```
cedricchevalier19: The handling of the host FFTW is a little bit weird. I thought it was possible to use a multi-threaded fftw from a serial kokkos node or a serial fftw in a kokkos serial context.

repos/spack_repo/builtin/packages/kokkos_fft/package.py
```diff
@@ -0,0 +1,82 @@
...
+            description="Enable device backend")
+    variant("tests", default=False, description="Enable tests")
+
+    depends_on("cxx", type="build")
+
+    depends_on("kokkos@4.4:4 +complex_align")
+    # kokkos-fft currently only supports compilation with the Kokkos nvcc wrapper
+    requires("^kokkos +serial", when="host_backend=fftw-serial")
+    requires("^kokkos +openmp", when="host_backend=fftw-openmp")
+    requires("^kokkos +cuda +wrapper", when="device_backend=cufft")
```
cedricchevalier19, line 44: Probably more something to do in kokkos-fft, but we should not require `+wrapper`.

repos/spack_repo/builtin/packages/kokkos_fft/package.py
```diff
@@ -0,0 +1,82 @@
...
+
+        if self.spec.satisfies("device_backend=hipfft"):
+            args.append(self.define("KokkosFFT_ENABLE_HIPFFT", "ON"))
+        else:
+            args.append(self.define("KokkosFFT_ENABLE_HIPFFT", "OFF"))
+
+        if self.spec.satisfies("device_backend=onemkl"):
+            args.append(self.define("KokkosFFT_ENABLE_ONEMKL", "ON"))
+        else:
+            args.append(self.define("KokkosFFT_ENABLE_ONEMKL", "OFF"))
```
tretre91: The if conditions can be directly given as argument to `self.define`
```suggestion
        args = [
            self.define("KokkosFFT_ENABLE_INTERNAL_KOKKOS", False),
            self.define_from_variant("KokkosFFT_ENABLE_TESTS", "tests"),
            self.define("KokkosFFT_ENABLE_DOCS", False),
            self.define("KokkosFFT_ENABLE_BENCHMARK", False),
            self.define("KokkosFFT_ENABLE_EXAMPLES", False),
            self.define(
                "KokkosFFT_ENABLE_FFTW",
                self.spec.satisfies("host_backend=fftw-serial")
                or self.spec.satisfies("host_backend=fftw-openmp"),
            ),
            self.define("KokkosFFT_ENABLE_CUFFT", self.spec.satisfies("device_backend=cufft")),
            self.define("KokkosFFT_ENABLE_HIPFFT", self.spec.satisfies("device_backend=hipfft")),
            self.define("KokkosFFT_ENABLE_ONEMKL", self.spec.satisfies("device_backend=onemkl")),
        ]
```

repos/spack_repo/builtin/packages/kokkos_fft/package.py
```diff
@@ -0,0 +1,82 @@
...
+    # kokkos-fft currently only supports compilation with the Kokkos nvcc wrapper
+    requires("^kokkos +serial", when="host_backend=fftw-serial")
+    requires("^kokkos +openmp", when="host_backend=fftw-openmp")
+    requires("^kokkos +cuda +wrapper", when="device_backend=cufft")
+    requires("^kokkos +rocm", when="device_backend=hipfft")
+    requires("^kokkos +sycl", when="device_backend=onemkl")
+    depends_on("googletest@1.15:1", when="+tests")
+
+    depends_on("fftw@3.3:3 ~mpi ~openmp precision=float,double", when="host_backend=fftw-serial")
+    depends_on("fftw@3.3:3 ~mpi +openmp precision=float,double", when="host_backend=fftw-openmp")
```
tretre91: Regarding the conflicting `~openmp` and `+openmp` variants we discussed (when `host_backend=fftw-serial,fftw-openmp`), you can use `host_backend==fftw-serial` to ensure the dependency applies when only the serial backend has been requested, for the second line it will apply if `host_backend` contains `fftw-openmp`
```suggestion
    depends_on("fftw@3.3:3 ~mpi ~openmp precision=float,double", when="host_backend==fftw-serial")
    depends_on("fftw@3.3:3 ~mpi +openmp precision=float,double", when="host_backend=fftw-openmp")
```

repos/spack_repo/builtin/packages/kokkos_fft/package.py
tpadioleau: Do you mean using a single-threaded fft from a `fftw +openmp` ?

repos/spack_repo/builtin/packages/kokkos_fft/package.py
tpadioleau: Actually kokkos-fft is header-only, it should only matter for the compilation of tests. Do you agree that for now we should keep it ?

repos/spack_repo/builtin/packages/kokkos_fft/package.py
tpadioleau: ok for me

repos/spack_repo/builtin/packages/kokkos_fft/package.py
tpadioleau: Thanks a lot, i did not know this notation

repos/spack_repo/builtin/packages/kokkos_fft/package.py
```diff
@@ -0,0 +1,70 @@
...
+                self.spec.satisfies("host_backend=fftw-serial")
+                or self.spec.satisfies("host_backend=fftw-openmp"),
+            ),
+            self.define("KokkosFFT_ENABLE_CUFFT", self.spec.satisfies("device_backend=cufft")),
+            self.define("KokkosFFT_ENABLE_HIPFFT", self.spec.satisfies("device_backend=hipfft")),
+            self.define("KokkosFFT_ENABLE_ONEMKL", self.spec.satisfies("device_backend=onemkl")),
+        ]
+
+        if self.spec.satisfies("^kokkos+rocm"):
+            args.append(self.define("CMAKE_CXX_COMPILER", self["hip"].hipcc))
```
tpadioleau, line 73: Actually not sure we need that. It seems to work if we use `rocmcc`. I wonder if Kokkos should also be using it instead of bypassing the Spack compiler ?

tldahlgren (dismissed):
Confirmed the homepage, version sha256, and use of defined variants.

tldahlgren:
@cedricchevalier19 @tpadioleau @yasahi-hpc Is this PR ready to merge?

tpadioleau:
@tldahlgren Thank you for the approval but I think we still need to work on it, let me put it back to draft. Sorry for the inconvenience. We will let you know when it ready.

repos/spack_repo/builtin/packages/kokkos_fft/package.py
cedricchevalier19: For now it is ok to keep it.

repos/spack_repo/builtin/packages/kokkos_fft/package.py
cedricchevalier19: No I meant an internally multithreaded fftw from a serial loop.

cedricchevalier19 (approved):
I think most of the "issues" should be tackled later as they probably require some changes in the upstream project.

repos/spack_repo/builtin/packages/kokkos_fft/package.py
tpadioleau: To be discussed in the repo kokkos-fft

tpadioleau:
We were wondering if we should depend on the virtual package `fftw-api@3` as there are a few providers:
```
fftw-api@3:
acfl  amdfftw  armpl-gcc  cray-fftw  fftw@3:  fujitsu-fftw  intel-oneapi-mkl  nvpl-fft
```

One problem is that we don't seem to be able to ensure that an implementation will provide the expected parallelism. For example trying `spack spec -I fftw-api +openmp` I get the following error
```
==> Error: Spack concretizer internal error. Please submit a bug report and include the command, environment if applicable and the following error message.
    fftw-api+openmp is unsatisfiable
```

I don't know if it is possible to depend on a whitelist of packages providing `fftw-api` ?

cedricchevalier19:
> One problem is that we don't seem to be able to ensure that an implementation will provide the expected parallelism. For example trying `spack spec -I fftw-api +openmp` I get the following error
> 
> ```
> ==> Error: Spack concretizer internal error. Please submit a bug report and include the command, environment if applicable and the following error message.
>     fftw-api+openmp is unsatisfiable
> ```
> 
> I don't know if it is possible to depend on a whitelist of packages providing `fftw-api` ?

We'll look at this later. The recipe is ready to be merged as is.

tpadioleau:
Hi @tldahlgren, I think we are ready to merge, may we have an other review ? Thanks

tldahlgren (approved):

PR #81, opened 2025-06-06
rivet: allow building without analyses

veprbl:

### DIFF
--- a/repos/spack_repo/builtin/packages/rivet/package.py
+++ b/repos/spack_repo/builtin/packages/rivet/package.py
@@ -50,8 +50,8 @@ class Rivet(AutotoolsPackage):
     )
 
     with when("@4.1:"):
-        variant("plugin-match", default="none", description="List of Rivet analyses to be included")
-        variant("plugin-unmatch", default="none", description="List of Rivet analyses to be excluded")
+        variant("plugin-match", default="none", multi=True, description="List of Rivet analyses to be included")
+        variant("plugin-unmatch", default="none", multi=True, description="List of Rivet analyses to be excluded")
 
     depends_on("c", type="build")  # generated
     depends_on("cxx", type="build")  # generated
@@ -164,7 +164,7 @@ class Rivet(AutotoolsPackage):
 
         val = self.spec.variants["plugin-unmatch"].value
         if val != "none":
-            args += [f"--with-plugin-unmatch={val}"]
+            args += [f"--with-plugin-unmatch={' '.join(val)}"]
 
         args += ["--with-fastjet=" + self.spec["fastjet"].prefix]
         args += ["--with-yoda=" + self.spec["yoda"].prefix]
--- a/repos/spack_repo/builtin/packages/rivet/package.py
+++ b/repos/spack_repo/builtin/packages/rivet/package.py
@@ -50,8 +50,8 @@ class Rivet(AutotoolsPackage):
     )
 
     with when("@4.1:"):
-        variant("plugin-match", default=[], multi=True, description="List of Rivet analyses to be included")
-        variant("plugin-unmatch", default=[], multi=True, description="List of Rivet analyses to be excluded")
+        variant("plugin-match", default="none", multi=True, description="List of Rivet analyses to be included")
+        variant("plugin-unmatch", default="none", multi=True, description="List of Rivet analyses to be excluded")
 
     depends_on("c", type="build")  # generated
     depends_on("cxx", type="build")  # generated
@@ -158,13 +158,15 @@ class Rivet(AutotoolsPackage):
             args += ["--with-hepmc3=" + self.spec["hepmc3"].prefix]
             args += ["--with-hepmc3-libpath=" + self.spec["hepmc3"].libs.directories[0]]
 
-        val = self.spec.variants["plugin-match"].value
-        if val:
-            args += [f"--with-plugin-match={' '.join(val)}"]
+        if "plugin-match" in self.spec.variants:
+            val = self.spec.variants["plugin-match"].value
+            if val != "none":
+                args += [f"--with-plugin-match={' '.join(val)}"]
 
-        val = self.spec.variants["plugin-unmatch"].value
-        if val:
-            args += [f"--with-plugin-unmatch={' '.join(val)}"]
+        if "plugin-unmatch" in self.spec.variants:
+            val = self.spec.variants["plugin-unmatch"].value
+            if val != "none":
+                args += [f"--with-plugin-unmatch={' '.join(val)}"]
 
         args += ["--with-fastjet=" + self.spec["fastjet"].prefix]
         args += ["--with-yoda=" + self.spec["yoda"].prefix]
--- a/repos/spack_repo/builtin/packages/rivet/package.py
+++ b/repos/spack_repo/builtin/packages/rivet/package.py
@@ -50,8 +50,8 @@ class Rivet(AutotoolsPackage):
     )
 
     with when("@4.1:"):
-        variant("plugin-match", default="none", multi=True, description="List of Rivet analyses to be included")
-        variant("plugin-unmatch", default="none", multi=True, description="List of Rivet analyses to be excluded")
+        variant("plugin-match", default=[], multi=True, description="List of Rivet analyses to be included")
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/rivet/package.py
```diff
@@ -154,6 +156,9 @@ def configure_args(self):
             args += ["--with-hepmc3=" + self.spec["hepmc3"].prefix]
             args += ["--with-hepmc3-libpath=" + self.spec["hepmc3"].libs.directories[0]]
 
+        if not self.spec.satisfies("+analyses"):
+            args += ["--with-analyses=no"]
```
wdconinc: Does this option support anything other than yes/no? Should we approach this like `openloops` and allow for users to specify subsets of analyses?

veprbl: There is a way to provide a regex: https://gitlab.com/hepcedar/rivet/-/blob/release-4-1-x/configure.ac?ref_type=heads#L286
Not sure if that can be part of a spec.

wdconinc: ... as long as it doesn't use any of the spack sigils? No `^`.

`bufr` seems to have a variant `test_files` which is supposed to be a free-form path...

wdconinc: And `psblas` accepts free-form `ccopt`, `cxxopt`, etc.

wdconinc: Alternative is to provide a few pre-selected options, e.g. `[exp for plugin${exp} in analyses]`.

veprbl: I suspect comment about regexes is not true even. In the end, they use `filter` function of make. There is no mention of regexes in GNU documentation https://www.gnu.org/software/make/manual/html_node/Text-Functions.html

repos/spack_repo/builtin/packages/rivet/package.py
```diff
@@ -49,6 +49,9 @@ class Rivet(AutotoolsPackage):
         description="HepMC version to link against",
     )
 
+    variant("plugin-match", default="none", description="List of Rivet analyses to be included")
+    variant("plugin-unmatch", default="none", description="List of Rivet analyses to be excluded")
```
wdconinc: Functionality only added in 4.1.0; https://gitlab.com/hepcedar/rivet/-/merge_requests/957.
```suggestion
    with when("@4.1:"):
        variant("plugin-match", default="none", description="List of Rivet analyses to be included")
        variant("plugin-unmatch", default="none", description="List of Rivet analyses to be excluded")
```

repos/spack_repo/builtin/packages/rivet/package.py
```diff
@@ -154,6 +157,14 @@ def configure_args(self):
             args += ["--with-hepmc3=" + self.spec["hepmc3"].prefix]
             args += ["--with-hepmc3-libpath=" + self.spec["hepmc3"].libs.directories[0]]
 
+        val = self.spec.variants["plugin-match"].value
+        if val != "none":
+            args += [f"--with-plugin-match={val}"]
```
wdconinc: If the variant is defined with `multi=True` then it can take comma-separated arguments and we can turn them into the space-separated patterns that rivet expects with
```suggestion
            args += [f"--with-plugin-match={' '.join(val)}"]
```
Tested and works for `plugin-match=HERA,LEP,MC`.

repos/spack_repo/builtin/packages/rivet/package.py
```diff
@@ -49,6 +49,10 @@ class Rivet(AutotoolsPackage):
         description="HepMC version to link against",
     )
 
+    with when("@4.1:"):
+        variant("plugin-match", default="none", description="List of Rivet analyses to be included")
+        variant("plugin-unmatch", default="none", description="List of Rivet analyses to be excluded")
```
wdconinc:
```suggestion
        variant("plugin-match", default="none", multi=True, description="List of Rivet analyses to be included")
        variant("plugin-unmatch", default="none", multi=True, description="List of Rivet analyses to be excluded")
```

repos/spack_repo/builtin/packages/rivet/package.py
```diff
@@ -154,6 +158,14 @@ def configure_args(self):
             args += ["--with-hepmc3=" + self.spec["hepmc3"].prefix]
             args += ["--with-hepmc3-libpath=" + self.spec["hepmc3"].libs.directories[0]]
 
+        val = self.spec.variants["plugin-match"].value
+        if val != "none":
+            args += [f"--with-plugin-match={' '.join(val)}"]
+
+        val = self.spec.variants["plugin-unmatch"].value
+        if val != "none":
+            args += [f"--with-plugin-unmatch={val}"]
```
wdconinc:
```suggestion
            args += [f"--with-plugin-unmatch={' '.join(val)}"]
```

wdconinc (commented):
Still need `multi` to accept lists. I also don't know if the `none` default check is correctly handled.

repos/spack_repo/builtin/packages/rivet/package.py
```diff
@@ -154,6 +158,14 @@ def configure_args(self):
             args += ["--with-hepmc3=" + self.spec["hepmc3"].prefix]
             args += ["--with-hepmc3-libpath=" + self.spec["hepmc3"].libs.directories[0]]
 
+        val = self.spec.variants["plugin-match"].value
+        if val != "none":
```
veprbl:
```suggestion
        if val:
```
?

wdconinc: I think they are tuples, but I'm not sure.

repos/spack_repo/builtin/packages/rivet/package.py
```diff
@@ -154,6 +158,14 @@ def configure_args(self):
...
+            args += [f"--with-plugin-match={' '.join(val)}"]
+
+        val = self.spec.variants["plugin-unmatch"].value
+        if val != "none":
```
veprbl:
```suggestion
        if val:
```

repos/spack_repo/builtin/packages/rivet/package.py
```diff
@@ -154,6 +158,14 @@ def configure_args(self):
             args += ["--with-hepmc3=" + self.spec["hepmc3"].prefix]
             args += ["--with-hepmc3-libpath=" + self.spec["hepmc3"].libs.directories[0]]
 
+        val = self.spec.variants["plugin-match"].value
+        if val:
+            args += [f"--with-plugin-match={' '.join(val)}"]
+
+        val = self.spec.variants["plugin-unmatch"].value
+        if val:
+            args += [f"--with-plugin-unmatch={' '.join(val)}"]
```
wdconinc: Hmm, https://gitlab.spack.io/spack/spack-packages/-/jobs/17135055#L7106 indicates that this needs the variant existence check after all.
```suggestion
        if "plugin-match" in self.spec.variants:
            val = self.spec.variants["plugin-match"].value
            if val != "none":
                args += [f"--with-plugin-match={' '.join(val)}"]

        if "plugin-unmatch" in self.spec.variants:
            val = self.spec.variants["plugin-unmatch"].value
            if val != "none":
                args += [f"--with-plugin-unmatch={' '.join(val)}"]
```

wdconinc: The failure at https://gitlab.spack.io/spack/spack-packages/-/jobs/17135069#L1961 may be due to the default value of the new variant. It seems we are now explicitly passing `"none"` in https://gitlab.spack.io/spack/spack-packages/-/jobs/17135069#L1170.

It appears sufficient to change the `if val:` to `if val != "none":`, as applied in the suggestion above as well.

wdconinc: Apparently because `val` is a tuple:
```diff
-            if val != "none":
+            if not "none" in val:
```

veprbl: Yeah, I suspected that.

wdconinc (approved):
Thanks! Appears to work well in CI (and locally).

veprbl:
Thanks to you for getting this over the finish line!

PR #86, opened 2025-06-06
py-astropy: add v7.0.1 and dependencies

lgarrison:
> [!NOTE]
> This PR was migrated from **https://github.com/spack/spack/pull/49594**.

https://github.com/astropy/astropy/releases/tag/v7.0.1
https://github.com/astropy/astropy-iers-data/releases/tag/v0.2025.3.10.0.29.26

In the `+all` extras, I couldn't get `py-ipydatagrid` to build, so I left that out. For a few astropy deps (like `sortedcontainers`), I took the "older versions may work" note in `pyproject.toml` at face value. There's also a pre-commit test dep that I think we don't need.

The `--use-system-libraries` `install_option` that forced `py-pip@:23` is replaced with `ASTROPY_USE_SYSTEM_ALL` since `py-astropy@4`. This lets us build with modern Python.

### DIFF
--- a/repos/spack_repo/builtin/packages/py_astropy/package.py
+++ b/repos/spack_repo/builtin/packages/py_astropy/package.py
@@ -18,8 +18,9 @@ class PyAstropy(PythonPackage):
     pypi = "astropy/astropy-4.0.1.post1.tar.gz"
     git = "https://github.com/astropy/astropy.git"
 
-    license("BSD-3-Clause")
+    license("BSD-3-Clause", checked_by="lgarrison")
 
+    version("7.0.1", sha256="392feeb443b2437cd4c2e0641a65e0f15ba791e148e9b1e5ed7de7dfcb38e460")
     version("6.1.0", sha256="6c3b915f10b1576190730ddce45f6245f9927dda3de6e3f692db45779708950f")
     version("5.1", sha256="1db1b2c7eddfc773ca66fa33bd07b25d5b9c3b5eee2b934e0ca277fa5b1b7b7e")
     version(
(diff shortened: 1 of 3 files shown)


repos/spack_repo/builtin/packages/py_astropy/package.py
```diff
@@ -32,20 +33,29 @@ class PyAstropy(PythonPackage):
...
 
     # Required dependencies
+    depends_on("python@3.11:", when="@7.0.1:", type=("build", "run"))
     depends_on("python@3.10:", when="@6.1.0:", type=("build", "run"))
     depends_on("python@3.8:", when="@5.1:", type=("build", "run"))
     depends_on("py-setuptools", type="build")
     depends_on("py-cython@0.29.13:", type="build")
     depends_on("py-cython@0.29.30", when="@5.1:6.0", type="build")
-    depends_on("py-cython@3.0.0", when="@6.1.0:", type="build")
+    depends_on("py-cython@3.0.0", when="@6.1.0:6", type="build")
```
Sinan81: in pyproject.toml, cython dependency is indicated as `"cython>=3.0.0,<3.1.0",` for version @6.1:6. it would be nice test that package actually builds with all cython@3.0.x versions.
https://github.com/astropy/astropy/blob/v6.0.0/pyproject.toml

repos/spack_repo/builtin/packages/py_astropy/package.py
```diff
@@ -32,20 +33,29 @@ class PyAstropy(PythonPackage):
...
+    depends_on("py-extension-helpers@1", when="@7.0.1:", type="build")
+    depends_on("pkgconfig", type="build")
 
     # in newer pip versions --install-option does not exist
-    depends_on("py-pip@:23.0", type="build")
+    depends_on("py-pip@:23.0", when="@:4.0", type="build")
 
+    depends_on("py-astropy-iers-data@0.2025.1.31.12.41.4:", when="@7.0.1:", type=("build", "run"))
     depends_on("py-astropy-iers-data", when="@6:", type=("build", "run"))
+    depends_on("py-numpy@1.23.2:", when="@7.0.1:", type=("build", "run"))
```
Sinan81: should this be just a run dependency since above py-numpy@2: is indicated as build dependency. Having two different versions of numpy seems confusing at first glance, but I can see this is in accordance with the pyproject.toml file.

repos/spack_repo/builtin/packages/py_astropy/package.py
lgarrison: Fixed, thanks!

Sinan81:
fyi: there's a newer 7x version: 7.1.0

lgarrison:
Yeah; I did this PR 4 months ago when 7.0.1 was the newest. I probably won't have time to add 7.1.0, although if you want to add it (maybe as a PR into this branch to build off of it), that would be great.

Hopefully a Spack maintainer will have time to look at this soon, but astropy has a lot of dependencies so I get why it's daunting to review. External reviews like yours help, though, so thank you for that!

Sinan81:
>Yeah; I did this PR 4 months ago when 7.0.1 was the newest. I probably won't have time to add 7.1.0, although if you want to add it (maybe as a PR into this branch to build off of it), that would be great.

I can add it in another PR along with a few older versions.

repos/spack_repo/builtin/packages/py_astropy/package.py
```diff
@@ -56,52 +66,74 @@ class PyAstropy(PythonPackage):
...
+        depends_on("py-ipywidgets", when="@7.0.1:", type=("build", "run"))
+        depends_on("py-ipykernel", when="@7.0.1:", type=("build", "run"))
+        # depends_on("py-ipydatagrid", when="@7.0.1:", type=("build", "run"))
+        depends_on("py-ipython@8:", when="@7.0.1:", type=("build", "run"))
         depends_on("py-ipython@4.2:", when="@4.3:", type=("build", "run"))
         depends_on("py-ipython", type=("build", "run"))
-        depends_on("py-pytest@7:", when="@5.0.2:", type=("build", "run"))
-        depends_on("py-pytest", type=("build", "run"))
+        depends_on("py-pytest@7:", when="@5.0.2:6", type=("build", "run"))
+        depends_on("py-pytest", when="@:6", type=("build", "run"))
```
Sinan81, line 133: nit-pick: I find this a bit confusing. in pyproject.tom, pytest >= 7.3 is required for astropy@7.0.1 as indicated in test section. However, under docs section pytest >= 7.0 is indicated which's not relevent I think.
https://github.com/astropy/astropy/blob/v7.0.1/pyproject.toml

Sinan81 (approved):
looks good. double checked all the modified lines. tested that it builds fine. Not sure if my approval helps with merging but approving just in case.

repos/spack_repo/builtin/packages/py_astropy/package.py
lgarrison: I think this doesn't reflect the docs section, but rather the "all" section of the 6.1.0 `pyproject.toml` (specifically [this line](https://github.com/astropy/astropy/blob/e97756f0ecaa401e453745dd34d01980657ffa35/pyproject.toml#L92)).

In other words, astropy@6 had pytest in the "all" extras. astropy@7 does not, so I added version caps at `astropy@:6`.

alalazo (approved):
It would be good, in a following PR, to group dependencies and constraints per version of `py-astropy`, for example:
```python
with when("@7.0.1:"):
    depends_on("python@3.11:", type=("build", "run"))
    depends_on("py-cython@3, type="build")
    ...
```
This makes it easy to deprecate, and then remove, old versions (since it's just a block of code that gets deleted).

PR #88, opened 2025-06-06
README: add instructions for linking spack to spack-packages clone

adamjstewart:
Happy to hear wordsmithing or placement suggestions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,6 +10,13 @@ To contribute, simply make a pull request to this repository with your package c
 We run continuous integration on this repository to test builds of a large number of
 Spack packages.
 
+If you want to test your package changes locally before submitting a pull request,
+simply change Spack's default package repo from the default cache location to your
+local git clone:
+```
+spack repo set --destination /path/to/local/spack-packages builtin
+```
+
 If you are migrating your pull requests from
 [github.com/spack/spack](https://github.com/spack/spack), it is recommended to use the [the migration tool](https://github.com/spack/migrate-package-prs).
 
--- a/README.md
+++ b/README.md
@@ -11,11 +11,12 @@ We run continuous integration on this repository to test builds of a large numbe
 Spack packages.
 
 If you want to test your package changes locally before submitting a pull request,
-simply change Spack's default package repo from the default cache location to your
-local git clone:
+simply change Spack's default package repo from the default cache location to the full
+path to your local git clone:
 ```
 spack repo set --destination /path/to/local/spack-packages builtin
 ```
+`$spack` can be used to form a relative path to your Spack root directory.
 
 If you are migrating your pull requests from
 [github.com/spack/spack](https://github.com/spack/spack), it is recommended to use the [the migration tool](https://github.com/spack/migrate-package-prs).


README.md
```diff
@@ -10,6 +10,13 @@ To contribute, simply make a pull request to this repository with your package c
 We run continuous integration on this repository to test builds of a large number of
 Spack packages.
 
+If you want to test your package changes locally before submitting a pull request,
+simply change Spack's default package repo from the default cache location to your
+local git clone:
+```
+spack repo set --destination /path/to/local/spack-packages builtin
+```
+
```
Chrismarsh, line 20: I think you should note that full paths are required and that you can use `$spack`, e.g., `spack repo set --destination '$spack/../spack-packages' builtin`

adamjstewart: I used ~/spack-packages and that worked, FWIW

kwryankrattiger:
Closing this for a minute, CI infra is choking on the PR for some reason.

README.md
Chrismarsh: I did `spack repo set --destination spack-packages builtin` which silently failed

README.md
```diff
@@ -10,6 +10,14 @@ To contribute, simply make a pull request to this repository with your package c
 We run continuous integration on this repository to test builds of a large number of
 Spack packages.
 
+If you want to test your package changes locally before submitting a pull request,
+simply change Spack's default package repo from the default cache location to the full
+path to your local git clone:
+```
+spack repo set --destination /path/to/local/spack-packages builtin
```
johnwparent, line 17: I fully acknowledge this is bikeshedding, feel free to ignore.

Do we want suggest that folks re-map their `builtin` repo to their packages checkout simply for the purposes of a one off test? The instructions for how to do this is general are very valuable, it's the first thing I did post repo split, but within the framing of "quickly test this" I don't know that we want to suggest a user overwrite their builtin repo with their specific checkout of the packages repo and then be responsible for maintaining that, particularly if they're not expecting this to modify their primary repository. To us it's pretty obvious since you're specifying `builtin` but that does require some degree of familiarity that might trip new users.

Chrismarsh: I think this is a good point. Adding the instructions to undo it is probably valuable -- I actually don't know how to get it back to the default

kwryankrattiger:
Releated https://github.com/spack/spack-packages/pull/295

README.md
johnwparent: I actually think we should suggest adding the repo as "test-packages" or similar. 
That said, an undo would also be fine.

@Chrismarsh you can just delete your local config for the builtin repo, or from the command line something like  `spack config remove repos:builtin:destination` would work

tgamblin: @johnwparent how about propose that in a follow-on?

tgamblin (approved):

tgamblin:
force-merged b/c I don't think we should have to wait for CI for README changes.

README.md
johnwparent: @tgamblin sure, didn't realize this was urgent, didn't intend to hold anything up.

PR #90, opened 2025-06-06
charmpp: disable fortran when @8.0.0: %oneapi

teabagk7:
Disable fortran v8.0.0 for oneapi compiler

Duplicate of https://github.com/spack/spack/pull/50657
Generated by migrate-pkg-prs


matthiasdiener (approved):

teabagk7:
@spackbot help

RMeli (dismissed):

RMeli:
@teabagk7 AFAIK @spackbot is not yet enabled here. What did you want to do?

teabagk7:
> @teabagk7 AFAIK @spackbot is not yet enabled here. What did you want to do?

Yeah, @spackbot is not yet here :(
I wanted to tag maintainers who can review and merge. Matthias already approved the change, so I thought I’d let you know.

RMeli:
Ok, in that case, I'm happy to merge once CI is green. I'll enable auto-merge. Thanks!

repos/spack_repo/builtin/packages/charmpp/package.py
```diff
@@ -376,6 +376,8 @@ def install(self, spec, prefix):
         # been implemented in v8.0.0
         if self.spec.satisfies("@8.0.0: %aocc"):
             options.append("--disable-fortran")
+        if self.spec.satisfies("@8.0.0: %oneapi"):
```
tgamblin: Is it just when the Fortran compiler is oneAPI?  If so this could be `%fortran=oneapi`

teabagk7: What would be the difference in that case?

tgamblin: You can have mixed compilers.  `%oneapi` will match if `oneapi` is one of them. `%fortran=oneapi` will match when it's used to satisfy your `fortran` dependency.

Which makes me wonder whether you have a `fortran` dependency -- the `charmpp` package has:

```python
```

Which were auto-generated by us after v0.23 (see [release notes here](https://github.com/spack/spack/releases/tag/v0.23.0)).  If those are correct, you should remove the `#generated` comments.

Can you say more about this constraint and why you want to disallow it?  I'm confused why you want to disable `fortran` given that `oneapi` has a `fortran` compiler.

teabagk7: Indeed, you are correct. To support mixed compilers, I need to modify the package accordingly.
The package has a Fortran dependency for the interfaces. However, the build fails with clang-based compilers. Therefore, I intend to disable Fortran support due to this issue.
The most widely used application that will utilize charmpp is NAMD, and Fortran interfaces are not necessary for its functionality.

teabagk7: @tgamblin, the working option I found to support mixed compilers:
```python
if self.spec.satisfies("@8.0.0: ^[virtuals=fortran] intel-oneapi-compilers")
```

Could you please share your thoughts on this?

tgamblin: See https://github.com/spack/spack/pull/50858 -- this is the same as what I am proposing (just longer syntax).

Did you try `@8.0.0: ^fortran=intel-oneapi-compilers`?

teabagk7: > See [spack/spack#50858](https://github.com/spack/spack/pull/50858) -- this is the same as what I am proposing (just longer syntax).
> 
> Did you try `@8.0.0: ^fortran=intel-oneapi-compilers`?

This one doesn't work for some reason...

tgamblin: ok let's go with the one that works -- I will look into why the other might not

tgamblin: any chance you can run `spack spec --json charmpp` and send the output, or send along the `spack.lock` file from your environment?   Would like to see what spec is not satisfying here.

teabagk7: Sure, here are the lock file and spec output (had to rename lock file to upload)
[spack.lock.txt](https://github.com/user-attachments/files/20833201/spack.lock.txt)
[spack_spec_charmpp_oneapi.json](https://github.com/user-attachments/files/20833202/spack_spec_charmpp_oneapi.json)

teabagk7: @tgamblin Did you have a chance to look into it?

teabagk7: @tgamblin

repos/spack_repo/builtin/packages/charmpp/package.py
```diff
@@ -376,6 +376,8 @@ def install(self, spec, prefix):
         # been implemented in v8.0.0
         if self.spec.satisfies("@8.0.0: %aocc"):
             options.append("--disable-fortran")
+        if self.spec.satisfies("@8.0.0: ^[virtuals=fortran] intel-oneapi-compilers"):
```
alalazo: Isn't it a bit unexpected to concretize with a fortran compiler, and then disable fortran in the `install` method? Can we use directives, and enforce these constraints at concretization time?

What I'm thinking is:
```python
variant("fortran", default=False, description="Enable Fortran support")
depends_on("fortran", type="build", when="+fortran")
conflicts("%fortran=intel-oneapi-compilers", when="+fortran")
```

teabagk7: @alalazo , yes a bit unexpected 😄  It would be more logical to disable dependency if we disable fortran.
`--disable-fortran` appeared only in version 8.0.0. If “when” can be used to define the variant, it can be done.

@matthiasdiener , What are your thoughts on that? If you’re okay, I can modify the recipe accordingly.

alalazo: > If “when” can be used to define the variant, it can be done.

Yes, you can:
```python
variant("fortran", default=False, when="@8:", description="Enable Fortran support")
```

matthiasdiener (approved):

alalazo (approved):

PR #91, opened 2025-06-06
fix ncurses build and convert a lot of dependencies into variants

cosmicexplorer:
> [!NOTE]
> This PR was migrated from **https://github.com/spack/spack/pull/50564**.

# Problem
The C++ bindings for `ncurses` currently don't build on gcc 15: https://gitlab.archlinux.org/archlinux/packaging/packages/ncurses/-/issues/3.

Also, `gdbm` and `gettext` have several dependencies that aren't declared.

# Solution
- **FIX:** Apply the workaround to build ncurses C++ on gcc 15 from https://gitlab.archlinux.org/archlinux/packaging/packages/ncurses/-/issues/3.
- **NON-BREAKING:** Declare hidden dependencies in `gdbm` and `gettext`.

# Result
There are fewer hidden dependencies, and the ncurses C++ build works again.

### DIFF
--- a/repos/spack_repo/builtin/packages/cmake/package.py
+++ b/repos/spack_repo/builtin/packages/cmake/package.py
@@ -140,11 +140,7 @@ class Cmake(Package):
         default=False,
         description="Enables the generation of html and man page documentation",
     )
-    variant(
-        "ncurses",
-        default=sys.platform != "win32",
-        description="Enables the build of the ncurses gui",
-    )
+    variant("ncurses", default=True, description="Enables the build of the ncurses gui")
     variant("qtgui", default=False, description="Enables the build of the Qt GUI")
 
     # Revert the change that introduced a regression when parsing mpi link
@@ -221,6 +217,10 @@ class Cmake(Package):
                 depends_on("rhash", when="@3.8.0:")
                 depends_on("jsoncpp build_system=meson", when="@3.2:")
 
+    conflicts(
+        "+ncurses platform=windows",
+        msg="ncurses GUI cannot be built on windows",
+    )
     depends_on("ncurses", when="+ncurses")
 
     with when("+doc"):
--- a/repos/spack_repo/builtin/packages/gdbm/package.py
+++ b/repos/spack_repo/builtin/packages/gdbm/package.py
@@ -19,6 +19,8 @@ class Gdbm(AutotoolsPackage, GNUMirrorPackage):
 
     license("GPL-3.0-or-later")
 
+    maintainers("cosmicexplorer")
+
     version("1.25", sha256="d02db3c5926ed877f8817b81cd1f92f53ef74ca8c6db543fbba0271b34f393ec")
     version("1.24", sha256="695e9827fdf763513f133910bc7e6cfdb9187943a4fec943e57449723d2b8dbf")
     version("1.23", sha256="74b1081d21fff13ae4bd7c16e5d6e504a4c26f7cde1dca0d963a484174bbcacd")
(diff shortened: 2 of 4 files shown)


repos/spack_repo/builtin/packages/cmake/package.py
```diff
@@ -103,11 +103,7 @@ class Cmake(Package):
         default=False,
         description="Enables the generation of html and man page documentation",
     )
-    variant(
-        "ncurses",
-        default=sys.platform != "win32",
-        description="Enables the build of the ncurses gui",
-    )
+    variant("ncurses", default=False, description="Enables the build of the ncurses gui")
```
cosmicexplorer: @johnwparent raised a discussion thread here (https://github.com/spack/spack/pull/50564#discussion_r2129100681):

> > means that bootstrapping a new spack install involves building tons of ncurses and gui deps we don't actually need.
> 
> Can the bootstrapping deps propagate a `cmake~ncurses` constraint? I can never remember if the concretizer is able to force that choice that far down the DAG.

Yes! So for context, I am using spack's `re2` and `hyperscan` for my Rust packages https://docs.rs/re2 and https://docs.rs/vectorscan-async which use my `spack-rs` build script crate. For this case, it would be sufficient to explicitly set `cmake~ncurses` in the `Cargo.toml` metadata that `spack-rs` consumes: https://github.com/cosmicexplorer/spack-rs/blob/f388fea6c962fb07618df1027b8e6cdeb4be99f6/re2/sys/Cargo.toml#L24

```diff
-spec                    = "re2@2024-07-02+shared+pic ^ abseil-cpp+shared cxxstd=20"
+spec                    = "re2@2024-07-02+shared+pic ^ abseil-cpp+shared cxxstd=20 ^ cmake~ncurses"
```

*I actually need to update that crate now to take advantage of the new spack-packages split, but for now it works fine.*

> If not IMO improving bootstrapping speed/building less is worth changing the default here.

Thanks so much for this feedback! ^_^ ! I am still going to propose a new feature for "variant default flavors" which will allow us to support multiple use cases, but I've taken this advice and set `cmake~ncurses` by default.

> > That's a case where spack is being used less interactively, so the cmake it bootstraps should imo not be pulling in all the really useful tools cmake has for interactive usage.
> 
> That's why I asked, I know for Windows the use case is rarely anything but a build requirement, and a lot of folks are using Spack in headless contexts, so maybe this is just a good default in general.

I'm *also* trying to remove the `readline` dependency from `gdbm` by default because it's not useful for headless use cases, but there is some issue which caused gitlab CI to fail. I will confirm that minimizing default dependencies still fails our gitlab CI with this new PR, and then I will see if flipping the defaults back makes it pass. Since I want this ncurses fix in, I will figure that out in a separate PR.

cosmicexplorer:
gitlab CI doesn't want to report a result, so I'm taking a different tack and removing the changes to variant defaults for now. I'm going to make a separate PR which changes the variant defaults to reduce the chance this will cause cascading breakage in the future.

cosmicexplorer:
Created #163 as a follow-up which will disable deps by default so this one can focus on:
(1) getting ncurses C++ to build
(2) codifying some undeclared dependencies

tldahlgren:
Added missing maintainers.

There are conflicts for `cmake` and `ncurses` that need to be resolved.

repos/spack_repo/builtin/packages/gdbm/package.py
```diff
@@ -37,20 +39,48 @@ class Gdbm(AutotoolsPackage, GNUMirrorPackage):
...
         # GDBM uses some non-standard GNU extensions,
         # enabled with -D_GNU_SOURCE.  See:
         #   https://patchwork.ozlabs.org/patch/771300/
         #   https://stackoverflow.com/questions/5582211
         #   https://www.gnu.org/software/automake/manual/html_node/Flag-Variables-Ordering.html
-        return ["--enable-libgdbm-compat", "CPPFLAGS=-D_GNU_SOURCE"]
+        opts.extend(["--enable-libgdbm-compat", "CPPFLAGS=-D_GNU_SOURCE"])
+
+        if "~readline" in self.spec:
+            opts.append("--without-readline")
```
alecbcs, line 71:
```suggestion
        if self.spec.satisfies("~readline"):
            opts.append("--without-readline")
```

Better to use the `satisfies()` directive in place of `x in self.spec`.

repos/spack_repo/builtin/packages/gdbm/package.py
```diff
@@ -37,20 +39,48 @@ class Gdbm(AutotoolsPackage, GNUMirrorPackage):
...
+
+        if "+nls" in self.spec:
+            libiconv_pfx = self.spec["libiconv"].prefix
```
alecbcs, line 74:
```suggestion
        if self.spec.satisfies("+nls"):
            libiconv_pfx = self.spec["libiconv"].prefix
```

Here too.

repos/spack_repo/builtin/packages/ncurses/package.py
```diff
@@ -111,7 +115,8 @@ def flag_handler(self, name, flags):
         if name == "cflags":
             flags.append(self["c"].pic_flag)
         elif name == "cxxflags":
-            flags.append(self["cxx"].pic_flag)
+            if "+cxx" in self.spec:
+                flags.append(self["cxx"].pic_flag)
```
alecbcs, line 119:
```suggestion
            if self.spec.satisfies("+cxx"):
                flags.append(self["cxx"].pic_flag)
```
Here too.

repos/spack_repo/builtin/packages/ncurses/package.py
```diff
@@ -121,21 +126,34 @@ def flag_handler(self, name, flags):
         # ncurses@:6.0 uses dynamic exception specifications not allowed in c++17
         if self.spec.satisfies("@:5"):
             if name == "cxxflags":
-                flags.append(self["cxx"].standard_flag(language="cxx", standard="14"))
+                if "+cxx" in self.spec:
+                    flags.append(self["cxx"].standard_flag(language="cxx", standard="14"))
```
alecbcs, line 130:
```suggestion
                if self.spec.satisfies("+cxx"):
                    flags.append(self["cxx"].standard_flag(language="cxx", standard="14"))
```

repos/spack_repo/builtin/packages/ncurses/package.py
```diff
@@ -121,21 +126,34 @@ def flag_handler(self, name, flags):
...
             "--disable-overwrite",
         ]
 
+        if "+shared" in spec:
+            opts.append("--with-shared")
+            if "+cxx" in spec:
+                opts.append("--with-cxx-shared")
+
+        if "~cxx" in spec:
+            opts.extend(["--without-cxx", "--without-cxx-binding"])
```
alecbcs, line 149:
```suggestion
        if spec.satisfies("+shared"):
            opts.append("--with-shared")
            if spec.satisfies("+shared"):
                opts.append("--with-cxx-shared")

        if spec.satisfies("~cxx"):
            opts.extend(["--without-cxx", "--without-cxx-binding"])
```

alecbcs (changes requested):
Couple of quick suggestions regarding the `spec.satisfies()` method.

PR #93, opened 2025-06-06
[Draft] deepmd: new package with deps

mtaillefumier:
> [!NOTE]
> This PR was migrated from **https://github.com/spack/spack/pull/49366**.

- This PR adds the package file to build the python package deepmd and the deepmd library. It does not support building the modules needed for gromacs and lammps yet. 
- cuda@12.8 and gcc@14 generate a lot of failure unless the C++ standard is set to c++14. Many packages still do not use the `CMAKE_CXX_STANDARD ` syntax which makes these changes tedious.

### DIFF
--- a/repos/spack_repo/builtin/packages/deepmd_lib/package.py
+++ b/repos/spack_repo/builtin/packages/deepmd_lib/package.py
@@ -1,0 +1,146 @@
+# Copyright 2013-2024 Lawrence Livermore National Security, LLC and other
+# Spack Project Developers. See the top-level COPYRIGHT file for details.
+#
+# SPDX-License-Identifier: (Apache-2.0 OR MIT)
+
+from spack_repo.builtin.build_systems.cmake import CMakePackage, generator
+from spack_repo.builtin.build_systems.cuda import CudaPackage
+from spack_repo.builtin.build_systems.rocm import ROCmPackage
+from spack.package import *
+
+
+class DeepmdLib(CudaPackage, ROCmPackage, CMakePackage):
+    """DeePMD-kit is a package written in Python/C++, designed to minimize the effort
+    required to build deep learning-based models of interatomic potential energy
+    and force field and to perform molecular dynamics (MD)"""
+
+    homepage = "https://docs.deepmodeling.com/projects/deepmd/en/stable/index.html#"
+    git = "https://github.com/deepmodeling/deepmd-kit.git"
+    url = "https://github.com/deepmodeling/deepmd-kit/archive/refs/tags/v2.2.11.tar.gz"
+
+    license("LGPL-3.0-only")
+
+    maintainers("mtaillefumier")
+    version("3.1.0", sha256="45f13df9ed011438d139a7f61416b8d7940f63c47fcde53180bfccd60c9d22ee")
+    version("3.0.3", sha256="f562ee5c07e77274425cf6d19e28ffcb07017a6f07c56fe82b5c9974dd0fb3b6")
+    version("3.0.2", sha256="b828d3a44730ea852505abbdb24ea5b556f2bf8b16de5a9c76018ed1ced7121b")
+    version("3.0.1", sha256="e842edbc2714bc948ce708c411e5fed751e67c88d5c493c2978f11c849027dca")
+    version("3.0.0", sha256="4df1091ce90dbea87734a20c6d826b8cbe80ac44646cb592c2e8586be319023c")
+    version("2.2.11", sha256="d22893a08c2556c5cb29682378105849cf672545c91ee52b10a97da6e9075ac3")
+
+    variant(
+        "tensorflow", default=True, description="Enable tensorflow support (original ML backend)"
+    )
+    variant(
+        "pytorch",
+        default=False,
+        description="Enable pytorch support (starting v3.0.0)",
+        when="@3.0:",
+    )
+    variant("cuda", default=False, description="Enable cuda support")
+    variant("rocm", default=False, description="Enable rocm support")
+    variant("gromacs", default=False, description="Enable gromacs plugins")
+    variant("horovod", default=True, description="Enable horovod support")
+    variant("jax", default=False, description="Enable JaX support", when="@3.0:")
+    variant("fp64", default=True, description="Enable double precision ops", when="+tensorflow")
+    variant("fp64", default=False, description="Enable double precision ops", when="+pytorch")
+    # deepmd library uses cmake as a build system but deepmd itself uses pip.
+
+    depends_on("c")
+    depends_on("cxx")
+
+    # Historical dependencies
+    depends_on("py-setuptools", type="build")
+    depends_on("py-tensorflow@2.16:", when="+tensorflow")
+    depends_on("py-tensorflow+mpi", when="+tensorflow")
+    depends_on("py-torch", when="+pytorch")
+    # depends_on("py-ase")
+    # depends_on("py-scipy")
+    # depends_on("py-numpy")
+    # depends_on("py-pyyaml")
+    # depends_on("py-args")
+    # depends_on("py-pyproject-metadata")
+    # depends_on("py-python-hostlist@1.21:")
+    # depends_on("py-typing-extensions", when="^python@:3.8")
+    # depends_on("py-importlib-metadata", when="^python@:3.8")
+    # depends_on("py-sphinx-argparse")
+    # depends_on("py-pygments")
+    # depends_on("py-sphinxcontrib-bibtex")
+    # depends_on("py-scikit-build-core")
+    # depends_on("py-setuptools-scm")
+    # depends_on("py-scikit-build")
+    # depends_on("py-hatch-fancy-pypi-readme")
+    # depends_on("py-pip", type="build")
+    # depends_on("python@3.10:")
+    # depends_on("py-h5py")
+    depends_on("py-jax", when="+jax")
+
+    # horovod needs some special settings
+    # depends_on("py-horovod controllers=mpi", when="+horovod")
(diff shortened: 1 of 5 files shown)


tldahlgren (changes requested):
This would be easier to review once you move the new packages to the new `builtin` directory.

alalazo:
@tldahlgren Thanks for catching this. We need an audit to prevent getting new packages with the old layout of the repo.

tldahlgren:
> @tldahlgren Thanks for catching this. We need an audit to prevent getting new packages with the old layout of the repo.

It's been added (and refined).

mtaillefumier:
@tldahlgren @alalazo, sorry for the delay. I moved the two packages to the new layout. 
Maybe merging the two packages in one would be a better solution as the dependencies are identical. The library only needs py-torch, tensorflow etc while the python package requires some python dependencies on the top.

tldahlgren:
> @tldahlgren @alalazo, sorry for the delay. I moved the two packages to the new layout. Maybe merging the two packages in one would be a better solution as the dependencies are identical. The library only needs py-torch, tensorflow etc while the python package requires some python dependencies on the top.

Np. Such packages are often separate so I would be inclined to not merge them.

repos/spack_repo/builtin/packages/deepmd_lib/package.py
```diff
@@ -0,0 +1,146 @@
+# Copyright 2013-2024 Lawrence Livermore National Security, LLC and other
+# Spack Project Developers. See the top-level COPYRIGHT file for details.
+#
+# SPDX-License-Identifier: (Apache-2.0 OR MIT)
+
+from spack_repo.builtin.build_systems.cmake import CMakePackage, generator
+from spack_repo.builtin.build_systems.cuda import CudaPackage
+from spack_repo.builtin.build_systems.rocm import ROCmPackage
+from spack.package import *
```
tldahlgren, line 10:
```suggestion

from spack.package import *
```

repos/spack_repo/builtin/packages/py_deepmd/package.py
```diff
@@ -0,0 +1,136 @@
+# Copyright Spack Project Developers. See COPYRIGHT file for details.
+#
+# SPDX-License-Identifier: (Apache-2.0 OR MIT)
+
+
+#from spack_repo.builtin.build_systems.cmake import CMakePackage, generator
+from spack_repo.builtin.build_systems.cuda import CudaPackage
+from spack_repo.builtin.build_systems.python import PythonPackage
+from spack_repo.builtin.build_systems.rocm import ROCmPackage
+from spack.package import *
```
tldahlgren, line 10:
```suggestion

from spack.package import *
```

repos/spack_repo/builtin/packages/py_deepmd/package.py
```diff
@@ -0,0 +1,136 @@
+# Copyright Spack Project Developers. See COPYRIGHT file for details.
+#
+# SPDX-License-Identifier: (Apache-2.0 OR MIT)
+
+
+#from spack_repo.builtin.build_systems.cmake import CMakePackage, generator
```
tldahlgren:
```suggestion
```

tldahlgren (commented):
Confirmed the homepages and version sha256s.

tldahlgren:
😞 I jumped the gun when I took the liberty of fixing the initial round of style issues:

```
==> running flake8
repos/spack_repo/builtin/packages/deepmd_lib/package.py:6: [F401] 'spack_repo.builtin.build_systems.cmake.generator' imported but unused
```

tldahlgren (changes requested):
After looking more closely at the packages I can see why you indicated uncertainty about having one or multiple spack package files. IMO  there are a number of considerations remaining, including:

- Why use GitHub over PyPI (https://pypi.org/project/deepmd-kit/)?
   https://spack.readthedocs.io/en/latest/build_systems/pythonpackage.html#id2
- If keeping two packages, where do the dependencies belong?
  (And remove comments from the package where they are not needed.)
 - Do the dependencies match those of the `pyproject.toml` for each version in terms type (https://spack.readthedocs.io/en/latest/packaging_guide_creation.html#dependency-types) as well as the packages and constraints?
 
 I can also see a need to organize more of the dependencies by variant (e.g., `+jax`) as you're doing with `cuda`, for example.
  
I say these things because I see discrepancies between the dependencies here and those in especially the `requires`, `dependencies` and variant-specific tool sections of the `pyproject.toml` files, i.e.,:

- https://github.com/deepmodeling/deepmd-kit/blob/v3.1.0/pyproject.toml
- https://github.com/deepmodeling/deepmd-kit/blob/v3.0.3/pyproject.toml
- https://github.com/deepmodeling/deepmd-kit/blob/v3.0.2/pyproject.toml
- https://github.com/deepmodeling/deepmd-kit/blob/v3.0.1/pyproject.toml
- https://github.com/deepmodeling/deepmd-kit/blob/v3.0.0/pyproject.toml
- https://github.com/deepmodeling/deepmd-kit/blob/v2.2.11/pyproject.toml

mtaillefumier:
The developers use github for development that why I used it in the package files. The python dependencies seem pretty similar between the different versions of deepmd so the best course of action is to go through the list of requirements and check if any specific version is required. I will work on this in the coming days.

repos/spack_repo/builtin/packages/py_dargs/package.py
```diff
@@ -0,0 +1,34 @@
...
+    version("0.4.10", sha256="7dc9a0b213bed71a281997dbf2be6612e988caadc566b4aec1bf253243a22ea0")
+    version("0.4.9", sha256="cbd68ed97fd49ba7b5701ea980ad495c64e41db2785357f2666490531995c407")
+    version("0.4.8", sha256="95f3085dc912af4bcefbd38ced60717f023b7b59ff9ce6aadabfaceca2f11243")
+    version("0.4.7", sha256="3132ea3f39dc1aaa342b3f045d262ed81c1d52fa2e817bae60839cf67c0d152b")
+    version("0.4.6", sha256="864dfa7a61f483d9a025f1f843afd6e9eefe728fdc9a06f991bcef92f0018bed")
+    version("0.4.5", sha256="46ddd9aba01be1c185b693cabee481c8a3d95b2d5f2a4f31d190f12c95fa23f3")
+    version("0.4.4", sha256="9cc8e9e36bfb401de19ca6e0de5cc5ab76871dacb9071c3f6ff078ca9aec38c0")
+    version("0.4.3", sha256="2c0b2f7ea2eef932f2a4d0a3a5801f00854bf2b2b69cb3c415ace0bd224afd7f")
+    version("0.4.2", sha256="285684743feec7d28cad31a090949c97a549fa94c9eea1c1d65fd8138f2458df")
+    version("0.4.1", sha256="6f04fa6a67b2a7dcb27b36896cd8062ee19f95cc74ab3ca2a08c3985da68a495")
```
bernhardkaindl, line 31: There should be no reason why we need so many version on a new package. Only the latest is enough. Dependencies get compilicated over time, but we don't have to start off already with lots of versions:
```suggestion
```

repos/spack_repo/builtin/packages/py_deepmd/package.py
```diff
@@ -0,0 +1,142 @@
...
+
+    license("LGPL-3.0-only")
+    maintainers("mtaillefumier")
+
+    version("3.1.0", sha256="45f13df9ed011438d139a7f61416b8d7940f63c47fcde53180bfccd60c9d22ee")
+    version("3.0.3", sha256="f562ee5c07e77274425cf6d19e28ffcb07017a6f07c56fe82b5c9974dd0fb3b6")
+    version("3.0.2", sha256="b828d3a44730ea852505abbdb24ea5b556f2bf8b16de5a9c76018ed1ced7121b")
+    version("3.0.1", sha256="e842edbc2714bc948ce708c411e5fed751e67c88d5c493c2978f11c849027dca")
+    version("3.0.0", sha256="4df1091ce90dbea87734a20c6d826b8cbe80ac44646cb592c2e8586be319023c")
+    version("2.2.11", sha256="d22893a08c2556c5cb29682378105849cf672545c91ee52b10a97da6e9075ac3")
```
bernhardkaindl, line 30: There should be no reason why we need so many version on a new package. Only the latest is enough. Dependencies get compilicated over time, but we don't have to start off already with lots of versions:
```suggestion
```

repos/spack_repo/builtin/packages/deepmd_lib/package.py
```diff
@@ -0,0 +1,121 @@
...
+
+    license("LGPL-3.0-only")
+
+    maintainers("mtaillefumier")
+    version("3.1.0", sha256="45f13df9ed011438d139a7f61416b8d7940f63c47fcde53180bfccd60c9d22ee")
+    version("3.0.3", sha256="f562ee5c07e77274425cf6d19e28ffcb07017a6f07c56fe82b5c9974dd0fb3b6")
+    version("3.0.2", sha256="b828d3a44730ea852505abbdb24ea5b556f2bf8b16de5a9c76018ed1ced7121b")
+    version("3.0.1", sha256="e842edbc2714bc948ce708c411e5fed751e67c88d5c493c2978f11c849027dca")
+    version("3.0.0", sha256="4df1091ce90dbea87734a20c6d826b8cbe80ac44646cb592c2e8586be319023c")
+    version("2.2.11", sha256="d22893a08c2556c5cb29682378105849cf672545c91ee52b10a97da6e9075ac3")
```
bernhardkaindl, line 30: There should be no reason why we need so many version on a new package. Only the latest is enough. Dependencies get compilicated over time, but we don't have to start off already with lots of versions:
```suggestion
```

bernhardkaindl (changes requested):

PR #109, opened 2025-06-06
root: Change the default value of the runtime_cxxmodules option

jmcarcell:
> [!NOTE]
> This PR was migrated from **https://github.com/spack/spack/pull/50432**.

Since ROOT 6.20 the default (https://github.com/root-project/root/pull/4452) is to have `runtime_cxxmodules` ON and we have it OFF by default with no way of configuring it, so I propose to follow the default value and set it to ON. I have found that in our stack having this set to off causes some slow downs in some IO operations (https://github.com/key4hep/key4hep-spack/issues/743). The PCH option was removed in https://github.com/root-project/root/commit/898171f5a4619ac07245d70b7f8943da067516ba so I propose not to set it.

### DIFF
--- a/repos/spack_repo/builtin/packages/root/package.py
+++ b/repos/spack_repo/builtin/packages/root/package.py
@@ -656,7 +656,7 @@ class Root(CMakePackage):
             define("libcxx", False),
             define("roottest", False),
             define_from_variant("rpath"),
-            define("runtime_cxxmodules", False),
+            define("runtime_cxxmodules", False if self.spec.satisfies("@:6.18") else True),
             define("shared", True),
             define("soversion", True),
             define("testing", self.run_tests),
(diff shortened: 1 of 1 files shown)


wdconinc (commented):
A similar trail of breadcrumbs as in the linked issue (https://github.com/key4hep/key4hep-spack/issues/743) led me here too. Aligning the defaults makes a lot of sense. I would, however, suggest that we actually turn this into a proper variant with the new default, and that would allow users to configure it. More detail on the tradeoffs is at https://github.com/root-project/root/blob/master/README/README.CXXMODULES.md (including here only for reference for when someone looks at why this variant was added).

wdconinc:
@jmcarcell I rebased this to see if this still fails as in https://gitlab.spack.io/spack/spack-packages/-/jobs/17042977.

jmcarcell:
@spackbot run pipeline

jmcarcell:
CI seems to be failing, I can see the same issue at least another PR updated recently

jmcarcell:
@spackbot run pipeline

jmcarcell:
I think this should be ready now

wdconinc (approved):

PR #110, opened 2025-06-06
madgraph5_aMC: add variant to install models

jmcarcell:
> [!NOTE]
> This PR was migrated from **https://github.com/spack/spack/pull/49568**.

The new variant `models` will install the models that are selected. This has the advantages that the models don't have to be downloaded every time (although they are quite small) a new user uses them, and they appear when running `display modellist`. Currently, if the madgraph installation folder is not writable I don't seem to be able to download and install them locally, that is a different problem that would also be solved by making the models available in the installation folder.

I ran this and it seems to work. I get weird ownership of the newly installed model (under `<install>/models) when running as root:

```
drwxr-sr-x. 10 root root  4.0K Mar 18 22:12 .
drwxr-sr-x. 17 root root  4.0K Mar 18 22:11 ..
drwxr-sr-x.  2 root root  4.0K Mar 18 22:11 MSSM_SLHA2
...
drwxr-sr-x.  4  501 games 4.0K Mar 18 22:12 loop_qcd_qed_sm_Gmu
```
But this also happens when running the command to install manually. Perhaps a better approach would be to unpack the models directly from Spack without calling madgraph to install them, but I don't know if madgraph does anything else besides unpacking them.


jmcarcell:
I had another look at this @wdconinc. I found that the default value is actually a tuple containing "None", so I changed the check to that, otherwise it wouldn't work with the default models. In addition, since compilation is done with `+pythia`, a dependency on `cxx` is needed. I was able to install it with `+pythia` and using one model in the list on Alma 9 and Ubuntu 24.

repos/spack_repo/builtin/packages/madgraph5amc/package.py
```diff
@@ -26,6 +26,7 @@ class Madgraph5amc(MakefilePackage):
     timeout = {"timeout": 60}
 
     with default_args(fetch_options=timeout):
+        version("3.6.2", sha256="1dde78f865278d8b9a59ca608076f495ad82cebe75f10469faba97210ddcdb6f")
```
tldahlgren: I am unable to confirm the sha256 and get `no remote versions` error when I try.

There is a newer version shown on the web site and it's sha256 is "207f990f7fb09164a54c5481e7fc6a41f000265b8ab3fdd77a844f505b06b169".

tldahlgren (changes requested):
See issue with version.

repos/spack_repo/builtin/packages/madgraph5amc/package.py
tldahlgren: I realize the check passed initially, but you can see from https://launchpad.net/mg5amcnlo/3.0/3.6.x that it is no longer listed.

tldahlgren:
`Update branch` to see if the `verify-checksum` check will pick up the issue. If it is able to confirmed the checksum then I'll merge even though I cannot checksum the package with either `curl` or `urllib`.

tldahlgren:
> `Update branch` to see if the `verify-checksum` check will pick up the issue. If it is able to confirmed the checksum then I'll merge even though I cannot checksum the package with either `curl` or `urllib`.

```
Run . spack-core/share/spack/setup-env.sh
==> Error: Could not fetch any versions for madgraph5amc
Error: Process completed with exit code 1.
```

jmcarcell:
This is very specific for us, and maybe it's not so important to upstream these changes.

PR #115, opened 2025-06-06
ucx: fixes for gcc15 and conflict

Chrismarsh:
`gcc@15` [makes a breaking change](https://gcc.gnu.org/gcc-15/changes.html#c) by defaulting to `-std=gnu23`. ucx and friends don't compile (summarized here https://github.com/openucx/ucx/issues/10589)

This PR fixes `kerberos` and `libpciaccess` by ensuring c++17 is used. However, there does not appear to be a fix for ucx yet (probably needs to wait until (1.19.x), so this conflicts against `gcc@15` for `@:1.18`

Maintainers:
@hppritcha


Chrismarsh:
@spackbot fix style

repos/spack_repo/builtin/packages/krb5/package.py
```diff
@@ -127,4 +127,11 @@ def configure_args(self):
...
             flags.append("-lintl")
-        return inject_flags(name, flags)
+
+        if name == "cflags":
+            if self.spec.satisfies("@:1.21.3 %gcc@15:"):
+                # gcc@15: is -std=gnu23 by default and
+                # up to at least 1.21.3 doesn't compile
+                flags.append("-std=gnu17")
+
+        return (flags, None, None)
```
johnwparent, line 137: Do we still need to call `inject_flags` here?

Chrismarsh: I didn't think so based on how I read [the docs](https://spack.readthedocs.io/en/latest/packaging_guide.html#compiler-flags). I understood it to be one can use `flag_handler` or `inject_flags`, and my changes swapped it to the more generic `flag_handler`. Indeed, I have not run into `flag_handler` calling `inject_flags` like this before. So maybe I have misunderstood?

johnwparent: Previously the behavior was still using `flag_handler` and calling `inject_flags` from the flag handler (to compose the flag list into the tuple you return). This achieves the same thing, maybe _need to_ was poorly phrased, want to for semantic purposes would probably have been a better phrasing. Either way, not a blocking comment.

johnwparent (approved):

hppritcha (approved):

PR #121, opened 2025-06-06
Python: Add Windows 'py' launcher to the PATH

johnwparent:
Python (on Windows) installs a dynamic python launcher into
the root of its install prefix
Some packages expect that launcher to be in the path as they
rely on that as a generic entrypoint to python.

This PR adds said launcher to the PATH for dependent
packages of Python

### DIFF
--- a/repos/spack_repo/builtin/packages/python/package.py
+++ b/repos/spack_repo/builtin/packages/python/package.py
@@ -1271,10 +1271,10 @@ print(json.dumps(config))
         """Set PYTHONPATH to include the site-packages directory for the
         extension and any other python extensions it depends on.
         """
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/python/package.py
```diff
@@ -1271,6 +1271,7 @@ def setup_dependent_build_environment(
         """Set PYTHONPATH to include the site-packages directory for the
         extension and any other python extensions it depends on.
         """
+        env.prepend_path("PATH", self.prefix)
```
scheibelp: Can these be made Windows-specific: in this case move it into the below `if` (for the other one, add an `if`).

scheibelp (approved):

PR #124, opened 2025-06-07
dd4hep: allow +utliityapps with root ~x ~opengl

wdconinc:
This PR removes the requirement in `dd4hep` that `+utlityapps` means `^root +x +opengl`. That was removed in https://github.com/AIDASoft/DD4hep/pull/1426, where the utility apps that depend on ROOT::Gui and ROOT::Eve are now only built when that is available in ROOT (i.e. `+utlityapps` and `+ddeve` are composable).


tldahlgren:
@drbenmorgan @jmcarcell @vvolkl Any objections to merging this PR?

vvolkl (approved):
None from my side.

PR #125, opened 2025-06-07
ghostscript: add variant to make gtk+ optional.

yizeyi18:
There are packages like texlive which only use ghostscript as a postscript2pdf converter, in which case gtk+ is not necessary.

This PR adds variant `gtk` to ghostscript, making gtkplus a optional dependency, to make packages like texlive could be installed without gtkplus. In order to maintain consistency with existing ghostscript install, the default value of `gtk` is set to be `True`.

### DIFF
--- a/repos/spack_repo/builtin/packages/ghostscript/package.py
+++ b/repos/spack_repo/builtin/packages/ghostscript/package.py
@@ -57,6 +57,7 @@ class Ghostscript(AutotoolsPackage):
 
     # https://www.ghostscript.com/ocr.html
     variant("tesseract", default=False, description="Use the Tesseract library for OCR")
+    variant("gtk", default=True, description="Enable gtk+ device for screen output")
 
     depends_on("c", type="build")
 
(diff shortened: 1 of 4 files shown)


repos/spack_repo/builtin/packages/ghostscript/package.py
```diff
@@ -70,7 +71,9 @@ class Ghostscript(AutotoolsPackage):
     depends_on("libtiff")
     depends_on("zlib-api")
     depends_on("libxext")
-    depends_on("gtkplus")
+    depends_on("gtkplus", type="link", when="+gtk")
+    depends_on("dbus", type="link")
+    depends_on("libiconv", type="link")
```
tldahlgren, line 76: Why are `dbus` and `libiconv` dependencies being added? Do they only apply when `~gtk` and or for a specific version on?

yizeyi18: Both `dbus` and `libiconv` are requested by `ghostscript` by default, though can be switched off by configure option.
 
With `gtkplus` as a link dependency(which also have `dbus` and `libiconv` as its dependency), the two packages are *implicit* dependencies of `ghostscript`. While `gtkplus` being removed, the two *implicit* dependencies are also removed, making builds fail. So I added the two dependencies here explicitly. 
Line 140 is added for similar reason: without a lib linked `libiconv`, ghostscript's configure script cannot automatically generate link flags of `libiconv` correctly, making builds fail at link stage. This configure option can help the script recognize `libiconv` to be linked.

To disable the two packages, configure options below are required:
 - `dbus`: `--disable-dbus`
 - `libiconv`: `--with-libiconv=no`

tldahlgren (approved):

yizeyi18:
A mistake was done in this PR: `fontconfig`, an optional dependent of ghostscript and provided as a *implicit* link dependent by gtk+, was made undetermined. A new PR was submitted to fix this.

PR #118, opened 2025-06-06
.gitattributes: Checkout LF (mostly)

johnwparent:
Set the gitattributes of this repo to ensure all checkouts of package recipes (and everything else but Windows shell files) are done with line feed (lf) endings.

PR #127, opened 2025-06-07
Revert ".gitattributes: Checkout LF (mostly)"

alalazo:
Reverts spack/spack-packages#118

This is likely causing issues in both `spack/spack` and `spack-packages` Gitlab pipelines, due to patches that are now checksummed differently:
- https://gitlab.spack.io/spack/spack-packages/-/jobs/17001967
- https://github.com/spack/spack/actions/runs/15505359709/job/43659490389

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,4 +1,0 @@
-* text eol=lf
-*.bat text eol=crlf
-*.pwsh text eol=crlf
-*.ps1 text eol=crlf


tgamblin (approved):

alalazo:
This also raises the question of how we can catch, in PRs to `spack-packages`, possible disruptions of `spack/spack`.

tgamblin:
I think the answers would be:
1. pin spack-packages in `spack/spack`, and only bump it deliberately, but this is slow and requires work.
2. Add some bootstrap CI checks here -- I think this makes more sense. Since Spack bootstrap still relies on `builtin` right now, we need to make sure we're not blowing that up.

PR #128, opened 2025-06-07
py_onnxruntime: add new versions and update recipe

jmcarcell:
This PR:
- Add versions 1.21.0 and 1.20.2, both of which I tested
- Change the version of abseil-cpp that is needed, that was changed in 1.20
- Add patches for the eigen hashes, that changed after gitlab implemented a new compression algorithm (see https://github.com/microsoft/onnxruntime/issues/24861 and related issues there)
- Add a patch for compatibility with the latest protobuf, since `ReleaseCleared()` has been removed in recent versions (see https://github.com/microsoft/onnxruntime/issues/21308). The patch may apply for earlier versions but we don't have them in the recipe.
- Add a dependency on binutils. This is necessary when the system assembler is not new enough, because then there is an issue with unknown instructions (https://github.com/microsoft/onnxruntime/issues/24275). This happens on Alma 9, where the system `as` has version `2.35`. I read somewhere that for these instructions the minimum is 2.36 but I'm not completely sure.


jmcarcell:
This is ready and CI passes except for one job where an unrelated failure happened before building anything.

tldahlgren (approved):
I'd like to hear from @adamjstewart or @afzpatel before merging, at least if they have time within the next few days.

repos/spack_repo/builtin/packages/py_onnxruntime/package.py
```diff
@@ -39,12 +41,16 @@ class PyOnnxruntime(CMakePackage, PythonExtension, ROCmPackage, CudaPackage):
     depends_on("c", type="build")  # generated
     depends_on("cxx", type="build")  # generated
 
+    depends_on("binutils@2.36:", type="build")
```
adamjstewart, line 44: We don't normally add binutils as a dependency since most Spack packages depend on binutils in one way or another. Unless this policy has changed.

tldahlgren: True, but the description states:

> * Add a dependency on binutils. This is necessary when the system assembler is not new enough, because then there is an issue with unknown instructions ([[Build] Building v1.21.0: unsupported instruction 'vpdpbusds' microsoft/onnxruntime#24275](https://github.com/microsoft/onnxruntime/issues/24275)). This happens on Alma 9, where the system `as` has version `2.35`. I read somewhere that for these instructions the minimum is 2.36 but I'm not completely sure.

jmcarcell: I don't know how to build on Alma 9 if it's not adding this dependency, if there is any other way I'm happy to add that.

PR #130, opened 2025-06-07
mmg: Enable Lagrangian mesh morphing

jcortial-safran:
> [!NOTE]
> This PR was migrated from **https://github.com/spack/spack/pull/39490**.

I would like to enable the Lagrangian mesh morphing feature in MMG3D. It requires to add two additional packages.


tldahlgren (approved):
Confirmed homepages and version sha256.

PR #136, opened 2025-06-07
mariadb: add diffutils as a dependency

jmcarcell:
Trying to build mariadb in a system without `diff` installed does not work. It complains that there is a different ABI between two files that it is diffing, of course what fails is `diff` not being found.

This is the error:

```
  >> 1694    CMake Error at /tmp/root/spack-stage/sp
             ack-stage-mariadb-11.3.2-hpik53di427np2
             qbaghi26jge7cmfpfe/spack-src/cmake/do_a
             bi_check.cmake:84 (MESSAGE):
     1695      ABI check found difference between
     1696      /tmp/root/spack-stage/spack-stage-mar
             iadb-11.3.2-hpik53di427np2qbaghi26jge7c
             mfpfe/spack-src/include/mysql/plugin_au
             dit.h.pp
     1697      and
     1698      /tmp/root/spack-stage/spack-stage-mar
             iadb-11.3.2-hpik53di427np2qbaghi26jge7c
             mfpfe/spack-build-hpik53d/abi_check.out
```


alalazo (approved):
Failures in pipelines seem unrelated

PR #138, opened 2025-06-07
py-rpds-py: depends_on rust type=build

wdconinc:
The docs of `py-rpds-py` claim that `rust` is only a [build-time dependency](https://pypi.org/project/rpds-py/):
> Note that if you install rpds-py from source, you will need a Rust toolchain installed, as it is a build-time dependency.


tldahlgren:
This needs conflicts to be resolved.

bernhardkaindl (approved):

PR #143, opened 2025-06-08
Mfem cmake

jacobmerson:
> [!NOTE]
> This PR was migrated from **https://github.com/spack/spack/pull/49790**.

Depends on https://github.com/spack/spack-packages/pull/496 (per https://github.com/spack/spack-packages/pull/143#issuecomment-3059528310)

We are developing tools that are built using CMake and link against mfem. We have found it is much more convenient to have a version of mfem that's built with CMake as integration into the downstream project is much more streamlined.

This is an initial implementation of splitting the spack package into the `Package` and `Builder` components to support both build systems. So far, I have not tested all variants. I'm hoping that the mfem team has CI/CD for their spack builds and can add `build_system=cmake` to the mix to validate everything is working as expected.


tldahlgren:
@acfisher @cyrush @jedbrown @markcmiller86 @rblake-llnl @white238  As maintainers on one or more of the affected packages, would you be interested in commenting on these changes?

repos/spack_repo/builtin/packages/mfem/package.py
```diff
@@ -632,18 +736,18 @@ def find_optional_library(name, prefix):
 
         # Determine C++ standard to use:
         cxxstd = None
-        if self.spec.satisfies("@4.0.0:"):
+        if spec.satisfies("@4.0.0:"):
```
markcmiller86, line 810: I wonder if it would be simpler here and perhaps a more correct expression of mfem's requirements to *default* to `cxxstd = 14` and set something other than that only under specific conditions. I mean, it is still ok as is but am wondering if simplifying would help.

repos/spack_repo/builtin/packages/mfem/package.py
```diff
@@ -518,6 +532,103 @@ class Mfem(Package, CudaPackage, ROCmPackage):
...
+        extra_install_tests = [self.examples_src_dir, self.examples_data_dir]
+        cache_extra_test_sources(self.pkg, extra_install_tests)
+
+
+# without splitting the cache_test_sources into AnyBuilder,
+# cache_extra_test_sources was not available in the GenericBuilder
+# took the idea from the superlu package
+class GenericBuilder(AnyBuilder, GenericBuilder):
+    #
+    # Note: Although MFEM does support CMake configuration, MFEM
```
markcmiller86: Is this comment regarding vanilla GNU make still relevant now that we appear to be switching mfem to use its CMake build?

repos/spack_repo/builtin/packages/visit/package.py
```diff
@@ -196,6 +196,7 @@ class Visit(CMakePackage):
     depends_on("mfem@4.4:", when="+mfem")
     depends_on("mfem+shared+exceptions+fms+conduit", when="+mfem")
     depends_on("libfms@0.2:", when="+mfem")
+    conflicts("^mfem build_system=cmake")
```
markcmiller86: I see this `conflicts()` line in `visit/package.py` and a few others. I guess it is telling Spack, if someone builds mfem with CMake, the resulting MFEM install cannot be used within these other packages? If so, I am wondering what difficulty is arising as a result? I mean, is MFEM's CMake install creating cmake content that is causing other packages to digest it incorrectly? If the rational could be elaborated a tad here, that might be useful.

cameronrutherford: I think @jacobmerson can clarify, but from what I can tell, MFEM has two builds that are supported: the CMake build, and the makefile based build. Both builds export different build objects, and do depending on how the parent project handles those builds, they make / break accordingly.

markcmiller86: Ah, ok. That makes sense. So, in the case of VisIt, the `package.py` appears to have logic that is dependent upon MFEM having been built and installed using the vanilla GNU make approach.

markcmiller86: So, maybe this is NOT the time/place to inquire about this but in looking at VisIt's `package.py` file, I honestly can't see anything there that would suggest a dependency in the build approach used for MFEM. I searched for all instances of `mfem`, and all I see are `variant()`, `depends_on()` statements and a logic block, conditional on `+mfem` being the `spec`, that defines variables relevant to VisIt's build.

So, I honestly can't see any reason why VisIt's `package.py` file is somehow dependent on MFEM's build/installation. Do I have the wrong expectations of Spack here?

cameronrutherford:
```suggestion
    depends_on("mfem+shared+exceptions+fms+conduit build_system=cmake", when="+mfem")
    depends_on("libfms@0.2:", when="+mfem")
```

Perhaps this is clearer?

jacobmerson:
@markcmiller86 I want to comment on my strategy for this pull request then I will address your specific comments/questions.

Currently, MFEM supports two build systems in parallel CMake and Makefiles. As far as I understand there is not 100% coverage of the MFEM features in the CMake build system. I'm not a MFEM developer, so I can't directly comment on if the Makefile based builds will eventually be phased out or not.

For this pull request, my strategy to add the CMake based build with the minimal amount of changes to the package logic to hopefully make it easier for the review/acceptance process.

repos/spack_repo/builtin/packages/mfem/package.py
jacobmerson: My preference is to keep the standard detection logic the same in this pull request as what's currently in the Makefile based builds and only have changes related to the addition of the CMake based builds. If you would prefer that this gets updated in this pull request let me know.

repos/spack_repo/builtin/packages/mfem/package.py
jacobmerson: I think someone from the MFEM team should comment on the accuracy of this. This pull request doesn't switch the build system in the spack package, but just adds an option to use `CMake` if desired.

repos/spack_repo/builtin/packages/visit/package.py
jacobmerson: @cameronrutherford is correct the packages that I added `conflict` statements to utilize build artifacts that are generated by the `Makefiles`. This constraint that the `Makefile` based artifacts are implicit in the `Makefiles` of the downstream packages like `visit`. The addition of the `conflicts` statement makes this constraint explicit for the `spack` package. Previously, `spack` could only build using the `Makefile` based approach so it was not necessary to have any explicit specification of the build system in the package.

jacobmerson: I think this changes the logic since it is forcing the package to use the `CMake` build rather than forcing it to not use the `CMake` build.

cyrush: VisIt can certainly build against a CMake built mfem -- we do this for windows. 

Same with Ascent, for our continuously tested builds on macOS, linux, and Windows we only use mfem's cmake build.

That said, I expect there are things that could complicate this in the spack package. 

It's fine to note the conflicts.

cameronrutherford: gah yes. sorry this would be better, or however it is you would express dependency logic on the GNU Make based build:

```suggestion
    depends_on("mfem+shared+exceptions+fms+conduit build_system=make", when="+mfem")
    depends_on("libfms@0.2:", when="+mfem")
```

v-dobrev:
I'd like to get the PR that adds MFEM v4.8, #496, merged before this PR. That's why I've been postponing the review of this PR for some time now.

repos/spack_repo/builtin/packages/visit/package.py
v-dobrev: I'm not convinced this conflict is actually needed for the `visit` package. @jacobmerson, did you actually try building `visit` with CMake built MFEM?

repos/spack_repo/builtin/packages/mfem/package.py
v-dobrev: This comment will probably need to be updated -- I can do that when I understand better what this PR does in terms of e.g. what variants are supported by the new CMake support which will require some testing to confirm what works and what doesn't.

tldahlgren:
FYI. Conflicts need to be resolved for `mfem`.

repos/spack_repo/builtin/packages/mfem/package.py
hughcars: mfem is now using c++17 since the introduction of "dfem"

jacobmerson:
I'm working on resolving the conflicts and trying to track down if I can remove some of the conflicts statements in the downstream packages.

repos/spack_repo/builtin/packages/visit/package.py
jacobmerson: I finally had a chance to test this out in a clean VM. looks like visit works with the `CMake` mfem built through spack. I'll remove the conflict.

tldahlgren:
FYI. There's a conflict that needs to be resolved.

hughcars:
@jacobmerson this would be really enabling for combining spack development with Palace!

jacobmerson:
@hughcars I can re-resolve the conflicts.

It would be good to have some indication from the MFEM team what they need for this to get merged.

tldahlgren:
There's also a conflict to be resolved with the `dray` package.

cyrush:
@tldahlgren  dray is no longer developed as a stand alone package, and wont be using newer mfem versions. Given this -- I support whatever the best path is to not block this pr.

tldahlgren:
There is a conflict (`dray`) that needs to be resolved.

jacobmerson:
@tldahlgren finally had a chance to update and remove the conflicts.

tldahlgren (dismissed):

tldahlgren:
Does anyone in reviewers list have an objection to this PR being merged?

v-dobrev:
Please do not merge yet. I'll review this soon -- I'll do it right after I'm done with the PR to add and test the new v4.9. (For now I just have the initial commit for that pushed: https://github.com/spack/spack-packages/tree/mfem-4.9.)

cameronrutherford:
I'll defer my review to @hughcars or @Sbozzolo since they're more involved at the moment.

PinguProgrammer:
Hi! Any updates on this?

jacobmerson:
Not on my side. I have been waiting for an indication from the mfem team that they are in a state they are willing to merge this.

@PinguProgrammer I can update the package again if it's useful to you.

PinguProgrammer:
@jacobmerson , that would be great if you have bandwidth and is easy. My goal was to attempt to compile on windows with cmake (since the current default is with autotools). Realistically, I should be able to try that with what you have already contributed -- I just haven't gotten around to trying it.

Fyi, it seems the mfem team did merge in update for v4.9 so they maybe open to it now? cc: @v-dobrev 

Thank you regardless!

v-dobrev:
One thing to note: the CMake build system installs both makefiles, `config.mk` and `test.mk` similar to the GNU make build system. For this reason, downstream packages that use these `.mk` files (like `glvis`) should still work fine with `mfem` built with CMake.

I'm making some of these adjustments locally and testing them and the overall CMake support. I'll push an update in the near future.

v-dobrev:
Hi @jacobmerson,

Is it okay with you if I push changes directly in your branch? Alternatively, I can create a branch (without a PR) from which you can cherry-pick changes I suggest.

jacobmerson:
@v-dobrev feel free to push to my branch. Let me know if you need additional permissions.

v-dobrev:
@spackbot fix style

v-dobrev:
@spackbot re-run pipeline

repos/spack_repo/builtin/packages/mfem/package.py
v-dobrev: Addressed in a0db032.

v-dobrev:
Update: I planning to run some tests with GPU builds and also see if more of the `conflicts("^mfem build_system=cmake")` are really necessary.

repos/spack_repo/builtin/packages/mfem/package.py
```diff
@@ -1510,3 +1539,100 @@ def all_headers(self, root_spec):
...
+            self.define_from_variant("MFEM_USE_LIBUNWIND", "libunwind"),
+            self.define_from_variant("MFEM_USE_FMS", "fms"),
+            self.define_from_variant("MFEM_USE_GINKGO", "ginkgo"),
+            self.define_from_variant("MFEM_USE_HIOP", "hiop"),
+            self.define_from_variant("MFEM_ENABLE_EXAMPLES", "examples"),
+            self.define_from_variant("MFEM_ENABLE_MINIAPPS", "miniapps"),
+            self.define_from_variant("MFEM_USE_EXCEPTIONS", "exceptions"),
+            self.define_from_variant("MFEM_PRECISION", "precision"),
+            self.define("MFEM_ENABLE_TESTING", True),
+        ]
```
hughcars, line 1580: The CMake builder needs to map `+cuda`, `+rocm`, and `+enzyme` to `MFEM_USE_CUDA`, `MFEM_USE_HIP`, and `MFEM_USE_ENZYME`. None of these appear in `cmake_args()`, and MFEM defaults all three to `OFF`, so the resulting build does not match the requested spec. The recipe also needs to call the CUDA/HIP architecture helpers explicitly; `CMakeBuilder.std_args()` does not forward `cuda_arch` or `amdgpu_target` automatically.

```suggestion
            self.define_from_variant("MFEM_USE_FMS", "fms"),
            self.define_from_variant("MFEM_USE_GINKGO", "ginkgo"),
            self.define_from_variant("MFEM_USE_HIOP", "hiop"),
            self.define_from_variant("MFEM_USE_CUDA", "cuda"),
            self.define_from_variant("MFEM_USE_HIP", "rocm"),
            self.define_from_variant("MFEM_USE_ENZYME", "enzyme"),
            self.define_from_variant("MFEM_ENABLE_EXAMPLES", "examples"),
            self.define_from_variant("MFEM_ENABLE_MINIAPPS", "miniapps"),
            self.define_from_variant("MFEM_USE_EXCEPTIONS", "exceptions"),
            self.define_from_variant("MFEM_PRECISION", "precision"),
            self.define("MFEM_ENABLE_TESTING", True),
        ]
        if "+cuda" in self.spec:
            args.append(CMakeBuilder.define_cuda_architectures(self))
        if "+rocm" in self.spec:
            args.append(CMakeBuilder.define_hip_architectures(self))
```

repos/spack_repo/builtin/packages/mfem/package.py
```diff
@@ -25,6 +26,8 @@ class Mfem(Package, CudaPackage, ROCmPackage):
 
     test_requires_compiler = True
 
+    build_system("generic", "cmake", default="generic")
```
hughcars, line 29: CMake needs to be rejected for MFEM releases before 3.3. Both 3.1 and 3.2 currently concretize with this build system, but upstream first added `CMakeLists.txt` in v3.3.

```console
$ spack install --until cmake -j4 'mfem@3.2 build_system=cmake ~zlib'
CMake Error: The source directory ".../spack-src" does not appear to contain CMakeLists.txt.
```

```suggestion
    build_system("generic", "cmake", default="generic")
    conflicts("build_system=cmake", when="@:3.2")
```

repos/spack_repo/builtin/packages/mfem/package.py
```diff
@@ -1510,3 +1539,100 @@ def all_headers(self, root_spec):
...
+            self.define_from_variant("MFEM_USE_MUMPS", "mumps"),
+            self.define_from_variant("MFEM_USE_SLEPC", "slepc"),
+            self.define_from_variant("MFEM_USE_SUNDIALS", "sundials"),
+            self.define_from_variant("MFEM_USE_PUMI", "pumi"),
+            self.define_from_variant("MFEM_USE_GSLIB", "gslib"),
+            self.define_from_variant("MFEM_USE_MPFR", "mpfr"),
+            self.define_from_variant("MFEM_USE_LAPACK", "lapack"),
+            self.define_from_variant("MFEM_USE_NETCDF", "netcdf"),
+            self.define_from_variant("MFEM_USE_CONDUIT", "conduit"),
+            self.define_from_variant("MFEM_USE_ZLIB", "zlib"),
```
hughcars, line 1569: This needs the same version split as the generic builder: `MFEM_USE_GZSTREAM` through MFEM 4.0 and `MFEM_USE_ZLIB` from 4.1. As written, `mfem@4.0.0 build_system=cmake +zlib` passes an option that the source never reads, leaving gzstream at its `OFF` default.

```console
$ git grep MFEM_USE_ZLIB v4.0 -- CMakeLists.txt config
# no matches
$ git grep MFEM_USE_GZSTREAM v4.0 -- CMakeLists.txt config
CMakeLists.txt:if (MFEM_USE_GZSTREAM)
config/defaults.cmake:option(MFEM_USE_GZSTREAM ... OFF)
```

```suggestion
            self.define_from_variant(
                "MFEM_USE_ZLIB"
                if self.spec.satisfies("@4.1.0:")
                else "MFEM_USE_GZSTREAM",
                "zlib",
            ),
```

repos/spack_repo/builtin/packages/mfem/package.py
```diff
@@ -583,6 +598,106 @@ class Mfem(Package, CudaPackage, ROCmPackage):
...
+        for d in dirs:
+            f = join_path(d, "test.mk")
+            if os.access(f, os.R_OK):
+                return FileList(f)
+        return FileList(find(self.prefix, "test.mk", recursive=True))
+
+    @property
+    def xlinker(self):
+        using_nvcc = "+cuda" in self.spec and "+enzyme" not in self.spec
+        return "-Wl," if not using_nvcc else "-Xlinker="
```
hughcars, line 668: This `patch()` needs to remain on `Mfem`, not `GenericBuilder`. Spack calls `pkg.do_patch()` before creating the builder, so the builder method is currently skipped for both build systems.

```console
$ spack patch 'mfem@4.1.0 build_system=generic'
==> No patches needed for mfem
$ src=$(spack location -c 'mfem@4.1.0 build_system=generic')
$ xxd -l3 "$src/fem/gslib.hpp"
00000000: efbb bf
```

```suggestion
        return "-Wl," if not using_nvcc else "-Xlinker="

    @when("@4.1.0")
    def patch(self):
        files_with_bom = [
            "fem/gslib.hpp",
            "fem/gslib.cpp",
            "linalg/hiop.hpp",
            "miniapps/gslib/field-diff.cpp",
            "miniapps/gslib/findpts.cpp",
            "miniapps/gslib/pfindpts.cpp",
        ]
        bom = "\ufeff"
        for f in files_with_bom:
            filter_file(bom, "", f)
```

repos/spack_repo/builtin/packages/mfem/package.py
```diff
@@ -1394,70 +1492,6 @@ def patch(self):
         for f in files_with_bom:
```
hughcars, line 1492: Remove this builder copy after moving the hook back to `Mfem`; it is not part of Spack's patch lifecycle.

repos/spack_repo/builtin/packages/mfem/package.py
```diff
@@ -583,6 +598,106 @@ class Mfem(Package, CudaPackage, ROCmPackage):
...
+        This property can be accessed using pkg["mfem"].test_mk.
+        In version 3.3.2 and newer, the location of test.mk is also defined
+        inside config.mk, variable MFEM_TEST_MK.
+        """
+        dirs = [self.prefix, self.prefix.share.mfem]
+        for d in dirs:
+            f = join_path(d, "test.mk")
+            if os.access(f, os.R_OK):
+                return FileList(f)
+        return FileList(find(self.prefix, "test.mk", recursive=True))
```
hughcars, line 663: `test_ex10` and its source/data attributes need to remain on `Mfem`. Spack discovers stand-alone tests on the package class, not the selected builder, so the current move drops the test entirely.

```console
$ spack python -c 'import spack.install_test, spack.repo; c=spack.repo.PATH.get_pkg_class("mfem"); print(spack.install_test.test_function_names(c))'
[]
```

```suggestion
        return FileList(find(self.prefix, "test.mk", recursive=True))

    examples_src_dir = "examples"
    examples_data_dir = "data"

    def test_ex10(self):
        """build and run ex10(p)"""
        test_dir = join_path(self.test_suite.current_test_cache_dir, self.examples_src_dir)
        mesh = join_path("..", self.examples_data_dir, "beam-quad.mesh")
        test_exe = "ex10p" if ("+mpi" in self.spec) else "ex10"

        with working_dir(test_dir):
            make = which("make", required=True)
            make(f"CONFIG_MK={self.config_mk}", test_exe, parallel=False)

            ex10 = which(test_exe, required=True)
            ex10("--mesh", mesh)
```

repos/spack_repo/builtin/packages/mfem/package.py
```diff
@@ -583,6 +598,106 @@ class Mfem(Package, CudaPackage, ROCmPackage):
...
+
+class AnyBuilder(BaseBuilder):
+    @run_after("install")
+    def cache_test_sources(self):
+        """Copy the example source files after the package is installed to an
+        install test subdirectory for use during `spack test run`."""
+        # Clean the 'examples' directory -- at least one example is always built
+        # and we do not want to cache executables.
+        make("examples/clean", parallel=False)
+        extra_install_tests = [self.examples_src_dir, self.examples_data_dir]
```
hughcars, line 684: Use the package-owned test paths after moving the attributes back to `Mfem`.

```suggestion
        extra_install_tests = [self.pkg.examples_src_dir, self.pkg.examples_data_dir]
```

repos/spack_repo/builtin/packages/mfem/package.py
```diff
@@ -1351,23 +1458,14 @@ def install(self, spec, prefix):
...
-        make("examples/clean", parallel=False)
-        cache_extra_test_sources(self, [self.examples_src_dir, self.examples_data_dir])
-
     def test_ex10(self):
         """build and run ex10(p)"""
         # MFEM has many examples to serve as a suitable smoke check. ex10
         # was chosen arbitrarily among the examples that work both with
         # MPI and without it
-        test_dir = join_path(self.test_suite.current_test_cache_dir, self.examples_src_dir)
+        test_dir = join_path(self.test_suite.current_test_cache_dir, self.pkg.examples_src_dir)
```
hughcars, line 1466: Remove the builder-owned `test_ex10` and path attributes after restoring them on `Mfem`; Spack does not discover stand-alone tests here.

repos/spack_repo/builtin/packages/mfem/package.py
```diff
@@ -1510,3 +1539,100 @@ def all_headers(self, root_spec):
             except NoHeadersError:
                 pass
         return all_hdrs
+
+
+class CMakeBuilder(CMakeBuilder):
```
hughcars, line 1544: The CMake builder needs its own callback to cache `examples/` and `data/`. It does not inherit `AnyBuilder`, and the generic `make("examples/clean")` step does not belong in an out-of-source CMake build.

```suggestion
class CMakeBuilder(CMakeBuilder):
    @run_after("install")
    def cache_test_sources(self):
        cache_extra_test_sources(
            self.pkg,
            [self.pkg.examples_src_dir, self.pkg.examples_data_dir],
        )
```

repos/spack_repo/builtin/packages/mfem/package.py
```diff
@@ -583,6 +598,106 @@ class Mfem(Package, CudaPackage, ROCmPackage):
...
+        """Export the main mfem header, mfem.hpp."""
+        hdrs = HeaderList(find(self.prefix.include, "mfem.hpp", recursive=False))
+        return hdrs or None
+
+    @property
+    def libs(self):
+        """Export the mfem library file."""
+        libs = find_libraries(
+            "libmfem", root=self.prefix.lib, shared=("+shared" in self.spec), recursive=False
+        )
```
hughcars, line 636: This query misses the native Windows CMake output. With the MSVC-compatible ABI, target `mfem` installs `lib/mfem.lib`; the current call instead searches for `lib/libmfem.lib` when static and `lib/libmfem.dll` when shared. The shared query also needs `runtime=False` to select the import library.

```suggestion
        name = (
            "mfem"
            if self.spec.satisfies("platform=windows build_system=cmake")
            else "libmfem"
        )
        libs = find_libraries(
            name,
            root=self.prefix.lib,
            shared=("+shared" in self.spec),
            recursive=False,
            runtime=False,
        )
```

hughcars (commented):
Me and my assistant tried a few things and have some suggestions, a few are backwards compatibility nits, but some are for flags that aren't being propagated and for locating options in more idiomatic places. Direct suggestions given where helpful.

jacobmerson:
@v-dobrev @hughcars edits make a lot of sense to me. Not sure how much progress you made. Do you want to apply these changes?

I can work through them if you want to push whatever other changes you wanted to make.

v-dobrev:
> @v-dobrev @hughcars edits make a lot of sense to me. Not sure how much progress you made. Do you want to apply these changes?
> 
> I can work through them if you want to push whatever other changes you wanted to make.

Thanks for the review and suggestions, @hughcars!

@jacobmerson, some of the suggested changes are to support mfem versions < 4.9.0. I don't think we really need to support the CMake build system with such older MFEM versions -- to me it makes sense to start with the current release, v4.9.0, and later update things as they evolve.

Other than that, if you want (and have time), go ahead and merge what you think makes sense.

I still want to test some GPU builds and maybe the proposed changes will help with that (I saw some changes related to cuda and hip).

PR #150, opened 2025-06-09
gha: set builtin path as part of checkout

haampie:
* In GitHub actions, the variable `${CI_PROJECT_DIR}` is not set, and as a result considered a literal path component.
* Every action seems to reinvent the wheel

So, simplify by making the checkout action take care of updating the path of builtin.

On the Spack side we should probably restore hard errors when no package repository is available.


alalazo:
Yes, having hard errors will prevent getting a random commit in CI, if the path we point to is wrong.

PR #156, opened 2025-06-09
flecsi: add v2.4.0

rbberger:


opensdh (approved):
I would exchange these two commits (or squash them) to avoid having a "2.4.0 exists but allows `networking=none`" state, but otherwise this is fine.

alalazo (approved):

PR #157, opened 2025-06-09
Add missing `depends_on("c")` for cutlass, libcxxwrap_julia, and py-pyside2

melven:
Both use CMake and fail to configure/compile without a working C compiler.


tldahlgren (approved):

tldahlgren:
Looks like a CI issue with a stack. Trying `Update branch` to see if that helps.

PR #163, opened 2025-06-09
disable ncurses default deps

cosmicexplorer:
*This is on top of #91. See the `+7/-8` diff [here](https://github.com/cosmicexplorer/spack-packages/compare/spack-pr-50564...disable-ncurses-default-deps?expand=1).*

# Problem
See the issue fixed in #91. `ncurses` is being added as a default dependency in many places which are not necessarily used in the terminal. If this sort of breakage occurs again, it could break quite a few packages that shouldn't need to worry about ncurses being broken.

See discussion at https://github.com/spack/spack-packages/pull/91#discussion_r2132453038, continued from https://github.com/spack/spack/pull/50564#discussion_r2129100681.

# Solution
- Disable ncurses C++ builds by default.
- Remove ncurses as a default dependency from other places like gettext and cmake which are not necessarily going to be used in an interactive terminal.
- Disable `readline` dependency from `gdbm` by default.

# Result
If this sort of ncurses breakage occurs again, it should be less likely to propagate to packages that shouldn't be depending on it.

### DIFF
--- a/repos/spack_repo/builtin/packages/cmake/package.py
+++ b/repos/spack_repo/builtin/packages/cmake/package.py
@@ -140,7 +140,7 @@ class Cmake(Package):
         default=False,
         description="Enables the generation of html and man page documentation",
     )
-    variant("ncurses", default=True, description="Enables the build of the ncurses gui")
+    variant("ncurses", default=False, description="Enables the build of the ncurses gui")
     variant("qtgui", default=False, description="Enables the build of the Qt GUI")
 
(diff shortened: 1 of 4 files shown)


tldahlgren:
Added missing maintainers.

tldahlgren:
There are conflicts that need to be resolved in `cmake` and `ncurses`.

johnwparent (approved):
Looks good from the CMake side

PR #168, opened 2025-06-09
qt: explicitly disable vulkan

AlexanderRichert-NOAA:
This PR explicitly disables vulkan in qt. In principle, this doesn't change anything, as no vulkan variant is defined, and it is (supposedly) off by default. I have however found that even for the same qt hash, the vulkan build option gets mysteriously enabled during some builds, and explicitly setting `-no-vulkan` resolves the issue (based on 10+ test rebuilds).

### DIFF
--- a/repos/spack_repo/builtin/packages/qt/package.py
+++ b/repos/spack_repo/builtin/packages/qt/package.py
@@ -617,6 +617,8 @@ class Qt(Package):
             use_spack_dep("freetype")
             if spec.satisfies("platform=linux") or spec.satisfies("platform=freebsd"):
                 config_args.append("-fontconfig")
+            # Avoid sporadic vkconvenience bug by explicitly disabling vulkan
+            config_args.append("-no-vulkan")
         else:
             config_args.append("-no-freetype")
             config_args.append("-no-gui")


repos/spack_repo/builtin/packages/qt/package.py
```diff
@@ -617,6 +617,8 @@ def common_config_args(self):
             use_spack_dep("freetype")
             if spec.satisfies("platform=linux") or spec.satisfies("platform=freebsd"):
                 config_args.append("-fontconfig")
+            # Avoid sporadic vkconvenience bug by explicitly disabling vulkan
+            config_args.append("-no-vulkan")
```
tldahlgren, line 621: Are you sure the option isn't `-no-feature-vulkan`?

tldahlgren (commented):
I have a question about the option.

repos/spack_repo/builtin/packages/qt/package.py
tldahlgren: I realize you've tested this based on your comment but I couldn't find the `-no-vulkan` option; hence, my question.

AlexanderRichert-NOAA: Yeah, good question, I got a bit turned around on this before since they don't give separate documentation for each and every feature/flag. I've found examples of both `-no-vulkan` and `-no-feature-vulkan` (though more of the former). According to qtbase/config_help.txt, some features have multiple synonyms ('-no-x' _and_ '-no-feature-x') so that may be the case here. I'd also be fine with using both flags :upside_down_face:

tldahlgren: Up to you. Shall I merge it as-is?

AlexanderRichert-NOAA: I'm happy with it as it is, thanks-- it's easy enough to add the other flag later if anybody else encounters this issue.

tldahlgren (approved):

PR #21, opened 2025-06-06
Try to build Kathleen's Spack PR

tgamblin:
Trying to build a full pipeline from a Spack in a fork.

See https://github.com/spack/spack/pull/50797

PR #170, opened 2025-06-10
CI: Make it easier to point to experimental refs

kwryankrattiger:
Add more configuration parameters based on https://github.com/spack/spack-packages/pull/21


.ci/env
```diff
@@ -1 +1,3 @@
 SPACK_CHECKOUT_VERSION=1559802c9cbb2996f9e29638cbc61a429e9388e8
+SPACK_CHECKOUT_SERVER=https://github.com
```
becker33: It doesn't look like this value is being used in the rest of the PR

kwryankrattiger: huh, that was supposed to be removed. I must have only dropped it fromthe `checkout_spack` action

becker33 (dismissed):

tgamblin (dismissed):

tgamblin (approved):

PR #192, opened 2025-06-10
Paraview 6 updates

kwryankrattiger:
Updates for ParaView 6

### DIFF
--- a/repos/spack_repo/builtin/packages/libproxy/package.py
+++ b/repos/spack_repo/builtin/packages/libproxy/package.py
@@ -58,17 +58,8 @@ class Libproxy(CMakePackage, MesonPackage):
     @property
     def libs(self):
         return find_libraries(
-            ["libproxy", "libpxbackend*"], root=self.prefix, shared=True, recursive=True
+            ["libproxy", "libproxy/*"], root=self.prefix, shared=True, recursive=True
         )
-
-    def setup_run_environment(self, env: EnvironmentModifications) -> None:
-        libs = self.spec["libproxy"].libs.directories[0]
-        if self.spec.satisfies("platform=darwin"):
-            env.prepend_path("DYLD_FALLBACK_LIBRARY_PATH", join_path(libs, "libproxy"))
-            env.prepend_path("DYLD_FALLBACK_LIBRARY_PATH", libs)
-        else:
-            env.prepend_path("LD_LIBRARY_PATH", join_path(libs, "libproxy"))
-            env.prepend_path("LD_LIBRARY_PATH", libs)
 
 
 class CMakeBuilder(cmake.CMakeBuilder):
--- a/repos/spack_repo/builtin/packages/libproxy/package.py
+++ b/repos/spack_repo/builtin/packages/libproxy/package.py
@@ -55,13 +55,20 @@ class Libproxy(CMakePackage, MesonPackage):
 
     extends("python", when="+python")
 
+    @property
+    def libs(self):
+        return find_libraries(
+            ["libproxy", "libpxbackend*"], root=self.prefix, shared=True, recursive=True
+        )
+
     def setup_run_environment(self, env: EnvironmentModifications) -> None:
-        if self.spec.satisfies("+python"):
-            libs = self.spec["libproxy"].libs.directories[0]
-            if self.spec.satisfies("platform=darwin"):
-                env.prepend_path("DYLD_FALLBACK_LIBRARY_PATH", libs)
-            else:
-                env.prepend_path("LD_LIBRARY_PATH", libs)
+        libs = self.spec["libproxy"].libs.directories[0]
+        if self.spec.satisfies("platform=darwin"):
+            env.prepend_path("DYLD_FALLBACK_LIBRARY_PATH", join_path(libs, "libproxy"))
+            env.prepend_path("DYLD_FALLBACK_LIBRARY_PATH", libs)
+        else:
+            env.prepend_path("LD_LIBRARY_PATH", join_path(libs, "libproxy"))
+            env.prepend_path("LD_LIBRARY_PATH", libs)
 
 
 class CMakeBuilder(cmake.CMakeBuilder):
--- a/repos/spack_repo/builtin/packages/paraview/package.py
+++ b/repos/spack_repo/builtin/packages/paraview/package.py
@@ -310,16 +310,11 @@ class Paraview(CMakePackage, CudaPackage, ROCmPackage):
     with when("@6:"):
         # ParaView 6 and later will not support Spack builds with Qt5.
         with when("+qt"):
-            # https://discourse.paraview.org/t/paraview-5-9-and-minimum-recommended-qt-version/5333
-            depends_on("qt@5.12: +sql+opengl", when="@6:")
-
-            # TODO: qt-tools creates an unconcretizable environment with openfoam %oneapi
-            #       These dependencies otherwise are correct
-            # depends_on("qt-base@6.9.0 +accessibility+gui+opengl+sql+network")
-            # depends_on("qt-tools+assistant")
-            # depends_on("qt-5compat")
-            # depends_on("qt-svg")
-            # depends_on("libxslt")
+            depends_on("qt-base@6.9.0 +accessibility+gui+opengl+sql+network")
+            depends_on("qt-tools+assistant")
+            depends_on("qt-5compat")
+            depends_on("qt-svg")
+            depends_on("libxslt")
 
         # ParaView@6: and later will depend on OSMesa as a fallback for
         # OpenGL.
--- a/repos/spack_repo/builtin/packages/paraview/package.py
+++ b/repos/spack_repo/builtin/packages/paraview/package.py
@@ -40,6 +40,10 @@ supported_cuda_archs = {
 }
 
 
+# This is a list of paraview variants that require the viskores library.
+viskores_dependency_variants = ["+cuda", "+fides", "+rocm"]
+
+
 class Paraview(CMakePackage, CudaPackage, ROCmPackage):
     """ParaView is an open-source, multi-platform data analysis and
     visualization application. This package includes the Catalyst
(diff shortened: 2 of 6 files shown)


repos/spack_repo/builtin/packages/paraview/package.py
```diff
@@ -236,26 +237,58 @@ class Paraview(CMakePackage, CudaPackage, ROCmPackage):
...
-
-    depends_on("gl@3.2:", when="+opengl2")
-    depends_on("gl@1.2:", when="~opengl2")
-    depends_on("glew")
-    depends_on("libxt", when="platform=linux ^[virtuals=gl] glx")
-
-    for plat in ["linux", "darwin", "freebsd"]:
+    # Handle X11 dependencies
+    # X is only used on Unix like platforms
+    conflicts("glx", when="~x")
```
vicentebolea, line 245: Is wayland unsupported?

repos/spack_repo/builtin/packages/qt_tools/package.py
```diff
@@ -52,17 +62,13 @@ class QtTools(QtPackage):
...
+        args = super().cmake_args() + [
+            self.define_feature("fullqthelp", False),
+            self.define_feature("qdoc"),
+            self.define_feature("clang", "qdoc"),
+            self.define_feature("assistant"),
+            self.define_feature("designer"),
+            self.define_feature("linguist"),
+        ]
 
         return args
```
vicentebolea:
```suggestion
        return super().cmake_args() + [
            self.define_feature("fullqthelp", False),
            self.define_feature("qdoc"),
            self.define_feature("clang", "qdoc"),
            self.define_feature("assistant"),
            self.define_feature("designer"),
            self.define_feature("linguist"),
        ]
```

vicentebolea (commented):
ParaView bites looks great

repos/spack_repo/builtin/packages/paraview/package.py
kwryankrattiger: It isn't _not_ supported, but I think the state of the art for gl on wayland is to use glx + xwayland.

vicentebolea (approved):

vicentebolea (dismissed):
libxslt is failing in the CI

tldahlgren:
Added remaining missing maintainer.

tldahlgren:
FYI. `qt-tools` has conflicts to resolve.

repos/spack_repo/builtin/packages/paraview/package.py
```diff
@@ -146,18 +170,19 @@ class Paraview(CMakePackage, CudaPackage, ROCmPackage):
     variant(
         "use_vtkm",
         default="default",
+        when="@5.3.0:5.13",
         multi=False,
         values=("default", "on", "off"),
-        description="Build VTK-m with ParaView by setting PARAVIEW_USE_VTKM=ON,OFF."
+        description="Build VTK-m(Viskores) with ParaView."
```
kmorel: Since the `use_vtkm` option only works with versions of ParaView that use VTK-m (and not Viskores), mentioning Viskores here will probably make things more confusing in the long term. Just let users forget about VTK-m and `use_vtkm` and instead let users focus on variants like `+cuda` and `+rocm`.

kmorel (approved):
Looks good for Viskores module and the parts of the ParaView module I looked at. I had a small comment, but not a big deal.

kmorel (approved):

kmorel:
@kwryankrattiger I can't help but note that this PR is quite old and we are still waiting for ParaView 6 to be available in Spack. What is the status of this?

tldahlgren:
There are a couple of files with conflicts that need to be resolved.

kwryankrattiger:
> @kwryankrattiger I can't help but note that this PR is quite old and we are still waiting for ParaView 6 to be available in Spack. What is the status of this?

Yes, it is quite old :(

The main blocker on this was that an unrelated package was getting stuck on a concretization error. To address this more systematically the X, GL, and Qt updates were extracted into separate PRs.

These two PRs that have since found themselves hung up in review:
* GL https://github.com/spack/spack-packages/pull/2057
* Qt https://github.com/spack/spack-packages/pull/2058

> There are a couple of files with conflicts that need to be resolved.

These are being ignored, there will be a rebase to only contain the paraview related changes at some point. I am hopeful to remaining two PRs will get merged sooner rather than later.

repos/spack_repo/builtin/packages/gobject_introspection/package.py
```diff
@@ -158,8 +158,19 @@ def setup_dependent_build_environment(
...
+        # system paths
+        # If it is built by Spack it should be safe to prepend.
         if self.spec.satisfies("^glib-bootstrap"):
-            env.append_path("PKG_CONFIG_PATH", self.spec["glib-bootstrap"].prefix.lib.pkgconfig)
+            glib_bootstrap_spec = self.spec["glib-bootstrap"]
+            pkg_config_path = self.spec["glib-bootstrap"].prefix.lib.pkgconfig
+            if glib_bootstrap_spec.external:
+                env.append_path("PKG_CONFIG_PATH", pkg_config_path)
+            else:
+                env.prepend_path("PKG_CONFIG_PATH", pkg_config_path)
```
michaelkuhn: Why is this necessary?

kwryankrattiger: The root of the issue is the link dep of a build dep is not included in the PKG_CONFIG_PATH variable. It is labeled, oddly, as a runtime only dep. To fix that is going to require a lot more thought as we have to change logic that is buried inside of build environment and my first pass at that was incomplete. This is easier and fixes the most immediate case.

You can still have a case where `glib` and `glib-bootstrap` are both in the deps and the glib-2.0 is always detected as the bootstrap version even if that isn't what is wanted.

--

So on my system I have an external package configured also in this stack. This creates a problem with pkg-config search for `glib-2.0` finds the wrong glib.

When we build the search path for spack we prepend all externals and then all non externals. This way we search the spack built prefixes first, then external prefixes. The external prefix that I have is not excluded and also contains the pkg-config files for all of my system packages. Including an external glib that doesn't satisfy the version constraint for gobject and is the wrong glib I'm general. So appending this package path always results in broken search in a common case.

So to "fix" this case I try to mimic the behavior of the build environment where spack installed packages are listed first, and externals are listed second. This is not a complete fix, but it gets past this common case for now.

repos/spack_repo/builtin/packages/openfoam/package.py
```diff
@@ -356,9 +356,9 @@ class Openfoam(Package):
     variant("scotch", default=True, description="With scotch/ptscotch decomposition")
     variant("zoltan", default=False, description="With zoltan renumbering")
     variant("mgridgen", default=False, description="With mgridgen support")
-    variant(
-        "paraview", default=False, description="Build paraview plugins and runtime post-processing"
-    )
+    # variant(
```
kwryankrattiger: @olesenm Are you okay with dropping this variant for now and opening a PR to re-add it?

Right now there is something going on in the concretizer that is overly constraining here such that bison cannot be concretized with GCC. As far as I can tell there is nothing wrong in packages but it is currently blocking the release of ParaView 6.

repos/spack_repo/builtin/packages/paraview/package.py
```diff
@@ -591,10 +666,10 @@ def use_x11():
             cmake_args.extend(
                 [
                     "-DPARAVIEW_ENABLE_EXAMPLES:BOOL=%s" % variant_bool("+examples"),
-                    "-DVTK_MODULE_USE_EXTERNAL_ParaView_cgns=OFF",
                     "-DVTK_MODULE_USE_EXTERNAL_VTK_gl2ps=OFF",
                     "-DVTK_MODULE_USE_EXTERNAL_VTK_libharu=OFF",
                     "-DVTK_MODULE_USE_EXTERNAL_VTK_utf8=OFF",
+                    "-DVTK_MODULE_USE_EXTERNAL_ParaView_cgns=OFF",
```
kmorel: I'm not sure why these are not left in alphabetical order, but OK.

repos/spack_repo/builtin/packages/paraview/package.py
```diff
@@ -667,10 +747,26 @@ def use_x11():
...
-            )
+        if spec.satisfies("@5.3.0:5") and spec.variants["use_vtkm"].value != "default":
+            if spec.satisfies("@6:"):
+                cmake_args.append(
+                    "-DPARAVIEW_USE_VISKORES:BOOL=%s" % spec.variants["use_vtkm"].value.upper()
+                )
+            else:
+                cmake_args.append(
+                    "-DPARAVIEW_USE_VTKM:BOOL=%s" % spec.variants["use_vtkm"].value.upper()
+                )
```
kmorel: I don't see how this `if` condition ever evaluates to true. Can't you just use the else clause?


```suggestion
            cmake_args.append(
                "-DPARAVIEW_USE_VTKM:BOOL=%s" % spec.variants["use_vtkm"].value.upper()
            )
```

kmorel (approved):
I have a couple of minor comments on the paraview package that don't affect its correctness. Otherwise, the changes to that and the viskores module look good to me.

repos/spack_repo/builtin/packages/paraview/package.py
kwryankrattiger: Ah this predated dropping the `use_vtkm` variant good catch.

repos/spack_repo/builtin/packages/paraview/package.py
kwryankrattiger: yeah, I wasn't sure where that came from. This has been rebased a lot, will reorder

repos/spack_repo/builtin/packages/gobject_introspection/package.py
haampie: > It is labeled, oddly, as a runtime only dep.

This isn't odd at all. Pure build dependencies have to run in the build environment. Their link deps are irrelevant for the to-be-built node (assuming they are not reachable as link deps from the root). Those link deps are only relevant for the to-be-executed build dependency and all it should run is `setup_run_environment` like functions. 

If you have pure build deps with isolated dependencies, their contributions to the build env is basically the same as `spack load build-dep1 build-dep2` followed by `setup_dependent_build_environment`.

haampie: I think you can fix this better by just making `glib-bootstrap` a build dependency of `glib` and let Spack run the partitioning. Can you please extract this as a separate PR?

kwryankrattiger: > This isn't odd at all. 

A pure build dependencies dependency is part of the runtime of the dependent, that is very strange. It shouldn't show up there at all. Maybe I read that wrong, but it was seeming strange.

> I think you can fix this better by just making glib-bootstrap a build dependency of glib

That actually doesn't really work. It was my first solution, but you would need to do the same thing for _all_ of the packages that depend on `gobject-introspection` which was a huge patch and it doesn't fully solve the issue any better either. It also becomes a bit opaque as not all of the libraries that depend on `gobject-introspection` are build only, so you need to be a bit careful about how it is applied.

I can extract it since it seems this PR is once again stalled indefinitely :/

repos/spack_repo/builtin/packages/openfoam/package.py
kwryankrattiger: Dropped this change here. Falling back to Qt5 fixes this for now..

repos/spack_repo/builtin/packages/libproxy/package.py
```diff
@@ -55,13 +55,20 @@ class Libproxy(CMakePackage, MesonPackage):
...
+        )
+
     def setup_run_environment(self, env: EnvironmentModifications) -> None:
-        if self.spec.satisfies("+python"):
-            libs = self.spec["libproxy"].libs.directories[0]
-            if self.spec.satisfies("platform=darwin"):
-                env.prepend_path("DYLD_FALLBACK_LIBRARY_PATH", libs)
-            else:
-                env.prepend_path("LD_LIBRARY_PATH", libs)
+        library_path_var = "LD_LIBRARY_PATH"
```
haampie: Why is LD_LIBRARY_PATH needed?

kwryankrattiger: It is because they were working around the issue that `libpxbackend` is needed at runtime but otherwise not found. I am not entirely sure if that is supposed to be swapable at runtime or if it is always okay to just use a library from that directory. I added the `libs` function (the spack way to handle this with rpaths) so it is possible we can drop this now. Will retest without it.

johnwparent (approved):

PR #197, opened 2025-06-10
updating latte package file

finkeljos:
migrating PR from spack/spack to spack/packages. fixed mpi build bug in latte package.py and made our latte branch available which can interface with sedacs

### DIFF
--- a/repos/spack_repo/builtin/packages/latte/package.py
+++ b/repos/spack_repo/builtin/packages/latte/package.py
@@ -3,6 +3,8 @@
 # SPDX-License-Identifier: (Apache-2.0 OR MIT)
 
 
+from spack_repo.builtin.build_systems.cmake import CMakePackage
+
 from spack.package import *
 
 
--- a/repos/spack_repo/builtin/packages/latte/package.py
+++ b/repos/spack_repo/builtin/packages/latte/package.py
@@ -2,11 +2,9 @@
 #
 # SPDX-License-Identifier: (Apache-2.0 OR MIT)
 
-
 from spack_repo.builtin.build_systems.cmake import CMakePackage
 
 from spack.package import *
-
 
 class Latte(CMakePackage):
     """Open source density functional tight binding molecular dynamics."""
@@ -14,24 +12,24 @@ class Latte(CMakePackage):
     homepage = "https://github.com/lanl/latte"
     url = "https://github.com/lanl/latte/tarball/v1.2.1"
     git = "https://github.com/lanl/latte.git"
-
+    
     maintainers("jeanlucf22", "finkeljos")
 
     tags = ["ecp", "ecp-apps"]
-
+    
     license("LGPL-2.0-or-later")
-
+    
     version("master", branch="master")
     version("lattepy", branch="lattepy")
     version("1.2.2", sha256="ab1346939dbd70ffd89c5e5bf8d24946cb3655dc25b203bec7fc59c6c63e4c79")
     version("1.2.1", sha256="a21dda5ebdcefa56e9ff7296d74ef03f89c200d2e110a02af7a84612668bf702")
     version("1.0.1", sha256="67b2957639ad8e36b69bc6ea9a13085183a881562af9ca6d2b90b412ff073789")
-
+    
     variant("interface", default=False, description="Build with interfacing to use with sedacs")
     variant("mpi", default=True, description="Build with mpi")
     variant("progress", default=False, description="Use progress for fast solvers")
     variant("shared", default=True, description="Build shared libs")
-
+    
     depends_on("c", type="build")  # generated
     depends_on("fortran", type="build")  # generated
     depends_on("cmake@3.1:", type="build")
--- a/repos/spack_repo/builtin/packages/latte/package.py
+++ b/repos/spack_repo/builtin/packages/latte/package.py
@@ -27,17 +27,16 @@ class Latte(CMakePackage):
     version("1.2.1", sha256="a21dda5ebdcefa56e9ff7296d74ef03f89c200d2e110a02af7a84612668bf702")
     version("1.0.1", sha256="67b2957639ad8e36b69bc6ea9a13085183a881562af9ca6d2b90b412ff073789")
 
+    variant("interface", default=False, description="Build with interfacing to use with sedacs")
+    variant("mpi", default=True, description="Build with mpi")
+    variant("progress", default=False, description="Use progress for fast solvers")
+    variant("shared", default=True, description="Build shared libs")
+
     depends_on("c", type="build")  # generated
     depends_on("fortran", type="build")  # generated
     depends_on("cmake@3.1:", type="build")
     depends_on("blas")
     depends_on("lapack")
-
-    variant("interface", default=False, description="Build with interfacing to use with sedacs")
-    variant("mpi", default=True, description="Build with mpi")
-    variant("progress", default=False, description="Use progress for fast solvers")
-    variant("shared", default=True, description="Build shared libs")
-    
     depends_on("mpi", when="+mpi")
     depends_on("qmd-progress", when="+progress")
 
--- a/repos/spack_repo/builtin/packages/latte/package.py
+++ b/repos/spack_repo/builtin/packages/latte/package.py
@@ -34,14 +34,12 @@ class Latte(CMakePackage):
     depends_on("lapack")
 
     variant("interface", default=False, description="Build with interfacing to use with sedacs")
-
     variant("mpi", default=True, description="Build with mpi")
-    depends_on("mpi", when="+mpi")
-
     variant("progress", default=False, description="Use progress for fast solvers")
-    depends_on("qmd-progress", when="+progress")
-
     variant("shared", default=True, description="Build shared libs")
+    
+    depends_on("mpi", when="+mpi")
+    depends_on("qmd-progress", when="+progress")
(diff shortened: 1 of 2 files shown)


repos/spack_repo/builtin/packages/latte/package.py
```diff
@@ -46,13 +52,21 @@ def cmake_args(self):
         else:
             options.append("-DBUILD_SHARED_LIBS=OFF")
         if self.spec.satisfies("+mpi"):
-            options.append("-DO_MPI=yes")
+            options.append("-DDO_MPI=ON")
+            options.append("-DCMAKE_C_COMPILER=%s" % self.spec["mpi"].mpicc)
+            options.append("-DCMAKE_CXX_COMPILER=%s" % self.spec["mpi"].mpicxx)
+            options.append("-DCMAKE_Fortran_COMPILER=%s" % self.spec["mpi"].mpifc)
```
tldahlgren, line 54: Nit. You could be using python f-strings here and elsewhere (see https://docs.python.org/3/tutorial/inputoutput.html#formatted-string-literals).

repos/spack_repo/builtin/packages/sedacs/package.py
```diff
@@ -0,0 +1,63 @@
+# Copyright Spack Project Developers. See COPYRIGHT file for details.
+#
+# SPDX-License-Identifier: (Apache-2.0 OR MIT)
+
+from spack.package import *
+from spack_repo.builtin.build_systems.python import PythonPackage
+from spack_repo.builtin.build_systems.cuda import CudaPackage
+
+
+class Sedacs(PythonPackage, CudaPackage):
```
tldahlgren, line 11: Have you reviewed the naming scheme (https://spack.readthedocs.io/en/latest/build_systems/pythonpackage.html#choosing-a-package-name) for python packages before choosing the file name?

repos/spack_repo/builtin/packages/sedacs/package.py
```diff
@@ -0,0 +1,63 @@
...
+        "cuda_arch=none", when="+gpulib",  msg="sedacs: Please select a CUDA arch value"
+    )
+
+    # build latte with qmd-progress and other branch/variant restrictions
+    variant("latte", default=False, description="Make latte available as a sedacs engine")
+    depends_on('latte@lattepy+interface+progress',when="+latte") 
+    depends_on('qmd-progress@master~benchmarks',when="+latte") 
+
+    # build system
+    build_system = 'pyproject'
```
tldahlgren: 🤔 I haven't seen this one before. The syntax is also different.

See https://spack.readthedocs.io/en/latest/packaging_guide_build.html#what-are-build-systems and https://spack.readthedocs.io/en/latest/packaging_guide_advanced.html#multiple-build-systems.

repos/spack_repo/builtin/packages/sedacs/package.py
```diff
@@ -0,0 +1,63 @@
...
+    version('main', branch='main')
+
+    maintainers("finkeljos", "cnegre")
+
+    tags = ["lanl"]
+
+    license("BSD 3-Clause")
+
+    # python dependencies
+    depends_on('python@3.9:', type=('build', 'run'))
```
tldahlgren: According to the `pyproject.toml` file, the python dependency should be `python@3.10:`.

```
dependencies = [
    "scipy>=1",
    "numpy>=1",
    "torch>=2",
]
requires-python = ">=3.10"
```

repos/spack_repo/builtin/packages/sedacs/package.py
```diff
@@ -0,0 +1,63 @@
...
+    depends_on('py-setuptools', type='build')
+    depends_on('py-wheel', type='build')
```
tldahlgren: The dependency on `py-wheel` is inherited from `PythonPackage`.

repos/spack_repo/builtin/packages/sedacs/package.py
```diff
@@ -0,0 +1,63 @@
...
+    depends_on('py-scikit-build-core', type='build')
+    depends_on('py-numpy', type=('build', 'run'))
+    depends_on('py-torch', type=('build', 'run'))
+    depends_on('py-scipy', type=('build', 'run'))
```
tldahlgren: Based on `pyproject.toml`, these should have minimum versions:

```suggestion
    depends_on('py-numpy@1:', type=('build', 'run'))
    depends_on('py-torch@2:', type=('build', 'run'))
    depends_on('py-scipy@1:', type=('build', 'run'))
```

 Though `py-torch` is the only package that lists older version so technically requires the minimum version.

repos/spack_repo/builtin/packages/sedacs/package.py
```diff
@@ -0,0 +1,63 @@
...
+    depends_on('cmake',when="+gpulib")
+    depends_on('nvhpc',when="+gpulib")
+    conflicts(
+        "cuda_arch=none", when="+gpulib",  msg="sedacs: Please select a CUDA arch value"
+    )
+
+    # build latte with qmd-progress and other branch/variant restrictions
+    variant("latte", default=False, description="Make latte available as a sedacs engine")
+    depends_on('latte@lattepy+interface+progress',when="+latte") 
+    depends_on('qmd-progress@master~benchmarks',when="+latte")
```
tldahlgren: Style-wise, variant directives are listed together before dependencies (`depends_on`).  See https://spack.readthedocs.io/en/latest/packaging_guide_creation.html#structure-of-a-package.

repos/spack_repo/builtin/packages/latte/package.py
```diff
@@ -15,28 +15,34 @@ class Latte(CMakePackage):
...
+
+    variant("interface", default=False, description="Build with interfacing to use with sedacs")
+
+    variant("mpi", default=True, description="Build with mpi")
     depends_on("mpi", when="+mpi")
+
+    variant("progress", default=False, description="Use progress for fast solvers")
     depends_on("qmd-progress", when="+progress")
 
+    variant("shared", default=True, description="Build shared libs")
```
tldahlgren: See comment on `sedacs` regarding ordering of `variant` and `depends_on directives`.

tldahlgren (changes requested):
Confirmed the homepage and branches but there are issues e.g. with dependencies, (maybe) naming, and directive ordering. See comments for more information.

repos/spack_repo/builtin/packages/sedacs/package.py
finkeljos: Yes, we thought about appending a py-, but this code is meant to interface to several low-level languages like fortran and C++/C. So we feel its more appropriate to not include py-.

repos/spack_repo/builtin/packages/latte/package.py
finkeljos: We have other packages in spack (LATTE, BML etc) which has this syntax. We wanted to stay consistent. Is it a problem this way?

tldahlgren: Not a problem. We've been switching spack core to use F strings and are encouraging packaged developers to do the same.

repos/spack_repo/builtin/packages/sedacs/package.py
finkeljos: fixed

repos/spack_repo/builtin/packages/sedacs/package.py
```diff
@@ -0,0 +1,64 @@
...
+    depends_on("nvhpc", when="+gpulib")
+    conflicts("cuda_arch=none", when="+gpulib", msg="sedacs: Please select a CUDA arch value")
+
+    # latte dependencies
+    depends_on("latte@lattepy+interface+progress", when="+latte")
+    depends_on("qmd-progress@master~benchmarks", when="+latte")
+
+    # build system
+    build_system = "pyproject"
+    build_backend = "scikit_build_core.build"
```
tldahlgren: Why are you specifying these?

See https://spack.readthedocs.io/en/latest/packaging_guide_build.html#what-are-build-systems for the meaning of `build_system`..

tldahlgren: You can also see the option chosen by the concretizer by running `spack spec sedacs`. For example, you can see it is set to `python_pip` -- `sedacs@main~cuda~gpulib~latte+mpi build_system=python_pip commit=02edd5337be04aa84c3c6c1047ff9096d8130ad5 arch=darwin-sequoia-m1` -- when I run on a laptop.

tldahlgren (changes requested):
Confirmed the homepage and branches. 

Still unclear why two properties are being set.  Please remove or add a comment to explain their purpose/use.

repos/spack_repo/builtin/packages/sedacs/package.py
finkeljos: Thanks for getting back to me so fast. I am specifying this because I need the line "scikit_build_core.build" in order to execute our build of a C++ library that interfaces to the python code properly. Although, this is only necessary when using +gpulib variant. If I don't include this, cmake doesn't properly build this library. What do you suggest?

finkeljos: Thanks for getting back to me so fast. comment added!

tldahlgren (approved):

PR #198, opened 2025-06-10
bisect spack/spack concretizer slowness

haampie:
looks like data-vis is slow to concretize, see if it's thanks to a spack bump.


haampie:
it's flaky on develop

PR #203, opened 2025-06-11
scorep: fix name of oneapi compilers in configure

melven:
If `intel-oneapi-compilers` is passed to the scorep configure script, one gets the following error:
```
1 error found in build log:
     87    checking how to create a pax tar archive... gnutar
     88    checking whether make supports nested variables... (cached) yes
     89    checking whether to enable maintainer-specific portions of Makefiles
           ... no
     90    checking build system type... x86_64-pc-linux-gnu
     91    checking for platform... linux (auto detected)
     92    checking for cross compilation... no
  >> 93    configure: error: compiler suite "intel-oneapi-compilers" not suppor
           ted by --with-nocross-compiler-suite.
```

This fixes it by setting the string to `oneapi` instead.


tldahlgren (dismissed):
@wrwilliams Any objections to this PR being merged?

repos/spack_repo/builtin/packages/scorep/package.py
```diff
@@ -207,6 +207,8 @@ def configure_args(self):
         ]
 
         cname = self.clean_compiler(spec.compiler.name)
+        if cname == "intel-oneapi-compilers":
+            cname = "oneapi"
```
wrwilliams: Please keep renames together in the `clean_compiler` helper. But the logic here should be fine.

wrwilliams (changes requested):

repos/spack_repo/builtin/packages/scorep/package.py
tldahlgren: Thanks for pointing that out. I'll wait for the change before merging.

melven: Done!

wrwilliams (approved):
Ready for merge as far as I'm concerned.

tldahlgren (approved):

PR #204, opened 2025-06-11
Add ParEval Package

Dando18:
Adds the [ParEval benchmark](https://github.com/parallelcodefoundry/pareval) as a package to Spack.

### DIFF
--- a/repos/spack_repo/builtin/packages/pareval/package.py
+++ b/repos/spack_repo/builtin/packages/pareval/package.py
@@ -12,6 +12,7 @@ from spack.package import *
 class Pareval(Package, CudaPackage, ROCmPackage):
     """ParEval benchmark for evaluating LLMs' parallel code generation capabilities"""
 
+    homepage = "https://github.com/parallelcodefoundry/pareval"
     url = "https://github.com/parallelcodefoundry/pareval"
     git = "https://github.com/parallelcodefoundry/pareval.git"
 
--- a/repos/spack_repo/builtin/packages/pareval/package.py
+++ b/repos/spack_repo/builtin/packages/pareval/package.py
@@ -1,0 +1,60 @@
+# Copyright Spack Project Developers. See COPYRIGHT file for details.
+#
+# SPDX-License-Identifier: (Apache-2.0 OR MIT)
+
+from spack_repo.builtin.build_systems.cuda import CudaPackage
+from spack_repo.builtin.build_systems.generic import Package
+from spack_repo.builtin.build_systems.rocm import ROCmPackage
+
+from spack.package import *
+
+
+class Pareval(Package, CudaPackage, ROCmPackage):
+    """ParEval benchmark for evaluating LLMs' parallel code generation capabilities"""
+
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/pareval/package.py
```diff
@@ -0,0 +1,61 @@
...
+    git = "https://github.com/parallelcodefoundry/pareval.git"
+
+    maintainers("Dando18")
+
+    license("MIT", checked_by="Dando18")
+
+    version("develop", branch="develop")
+    version("1.2", tag="v1.2")
+    version("1.1", tag="v1.1")
+    version("1.0", tag="v1.0")
```
tldahlgren: Is there a reason for using tags instead of hashes?

We prefer the use of safe versions, which are those with `sha256` or `commit`. See https://spack.readthedocs.io/en/latest/packaging_guide_creation.html#checksum-verification.

tldahlgren (changes requested):
Confirmed the  homepage, branch, and tags; however, we generally prefer safe versions (i.e., ones with hashes). See the comment for more information.

repos/spack_repo/builtin/packages/pareval/package.py
```diff
@@ -0,0 +1,61 @@
...
+from spack_repo.builtin.build_systems.rocm import ROCmPackage
+
+from spack.package import *
+
+
+class Pareval(Package, CudaPackage, ROCmPackage):
+    """ParEval benchmark for evaluating LLMs' parallel code generation capabilities"""
+
+    homepage = "https://github.com/parallelcodefoundry/pareval"
+    url = "https://github.com/parallelcodefoundry/pareval"
```
tldahlgren: Based on the links for each version at https://github.com/parallelcodefoundry/ParEval/releases, the URL should be:

```suggestion
    url = "https://github.com/parallelcodefoundry/ParEval/archive/refs/tags/v1.2.tar.gz"
```

repos/spack_repo/builtin/packages/pareval/package.py
tldahlgren: Correcting the URL property yields the following sha256s:

```suggestion
    version("1.2", sha256="8b374afc07bc177a3ba335ff1fe782b3a1943bbcd2dd7ac0a62dedfa4f44f651")
    version("1.1", sha256="c07bf9b1027afed97281eb2fbcbe8f3d819ac12045ba9d1a7a90d5488f92cc53")
    version("1.0", sha256="314cf62e421c9382f4344d7da57f3dcdf9ab3e1272afd6a7badabf0ffe80f294")
```

after running `spack checksum pareval`.

Dando18: No particular reason, I've updated it to hashes. Thanks!

repos/spack_repo/builtin/packages/pareval/package.py
Dando18: I've updated this as well.

tldahlgren (approved):
(Re) con firmed homepage, branch, and sha256s.

PR #208, opened 2025-06-11
cpu-features: add version 0.10.1, update deps, and cleanup

pranav-sivaraman:
Built on my M1 laptop with all the tests passing.


tldahlgren (approved):
Confirmed the sha256.

PR #212, opened 2025-06-11
maker: add v2.31.11

mikerenfro:
Migrated from https://github.com/spack/spack/pull/40236


repos/spack_repo/builtin/packages/maker/package.py
```diff
@@ -35,8 +35,9 @@ class Maker(Package):
     homepage = "https://www.yandell-lab.org/software/maker.html"
     manual_download = True
 
-    version("3.01.04", sha256="87be5b0fad92d7d7b359f233877830e8a353a80277c3c88aa89f359899fa8bfb")
+    version("3.01.04", sha256="3d1f90a5ced5a7ed00e1136700d76d4d82315d76690223dcc50ceb712d5b5611")
```
tldahlgren, line 38: This version is newer than the last one shown at https://www.yandell-lab.org/software/maker.html.

tldahlgren: Please add a comment noting why there is a difference.

tldahlgren (changes requested):
Just need a comment.

PR #214, opened 2025-06-12
kokkos-kernels: declare dependency on c/fortran when using tpl blas libraries

msimberg:
cf. https://github.com/kokkos/kokkos-kernels/blob/e94234855d50831e6ad4ed0d78777cc84ae3f3c8/cmake/kokkoskernels_tpls.cmake#L434-L436.

For some third-party libraries kokkos-kernels enables the CMake C and Fortran languages. kokkos-kernels currently fails to build with those libraries enabled.

In theory the dependencies should be added for magma and armpl as well, but those are not supported by the spack package yet, and I'm not planning to add support for those in this PR. Hence the dependencies are only added for `+blas` and `+mkl`.

CC @lucbv @srajama1 @brian-kelley

### DIFF
--- a/repos/spack_repo/builtin/packages/kokkos_kernels/package.py
+++ b/repos/spack_repo/builtin/packages/kokkos_kernels/package.py
@@ -186,9 +186,9 @@ class KokkosKernels(CMakePackage, CudaPackage):
         description="Scalars",
     )
 
-    depends_on("c", type="build")
+    depends_on("cxx", type="build")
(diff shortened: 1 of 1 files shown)


msimberg:
Ping @brian-kelley @srajama1 @lucbv.

albestro (approved):
LGTM, waiting for maintainers before proceeding with this.

@brian-kelley @srajama1 @lucbv

lucbv (approved):

PR #221, opened 2025-06-12
Ipopt: fix spacing between mumps_flags when ^mumps~mpi

nkoukpaizan:
This fixes a bug encountered when trying to build Ipopt with ^mumps~mpi. The current version results in `-lpord-lmpiseq`, which is not found and should be `-lpord -lmpiseq`.


tldahlgren (approved):
Great catch!

PR #223, opened 2025-06-13
cgns: patch for compiling with nvfortran

melven:
Fixes compiling with Fortran enabled and nvhpc (tested with nvhpc24.5)

There is a space in a `bind(C, name="...")` which shouldn't be there. Other Fortran compilers seem to be less critical about it.
(at least while compiling, I didn't check if they create a wrong function name that has a space in it!)


tldahlgren (approved):
LGTM. @gsjaardema Any objections to this PR being merged?

gsjaardema (approved):
Looks good to me

PR #224, opened 2025-06-13
gdb: add version 16.2

melven:
Tested compiling and using it with gcc 13, 14 and 15.


tldahlgren (approved):
Confirmed the version sha256.

@robertu94 Any objections to this PR being merged?

robertu94 (approved):

PR #226, opened 2025-06-13
pdi: Add version 1.9.2

jbigot:
Hi @alalazo , here's another release of PDI. Would love to have your feedback.

Thanks!


tldahlgren (commented):
Confirmed the new sha256s. 

Why switch the other packages to relative imports?

jbigot:
The relative imports keeps those packages closer to those in our [dedicated repo](https://github.com/pdidev/spack) that we provide for users of older spack. This reduces the maintainance burden and ensure our test there translate to better quality here.

tldahlgren (approved):

PR #229, opened 2025-06-13
fastjet: add v3.5.0

m-fila:
Migrated from https://github.com/spack/spack/pull/50707


wdconinc (approved):
Thanks for the new version. Looks good to me.

Deferring to maintainers @vvolkl @drbenmorgan for approval.

(Optional) fastjet supports cmake builds (in addition to autotools) with this version, so it may be nice to add that support in the spack package too. See https://spack.readthedocs.io/en/latest/packaging_guide.html#multiple-build-systems for more info.

wdconinc:
Rebasing so the pipelines get unstuck.

PR #231, opened 2025-06-13
fmi4cpp: Adding missing C dependency

teabagk7:
Build fails without the change:
```
spack/cc: line 401: SPACK_CC_LINKER_ARG: ERROR: LINKER ARG WAS NOT SET, MAYBE THE PACKAGE DOES NOT DEPEND ON CC?
```


teabagk7:
Requesting review from @prudhomm

tldahlgren (approved):

PR #233, opened 2025-06-13
Snapped spackage

carlos-delgado15:


tgamblin:
We can't use this code as-is, because the only license is "all rights reserved, Paratools, Inc."  The authors should use a proper open source license.

Closing until the license is fixed.

PR #238, opened 2025-06-14
gdbm: conflicts with gcc@15: when @:1.24 (since C23 compatibility)

wdconinc:
This PR adds to `gdbm` a conflict with `gcc@15:` when `@:1.24`. C23 compability (default in gcc-15) was added in https://git.gnu.org.ua/gdbm.git/commit/?id=734f69451392ffb6280d8a3aef51e2673c9b8554.

### DIFF
--- a/repos/spack_repo/builtin/packages/gdbm/package.py
+++ b/repos/spack_repo/builtin/packages/gdbm/package.py
@@ -39,6 +39,10 @@ class Gdbm(AutotoolsPackage, GNUMirrorPackage):
 
     depends_on("readline")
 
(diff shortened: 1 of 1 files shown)


michaelkuhn:
Could you give #557 a try? I needed gdbm@1.23 since perl seems to still depend on it.

wdconinc:
> Could you give #557 a try? I needed gdbm@1.23 since perl seems to still depend on it.

I finally had a chance to test this and gdbm 1.23, 1.24, 1.25 all install fine with gcc@15 after #557 was merged.

PR #240, opened 2025-06-15
gcc: add c17 and c23 support

RobertMaaskant:
- https://gcc.gnu.org/onlinedocs/gcc-8.1.0/gcc/C-Dialect-Options.html#index-std-1
- https://gcc.gnu.org/onlinedocs/gcc-14.1.0/gcc/C-Dialect-Options.html#index-std-1


RobertMaaskant:
@spackbot re-run pipeline -- unrelated failures

becker33 (approved):

PR #243, opened 2025-06-15
wayland-protocols: add v1.4[1-5]

wdconinc:
This PR adds new `wayland-protocols`, up to v1.45. Diff at https://gitlab.freedesktop.org/wayland/wayland-protocols/-/compare/1.40...1.45.


tldahlgren (approved):
Confirmed all of the new version sha256s.

PR #245, opened 2025-06-15
harfbuzz: add v10.[3-4].*, v11.[0-2].*

wdconinc:
This PR adds newer `harfbuzz` versions, through v11.2.1. No addition of new meson options `fontations` and `harfruzz` since neither of those have dependencies in spack right now.


tldahlgren (commented):
Cannot confirm the version sha256s with `curl` but can with `urllib`.  (`spack checksum` and `verify-checksums` do confirm them.)

Is there a reason to add so many versions of the package?

wdconinc:
> Is there a reason to add so many versions of the package?

Mainly for completeness and to aid in identifying with which version issues were introduced. In this case, adding the later v10 versions adds updates for those packages/users not able/wanting to update to v11 yet.

tldahlgren (approved):

PR #246, opened 2025-06-16
node-js: add v24.2.0

wdconinc:
This PR adds `node-js`, v24.2.0, and makes it preferred.


tldahlgren (approved):
Confirmed the sha256s.

@cosmicexplorer Any objections to this PR being merged as is?

PR #247, opened 2025-06-16
Add actions/stale to GitHub actions

alalazo:
As voted in November 2024 by the steering committee, add actions/stale and:
- [x] Wait 6 months before marking an issue / PR as stale
- [x] Wait an additional month before closing it

Issues / PRs labeled as "pinned", or "bug", are exempted from stalebot checks.

### DIFF
--- a/.github/workflows/stale.yaml
+++ b/.github/workflows/stale.yaml
@@ -1,0 +1,40 @@
+name: 'Close stale issues and PRs'
+on:
+  schedule:
+    # Run every day at 2:39 UTC
+    - cron: '39 2 * * *'
+  workflow_dispatch:
+
+jobs:
+  stale:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/stale@5bef64f19d7facfb25b37b414482c7164d639639
+        with:
+          # Issues configuration
+          stale-issue-message: |
+            This issue has been automatically marked as stale because it has not had 
+            recent activity in the last 6 months. It will be closed if no further activity occurs.
+            Thank you for your contributions.
+          close-issue-message: |
+            This issue was closed because it has been stalled for 30 days with no activity.
+          stale-issue-label: 'stale'
+          exempt-issue-labels: 'pinned,bug'
+
+          # Pull requests configuration
+          stale-pr-message: |
+            This pull request has been automatically marked as stale because it has not had 
+            recent activity in the last 6 months. It will be closed if no further activity occurs.
+            Thank you for your contributions.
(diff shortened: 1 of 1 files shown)


haampie (approved):

.github/workflows/stale.yaml
```diff
@@ -0,0 +1,40 @@
...
+  schedule:
+    # Run every day at 2:39 UTC
+    - cron: '39 2 * * *'
+  workflow_dispatch:
+
+jobs:
+  stale:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/stale@5bef64f19d7facfb25b37b414482c7164d639639
```
adamjstewart, line 12: Does dependabot update this? Do we have dependabot set up on this repo yet?

.github/workflows/stale.yaml
```diff
@@ -0,0 +1,40 @@
...
+        with:
+          # Issues configuration
+          stale-issue-message: |
+            This issue has been automatically marked as stale because it has not had 
+            recent activity in the last 6 months. It will be closed if no further activity occurs.
+            Thank you for your contributions.
```
adamjstewart, line 18: Can we add instructions to add "pinned" if someone wants to keep this permanently open?

.github/workflows/stale.yaml
```diff
@@ -0,0 +1,40 @@
...
+          close-issue-message: |
+            This issue was closed because it has been stalled for 30 days with no activity.
+          stale-issue-label: 'stale'
+          exempt-issue-labels: 'pinned,bug'
```
adamjstewart, line 22: Personally I don't like making exceptions for "bug". I also think that for the package repo, all bugs and build bugs, so this might not even make sense. Do we have all of the same labels in this repo we had for core?

.github/workflows/stale.yaml
```diff
@@ -0,0 +1,40 @@
...
+
+          # Pull requests configuration
+          stale-pr-message: |
+            This pull request has been automatically marked as stale because it has not had 
+            recent activity in the last 6 months. It will be closed if no further activity occurs.
+            Thank you for your contributions.
```
adamjstewart, line 28: Same here, can we add instructions for adding "pinned" for things we want to keep permanently open?

.github/workflows/stale.yaml
alalazo: Build systems now are part of the repository, that's why I left the "bug" label in. We run unit tests over them.

PR #250, opened 2025-06-16
geomodel: add v6.11.0 through v6.14.0

stephenswat:
This commit adds v6.11.0, v6.11.1, v6.12.0, v6.13.0, and v6.14.0 of the geomodel package. The visualisation variant had to be disabled in the CI to handle the newly introduced EGL dependency.

### DIFF
--- a/repos/spack_repo/builtin/packages/geomodel/package.py
+++ b/repos/spack_repo/builtin/packages/geomodel/package.py
@@ -19,6 +19,11 @@ class Geomodel(CMakePackage):
 
     license("Apache-2.0", checked_by="wdconinc")
 
+    version("6.14.0", sha256="b294f624145f922efd1da4016b79e698cabb0034cc6791841648d7845fd9fb15")
+    version("6.13.0", sha256="8f1ebfe7fb502af078ed535410160becca2eb81286b94154b43c89873a3c45ad")
+    version("6.12.0", sha256="4fa32672c6cb3d9b89ddf5e9566d0c283e5769768db6236883d72fecdd6207cd")
+    version("6.11.1", sha256="f7b47ff4ed264e1bfb013e6cdb724e7532109f1d58ab5774656e6ff871afb362")
+    version("6.11.0", sha256="fc9fdd7d64b623586089949d9790182dcd93ebb35a05198c91eac8adbbbfd778")
     version("6.10.0", sha256="968a0f7c8108b14f22041ca0c6ae8a3293175131c6f61055527ecdefe8c7839a")
     version("6.9.0", sha256="ea34dad8a0cd392e06794b8a1b7407dd6ad617fefd19fb4cccdf36b154749793")
     version("6.8.0", sha256="4dfd5a932955ee2618a880bb210aed9ce7087cfadd31f23f92e5ff009c8384eb")
@@ -65,12 +70,16 @@ class Geomodel(CMakePackage):
         description="Use the specified C++ standard when building",
     )
 
+    # GeoModel 6.11 drops support for C++17.
+    conflicts("cxxstd=17", when="@6.11:")
+
     conflicts("+fullsimlight", when="+fsl", msg="FSL triggers the build of the FullSimLight")
 
     depends_on("c", type="build")
     depends_on("cxx", type="build")
 
     depends_on("cmake@3.16:", type="build")
+    depends_on("cmake@:3", when="@:6.10", type="build")
 
     depends_on("eigen@3.2.9:")
     depends_on("nlohmann-json@3.6.1:")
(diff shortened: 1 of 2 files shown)


repos/spack_repo/builtin/packages/geomodel/package.py
```diff
@@ -93,6 +100,9 @@ class Geomodel(CMakePackage):
         depends_on("coin3d")
         depends_on("soqt")
         depends_on("gl")
+        depends_on("egl", when="@6.11:")
```
wdconinc, line 108: I'm not sure if this is the right way to approach this. Since `egl` is only able to be provided as an external package, this would mean that `@6.11:` cannot be built anymore without external packages.

Tagging some maintainers for their input on this: @biddisco @snehring

snehring: That is correct. egl is currently external only, so anyone who would want to install this would need to install whatever package implements egl in their ecosystem and configure it as an external.

wdconinc: @stephenswat In that case, can you please change `.ci/gitlab/stacks/hep/spack.yaml` to use `~visualization`? Then these newer versions will be used in CI, which I think it preferred over continuously testing the last 6.10 version.

paulgessinger:
Could you add verions 6.13.0 and 6.14.0 as well @stephenswat?

stephenswat:
Updated as requested by @paulgessinger and @wdconinc.

paulgessinger:
Build log reports:

```
>> 538    /tmp/root/spack-stage/spack-stage-geomodel-6.14.0-xibsxttctjylt2pyt
            sp7ojxm6nhwflo3/spack-src/GeoModelCore/GeoModelHelpers/GeoModelHelp
            ers/StringUtils.icc:8:10: fatal error: format: No such file or dire
            ctory
     539        8 | #include <format>
     540          |          ^~~~~~~~
     541    compilation terminated.
```

GeoModel started using `std::format`, which I believe means it's no longer compatible with GCC11.
This was introduced in 6.11.0 (https://gitlab.cern.ch/GeoModelDev/GeoModel/-/merge_requests/432).

stephenswat:
> GeoModel started using `std::format`, which I believe means it's no longer compatible with GCC11.

Added a conflict with gcc 12 and lower and clang 16 and lower to resolve this. Thanks!

wdconinc:
> Build log reports:
> 
> ```
> >> 538    /tmp/root/spack-stage/spack-stage-geomodel-6.14.0-xibsxttctjylt2pyt
>             sp7ojxm6nhwflo3/spack-src/GeoModelCore/GeoModelHelpers/GeoModelHelp
>             ers/StringUtils.icc:8:10: fatal error: format: No such file or dire
>             ctory
>      539        8 | #include <format>
>      540          |          ^~~~~~~~
>      541    compilation terminated.
> ```
> 
> GeoModel started using `std::format`, which I believe means it's no longer compatible with GCC11.
> This was introduced in 6.11.0 (https://gitlab.cern.ch/GeoModelDev/GeoModel/-/merge_requests/432).

If you want (in a separate pr) we can update to ghcr.io/spack/ubuntu-24.04:v2024-09-05-v2 I stead of the current ubuntu-22.04 image with gcc11. I think this already came up once before so it's probably time.

wdconinc:
> If you want (in a separate pr) we can update to ghcr.io/spack/ubuntu-24.04:v2024-09-05-v2 I stead of the current ubuntu-22.04 image with gcc11. I think this already came up once before so it's probably time.

See https://github.com/spack/spack-packages/pull/372

wdconinc:
@spackbot run pipeline

repos/spack_repo/builtin/packages/geomodel/package.py
```diff
@@ -93,6 +105,9 @@ class Geomodel(CMakePackage):
...
+
+    depends_on("googletest", when="@6.11:")
```
paulgessinger: I think this is then ONLY needed on 6.11, right? Starting from 6.12 the tests are not built, so GTest is not needed.

stephenswat: Resolved by making it a build dependency in 6.11 and a test dependency in 6.12:.

repos/spack_repo/builtin/packages/geomodel/package.py
```diff
@@ -107,4 +123,8 @@ def cmake_args(self):
                 "GEOMODEL_USE_QT6", self.spec.satisfies("+visualization ^[virtuals=qmake] qt-base")
             ),
         ]
+
+        if self.spec.satisfies("@6.12:"):
+            args.append(self.define("GEOMODEL_BUILD_TESTING", "OFF"))
```
wdconinc:
```suggestion
            args.append(self.define("GEOMODEL_BUILD_TESTING", self.run_tests))
```

wdconinc (dismissed):
Looks good. Minor suggestion.

repos/spack_repo/builtin/packages/geomodel/package.py
stephenswat: Good idea; implemented.

wdconinc (approved):

PR #261, opened 2025-06-16
cpptrace: v1.0.0

RMeli:
https://github.com/jeremy-rifkin/cpptrace/releases/tag/v1.0.0

### DIFF
--- a/repos/spack_repo/builtin/packages/cpptrace/package.py
+++ b/repos/spack_repo/builtin/packages/cpptrace/package.py
@@ -1,0 +1,40 @@
+# Copyright Spack Project Developers. See COPYRIGHT file for details.
+#
+# SPDX-License-Identifier: (Apache-2.0 OR MIT)
+
+from spack_repo.builtin.build_systems.cmake import CMakePackage, generator
+
+from spack.package import *
+
+
+class Cpptrace(CMakePackage):
+    """Simple, portable, and self-contained stacktrace library for C++11 and newer."""
+
+    homepage = "https://github.com/jeremy-rifkin/cpptrace"
+    url = "https://github.com/jeremy-rifkin/cpptrace/archive/refs/tags/v1.0.0.tar.gz"
+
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/cpptrace/package.py
```diff
@@ -0,0 +1,40 @@
...
+
+    maintainers("RMeli")
+
+    license("MIT", checked_by="RMeli")
+
+    version("1.0.0", sha256="0e11aebb6b9b98ce9134a58532b63982365aadc76533a4fbb7f6fb6edb32de2e")
+
+    variant("shared", default=False, description="Build shared libraries")
+    variant("external_libdwarf", default=False, description="Use external libdwarf")
+    variant("external_zstd", default=False, description="Use external zstd")
```
albestro, line 24: Not a request, just wondering if there is any benefit in using "bundled" ones.

In general I would prefer having them managed by spack instead of getting dependencies "hidden" in packages. The only reason I can think of is about compatibility, but I would solve it with a conflict/requirement.

pranav-sivaraman:
This [PR](https://github.com/spack/spack/pull/45605) is my previous attempt at adding this package. I was eventually working towards adding `libassert`. I think one thing I was caught up on are the implicit system dependencies that exist and trying to model those. 

Also I am pretty sure `zstd` is a transitive dependency from `libdwarf` and not a direct dependency.

tldahlgren (approved):
Confirmed the homepage and version sha256.

tldahlgren:
Will defer merge decision since @albestro and @pranav-sivaraman have comments.

tldahlgren:
It's been two weeks without additional comments. Will merge with the expectation of a follow-on PR to address any new issues.

RMeli:
Apologies, this PR went through the cracks on my side. I can apply the changes suggested by @albestro on a follow-up PR when I have time.

PR #262, opened 2025-06-16
repeatmodeler: patch perl shebangs

snehring:
migrate from spack repo (https://github.com/spack/spack/pull/38420)

adds a list of files to patch (if present) to replace shebangs. There was some discussion in the original PR about reworking how this package worked but I do not believe a pr was ever submitted.

### DIFF
--- a/repos/spack_repo/builtin/packages/repeatmodeler/package.py
+++ b/repos/spack_repo/builtin/packages/repeatmodeler/package.py
@@ -85,7 +85,7 @@ class Repeatmodeler(Package):
             "util/visualizeAlignPNG.pl",
         ]
         for f in file_list:
-            filter_file("^#!/usr/bin/perl.*$", "#!/usr/bin/env perl", f)
+            filter_file("^#!/usr/bin/perl.*$", "#!/usr/bin/env perl", f, ignore_absent=True)
 
     def install(self, spec, prefix):
         # interactive configuration script
--- a/repos/spack_repo/builtin/packages/repeatmodeler/package.py
+++ b/repos/spack_repo/builtin/packages/repeatmodeler/package.py
@@ -83,9 +83,13 @@ class Repeatmodeler(Package):
             "util/TSD.pl",
             "util/viewMSA.pl",
             "util/visualizeAlignPNG.pl",
+            "configure",
         ]
         for f in file_list:
-            filter_file("^#!/usr/bin/perl.*$", "#!/usr/bin/env perl", f, ignore_absent=True)
+            filter_file(r"^#!.*perl( -w)?.*$", fr"#!{self.spec['perl'].prefix.bin.perl}\1", f, ignore_absent=True)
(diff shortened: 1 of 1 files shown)


tldahlgren:
Added previous PR and @EbiArnie since reviewed it.

tldahlgren:
Very strange that the pipeline is reported as failed but no failures appear when the link is followed.

tldahlgren:
@spackbot run pipeline

tldahlgren:
@snehring @EbiArnie Is this PR ready to merge (once CI passes)?

tldahlgren:
🤔 CI actually passed (https://gitlab.spack.io/spack/spack-packages/-/pipelines/1140306) but the status didn't get reported back to this PR.  

It also looks like spackbot didn't restart the pipeline so will close and re-open.

tldahlgren (dismissed):

repos/spack_repo/builtin/packages/repeatmodeler/package.py
```diff
@@ -47,6 +47,46 @@ class Repeatmodeler(Package):
...
+            "util/Linup",
+            "util/renameIds.pl",
+            "util/resolveIndels.pl",
+            "util/rmblast.pl",
+            "util/TSD.pl",
+            "util/viewMSA.pl",
+            "util/visualizeAlignPNG.pl",
+        ]
+        for f in file_list:
+            filter_file("^#!/usr/bin/perl.*$", "#!/usr/bin/env perl", f, ignore_absent=True)
```
EbiArnie: Some of the files have 

> /usr/local/bin/perl

and this would not work there.

Also, we're dropping the '-w', that's not ideal.

Maybe something like this:
`filter_file("^#!/usr/(?:local/)?bin/perl( -w)?.*$", "#!/usr/bin/env perl\1", f, ignore_absent=True)`

Although I'm not sure if we can use backrefs here at all. Needs to be tested :-)

EbiArnie (changes requested):

repos/spack_repo/builtin/packages/repeatmodeler/package.py
snehring: I bet this will work fine, I'm dealing with some issues with a dependency so testing is pending that getting resolved.

tldahlgren:
Ping me when this is ready to be re-reviewed/merged.

repos/spack_repo/builtin/packages/repeatmodeler/package.py
snehring: I modified it a bit to replace any shebang with perl in it to the actual perl in the spec (instead of using env), -w should be added when present (and seems to be in testing)

snehring:
I think I'll update this to 2.0.7 as well, but in a separate PR once this is merged.

EbiArnie (approved):
Looks good

EbiArnie:
Thanks!

PR #268, opened 2025-06-17
beast2: new versions need the jre subfolder installed

adnan-umich:
@snehring 

Issue when installing beast2 v2.7+ binary

- Added support for jre folder, now required by modern beast2 versions. 
- Fixed installation issues, where beast2 package manager defaults to users ~/.beast folder and throws an error if ~/.beast/2.7/toDeleteList is missing.

### DIFF
--- a/repos/spack_repo/builtin/packages/beast2/package.py
+++ b/repos/spack_repo/builtin/packages/beast2/package.py
@@ -50,6 +50,15 @@ class Beast2(Package):
                     filter_file(
                         r"(beast\.pkgmgmt.*\b)|(viz.*\b)", "{0} {1} \\1".format(javafx, modules), i
                     )
+                    
+        if self.spec.satisfies("@2.7.0:"):
+            # print a message that indicates you need to create /home/${USER}/.beast/2.X/toDeleteList 
+            # to avoid a warning about the toDeleteList file
+            print(
+                "To avoid an error about the toDeleteList file, create the directory ~/.beast/2.7 "
+                "and an empty file named toDeleteList in that directory."
+                "On Linux, run mkdir -p ~/.beast/2.7 && touch ~/.beast/2.7/toDeleteList"
+            )
 
     def setup_run_environment(self, env: EnvironmentModifications) -> None:
         env.set("BEAST", self.prefix)
@@ -63,4 +72,8 @@ class Beast2(Package):
             template_dir = "templates"
         else:
             template_dir = "fxtemplates"
+            
+        if spec.satisfies("@2.7.0:"):
+            install_tree("jre", join_path(self.prefix, "jre"))
+
         install_tree(template_dir, join_path(self.prefix, template_dir))


alalazo:
Could you improve the title?

repos/spack_repo/builtin/packages/beast2/package.py
```diff
@@ -50,6 +50,15 @@ def patch(self):
...
                     )
+                    
+        if self.spec.satisfies("@2.7.0:"):
+            # print a message that indicates you need to create /home/${USER}/.beast/2.X/toDeleteList 
+            # to avoid a warning about the toDeleteList file
+            print(
+                "To avoid an error about the toDeleteList file, create the directory ~/.beast/2.7 "
+                "and an empty file named toDeleteList in that directory."
+                "On Linux, run mkdir -p ~/.beast/2.7 && touch ~/.beast/2.7/toDeleteList"
+            )
```
tldahlgren, line 62: Is there a reason to not automatically create the directory and file before it is needed IF it isn't present?

snehring:
@adnan-umich do you know what exactly it's looking for in the jre directory? afaik that's a bundle jre that we're trying to avoid since we build our own java.

adnan-umich:
@snehring the program looks for a jre folder, which contains the following files and directories; 

 ._bin  ._conf  ._DISCLAIMER  ._legal  ._lib  ._OPENJFX_LICENSE  ._readme.txt  ._release  ._Welcome.html
 bin    conf    DISCLAIMER    legal    lib    OPENJFX_LICENSE    readme.txt    release    Welcome.html

These were files introduced in the newer versions of beast2. I noticed an older install didn't have these file requirements.

snehring:
Can you share the errors you're getting and what you're doing to trigger them? I've tested a few of the included examples and they seem to complete without issue with the package as is.

adnan-umich:
Here's how I can reproduce it. 
```
ml spack/0.23
spack install beast2

spack spec ./spec.json
[+]  beast2@2.7.4%gcc@13.2.0+beagle build_system=generic arch=linux-rhel8-skylake_avx512
[+]      ^gcc-runtime@13.2.0%gcc@13.2.0 build_system=generic arch=linux-rhel8-skylake_avx512
[e]      ^glibc@2.28%gcc@13.2.0 build_system=autotools arch=linux-rhel8-skylake_avx512
[+]      ^javafx@20.0.1%gcc@13.2.0 build_system=generic arch=linux-rhel8-skylake_avx512
[+]      ^libbeagle@3.1.2%gcc@13.2.0~cuda~opencl build_system=autotools cuda_arch=none arch=linux-rhel8-skylake_avx512
[e]          ^autoconf@2.69%gcc@13.2.0 build_system=autotools patches=7793209 arch=linux-rhel8-skylake_avx512
[e]          ^automake@1.16.1%gcc@13.2.0 build_system=autotools arch=linux-rhel8-skylake_avx512
[e]          ^gmake@4.2.1%gcc@13.2.0~guile build_system=generic patches=ca60bd9,fe5b60d arch=linux-rhel8-skylake_avx512
[e]          ^libtool@2.4.6%gcc@13.2.0 build_system=autotools arch=linux-rhel8-skylake_avx512
[e]          ^m4@1.4.18%gcc@13.2.0+sigsegv build_system=autotools patches=3877ab5,fc9b616 arch=linux-rhel8-skylake_avx512
[e]          ^pkgconf@1.4.2%gcc@13.2.0 build_system=autotools arch=linux-rhel8-skylake_avx512
[e]          ^subversion@1.10.2%gcc@13.2.0+apxs+nls~perl+serf build_system=autotools arch=linux-rhel8-skylake_avx512
[+]      ^openjdk@17.0.11_9%gcc@13.2.0 build_system=generic certs=none arch=linux-rhel8-skylake_avx512
```

Then after this installation - (there isn't a jre folder in the installation directory).`packagemanager` fails with the following error: `/home/adnanzai/.spack/opt/spack/gcc-13.2.0/beast2/2.7.4-snhq/bin/packagemanager: line 33: /home/adnanzai/.spack/opt/spack/gcc-13.2.0/beast2/2.7.4-snhq/jre/bin/java: No such file or directory`

But If I place the jre folder back into the root installation directory the command works again.

snehring:
It seems to be a few of the scripts are indiscriminately overriding JAVA_HOME. Notably the beast script's logic is incorrect which is why it continues to work without issue. #604 should fix the issue

tldahlgren (commented):
🤔 Thought I submitted this comment days ago.

repos/spack_repo/builtin/packages/beast2/package.py
snehring: I don't believe this is in scope of the spack package, given that the installing user may not be the user that's ultimately running the package. It's also not a path that is touched by the install. If this issue occurs I believe it to be a bug in beast2 rather than the spack recipe.

tldahlgren:
@snehring Any objections then to having this PR merged as is?

snehring:
#604 should supersede this

adnan-umich:
If #604 solves the java issue then I don't think we need this PR anymore.

tldahlgren:
> If #604 solves the java issue then I don't think we need this PR anymore.

@adnan-umich Can you confirm #604 works for you?

snehring:
> If #604 solves the java issue then I don't think we need this PR anymore.

Can you test #604 to verify it works for you? You could also just modify the wrapper scripts (eg packagemanager) as is to remove the line that sets JAVA_HOME

snehring:
@alecbcs this should be resolved by #604, I believe this PR can be closed.

PR #270, opened 2025-06-17
bowtie: add 2.5.4

nboelte:


tldahlgren (approved):
Sha256 is confirmed.

tldahlgren:
Interesting. I re-ran the individual stack that had failed and it passed but it is not reflected in an update to this PRs checks. Perhaps because the original run is 3 weeks old?  @zackgalbreath

tldahlgren:
> Interesting. I re-ran the individual stack that had failed and it passed but it is not reflected in an update to this PRs checks. Perhaps because the original run is 3 weeks old? @zackgalbreath

Nvm. Does appear to have been a delay.

PR #271, opened 2025-06-17
cp2k: Fix basis molopt uzh in 25.01

nboelte:
This fixes a typo in a data file of cp2k. It is important for one of our users, but I'm not familiar enough with CP2K to know if its important enough to be included in spack-packages or not.

### DIFF
--- a/repos/spack_repo/builtin/packages/cp2k/package.py
+++ b/repos/spack_repo/builtin/packages/cp2k/package.py
@@ -503,6 +503,13 @@ class Cp2k(MakefilePackage, CMakePackage, CudaPackage, ROCmPackage):
     # https://github.com/electronic-structure/SIRIUS/pull/1048
     patch("sirius-api-7.7.0.patch", when="@2024.2:2025.1 ^sirius@7.7.0")
 
+    # Fix missing S in data/BASIS_MOLOPT_UZH
+    patch(
+        "https://github.com/cp2k/cp2k/commit/da03128481adf8f78a8a04ebeae0490480c03b89.patch?full_index=1",
+        sha256="0d542c414216866953c95e642d2590b3d313717dfaebbf12cfafedbdd3bf54a3",
+        when="@=2025.1",
+    )
+
     def patch(self):
         # Patch for an undefined constant due to incompatible changes in ELPA
         if self.spec.satisfies("@9.1:2022.2 +elpa"):
--- a/repos/spack_repo/builtin/packages/cp2k/package.py
+++ b/repos/spack_repo/builtin/packages/cp2k/package.py
@@ -504,6 +504,7 @@ class Cp2k(MakefilePackage, CMakePackage, CudaPackage, ROCmPackage):
     patch("sirius-api-7.7.0.patch", when="@2024.2:2025.1 ^sirius@7.7.0")
 
     # Fix missing S in data/BASIS_MOLOPT_UZH
+    # https://github.com/cp2k/cp2k/pull/4140
     patch(
         "https://github.com/cp2k/cp2k/commit/da03128481adf8f78a8a04ebeae0490480c03b89.patch?full_index=1",
         sha256="0d542c414216866953c95e642d2590b3d313717dfaebbf12cfafedbdd3bf54a3",


tldahlgren:
Added missing maintainers to reviewer list.

mtaillefumier (approved):

alalazo:
Failures in pipelines, e.g.: https://gitlab.spack.io/spack/spack-packages/-/jobs/17108176

mtaillefumier:
the patch and checksum are correct.

repos/spack_repo/builtin/packages/cp2k/package.py
```diff
@@ -464,6 +464,14 @@ class Cp2k(MakefilePackage, CMakePackage, CudaPackage, ROCmPackage):
     # https://github.com/electronic-structure/SIRIUS/pull/1048
     patch("sirius-api-7.7.0.patch", when="@2024.2:2025.1 ^sirius@7.7.0")
 
+    # Fix missing S in data/BASIS_MOLOPT_UZH
+    # https://github.com/cp2k/cp2k/pull/4140
+    patch(
+        "https://github.com/cp2k/cp2k/commit/da03128481adf8f78a8a04ebeae0490480c03b89.patch?full_index=1",
+        sha256="0d542c414216866953c95e642d2590b3d313717dfaebbf12cfafedbdd3bf54a3",
+        when="@=2025.1",
```
mtaillefumier, line 511: when="@2025.1"

mtaillefumier (changes requested):

abussy (approved):

repos/spack_repo/builtin/packages/cp2k/package.py
RMeli: I think this is technically correct. `@2025.1` is a version range, so it will also match `@2025.1.1` etc, while `@=2025.1` is the specific version `2025.1` only. I don't think CP2K does patch releases, so I don't mind either way.

hfp (approved):

repos/spack_repo/builtin/packages/cp2k/package.py
mtaillefumier: interesting that the syntax `"@="` is allowed. I always thought that `when="@number" ` was sufficient.

mtaillefumier (approved):

mtaillefumier:
spack does not find the patch but the patch exists and the checksum is also correct.

my checksum: `0d542c414216866953c95e642d2590b3d313717dfaebbf12cfafedbdd3bf54a3`
checksum in the recipe: `0d542c414216866953c95e642d2590b3d313717dfaebbf12cfafedbdd3bf54a3`

repos/spack_repo/builtin/packages/cp2k/package.py
nboelte: The syntax is sometimes necessary. At least in earlier spack versions, `gromacs@2023+plumed` will try to install gromacs version 2023.3, which is incompatible with +plumed and fails, so it must be specified as `gromacs@=2023+plumed`. Of course a less strange versioning scheme of gromacs would also have prevented that problem.

I used it here because i only tested that the patch works for 2025.1 - but as @RMeli said it is probably the same either way in this particular case.

nboelte: If you still want it changed, please let me know, i definitely don't feel strongly about it.

tldahlgren (approved):

tldahlgren:
> spack does not find the patch but the patch exists and the checksum is also correct.
> 
> my checksum: `0d542c414216866953c95e642d2590b3d313717dfaebbf12cfafedbdd3bf54a3` checksum in the recipe: `0d542c414216866953c95e642d2590b3d313717dfaebbf12cfafedbdd3bf54a3`

Details:

```
==> Error: Couldn't find patch for package builtin.cp2k with sha256: 0d542c414216866953c95e642d2590b3d313717dfaebbf12cfafedbdd3bf54a3. This usually means the patch was modified or removed. To fix this, either reconcretize or use the original package repository
  File "/builds/spack/spack-packages/spack/lib/spack/spack/spec.py", line 3643, in patches
    patch = index.patch_for_package(sha256, pkg_cls)
  File "/builds/spack/spack-packages/spack/lib/spack/spack/patch.py", line 492, in patch_for_package
    raise spack.error.PatchLookupError(
spack.error.PatchLookupError: Couldn't find patch for package builtin.cp2k with sha256
```

Which IS weird because the download of the patch with `?index=1` is the provided sha256.

mtaillefumier (approved):
It is a typo in one file with well defined patch so it should work.

tldahlgren:
Afraid CI is still failing on `cp2k` because of the patch (see https://gitlab.spack.io/spack/spack-packages/-/pipelines/1127370).

mtaillefumier:
the checksum are correct so I do not understand why the ci/cd fails for that patch. I do not see anything wrong with the additional code either.

tldahlgren:
@nboelte Can you rebase this PR to develop and see if that unblocks the problem?

This PR's CI is relying on an old version of Spack (Jun 10th) and there were patch-related changes merged on Jun 19 that *may* help (can't be certain). 🤞

mtaillefumier (approved):

nboelte:
Seems like the rebase did the job. Thanks everyone for your patience!

PR #281, opened 2025-06-17
plink: add 1.9-beta7.7

nboelte:
The new plink version has a higher version of the bundled zlib library, so there is also a new version of the patch.


tldahlgren (changes requested):
Confirmed the commit exists but, since there are now multiple version directives using the `git` argument, can you switch to using the `git` property instead?

nboelte:
Good catch! I moved the git repository to the property.

tldahlgren (approved):
Thanks.

PR #284, opened 2025-06-17
spfft: enable nvpl

RMeli:
Manual port of PR https://github.com/spack/spack/pull/49334, which was never merged.


tldahlgren (approved):
The only difference between this raw file and that of the original -- approved by @AdhocMan -- is the inclusion of the build system imports.

PR #290, opened 2025-06-17
Added build_index64

MBlaschek:
as found in the CMAKElists.txt of netlib-lapack

```cmake
# By default build index32 library
option(BUILD_INDEX64 "Build Index-64 API libraries" OFF)
if(BUILD_INDEX64)
  set(BLASLIB "blas64")
  set(CBLASLIB "cblas64")
  set(LAPACKLIB "lapack64")
  set(LAPACKELIB "lapacke64")
  set(TMGLIB "tmglib64")
  set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -DWeirdNEC -DLAPACK_ILP64 -DHAVE_LAPACK_CONFIG_H")
  set(FORTRAN_ILP TRUE)
else()
  set(BLASLIB "blas")
  set(CBLASLIB "cblas")
  set(LAPACKLIB "lapack")
  set(LAPACKELIB "lapacke")
  set(TMGLIB "tmglib")
endif()
```
Should not impact the default behavior, only allow to build the index64 libraries.

related to this: https://github.com/Reference-LAPACK/lapack/issues/461


repos/spack_repo/builtin/packages/netlib_lapack/package.py
```diff
@@ -97,6 +97,7 @@ class NetlibLapack(CMakePackage):
     variant("shared", default=True, description="Build shared library version")
     variant("pic", default=True, description="Produce position-independent code")
     variant("external-blas", default=False, description="Build lapack with an external blas")
+    variant("index64", default=False, description="Build lapack with index-64 integers")
```
alalazo, line 100: Can we use `ilp64` as a name, to be consistent with `openblas`?

MBlaschek: not sure what you mean?
The libraries will be called: `liblapack64.so, liblapacke64.so` in lib64 directory.

MBlaschek: oh, now i understand. yes of course! the variant name.

PR #295, opened 2025-06-17
CI: Use set destination to override repo for gitlab CI

kwryankrattiger:
Hopefully this fixes CI and actually test the repo. The override config was not being considered correctly and the head of develop was being cloned instead of using the spack-package PR.

### DIFF
--- a/.ci/gitlab/.gitlab-ci.yml
+++ b/.ci/gitlab/.gitlab-ci.yml
@@ -21,7 +21,6 @@ default:
   - echo ${SPACK_CHECKOUT_VERSION}
   - git fetch --depth 1 origin ${SPACK_CHECKOUT_VERSION}
   - git checkout FETCH_HEAD
-  - spack repo set --destination "${CI_PROJECT_DIR}" --scope site builtin
   - cd ..
 
 ########################################
@@ -123,6 +122,7 @@ default:
     - cat /proc/meminfo | grep 'MemTotal\|MemFree' || true
     - *clone_spack
     - . "./spack/share/spack/setup-env.sh"
+    - spack repo set --destination "${CI_PROJECT_DIR}" --scope site builtin
   after_script:
     - cat /proc/loadavg || true
     - cat /proc/meminfo | grep 'MemTotal\|MemFree' || true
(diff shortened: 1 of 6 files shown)


.ci/gitlab/.gitlab-ci.yml
```diff
@@ -21,6 +21,7 @@ default:
   - echo ${SPACK_CHECKOUT_VERSION}
   - git fetch --depth 1 origin ${SPACK_CHECKOUT_VERSION}
   - git checkout FETCH_HEAD
+  - spack repo set --destination "${CI_PROJECT_DIR}" --scope site builtin
```
alalazo:
```
. "./spack/share/spack/setup-env.sh"
```

alalazo:
Can you explain:

> The override config was not being considered correctly and the head of develop was being cloned instead of using the spack-package PR.

Where did you see this? I have been working on PRs lately that modify pipelines etc. and we were always using the code from the PR :thinking:

zackgalbreath:
> Can you explain:
> 
> > The override config was not being considered correctly and the head of develop was being cloned instead of using the spack-package PR.
> 
> Where did you see this? I have been working on PRs lately that modify pipelines etc. and we were always using the code from the PR 🤔

Here's two relatively recent examples I noticed that prompted this investigation:

- https://gitlab.spack.io/spack/spack-packages/-/jobs/17102330#L106 (generate job)
- https://gitlab.spack.io/spack/spack-packages/-/jobs/17102618#L78 (rebuild job)

In both cases, the evidence is the following line in the logs.

```
Cloning into '<snip>/tmp/_user_cache/package_repos/fncqgg4'...
```

alalazo:
Hmm, do we know why this doesn't always happen? I just checked two recent PRs of mine, and can't see that line. For instance:
- https://gitlab.spack.io/spack/spack-packages/-/jobs/17114418

Also, it would be good to have a flag that prevents cloning a new repo, and just errors out if a repo isn't found where expected. We should add something like that to `spack/spack`.

zackgalbreath:
> Hmm, do we know why this doesn't always happen? I just checked two recent PRs of mine, and can't see that line. For instance:
> 
> * https://gitlab.spack.io/spack/spack-packages/-/jobs/17114418

Thanks for pointing this out. The pipeline you linked above is for a PR that bumps the version of Spack we use in CI here. Looking more closely at the diff between these two versions of Spack, my hunch is that https://github.com/spack/spack/pull/50886 fixes the issue this PR is attempting to address.

haampie:
> https://gitlab.spack.io/spack/spack-packages/-/jobs/17102330#L106 (generate job)
> https://gitlab.spack.io/spack/spack-packages/-/jobs/17102618#L78 (rebuild job)

These are both older than https://github.com/spack/spack-packages/commit/095fae4b1cdb30b0f628b7b9c5c5a9a0aa81ef8c, irrelevant.

PR #296, opened 2025-06-17
README: Update Spack Package Repo Checkout Instructions

johnwparent:
Follow up to https://github.com/spack/spack-packages/pull/88

#88 is very helpful, however it does suggest the users make a change to their config that will persist after the one-off use case the operation is suggested to be used in the context of. 

This PR just adds instructions on how to undo the change.

Alternatively we could:

* Have the users add their local `spack-packages` checkout as another repo e.g. `test` (would that work? Can spack prefer repos?)
* Instead of refactoring the instructions, just rephase the use case to "If you want Spack to use your local spack-packages instead of its own checkout going forward..." rather than suggesting the behavior is one-off/temporary.

I think the un-do instruction is useful, but would probably be better lived in core spack's docs, so I don't really have a strong opinion which of these options we go with, just that we rephase this slightly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@ simply change Spack's default package repo from the default cache location to th
 spack repo set --scope user --destination /path/to/local/spack-packages builtin
 ```
 
-This updates `~/.spack/repos.yaml` to look like:
+This updates `~/.spack/repos.yaml` config to look like:
 
 ```yaml
 repos:
--- a/README.md
+++ b/README.md
@@ -14,16 +14,21 @@ If you want to test your package changes locally before submitting a pull reques
 simply change Spack's default package repo from the default cache location to the full path to your local git clone:
 
 ```shell
-spack repo set --destination /path/to/local/spack-packages builtin
+spack repo set --scope user --destination /path/to/local/spack-packages builtin
 ```
 
-Your Spack installation will be configured to pull from your local Spack
-packages checkout in perpetuity, meaning any updates/other management of the Spack
-packages repo will now be manual. To undo this, simply remove the local modification
-done to the repos configuration or run
+This updates `~/.spack/repos.yaml` to look like:
+
+```yaml
+repos:
+  builtin:
+    destination: /path/to/local/spack-packages
+```
+
+To undo this, you can run:
 
 ```shell
-spack config remove repos:builtin:destination
+spack config --scope user remove repos:builtin:destination
 ```
 
 `$spack` can be used to form a relative path to your Spack root directory.
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@ We run continuous integration on this repository to test builds of a large numbe
 Spack packages.
 
 If you want to test your package changes locally before submitting a pull request,
-simply configure Spack's default package repo to the full path to your local git clone:
+simply configure Spack's default package repo to the full path of your local git clone:
 
 ```shell
 spack repo set --scope user --destination /path/to/local/spack-packages builtin
--- a/README.md
+++ b/README.md
@@ -11,11 +11,21 @@ We run continuous integration on this repository to test builds of a large numbe
 Spack packages.
 
 If you want to test your package changes locally before submitting a pull request,
-simply change Spack's default package repo from the default cache location to the full
-path to your local git clone:
-```
+simply change Spack's default package repo from the default cache location to the full path to your local git clone:
+
+```shell
 spack repo set --destination /path/to/local/spack-packages builtin
 ```
+
+Your Spack installation will be configured to pull from your local Spack
+packages checkout in perpetuity, meaning any updates/other management of the Spack
+packages repo will now be manual. To undo this, simply remove the local modification
+done to the repos configuration or run
+
+```shell
+spack config remove repos:builtin:destination
+```
+
 `$spack` can be used to form a relative path to your Spack root directory.
 
 If you are migrating your pull requests from
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@ We run continuous integration on this repository to test builds of a large numbe
 Spack packages.
 
 If you want to test your package changes locally before submitting a pull request,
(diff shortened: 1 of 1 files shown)


johnwparent:
@Chrismarsh @tgamblin

Chrismarsh:
Having duplicate packages in two repos seems confusing.

IMO I like "after testing, to revert to default do x,y,z" would be simplest and most clear. 

Having details on how to revert in two places seems ok? I often struggle to find what I want in the docs because the details are in a similar but different spot. Maybe that's a skill issue though.

adamjstewart:
I'm one of those frequent contributors who want to make this change and then never undo it, so for me this PR doesn't really matter. I'll let others decide whether and how to add undo instructions.

haampie:
I don't get it, why would you revert it?

johnwparent:
> I don't get it, why would you revert it?

Personally? I wouldn't. 
But for users who aren't frequent contributors, there might not be interest in maintaining their own checkout of the spack-packages repository beyond testing infrequent package PRs, and thus, a desire to undo this setting once they've verified local changes. 

I don't see the harm in providing instructions for how to undo setting the builtin repo to a local checkout rather than a core managed checkout.

Chrismarsh:
Another use case would be someone pulling a PR to test a fix, but then doesn't have the desire to contribute etc.

README.md
```diff
@@ -11,11 +11,21 @@ We run continuous integration on this repository to test builds of a large numbe
...
+
+Your Spack installation will be configured to pull from your local Spack
+packages checkout in perpetuity, meaning any updates/other management of the Spack
+packages repo will now be manual. To undo this, simply remove the local modification
+done to the repos configuration or run
+
+```shell
+spack config remove repos:builtin:destination
+```
+
```
tgamblin: I think keeping it simple and just giving the users commands here is best -- I don't think we need to get into the various use cases or explain too much.  They can look up the commands if they feel so inclined.

````suggestion
spack repo set --scope user --destination /path/to/local/spack-packages builtin
```

This updates `~/.spack/repos.yaml` to look like:

```yaml
repos:
  builtin:
    destination: /path/to/local/spack-packages
```

To undo this, you can run:

```shell
spack config --scope user remove repos:builtin:destination
```

````

**Review by tgamblin — CHANGES_REQUESTED** [review]

File: README.md (hunk 1 above)
Comment (johnwparent) [reply README.md]: I was thinking that since we explain that this is the mechanism to test your changes locally above, that we should at least make them aware this will have more of an impact than a one off test use case might suggest.

Comment (tgamblin) [reply README.md]: If we tell them it modifies their user config I do not see how that can be lost on them

Comment (johnwparent) [reply README.md]: Makes sense, accepted the commit and made a few small tweaks to earlier phrasing.

**Review by bernhardkaindl — APPROVED** [review]
I think there's nothing left to let this waiting.

The added information is much needed as well.

**Review by bernhardkaindl — DISMISSED** [review]
@tgamblin: @johnwparent accepted your commit, it would be ready to merge with changing your changes requested state to approval:
> Makes sense, accepted the commit and made a few small tweaks to earlier phrasing.

File: README.md [hunk 2]
```diff
@@ -11,11 +11,26 @@ We run continuous integration on this repository to test builds of a large numbe
 Spack packages.
 
 If you want to test your package changes locally before submitting a pull request,
-simply change Spack's default package repo from the default cache location to the full
-path to your local git clone:
+simply configure Spack's default package repo to the full path to your local git clone:
```
tldahlgren:
```suggestion
simply configure Spack's default package repo to the full path of your local git clone:
```

README.md
```diff
@@ -11,11 +11,26 @@ We run continuous integration on this repository to test builds of a large numbe
 Spack packages.
 
 If you want to test your package changes locally before submitting a pull request,
-simply change Spack's default package repo from the default cache location to the full
-path to your local git clone:
+simply configure Spack's default package repo to the full path of your local git clone:
+
+```shell
+spack repo set --scope user --destination /path/to/local/spack-packages builtin
```
tldahlgren, line 17: Why is `--scope user` required?

README.md
```diff
@@ -11,11 +11,26 @@ We run continuous integration on this repository to test builds of a large numbe
...
+```yaml
+repos:
+  builtin:
+    destination: /path/to/local/spack-packages
 ```
+
+To undo this, you can run:
+
+```shell
+spack config --scope user remove repos:builtin:destination
```
tldahlgren, line 31: You can simply run `spack repo remove builtin` (see https://spack.readthedocs.io/en/latest/repositories.html#spack-repo-remove) and it will remove `builtin` from `$HOME/.spack/repos.yaml`.

README.md
tldahlgren, line 17: Why is `--scope user` provided?

I don't need it to have my `$HOME/.spack/repos.yaml` updated.

tldahlgren: This example is explicitly provided in https://spack.readthedocs.io/en/latest/repositories.html#spack-repo-set and it doesn't include `--scope user`. 

I recommend being consistent with the `spack/spack` docs (to the degree possible).

tldahlgren (changes requested):
The user scope isn't needed. `spack repo` commands are sufficient. See the comments.

(FWIW. There's also new material in the tutorial at https://spack-tutorial.readthedocs.io/en/latest/tutorial_packaging.html#getting-started that includes links to the docs for `spack repo` and the repository structure.)

README.md
tldahlgren: If there are no other repositories listed, it will result in `$HOME/.spack/repos.yaml`:

```
repos: {}
```

*but* that's okay as long as the default repos.yaml has a builtin:

```
$ spack config blame repos
---                                                                 repos:
$SPACK_ROOT/etc/spack/defaults/repos.yaml:14    builtin:
$SPACK_ROOT/etc/spack/defaults/repos.yaml:15      git: https://github.com/spack/spack-packages.git
```

PR #298, opened 2025-06-17
Windows: Initial attempt to stand up oneAPI

johnwparent:
Spun from https://github.com/spack/spack/pull/50264

 * adds installation support on Windows (only viable if you're an administrator at the moment)
 * allows oneapi compilers to be externally detected on Windows
 * allows oneapi compilers to be used with MSVC compilers on Windows
 * Ensures proper ordering of oneAPI and msvc environment setup by injecting MSVC as a oneapi dependency. *

* This isn't actually true from a strict sense, but if MSVC is going to be used in the same environment as MSVC, MSVC's setup environment methods _need_ to be run first. Any suggestions for improvement there are more than welcome.

### DIFF
--- a/repos/spack_repo/builtin/build_systems/compiler.py
+++ b/repos/spack_repo/builtin/build_systems/compiler.py
@@ -219,7 +219,7 @@ class CompilerPackage(PackageBase):
 
 @memoized
 def _compiler_output(
-    compiler_path: Path, *, version_argument: str, ignore_errors: Tuple[int, ...] = ()
+    compiler_path: Path, *, version_argument: Optional[str], ignore_errors: Tuple[int, ...] = ()
 ) -> str:
     """Returns the output from the compiler invoked with the given version argument.
 
@@ -243,7 +243,7 @@ def _compiler_output(
 
 
 def compiler_output(
-    compiler_path: Path, *, version_argument: str, ignore_errors: Tuple[int, ...] = ()
+    compiler_path: Path, *, version_argument: Optional[str], ignore_errors: Tuple[int, ...] = ()
 ) -> str:
     """Wrapper for _get_compiler_version_output()."""
     # This ensures that we memoize compiler output by *absolute path*,
--- a/repos/spack_repo/builtin/build_systems/oneapi.py
+++ b/repos/spack_repo/builtin/build_systems/oneapi.py
@@ -48,6 +48,7 @@ class IntelOneApiPackage(Package):
         conflicts(c, msg="This package in only available for x86_64 and Linux/Windows")
 
     # env mods are mandatory on Windows, no variant option
+    # and oneAPI is currently unsupported on Darwin
     for plat in ["linux", "freebsd"]:
         # Add variant to toggle environment modifications from vars.sh
         variant(
--- a/repos/spack_repo/builtin/build_systems/oneapi.py
+++ b/repos/spack_repo/builtin/build_systems/oneapi.py
@@ -48,12 +48,13 @@ class IntelOneApiPackage(Package):
         "target=ppc64le:",
         "target=aarch64:",
         "platform=darwin",
-        "platform=windows",
     ]:
-        conflicts(c, msg="This package in only available for x86_64 and Linux")
+        conflicts(c, msg="This package in only available for x86_64 and Linux/Windows")
 
-    # Add variant to toggle environment modifications from vars.sh
-    variant("envmods", default=True, description="Toggles environment modifications")
+    # env mods are not an option on Windows AFAICT
+    for plat in ["linux", "freebsd"]:
+        # Add variant to toggle environment modifications from vars.sh
+        variant("envmods", default=True, description="Toggles environment modifications", when=f"platform={plat}")
 
     @staticmethod
     def update_description(cls):
--- a/repos/spack_repo/builtin/build_systems/oneapi.py
+++ b/repos/spack_repo/builtin/build_systems/oneapi.py
@@ -140,15 +140,16 @@ class IntelOneApiPackage(Package):
         elif sys.platform == "win32":
             # Installer can be executed directly
             # as it's a standalone executable
-            installer_extraction_path = os.path.join(self.stage.source_path, f"installer-{os.path.basename(self.component_dir)}"),
+            installer_extraction_path = os.path.join(self.stage.source_path, f"installer-{os.path.basename(self.component_dir)}")
             installer = Executable(installer_path)
             installer_args = [
-                "--extract-folder",
-                "--remove-extracted-files",
-                installer_extraction_path,
-                "yes",
-                "-a",
                 "-s",
+                "--extract-folder",
+                installer_extraction_path,
+                "--remove-extracted-files",
+                "yes",
+                "--a",
+                "--s",
                 "--eula",
                 "accept",
                 "--install-dir",
--- a/repos/spack_repo/builtin/build_systems/oneapi.py
+++ b/repos/spack_repo/builtin/build_systems/oneapi.py
@@ -5,8 +5,8 @@
 import os
 import platform
 import shutil
-from os.path import basename, isdir
 import sys
+from os.path import basename, isdir
 
 from llnl.util.link_tree import LinkTree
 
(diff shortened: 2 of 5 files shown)


repos/spack_repo/builtin/packages/msvc/package.py
```diff
@@ -55,7 +54,7 @@ def install(self, spec, prefix):
     # compiler wrappers
     compiler_wrapper_link_paths = {"c": "", "cxx": "", "fortran": ""}
```
scheibelp: Should fortran also be omitted here? That might be more of a tidyness concern than a correctness one.

And/or should `intel-oneapi-compilers` be doing this for `fortran` on `windows`?

repos/spack_repo/builtin/build_systems/oneapi.py
```diff
@@ -43,17 +44,18 @@ class IntelOneApiPackage(Package):
...
-        "platform=windows",
-    ]:
-        conflicts(c, msg="This package in only available for x86_64 and Linux")
-
-    # Add variant to toggle environment modifications from vars.sh
-    variant("envmods", default=True, description="Toggles environment modifications")
+    for c in ["target=ppc64:", "target=ppc64le:", "target=aarch64:", "platform=darwin"]:
+        conflicts(c, msg="This package in only available for x86_64 and Linux/Windows")
+
+    # env mods are mandatory on Windows, no variant option
```
scheibelp, line 50: Could this comment be updated to also mention that MacOS is excluded entirely? It might be worth including it in the loop here even though other constraints forbid it (just to avoid explaining that).

repos/spack_repo/builtin/build_systems/oneapi.py
```diff
@@ -150,13 +173,17 @@ def setup_run_environment(self, env: EnvironmentModifications) -> None:
...
            $ source {prefix}/{component}/{version}/env/vars.sh
         """
-        # Only if environment modifications are desired (default is +envmods)
-        if "+envmods" in self.spec:
-            env.extend(
-                EnvironmentModifications.from_sourcing_file(
-                    self.component_prefix.env.join("vars.sh"), *self.env_script_args
+        if sys.platform == "win32":
+            # env mods can't be conditional on Windows
+            pass
```
scheibelp, line 179: Sorry I'm not seeing where this is happening (I would have assumed that `vars.sh` would be sourced here, just not conditioned on `+envmods` - is that handled elsewhere? I don't see it in the diff.

repos/spack_repo/builtin/packages/intel_oneapi_compilers/package.py
```diff
@@ -351,13 +79,30 @@ class IntelOneapiCompilers(IntelOneApiPackage, CompilerPackage):
...
-    debug_flags = ["-debug", "-g", "-g0", "-g1", "-g2", "-g3"]
-    opt_flags = ["-O", "-O0", "-O1", "-O2", "-O3", "-Ofast", "-Os"]
+    debug_flags = [
+        DebugFlags.DEBUG,
+        DebugFlags.G,
+        DebugFlags.G0,
+        DebugFlags.G1,
+        DebugFlags.G2,
+        DebugFlags.G3,
+    ]
```
scheibelp: Could you have the `Debug/OptFlags` classes define a method that automatically expands this, e.g. `debug_flags = DebugFlags.values`? I'm not sure if there will be direct analogs for each individual flag across both operating systems, and I worry that if new flags are added it would be easy to miss here.

repos/spack_repo/builtin/build_systems/compiler.py
```diff
@@ -219,7 +219,7 @@ def _fortran_path(self) -> Optional[str]:
 
 @memoized
 def _compiler_output(
-    compiler_path: Path, *, version_argument: str, ignore_errors: Tuple[int, ...] = ()
+    compiler_path: Path, *, version_argument: Optional[str], ignore_errors: Tuple[int, ...] = ()
```
scheibelp, line 222: A couple things:

* Looking at the body of this, there appears to be a sort-of optional mechanism already because it checks `if not version_argument` and then invokes `compiler()` without it if so (I could see an argument made for refactoring that, but for now, IMO we might as well follow what this allows for)
* (question) given prior point, and that you set the version arg to "" for windows, I would think this wouldn't have required overriding `determine_version`, but did an error occur when you didn't?

repos/spack_repo/builtin/packages/intel_oneapi_compilers/oneapi_versions.py
```diff
@@ -0,0 +1,473 @@
...
+                    name="amd-plugin-installer",
+                    placement="amd-plugin-installer",
+                    when=f"@{v.version}",
+                    expand=False,
+                    url=v.amd.url,
+                    sha256=v.amd.sha,
+                )
+
+
+class oneAPIInstaller:
```
scheibelp, line 52: I think `InstallResource` might be clearer than `oneAPIInstaller` (I think `oneapi` can be dropped because other packages won't `import *` from this). A short docstring would also be helpful (e.g. "For oneapi resource locations that are not complete, prepends common resource location")

repos/spack_repo/builtin/packages/intel_oneapi_compilers/oneapi_versions.py
```diff
@@ -0,0 +1,473 @@
...
+        cxx: oneAPIInstaller,
+        fortran: Optional[oneAPIInstaller] = None,
+        nvidia: Optional[oneAPIInstaller] = None,
+        amd: Optional[oneAPIInstaller] = None,
+    ):
+        self.version = version
+        assert platform.lower() in (
+            "windows",
+            "linux",
+        ), "OneAPI is only compatible with Windows or Linux"
```
scheibelp: I thought `freebsd` was ok too

* `darwin` was conflicted although versions were still declared for it
* `freebsd` did not have a conflict (and I assume the versions that install on `linux` would install there too)

I think that would be skipped in

```
        for v in filter(
            lambda x: x.platform == platform.system().lower(), oneAPIVersions.versions
        ):
```

or does `platform.system()` report `"linux"` on `freebsd`? https://forums.freebsd.org/threads/uname-unchanged-after-freebsd-update.87449/  suggests not.

Also I know we have

1. `spack.platform`
2. `sys.platform`
3. `platform.system()`

and there is already precedent for using both of [2-3] in the package repo, although in all our reviews I think you've been using `sys.platform`, so is it possible to use that  here too, or would that have issues?

scheibelp (changes requested):
A few questions/requests

repos/spack_repo/builtin/packages/msvc/package.py
johnwparent: Tidyness wise? Yes, that said I was hesitant to do so for functionality concerns. 
I'll pull this out and correct the `intel-oneapi-compilers`

repos/spack_repo/builtin/build_systems/oneapi.py
johnwparent: Makes sense, will do.

repos/spack_repo/builtin/packages/intel_oneapi_compilers/package.py
johnwparent: Good idea!

repos/spack_repo/builtin/packages/intel_oneapi_compilers/oneapi_versions.py
johnwparent: I favored the more specific `oneAPIInstaller` so future readers of this package recipe don't see the generic name `InstallerResource` and presume that we have some type of generic mechanic to handle something like this. 

Docstring being added.

repos/spack_repo/builtin/packages/intel_oneapi_compilers/oneapi_versions.py
johnwparent: It's possible to use `sys.platform` here as well, but I chose not to deliberately. My thinking was along the lines that assigning the platform to a given instantiation of oneAPIVersions is a user facing API, and having consistent naming scheme (i.e. everything fully spelled out like "linux", "windows", "freebsd" was a little nicer than "win32", "linux", etc) 

The version filter line you link above has already been refactored as we needed to define these versions for mac to pass the audit check. Also missing freebsd was more an oversight than any assumption about generic linux handling.

repos/spack_repo/builtin/build_systems/compiler.py
johnwparent: I find passing `""` as an alias for `None` and relying on unadvertised, subtle behavior for that to be a valid check to be fragile/error prone/hard to read/reason about vs being explicit that the given value is optional. 

For Windows we need to ignore errors from the compiler invocation. The intel and MSVC compilers both have no version argument, the version is determined by invoking the compiler with `LOGO` enabled (default behavior without `/NOLOGO`). The best way to do so is to just invoke the compiler with no arguments, but that returns an error code since the compiler complains about no provided source file.

johnwparent:
@scheibelp Updated based off review comments, ready for another round when you are

repos/spack_repo/builtin/build_systems/oneapi.py
johnwparent: Updating this to use `EnvironmentModifications.from_sourcing_file`. In `setup_dependent_run_environment` , does the `env` attribute include the modifications made by things oneAPI depends on? 
For oneAPI + MSVC, oneAPI needs to be aware of the selected version of MSVC and the SDK. MSVC's setup_environment does this already, so if we can just use the existing `env` to inform oneAPI's setvars of that, that would be ideal.

PR #305, opened 2025-06-18
Vecgeom: add patched 2.0.0-rc.4

sethrj:
v2.0.0-rc.4 fails to configure downstream due to a bug in its cmake. This slightly updated version also includes a few navigation fixes.


sethrj:
Actually I'll just wait on the next rc.

PR #309, opened 2025-06-19
ninja: add version 1.13.0

kshea21:
Add latest version of ninja, 1.13.0, to spack-packages repository.

### DIFF
--- a/repos/spack_repo/builtin/packages/ninja/package.py
+++ b/repos/spack_repo/builtin/packages/ninja/package.py
@@ -37,7 +37,12 @@ class Ninja(Package):
     version("1.8.2", sha256="86b8700c3d0880c2b44c2ff67ce42774aaf8c28cbf57725cb881569288c1c6f4")
     version("1.7.2", sha256="2edda0a5421ace3cf428309211270772dd35a91af60c96f93f90df6bc41b16d9")
     version("1.6.0", sha256="b43e88fb068fe4d92a3dfd9eb4d19755dae5c33415db2e9b7b61b4659009cde7")
-    version("kitware", branch="features-for-fortran", git="https://github.com/Kitware/ninja.git", deprecated=True)
+    version(
+        "kitware",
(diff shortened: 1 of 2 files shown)


kwryankrattiger:
So this also contains the job server now, we can probably remove the Kitware version in favor of preferring 1.13.0

mathomp4:
A query for @kshea21: once this gets in, do we now get More Speed™ for free? (I've been liking the concurrent builds). Or will there need to be internal Spack tweaks for the Ninja GNU Make jobserver support?

msimberg:
> So this also contains the job server now, we can probably remove the Kitware version in favor of preferring 1.13.0

Maybe the "kitware" version can be marked deprecated? I have no personal use for keeping the custom version and favour removing it altogether, but may be smoother for anyone that has been relying on it?

repos/spack_repo/builtin/packages/ninja/package.py
```diff
@@ -26,6 +26,7 @@ class Ninja(Package):
 
     version("kitware", branch="features-for-fortran", git="https://github.com/Kitware/ninja.git")
```
kwryankrattiger: I think it is a good idea to mark this as deprecated since the Ninja upstream supports `pipe` and `fifo`, where as this only supports `pipe`.

```suggestion
    version("kitware", branch="features-for-fortran", git="https://github.com/Kitware/ninja.git", deprecated=True)
```

kshea21:
@mathomp4 there are still some internal spack tweaks that need to be made, I'm working on it over at #50708. We should see More Speed™ for free by 1.1 or sooner in develop. I'm so glad that you've been enjoying concurrent builds!!

kwryankrattiger (approved):

kwryankrattiger:
@spackbot run pipeline

kshea21:
@spackbot run pipeline

kshea21:
@spackbot run pipeline

PR #320, opened 2025-06-19
libdwarf: add versions 0.12 and 2.0 and remove deprecated versions

pranav-sivaraman:
This is tested in CI, so version compatibility bounds may need to be added to other packages.


tldahlgren (approved):
Checksums confirmed in the check (but I can only with `curl` as the `url_fetch_method`). Also confirmed use of all of the affected variants.

pranav-sivaraman:
Oh I can remove the `url_for_version` since I've deprecated the old versions. Let me make a followup PR.

PR #321, opened 2025-06-19
MSVC compiler wrapper package

johnwparent:
Updates the compiler wrapper package to handle
Windows compiler wrappers

Adds builders for the new compiler wrapper flavor

Adds proper env setup for the wrapper

Updates the MSVC package for compiler wrapper
support

Sibling to https://github.com/spack/spack/pull/38495


johnwparent:
cc @scheibelp

johnwparent:
@kwryankrattiger sorry for the ping, feel free to review if you want, but I'm just messing with the stacks to debug, all the stack changes will be reverted before merge

johnwparent:
Trying a "no cache" run to see if that's the cause of our issues

johnwparent:
@spackbot rebuild everything

johnwparent:
Bootstrap failure currently makes sense given we've changed the compiler-wrapper versions available. Need to regenerate the clingo.json once we've cut an official release of the wrapper repo.

haampie:
Does it make sense to merge these two repos? `spack/compiler-wrapper` and this one

johnwparent:
> Does it make sense to merge these two repos? `spack/compiler-wrapper` and this one

It's a good question, I've gone back and forth a few times but I think I've convinced myself the answer is no. Merging would make this recipe a lot cleaner, but there's basically no overlap between the contents of the repos despite their shared purpose so it would probably end up looking something like:
```
compiler-wrapper/
├── windows/
│   ├── src/
│   │   ├── ....
|   | - README.md
│   └── test/
├── unix/
|   |── cc.sh
|   |── README.md
└── README.md
```
Where the projects are effectively split in two just within the same repo.
It's not my preferred layout (although my preference means little here), plus the Windows side has a much busier history and they'll likely be "released" at very different cadences. 

All that to say, while I think they should not be merged, I'm happy to hear an opinion to the contrary.

haampie:
Ok, wouldn't it be cleaner then to put it in a different package altogether? That way package hash changes for msvc things don't affect posix

johnwparent:
> Ok, wouldn't it be cleaner then to put it in a different package altogether? That way package hash changes for msvc things don't affect posix

it would be much cleaner, I assumed there would be issues if we made the compiler-wrapper package a virtual or there were two of them floating around, but if this is as easy as using a virtual or changing every reference to the wrapper's package name (for which this makes sense) to "if IS_WINDOWS: compiler-wrapper-name = "win-wrapper"" then I'm all for it

johnwparent:
@spackbot run pipeline

johnwparent:
Closing to clear PR buildcache in Gitlab CI

johnwparent:
Hopefully the script will have cleared the cache by now

repos/spack_repo/builtin/packages/win_gpg/package.py
scheibelp, line 28: Sorry if I forgot: why is this removed? Does it create a cyclic dependency?

repos/spack_repo/builtin/packages/msmpi/ifort_nd_compat.patch
```diff
@@ -37,7 +37,7 @@ index 24bd29d..319153d 100644
 +      <Fort_Flags>/D _WIN64=1 /D _AMD64_=1 /D AMD64=1</Fort_Flags>
      </PropertyGroup>
 -    <Exec Command="$(GFORTRAN_BIN)\gfortran.exe -I$(MPI_INC_ROOT) -c %(ForCompile.Identity) $(Fort_Flags)  -o $(O)\%(ForCompile.FileName).obj" />
-+    <Exec Command="$(IFORT_BIN)\ifx.exe /I$(MPI_INC_ROOT) /c %(ForCompile.Identity) $(Fort_Flags) /names:lowercase /assume:nounderscore /o $(O)\%(ForCompile.FileName).obj" />
++    <Exec Command="$(IFORT_BIN)\ifx.exe /I&quot;$(MPI_INC_ROOT)&quot; /c %(ForCompile.Identity) $(Fort_Flags) /names:lowercase /assume:nounderscore /o $(O)\%(ForCompile.FileName).obj" />
```
scheibelp: It occurs to me: would this be an opportunity to redirect to a more-generic variable like `$FC`, that would be set by the `msvc` package?

repos/spack_repo/builtin/packages/msvc/package.py
```diff
@@ -264,6 +276,17 @@ def platform_toolset_ver(self):
         vs22_toolset = Version(toolset_ver) > Version("142")
         return toolset_ver if not vs22_toolset else "143"
 
+    @property
+    def ld(self):
+        assert self.spec.concrete, "cannot retrieve C++ linker, spec is not concrete"
+        assert self.spec.external, (
```
scheibelp: I thought packages usually handle this by adding a check in the `install()` method - was there a reason it was done differently for this package?

repos/spack_repo/builtin/packages/compiler_wrapper/package.py
```diff
@@ -172,6 +124,9 @@ def setup_dependent_build_environment(
             compiler = getattr(compiler_pkg, attr_name)
             env.set(spack_var_name, compiler)
 
+            if hasattr(compiler_pkg, "ld"):
+                env.set("SPACK_LD", compiler_pkg.ld)
+
             # -frandom-seed= is needed for deterministic builds with GCC
             if compiler_pkg.name == "gcc" and self.spec.satisfies("@1.1:"):
```
scheibelp: This is being used as a builder mixin, so `self.spec` -> `self.pkg.spec`

repos/spack_repo/builtin/packages/win_gpg/package.py
johnwparent: I think I did this while testing that concretization failure we were trying to diagnose a while back. 
Testing cruft, removed.

repos/spack_repo/builtin/packages/compiler_wrapper/package.py
johnwparent: Right, thanks.

repos/spack_repo/builtin/packages/msmpi/ifort_nd_compat.patch
johnwparent: Yeah, that's probably a good idea. I'll make a followup PR to specifically the MSMPI package for that though. In fact, this change should be pulled out of this PR now that we've validated it works.

repos/spack_repo/builtin/packages/msvc/package.py
johnwparent: I think this is an errant change from my work making visual studio an installable spack package. Reverting.

PR #323, opened 2025-06-20
Fix pipelines to account for new cache layout

alalazo:


haampie:
See https://github.com/spack/spack/pull/50919.

This does not fix uploading the right files.

Also even if you fix that, it's likely that the cache is unused because Spack may be cloned in a different dir during each build.

PR #326, opened 2025-06-20
whizard: Make sure to set the LD_LIBRARY_PATH for run environment

tmadlener:
whizard needs the `LD_LIBRARY_PATH` to be set in order to run in it without issues downstream. This is also done in the `whizard_setup.sh` script that is generated by whizard during its build.

### DIFF
--- a/repos/spack_repo/builtin/packages/whizard/package.py
+++ b/repos/spack_repo/builtin/packages/whizard/package.py
@@ -147,6 +147,12 @@ class Whizard(AutotoolsPackage):
         env.set("FC", self.compiler.fc)
         env.set("F77", self.compiler.fc)
 
+    def setup_run_environment(self, env: EnvironmentModifications) -> None:
+        # Whizard creates a whizard_setup.[c]sh script that sets the
+        # LD_LIBRARY_PATH that is necessary because whizard builds more
+        # libraries on invocation, which are usually not linked via RPATH
+        env.prepend_path("LD_LIBRARY_PATH", self.spec["whizard"].libs.directories[0])
+
     @run_before("autoreconf")
     def prepare_whizard(self):
         # As described in the manual (SVN Repository version)


bernhardkaindl (approved):
@vvolkl As this has no responses yet, I guess there is no objection.

bernhardkaindl (commented):
@tmadlener, see #427. It looks like as @paulgessinger says in https://github.com/spack/spack-packages/pull/427#issuecomment-3162919856, that 

https://github.com/spack/spack-packages/blob/7aff9dd7017886ac70d2f565516bc78124b3d30d/repos/spack_repo/builtin/packages/whizard/package.py#L122-L127

is not effective, the compiler flags determined by whizard's `configure` script still only get `-std=c++11`, even though the detected value here is 20. The flag injection does not seem to be working here.

He points out that whizard's recipe is bypassing spack's compiler wrapper, as he sees direct calls to `/usr/bin/g++` in the log, which I found here:

https://github.com/spack/spack-packages/blob/7aff9dd7017886ac70d2f565516bc78124b3d30d/repos/spack_repo/builtin/packages/whizard/package.py#L141-L148

I think this should be fixed to either use `filter_compiler_wrappers` or, if needed, patch the code of whizard to use the correct compiler calls at runtime.

I think would fix it all properly and LD_LIBRARY_PATH would not be needed, because then, spack's compiler wrappers would pass spack''s `rpath` to the compile of whizard, hence LD_LIBRARY_PATH would then be obsolete to run link the whizard program itself.

bernhardkaindl:
This was added in https://github.com/spack/spack/pull/17300, as it seems (AFAICS using `grep` and `strings`) that the used compilers are apparently in the strings included in libwhizard and other libs and many other files.

Adam commented about it:
> In the long run, it might be a good idea to contact the developers and ask them if there's a way to change the compiler used at run-time. Often there's an env var or file that can be patched.

Maybe, it is enough to just use `filter_compiler_wrappers` on the installed scripts:

```py
grep -r -e 'g[+c][+c]-13"' -e gfortran-13 *
bin/whizard-config:         echo "/usr/bin/gfortran-13"
grep: lib/libwhizard.so.2.0.2: binary file matches
lib/whizard/libtool:LTCC="/usr/bin/gcc-13"
lib/whizard/libtool:CC="/usr/bin/gcc-13"
lib/whizard/libtool:CC="/usr/bin/gfortran-13"
lib/whizard/libtool:CC="/usr/bin/gfortran-13"
lib/whizard/libtool:CC="/usr/bin/g++-13"
```
`grep -r '-D[^ ]-13'` could not find any `-D` flags that pass the compilers, and `grep` could find them in `.h*`, or '*.c*` files, and I could not find any place where the strings would be generated. So I assume that the strings would have been compiled into the binaries using `libtool` to compile Whizard. So I could not find any evidence of a deliberate inclusion of the compilers besides `bin/whizard-config` and `lib/whizard/libtool` that are easy to fix by `filter_compiler_wrappers`.

@vvolkl, you did #17300, could you shed some more light on this (why you did/could not use `filter_compiler_wrappers`)?

tmadlener:
I have asked one of the whizard developers about this to see if we are missing something obvious or whether a fix is needed (either here or in the whizard build).

wdconinc:
Now that #427 is merged, which restricts hepmc3 to 3.3.0 in the HEP stack in CI, can we make sure to remove that version pinning as part of this PR?

wdconinc (approved):

PR #329, opened 2025-06-20
Gotcha 1.0.8 add commit hash

hjjvandam:
Gotcha version 1.0.8 was added to the gotcha package.py without the corresponding commit hash.
This PR fixes that omission.


tldahlgren (approved):
Commit hash has been confirmed.

PR #342, opened 2025-06-23
CI: Make it easier to point to experimental refs

kwryankrattiger:
This requires changes from https://github.com/spack/spack/pull/51117 in order to work correctly.


kwryankrattiger:
Error
```
$ git remote add origin https://github.com/${SPACK_CHECKOUT_REPO}.git
$ git fetch --depth 1 origin ${SPACK_CHECKOUT_VERSION}
remote: Not Found
fatal: repository 'https://github.com/.git/' not found
```

It seems like the variable defined in the dotenv is not being forwarded from the parent pipeline to the child pipeline. This should be covered by specifying that `pipeline_variables` should be forwarded but it is not ([ref](https://docs.gitlab.com/ci/yaml/#triggerforward)).

```
.trigger_build: &trigger_build
  strategy: depend
  forward:
    pipeline_variables: true
```

kwryankrattiger:
Seems fixed!

```
Initialized empty Git repository in /builds/spack/spack-packages/spack/.git/
[32;1m$ git remote add origin https://github.com/${SPACK_CHECKOUT_REPO}.git[0;m
[32;1m$ git fetch --depth 1 origin ${SPACK_CHECKOUT_VERSION}[0;m
From https://github.com/spack/spack
 * branch            refs/pull/51117/head -> FETCH_HEAD
[32;1m$ git checkout FETCH_HEAD[0;m
Note: switching to 'FETCH_HEAD'.
```

kwryankrattiger:
[Passing pipeline](https://gitlab.spack.io/spack/spack/-/pipelines/1207711) in spack/spack

kwryankrattiger:
@spackbot run pipeline

kwryankrattiger:
@spackbot run pipeline

repos/spack_repo/builtin/packages/abseil_cpp/package.py
```diff
@@ -78,6 +78,7 @@ class AbseilCpp(CMakePackage):
     )
 
     variant("shared", default=True, description="Build shared instead of static libraries")
+    variant("xx", default=True, description="Build shared instead of static libraries")
```
becker33: This looks like it's leftover from testing

.ci/gitlab/scripts/common/write_dotenv.py
```diff
@@ -0,0 +1,48 @@
...
+    for line in file:
+        if "=" in line:
+            key, value = line.rsplit("=", 1)
+            dotenv[key] = value.strip()
+    return dotenv
+
+
+def write_dotenv(env, file):
+    for var in dotenv:
+        fd.write(f"{var}={dotenv[var]}\n")
```
becker33:
```suggestion
    for key, value in env.items():
        file.write(f"{key}={value}\n")
```
mismatch between `fd` and `file`, but also changed how we iterate for better readability.

.ci/gitlab/.gitlab-ci.yml
```diff
@@ -186,10 +209,12 @@ dotenv:
     KUBERNETES_MEMORY_REQUEST: "5M"
   artifacts:
     reports:
-      dotenv: .ci/env
+      dotenv: env
   tags: [ "spack", "service" ]
   script:
-  - cat .ci/env
+  - curl -LSsO https://raw.githubusercontent.com/spack/spack-packages/refs/pull/342/head/.ci/gitlab/scripts/common/write_dotenv.py
+  - python3 write_dotenv.py .ci/env
```
becker33: Why not just do this relative to `$SPACK_PACKAGES_ROOT`?

.ci/gitlab/scripts/common/write_dotenv.py
```diff
@@ -0,0 +1,48 @@
...
+        print(f"Found base dotenv: {base_dotenv}")
+        with open(base_dotenv, "r", encoding="utf-8") as fd:
+            dotenv = read_dotenv(fd)
+
+    repo_override = os.environ.get("CI_PROJECT_NAME")
+    if repo_override:
+        if repo_override.endswith("packages"):
+            configure_packages_repo(dotenv)
+        else:
+            configure_core_repo(dotenv)
```
becker33: I'm not seeing where these variables are set in the non-override case

becker33 (dismissed):

repos/spack_repo/builtin/packages/abseil_cpp/package.py
kwryankrattiger: Yeah, I need to repush once https://github.com/spack/spack/pull/51219 is merged to update the spack ref so it is just hanging out for now

.ci/gitlab/.gitlab-ci.yml
kwryankrattiger: ~Yeah I can probably do that now. In the beginning that variable didn't exist~

That variable is set by this script, and we may not have it when this runs.

.ci/gitlab/scripts/common/write_dotenv.py
kwryankrattiger: I am not sure how this was working...

.ci/gitlab/scripts/common/write_dotenv.py
kwryankrattiger: Good catch!

.ci/gitlab/scripts/common/write_dotenv.py
```diff
@@ -0,0 +1,50 @@
...
+    else:
+        configure_packages_repo(dotenv)
```
becker33:
```suggestion
    if repo_override and not repo_override.endswith("packages"): 
        configure_core_repo(dotenv)
    else:
        configure_packages_repo(dotenv)
```
It might also be better to validate that `CI_PROJECT_NAME` is valid and use the full string instead of `endswith`

.ci/gitlab/.gitlab-ci.yml
becker33: What about something like this to ensure you get the current version from whichever source is appropriate? I really don't like hardcoding the url to the original PR in here.

```suggestion
  - dotenv_py=$(ls $CI_PROJECT_DIR/.ci/gitlab/scripts/common/write_dotenv.py 2>/dev/null || ls $CU_AUX_DIR/.ci/gitlab/scripts/common/write_dotenv.py 2>/dev/null)
  - python3 $dotenv_py .ci/env
```

.ci/gitlab/scripts/common/write_dotenv.py
kwryankrattiger: I am not too overly concerned about this script being very robust right now since it is only going to be run by our CI. If there is interest by others to use it for some reason then I think it would be worth cleaning up and making more robust.

.ci/gitlab/.gitlab-ci.yml
kwryankrattiger: So, like I said, we don't know where this script is in the context of this job. It is possible it is running in the spack/spack context or the spack/spack-packages context and the point is the script determines that and then sets up the subsequent jobs accordingly.

The reason it points to the PR right now is it does not yet exist on develop. Once this is merged we can point it to look at the `HEAD` of develop. I will put in a comment similar to what @alecbcs has for the triage job.

becker33: That's why my proposal was to check for the script in each context, rather than pulling it from a third source.

.ci/gitlab/.gitlab-ci.yml
```diff
@@ -160,7 +183,7 @@ default:
     - (Get-WmiObject Win32_PhysicalMemory | measure-object Capacity -sum).sum/1kb
     - $ErrorActionPreference=$ErrorActionOld
     - *clone_spack
-    - . .\spack\share\spack\setup-env.ps1
+    - . $SPACK_ROOT\share\spack\setup-env.ps1
```
johnwparent, line 186: This is set in the environment, not in the shell right?
If so needs to be:
```suggestion
    - . $Env:SPACK_ROOT\share\spack\setup-env.ps1
```

.ci/gitlab/.gitlab-ci.yml
johnwparent: > Once this is merged we can point it to look at the HEAD of develop.

I'm not really a fan of this either, pulling from a commit where we don't have a guarantee this file wasn't modified unexpectedly for a given PR context (not likely, but it could happen) seems like a recipe for future headaches. 

If we're _never_ using this script from Spack-core/Spack-packages directly, why not put it in a separate repo/gist/whatever rather than have to deal with sync or location issues?

Alternatively is `.ci/gitlab/scripts/common/write_dotenv.py` this the same for both repos? We should know from Gitlab where the clone root is. 

IIRC `.pre` is run after the clone, so that should work right?

kwryankrattiger: `CI_AUX_DIR` is determined by that script, it is a chicken-egg problem. This entire step is to setup the context of the pipeline for all of the downstream jobs.

The reason for curl'ing the file directly is to avoid the alternative, which would be to clone the entire develop branch of packages to a known location and then call it from there.

.ci/gitlab/.gitlab-ci.yml
```diff
@@ -14,29 +14,50 @@ default:
...
   - echo ${SPACK_CHECKOUT_VERSION}
   - git fetch --depth 1 origin ${SPACK_CHECKOUT_VERSION}
   - git checkout FETCH_HEAD
-  - cd ..
-
-########################################
+  - cd ${CI_PROJECT_DIR}
+
+.trigger_build: &trigger_build
+  strategy: depend
```
johnwparent, line 31: This is nice

.ci/gitlab/scripts/common/write_dotenv.py
kwryankrattiger: The default of what happens here now is nothing is configured/the base dot env is written as a pass-through. I think that is the correct behavior.

.github/actions/checkout-spack/action.yml
```diff
@@ -22,17 +22,18 @@ runs:
       run: |
         source .ci/env
         echo "version=${SPACK_CHECKOUT_VERSION}" >> $GITHUB_OUTPUT
+        echo "gh-repo=${SPACK_CHECKOUT_REPO}" >> $GITHUB_OUTPUT
     - name: Checkout Spack
       uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683
       with:
-        repository: spack/spack
+        repository: ${{ steps.spack-version.outputs.gh-repo}}
```
johnwparent, line 29: I believe you need to explicitly denote these outputs somewhere using the `outputs:` key.

johnwparent: See: https://docs.github.com/en/actions/reference/workflows-and-actions/metadata-syntax#outputs-for-composite-actions

johnwparent (commented):
All but one of my comments are non-blocking, looks great!

.github/actions/checkout-spack/action.yml
kwryankrattiger: ah, I forgot this was still in this PR, I don't think it should be anymore since we should never need to use it this way.

.ci/gitlab/.gitlab-ci.yml
kwryankrattiger: So, I am not a huge fan of this either but some of the alternatives I had thought of...
* using a gist
* requiring the script to exist and be mounted for all runners
* doing the same thing, but cloning the entire repo
* encoding the script into a base64 string, then decoding it to a file and execute it with python

We can change this to pull from a commit, but again we can't really change that until _after_ this is merged so that the script lives on a develop commit in `spack/spack-packages`. This is supposed to only be a temporary solution so we can beat the chicken-egg problem.

If it is a huge deal, I can break this script out into another PR, merge that, and then we reference that commit here.

johnwparent: As long as we're not pointing at this PR or an endpoint that can change without deliberate intent (or do one of those things now and fix it as soon as this lands), I'm happy, opinions beyond that I will leave to other folks.

(I know I suggested baking into runner images but I was like 80% unserious, that seems like a massive pain to update)

.github/actions/checkout-spack/action.yml
kwryankrattiger: actually, for passing between steps this is fine. you only need to define outputs if they are passed out of the action/job

kwryankrattiger: and..the failure I was seeing was a bad ref. Not sure where I got that one from.

kwryankrattiger:
CI failures unrelated to this PR. See [develop failing](https://gitlab.spack.io/spack/spack-packages/-/pipelines/1256802) for cabana/cray

@becker33 or @johnwparent either one of you open to merging this?

zackgalbreath (approved):

tgamblin (approved):

johnwparent (approved):

PR #350, opened 2025-06-24
py-torchgeo: add v0.7.1

ashnair1:
https://github.com/microsoft/torchgeo/releases/tag/v0.7.1

### DIFF
--- a/repos/spack_repo/builtin/packages/py_torchgeo/package.py
+++ b/repos/spack_repo/builtin/packages/py_torchgeo/package.py
@@ -18,6 +18,7 @@ class PyTorchgeo(PythonPackage):
     maintainers("adamjstewart", "calebrob6")
 
     version("main", branch="main")
+    version("0.7.1", sha256="05f645868a6dff083d4d0529662bde1b502e1f33ef260ebc735065e05d84176e")
     version("0.7.0", sha256="4ba0e96ea826080f393b1bb719a3f8c364637112710b1ac38c56b9590a638e29")
     version("0.6.2", sha256="82f49f0d18d2c22cc70fc0690641e8dd60e4904a9c50d32c79ebd5020ac10fa7")
     version("0.6.1", sha256="38c930917ea341d05a7a611ff74c017f29482df7455d50e287ea79dec7d0a14b")
@@ -46,6 +47,7 @@ class PyTorchgeo(PythonPackage):
 
     # Required dependencies
     with default_args(type="build"):
+        depends_on("py-setuptools@77.0.1:", when="@0.7.1:")
         depends_on("py-setuptools@61:", when="@0.5:")
         depends_on("py-setuptools@42:")
 
--- a/repos/spack_repo/builtin/packages/py_torchgeo/package.py
+++ b/repos/spack_repo/builtin/packages/py_torchgeo/package.py
@@ -15,7 +15,7 @@ class PyTorchgeo(PythonPackage):
     git = "https://github.com/microsoft/torchgeo.git"
 
     license("MIT")
-    maintainers("adamjstewart", "calebrob6")
+    maintainers("adamjstewart", "calebrob6", "ashnair1")
 
     version("main", branch="main")
     version("0.7.1", sha256="05f645868a6dff083d4d0529662bde1b502e1f33ef260ebc735065e05d84176e")


repos/spack_repo/builtin/packages/py_torchgeo/package.py
```diff
@@ -18,6 +18,7 @@ class PyTorchgeo(PythonPackage):
     maintainers("adamjstewart", "calebrob6")
```
adamjstewart: You can also add yourself to the list of maintainers if you want. These people get notified to review PRs when the package recipe is updated.

repos/spack_repo/builtin/packages/py_torchgeo/package.py
```diff
@@ -44,6 +45,7 @@ class PyTorchgeo(PythonPackage):
 
     # Required dependencies
     with default_args(type="build"):
+        depends_on("py-setuptools@77.0.1:", when="@0.7.1:")
         depends_on("py-setuptools@61:", when="@0.5:")
         depends_on("py-setuptools@42:")
```
adamjstewart, line 53: I actually made a mistake down below. It should actually be:
```python
depends_on("py-segmentation-models-pytorch@0.5:", when="@0.7.1:")
# https://github.com/microsoft/torchgeo/pull/2740
depends_on("py-segmentation-models-pytorch@0.3.3:0.4", when="@0.7.0")
depends_on("py-segmentation-models-pytorch@0.2:0.4", when="@:0.6")
```
with `when` being added to the last line. Can you fix this?

repos/spack_repo/builtin/packages/py_torchgeo/package.py
```diff
@@ -99,6 +101,7 @@ class PyTorchgeo(PythonPackage):
         depends_on("py-segmentation-models-pytorch@0.5:", when="@0.7.1:")
         # https://github.com/microsoft/torchgeo/pull/2740
         depends_on("py-segmentation-models-pytorch@0.3.3:0.4", when="@0.7.0")
+        depends_on("py-segmentation-models-pytorch@0.2:0.4", when="@:0.6")
         depends_on("py-segmentation-models-pytorch@0.2:0.4")
```
adamjstewart: You'll have to delete this line, otherwise all lines will be merged and the result will be that all versions require SMP 0.2–0.4.

adamjstewart (approved):
Looks good now, let's see if CI passes.

ashnair1:
Seems like two versions of `setuptools` are installed in the gitlab runner (`79.0.1` and `63.4.3`) and the latter is being used for building the package and raises an error with the license field as it's not compatible.

adamjstewart:
Pinged some people to investigate this.

adamjstewart:
@ashnair1 can you rebase and see if the problem magically goes away?

adamjstewart (approved):

adamjstewart:
Still seeing the same issue. Setuptools 80.9.0 is the only version in `CMAKE_PREFIX_PATH`, but `PYTHONPATH` has both setuptools 63.4.3 and 80.9.0. There is either a bug in the concretizer or in one of the package recipes causing setuptools 63.4.3 to be added to `PYTHONPATH`. @alalazo @haampie can you investigate?

alalazo:
I could reproduce this issue on:

* **Spack:** 1.1.0.dev0 (https://github.com/spack/spack/commit/99bc917a4f7ddf468478d2ce963d4240c6ad3a32)
* **Builtin repo:** https://github.com/spack/spack-packages/commit/c9635b5fe1d11412c1d70cdc547c936f5870cead
* **Python:** 3.13.5
* **Platform:** darwin-sequoia-m1

It seems the question "Why do we get an old version of `py-networkx` in the DAG ?" is answered by our optimization criteria, and by the amount of constraints the Python ecosystem imposes on packages. 

One thing that we might want to rework in our optimization criteria is a penalty given when a node uses 2 or more compilers:
```
opt_criterion(46, "number of compilers used on the same node").
#minimize{ 0@246: #true }.
#minimize{ 0@46: #true }.
#minimize{
    Penalty@46+Priority,PackageNode
    : compiler_penalty(PackageNode, Penalty), build_priority(PackageNode, Priority)
}.
```
On Darwin, this tries to minimize the number of nodes that are built with `%fortran=gcc` with a certain priority. 

Besides that, Python packages are often so constrained that we are forced for a node to bring in dependencies that are older than their latest known version. Sometimes penalties sum up such that taking a big penalty on that node version is "better" than taking no penalty on that node, but bring in X dependencies, each one with a small penalty on their versions :melting_face: 

The workaround for this would be to pin the desired versions, but I agree that the outcome is not what one would expect. I'll have a look if adding a `max` kind of criteria on versions slows down concretization.

adamjstewart:
Same concretizer/PYTHONPATH bug in #695 and #726

adamjstewart:
@ashnair1 do you want to add v0.7.2 now that it's out? If not, @isaaccorley can also do it.

ashnair1:
> @ashnair1 do you want to add v0.7.2 now that it's out? If not, @isaaccorley can also do it.

@adamjstewart I'll add v0.7.2 once this is merged

adamjstewart:
Also, #2340 should finally fix CI here. It forces us to use Python 3.13+, therefore numpy 2.1+, therefore newer setuptools. We still have no idea why CI was picking ancient versions of all 3.

adamjstewart:
Also need to update dependencies for 0.7.2 and add conflicts on lightning

adamjstewart (changes requested):
Still need to update minimum supported versions for many dependencies

adamjstewart (approved):
We'll do 0.7.2 in a separate PR.

PR #353, opened 2025-06-24
vtk: add latest version v9.5.0

vicentebolea:
🎉 🎉 🎉

VTK 9.5.0 is out

This PR updates the VTK Spack package to include this new release and set VTK v9.5.0 as the preferred version.

Let me know if we need to change anything else in the package.

🎉 🎉 🎉

### DIFF
--- a/repos/spack_repo/builtin/packages/vtk/package.py
+++ b/repos/spack_repo/builtin/packages/vtk/package.py
@@ -26,10 +26,12 @@ class Vtk(CMakePackage):
     license("BSD-3-Clause")
 
     version(
-        "9.4.1",
-        sha256="c253b0c8d002aaf98871c6d0cb76afc4936c301b72358a08d5f3f72ef8bc4529",
+        "9.5.0",
+        sha256="04ae86246b9557c6b61afbc534a6df099244fbc8f3937f82e6bc0570953af87d",
         preferred=True,
     )
+
+    version("9.4.1", sha256="c253b0c8d002aaf98871c6d0cb76afc4936c301b72358a08d5f3f72ef8bc4529")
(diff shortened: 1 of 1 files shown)


vicentebolea:
VTK builds: https://gitlab.spack.io/spack/spack-packages/-/pipelines/1134775

johnwparent (changes requested):
Looks like only e4s got rebuilt in Gitlab, I'd like to see at least a windows build with this version run. 

You can do that by editing:
`.ci\gitlab\stacks\windows-vis\spack.yaml` 
to use

```yaml
  specs:
  - "vtk@9.5.0 ~mpi"
```
instead of
```yaml
  specs:
  - "vtk@9: ~mpi"
 ```

I'd make a commitable suggestion, but somehow in 2025, Github's review tools are still super limited.

johnwparent:
Thanks @vicentebolea! Appreciate that. We'll need to revert that commit before we merge but we'll know it builds on Windows so IMO worth it.

vicentebolea:
@johnwparent unfortunately we cannot concretize vtk 9.5.0 in windows due to it depending on the a seacas version which conflicts with Windows. We could in this case use the internal seacas in VTK

johnwparent:
> @johnwparent unfortunately we cannot concretize vtk 9.5.0 in windows due to it depending on the a seacas version which conflicts with Windows. We could in this case use the internal seacas in VTK

That's alright, it explains the lack of concretization to the preferred version. I'd rather first check to see if I can make that seacas work on Windws (or explore other seacas' VTK could potentially work with). 
Let's just revert the Windows CI VTK version change commit and then LGTM.

Thanks for investigating!

johnwparent (approved):
LGTM, thanks @vicentebolea

PR #356, opened 2025-06-25
py-numpy: Fix atomic_load const issue for Fujitsu compiler

otsukay:
### Summary

This PR adds a conditional patch to `py-numpy` that fixes a build error when using the Fujitsu compiler (`%fj`) on A64FX systems.

### Problem

When building `py-numpy` with `%fj`, compilation fails due to the use of `const _Atomic(...) *` in calls to `atomic_load`. The Fujitsu compiler (based on old LLVM) rejects this usage with an error like:

### Update
Adjusted the patch condition:
```
- patch("fix-atomic-const.patch", when="%fj")
+ patch("fix-atomic-const.patch", when="@2.1.0: %fj")
```
because the file the patch targets was added in @2.1.0.
Thanks again for your time.


tldahlgren (dismissed):
@adamjstewart @rgommers Any objections to this PR being merged?

adamjstewart:
Has this patch been submitted upstream?

rgommers (approved):
This fix seems fine to me. If it's not fixed in any compiler version and not yet reported to Fujitsu, that would be useful to do.

adamjstewart:
CI errors seem unrelated, rust is pulling in whatever it wants from the internet and it seems the latest version of something broke.

adamjstewart (approved):

tldahlgren:
Let's see if `Update branch` helps before trying something else.

PR #357, opened 2025-06-25
openmpi: add variant to use cray-xpmem name for package configure when cray-xpmem enabled

biddisco:
on cray systems the xpmem package configure file is named 
```
/usr/lib64/pkgconfig/cray-xpmem.pc
```
or similar, this PR adds a avariant that changes the xpmem search from `xpmem` to `cray-xpmem` and thus allows the build to proceed

### DIFF
--- a/repos/spack_repo/builtin/packages/openmpi/package.py
+++ b/repos/spack_repo/builtin/packages/openmpi/package.py
@@ -598,6 +598,8 @@ with '-Wl,-commons,use_dylibs' and without
 '-Wl,-flat_namespace'.""",
     )
 
+    variant("cray-xpmem", default=False, description="use cray-xpmem instead of xpmem configure flag (if fabrics=xpmem enabled)")
+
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/openmpi/package.py
```diff
@@ -598,6 +598,8 @@ class Openmpi(AutotoolsPackage, CudaPackage):
 '-Wl,-flat_namespace'.""",
     )
 
+    variant("cray-xpmem", default=False, description="use cray-xpmem instead of xpmem configure flag (if fabrics=xpmem enabled)")
```
msimberg: I think the comment can be put into a `when`?
```suggestion
    variant("cray-xpmem", default=False, when="fabrics=xpmem", description="use cray-xpmem instead of xpmem configure flag")
```

hppritcha (approved):

msimberg (approved):

PR #358, opened 2025-06-25
angsd: add version 0.940

downloadico:
Setting HTSSRC to "systemwide" so that angsd doesn't try to build its own htslib.

### DIFF
--- a/repos/spack_repo/builtin/packages/angsd/package.py
+++ b/repos/spack_repo/builtin/packages/angsd/package.py
@@ -16,8 +16,9 @@ class Angsd(MakefilePackage):
     """
 
     homepage = "https://github.com/ANGSD/angsd"
-    url = "https://github.com/ANGSD/angsd/archive/0.935.tar.gz"
+    url = "https://github.com/ANGSD/angsd/archive/0.940.tar.gz"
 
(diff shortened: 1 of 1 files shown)


johnwparent (commented):
Is there a way to set a specific path to htslib for the build? 
If not then LGTM. If so, could you add it?

downloadico:
Hi John,

I don't see such a mechanism.

Thanks,
Jim

johnwparent:
>  I don't see such a mechanism

From reading the makefile it looks like https://github.com/ANGSD/angsd/blob/master/Makefile#L38 points to an instruction on how to point `angsd` to a local install of `htslib`.

downloadico:
I'll take another look.

Thanks,
Jim

downloadico:
Hi John,

That's what I thought at first too.  However, when HTSSRC is set, the Makefile assumes that you've got libhts.a and libhts.h in that directory, and not in lib/libhts.a and include/libhts.h like how it is when libhts is installed.

Jim

johnwparent:
Hmmm, that's unfortunate, thanks for trying that out. IMO this makefile should probably be patched to conform with how Spack is installing `htslib` but if you're not interested in that approach then LGTM.

downloadico:
Hi John,

This Makefile is a mess IMHO.  I'm not really interested in fixing it.

Jim

becker33 (approved):
Generally agree with @johnwparent

PR #369, opened 2025-06-26
acts-algebra-plugins: add v0.28.0

stephenswat:
This commit adds ACTS algebra plugins version 0.28.0.

### DIFF
--- a/repos/spack_repo/builtin/packages/acts_algebra_plugins/package.py
+++ b/repos/spack_repo/builtin/packages/acts_algebra_plugins/package.py
@@ -18,6 +18,7 @@ class ActsAlgebraPlugins(CMakePackage):
 
     license("MPL-2.0", checked_by="stephenswat")
 
+    version("0.28.0", sha256="d798ba2129bf922f54627233ef947b8bb2345db9199e3868cc48bc1da86d5f15")
     version("0.27.0", sha256="c2081b399b7f4e004bebd5bf8250ed9596b113002fe445bca7fdac24d2c5932c")
     version("0.26.2", sha256="0170f22e1a75493b86464f27991117bc2c5a9d52554c75786e321d4c591990e7")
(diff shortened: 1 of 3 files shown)


repos/spack_repo/builtin/packages/acts_algebra_plugins/package.py
```diff
@@ -18,6 +18,7 @@ class ActsAlgebraPlugins(CMakePackage):
 
     license("MPL-2.0", checked_by="stephenswat")
 
+    version("0.28.0", sha256="d798ba2129bf922f54627233ef947b8bb2345db9199e3868cc48bc1da86d5f15")
```
wdconinc, line 21: Because the internal version number of algebra-plugins [was only updated](https://github.com/acts-project/algebra-plugins/compare/v0.27.0...v0.28.0#diff-1e7de1ae2d059d21e1dd75d5812d5a34b0222cef273b7c3a2af62eb747f9d20aL9) with this version upgrade, from 0.23 to 0.28, it means that `acts +traccc` and `detray` will still fail to build with versions e.g. 0.27 allowed by spack since CMake will flag them as insufficient. See:
https://github.com/spack/spack-packages/blob/d6e6a1eceedc6129f05259dd7fda7958e2336418/repos/spack_repo/builtin/packages/acts/package.py#L371
and
https://github.com/spack/spack-packages/blob/d6e6a1eceedc6129f05259dd7fda7958e2336418/repos/spack_repo/builtin/packages/detray/package.py#L93

Can you update those dependencies between 0.24 and 0.27 to all require 0.28, with a comment indicating why?

stephenswat: Done, thanks!

wdconinc (dismissed):

wdconinc (approved):

PR #371, opened 2025-06-26
Update of CP2K and its dependencies

mtaillefumier:
This PR contains various updates in cp2k dependencies. 

- CP2K
    - tblite library
    - SIRIUS modules
         - Added dftd4 and dftd3 support
         - Added nlcglib support
         - Added variable cell optimization
    - Added Mi300 support 
    - Added the possibility to compile dbcsr cpu only even when `+cuda` or `+rocm` are specified
    - Updated to the latest version of cp2k 2025.2
- multicharge
    - Add toml-f library
    - Update mctc-lib version
    - Update multicharge version
    - Add jonquil dependency
    - Add missing depends_on("fortran") in multicharge
- tblite: Added cmake build system
- DBCSR
    - Added Mi300 support


repos/spack_repo/builtin/packages/mctc_lib/package.py
```diff
@@ -23,6 +23,27 @@ class MctcLib(MesonPackage, CMakePackage):
...
+        sha256="1ead2810a0295484b5fc78a9b017f9e14152fda66ad0cecdf24e733ec2a38498",
+    )
+    version(
+        "0.4.0-gnu-12-macos-latest",
+        sha256="fc337b2b46bd664446b866d1c7a1d5cdfb1837e66d114e0d9fc6fbc1e7b47be5",
+    )
+    version(
+        "0.4.0-gnu-10-ubuntu-latest",
+        sha256="74fb14b1b2f7410cbeed0ebb854acf484989462baff46d9268f6591e83c992e7",
+    )
```
awvwgk: I think those should not be included here.

awvwgk:
```suggestion
```

repos/spack_repo/builtin/packages/mctc_lib/package.py
```diff
@@ -23,6 +23,27 @@ class MctcLib(MesonPackage, CMakePackage):
     build_system("cmake", "meson", default="meson")
 
     version("main", branch="main")
+
+    version("0.4.2", sha256="ce1e962c79d871d3705be590aef44f07ca296843b85e164307290f8324769406")
```
awvwgk, line 27: From 0.4.2 the JSON dependency changes from json-fortran to jonquil/toml-f

mtaillefumier: updated. Thanks

mtaillefumier:
@spackbot fix style

mtaillefumier:
@spackbot fix style

repos/spack_repo/builtin/packages/jonquil/package.py
```diff
@@ -0,0 +1,44 @@
...
+    license("Apache-2.0")
+
+    build_system("cmake", "meson", default="meson")
+
+    version("main", branch="main")
+    version("0.3.0", sha256="4bc3f0ae47ac2e009a0dc733ad9d0f16db4dfed13b50f58b9a06bb3a579eec47")
+    version("0.2.0", sha256="68448be7f399942e15a05ed7a149cc226a8ee81a8ce66cd68a2d01d9fc86527e")
+    version("0.1.0", sha256="0c8854da047306cad357143fe56f7afe3d323d89aa7383b6614b2b587f580044")
+
+    depends_on("fortran", type="build")  # generated
```
RMeli:
```suggestion
    depends_on("fortran", type="build")
```

repos/spack_repo/builtin/packages/mctc_lib/package.py
```diff
@@ -23,24 +23,39 @@ class MctcLib(MesonPackage, CMakePackage):
...
+            self.define_from_variant("WITH_OpenMP", "openmp"),
+        ]
 
 
 class MesonBuilder(meson.MesonBuilder):
     def meson_args(self):
-        return ["-Djson={0}".format("enabled" if "+json" in self.spec else "disabled")]
+        return [
+            "-Djson={0}".format("enabled" if "+json" in self.spec else "disabled"),
+            "-Dopenmp={0}".format("true" if "json" in self.spec else "false"),
```
RMeli:
```suggestion
            "-Dopenmp={0}".format("true" if "+openmp" in self.spec else "false"),
```
Is `true` the correct value as opposed as `enabled`/`disabled`?

repos/spack_repo/builtin/packages/toml_f/package.py
```diff
@@ -2,30 +2,46 @@
...
 
-    homepage = "https://toml-f.readthedocs.io"
-    url = "https://github.com/toml-f/toml-f/releases/download/v0.3.1/toml-f-0.3.1.tar.xz"
-    git = "https://github.com/toml-f/toml-f.git"
+    homepage = "https://toml-f.readthedocs.io/"
+    url = "https://github.com/toml-f/toml-f/releases/download/v0.4.2/toml-f-0.4.2.tar.xz"
+    git = "https://github.com/toml-f/toml-f/"
 
-    maintainers("awvwgk")
+    maintainers("mtaillefumier")
```
RMeli:
```suggestion
    maintainers("awvwgk", "mtaillefumier")
```
? Unless @awvwgk wants to be removed...

repos/spack_repo/builtin/packages/mstore/package.py
```diff
@@ -0,0 +1,49 @@
...
+    build_system("cmake", "meson", default="meson")
+
+    version("main", branch="main")
+    version("0.3.0", sha256="8cbae54a47339de0f47457d7d0931fb5ac23a6cfc8b54872b925528cf5138523")
+    version("0.2.0", sha256="932ec27cb327f7bfcdc316ae7d39a13cff03b95946bda79b52fd2fa43c4fd4d4")
+
+    variant("openmp", default=True, description="Use OpenMP parallelisation")
+
+    depends_on("c", type="build")  # generated
+    depends_on("fortran", type="build")  # generated
```
RMeli:
```suggestion
    depends_on("c", type="build")
    depends_on("fortran", type="build")
```

repos/spack_repo/builtin/packages/cp2k/package.py
```diff
@@ -225,8 +229,8 @@ class Cp2k(MakefilePackage, CMakePackage, CudaPackage, ROCmPackage):
     depends_on("greenx", when="+greenx")
     depends_on("hdf5+hl+fortran", when="+hdf5")
     depends_on("trexio", when="+trexio")
-    depends_on("libvdwxc", when="+vdwxc")
```
RMeli, line 236: The variant is still needed to explicitly set `CP2K_USE_LIBVDWXC` but the dependency is taken care of by SIRIUS, is this correct?

repos/spack_repo/builtin/packages/cp2k/package.py
```diff
@@ -337,18 +341,21 @@ class Cp2k(MakefilePackage, CMakePackage, CudaPackage, ROCmPackage):
...
         depends_on("sirius+openmp", when="+openmp")
         depends_on("sirius~openmp", when="~openmp")
         depends_on("sirius@7.3:", when="@9.1")
-        depends_on("sirius@7.4:7.5", when="@2023.2")
+        depends_on("sirius@7.4:", when="@2023.2")
         depends_on("sirius@7.5:", when="@2024.1:")
-        depends_on("sirius@7.6:7.7 +pugixml", when="@2024.2:")
-        depends_on("sirius@7.7: +pugixml", when="@2025.2:")
+        depends_on("sirius@7.6:+pugixml", when="@2024.2:2025.2")
+        depends_on("sirius@7.7:+pugixml", when="@2025.2:")
```
RMeli: Ranges are inclusive, so these overlap. 
```suggestion
        depends_on("sirius@7.6:+pugixml", when="@2024.2:")
        depends_on("sirius@7.7:+pugixml", when="@2025.2:")
```
?

repos/spack_repo/builtin/packages/dbcsr/package.py
```diff
@@ -197,6 +204,7 @@ def cmake_args(self):
                 "gfx90a": "Mi250",
                 "gfx90a:xnack-": "Mi250",
                 "gfx90a:xnack+": "Mi250",
+                "gfx942": "Mi250",
```
RMeli:
```suggestion
                "gfx942": "Mi300",
```
?

repos/spack_repo/builtin/packages/dftd4/package.py
```diff
@@ -56,6 +56,8 @@ class Dftd4(MesonPackage, CMakePackage):
         depends_on(f"mctc-lib build_system={build_system}", when=f"build_system={build_system}")
         depends_on(f"multicharge build_system={build_system}", when=f"build_system={build_system}")
 
+    depends_on("mctc-lib@0.3", when="@:3.8")
+    depends_on("multicharge@0.3", when="@:3.8")
```
RMeli, line 60: It depends only on these specific versions? Or it should be a version range?

repos/spack_repo/builtin/packages/dbcsr/package.py
mtaillefumier: The dbcsr build system does not support Mi300 yet for various reasons. the `"gfx942": "Mi250",` is needed here :(

repos/spack_repo/builtin/packages/cp2k/package.py
mtaillefumier: true. Thanks for fixing this.

repos/spack_repo/builtin/packages/dbcsr/package.py
RMeli: Ah, I see. In this case I think it would be better to have a conflict on `amdgpu_target=gfx942` then.

repos/spack_repo/builtin/packages/dftd4/package.py
mtaillefumier: The versions of the dependencies are locked unfortunately. The developers also changed the api of these libraries which lead to compilation failure for new versions of these packages.

repos/spack_repo/builtin/packages/dbcsr/package.py
mtaillefumier: I am reluctant to add a conflict for this as the only thing that this parameter controls is the file containing the optimized parameters for each GPU and the architecture during JIT time. 
The lack of optimized parameters for `Mi300` should not block users from using `dbcsr`. The only difference will be a performance "penalty" if any.  Ideally, the build system should take care of this but it is not yet present in the main repository. 
I also want to be able to compile cp2k on mi300 without having to modify the recipe locally as I am forced to do right now. 
I acknowledge though it is not the best solution.

RMeli: > The lack of optimized parameters for Mi300 should not block users from using dbcsr.

Users will not be blocked to use `dbcsr`, but they will have to set `dbcsr amdgpu_target=gfx90a`. In this way, you will not magically change the meaning of setting `amdgpu_target=gfx942` once the architecture is properly supported. The user would be consciously using the parameters for another architecture (you can also add this suggestion to the conflict message).

Something like
```
cp2k +rocm amdgpu_target=gfx942 ^dbcsr +rocm amdgpu_target=gfx90a
```
should work, without the need to change the recipe.

mtaillefumier: added a conflict and a message suggesting to use `gfx90a` when dbcsr is compiled for mi300. 
NB: Using parameters from older architecture should not lead to a bad behavior otherwise it means the code is buggy.

repos/spack_repo/builtin/packages/dbcsr/package.py
```diff
@@ -108,6 +108,12 @@ class Dbcsr(CMakePackage, CudaPackage, ROCmPackage):
     dbcsr_amdgpu_targets = ("gfx906", "gfx910", "gfx90a", "gfx90a:xnack-", "gfx90a:xnack+")
     amd_msg = f"DBCSR supports these AMD gpu targets:  {', '.join(dbcsr_amdgpu_targets)}"
 
+    conflicts(
+        "+rocm",
+        when="amdgpu_target=gfx942",
+        msg="DBCSR does not support gfx942. Please use amdgpu_target=gfx90a",
+    )
+
```
RMeli: Actually, I did not realize this before from your PR, but I think the conflict below already takes care of this, since `gfx942` is not added to `dbcsr_amdgpu_targets`?

repos/spack_repo/builtin/packages/dbcsr/package.py
```diff
@@ -106,13 +106,7 @@ class Dbcsr(CMakePackage, CudaPackage, ROCmPackage):
...
 
     dbcsr_amdgpu_targets = ("gfx906", "gfx910", "gfx90a", "gfx90a:xnack-", "gfx90a:xnack+")
-    amd_msg = f"DBCSR supports these AMD gpu targets:  {', '.join(dbcsr_amdgpu_targets)}"
-
-    conflicts(
-        "+rocm",
-        when="amdgpu_target=gfx942",
-        msg="DBCSR does not support gfx942. Please use amdgpu_target=gfx90a",
-    )
+    amd_msg = f"DBCSR supports these AMD gpu targets:  {', '.join(dbcsr_amdgpu_targets)}. Use amdgpu_target=gfx942 to enable gfx942 (mi300) support in DBCSR"
```
RMeli:
```suggestion
    amd_msg = f"DBCSR supports these AMD GPU targets:  {', '.join(dbcsr_amdgpu_targets)}. Set 'amdgpu_target' explicitly to one of the supported targets."
```
? I think this would be a bit more general than just for Mi300.

RMeli (dismissed):
LGTM. I'll let @awvwgk comment on the DFTD and DFTD-related packages.

repos/spack_repo/builtin/packages/jonquil/package.py
```diff
@@ -0,0 +1,44 @@
...
+
+    version("main", branch="main")
+    version("0.3.0", sha256="4bc3f0ae47ac2e009a0dc733ad9d0f16db4dfed13b50f58b9a06bb3a579eec47")
+    version("0.2.0", sha256="68448be7f399942e15a05ed7a149cc226a8ee81a8ce66cd68a2d01d9fc86527e")
+    version("0.1.0", sha256="0c8854da047306cad357143fe56f7afe3d323d89aa7383b6614b2b587f580044")
+
+    depends_on("fortran", type="build")
+    depends_on("meson@0.57.2:", type="build", when="build_system=meson")
+
+    depends_on("toml-f")
```
awvwgk: Needs version 0.4 at least

awvwgk (approved):

RMeli (approved):

Issue #316, opened 2025-06-18
pycuda: SystemError: type Boost.Python.enum has the Py_TPFLAGS_HAVE_GC flag but has no traverse function


Sinan81:
### Steps to reproduce

```
$ spack install pycuda
$ python
>>> import pycuda.driver
```

- pycuda@2024.1.1 and 2025.1.1
- numpy@1.26.4
- boost@1.88 and 1.80
- python@3.11

### Error message

```
$ python                                                                                       
Python 3.11.11 (main, Jun 17 2025, 00:46:57) [GCC 11.5.0] on linux                                                                                                                            
Type "help", "copyright", "credits" or "license" for more information.                                                                                                                        
>>> import pycuda.driver                                                                                                                                                                  
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/path/to/spack/installation/var/spack/environments/test_pycuda/.spack-env/view/lib/python3.11/site-packages/pycuda/driver.py", line 70, in <module>
    from pycuda._driver import *  # noqa
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
SystemError: type Boost.Python.enum has the Py_TPFLAGS_HAVE_GC flag but has no traverse function
```

### Information on your system

$ spack debug report
* **Spack:** 1.0.0.dev0 (99d06b95a3635055b90772de2cfb355bbf9f3bb6)
* **Python:** 3.9.21
* **Platform:** linux-rhel9-ivybridge

### General information

- [x] I have run `spack debug report` and reported the version of Spack/Python/Platform
- [x] I have searched the issues of this repo and believe this is not a duplicate
- [x] I have run the failing commands in debug mode and reported the output

Sinan81:
related issue in boost repo: https://github.com/boostorg/python/issues/400

Sinan81:
it looks like python@3.11 & boost.python compatibility issue is fixed as of boost@1.81.0 so I'm not sure why this is still an issue.

Sinan81:
Note that  `Py_TPFLAGS_HAVE_GC` on line 116 in https://github.com/boostorg/python/blob/boost-1.80.0/src/object/enum.cpp is removed in 
https://github.com/boostorg/python/blob/boost-1.81.0/src/object/enum.cpp

Sinan81:
it seems like boost installation is using hpp files rather than cpp. hence the above fix isn't working.

Sinan81:
[...]

PR #376, opened 2025-06-27
package/boost: fix bug for python@3.11:

Sinan81:
fix bug for python@3.11 or newer:
https://github.com/spack/spack-packages/issues/316

### DIFF
--- a/repos/spack_repo/builtin/packages/boost/package.py
+++ b/repos/spack_repo/builtin/packages/boost/package.py
@@ -318,7 +318,7 @@ class Boost(Package):
     conflicts("+clanglibcpp", when="@1.85: +stacktrace")
 
     # https://github.com/boostorg/python/issues/400
-    conflicts("@:1.80.0", when="+python ^python@3.11:", msg="Boost.python.enum has a known bug for older verions")
+    conflicts("@:1.80.0", when="+python ^python@3.11:", msg="Boost.python.enum has a known bug for older versions")
 
     # On Windows, the signals variant is required when building any of
     # the all_libs variants.
--- a/repos/spack_repo/builtin/packages/boost/package.py
+++ b/repos/spack_repo/builtin/packages/boost/package.py
@@ -318,7 +318,7 @@ class Boost(Package):
     conflicts("+clanglibcpp", when="@1.85: +stacktrace")
 
     # https://github.com/boostorg/python/issues/400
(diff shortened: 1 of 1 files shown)


Sinan81:
I'm inclining towards creating a patch for boost <= 1.80. :thinking:

repos/spack_repo/builtin/packages/boost/package.py
```diff
@@ -317,6 +317,9 @@ def libs(self):
     # safe to do so on affected platforms.
     conflicts("+clanglibcpp", when="@1.85: +stacktrace")
 
+    # https://github.com/boostorg/python/issues/400
+    conflicts("@:1.80.0", when="+python ^python@3.11:", msg="Boost.python.enum has a known bug for older versions")
```
hainest: A small clarification to the message.

```suggestion
    conflicts("@:1.80.0", when="+python ^python@3.11:", msg="Boost.python.enum has a known bug for python before 3.11")
```

hainest (changes requested):

hainest:
> I'm inclining towards creating a patch for boost <= 1.80. 🤔

What's the oldest combination of versions of Boost and Python you'd like to support?

Sinan81:
> > I'm inclining towards creating a patch for boost <= 1.80. 🤔
> 
> What's the oldest combination of versions of Boost and Python you'd like to support?

Down to 1.77 for sure. and if the patch is applicable for any version lower than that without modification, than we can continue to apply it. accordingly we can adjust the version range in conflict statement.

repos/spack_repo/builtin/packages/boost/package.py
Sinan81: Still not clear. how about:
"Boost.python.enum has a known bug for python 3.11"

hainest: It's any version of python newer than 3.11. What if we use the language of spack and just say

```suggestion
    conflicts("@:1.80.0", when="+python ^python@3.11:", msg="Boost.python.enum has a known bug for boost@:1.80.0 and python@3.11:")
```

hainest (approved):

Sinan81:
actually, spack's boost installation is working with header files as opposed to cpp. so it's not clear how it can be patched. It can be done in a different. PR later.

tldahlgren (approved):

PR #384, opened 2025-06-27
py-setproctitle: add v1.2.2 and v1.3.6

viperML:
This PR adds the latest version of py-setproctitle (1.3.6), and 1.2.2 which will be pinned by a py-ray update.

CC @theabm


RikkiButler20 (approved):
Verified sha256

viperML:
Thanks again!

PR #387, opened 2025-06-27
meep: make meep depend on Fortran

downloadico:
Meep needs Fortran to build.


wdconinc (approved):
Confirmed, https://github.com/NanoComp/meep/blob/master/configure.ac#L159.

PR #398, opened 2025-06-29
hpx: add v1.11.0

hkaiser:

### DIFF
--- a/repos/spack_repo/builtin/packages/hpx/package.py
+++ b/repos/spack_repo/builtin/packages/hpx/package.py
@@ -176,7 +176,7 @@ class Hpx(CMakePackage, CudaPackage, ROCmPackage):
 
     # Asio 1.34.0 removed io_context::work, used by HPX:
     # https://github.com/chriskohlhoff/asio/commit/a70f2df321ff40c1809773c2c09986745abf8d20.
-    conflicts("^asio@1.34:", when="@:1.10")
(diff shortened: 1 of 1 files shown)


rbberger:
@hkaiser looks like you need to update this asio conflict to 1.11. If there is no HPX version that works with this, just remove the `when` clause.
https://github.com/spack/spack-packages/blob/25b2c5b39ef29f02c26169b9d00734b73c5fd75a/repos/spack_repo/builtin/packages/hpx/package.py#L177-L179

hkaiser:
> @hkaiser looks like you need to update this asio conflict to 1.11. If there is no HPX version that works with this, just remove the `when` clause.
> 
> https://github.com/spack/spack-packages/blob/25b2c5b39ef29f02c26169b9d00734b73c5fd75a/repos/spack_repo/builtin/packages/hpx/package.py#L177-L179

The new release does work with all Asio versions, V1.10 does not. I believe what we have is correct.

rbberger:
@hkaiser well, the CI errors out with asio errors

hkaiser:
> @hkaiser well, the CI errors out with asio errors

Right. This is a new error that I have not seen before. I'll investigate. For now let's do what you suggested and remove the `when="@:1.10"` from the spec file. I created an issue for HPX (https://github.com/STEllAR-GROUP/hpx/issues/6730)

rbberger (approved):

PR #401, opened 2025-06-30
intel-oneapi-mkl: add compiler dependency when threads=openmp

msimberg:
Separating the compiler dependency from #16.


rscohn2 (approved):

msimberg:
This seems to not be any better than #16 and still fails in CI. See discussion in #16 for more context.

PR #402, opened 2025-06-30
update nlcglib,sirius,q-e-sirius

simonpintarelli:
- `q-e-sirius` versions `1.0.0` and `1.0.1`
- `sirius` version `7.8.0`
- `nlcglib` version `1.3.0`


repos/spack_repo/builtin/packages/nlcglib/package.py
```diff
@@ -44,7 +45,7 @@ class Nlcglib(CMakePackage, CudaPackage, ROCmPackage):
 
     depends_on("cxx", type="build")  # generated
 
-    depends_on("cmake@3.21:", type="build")
+    depends_on("cmake@3.23:", type="build")
```
RMeli, line 48: Does this only applies to the latest versions? If yes, you could use `when=...`.

RMeli (approved):

repos/spack_repo/builtin/packages/nlcglib/package.py
simonpintarelli: True, I just made it the same as sirius, that constraint will be present anyway.

RMeli (approved):

RMeli (approved):

PR #403, opened 2025-06-30
Revert style change

kwryankrattiger:
Revert the previous attempt to fix this and remove `set -e` which was causing the incorrect exit code when no changed files.


kwryankrattiger:
Redundant with https://github.com/spack/spack-packages/pull/410

PR #407, opened 2025-06-30
Geant4-VMC: add version 6-7-p1, fix geant4 version dependency

jezwilkinson:
Adds version 6-7-p1 to spec for Geant4-VMC: https://github.com/vmc-project/geant4_vmc/releases/tag/v6-7-p1
Also adds an explicit version requirement for the Geant4 dependency for this version of geant4-vmc (6-7 fails to compile if concretised with earlier versions of Geant4)

### DIFF
--- a/repos/spack_repo/builtin/packages/geant4_vmc/package.py
+++ b/repos/spack_repo/builtin/packages/geant4_vmc/package.py
@@ -42,6 +42,7 @@ class Geant4Vmc(CMakePackage):
     depends_on("cxx", type="build")  # generated
 
     depends_on("cmake@3.3:", type="build")
+    depends_on("cmake@3.16:", type="build", when="@6-7:")
     depends_on("geant4")
     depends_on("geant4@11.2.0:", when="@6-7:")
     depends_on("vmc")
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/geant4_vmc/package.py
```diff
@@ -36,10 +38,12 @@ class Geant4Vmc(CMakePackage):
     version("4-0-p2", sha256="cdd73c499cd296f13b6c0d37e161e7d94343f85617b2a7577ded8312248f9b9b")
     version("3-6-p6", sha256="e62a62ff7075ff9afb2ffe420610374f62136094a447bbbc5f739a2238ddb0f0")
 
+    depends_on("c", type="build")
     depends_on("cxx", type="build")  # generated
 
     depends_on("cmake@3.3:", type="build")
```
wdconinc, line 44: Noticed that a newer version of CMake is required but did not verify that v6-7 is when it changed:
```suggestion
    depends_on("cmake@3.3:", type="build")
    depends_on("cmake@3.16:", type="build", when="@6-7:")
```

wdconinc (commented):
Looks good to me. One change to CMake minimum version.

wdconinc (approved):

PR #408, opened 2025-06-30
Fix style_check.sh

alalazo:


haampie:
Let's go with #410. This is still missing flake8/isort.

PR #409, opened 2025-06-30
Geant3: add version 4-4, fix deprecated ROOT variant dependency

jezwilkinson:
Add version 4-4 to spec for Geant3: https://github.com/vmc-project/geant3/releases/tag/v4-4
Also fix reference to deprecated ROOT variant "vmc" (this variant is not defined as of ROOT 6.26, so currently specifying "root~vmc" will not allow concretisation with newer ROOT versions)

### DIFF
--- a/repos/spack_repo/builtin/packages/geant3/package.py
+++ b/repos/spack_repo/builtin/packages/geant3/package.py
@@ -41,7 +41,7 @@ class Geant3(CMakePackage):
     def cmake_args(self):
         args = []
         if self.spec.satisfies("%gcc@10:"):
-            args.append('-DCMAKE_Fortran_FLAGS="-std=legacy"')
+            args.append('-DCMAKE_Fortran_FLAGS=-fallow-argument-mismatch -fallow-invalid-boz')
         args.append(self.define_from_variant("CMAKE_CXX_STANDARD", "cxxstd"))
         return args
 
--- a/repos/spack_repo/builtin/packages/geant3/package.py
+++ b/repos/spack_repo/builtin/packages/geant3/package.py
@@ -13,7 +13,8 @@ class Geant3(CMakePackage):
 
     homepage = "https://github.com/vmc-project/geant3"
     url = "https://github.com/vmc-project/geant3/archive/v3-7.tar.gz"
-
+    
+    version("4-4", sha256="31ad296d856aa282a347abcb451809e110237d8a3fd7c47458777c2c29ae47f2")
     version("4-1", sha256="a1dcab7bc7a7493e4c78d7bec22cd816e79e40992bf9db0d616e2a0125fcdf50")
     version("3-8", sha256="6ff6745eef59139d791bef043b405f6d515be1d98096cf4e82ac4c1f61f737dc")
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/geant3/package.py
```diff
@@ -39,7 +41,7 @@ class Geant3(CMakePackage):
     def cmake_args(self):
         args = []
         if self.spec.satisfies("%gcc@10:"):
-            args.append('-DCMAKE_Fortran_FLAGS="-fallow-argument-mismatch -fallow-invalid-boz"')
+            args.append('-DCMAKE_Fortran_FLAGS="-std=legacy"')
```
wdconinc: Is this really equivalent? I see in https://manpages.ubuntu.com/manpages/jammy/en/man1/alpha-linux-gnu-gfortran.1.html that legacy implies `allow-argument-mismatch`, but that's not mentioned for `allow-invalid-boz`.

jezwilkinson: OK, i think i see what happened here now... The actual issue I was seeing with the flags wasn't actually because of the flags themselves, but the quotes around them (the error I see on Ubuntu was ` gfortran: error: unrecognized command-line option '-fallow-argument-mismatch -fallow-invalid-boz'`, and it seems like a) all three of these flags are being input as standard in the command process anyway, and b) when multiple of them are set it's also adding the quote marks into the commandline options and *that* seems to be the root cause of the error:

```
-I/tmp/jeremy/spack-stage/spack-stage-geant3-4-4-3h2vhel2orgcyfyyjwpgfznjbmqk4zfv/spack-src/TGeant3 -I/tmp/jeremy/spack-stage/spack-stage-geant3-4-4-3h2vhel2orgcyfyyjwpgfznjbmqk4zfv/spack-build-3h2vhel
-I/data/Work/dev/FairSoft_spack/spack/opt/spack/linux-skylake/root-6.36.00-e3nou277fyfxe2nfmvuovxylemcxqvjk/include/root
"-fallow-argument-mismatch -fallow-invalid-boz"  -std=legacy -finit-local-zero -fno-strict-overflow -fallow-argument-mismatch -fallow-invalid-boz -O3
```

If I just remove the quote marks in this line rather than changing the flags, it also seems to work. So I'll revert this and remove the quotes instead.

repos/spack_repo/builtin/packages/geant3/package.py
```diff
@@ -39,7 +41,7 @@ class Geant3(CMakePackage):
     def cmake_args(self):
         args = []
         if self.spec.satisfies("%gcc@10:"):
-            args.append('-DCMAKE_Fortran_FLAGS="-fallow-argument-mismatch -fallow-invalid-boz"')
+            args.append("-DCMAKE_Fortran_FLAGS=-fallow-argument-mismatch -fallow-invalid-boz")
```
wdconinc: Even better (and avoiding quote issues) is to use the `self.define` helper:
```suggestion
            args.append(self.define("CMAKE_Fortran_FLAGS", "-fallow-argument-mismatch -fallow-invalid-boz"))
```

jezwilkinson: Ah even better, thanks!

wdconinc (approved):

PR #420, opened 2025-06-30
openturbine: add cuDSS variant and update ADI and ROSCO variants

ddement:
Update OpenTurbine package to add cuDSS variant and to update ADI/ROSCO variants


tldahlgren (approved):
Confirmed variants are used.

PR #422, opened 2025-06-30
sundials: add v7.4.0

balos1:


gardner48 (approved):
Confirmed the commit hash is correct for v7.4.0

RikkiButler20 (approved):
Verified checksum

PR #425, opened 2025-07-01
samurai: add v0.24

sbstndb:
Add new version of samurai


alalazo:
@spackbot run pipeline

PR #427, opened 2025-07-01
hepmc3: Add version 3.3.1

paulgessinger:


alalazo (dismissed):

paulgessinger:
Whizard still seems to not build with this change. I don't understand the error message though, as there are additional fatal messages complaining about the fact that ROOT was built with C++17 but then whizards hepmc3 related code is not.

Locally, I can build `whizard ^hepmc3@3.3.1` without issues.

paulgessinger:
Ah, I can get the warning when building with gcc11. In my case, the ROOT `#error` cases are not present, I only get 

```
HepMC3Wrap.cpp: In function 'double gen_event_weight(HepMC3::GenEvent*, int)':
HepMC3Wrap.cpp:158:19: warning: comparison of integer expressions of different signedness: 'int' and 'std::vector<double>::size_type' {aka 'long unsigned int'} [-Wsign-compare]
  158 |   if (0 <= i && i <= evt->weights().size()) {
      |                 ~~^~~~~~~~~~~~~~~~~~~~~~~~
``` 

which is not fatal.

The ROOT `#error` seems to be related  to the whizard package not figuring out which standard version to use for the C++ code, the log indicates it runs on C++11.

paulgessinger:
whizard's `package.py` tries to handle this via

https://github.com/spack/spack-packages/blob/7aff9dd7017886ac70d2f565516bc78124b3d30d/repos/spack_repo/builtin/packages/whizard/package.py#L122-L127

but that's not effective, the compiler flags determined by whizard's `configure` script still only get `-std=c++11`, even though the detected value here is 20. The flag injection does not seem to be working here.

I'm not entirely sure, but I think the whizard build is bypassing spack's compiler wrapper, as I see direct calls to `/usr/bin/g++` in the log.

paulgessinger:
I'm fairly certain now this is a whizard issue, not a hepmc3 issue. 

What do I do in this case?

alalazo:
> I'm fairly certain now this is a whizard issue, not a hepmc3 issue. What do I do in this case?

Not sure what's the best option, but here are two alternatives:
- Fix `whizard` recipe in this PR, if that is also of interest in your workflow
- Pin `hepmc3` version in the `hep` pipeline to the previous version, and leave traces why that was done

@wdconinc Do you have any opinion on this matter?

bernhardkaindl:
@paulgessinger, [whizard](https://github.com/spack/spack-packages/blob/develop/repos/spack_repo/builtin/packages/whizard/package.py) has #326 (for adding LD_LIBRARY_PATH for run environment) as it is set up to use the actual compilers directly, claimed on "seems incompatible with filter_compiler_wrappers":

https://github.com/spack/spack-packages/blob/948d4ea14409e38d47882b5a5c2d61d99d02b30b/repos/spack_repo/builtin/packages/whizard/package.py#L141-L148

I think this should be fixed to either use `filter_compiler_wrappers` or, if needed, patch the code of whizard to use the correct compiler calls at runtime.

This was added in https://github.com/spack/spack/pull/17300, as it seems (AFAICS using `grep` and `strings`) that the used compilers are apparently in the strings included in libwhizard and other libs and many other files.

Adam commented about it:
> In the long run, it might be a good idea to contact the developers and ask them if there's a way to change the compiler used at run-time. Often there's an env var or file that can be patched.

Maybe, it is enough to just use `filter_compiler_wrappers` on the installed scripts:

```py
grep -r -e 'g[+c][+c]-13"' -e gfortran-13 *
bin/whizard-config:         echo "/usr/bin/gfortran-13"
grep: lib/libwhizard.so.2.0.2: binary file matches
lib/whizard/libtool:LTCC="/usr/bin/gcc-13"
lib/whizard/libtool:CC="/usr/bin/gcc-13"
lib/whizard/libtool:CC="/usr/bin/gfortran-13"
lib/whizard/libtool:CC="/usr/bin/gfortran-13"
lib/whizard/libtool:CC="/usr/bin/g++-13"
```
`grep -r '-D[^ ]-13'` could not find any `-D` flags that pass the compilers, and `grep` could find them in `.h*`, or '*.c*` files, and I could not find any place where the strings would be generated. So I assume that the strings would have been compiled into the binaries using `libtool` to compile Whizard. So I could not find any evidence of a deliberate inclusion of the compilers besides `bin/whizard-config` and `lib/whizard/libtool` that are easy to fix by `filter_compiler_wrappers`.

@vvolkl, you did #17300, could you shed some more light on this (why you did/could not use `filter_compiler_wrappers`)?

paulgessinger:
@alalazo I pinned the hepmc3 version to 3.3.0 in the stack description now, let's see if that fixes the job. @wdconinc

alalazo (approved):

alalazo:
LGTM, now it's up to @wdconinc

wdconinc (approved):
Thanks. Now on to fixing whizard...

PR #429, opened 2025-07-01
py-grpcio: add internal-re2 variant

viperML:
As explained on the comment, grpcio crashes at runtime with the error described at https://github.com/spack/spack/issues/49338.

The re2 dependency is optional so it can be safely disabled by default.

Found this when updating py-ray.
CC @theabm

### DIFF
--- a/repos/spack_repo/builtin/packages/py_grpcio/package.py
+++ b/repos/spack_repo/builtin/packages/py_grpcio/package.py
@@ -58,6 +58,10 @@ class PyGrpcio(PythonPackage):
         # Released 6ish years ago and does not install for python 3.8 with gcc11
     )
 
+    # Spack's re2 causes crashes, even pinned with the same version
+    # https://github.com/spack/spack/issues/49338
+    variant("internal-re2", default=True, description="Use internal re2")
+
     depends_on("c", type="build")  # generated
     depends_on("cxx", type="build")  # generated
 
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/py_grpcio/package.py
```diff
@@ -106,24 +110,25 @@ def setup_build_environment(self, env: EnvironmentModifications) -> None:
...
-            r"(\s+CARES_INCLUDE = ).*",
-            r"\1('{0}',)".format(self.spec["c-ares"].prefix.include),
-            "setup.py",
-        )
-        filter_file(
-            r"(\s+RE2_INCLUDE = ).*",
-            r"\1('{0}',)".format(self.spec["re2"].prefix.include),
-            "setup.py",
-        )
+        with when("+re2"):
```
AbhinavPraveen: Is this line in the wrong place? Shouldn't it be above only the `filter_file` for `RE2_INCLUDE`?

alalazo: The `when` context manager cannot be used in a method. It's meant for directives in the class scope  @viperML Did you check the package behave as intended?

alalazo (changes requested):

repos/spack_repo/builtin/packages/py_grpcio/package.py
viperML: Oops, should had taken another coffee...

viperML:
@AbhinavPraveen @alalazo

Summary of the last force-push:

- Fixed the position and removed the bad `with`.
- Renamed the variant to `internal-re2`. This is probably in line with the naming conventions for other packages, i.e. `openmpi +internal-hwloc`.
- I built and ran the package with both specs. `~internal-re2` crashes with the mentioned bug, `+internal-re2` works OK, and is the new default.
- I tried pinning re2 to the same git submodule that upstream uses (`re2@2022-04-01`, https://github.com/grpc/grpc/pull/31665), but still crashes with the same error.

PR #430, opened 2025-07-01
py-opentelemetry-{api, sdk,proto,semantic-...}: init with 1.34.1

viperML:
Simple Python packages with not many dependencies.

Needed for a py-ray update, CC @theabm


alalazo:
@spackbot fix style

viperML:
Manually ran `spack style`, hopefully works now.

PR #363, opened 2025-06-26
py-cython: add v3.0.12

viperML:
I need this as a dependency for a future update of py-ray. We already have cython 3.1.X, but I can imagine this may be easier to merge.

Changelog from upstream: https://github.com/cython/cython/releases/tag/3.0.12

```
3.0.12 (2025-02-11)

===================

Bugs fixed
----------

* Release 3.0.11 introduced some incorrect ``noexcept`` warnings.
  (Github issue :issue:`6335`)

* Conditional assignments to variables using the walrus operator could crash.
  (Github issue :issue:`6094`)

* Dict assignments to struct members with reserved C names could generate invalid C code.

* Fused ctuples with the same entry types but different sizes could fail to compile.
  (Github issue :issue:`6328`)

* In Py3, `pyximport` was not searching `sys.path` when looking for importable source files.
  (Github issue :issue:`5615`)

* Using `& 0` on integers produced with `int.from_bytes()` could read invalid memory on Python 3.10.
  (Github issue :issue:`6480`)

* Modules could fail to compile in PyPy 3.11 due to missing CPython specific header files.
  Patch by Matti Picus.  (Github issue :issue:`6482`)

* Minor fix in C++ ``partial_sum()`` declaration.
```

cc @theabm

PR #432, opened 2025-07-01
py-ray: add version 2.47.1, remove old versions and refactor definition

viperML:
This PR updates the package for Ray, also removing the cruft and old versions.

Depends on the following PR's to be merged before:

- #430
- #429 
- #428 
- #384
- #363

I'll try to keep this other branch with all the backports, to be able to test the package in the meantime: https://github.com/viperML/spack-packages/tree/py-ray-update

The current state of the PR works, but marking as draft as I may add the optional dependencies corresponding to the groups [data,train,tune,serve] which are recommended by Ray usptream.

CC @theabm

### DIFF
--- a/repos/spack_repo/builtin/packages/py_ray/package.py
+++ b/repos/spack_repo/builtin/packages/py_ray/package.py
@@ -16,93 +16,106 @@ class PyRay(PythonPackage):
 
     license("Apache-2.0")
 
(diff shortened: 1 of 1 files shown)


viperML:
I've finally added all the variants that Ray exposes, except for the rllib. From my testing, the package seems to work fine now.

theabm (approved):

PR #435, opened 2025-07-01
Tidyup NCCL support for aws-ofi-nccl plugin

green-br:
Following work in spack/spack#47527 these are the changes to tidyup the support for aws-ofi-nccl plugin which is used on machines such as HPE with Slingshot 11.  Tested in environment with:

```
  specs:
  - aws-ofi-nccl ^nccl+cuda cuda_arch=90
  - nccl-tests+mpi cuda_arch=90
```

### DIFF
--- a/repos/spack_repo/builtin/packages/aws_ofi_nccl/package.py
+++ b/repos/spack_repo/builtin/packages/aws_ofi_nccl/package.py
@@ -46,7 +46,7 @@ class AwsOfiNccl(AutotoolsPackage):
 
     depends_on("libfabric")
     depends_on("cuda")
-    depends_on("nccl")
+    depends_on("nccl fabrics=auto")
     depends_on("mpi")
     depends_on("hwloc", when="@1.7:")
     depends_on("autoconf", type="build")
@@ -62,6 +62,8 @@ class AwsOfiNccl(AutotoolsPackage):
     # To enable this plug-in to work with NCCL add it to the LD_LIBRARY_PATH
     def setup_run_environment(self, env: EnvironmentModifications) -> None:
         env.append_path("LD_LIBRARY_PATH", self.prefix.lib)
+        # Set the this network so that NCCL doesnt pick up anything else.
+        env.set("NCCL_NET", "AWS Libfabric")
 
     # To enable this plug-in to work with NCCL add it to the LD_LIBRARY_PATH
     def setup_dependent_run_environment(
--- a/repos/spack_repo/builtin/packages/aws_ofi_nccl/package.py
+++ b/repos/spack_repo/builtin/packages/aws_ofi_nccl/package.py
@@ -62,7 +62,7 @@ class AwsOfiNccl(AutotoolsPackage):
     # To enable this plug-in to work with NCCL add it to the LD_LIBRARY_PATH
     def setup_run_environment(self, env: EnvironmentModifications) -> None:
         env.append_path("LD_LIBRARY_PATH", self.prefix.lib)
-        # Set the this network so that NCCL doesnt pick up anything else.
+        # Set the network so that NCCL doesn't pick up anything else.
         env.set("NCCL_NET", "AWS Libfabric")
 
     # To enable this plug-in to work with NCCL add it to the LD_LIBRARY_PATH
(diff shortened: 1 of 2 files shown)


green-br:
@spackbot fix style

repos/spack_repo/builtin/packages/nccl/package.py
```diff
@@ -66,7 +66,15 @@ class Nccl(MakefilePackage, CudaPackage):
...
     depends_on("cxx", type="build")  # generated
 
-    depends_on("rdma-core")
+    # Make verbs default but package like aws-ofi-nccl can be used instead which loads
+    # another library to use instead at runtime.
+    values=disjoint_sets(("auto",), ("verbs",))  # supported transports
+        .with_default("verbs")
+        .with_non_feature_values("auto"),
+        description="List of fabrics that are enabled; " "'auto' lets nccl determine at runtime",
+    )
```
alalazo: :thinking: It seems like something was removed from the intended changeset...

green-br: Sorry for confusion - failed attempt at fixing the style issue picked up in checks.  Just fixed, ran `spack style` and retested with same Spack environment.  Hopefully makes more sense now.

repos/spack_repo/builtin/packages/nccl/package.py
```diff
@@ -66,7 +66,17 @@ class Nccl(MakefilePackage, CudaPackage):
     depends_on("c", type="build")  # generated
     depends_on("cxx", type="build")  # generated
 
-    depends_on("rdma-core")
+    # Make verbs default but package like aws-ofi-nccl can be used instead which loads
```
adamjstewart:
```suggestion
    # Make verbs default but packages like aws-ofi-nccl can be used instead which loads
```

repos/spack_repo/builtin/packages/aws_ofi_nccl/package.py
```diff
@@ -62,6 +62,8 @@ def url_for_version(self, version):
     # To enable this plug-in to work with NCCL add it to the LD_LIBRARY_PATH
     def setup_run_environment(self, env: EnvironmentModifications) -> None:
         env.append_path("LD_LIBRARY_PATH", self.prefix.lib)
+        # Set the this network so that NCCL doesnt pick up anything else.
```
adamjstewart:
```suggestion
        # Set the network so that NCCL doesn't pick up anything else.
```

repos/spack_repo/builtin/packages/nccl/package.py
```diff
@@ -66,7 +66,17 @@ class Nccl(MakefilePackage, CudaPackage):
...
+    # another library to use instead at runtime.
+    variant(
+        "fabrics",
+        values=disjoint_sets(("auto",), ("verbs",))  # supported transports
+        .with_default("verbs")
+        .with_non_feature_values("auto"),
+        description="List of fabrics that are enabled; " "'auto' lets nccl determine at runtime",
```
adamjstewart:
```suggestion
        description="List of fabrics that are enabled; 'auto' lets nccl determine at runtime",
```

adamjstewart (commented):
I don't love "auto" as it challenges reproducibility, but it seems to be required for aws-ofi-nccl. As long as the default remains the same, it should be fine.

repos/spack_repo/builtin/packages/nccl/package.py
msimberg: To follow up on @alalazo's comment, what does the `fabrics` variant actually change? I can't see any configure/build/etc. options set based on the variant.

green-br: It changes the dependency on rdma-core package.  At the moment I think nccl only depends on rdma-core at runtime since `RDMA_CORE` Makefile variable is not set anyway.  I don't have access to a system where I can check so have left the current logic but allowed for other networks to be used without pulling in the whole rdma-core dependency. On Slingshot 11 from HPE we can build NCCL without rdma-core and then set NCCL to use the library from aws-ofi-nccl.

msimberg: Ah, sorry, I missed that. That makes sense.

green-br:
> I don't love "auto" as it challenges reproducibility, but it seems to be required for aws-ofi-nccl. As long as the default remains the same, it should be fine.

Thanks for taking a look.  Just committed suggestions.

adamjstewart (approved):

Issue #215, opened 2025-06-12
Installation issue: ROOT 6.36.00 on Darwin


jezwilkinson:
### Steps to reproduce the issue

On Mac OSX systems, the build for ROOT v6.36.00 through Spack completes without errors, but the RPATH for its libraries does not get installed correctly. This causes ROOT to crash at runtime.

```console
$ spack add root@6.36.00
$ spack install
$ root
...
```

### Error message

<details><summary>Error message</summary>

<pre>
alfaci@citmac008 spack % root
dyld[79397]: Library not loaded: @rpath/libRint.6.36.so
  Referenced from: <27660B66-7EE2-3FB4-9855-E43EF17D9259> /Users/alfaci/jwilkins/spack/opt/spack/darwin-m1/root-6.36.00-s4iedayuyeec6kyav2zahhnvnepvy2gi/bin/root.exe
  Reason: tried: '/private/var/folders/8c/3pr52n2x4ng0j9p3rmqtfyvc0000gn/T/alfaci/spack-stage/spack-stage-root-6.36.00-s4iedayuyeec6kyav2zahhnvnepvy2gi/spack-build-s4ieday/lib/libRint.6.36.so' (no such file) [...]
</pre></details>

### Information on your system

* **Spack:** 1.0.0.dev0 (a40b98368c75dc1af94b48799f14e7bee57fca68)
* **Builtin repo:** 1266998ab140209b68ad7a755e5a4ff0f5e8b8bc
* **Python:** 3.13.3
* **Platform:** darwin-sequoia-m1 (also on darwin-sequoia-skylake and darwin-sonoma-skylake)

### Additional information

I can fix the issue on my side if I edit the ROOT recipe for 6.36.00 to apply [a patch](https://github.com/jezwilkinson/spack/blob/fairsoft_cleanup/var/spack/repos/spack_repo/builtin/packages/root/RootBuildOptions.patch) to `RootBuildOptions.cmake` that effectively reverts https://github.com/root-project/root/pull/18277/commits/801684cbcb437a27dca8dbfe3f4fc77a5c14aa32 - I'm not sure if there's just some change that should be added in the build options in the Spack recipe to account for the new handling of the `CMAKE_INSTALL_RPATH` from that commit, rather than reverting it.
ROOT release versions prior to 6.36 (which do not include the commit mentioned above) build and run without problems.

Maintainers for root: @drbenmorgan @gartung @greenc-FNAL @marcmengel @vitodb @vvolkl 

[...]

PR #436, opened 2025-07-01
root: patch handling of CMAKE_INSTALL_RPATH

jezwilkinson:
Potential fix to address https://github.com/spack/spack-packages/issues/215 (issues installing ROOT >=6.36 library RPaths on Mac OSX).

Currently, the situation is the following:
- ROOT installs its libraries into `$prefix/lib/root` due to the "gnuinstall" install option
- Spack's CMake package builder only allows for `CMAKE_INSTALL_RPATH` values of `$prefix/lib` and `$prefix/lib64` (hardcoded)
- Prior to ROOT 6.36, ROOT overrode the `CMAKE_INSTALL_RPATH` with `@loader_path` and `@loader_path/${BINDIR_TO_LIBDIR}` in its CMake configuration. Since 6.36, these are instead _appended_ to the INSTALL_RPATH.
- At install time, `install_name_tool` sees `$prefix/lib` in both `CMAKE_INSTALL_RPATH` and in the paths already included from `CMAKE_INSTALL_RPATH_USE_LINK_PATH`, and refuses to add the `@loader_path` paths to the binary as a result. This means that the libraries don't get linked properly and ROOT fails at runtime.

This PR would override the `CMAKE_INSTALL_RPATH` with `$prefix/lib/root` by adding it to the CMake options on the RPath variant. The existing `$prefix/lib` and `/lib64` are still part of the link path directories, but since there's no longer the clash with the install RPaths, all of the correct paths get added to the installation.

Admittedly this is probably not the cleanest way of dealing with this issue, but I couldn't find a better way of overriding or adding the correct library dir to the `CMAKE_INSTALL_PATH` with the current version of the CMakePackage build system in Spack.

### DIFF
--- a/repos/spack_repo/builtin/packages/root/package.py
+++ b/repos/spack_repo/builtin/packages/root/package.py
@@ -976,6 +976,10 @@ class Root(CMakePackage):
             options.append(define("FTGL_ROOT_DIR", ftgl_prefix))
             options.append(define("FTGL_INCLUDE_DIR", ftgl_prefix.include))
 
+        # Fix RPath handling with gnuinstall
+        if "+rpath" in self.spec:
+            options.append(define("CMAKE_INSTALL_RPATH", self.prefix.lib.root))
+
         return options
 
     def setup_build_environment(self, env: EnvironmentModifications) -> None:


marcmengel (approved):
Seems like a reasonable workaround for this problem.

tldahlgren (approved):

PR #438, opened 2025-07-01
lammps: add missing C dependency

rbberger:

### DIFF
--- a/repos/spack_repo/builtin/packages/lammps/package.py
+++ b/repos/spack_repo/builtin/packages/lammps/package.py
@@ -415,23 +415,8 @@ class Lammps(CMakePackage, CudaPackage, ROCmPackage, PythonExtension):
         deprecated=True,
     )
 
-    depends_on("c", type="build", when="@20250612:")
+    depends_on("c", type="build")
     depends_on("cxx", type="build")
-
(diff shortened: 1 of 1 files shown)


kwryankrattiger (dismissed):

alalazo:
@spackbot run pipeline

repos/spack_repo/builtin/packages/lammps/package.py
```diff
@@ -415,6 +415,7 @@ class Lammps(CMakePackage, CudaPackage, ROCmPackage, PythonExtension):
         deprecated=True,
     )
 
+    depends_on("c", type="build", when="@20250612:")
```
alalazo: Did we consider simplifying this whole thing and:
```python
depends_on("c", type="build")
```
plus delete the conditional dependency below on the C language? This will simplify the maintenance of the package, without imposing more constraints, practically speaking (usually, if there's a C++ compiler available, there's also the C compiler of the same toolchain)

rbberger: Given that new situation, you're right. Let's just simplify it.

alalazo (approved):

PR #439, opened 2025-07-01
armpl-gcc: add v25.04

paolotricerri:
This patch introduces the possibility of installing armpl_gcc and acfl 25.04 through spack. It also introduces the support of installing armpl-gcc package on Ubuntu 24.04


repos/spack_repo/builtin/packages/armpl_gcc/package.py
```diff
@@ -466,7 +474,15 @@ def blas_libs(self):
...
-        armpl_libs += find_system_libraries(["libm"])
+        # Link the same libraries as the gcc used for Arm PL
+        gcc_compiler = self.compiler.cc
+        gcc_path = os.path.dirname(os.path.dirname(gcc_compiler))
+        armpl_libs += find_libraries(
+            ["libstdc++", "libgomp", "libm"],
+            root=gcc_path,
+            shared=self.spec.satisfies("+shared"),
+            recursive=True,
+        )
```
alalazo:
```suggestion
        # Link the same libraries as the gcc used for Arm PL
        armpl_libs += find_libraries(
            ["libstdc++", "libgomp", "libm"],
            root=self["gcc"].prefix,
            shared=self.spec.satisfies("+shared"),
            recursive=True,
        )
```

alalazo (commented):
A small suggestion to shorten the code

alalazo (approved):

PR #444, opened 2025-07-01
Add new versions to survey package.py

jgalarowicz:
Add new versions to survey package.py


bernhardkaindl (approved):
> Survey is a licensed product with the source not openly available.

Therefore we can't check anything, might as well merge as-is.

PR #446, opened 2025-07-02
libssh2: deprecate older versions due to CVEs

drkrynstrng:
Deprecating older versions of libssh2 due to CVEs.


drkrynstrng:
@spackbot fix style

alalazo:
The failure is because this PR deprecates a version of `libssh2` that is needed to build `julia@1.9`, which we use in the tutorial.

@giordano Can we relax the dependency ranges in `julia`? Also, would it be fine if `julia @1.9` becomes unsat, unless `--deprecated` is used?

giordano:
I need to double check libssh2 compatibility, but we may be able to relax the compatibility bounds in the Julia recipes, I'll try to check that later today.

drkrynstrng:
Ah, I see. Maybe there are patches that could be applied to the older versions, instead of deprecating them.

bernhardkaindl (approved):

PR #447, opened 2025-07-02
[cp2k] Add tblite dependency

mtaillefumier:
- Add the Mi300 target to cp2k and dbcsr
- Add tblite to cp2k recipe
- tblite can be build with either cmake or meson

### DIFF
--- a/repos/spack_repo/builtin/packages/tblite/package.py
+++ b/repos/spack_repo/builtin/packages/tblite/package.py
@@ -2,12 +2,14 @@
 #
 # SPDX-License-Identifier: (Apache-2.0 OR MIT)
 
+from spack_repo.builtin.build_systems import cmake, meson
+from spack_repo.builtin.build_systems.cmake import CMakePackage
 from spack_repo.builtin.build_systems.meson import MesonPackage
 
 from spack.package import *
 
 
-class Tblite(MesonPackage):
+class Tblite(CMakePackage, MesonPackage):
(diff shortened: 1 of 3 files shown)


repos/spack_repo/builtin/packages/tblite/package.py
```diff
@@ -20,26 +22,37 @@ class Tblite(MesonPackage):
...
+
+    for build_system in ["cmake", "meson"]:
+        depends_on(
+            f"mctc-lib@0.3: build_system={build_system}", when=f"build_system={build_system}"
+        )
+        depends_on(
+            f"simple-dftd3@0.3: build_system={build_system}", when=f"build_system={build_system}"
+        )
+        depends_on(f"dftd4@3: build_system={build_system}", when=f"build_system={build_system}")
+        depends_on(f"toml-f build_system={build_system}", when=f"build_system={build_system}")
```
alalazo, line 44: Some of these build systems are not supported, see issues in CI.

alalazo (changes requested):

repos/spack_repo/builtin/packages/tblite/package.py
mtaillefumier: Thanks for pointing this out. This PR depends on https://github.com/spack/spack-packages/pull/371 which is still under review. I realized this afterwards.

mtaillefumier:
I combined this PR with PR #371

PR #345, opened 2025-06-24
CI: Add debug to darwin for signing

kwryankrattiger:
Debug info for mac only.

PR #396, opened 2025-06-28
remove ci_kick_package

haampie:

PR #449, opened 2025-07-02
Remove ci-kick-package

alalazo:
This package was added in #345, for debugging purposes. The PR was partially reverted in #415, but the package was not removed.

Hence, remove the package here, since it should not be part of builtin.


eugeneswalker (approved):

kwryankrattiger:
I had left this on purpose as it is a convenient way to force things in CI to run that doesn't ping other package maintainers but it does ping CI maintainers. I would like to keep this in the repo if possible even if it isn't meant to be installed by anyone in a real environment rather than have to re-create it every time we want to use it for CI purposes.

This package was previously PR'd to be removed and I closed that PR with a similar comment. #396

alalazo:
Discussed in DM to have it in a separate repo, since it's just for debugging pipelines.

PR #455, opened 2025-07-02
julia: relax compatibility with libssh2

giordano:
Ref: https://github.com/spack/spack-packages/pull/446#issuecomment-3028008667.  As far as I could see, we never had compatibility problems with libssh2, so it should hopefully be safe to relax the compatibility specification.

I hope I got the syntax right: with, say, `libssh2@1.10.0:1.10` I intend to allow from v1.10.0 to any following release in the v1 series, excluding a hypothetical future version v2 or later, is that correct?

CC: @alalazo @haampie.


giordano:
Build failure in vtk feels unrelated (but I can't see the full log?).

alalazo (approved):

alalazo:
It seems it's taking more than 6hrs to build, and yeah - it feels unrelated to this PR :disappointed: I just restarted that pipeline.

PR #464, opened 2025-07-03
hypre: post install may not depend on CXX

cessenat:
When package does not depend on cxx, the run after install cache_test_sources function filters CXX with a non existing variable. This PR intends to make sure os.environ['CXX'] exists.


repos/spack_repo/builtin/packages/hypre/package.py
```diff
@@ -402,7 +402,8 @@ def cache_test_sources(self):
         makefile = join_path(install_test_root(self), self.extra_install_tests, "Makefile")
         filter_file(r"^HYPRE_DIR\s* =.*", f"HYPRE_DIR = {self.prefix}", makefile)
         filter_file(r"^CC\s*=.*", f"CC = {os.environ['CC']}", makefile)
-        filter_file(r"^CXX\s*=.*", f"CXX = {os.environ['CXX']}", makefile)
+        if "CXX" in os.environ:
```
alalazo: Maybe:
```suggestion
        if "cxx" in self.spec:
```
?

alalazo (approved):

alalazo:
Unrelated failure in pipelines :melting_face:

PR #473, opened 2025-07-03
Cleanup style check some more

kwryankrattiger:
- remove unnecessary check for spack-core
- run all checks even if one fails, report aggregated status
- print which check(s) are running


.ci/style_check.sh
```diff
@@ -1,12 +1,16 @@
 #!/bin/sh -e
 ref="${1:-develop}"
-if [ ! -d "spack-core" ]; then
-  printf "No 'spack-core' dir found, should be a symlink or clone of 'spack/spack'\n" >&2
```
haampie, line 4: Required for flake8 plugins from spack/spack: https://github.com/spack/spack-packages/blob/257f242f3baa7f71f0ca02c3f7c81807b0534d25/.flake8#L24

kwryankrattiger: hm, could we moves those to packages? It seems it is primarily/only handling the `@when` creating multiple definitions which is a packages issue more than a core issue.

kwryankrattiger:
#546

PR #474, opened 2025-07-03
Qt: fixup win masm patch

johnwparent:
Adds escaped quoting to the masm.pri python command line definition involving file redirection so the generated NMake is able to pass the command line to Python correctly.
Escaped quotes are required so that both the PRI and NMake files interpret each command component correctly
Without quotes NMake python command arguments  are processed as a single arguent to Python and evaluated incorrectly


repos/spack_repo/builtin/packages/icu4c/Quote_datagen.patch
```diff
@@ -0,0 +1,184 @@
+diff --git a/source/data/makedata.mak b/source/data/makedata.mak
+index 1fe8c88593c..36dcd4cd7b8 100644
+--- a/source/data/makedata.mak
++++ b/source/data/makedata.mak
+@@ -285,7 +285,7 @@ ICU_DATA_BUILD_VERBOSE=
+ # Keep track of whether they are built via timestamp files.
+ 
+ $(TOOLS_TS): "$(ICUTOOLS)\genrb\$(CFGTOOLS)\genrb.exe" "$(ICUTOOLS)\gencnval\$(CFGTOOLS)\gencnval.exe" "$(ICUTOOLS)\gencfu\$(CFGTOOLS)\gencfu.exe" "$(ICUTOOLS)\icupkg\$(CFGTOOLS)\icupkg.exe" "$(ICUTOOLS)\makeconv\$(CFGTOOLS)\makeconv.exe" "$(ICUPBIN)\pkgdata.exe"
+-	@echo "timestamp" > $(TOOLS_TS)
++	@echo "timestamp" > "$(TOOLS_TS)"
```
scheibelp, line 10: (question) Based on the description this sounds like something that would also break outside of spack - is this something that should be upstreamed?

Also, do you think it would be worthwhile at all to encode this as a `FilterFile` since it seems to follow a pattern: e.g. wrap everything like `$(.*)` in quotes.

repos/spack_repo/builtin/packages/qt/qt515_masm_python.patch
```diff
@@ -16,7 +16,7 @@ index b67ee79..b757ee5 100644
  retgen.input = retgen.script
  retgen.CONFIG += no_link
 -retgen.commands = $$QMAKE_PYTHON $$retgen.script > ${QMAKE_FILE_OUT}
-+retgen.commands = "\"$$QMAKE_PYTHON\"" $$retgen.script > ${QMAKE_FILE_OUT}
++retgen.commands = "\"$$QMAKE_PYTHON\"" "\"$$retgen.script\"" > "\"${QMAKE_FILE_OUT}\""
```
scheibelp: (question) I guess it would be difficult to come up with a function/filter that handles this and `icu4c` because it looks like these args need an extra layer of wrapping when assigned as a variable.

scheibelp (commented):
I have a couple questions about potentially extracting this into something general-purpose, since it looks like this is an issue that could affect many package builds (on Windows).

repos/spack_repo/builtin/packages/icu4c/Quote_datagen.patch
johnwparent: Yes, this is non Spack specific.
Qt6 switched to a CMake based build and I don't believe they're still updating their buildsystem for the 5.x release series. That said, I'll look into it. If they're still accepting contributions I agree this would be valuable to upstream.

repos/spack_repo/builtin/packages/qt/qt515_masm_python.patch
johnwparent: Maybe the logic could be genericized? It would be an incredibly painstaking process however, quoting has long been a thorn in Windows development and this situation is somewhat difficult to identify programmatically (or even as developers, as evidenced by the number of changes I've had to make here)

scheibelp (commented):
Reminder to myself: this includes https://github.com/spack/spack-packages/pull/541 and I should approve this once that merges (everything here on top of that looks good).

repos/spack_repo/builtin/packages/qt/package.py
```diff
@@ -135,6 +135,8 @@ class Qt(Package):
     patch("qt515-quick3d-assimp.patch", when="@5.15:5+opengl")
     # https://forum.qt.io/topic/130793/a-problem-with-python-path-when-i-try-to-build-qt-from-source-e-program-is-not-recognized-as-an-internal-or-external-command?_=1722965446110&lang=en-US
     patch("qt515_masm_python.patch", when="@5.15 platform=windows")
```
scheibelp: I think it would be good to use a naming scheme to indicate all the patches dealing with quoting, e.g. putting `quote_` as a prefix

repos/spack_repo/builtin/packages/qt/root_configure_path.patch
```diff
@@ -0,0 +1,14 @@
+diff --git a/configure.bat b/configure.bat
+index 500d627f..3d431df4 100644
+--- a/configure.bat
++++ b/configure.bat
+@@ -27,7 +27,7 @@
+ ::
+ :::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::
+ 
+-set "srcpath=%~dp0"
++set "srcpath=%~sdp0"
```
scheibelp: The quoting ones are somewhat self-explanatory, but I'm not sure what this one does from looking at the diff (it makes sense to redirect a path used at build time, but I can't recall spack setting `%~sdp0`)

scheibelp (changes requested):
I have a couple requests (docs and naming)

repos/spack_repo/builtin/packages/qt/package.py
johnwparent: Good idea, will do.

repos/spack_repo/builtin/packages/qt/root_configure_path.patch
johnwparent: `%` is batch variable dereference syntax. `0` is the argument we're dereferencing, which inside of a batch script, much the same as sh/bash/zsh/etc,  `0` is the name of the script. `~` Is an operator that allows us to modify the result of the dereferenced variable, and `sdp` are each operators on the value inside `%0`.
`~s` uses 8.3 or short file names if the variable is a path. 8.3 names are DOS hold overs that are valid names that just don't use spaces, so it's a workaround to quoting. Quoting is also a valid approach here, but I wanted to determine if we could have a more generic approach.
`~d` gives us the drive of the path
`~p` gives us the dirname without the drive.
Put together `~sdp` gives us the short file version of the drive and dirname of the script currently being executed. 
For more context see: https://ss64.com/nt/syntax-args.html


My investigation into the above translated into a branch. Basically I think we can almost always substitute `%~dp0` for `%~sdp0`, programmatically with filter file or similar. Currently looking into the best way to achieve that, whether it's just something we manually do for each package we port, or if it should be a default Spack behavior. 
I'll look into it a bit more and we can discuss at a later point.

repos/spack_repo/builtin/packages/qt/package.py
johnwparent: Pushed!

scheibelp (dismissed):

johnwparent:
@scheibelp everything is good on CI now, LGTM!

repos/spack_repo/builtin/packages/qt/package.py
```diff
@@ -575,15 +579,47 @@ def patch(self):
...
+                return False
+            return True if re.search(r"[ <>^:\"|?*]", arg) else False
+
+        return [quote(x) if has_space(x) or has_reserved(x) else x for x in args]
+
+    def _split_link_args(self, arg_str):
+        """Returns a list of the -L
+        arguments included in arg_str with proper
+        handling for paths with spaces"""
+        return ["-L" + x for x in arg_str.split("-L") if x]
```
scheibelp: I think this assumes no path has `-L` "in the middle" of it.

`search_flags`/`include_flags` are `-L`/`-I` of `.directories` (of the respective objects), so it would be safer to access those directly IMO

scheibelp (changes requested):

repos/spack_repo/builtin/packages/qt/package.py
johnwparent: Hmm, that a good point that `my-Library` is absolutely a legal path that could be in an install dir. 
I'll refactor. See also https://github.com/spack/spack/pull/51078 where I try to resolve this for the general case, not just Qt, without resorting to this

scheibelp (approved):

PR #477, opened 2025-07-04
vdt: add licence information

vvolkl:
See https://github.com/dpiparo/vdt/blob/master/Licence.txt


repos/spack_repo/builtin/packages/vdt/package.py
```diff
@@ -18,6 +18,8 @@ class Vdt(CMakePackage):
 
     maintainers("wdconinc")
 
+    license("LGPL-3.0-or-later")
```
vvolkl:
```suggestion
    license("LGPL-3.0-or-later", checked_by="vvolkl")
```

alalazo (approved):

PR #478, opened 2025-07-04
tandem: improve memalign, add version, and slight fixes

Thomas-Ulrich:
- fix compilation of tandem@1.0 %gcc@12
- narrow the range of petsc version available for each tandem version
- compile mpi with cuda_arch if available.

(https://github.com/spack/spack/pull/47189)


repos/spack_repo/builtin/packages/tandem/package.py
```diff
@@ -30,6 +31,7 @@ class Tandem(CMakePackage, CudaPackage, ROCmPackage):
     version("1.0", tag="v1.0", commit="eccab10cbdf5842ed9903fac7a023be5e2779f36", submodules=True)
 
     patch("fix_v1.0_compilation.diff", when="@1.0")
+    patch("fix_v1.1_compilation.diff", when="@1.1")
```
alalazo, line 37: Can you add a brief comment above the `patch` directive explaining:
2. Why it is needed
3. Where it was taken from

?

repos/spack_repo/builtin/packages/tandem/package.py
```diff
@@ -65,7 +76,16 @@ class Tandem(CMakePackage, CudaPackage, ROCmPackage):
...
 
     depends_on("zlib-api")
-    depends_on("petsc@3.16: +int64 +mumps +scalapack memalign=32")
+
+    conflicts("^petsc memalign=none")
+    conflicts("^petsc memalign=4")
+    conflicts("^petsc memalign=8")
+    conflicts("^petsc memalign=16")
+
+    depends_on("petsc +int64 +mumps +scalapack memalign=32")
```
alalazo: Why the conflicts? Those should be redundant because:
- `memalign` is a single valued variant
- which is specified in an unconditional `depend_on` directive

alalazo (changes requested):
Just a couple of questions

alalazo (approved):

PR #481, opened 2025-07-04
py_uv: add 0.8.8 (but needs rust-1.86+)

afuetterer:
Ref: https://pypi.org/project/uv/0.7.19/#uv-0.7.19.tar.gz

Edit by @bernhardkaindl: It should likely depend on #1028 to not build rust-nightly but use a stable version of rust instead when building uv.

### DIFF
--- a/repos/spack_repo/builtin/packages/py_uv/package.py
+++ b/repos/spack_repo/builtin/packages/py_uv/package.py
@@ -11,10 +11,11 @@ class PyUv(PythonPackage):
     """An extremely fast Python package and project manager, written in Rust."""
 
     homepage = "https://github.com/astral-sh/uv"
-    pypi = "uv/0.4.15.tar.gz"
+    pypi = "uv/uv-0.8.8.tar.gz"
 
     license("APACHE 2.0 or MIT")
 
+    version("0.8.8", sha256="6880e96cd994e53445d364206ddb4b2fff89fd2fbc74a74bef4a6f86384b07d9")
     version("0.6.8", sha256="45ecd70cfe42132ff84083ecb37fe7a8d2feac3eacd7a5872e7a002fb260940f")
     version("0.4.27", sha256="c13eea45257362ecfa2a2b31de9b62fbd0542e211a573562d98ab7c8fc50d8fc")
     version("0.4.17", sha256="01564bd760eff885ad61f44173647a569732934d1a4a558839c8088fbf75e53f")
@@ -22,6 +23,7 @@ class PyUv(PythonPackage):
     version("0.4.15", sha256="8e36b8e07595fc6216d01e729c81a0b4ff029a93cc2ef987a73d3b650d6d559c")
 
     depends_on("rust", type=("build", "run"))
+    depends_on("rust@1.86:", when="@0.7:")
     depends_on("python@3.8:", type=("build", "run"))
     depends_on("py-maturin@1:1", type="build")
     depends_on("cmake", type="build", when="@:0.6.3")


afuetterer:
I simply added a new version to package.py. Did CI succeed before on this package.py? Is it failing because there is no "py-uv" on pypi? A lot of spack packages have this py prefix. What I am doing wrong?

```
Error: Could not fetch any versions for py-uv
```

rbberger:
Manually verified checksum. @kwryankrattiger any idea what's going with the CI here?

bernhardkaindl (changes requested):
First of all, 0.7.19 needs at least rust-1.86 to build (it says so), while spack currently only has 1.85. So until rust is updated, it can't be built.

IF a package is submitted, the submitter should have ideally built it, not just add the checksum.

Secondly, it's already outdated (due to the time passed). At the moment 0.8.7 is the current version.

bernhardkaindl:
@afuetterer:
> I simply added a new version to package.py. Did CI succeed before on this package.py? Is it failing because there is no "py-uv" on pypi? A lot of spack packages have this py prefix. What I am doing wrong?
> ```
> Error: Could not fetch any versions for py-uv
> ```

Q:
> Did CI succeed before on this package.py?

No, not the current new CI, that runs `spack ci verify-versions HEAD~ HEAD`.
The old CI lacked this command, failed to check checksums, and was unable to detect if `spack checksum` was broken for that package.

> Is it failing because there is no "py-uv" on pypi?

No, it failed because, with the broken
```py
    pypi = "uv/0.4.15.tar.gz"
```

it checked
- https://files.pythonhosted.org/packages/source/u/uv/0.8.8.tar.gz which does not exist.
- https://files.pythonhosted.org/packages/source/u/uv/uv-0.8.8.tar.gz is the correct URL.

Therefore, it needed to be fixed to:
```py
    pypi = "uv/uv-0.8.8.tar.gz"
```

@rbberger:
> Manually verified checksum. kwryankrattiger any idea what's going with the CI here?

CI is fine, as explained. The issue was with the package, specifically the spack checksum, which the new check `spack ci verify-versions HEAD~1 HEAD` addresses.

To see the failing command `spack ci verify-versions HEAD~1 HEAD` itself, you had to click on the triangle to unfold the folded message where the command was started by CI.

afuetterer:
Sorry for the faulty PR and the extra work @bernhardkaindl. Thanks a lot for the fix and explanations.

bernhardkaindl (approved):
> Sorry for the faulty PR and the extra work @bernhardkaindl. Thanks a lot for the fix and explanations.

Yeah, that was tricky. I think the error message shown by GitLab CI should be improved. This update could be submitted to the main spack repo.

I as the spack package maintainer of rust will want to review the new rust version, I moved the new rust version to #1028:

Until the update of the rust recipes with #1028 is also applied to a spack-packages repo, spack will resort build `rust-nightly`, which of course isn't a stable version, but might be sufficient for building uv.

Alternatively, if the change to the rust recipes with #1028 is applied to the tree to build uv, rust-1.89.0 will be built. Even if #1028 is not applied, once spack has its build of rust-1.89.0, it can use it to build uv.

So this PR can already be merged with this caveat.

bernhardkaindl:
> Sorry for the faulty PR and the extra work @bernhardkaindl. Thanks a lot for the fix and explanations.

bernhardkaindl:
As building `rust-nightly` is probably not what everyone wants, I'll leave auto-merge off for now to possibly delay merging until after #1028 is merged.

afuetterer:
Hi, I saw that there now is a newer rust version available:
https://github.com/spack/spack-packages/blob/e79ba9fc76dd02568da737230bd5f570822fcbb9/repos/spack_repo/builtin/packages/rust/package.py#L41

Unfortunately I was not able to build py-uv using Rust 1.9.2.

I get several warnings like this `warning: value assigned to [...] is never read`, ..., `maybe it is overwritten before being read?` Then the build just hangs.

Should I close this PR?

alalazo:
> Unfortunately I was not able to build py-uv using Rust 1.9.2.
> Should I close this PR?

An alternative would be to update this PR with upper bounds:
```
depends_on("rust@:X.Y", when="@:A.B")
```

afuetterer:
I understand. But there is no upper bound to use. Because (as @bernhardkaindl pointed out) 1.86+ is needed, the next version of rust is 1.92 already.

```
version("1.92.0", sha256="9e0d2ca75c7e275fdc758255bf4b03afb3d65d1543602746907c933b6901c3b8")
version("1.86.0", sha256="022a27286df67900a044d227d9db69d4732ec3d833e4ffc259c4425ed71eed80")
```

bernhardkaindl:
@afuetterer:
> Should I close this PR?

Yes, please close this PR.
- uv-0.7.22 has been merged into this repo already
- 0.8.8 (the version in this PR) is already obsolte: Upstream has: uv-0.9.27:
  > https://pypi.org/project/uv/
  
 If you are able to do a working build of uv-0.9.27 in spack, you best submit that as a new PR.

alalazo:
Since it seems there is agreement on closing the PR, I'll just close. Feel free to reopen if I misunderstood the discussion above.

afuetterer:
Thanks, I will open a new PR for a recent version of uv, once I can built it locally.

PR #482, opened 2025-07-04
snakemake: add 9.6.3

afuetterer:
Ref: https://pypi.org/project/snakemake/9.6.3/#snakemake-9.6.3.tar.gz

### DIFF
--- a/repos/spack_repo/builtin/packages/snakemake/package.py
+++ b/repos/spack_repo/builtin/packages/snakemake/package.py
@@ -11,11 +11,12 @@ class Snakemake(PythonPackage):
     """Workflow management system to create reproducible and scalable data analyses."""
 
     homepage = "https://snakemake.readthedocs.io/en"
-    pypi = "snakemake/snakemake-8.25.2.tar.gz"
+    pypi = "snakemake/snakemake-9.6.3.tar.gz"
     maintainers("marcusboden", "w8jcik")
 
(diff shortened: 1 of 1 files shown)


alalazo (approved):

wdconinc:
This PR should have gotten a better review (it may make sense to add snakemake to a CI stack).

This is a major version upgrade, but none of the dependencies are updated as required per the setup.cfg to pyproject.toml transition and changes. The package can't run when installed, since the required py-snakemake-interface-common version isn't available in spack. The maintainers were not offered a chance to review, @marcusboden @w8jcik, and would have likely caught this.

I am going to mark the most recent 8-series version as preferred until someone has the time to fix this. See https://github.com/spack/spack-packages/pull/1191.

afuetterer:
Oh no, I am sorry for messing up here. Thanks for pointing that out @wdconinc,

wdconinc:
> Oh no, I am sorry for messing up here. Thanks for pointing that out @wdconinc,

No problem. We could have given better feedback ;-) Do you want to open a PR to add the updates to e.g. py-snakemake-interface-common (1.20 required now) and verify the other dependencies? I can give it a review, but I won't have time myself to do the digging to determine what's all needed.

afuetterer:
Thanks, yes, let me fix that myself.

I will open a new PR for that.

PR #485, opened 2025-07-04
aws-ofi-nccl: add v1.14.0, 1.14.1, 1.16.0

msimberg:
Adding 1.16.0 plus a couple of intermediate 1.14.X releases as it turns out the minor version is more significant than it seems (1.14.2 does not work exactly the same or better than 1.14.1 and 1.14.0).


albestro (dismissed):

alalazo (approved):
This seems stuck rebuilding the unrelated `hipblaslt` which frequently times out after 6hrs :disappointed: Merging the PR.

PR #487, opened 2025-07-04
hdfview: add new version and java usage

Chrismarsh:
This
- adds version 3.3.2
- ensure java as a direct dependency and ensure the correct version is used
- fixes how java is specified in the `hdfview.sh` wrapper script. Even with 3.3.0 I couldn't launch the GUI because the classpaths were wrong. It's not clear to me when this might have changed, or that it ever worked before as paths that don't exist are used as cps.
- fixes JAVABIN in wrapper script to point to java. Also unclear that this ever worked previously.
- The URL for 3.3.2+ has updated so now directly uses the github release source tarball

### DIFF
--- a/repos/spack_repo/builtin/packages/hdfview/package.py
+++ b/repos/spack_repo/builtin/packages/hdfview/package.py
@@ -59,7 +59,7 @@ class Hdfview(Package):
 
         # set the javabin to be spack's java
         filter_file(
-            r"export JAVABIN.+", f"export JAVABIN={self.spec["java"].prefix}/bin/", hdfview
+            r"export JAVABIN=.+", f"export JAVABIN={self.spec['java'].prefix}/bin", hdfview
         )
 
         mkdirp(prefix.bin)
--- a/repos/spack_repo/builtin/packages/hdfview/package.py
+++ b/repos/spack_repo/builtin/packages/hdfview/package.py
@@ -58,9 +58,9 @@ class Hdfview(Package):
         filter_file(r"/lib/app", "", hdfview)
 
         # set the javabin to be spack's java
-        filter_file(r"export JAVABIN.+", 
-            f"export JAVABIN={self.spec["java"].prefix}/bin/",
-             hdfview)
+        filter_file(r"export JAVABIN.+",
+                    f"export JAVABIN={self.spec["java"].prefix}/bin/",
+                    hdfview)
 
         mkdirp(prefix.bin)
         install(hdfview, prefix.bin.hdfview)
@@ -83,4 +83,4 @@ class Hdfview(Package):
 
         url_fmt = "https://support.hdfgroup.org/ftp/HDF5/releases/HDF-JAVA/hdfview-{0}/src/hdfview-{0}.tar.gz"
         return url_fmt.format(version)
-        
+
--- a/repos/spack_repo/builtin/packages/hdfview/package.py
+++ b/repos/spack_repo/builtin/packages/hdfview/package.py
@@ -78,7 +78,7 @@ class Hdfview(Package):
 
         # the new versions have a complex https://objects.githubusercontent.com url so use the
         # github release
-        if version >= Version("3.3.0"):
+        if version > Version("3.3.0"):
             return super().url_for_version(version)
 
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/hdfview/package.py
```diff
@@ -45,8 +50,18 @@ def install(self, spec, prefix):
...
+        # set the internal dir to be the spack install prefix
         filter_file(r"\$dir", prefix, hdfview)
 
+        # the wrapper script looks for these subdirectories, however they are not
+        # in the spackk install prefix, so filter out
+        filter_file(r"/lib/app", "", hdfview)
+
+        # set the javabin to be spack's java
+        filter_file(
+            r"export JAVABIN.+", f"export JAVABIN={self.spec["java"].prefix}/bin/", hdfview
```
alalazo: Syntax error :disappointed:

alalazo (changes requested):
https://gitlab.spack.io/spack/spack-packages/-/jobs/17269530

repos/spack_repo/builtin/packages/hdfview/package.py
Chrismarsh: Ooof sorry about that. I'm on 3.12 which [allows quote reuse in fstrings](https://docs.python.org/3/whatsnew/3.12.html#whatsnew312-pep701), which is why I didn't catch this.
I wonder if this can/should be added to the style checks?

I have switched to single quotes so hopefully all sorted now

tgamblin: we run vermin on both core and packages... wonder why this wasn't caught.

kwryankrattiger: This is the vermin command we run in CI

```
vermin --backport importlib --backport argparse --violations --backport typing -t=3.6- -vvv repos tests
```

I had copied that from spack core GHA and updated the paths, maybe I did it wrong?

Also there is an error reported  by vermin about a patch file here. I don't see an exclusions file so maybe there is an issue with the exit code?

```
Analyzing 8568 files using 4 processes..
/home/runner/work/spack-packages/spack-packages/repos/spack_repo/builtin/packages/phist/sbang.patch:2:7:error: invalid decimal literal: index 6eb9051..874ce48 100755
```

Chrismarsh: It looks like vermin doesn't detect the nested fstring quotes 
https://github.com/netromdk/vermin/issues/264

repos/spack_repo/builtin/packages/hdfview/package.py
```diff
@@ -57,3 +72,14 @@ def setup_build_environment(self, env: EnvironmentModifications) -> None:
         env.set("HDF5LIBS", self.spec["hdf5"].prefix)
         env.set("HDFLIBS", self.spec["hdf"].prefix)
         env.set("ANT_HOME", self.spec["ant"].prefix)
+        env.set("JAVA_HOME ", self.spec["java"].prefix)
+
+    def url_for_version(self, version):
+
+        # the new versions have a complex https://objects.githubusercontent.com url so use the
+        # github release
+        if version >= Version("3.3.0"):
```
tgamblin: I wasn't able to curl 3.3.0 from the URL displayed in `spack info` -- is this function right?

Chrismarsh: Great catch thanks. It should be `>` .  I didn't catch the problem as when I tested it (and confirmed just now), spack used its cache papering over the issue

```
==> No binary for hdfview-3.3.0-ycvmnj7277lubbwb4q5bvfqt7qspa6uz found: installing from source
==> Installing hdfview-3.3.0-ycvmnj7277lubbwb4q5bvfqt7qspa6uz [31/31]
==> Fetching https://mirror.spack.io/_source-cache/archive/09/0916161861c21fa8dd354b445b48eff5a53d80a5c0b383e79eb64b7b108e2430.tar.gz
    [100%]  123.83 MB @   17.9 MB/s
==> No patches needed for hdfview

```

Chrismarsh:
@alalazo do these changes work for you?

alalazo (approved):

PR #495, opened 2025-07-06
lhapdf: add gpu version

ebothmann:
This adds the `gpu` version, which installs a feature branch of LHAPDF with support for usage on GPU (via CUDA and via Kokkos). This feature has not yet been integrated into an LHAPDF release, but it is very useful, e.g. to run parton-level event generators with GPU acceleration, such as [Pepper](https://gitlab.com/spice-mc/pepper). If this Pull Request is accepted, I plan to submit another one which adds a Spack package for Pepper.

This is my first contribution to `spack-packages`.

### DIFF
--- a/repos/spack_repo/builtin/packages/lhapdf/package.py
+++ b/repos/spack_repo/builtin/packages/lhapdf/package.py
@@ -24,6 +24,7 @@ class Lhapdf(AutotoolsPackage):
 
     license("GPL-3.0-or-later")
 
+    version("gpu", branch="kokkos_version")
     version("6.5.5", sha256="d20d8fb71936403274caec5bd584c891592b96c6319175df51d9bb69db869bd8")
     version("6.5.4", sha256="ace8913781044ad542e378697fcd95a8535d510818bb74a6665f9fd2b132ac0f")
     version("6.5.3", sha256="90fe7254d5a48a9b2d424fcbac1bf9708b0e54690efec4c78e9ad28b9203bfcd")


wdconinc:
Thanks for the contribution! In general we are a bit hesitant to include versions on development branches since the dependencies are often still in flux. In any case, here those dependencies do need to be explicitly depended on.

ebothmann:
Thank you for your review, @wdconinc.

I completely understand the reservation about development branches. However, I’m in contact with the LHAPDF authors, and it will take a while until this development branch is integrated into upstream, due to the chronic lack of time for technical work in theoretical physics 😓 (tagging @maxkno as the author of the development branch)

At the same time, the currently very active development of GPU/HPC accelerated Monte-Carlo event generators (in view of the upcoming High Luminosity phase of the LHC) would profit from this development version being available via Spack, as this takes away some of the friction when deploying to a new HPC machine.

Regarding the dependencies, there are in fact none (or at least no additional ones with respect to the release versions). The changes to support evaluation on CUDA or via Kokkos are entirely header-based, and CUDA etc. does not need to be available at compile time of the LHAPDF package. Only software that wants to link to LHAPDF and wants to use CUDA etc. will need to address this. That’s actually the beauty of @maxkno’s changes.

What do you think?

wdconinc:
> Regarding the dependencies, there are in fact none (or at least no additional ones with respect to the release versions). The changes to support evaluation on CUDA or via Kokkos are entirely header-based, and CUDA etc. does not need to be available at compile time of the LHAPDF package. Only software that wants to link to LHAPDF and wants to use CUDA etc. will need to address this. That’s actually the beauty of @maxkno’s changes.

I had a quick look at https://gitlab.com/hepcedar/lhapdf/-/merge_requests/91 and I understand what you mean about the header-only nature. Nevertheless, I think from a spack user point of view, I think we will want a dependency on CUDA and Kokkos, for two reasons:
- if there is (ever, maybe not now) a requirement for a specific minimum version of either CUDA or Kokkos within the LHAPDF code, then it will have to be specified somewhere, and we can't expect downstream code that includes these headers to have to take on that role.
- if a user installs `lhapdf@gpu`, then we want that user to end up with a reasonable set of dependencies to use wha they just installed; if we don't do that, then it's only a matter of time before someone will file a big report adding the dependencies since that's their expectation.

It is not a problem that the code that *uses* this new LHAPDF version will likely *also* want to specify the dependency on CUDA and/or Kokkos.

So, for the CUDA dependency, I think it will be sufficient to inherit `CUDAPackage` as well, and for the Kokkos dependency, I'd add a `depends_on("kokkos", when="+cuda")`.

wdconinc:
@ebothmann Before the year runs out, I'm just checking on some long-open PRs. Has there been any movement towards integrating this functionality into lhapdf directly? From our point of view it would be easiest if this becomes functionality that is avialable as of the next lhapdf release version.

ebothmann:
Dear @wdconinc, Happy New Year!

I will try to make this happen, but it’ll likely take a few more months to get back to this. So I’d close this PR until then. Sorry for leaving this open so long.

PR #498, opened 2025-07-07
Bump Spack commit (fix pipelines pruning in PRs)

alalazo:


alalazo:
Without package changes it behaves as expected:

Screenshot from 2025-07-07 11-48-58

Now pushing a package change to check everything becomes "no spec to rebuild"

alalazo:
It seems to work:
Screenshot from 2025-07-07 13-51-22

PR #500, opened 2025-07-07
omb-micro-benchmarks: add v7.5.1

MatthewLieber:
Adding the hash for the v7.5.1 release of OSU Mircobenchmarks

### DIFF
--- a/repos/spack_repo/builtin/packages/osu_micro_benchmarks/package.py
+++ b/repos/spack_repo/builtin/packages/osu_micro_benchmarks/package.py
@@ -23,7 +23,7 @@ class OsuMicroBenchmarks(AutotoolsPackage, CudaPackage, ROCmPackage):
 
     maintainers("natshineman", "harisubramoni", "MatthewLieber")
 
-    version("v7.5.1", sha256="b4c696b3d9ceabb8263ab24e01436871951c02272bb7afc47a25a4900b54ec6e")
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/osu_micro_benchmarks/package.py
```diff
@@ -23,6 +23,7 @@ class OsuMicroBenchmarks(AutotoolsPackage, CudaPackage, ROCmPackage):
 
     maintainers("natshineman", "harisubramoni", "MatthewLieber")
 
+    version("v7.5.1", sha256="b4c696b3d9ceabb8263ab24e01436871951c02272bb7afc47a25a4900b54ec6e")
```
rbberger: Shouldn't this just be `7.5.1` instead of `v7.5.1`?

rbberger (changes requested):

repos/spack_repo/builtin/packages/osu_micro_benchmarks/package.py
rbberger: @MatthewLieber ping.

MatthewLieber: @rbberger sorry for the late reply thanks for catching this I have pushed the change to keep the naming convention constant

rbberger (approved):

PR #503, opened 2025-07-07
BlisBase: Exposing the optional ability to target alternative build configurations

G-Ragghianti:
As detailed in issue https://github.com/spack/spack/issues/49376 .  The current spack recipe forces the use of the "auto" build configuration which makes cross-compilation impossible.  This change adds a new variant which will allow the selection of alternative build configurations.  The original behavior of the package is preserved.

### DIFF
--- a/repos/spack_repo/builtin/packages/blis/package.py
+++ b/repos/spack_repo/builtin/packages/blis/package.py
@@ -11,6 +11,13 @@ from spack.package import *
 # https://github.com/flame/blis/issues/195
 # https://github.com/flame/blis/issues/197
 
+# Define a mapping between the selected architecture and the Blis config target
+_targets = [
+    # Arch, Target
+    ["x86_64", "x86_64"],
+    ["zen", "zen"],
+]
+
 
 class BlisBase(MakefilePackage):
     """Base class for building BLIS, shared with the AMD optimized version
@@ -83,8 +90,12 @@ class BlisBase(MakefilePackage):
         return config_args
 
     def edit(self, spec, prefix):
-        # To ensure auto should always be the last argument for base and derived class
-        config_args = self.configure_args() + ["auto"]
+        # To ensure the target should always be the last argument for base and derived class
+        target = "auto"
+        for _arch, _target in _targets:
+            if self.spec.satisfies(f"target={_target}"):
+                target = _target
+        config_args = self.configure_args() + [target]
         configure("--prefix={0}".format(prefix), *config_args)
 
     @run_after("install")
--- a/repos/spack_repo/builtin/packages/blis/package.py
+++ b/repos/spack_repo/builtin/packages/blis/package.py
@@ -11,9 +11,9 @@ from spack.package import *
 # https://github.com/flame/blis/issues/195
 # https://github.com/flame/blis/issues/197
 
-# If the spack target architecture matches one of these values, 
-# provide this target to the Blis build as the Blis config target 
-# instead of using automatic configuration detection. 
+# If the spack target architecture matches one of these values,
+# provide this target to the Blis build as the Blis config target
+# instead of using automatic configuration detection.
 _targets = ["x86_64", "zen", "zen2"]
 
 
--- a/repos/spack_repo/builtin/packages/blis/package.py
+++ b/repos/spack_repo/builtin/packages/blis/package.py
@@ -11,12 +11,10 @@ from spack.package import *
 # https://github.com/flame/blis/issues/195
 # https://github.com/flame/blis/issues/197
 
-# Define a mapping between the selected architecture and the Blis config target
-_targets = [
-    # Arch, Target
-    ["x86_64", "x86_64"],
-    ["zen", "zen"],
-]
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/blis/package.py
```diff
@@ -84,7 +85,8 @@ def configure_args(self):
 
     def edit(self, spec, prefix):
         # To ensure auto should always be the last argument for base and derived class
-        config_args = self.configure_args() + ["auto"]
+        mconfig = self.spec.variants["config"].value
```
alalazo: I think we should check the `target`, not add a new variant for `blis`, right?

G-Ragghianti: Wouldn't this require setting up and maintaining a lookup table mapping all the possible targets to the configurations that come with amdblis?  This seems like a lot of work and quite fragile.  Also, some of the included configurations include build parameters that are not purely a function of the target architecture.

The method here just exposes an existing configuration parameter of the package(s), and retains the existing behavior which the blis team seems very interested in maintaining.

alalazo: Doesn't this interfere with `target=` ? What happens if I set `config=<icelake>` and `target=x86_64`? Are those allowed configurations?

Also how do we know which values are allowed? Downloading `blis` and running configure to know which values I can use in a package manager that should manage the builds for me doesn't seem a good option :slightly_smiling_face:

G-Ragghianti: Right now, it is essentially ignoring the target.  Using target= doesn't allow cross compiling because the build system for Blis doesn't just use the compiler flags for optimizations, but also builds in optimizations targeting specific arch using the "config" setting during the build.  This is because the hard-coded value is config=auto which just looks at the CPU arch of the system you are building on.  There is no way to override this (in spack) to set any other CPU arch.  If you are building it by hand, then you can just set config to the appropriate value for your site.  I'm just trying to expose this functionality to make the package usable.

We can enumerate the currently available config values in the amdblis package, but this would have to be maintained.  Also, because this is actually in BlisBase class, other packages than amdblis may have different values.  To avoid this problem, we could just implement this in the amdblis package (suffficient for us).

G-Ragghianti: If you want to do this by checking the target and mapping that to a blis config, where do we get a comprehensive list of the possible targets? Can we just implement a limited subset of mappings and open the possibility of adding other maps in the future?  That would be sufficient for us.

alalazo: > We can enumerate the currently available config values in the amdblis package, but this would have to be maintained.

Yes, that's what I am suggesting. We can select the appropriate option based on the value of `target=`, and avoid adding a variant. If that needs to be in a derived class, that's fine.

Using a generic string value doesn't seem a good option because:
1. Users would have no clue about possible values, when using `spack info`
2. We may conflict with the guarantee that `target=` is supposed to give.

alalazo: > where do we get a comprehensive list of the possible targets?
```console
$ spack arch --known-targets
```
should give you the complete list of targets known by the version of Spack you're using.

> Can we just implement a limited subset of mappings and open the possibility of adding other maps in the future? 

Sure. If that is sufficient for your use case you can just check generic targets, e.g.:
```python
if self.spec.satisfies("target=x86_64_v4:"):
    # options for AVX512
elif self.spec.satisfies("target=x86_64_v3:"):
    # options for AVX2
```
etc. Does that seem a reasonable way forward?

G-Ragghianti: > > Can we just implement a limited subset of mappings and open the possibility of adding other maps in the future?
> 
> Sure. If that is sufficient for your use case you can just check generic targets, e.g.:
> 
> ```python
> if self.spec.satisfies("target=x86_64_v4:"):
>     # options for AVX512
> elif self.spec.satisfies("target=x86_64_v3:"):
>     # options for AVX2
> ```
> 
> etc. Does that seem a reasonable way forward?

Yes, I'll look into implementing it that way.

G-Ragghianti: @alalazo OK, this is implemented and tested.  Looks good?

repos/spack_repo/builtin/packages/blis/package.py
```diff
@@ -83,8 +90,12 @@ def configure_args(self):
         return config_args
 
     def edit(self, spec, prefix):
-        # To ensure auto should always be the last argument for base and derived class
-        config_args = self.configure_args() + ["auto"]
+        # To ensure the target should always be the last argument for base and derived class
+        target = "auto"
+        for _arch, _target in _targets:
+            if self.spec.satisfies(f"target={_target}"):
+                target = _target
```
alalazo, line 94: I assume this is to map a Spack name to a Blis name, and you meant:
```suggestion
                target = _arch
```
? Otherwise I don't see where `_arch` is used...

alalazo (approved):

PR #504, opened 2025-07-07
nvshmem: introduce cmake, add new versions 3 and libfabric variant

albestro:
It's far from being nice and perfect, but at least it is a starting point for having newer `nvshmem` versions available, together with some of its features not currently exposed (e.g. `libfabric` support).

The ugliest thing is that, since filename has been changed from the simple  `nvshmem_src_{version}.txz"` to `nvshmem_src_cuda12-all-all-{version}.tar.gz`, I opted for an "on-spot" solution.

### DIFF
--- a/repos/spack_repo/builtin/packages/nvshmem/package.py
+++ b/repos/spack_repo/builtin/packages/nvshmem/package.py
@@ -3,7 +3,7 @@
 # SPDX-License-Identifier: (Apache-2.0 OR MIT)
 
 from spack_repo.builtin.build_systems import cmake, makefile
-from spack_repo.builtin.build_systems.cmake import CMakePackage
+from spack_repo.builtin.build_systems.cmake import CMakePackage, generator
 from spack_repo.builtin.build_systems.cuda import CudaPackage, conflicts
 from spack_repo.builtin.build_systems.makefile import MakefilePackage
 
@@ -54,6 +54,8 @@ class Nvshmem(MakefilePackage, CMakePackage, CudaPackage):
     )
     variant("libfabric", default=False, description="Build with Libfabric support")
 
+    generator("ninja")
+
     conflicts("~cuda")
 
     def url_for_version(self, version):
@@ -70,7 +72,9 @@ class Nvshmem(MakefilePackage, CMakePackage, CudaPackage):
 
     depends_on("cuda@11:", when="@3.2.5:")
 
-    depends_on("cmake@3.19:", when="build_system=cmake")
+    with default_args(when="build_system=cmake", type="build"):
+        depends_on("cmake@3.19:")
+        depends_on("ninja")
 
     depends_on("mpi", when="+mpi")
 
--- a/repos/spack_repo/builtin/packages/nvshmem/package.py
+++ b/repos/spack_repo/builtin/packages/nvshmem/package.py
@@ -98,7 +98,6 @@ class CMakeBuilder(cmake.CMakeBuilder):
             self.define_from_variant("NVSHMEM_USE_NCCL", "nccl"),
             self.define_from_variant("NVSHMEM_USE_GDRCOPY", "gdrcopy"),
             self.define_from_variant("NVSHMEM_SHMEM_SUPPORT", "shmem"),
-            self.define("NVSHMEM_PMIX_SUPPORT", True),
             self.define("NVSHMEM_IBRC_SUPPORT", False),
             self.define("NVSHMEM_BUILD_PYTHON_LIB", False),
             self.define("NVSHMEM_BUILD_EXAMPLES", False),
--- a/repos/spack_repo/builtin/packages/nvshmem/package.py
+++ b/repos/spack_repo/builtin/packages/nvshmem/package.py
@@ -56,7 +56,7 @@ class Nvshmem(MakefilePackage, CMakePackage, CudaPackage):
 
     generator("ninja")
 
-    conflicts("~cuda")
+    conflicts("~cuda", msg="NVSHMEM requires CUDA")
 
     def url_for_version(self, version):
         ver_str = "{0}".format(version)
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/nvshmem/package.py
```diff
@@ -42,22 +52,83 @@ class Nvshmem(MakefilePackage, CudaPackage):
...
+    def cmake_args(self):
+        config = [
+            self.define("CMAKE_CUDA_ARCHITECTURES", self.spec.variants["cuda_arch"].value),
+            self.define_from_variant("NVSHMEM_MPI_SUPPORT", "mpi"),
+            self.define_from_variant("NVSHMEM_LIBFABRIC_SUPPORT", "libfabric"),
+            self.define_from_variant("NVSHMEM_UCX_SUPPORT", "ucx"),
+            self.define_from_variant("NVSHMEM_USE_NCCL", "nccl"),
+            self.define_from_variant("NVSHMEM_USE_GDRCOPY", "gdrcopy"),
+            self.define_from_variant("NVSHMEM_SHMEM_SUPPORT", "shmem"),
+            self.define("NVSHMEM_PMIX_SUPPORT", True),
```
msimberg: I guess you're trying to minimize the number of new variants introduced in this PR, so hardcoding makes sense for the below options. But if this is true, what pmix is this going to pick up? There's a pmix package, so should this maybe be behind a variant and `depends_on("pmix", when="@3: +pmix")`?

repos/spack_repo/builtin/packages/nvshmem/package.py
```diff
@@ -42,22 +52,83 @@ class Nvshmem(MakefilePackage, CudaPackage):
         when="@2.6:",
         description="Build with support for GPU initiated communication",
     )
+    variant("libfabric", default=False, description="Build with Libfabric support")
```
msimberg, line 55: Do you know if `+libfabric` conflicts with `+ucx` or vice versa?

albestro: AFAICT, they are completely independent.

repos/spack_repo/builtin/packages/nvshmem/package.py
albestro: Yes, exactly. It's a good point. I have to dig a bit more.

From [NVSHMEM doc](https://docs.nvidia.com/nvshmem/archives/nvshmem-241/pdf/NVSHMEM-Installation-Guide.pdf)

> NVSHMEM supports the following methods to launch your application:
> ‣ Use of a PMI-1 compatible launcher, such as Hydra.
> ‣ Use of a PMI-2 compatible launcher, such as Slurm.
> ‣ Use of a PMIx compatible launcher, such as Slurm or Open MPI mpirun.
> ‣ Launching as part of an existing MPI application.
> ‣ Launching as part of an existing OpenSHMEM application.

and about `PMI*` it also reports

> The PMI-1 and PMI-2 clients are in NVSHMEM and are automatically built as part of the
build process. A PMIx client must be provided by the user by installing Open PMIx or by using the PMIx client that is installed by Open MPI or Slurm. 

and for what concerns me more in limitations section of the [last release notes](https://docs.nvidia.com/nvshmem/release-notes-install-guide/release-notes/release-3309.html#limitations)

> NVSHMEM is not compatible with the PMI client library on Cray systems and must use the NVSHMEM internal PMI-2 client library.

Which is not perfectly clear to me what it means (due to a lack of knowledge on my side).

Again, I have to dig a bit more, but I would say that we can either:
- drop `pmix` and add a dependency on `pmi` somehow (it is not a virtual afaik)
- add a dependency on `pmix`, set it as default

The latter solution seems more generic and easier to describe.

msimberg: > - add a dependency on pmix, set it as default
>
> The latter solution seems more generic and easier to describe.

Yep, this sounds correct to me as well.

albestro: In the end, I quickly tested it without PMIx and it works. I would go as simple as that for the moment.

repos/spack_repo/builtin/packages/nvshmem/package.py
```diff
@@ -42,22 +52,86 @@ class Nvshmem(MakefilePackage, CudaPackage):
...
+
     depends_on("gdrcopy", when="+gdrcopy")
+    conflicts("~gdrcopy", when="~ucx")
+    depends_on("gdrcopy@2.0:", when="@3: +gdrcopy")
+
     depends_on("nccl", when="+nccl")
+    depends_on("nccl@2.0:", when="@3: +nccl")
+
+    depends_on("libfabric", when="+libfabric")
+    depends_on("libfabric@1.15.0.0:", when="@3: +libfabric")
```
msimberg: Minor: can this be
```suggestion
    depends_on("libfabric@1.15:", when="@3: +libfabric")
```
?

Same for `nccl`, `gdrcopy`, and `ucx` above?

repos/spack_repo/builtin/packages/nvshmem/package.py
msimberg: Sounds good. And PMIX is disabled by default in CMake (I ask since it's not explicitly set to false here)?

albestro: > PMIX is disabled by default in CMake

Yes, I checked and at least since 2.9.0 (the first version supporting CMake) PMIX_SUPPORT should be disabled by default.

repos/spack_repo/builtin/packages/nvshmem/package.py
albestro: see https://github.com/spack/spack-packages/pull/504/commits/e61b744a68d74ffcafea2a0ef313f2650822e70d

msimberg (dismissed):
Looks good to me. Waiting to see if @bvanessen has comments.

msimberg (approved):

PR #505, opened 2025-07-07
Disable running hipblaslt tests as they always timeout

kwryankrattiger:
This test is broken, so disabling for now.


alalazo:
I don't think it's a post install test that breaks - it seems more a sort of test run during the build of the package:
- https://gitlab.spack.io/spack/spack-packages/-/jobs/17259654

kwryankrattiger:
Hm yeah I think you are right there. That is unfortunate.

PR #507, opened 2025-07-07
ci: add triage action to label pull requests

alecbcs:
Add a triage job to label pull requests with a GitHub Action.

### DIFF
--- a/.github/workflows/bin/spack-labeler.py
+++ b/.github/workflows/bin/spack-labeler.py
@@ -1,0 +1,99 @@
+# Copyright Spack Project Developers. See COPYRIGHT file for details.
+#
+# SPDX-License-Identifier: (Apache-2.0 OR MIT)
+
+import os
+import re
+
+import requests
+import yaml
+
+
+def import_labels_config(path: str):
+    with open(path) as fd:
+        try:
+            label_patterns = yaml.safe_load(fd)
+        except yaml.YAMLError as exc:
+            raise Exception(f"Unable to load {path}\n{exc}")
+
+        # compile all the regexes above, and ensure that all pattern dict values are lists
+        for label, pattern_dict in label_patterns.items():
+            for attr in pattern_dict.keys():
+                patterns = pattern_dict[attr]
+                if not isinstance(patterns, list):
+                    patterns = [patterns]
+                pattern_dict[attr] = [re.compile(s) for s in patterns]
+
+    return label_patterns
+
+
+def main():
+    labels_config_path = os.environ.get("LABELS_CONFIG")
+    repository = os.environ.get("GH_REPO")
+    pr_number = os.environ.get("GH_PR_NUMBER")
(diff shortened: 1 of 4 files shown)


kwryankrattiger:
@spackbot fix style

adamjstewart:
If we're moving away from spackbot, I would be in favor of using something like https://github.com/actions/labeler. Simpler, but sometimes simple is nice.

alecbcs:
I use labeler in other projects and really like it, but it doesn't support tags that come from the patch of the PR. For example, we could have `update-package`, but we couldn't have `new-version` because it only looks at the names of the files modified.

I find labels like `new-version`, `conflicts`, and `patch` helpful when reviewing a PR and would prefer to keep if possible.

adamjstewart:
I guess I've never personally found those hyper-specific tags to be very useful. Most PRs end up having several tags. But if others want this, then I'm fine with your PR.

alecbcs:
Talked about this in the meeting today, general consensus was to keep the existing hyper-specific tags for now at least.

Waiting on final say from @tldahlgren and @adamjstewart. I'm not sure why the GitLab check is failing but appears unrelated to this PR as we didn't modify any packages.

adamjstewart (dismissed):
Can't easily validate this myself, but I trust you.

.github/workflows/bin/spack-labeler.py
```diff
@@ -0,0 +1,98 @@
+# Copyright 2013-2021 Lawrence Livermore National Security, LLC and other
```
tldahlgren: Shouldn't this line be removed?

tldahlgren (changes requested):
See nit. Don't need two references to the Copyright .

tldahlgren:
> Can't easily validate this myself, but I trust you.

Same here.

.github/workflows/bin/spack-labeler.py
```diff
@@ -0,0 +1,98 @@
...
+# Spack Project Developers. See the top-level COPYRIGHT file for details.
```
tldahlgren:
```suggestion
# Copyright Spack Project Developers. See the top-level COPYRIGHT file for details.
```

tldahlgren (approved):

.github/workflows/bin/spack-labeler.py
alecbcs: Good catch. Fixed the license header!

alecbcs:
> > Can't easily validate this myself, but I trust you.
> 
> Same here.

If you all are interested you can run the script by creating a PAT and running,

```bash
GH_PR_NUMBER=123 GH_REPO=spack/spack-packages LABELS_CONFIG=.github/labels.yml GH_TOKEN=<<PERSONAL_ACCESS_TOKEN>> python3 .github/workflows/bin/spack-labeler.py
```

tgamblin (approved):

PR #297, opened 2025-06-17
gcc: add +amdgcn to enable omp offload to AMD GPUs; update +nvptx logic

eugeneswalker:
GCC: add `+amdgcn` variant to enable OpenMP offload to AMD GPUs.

Originally published to spack/spack by @wyphan 
* https://github.com/spack/spack/pull/35919

This is a work-in-progress. There are still some issues that need to be worked out, which will be documented in comments on this PR. Don't merge yet. 

CC @kweide

PR #508, opened 2025-07-07
gcc: update +nvptx logic

kweide:
I found that some of these `+nvpxt` changes appeared to be necessary for building a functioning GCC (including gfortran) with support for OpenMP offloading to an Nvidia GPU. Problems with AMD support should not hold up getting this into the package repository.

This commit takes just the changes from PR #297 that are relevant to +nvptx, omitting the changes that add support for +amdgcn.

The commit was constructed by taking the changes from the mentioned PR (basically commit c4dd45c1a4e9917b1f3a6fe63659eed150714c68) and then backing out the changes relevant to +amdgcn. Thus
 - some of the changes here may appear a bit unmotivated when viewed on their own, but
 - adding support for +amdgcn later, perhaps with a reworked PR #297, should be straightforward.


kweide:
CC @eugeneswalker

eugeneswalker (approved):
Thank you for doing this!

kweide:
@alalazo  @michaelkuhn
Can this be merged now?
Should some test(s) be added for the updated feature (+nvptx)?  (I don't know how to something like that.)

PR #512, opened 2025-07-07
legion: various minor updates

rbberger:
- make `fortran` dependency optional (it's been causing problems in some mixed compiler configurations on `develop`)
- correct mistake that was pulling in UCC prior to 25.03.0
- some cleanup of existing `cmake_args` logic
- remove redundant requirements for `stable`.

### DIFF
--- a/repos/spack_repo/builtin/packages/legion/package.py
+++ b/repos/spack_repo/builtin/packages/legion/package.py
@@ -392,7 +392,25 @@ class Legion(CMakePackage, ROCmPackage):
     def cmake_args(self):
         spec = self.spec
         from_variant = self.define_from_variant
-        options = [from_variant("CMAKE_CXX_STANDARD", "cxxstd")]
+        options = [
+            from_variant("CMAKE_CXX_STANDARD", "cxxstd"),
+            from_variant("BUILD_SHARED_LIBS", "shared"),
+            from_variant("Legion_BOUNDS_CHECKS", "bounds_checks"),
+            from_variant("Legion_PRIVILEGE_CHECKS", "privilege_checks"),
+            from_variant("Legion_SPY", "spy"),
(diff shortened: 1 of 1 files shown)


elliottslaughter (approved):
LGTM.

repos/spack_repo/builtin/packages/legion/package.py
```diff
@@ -138,7 +138,8 @@ class Legion(CMakePackage, ROCmPackage):
     # cuda-centric
     cuda_arch_list = CudaPackage.cuda_arch_values
     for arch in cuda_arch_list:
-        depends_on(f"ucc cuda_arch={arch}", when=f"network=ucx +cuda cuda_arch={arch}")
+        depends_on(f"ucc cuda_arch={arch}", when=f"@25.03.0: network=ucx +cuda cuda_arch={arch}")
+        depends_on(f"ucc cuda_arch={arch}", when=f"@stable network=ucx +cuda cuda_arch={arch}")
```
kwryankrattiger: This shouldn't be required, Spack recognizes `stable` as an infinity version, ie it should be captured already with `@25.03.0:`

kwryankrattiger: ref. https://github.com/spack/spack/blob/develop/lib/spack/spack/version/common.py#L13

```
infinity_versions = ["stable", "nightly", "trunk", "head", "master", "main", "develop"]
```

rbberger: Ah, good. I didn't know it was part of that set. More cleanup incoming...

kwryankrattiger (approved):

repos/spack_repo/builtin/packages/legion/package.py
elliottslaughter: It's great that we can remove the explicit `stable`, thanks!

PR #513, opened 2025-07-07
[Draft] ENH: manifold: Add package for working with meshes

moloney:
Adds new package "Manifold" which is a library that is used by a number of software packages for working with manifold meshes, including for example py-trimesh through its Python bindings.

PR #514, opened 2025-07-07
[Draft/WIP] ENH: py-trimesh: add variant for optional dependencies

moloney:
Adds "recommended" variant that will install more optional deps, plus some of the ones that were previously classified as "easy" (as defined by the trimesh developers).

Depends on PR #513 for supporting new optional dep "manifold", PR #527 for supporting "py-glooey", and PR #528 for supporting "py-vhacdx".

### DIFF
--- a/repos/spack_repo/builtin/packages/py_trimesh/package.py
+++ b/repos/spack_repo/builtin/packages/py_trimesh/package.py
@@ -60,4 +60,3 @@ class PyTrimesh(PythonPackage):
             depends_on("py-psutil")
             depends_on("py-vhacdx")
             depends_on("manifold +python")
-
--- a/repos/spack_repo/builtin/packages/py_trimesh/package.py
+++ b/repos/spack_repo/builtin/packages/py_trimesh/package.py
@@ -21,27 +21,43 @@ class PyTrimesh(PythonPackage):
     variant(
         "easy",
         default=False,
-        description="Install soft dependencies and unlock extra functionality",
+        description="Install all 'easy' soft dependencies",
+    )
+    variant(
+        "recommended",
+        default=False,
+        description="Install most 'recommended' (and all easy) soft dependencies",
+        when="+easy",
     )
 
     depends_on("py-setuptools@40.8:", type="build")
+    with default_args(type=("build", "run")):
+        depends_on("py-numpy")
+        with when("+easy"):
+            depends_on("py-chardet")
+            depends_on("py-colorlog")
+            depends_on("py-jsonschema")
+            depends_on("py-lxml")
+            depends_on("py-mapbox-earcut")
+            depends_on("py-msgpack")
+            depends_on("py-networkx")
+            depends_on("pil")
+            depends_on("py-pycollada")
+            depends_on("py-requests")
+            depends_on("py-rtree")
+            depends_on("py-scipy")
+            depends_on("py-setuptools")
+            depends_on("py-shapely")
+            depends_on("py-svgpath")
+            depends_on("py-xxhash")
 
-    depends_on("py-chardet", type=("build", "run"), when="+easy")
-    depends_on("py-colorlog", type=("build", "run"), when="+easy")
-    depends_on("py-jsonschema", type=("build", "run"), when="+easy")
-    depends_on("py-lxml", type=("build", "run"), when="+easy")
-    depends_on("py-mapbox-earcut", type=("build", "run"), when="+easy")
-    depends_on("py-msgpack", type=("build", "run"), when="+easy")
-    depends_on("py-networkx", type=("build", "run"), when="+easy")
-    depends_on("py-numpy", type=("build", "run"))
-    depends_on("pil", type=("build", "run"), when="+easy")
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/py_trimesh/package.py
```diff
@@ -21,27 +21,42 @@ class PyTrimesh(PythonPackage):
     variant(
         "easy",
         default=False,
-        description="Install soft dependencies and unlock extra functionality",
+        description="Install all 'easy' soft dependencies",
```
tldahlgren, line 24: Suggest changing `soft` -> `software` or dropping it entirely.

repos/spack_repo/builtin/packages/py_trimesh/package.py
```diff
@@ -21,27 +21,42 @@ class PyTrimesh(PythonPackage):
...
+    )
+    variant(
+        "recommended",
+        default=False,
+        description="Install most 'recommended' (and all easy) soft dependencies",
```
tldahlgren, line 29: Suggest changing `soft` -> `software` or dropping it entirely.

repos/spack_repo/builtin/packages/py_trimesh/package.py
```diff
@@ -21,27 +21,42 @@ class PyTrimesh(PythonPackage):
...
+            depends_on("py-networkx")
+            depends_on("pil")
+            depends_on("py-pycollada")
+            depends_on("py-requests")
+            depends_on("py-rtree")
+            depends_on("py-scipy")
+            depends_on("py-setuptools")
+            depends_on("py-shapely")
+            depends_on("py-svgpath")
+            depends_on("py-xxhash")
```
tldahlgren, line 52: Looks like `py-pyglet@:1` and `py-sympy` are missing here based on `3.17.1`'s `setup.py` (https://github.com/mikedh/trimesh/blob/3.17.1/setup.py):

```
requirements_easy = set([
    'scipy',     # provide convex hulls, fast graph ops, etc
    'networkx',  # provide slow graph ops with a nice API
    'lxml',      # handle XML better and faster than built- in XML
    'pyglet<2',  # render preview windows nicely : note pyglet 2.0 is basically a re-write
    'shapely',   # handle 2D polygons robustly
    'rtree',     # create N-dimension trees for broad-phase queries
    'svg.path',  # handle SVG format path strings
    'sympy',     # do analytical math
    'msgpack',   # serialize into msgpack
    'pillow',    # load images
    'requests',  # do network requests
    'xxhash',    # hash ndarrays faster than built-in MD5/CRC
    'setuptools',  # do setuptools stuff
    'jsonschema',  # validate JSON schemas like GLTF
    'pycollada',   # parse collada/dae/zae files
    'chardet',     # figure out if someone used UTF-16
    'mapbox-earcut',  # fast 2D triangulations of polygons
    'colorlog'])   # log in pretty colors
```

repos/spack_repo/builtin/packages/py_trimesh/package.py
```diff
@@ -21,27 +21,42 @@ class PyTrimesh(PythonPackage):
...
+
+        with when("+recommended"):
```
tldahlgren, line 54: These do not appear to be "recommended" for the listed version(s), though some do appear in the `requirements_all` in https://github.com/mikedh/trimesh/blob/3.17.1/setup.py.

tldahlgren (changes requested):
The changes don't match the options in the listed versions of the package, which are 3-6 years old. 

The only package (`spack dependents py-trimesh`) depending on this one is `py-graphcast`, which has a version that is about 2 years old but doesn't explicitly limit the version of `py-trimesh`.

You might want to consider deprecating the older versions (https://spack.readthedocs.io/en/latest/packaging_guide_creation.html#deprecating-old-versions) and adding newer ones that are more in line with the changes you want to make. Just make sure that the dependencies in this file match those in the software's `pyproject.toml`/`setup.*`.

repos/spack_repo/builtin/packages/py_trimesh/package.py
```diff
@@ -21,27 +21,42 @@ class PyTrimesh(PythonPackage):
...
+            depends_on("py-sympy")
+            depends_on("py-pyglet@:1")
+            depends_on("py-glooey")
+            depends_on("py-meshio")
+            depends_on("py-scikit-image")
+            depends_on("py-psutil")
+            depends_on("py-vhacdx")
+            depends_on("manifold +python")
```
tldahlgren, line 62: These do not appear to be "recommended" for the listed version(s), though some do appear in the `requirements_all` in https://github.com/mikedh/trimesh/blob/3.17.1/setup.py.

```
requirements_all = requirements_easy.union([
    'python-fcl',    # do fast 3D collision queries
    'psutil',        # figure out how much memory we have
    'glooey',        # make GUI applications with 3D stuff
    'meshio',        # load a number of additional mesh formats; Python 3.5+
    'scikit-image',  # marching cubes and other nice stuff
    'xatlas',        # texture unwrapping
])
```

PR #515, opened 2025-07-07
lfortran: add kokkos variant

junghans:


rbberger (commented):
Isn't this missing a change in `cmake_args` that does
```python
self.define_from_variant("WITH_KOKKOS", "kokkos")
```
?

rbberger (dismissed):

rbberger (approved):

PR #516, opened 2025-07-08
[WIP] pbvr,kvs: new packages. Fixes for qt-base and qt-svg.

sakamoto-naohito:
pbvr: New package
kvs: New package
qt-base: Fixed issues with linking OpenGL using qmake; addressed bugs in kernel versions prior to 4.18.0-392.
qt-svg: Added version 6.2.4.


repos/spack_repo/builtin/packages/qt_base/package.py
```diff
@@ -324,6 +333,20 @@ def define_feature(key, variant=None):
         if "~opengl" in spec:
             args.append(self.define("INPUT_opengl", "no"))
 
+        # Workaround for qmake's $$system() not capturing command output correctly.
+        # This issue is caused by a known bug in certain Linux kernel versions.
+        # To avoid it, Qt is built with the '-no-feature-forkfd_pidfd' option to disable
+        # the use of new process management features that rely on pidfd.
+        # Note: Red Hat fixed this kernel issue in version 4.18.0-392 and later.
```
wdconinc: Could you provide some info on which (currently supported) operating systems are shipping these older kernels? Mostly curious, but also so we can add info and a sunset strategy for this part of the package.

repos/spack_repo/builtin/packages/qt_base/package.py
```diff
@@ -258,6 +260,13 @@ class QtBase(QtPackage):
             sha256="e4d9f1aee0566558e77eef5609b63c1fde3f3986bea1b9d5d7930b297f916a5e",
         )
 
+    # Patch to override incorrect OpenGL linkage in FindWrapOpenGL.cmake.
+    # The original CMake script links against OpenGL::GL,
+    # which may resolve to `-lOpenGL -lGLX` on some systems, causing linker errors.
```
wdconinc: I'd argue that this is intended behavior. When the `OpenGL::GL` target resolves to `-lOpenGL -lGLX` then we should use that, because that's the OpenGL installation indicating that it needs both of those. Isn't this a reflection of another issue, with the OpenGL being used? I am concerned that this will end up introducing linking OpenGL errors for other systems.

wdconinc (commented):
`qt-svg` version looks good. Some concerns about the `qt-base` changes.

repos/spack_repo/builtin/packages/qt_base/package.py
sakamoto-naohito: Sorry for the late reply.
This bug occurs on Red Hat Enterprise Linux 8.

The Red Hat Enterprise Linux life cycle is as follows.
End of Full Support                         : 31 May 2024
End of Maintenance Support          : 31 May 2029
End of Extended Lifecycle Support : 31 May 2032
Link:https://access.redhat.com/support/policy/updates/errata

The kernel version is as follows.
Red Hat Enterprise Linux 8.6                  : 4.18.0-372.9.1
The version in which the bug was fixed : 4.18.0-392
Red Hat Enterprise Linux 8.7                  : 4.18.0-425.3.1
Link:https://access.redhat.com/ja/articles/16476

I referred to the following link regarding the bug fix.
LINK:https://forum.qt.io/topic/138533/qt-creator-6-7-8-qt-6-qmake-doesn-t-exist-isn-t-executable-or-has-no-output

Please wait a moment for the other reply.

wdconinc: If this affects only up to RHEL 8.6, which is out of support for more than a year, should we fix this as a bug in spack? Not only is RHEL 8.6 out of support, but also RHEL 8.7, 8.8, and 8.9 are already out of support. I understand the business and stability considerations of sticking to an older operating system release, but isn't it just really time to upgrade?

We remove legacy python support from spack packages according to the python EOL timeline, so I would suggest we should do this similarly for kernel versions in use by major operating system releases. If we add this here, what would be our sunset strategy? How many years out of support should we allow?

repos/spack_repo/builtin/packages/qt_base/package.py
sakamoto-naohito: The modification to the qt-base package was incorrect and should have been made to the pbvr package instead, so the patch to qt-base has been reverted and reapplied to pbvr.

repos/spack_repo/builtin/packages/qt_base/package.py
sakamoto-naohito: I am removing a fix for a Linux kernel error from the qt-base package. 
As I do not manage the system itself, I cannot perform system updates, nor am I aware of the schedule for any future updates. 
Therefore, I will handle this error in a new package, pbvr, to minimize the scope of the impact. 
Once the system update is completed, I will remove the corresponding error handling from the new pbvr package.

repos/spack_repo/builtin/packages/kvs/package.py
```diff
@@ -0,0 +1,72 @@
...
+
+class Kvs(Package):
+    """Kyoto Visualization System (KVS) is a multi-platform, open-source C++ library
+    for developing scientific visualization applications. KVS provides various classes
+    and interfaces, such as isosurace extraction, streamlines and volume rendering,
+    to visualize medical data obtained by computerized tomography or magnetic resonance imaging,
+    numerical data from computational fluid dynamics and so on."""
+
+    homepage = "https://github.com/TO0603/KVS"
+    url = "https://github.com/TO0603/KVS/archive/refs/tags/forDev.tar.gz"
```
wdconinc: We don't typically distribute forks. Can you please justify why we should not use https://github.com/naohisas/KVS here?

repos/spack_repo/builtin/packages/kvs/package.py
```diff
@@ -0,0 +1,72 @@
...
+#
+# SPDX-License-Identifier: (Apache-2.0 OR MIT)
+
+from spack_repo.builtin.build_systems.generic import Package
+
+from spack.package import *
+from spack.util.environment import set_env
+
+
+class Kvs(Package):
```
wdconinc: This package appears to support CMake as a build system with the usual way to set options. Is there a reason to go through the patch trouble to configure this?

sakamoto-naohito: This package does not support CMake.  
The official build system is plain `make`, and the build options are controlled only by editing a `kvs.conf` file.  
In this file, the compiler and compilation flags are specified, including whether to enable or disable certain features through preprocessor options (`-D`).
Therefore, we cannot use the standard CMake arguments mechanism in Spack, and patching the configuration file is required to adjust the build options.

wdconinc: In that case it is more appropriate to use a [MakefilePackage](https://spack.readthedocs.io/en/latest/build_systems/makefilepackage.html).

repos/spack_repo/builtin/packages/kvs/package.py
sakamoto-naohito: We are developing this fork specifically as KVS for PBVR, and it has diverged significantly from the upstream repository. 

The main differences include:
・Support for OpenXR, allowing HMDs to be selected as display devices
・Support for VTK file format I/O
・Support for 3D point cloud data (LAS format) I/O and visualization
・Extended functionality for managing visualization target objects

No plan to merge these changes back upstream; the fork will be maintained independently.
In the upcoming PBVR release (version 3.5.0), this forked KVS will be bundled.

For these reasons, we believe it is appropriate to use this fork rather than the upstream repository.

repos/spack_repo/builtin/packages/kvs/package.py
sakamoto-naohito: I previously stated that this package uses plain make, but in fact, qmake is also invoked during the build process.
Therefore, the build cannot be completed using a standard Makefile alone.
Using MakefilePackage caused issues during installation.
To properly support the KVS build process in Spack, I think it is appropriate to keep using the Package class.

repos/spack_repo/builtin/packages/kvs/package.py
wdconinc: > We are developing this fork specifically as KVS for PBVR, and it has diverged significantly from the upstream repository.

Spack cannot endorse your fork over the original project without permission from the origin project. It's in the [KVS license](https://github.com/naohisas/KVS/blob/develop/LICENSE).

As a packaging solution, we cannot ignore what users expect to get when they install a package by the name of `kvs`, and we have to be careful to respect the intellectual property of authors of active projects.

This is not the situation of an unmaintained original project where you take over maintainership: there are more recent commits in the original project than in your fork.

You should package KVS with the original project location as source URL, and make abundantly clear that your version 3.5 is not the same as a potential future version 3.5 in the original project.

wdconinc: After some more discussion with other spack maintainers, here are some options for how to handle this:
- define a specific `pbvr` version or prefixed version series that tracks your fork, similar to the `kitware` version in the `ninja` package, or the `compat-*` series in the `mgard` package, https://github.com/spack/spack-packages/blob/01d5f5bc9fadc15c9a8e023dfa128daea42d8159/repos/spack_repo/builtin/packages/ninja/package.py#L40-L45
https://github.com/spack/spack-packages/blob/c71f8e48245012565ac7b0648dfda137b0071de7/repos/spack_repo/builtin/packages/mgard/package.py#L36-L41
- define the change set between original and fork as a set of patches that `pbvr` applies to `kvs`, similar to how `julia` requires specific patches on `llvm`, https://github.com/spack/spack-packages/blob/c71f8e48245012565ac7b0648dfda137b0071de7/repos/spack_repo/builtin/packages/julia/package.py#L224-L231

The main thing that needs to be ensured is that `spack install kvs` installs the original project by default, not your fork.

If there are other approaches that come up, I'll post them here too.

sakamoto-naohito: After consideration, we will proceed as follows:

- We will not create a kvs package.
- We will change the pbvr package to version 3.5.0 and bundle kvs with it.

pbvr version 3.5.0 has already been released. Please wait while we update the pbvr package.

sakamoto-naohito: I removed the kvs package and updated the pbvr package from version 3.4.0 to 3.5.0, bundling kvs within it.

repos/spack_repo/builtin/packages/pbvr/package.py
```diff
@@ -0,0 +1,148 @@
...
+
+    patch("kvs-conf.patch", when="~client~extended_fileformat")
+    patch("kvs-extended-fileformat-conf.patch", when="~client+extended_fileformat")
+    patch("kvs-client-conf.patch", when="+client~extended_fileformat")
+    patch("kvs-client-extended-fileformat-conf.patch", when="+client+extended_fileformat")
+
+    patch("pbvr-conf.patch", when="~mpi")
+    patch("pbvr-conf-mpi.patch", when="+mpi")
+    patch("makefile-machine-gcc-omp.patch", when="~mpi")
+    patch("makefile-machine-gcc-mpi-omp.patch", when="+mpi")
```
wdconinc: Have you considered avoiding these patches (which are fragile) and instead implementing a `patch` method that does filter_file operations to make the individual changes?
```py
    def patch(self):
        filter_file(
            "KVS_ENABLE_GLEW=.*$",
            "KVS_ENABLE_GLEW=0",
            "KVS/kvs.conf"
        )
        filter_file(
            "KVS_SUPPORT_OPENXR=.*$",
            "KVS_SUPPORT_OPENXR=0",
            "KVS/kvs.conf"
        )
        if self.satisfies("+client"):
            filter_file(
                "KVS_SUPPORT_QT=.*$",
                "KVS_SUPPORT_QT=1",
                "KVS/kvs.conf"
            )
        if self.satisfies("+extended_fileformat"):
            filter_file(
                "KVS_SUPPORT_EXTENDED_FILE_FORMAT=.*$",
                "KVS_SUPPORT_EXTENDED_FILE_FORMAT=1",
                "KVS/kvs.conf"
            )
```
This might be more composable and maintainable when there are upstream changes to the configuration files. Patches like this can often turn into headaches.

repos/spack_repo/builtin/packages/pbvr/package.py
```diff
@@ -0,0 +1,148 @@
...
+
+    def setup_build_environment(self, env):
+        compiler_name = self.spec.compiler.name
+        if compiler_name != "gcc":
+            raise InstallError(f"{self.name} is only supported with the GCC compiler.")
```
wdconinc: This is better handled as a package requirement so the error is emitted on concretization, not building.
```suggestion
    requires("%cxx=gcc")
```

repos/spack_repo/builtin/packages/pbvr/package.py
```diff
@@ -0,0 +1,148 @@
...
+                    "https://github.com/CCSEPBVR/CS-IS-PBVR/wiki/BuildforLinux_JP"
+                )
+
+        with set_env(
+            KVS_CPP="g++",
+            SPACK_KVS_DIR=prefix,
+            VTK_VERSION="9.3",
+            VTK_INCLUDE_PATH=str(spec["vtk"].prefix.include) + "/vtk-9.3",
+            VTK_LIB_PATH=str(spec["vtk"].prefix.lib),
+        ):
```
wdconinc: This must be done in `setup_build_environment`.

repos/spack_repo/builtin/packages/pbvr/package.py
```diff
@@ -0,0 +1,148 @@
...
+        for root, dirs, files in os.walk(source_dir):
+            for fname in files:
+                path = os.path.join(root, fname)
+                try:
+                    with open(path, "r", encoding="utf-8") as f:
+                        content = f.read()
+                    if "KVS_DIR" in content:
+                        filter_file("KVS_DIR", "SPACK_KVS_DIR", path)
+                except Exception:
+                    pass
```
wdconinc: This is potentially expensive and not canonical. At least we need a line of comment/documentation on why we cannot just use the regular `KVS_DIR` variable.

repos/spack_repo/builtin/packages/pbvr/package.py
```diff
@@ -0,0 +1,148 @@
...
+    def build(self, spec, prefix):
+        # Workaround for qmake's $$system() not capturing command output correctly.
+        # This issue is caused by a known bug in certain Linux kernel versions.
+        # To avoid it, Qt is built with the '-no-feature-forkfd_pidfd' option to disable
+        # the use of new process management features that rely on pidfd.
+        # Note: Red Hat fixed this kernel issue in version 4.18.0-392 and later.
+        is_rhel8_bug_fixed = False
+        release_str = platform.uname().release
+        match = re.match(r"^(\d+\.\d+\.\d+)-([\d\.]+)\.el8", release_str)
+        if match and not is_rhel8_bug_fixed:
```
wdconinc: `is_rhel8_bug_fixed` is always false in this conditional.

repos/spack_repo/builtin/packages/pbvr/package.py
```diff
@@ -0,0 +1,148 @@
...
+
+        if "+client" in spec:
+            install("Client/build/App/pbvr_client", prefix.bin)
+            src = self.stage.source_path
+            install_tree(
+                os.path.join(src, "Client/build/App/Shader"), os.path.join(prefix.bin, "Shader")
+            )
+            install_tree(
+                os.path.join(src, "Client/build/App/Font"), os.path.join(prefix.bin, "Font")
+            )
```
wdconinc, line 182: Prefer using `join_path` inside the spack DSL.
```suggestion
            install_tree(
                join_path(src, "Client/build/App/Shader"), join_path(prefix.bin, "Shader")
            )
            install_tree(
                join_path(src, "Client/build/App/Font"), join_path(prefix.bin, "Font")
            )
```

repos/spack_repo/builtin/packages/pbvr/package.py
```diff
@@ -0,0 +1,148 @@
...
+            if "+client" in spec:
+                qmake = Executable(spec["qt-base"].prefix.bin.qmake)
+                build_dir = join_path(self.stage.source_path, "Client/build")
+                os.makedirs(build_dir)
+                with working_dir(build_dir):
+                    qmake("../pbvr_client.pro")
+                    make()
+
+            # Build Sevrer
+            make("-C", "CS_server")
```
wdconinc: For consistency and to avoid the dependency on the present working directory:
```suggestion
            with working_dir(join_path(self.stage.source_path, "CS_server")):
                make()
```

repos/spack_repo/builtin/packages/pbvr/package.py
```diff
@@ -0,0 +1,148 @@
...
+    for Computational Science and e-Systems of the Japan Atomic Energy Agency."""
+
+    phases = ["build", "install"]
+
+    homepage = "https://github.com/CCSEPBVR/CS-IS-PBVR"
+    url = "https://github.com/CCSEPBVR/CS-IS-PBVR/archive/refs/tags/v3.5.0.tar.gz"
+
+    maintainers("sakamoto-naohito")
+
+    license("LGPL-3.0-only")
```
wdconinc: Since this now vendors KVS, the license should reflect this:
```suggestion
    license("LGPL-3.0-only AND BSD-3-Clause")
```

repos/spack_repo/builtin/packages/pbvr/package.py
sakamoto-naohito: I've implemented the suggested patch method.

repos/spack_repo/builtin/packages/pbvr/package.py
sakamoto-naohito: I've updated the code to set the environment variables within setup_build_environment as you suggested.

repos/spack_repo/builtin/packages/pbvr/package.py
sakamoto-naohito: I removed the variable is_rhel8_bug_fixed.

repos/spack_repo/builtin/packages/pbvr/package.py
sakamoto-naohito: I've modified the filter_file function to allow specifying the target file. I've added a comment explaining the reason for changing KVS_DIR to SPACK_KVS_DIR within the filter_file function.

sakamoto-naohito:
It's been about a month since I pushed the fix.
Are there any other required changes, or is this considered complete?

bernhardkaindl:
@sakamoto-naohito, I'm sorry that this has been automatically closed after you updated the PR and asked for review.

Please comment if you'd be interested in finishing this PR to get it merged, thanks!

bernhardkaindl:
Hi @sakamoto-naohito, many fixes went into the Qt packages since you opened this PR, so it should be easier now.

However, do you plan to resume / revive this PR, or shall we close it again?

sakamoto-naohito:
Thank you for following up. I no longer need this PR to be merged, so please feel free to close it.

PR #518, opened 2025-07-08
py_contourpy: Adjust py-setuptools dependency

JakobKai:
This dependency is not defined in contourpy itself. But starting from version 1.0.7 the [project] section is used in pyproject.toml. This requires `py-setuptools@61:`.

### DIFF
--- a/repos/spack_repo/builtin/packages/py_contourpy/package.py
+++ b/repos/spack_repo/builtin/packages/py_contourpy/package.py
@@ -31,6 +31,8 @@ class PyContourpy(PythonPackage):
 
         # Historical dependencies
         depends_on("py-setuptools@42:", when="@:1.0")
+        # not in original requirements, but pyproject.toml [project] requires py-setuptools@61:
+        depends_on("py-setuptools@61:", when="@1.0.7:1.0")
         depends_on("py-build", when="@:1.0.5")
 
     with default_args(type=("build", "link", "run")):


repos/spack_repo/builtin/packages/py_contourpy/package.py
```diff
@@ -31,6 +31,8 @@ class PyContourpy(PythonPackage):
 
         # Historical dependencies
         depends_on("py-setuptools@42:", when="@:1.0")
+        # not in original requirements, but pyproject.toml [project] requires py-setuptools@61:
+        depends_on("py-setuptools@61:", when="@1.0.7:1.0")
```
manuelakuhn, line 35: This restriction is not needed, `py-contourpy` builds fine with `py-setuptools` < 61 because a `setup.py` is still provided with versions `1.0`.
E.g. for me `py-contourpy@1.0.7` builds fine with `py-setuptools@59`

JakobKai: Spack does not report an error during the installation and a simple import (`import contourpy`) works, but setuptools@:60 seems to be unable to extract the correct package name when the metadata is in `pyproject.toml` and the installation is broken.

I tested to configurations and have the following differences:

## `spack install py-contourpy@1.0.7 ^py-setuptools@61:`
- uses `py-setuptools@80.9.0`
- content of `lib/python3.11/site-packages/`:  `contourpy` and  `contourpy-1.0.7.dist-info`
- `from contourpy import contour_generator` works

## `spack install py-contourpy@1.0.7 ^py-setuptools@:60`
- uses `py-setuptools@59.4.0`
- content of `lib/python3.11/site-packages/`: `contourpy` and  `UNKNOWN-1.0.7.dist-info`
- `from contourpy import contour_generator` fails

manuelakuhn: Ah ok, I see what you mean. I only tested `python -c "import contourpy"` which worked perfectly fine without error with the old `py-setuptools` version.
I thought `import contourpy` would load everything, but there seems to be a difference to `from contourpy import contour_generator`...

Thanks for catching that!

manuelakuhn (approved):

repos/spack_repo/builtin/packages/py_contourpy/package.py
JakobKai: Thank you for checking and validating this change :)

PR #525, opened 2025-07-08
Boost: fixup Windows build

johnwparent:
* Refactor compiler attributes access to use new compiler interface
* Update handling of new and deprecated boost modules

### DIFF
--- a/repos/spack_repo/builtin/packages/boost/package.py
+++ b/repos/spack_repo/builtin/packages/boost/package.py
@@ -145,6 +145,7 @@ class Boost(Package):
         "regex",
         "serialization",
         "signals",
+        "signals2",
         "stacktrace",
         "system",
         "test",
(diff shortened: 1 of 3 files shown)


repos/spack_repo/builtin/packages/boost/package.py
```diff
@@ -320,7 +327,10 @@ def libs(self):
     # On Windows, the signals variant is required when building any of
     # the all_libs variants.
     for lib in all_libs:
-        requires("+signals", when=f"+{lib} platform=windows")
+        if lib != "signals2":
+            requires("+signals", when=f"@:1.68 +{lib} platform=windows")
+        if lib != "signals":
+            requires("+signals2", when=f"@1.69: +{lib} platform=windows")
```
scheibelp, line 333: (refactor request)

```
if lib not in ["signals", "signals2"]:
    # <= 1.68 needs signals, after that needs signals2
    requires("+signals", when=f"@:1.68 +{lib} platform=windows")
    requires("+signals2", when=f"@1.69: +{lib} platform=windows")
```

I think this would be clearer, and would avoid the "junk" constraint `requires("+signals", when=f"@:1.68 +signals platform=windows")`

scheibelp (changes requested):
one refactoring request

repos/spack_repo/builtin/packages/boost/package.py
johnwparent: Good idea! Addressed and pushed

scheibelp (approved):

scheibelp (approved):

PR #527, opened 2025-07-08
[Draft/WIP] ENH: py-glooey: new package

moloney:
Adds new spec for python "glooey" package.

### DIFF
--- a/repos/spack_repo/builtin/packages/py_glooey/package.py
+++ b/repos/spack_repo/builtin/packages/py_glooey/package.py
@@ -2,24 +2,8 @@
 #
 # SPDX-License-Identifier: (Apache-2.0 OR MIT)
 
-# ----------------------------------------------------------------------------
-# If you submit this package back to Spack as a pull request,
-# please first remove this boilerplate and all FIXME comments.
-#
-# This is a template package file for Spack.  We've put "FIXME"
-# next to all the things you'll want to change. Once you've handled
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/py_glooey/package.py
```diff
@@ -0,0 +1,27 @@
...
+    license("MIT", checked_by="moloney")
+
+    version("0.3.6", sha256="c025aca6c0e8d73f99793cb7b15aa547195da5499a11384379f2d3446f864520")
+
+    depends_on("python@3.6:", type=("build", "run"))
+    depends_on("py-flit", type="build")
+    with default_args(type=("build", "run")):
+        depends_on("py-pyglet")
+        depends_on("py-more-itertools")
+        depends_on("py-pyyaml")
```
tldahlgren, line 27: You're missing a couple of dependencies according to the `pyproject.toml` (https://github.com/kxgames/glooey/blob/v0.3.6/pyproject.toml):

```
requires = [
  'pyglet',
  'more_itertools',
  'vecrec',    # need to add this package
  'autoprop',  # need to add this package
  'pyyaml',
]
```

tldahlgren (changes requested):
Confirmed the homepage exists and the sha256 is correct; however, there are missing dependencies that require the packages to be added (see comments).

PR #529, opened 2025-07-09
petsc: depends_on mumps+int64 when +mumps+int64

Briffou:
petsc+mumps+int64 was broken I think when with +metis+openmp for instance


Briffou:
Humm, just saw in petsc code:

```
#error "PETSc has not been tested with full 64-bit MUMPS and we choose to error out"
```

so I think `~int64` was deliberate.

PR #537, opened 2025-07-09
sirius require cmake>=3.25 

simonpintarelli:
sirus is using `CUDA::nvtx3` which has been introduced in cmake 3.25.

### DIFF
--- a/repos/spack_repo/builtin/packages/sirius/package.py
+++ b/repos/spack_repo/builtin/packages/sirius/package.py
@@ -80,7 +80,7 @@ class Sirius(CMakePackage, CudaPackage, ROCmPackage):
     depends_on("c", type="build")
     depends_on("fortran", type="build")
 
-    depends_on("cmake@3.23:", type="build")
+    depends_on("cmake@3.25:", type="build")
     depends_on("mpi")
     depends_on("gsl")
     depends_on("blas")


repos/spack_repo/builtin/packages/sirius/package.py
```diff
@@ -80,7 +80,7 @@ class Sirius(CMakePackage, CudaPackage, ROCmPackage):
     depends_on("c", type="build")
     depends_on("fortran", type="build")
 
-    depends_on("cmake@3.23:", type="build")
+    depends_on("cmake@3.25:", type="build")
```
msimberg, line 83: Is this sirius-version dependent? I see sirius still links to `nvToolsExt` instead of `nvtx3` as I would expect: https://github.com/electronic-structure/SIRIUS/blob/dfb1f492205a0541f574c76dd17e15381aeef2e7/src/CMakeLists.txt#L126. Or is that another cmake target?

simonpintarelli: thanks @msimberg and sorry for the late reply. The PR to link against nvtx3 isn't yet merged.

simonpintarelli: https://cmake.org/cmake/help/v3.25/module/FindCUDAToolkit.html?highlight=nvtx#nvtoolsext 
the old nvToolsExt target still exists, but is deprecated. I think it should still work for older sirius versions, so no need to make a distinction.

msimberg (approved):

PR #541, opened 2025-07-09
Python: overhaul windows installer

johnwparent:
This PR

* Adds the installation of PDB files that were previously neglected. These are installed by the official Windows Python Installers, so Spack should as well
* Python 3.13+ add two "new" binaries, `venvlauncher.exe` and `venvwlauncher.exe`, which are just renamed `python.exe` and `pythonw.exe` respectively, for use in venv creation. Update the installer to install these into the venv stdlib module. Additionally install `python.exe` and `pythonw.exe` into the venv stdlib module. Not sure how venv's were ever working on Windows, I think we just never actually tested installing anything into them.
* Refactors installer to use pathlib
* Reorganizes installer for a more logical approach to installing Python
* Adds a Python package to Gitlab Ci for a more stable Python+ ecosystem on Windows

### DIFF
--- a/repos/spack_repo/builtin/packages/python/package.py
+++ b/repos/spack_repo/builtin/packages/python/package.py
@@ -660,20 +660,21 @@ class Python(Package):
             # are created
             venv_binaries = ("venvlauncher.exe", "venvwlauncher.exe")
         for binary in venv_binaries:
-                copy(str(build_root / binary), prefix.Lib.venv.scripts.nt)
+            copy(str(build_root / binary), prefix.Lib.venv.scripts.nt)
 
         # handle shared libraries
-        shared_libraries:List[str] = []
-        shared_libraries.extend(glob.glob(f"{str(build_root)}\\*.dll"))
+        shared_libraries = glob.glob(f"{str(build_root)}\\*.dll")
         shared_libraries.extend(glob.glob(f"{str(build_root)}\\*.pyd"))
         os.makedirs(prefix.DLLs)
         for lib in shared_libraries:
+            libname = os.path.basename(lib)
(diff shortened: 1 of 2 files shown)


johnwparent:
@spackbot fix style

repos/spack_repo/builtin/packages/python/package.py
```diff
@@ -614,41 +615,76 @@ def win_installer(self, prefix):
...
+        copy_tree(str(doc_dir), prefix.Doc)
+        # install tools
+        tools_dir = proj_root / "Tools"
+        copy_tree(str(tools_dir), prefix.Tools)
+        # install stdlib python modules
+        lib_dir = proj_root / "Lib"
+        copy_tree(str(lib_dir), prefix.Lib)
+
+        # locate and track all pdb files
+        pdbs = glob.glob(f"{str(build_root)}\\*.pdb")
```
scheibelp, line 637: (question) It occurs to me if we are copying `.pdb` files (which seems like a good idea) then we may need to communicate to the python install where the source is going to end up (which I think `pathmap` can be used for w/ `cl`).

That doesn't have to be done in this PR, but I wanted to confirm that you agree this would be an issue (or if I'm missing somewhere that is already happening).

scheibelp (dismissed):
I had a question (but overall approve)

scheibelp (dismissed):

johnwparent:
Pipeline is now green on Gitlab, modifying Github side PR to trigger sync script

johnwparent:
@scheibelp green ci! LGTM before we get another conflict 😁 (that last conflict was my fault anyway lol)

scheibelp (approved):

PR #544, opened 2025-07-10
remove spack.platforms.cray.slingshot_network

haampie:
Spack packages definitions should not depend on the state of the
filesystem.

Instead of branching on `spack.platforms.cray.slingshot_network`, use a
variant `+slingshot` which users can prefer in config for particular
systems.

Further, this is private Spack API.


alalazo (approved):

alalazo:
Lot of shepherding to get pipelines :heavy_check_mark: but now it's done. Hopefully the status here will change too :crossed_fingers:

PR #545, opened 2025-07-10
builtin: do not use spack.tengine

haampie:
tengine is private Spack API, use f-strings instead.


alalazo (approved):

alalazo:
Works fine for me:
Screenshot from 2025-07-10 13-39-50

PR #546, opened 2025-07-10
Always run all style checks

haampie:
* Run all style checks also when any of them fails.
* Print which check is running
* Colors
* `--fix`

Replaces #543.

### DIFF
--- a/.ci/style_check.sh
+++ b/.ci/style_check.sh
@@ -1,12 +1,34 @@
 #!/bin/sh -e
-ref="${1:-develop}"
-if [ ! -d "spack-core" ]; then
-  printf "No 'spack-core' dir found, should be a symlink or clone of 'spack/spack'\n" >&2
-  exit 1
-fi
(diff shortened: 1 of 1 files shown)


.ci/style_check.sh
```diff
@@ -1,12 +1,34 @@
...
+      exit 0
+      ;;
+    --fix) flags= ;;
+    -*) die "unknown option: $1" ;;
+    *) ref="$1" ;;
+  esac
+  shift
+done
+[ -d "spack-core" ] ||  die "no 'spack-core' dir found: should be a clone of 'spack/spack'"
+if ! python_files > /dev/null; then
```
kwryankrattiger, line 21: if not python files changed we can also skip the `spack-core` check.

haampie: it's just one stat call, but ok

kwryankrattiger: Not so much to avoid the stat but to not error if there is nothing to check.

PR #547, opened 2025-07-10
libint: control/introduced optional variants

hfp:
- Improved control of optional components.
- Enabled FMA by default (to be leveraged).
- Optimized recipe for performance.
- Modernized code (format strings).
- New release(s) of LIBINT.

### DIFF
--- a/repos/spack_repo/builtin/packages/libint/package.py
+++ b/repos/spack_repo/builtin/packages/libint/package.py
@@ -5,21 +5,12 @@
 import os
 
 from spack_repo.builtin.build_systems.autotools import AutotoolsPackage
-from spack_repo.builtin.packages.boost.package import Boost
 
 from spack.package import *
 
-TUNE_VARIANTS = (
-    "none",
-    "cp2k-lmax-4",
-    "cp2k-lmax-5",
-    "cp2k-lmax-6",
-    "cp2k-lmax-7",
-    "molgw-lmax-4",
-    "molgw-lmax-5",
-    "molgw-lmax-6",
-    "molgw-lmax-7",
-)
+TUNE_VARIANTS_CP2K = ("cp2k-lmax-4", "cp2k-lmax-5", "cp2k-lmax-6", "cp2k-lmax-7")
+TUNE_VARIANTS_MOLGW = ("molgw-lmax-4", "molgw-lmax-5", "molgw-lmax-6", "molgw-lmax-7")
+TUNE_VARIANTS = tuple(["none"]) + TUNE_VARIANTS_CP2K + TUNE_VARIANTS_MOLGW
 
 
 class Libint(AutotoolsPackage):
@@ -28,12 +19,13 @@ class Libint(AutotoolsPackage):
     """
 
     homepage = "https://github.com/evaleev/libint"
-    url = "https://github.com/evaleev/libint/archive/v2.1.0.tar.gz"
+    url = "https://github.com/evaleev/libint/archive/v2.11.1.tar.gz"
 
-    maintainers("dev-zero")
+    maintainers("dev-zero", "hfp")
 
     license("LGPL-3.0-only")
 
+    version("2.11.1", sha256="58ab0f893d94cbed3ab35a6c26ec5e4d8541c59889407a6d30c50b8ea415bdf3")
     version("2.9.0", sha256="4929b2f2d3e53479270be052e366e8c70fa154a7f309e5c2c23b7d394159687d")
     version("2.6.0", sha256="4ae47e8f0b5632c3d2a956469a7920896708e9f0e396ec10071b8181e4c8d9fa")
     version("2.4.2", sha256="86dff38065e69a3a51d15cfdc638f766044cb87e5c6682d960c14f9847e2eac3")
@@ -44,8 +36,9 @@ class Libint(AutotoolsPackage):
     version("1.1.6", sha256="f201b0c621df678cfe8bdf3990796b8976ff194aba357ae398f2f29b0e2985a6")
     version("1.1.5", sha256="ec8cd4a4ba1e1a98230165210c293632372f0e573acd878ed62e5ec6f8b6174b")
 
-    variant("debug", default=False, description="Enable building with debug symbols")
-    variant("fortran", default=False, description="Build & install Fortran bindings")
+    variant("generic", default=False, description="Avoid specialization and larger code size")
+    variant("shared", default=True, description="Build shared library")
+    variant("debug", default=False, description="Enable debug symbols")
     variant(
         "tune",
         default="none",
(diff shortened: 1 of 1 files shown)


hfp:
Configuring with `--enable-generic-code` vs not providing this option is about ~40% code size increase (on x86-64), which is in an acceptable range vs leaving performance on the table. Also, enabling FMA on CPUs is not about "risk" but it is fully covered by LIBINT's tests ("GPU" is a non-topic if "FMA" is left on the table).

alalazo:
@dev-zero

hfp:
> @dev-zero

It might be Tiziano is busy with little time to review (also [elsewhere](https://github.com/cp2k/dbcsr/pull/931)). Perhaps @RMeli can help with the review and approval?

On a side-note, I intend to help maintaining this package going forward hence the maintainer list was updated.

repos/spack_repo/builtin/packages/libint/package.py
```diff
@@ -44,46 +36,51 @@ class Libint(AutotoolsPackage):
...
+    variant("fortran", default=False, description=description_fortran)
+    for tune in TUNE_VARIANTS_CP2K:
+        variant("fortran", default=True, description=description_fortran, when=f"tune={tune}")
 
     # Build dependencies
-    depends_on("c", type="build")  # generated
+    depends_on("fortran", type="build", when="+fortran")
     depends_on("cxx", type="build")  # generated
-    depends_on("fortran", type="build")  # generated
+    depends_on("c", type="build")  # generated
```
RMeli:
```suggestion
    depends_on("c", type="build")
```

repos/spack_repo/builtin/packages/libint/package.py
```diff
@@ -44,46 +36,51 @@ class Libint(AutotoolsPackage):
...
-    variant(
-        "fma",
-        default=False,
-        description=(
-            "Generate code utilizing FMA (requires capable CPU and recent enough compiler)"
-        ),
-    )
+    variant("fma", default=True, description="Generate code utilizing FMA")
+    variant("cxx", default=False, description="Build and install the C++ API")
+    variant("eigen", default=False, description="Eigen used by implementation", when="+cxx")
```
RMeli: Looks like Eigen is a required dependency with `+cxx`. `spack install libint@2.11.1+cxx+fortran tune=cp2k-lmax-6` fails:
```
==> Installing libint-2.11.1-itdfatboff55cogmv7m7lskxyj5huexa [37/37]
==> Fetching https://github.com/evaleev/libint/archive/v2.11.1.tar.gz
    [100%]   21.96 MB @    8.1 MB/s
==> No patches needed for libint
==> libint: Executing phase: 'autoreconf'
==> libint: Executing phase: 'configure'
==> libint: Executing phase: 'build'
    'cmake' '..' '-DCMAKE_INSTALL_PREFIX=/home/rmeli/spack/opt/spack/linux-zen4/libint-2.11.1-itdfatboff55cogmv7m7lskxyj5huexa' '-DLIBINT2_BUILD_SHARED_AND_STATIC_LIBS=ON' '-DENABLE_FORTRAN=ON'

1 error found in build log:
     8102      Add the installation prefix of "Eigen3" to CMAKE_PREFIX_PATH or set
     8103      "Eigen3_DIR" to a directory containing one of the above files.  If "Eigen3"
     8104      provides a separate development package or SDK, be sure it has been
     8105      installed.
     8106    
     8107    
  >> 8108    CMake Error at CMakeLists.txt:120 (message):
     8109      C++ API cannot be built without Eigen3; configure (via CMake) and install
     8110      Eigen3 and add the install prefix to CMAKE_PREFIX_PATH, or add
     8111      -DREQUIRE_CXX_API=OFF to the CMake command line if the C++ API is not
     8112      required
     8113    
     8114    
```

repos/spack_repo/builtin/packages/libint/package.py
```diff
@@ -44,46 +36,51 @@ class Libint(AutotoolsPackage):
...
-    # Libint 2 dependencies
-    # Fixme: Can maintainers please confirm that this is a required dependency
-    depends_on(Boost.with_default_variants, when="@2:")
+    # Libint2 dependencies
+    # Boost is only needed if the Fortran interface is built
+    # A header-only installation (with no specific libraries)
+    # is sufficient (Boost.with_default_variants not needed)
+    depends_on("boost", when="@2: +fortran")
+    # Eigen is optional and not strictly necessary
+    depends_on("eigen", when="@2.7.0: +eigen")
```
RMeli:
```suggestion
    depends_on("eigen", when="@2.7.0: +cxx")
```
? See comment above.

repos/spack_repo/builtin/packages/libint/package.py
RMeli: `libint@2.11.1+cxx+fortran tune=cp2k-lmax-6 +eigen` worked.

RMeli (changes requested):

repos/spack_repo/builtin/packages/libint/package.py
hfp: I removed these comments. Thank you!

repos/spack_repo/builtin/packages/libint/package.py
hfp: > Looks like Eigen is a required dependency with `+cxx`. ``

... and in turn `+cxx` is needed if the Fortran interface is requested (`+fortran`). Older LIBINT built the Fortran interface right away, whereas newer versions require the full two-phase build (which depends on `+cxx`, etc).

repos/spack_repo/builtin/packages/libint/package.py
hfp: Thank you, it's taken into the next (forced) commit.

repos/spack_repo/builtin/packages/libint/package.py
hfp: I am currently testing the suggested changes locally and specifically the case you pointed out.

hfp: The change just pushed includes all of your suggestions. Now, `spack install libint@2.11.1+cxx+fortran tune=cp2k-lmax-6` passes. Thank you!

hfp:
Should be ready now.

repos/spack_repo/builtin/packages/libint/package.py
RMeli:
```suggestion
```
The variant is now unused, isn't it? It should be removed, I think.

RMeli: I can confirm that I was able to build the same spec locally.

hfp: > The variant is now unused, isn't it? It should be removed, I think.

Sure, I can do this. However, libint also treats Eigen as a feature. For this reason, I kept the option. Honestly, I do not know how and when it is used other than as "vectorization helper" or "for convenience".

I will remove it anyway as we should not broadcast features we know nothing about ;-)

RMeli (approved):

RMeli:
Thanks @hfp, LGTM.

hfp:
> Thanks @hfp, LGTM.

Thank you for the review and the good suggestions!

PR #372, opened 2025-06-26
HEP: upgrade gitlab runner image to ubuntu-24.04:v2024-09-05-v2

wdconinc:
This PR upgrades the HEP gitlab runner image from ubuntu22.04-runner-amd64-gcc-11.4:2024.03.01 to ubuntu-24.04:v2024-09-05-v2. The main motivation is a more recent system gcc that supports C++20 more fully.

Needs:
- [x] https://github.com/spack/spack/pull/50094

PR #548, opened 2025-07-10
geomodel: Add version 6.15.0

paulgessinger:


stephenswat (approved):

wdconinc (commented):
Maintainer approved. Looks good to me.

However, it is not possible to rely on CI for this, since only geomodel-6.10 is built there. See https://gitlab.spack.io/spack/spack-packages/-/jobs/17342178#L224. Can we ensure that the latest (newly added) versions of geomodel are built? Is this related to the egl issue? Should we simply disable visualization?

paulgessinger:
@wdconinc How do I go about this? I'm not aware how/where the CI build ranges are configured.

wdconinc:
> @wdconinc How do I go about this? I'm not aware how/where the CI build ranges are configured.

This one is defined at https://github.com/paulgessinger/spack-packages/blob/geomodel/6.15.0/.ci/gitlab/stacks/hep/spack.yaml#L92.

But it seems that it already had visualization disabled. Actually, I think it's due to:
```
    # GeoModel 6.11 requires std::format and C++20
    with when("@6.11:"):
        conflicts("cxxstd=17")
        conflicts("%gcc@:12")
        conflicts("%clang@:16")
```
We appear to have already discussed this in https://github.com/spack/spack-packages/pull/250 and it's waiting for https://github.com/spack/spack-packages/pull/372. So, it's a known issue and I will approve this version addition here.

wdconinc (approved):

PR #552, opened 2025-07-10
builtin: remove the use of `set_env`

alalazo:
See https://github.com/spack/spack/issues/50388

`set_env` is considered internal Spack API, so remove its use from packages

### DIFF
--- a/repos/spack_repo/builtin/packages/chapel/package.py
+++ b/repos/spack_repo/builtin/packages/chapel/package.py
@@ -15,7 +15,7 @@ import llnl.util.lang
 import spack.platforms
 import spack.platforms.cray
 from spack.package import *
-from spack.util.environment import is_system_path, set_env
+from spack.util.environment import is_system_path
 
 
 @llnl.util.lang.memoized
@@ -584,32 +584,24 @@ class Chapel(AutotoolsPackage, CudaPackage, ROCmPackage):
             env.unset(var)
 
     def build(self, spec, prefix):
-        with set_env(CHPL_MAKE_THIRD_PARTY=join_path(self.build_directory, "third-party")):
-            make()
-            with set_env(CHPL_HOME=self.build_directory):
-                if spec.satisfies("+chpldoc"):
-                    make("chpldoc")
-                if spec.satisfies("+python-bindings"):
-                    make("chapel-py-venv")
-                    python("-m", "ensurepip", "--default-pip")
-                    python("-m", "pip", "install", "tools/chapel-py")
+        make()
+        if spec.satisfies("+chpldoc"):
+            make("chpldoc")
+        if spec.satisfies("+python-bindings"):
+            make("chapel-py-venv")
+            python("-m", "ensurepip", "--default-pip")
+            python("-m", "pip", "install", "tools/chapel-py")
 
     def install(self, spec, prefix):
-        # Needed to prevent invalidating cache on subsequent cmake runs due to
-        # env changes, likely pertaining to spack compiler wrappers.
-        with set_env(CHPL_CMAKE_USE_CC_CXX="1"):
-            make("install")
-            # We install CMakeLists.txt so we can later lookup the version number
-            # if working from a non-versioned release/branch (such as main)
-            if not self.is_versioned_release():
-                install("CMakeLists.txt", join_path(prefix.share, "chapel"))
-            install_tree(
-                "doc", join_path(prefix.share, "chapel", self._output_version_short, "doc")
-            )
-            install_tree(
-                "examples",
-                join_path(prefix.share, "chapel", self._output_version_short, "examples"),
-            )
+        make("install")
+        # We install CMakeLists.txt so we can later lookup the version number
+        # if working from a non-versioned release/branch (such as main)
+        if not self.is_versioned_release():
+            install("CMakeLists.txt", join_path(prefix.share, "chapel"))
+        install_tree("doc", join_path(prefix.share, "chapel", self._output_version_short, "doc"))
+        install_tree(
+            "examples", join_path(prefix.share, "chapel", self._output_version_short, "examples")
+        )
 
     def setup_chpl_platform(self, env):
         if self.spec.variants["host_platform"].value == "unset":
@@ -770,6 +762,11 @@ class Chapel(AutotoolsPackage, CudaPackage, ROCmPackage):
     def setup_build_environment(self, env: EnvironmentModifications) -> None:
         self.unset_chpl_env_vars(env)
         self.setup_env_vars(env)
+        env.set("CHPL_HOME", self.build_directory)
+        env.set("CHPL_MAKE_THIRD_PARTY", join_path(self.build_directory, "third-party"))
+        # Needed to prevent invalidating cache on subsequent cmake runs due to
+        # env changes, likely pertaining to spack compiler wrappers.
+        env.set("CHPL_CMAKE_USE_CC_CXX", "1")
 
     def setup_run_environment(self, env: EnvironmentModifications) -> None:
         self.setup_env_vars(env)
@@ -781,6 +778,11 @@ class Chapel(AutotoolsPackage, CudaPackage, ROCmPackage):
         )
         with when("@2.2:2.5"):
             env.set("CHPL_HOME", chpl_home)
+
+        # Needed for standalone tests
+        env.set("CHPL_CHECK_HOME", self.test_suite.current_test_cache_dir)
+        if self.spec.satisfies("+cuda") or self.spec.satisfies("+rocm"):
+            env.set("COMP_FLAGS", "--no-checks --no-compiler-driver")
 
     def get_chpl_version_from_cmakelists(self) -> str:
         cmake_lists = None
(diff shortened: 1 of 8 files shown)


alalazo:
@spackbot help

alalazo:
@spackbot maintainers

alalazo:
I didn't test all of the packages. If there's anything that seems controversial or wrong, for any recipe, let me know and I'll modify asap

ashermancinelli (approved):
TYVM!

G-Ragghianti (approved):

repos/spack_repo/builtin/packages/chapel/package.py
```diff
@@ -770,6 +762,11 @@ def setup_env_vars(self, env):
     def setup_build_environment(self, env: EnvironmentModifications) -> None:
         self.unset_chpl_env_vars(env)
         self.setup_env_vars(env)
+        env.set("CHPL_HOME", self.build_directory)
+        env.set("CHPL_MAKE_THIRD_PARTY", join_path(self.build_directory, "third-party"))
+        # Needed to prevent invalidating cache on subsequent cmake runs due to
+        # env changes, likely pertaining to spack compiler wrappers.
+        env.set("CHPL_CMAKE_USE_CC_CXX", "1")
```
bonachea: IIUC you've taken two environment settings that were previously only set for certain commands in  the build step and one that was only set for the install step, and hoisted all three to be set for all parts of both steps.

This looks at the very least inequivalent to the old behavior, and quite possibly breaks the build. 

@arifthpe you've worked on these changes most recently, can you please comment?

repos/spack_repo/builtin/packages/chapel/package.py
```diff
@@ -782,6 +779,11 @@ def setup_run_environment(self, env: EnvironmentModifications) -> None:
         with when("@2.2:2.5"):
             env.set("CHPL_HOME", chpl_home)
 
+        # Needed for standalone tests
+        env.set("CHPL_CHECK_HOME", self.test_suite.current_test_cache_dir)
+        if self.spec.satisfies("+cuda") or self.spec.satisfies("+rocm"):
+            env.set("COMP_FLAGS", "--no-checks --no-compiler-driver")
```
bonachea: In this case it appears you've hoisted an environment setting out of its surgical location in a stand-alone test where it was apparently needed, to the general run environment which is used for Spack's chapel module (which may or may not be appropriate, I'm not sure).

bonachea (changes requested):
This PR makes non-trivial behavioral changes to the operation of the Chapel spackage, and these changes reportedly have not been thoroughly tested for correctness. Absent any testing, these could easily break the operation of the package.

I don't personally expect to have time to test these proposed changes in the near future, but someone should before this is merged.

mtaillefumier (approved):
No issues on my side for CP2K.

repos/spack_repo/builtin/packages/chapel/package.py
alalazo: > This looks at the very least inequivalent to the old behavior, and quite possibly breaks the build.

The last commit should restore previous behavior exactly. If there are still concerns we'll just wait to merge this until tested.

repos/spack_repo/builtin/packages/chapel/package.py
alalazo: https://github.com/spack/spack-packages/pull/552#discussion_r2199824523

repos/spack_repo/builtin/packages/chapel/package.py
```diff
@@ -865,38 +863,39 @@ def check(self):
...
+        env.update(env_update)
+        return subprocess.run(["util/test/checkChplInstall"], env=env)
 
-    def check_chpl_install(self):
+    def check_chpl_install(self, env_update):
         if self.spec.variants["comm"].value != "none":
-            return self.check_chpl_install_gasnet()
+            return self.check_chpl_install_gasnet(env_update=env_update)
         else:
             return subprocess.run(["util/test/checkChplInstall"])
```
arifthpe: IIUC this call is missing the env it had previously

repos/spack_repo/builtin/packages/chapel/package.py
```diff
@@ -584,32 +584,30 @@ def unset_chpl_env_vars(self, env):
...
-            if not self.is_versioned_release():
-                install("CMakeLists.txt", join_path(prefix.share, "chapel"))
-            install_tree(
-                "doc", join_path(prefix.share, "chapel", self._output_version_short, "doc")
-            )
-            install_tree(
-                "examples",
-                join_path(prefix.share, "chapel", self._output_version_short, "examples"),
-            )
+        make("install", extra_env={"CHPL_CMAKE_USE_CC_CXX": "1"})
```
arifthpe: This `extra_env` should also be applied to the `install` and `install_tree` calls following to preserve previous semantics.

arifthpe (changes requested):
More `chapel` semantic changes

repos/spack_repo/builtin/packages/chapel/package.py
bonachea: I agree with @arifthpe. 

I think this mistake also demonstrates that the suggested replacement idioms for `with set_env(...)` are more error-prone, in addition to being less concise.

Is there a solid technical rationale for removing/disallowing `with set_env(...)`?
This PR seems to demonstrate that removing it just makes the spackages harder to maintain.

tgamblin (changes requested):
@alalazo I rather agree with @bonachea that `set_env` was useful for packagers.  Can we not just put it in `spack.package` and let these packages continue to exist unmodified?  It is useful, IMO, to have a context manager for this, and I don't see why we wouldn't keep this functionality around in core spack so I don't see why we need to remove it from the package API.

What we *don't* want is packages importing from `spack.util.environment` -- we are trying to narrow the package API down to one place so that we can make a simple guarantee for 1.0. I think we should do that with as little disruption to packages as possible.

repos/spack_repo/builtin/packages/chapel/package.py
alalazo: Why do you say so? `install` and `install_tree` are Python functions, where should we apply the environment variable and how does this affect copying?

For instance, this is `install`: 
- https://github.com/spack/spack/blob/570fa283b5787581c6ea4c50ddac7e10c8daa814/lib/spack/llnl/util/filesystem.py#L705-L721

alalazo:
> It is useful, IMO, to have a context manager for this, and I don't see why we wouldn't keep this functionality around in core spack so I don't see why we need to remove it from the package API.

@tgamblin It's used only in a few tests in core Spack, so a good candidate for removal. If we want to keep it in the package API, that's fine - we should just group what should go in `spack.package` from https://github.com/spack/spack/issues/50388 instead of being removed, and do it in one PR.

I might even do it in https://github.com/spack/spack/pull/50976 and rename the PR `Package API v2.2`

haampie:
>  It is useful, IMO, to have a context manager for this, and I don't see why we wouldn't keep this functionality around in core spack so I don't see why we need to remove it from the package API.

`setenv` + `fork` + `execve` + `setenv` instead of simply `fork` + `execve` is an anti-pattern we shouldn't encourage. It's problematic would we ever do parallel loading/setup of packages.

repos/spack_repo/builtin/packages/chapel/package.py
```diff
@@ -584,32 +584,30 @@ def unset_chpl_env_vars(self, env):
...
-                    make("chapel-py-venv")
-                    python("-m", "ensurepip", "--default-pip")
-                    python("-m", "pip", "install", "tools/chapel-py")
+        make(extra_env={"CHPL_MAKE_THIRD_PARTY": join_path(self.build_directory, "third-party")})
+        extra_env = {
+            "CHPL_MAKE_THIRD_PARTY": join_path(self.build_directory, "third-party"),
+            "CHPL_HOME": self.build_directory,
+        }
+        if spec.satisfies("+chpldoc"):
+            make("chpldoc", extra_env=extra_env)
```
haampie: More standard would be `make FOO=BAR`, which would also show up in build logs.

repos/spack_repo/builtin/packages/chapel/package.py
arifthpe: Oh, my bad, I thought those invoke some Make target

bonachea:
> `setenv` + `fork` + `execve` + `setenv` instead of simply `fork` + `execve` is an anti-pattern we shouldn't encourage.

OK, but based on a quick grep of the builtin package repo there are over 170 direct uses of `os.environ`, so hasn't that ship already sailed?

At least `set_env` maintains a principled, composable context for `os.environ` manipulations. It also ensures the modifications don't "leak" outside the lexical scope of `with set_env`, in a manner that is arguably easier to read/audit than explicitly passing around mutable environment objects.

haampie:
> so hasn't that ship already sailed?

No

repos/spack_repo/builtin/packages/chapel/package.py
haampie: It's error-prone because of how this package.py is written.

haampie:
I'll leave `set_env` in the package api :(

bonachea:
It appears this PR has been superseded by #578

tgamblin:
Yep, closing!  If you as package maintainers want to remove `set_env`, feel free to submit a PR after #578.

PR #556, opened 2025-07-10
ncurses: add 6.5-20250705

michaelkuhn:
This fixes building with gcc@15.


manuelakuhn (approved):
Tested on rocky8 with gcc@15 and ubuntu 24.04 with gcc@13

PR #557, opened 2025-07-10
gdbm: fix build with gcc@15

michaelkuhn:


repos/spack_repo/builtin/packages/gdbm/package.py
```diff
@@ -47,6 +47,14 @@ class Gdbm(AutotoolsPackage, GNUMirrorPackage):
     patch("gdbm.patch", when="@:1.18 %oneapi")
     patch("gdbm.patch", when="@:1.18 %arm@21:")
 
+    def flag_handler(self, name, flags):
+        # See https://src.fedoraproject.org/rpms/gdbm/blob/44ea7380c69b1c139fe385bc1c5940070b36c626/f/gdbm.spec#_62
+        if self.spec.satisfies("@:1.24 %gcc@15:"):
+            if name == "cflags":
```
alecbcs:
```suggestion
        if name == "cflags":
            if self.spec.satisfies("@:1.24 %gcc@15:"):
```
I'd flip these two conditional statements to allow for easier additions later on.

alecbcs (changes requested):
Quick suggestion for the flag conditionals. Otherwise looks good to me.

repos/spack_repo/builtin/packages/gdbm/package.py
michaelkuhn: Done

alecbcs (approved):
Looks good to me. Thanks @michaelkuhn!

PR #558, opened 2025-07-11
Add back slingshot network detection.

tgamblin:
Restore functionality lost in #544.

Packages shouldn't *silently* depend on host-based detection, as that discards build provenance. Detected options should be modeled as variants, or at least through specs metadata in some way, as #544 did.

Packages *can* use host detection to set defaults for variants, as that preserves build provenance and allows a user to override the setting if needed. It also makes it easy for users to `spack install` on a given machine and get a sensible default.

`slingshot_network()` as currently written probably shouldn't be in the Spack package API -- it isn't really detecting hardware; it's just the best way we know to test for slinghost on Cray EX. So, not importing `spack.platforms.cray.slingshot_network()` is a good idea.

However, the function is actually needed by users of several packages, so move it *into* the packages for now so as not to disrupt them.

We should have a discussion, eventually, of how best to provide hardware information to packages for configuration hints.  It should likely be something more dynamic than a hard-coded function, so that we don't have to keep functions like `slingshot_network()` around forever. e.g., we might supply a dictionary of detected hardware features to packages, or some string-based query function -- something with a small and sustainable API surface. We can leave that until after 1.0.

### DIFF
--- a/repos/spack_repo/builtin/packages/aluminum/package.py
+++ b/repos/spack_repo/builtin/packages/aluminum/package.py
@@ -16,6 +16,12 @@ from spack_repo.builtin.build_systems.rocm import ROCmPackage
 from spack.package import *
 
 
+def slingshot_network():
+    return os.path.exists("/opt/cray/pe") and (
+        os.path.exists("/lib64/libcxi.so") or os.path.exists("/usr/lib64/libcxi.so")
+    )
+
+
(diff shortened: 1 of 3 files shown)


haampie:
Hm, disagreed. We have `libcxi` as a package, and externals. There's, as always, nothing special about cray libraries.

haampie (approved):

PR #564, opened 2025-07-11
freetype: add zlib variant

hfbauman:
Add variant to enable, disable, or automatically configure zlib. Defaults to "no" to be consistent with the previous hardcoded arguments.

### DIFF
--- a/repos/spack_repo/builtin/packages/freetype/package.py
+++ b/repos/spack_repo/builtin/packages/freetype/package.py
@@ -66,6 +66,13 @@ class Freetype(AutotoolsPackage, CMakePackage):
 
     variant("shared", default=True, description="Build shared libraries")
     variant("pic", default=True, description="Enable position-independent code (PIC)")
+    variant(
+        "zlib",
+        default="no",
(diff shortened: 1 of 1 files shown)


alecbcs:
What is the current behavior of auto? I'd vote that we make the `zlib` a binary variant that adds `zlib-api` as an explicit dependency to the package thus adding it in to the build environment and enabling it with the configure flag.

bernhardkaindl (changes requested):
> What is the current behavior of auto? I'd vote that we make the zlib a binary variant that adds zlib-api as an explicit dependency to the package thus adding it in to the build environment and enabling it with the configure flag.

I also think that auto should not be needed, as yes definitely needs an explicit dependency on `zlib-api`. The PR should also state what are the motivations for this change.

tldahlgren:
@spackbot maintainers

michaelkuhn:
I agree with @alecbcs and @bernhardkaindl, this needs an explicit zlib-api dependency. I would also go for a boolean variant here.

PR #567, opened 2025-07-11
Get a working zsh 5.9 on frontier

PDoakORNL:
On Suse Linux enterprise server 15 SP6 with whatever cray has done to it the zsh 5.9 require patches to build with GCC14: and to fix a hang on any command substitution view "subshell" i.e. `$(...)`

The boolcodes_error.patch is needed to compile at all with gcc@14 back to at least 5.3.1


alecbcs:
Have any of these patches made their way back upstream into zsh? If so it'd be great to reference the upstream merge commit instead of copying the patches into the packages repository.

PDoakORNL:
5.9 is still the newest released version.
https://zsh.sourceforge.io/FAQ/zshfaq01.html#l6

The current git master does work, not sure about the 5.9.0.2-test tarball.
But in light of that I'm going to close this PR and open a new one later with a version from the git repo.

PR #568, opened 2025-07-11
chapel: change maintainers list order

arifthpe:
Change order of `chapel` maintainers list so the current primary maintainers are the first two, to work around [spackbot only requesting review from the first two maintainers](https://github.com/spack/spackbot/issues/109).


arifthpe:
My understanding of the spackbot behavior was incorrect, there's not a limitation on how many maintainers get requested -- see comments on https://github.com/spack/spackbot/issues/109. So this PR is unnecessary

PR #570, opened 2025-07-11
[Draft/WIP] opencascade: Add CMake patch for finding freetype

hfbauman:


wdconinc (commented):
Please add a description with some information on what problem this solves.

There is already a `freetype` variant (default to false), and there is an `adm/cmake/freetype.cmake`, so I don't understand (without additional info) why this patch would need to be added only when `+vtk` and why only to `adm/cmake/vtk.cmake`. If the problem is that `+vtk` should automatically require `+freetype`, then this patch is not the right approach.

PR #571, opened 2025-07-11
[Draft] py-setuptools: Add dependency to py-packaging

kmorel:
The latest version of py-setuptools depends on a recent version of py-packaging. If you fail to load a valid version of packaging, you get an error like the following.

```
  ImportError: Cannot import `packaging.licenses`.
          Setuptools>=77.0.0 requires "packaging>=24.2" to work properly.
          Please make sure you have a suitable version installed.
```

This causes other packages that use pip to install themselves to fail.

### DIFF
--- a/repos/spack_repo/builtin/packages/py_setuptools/package.py
+++ b/repos/spack_repo/builtin/packages/py_setuptools/package.py
@@ -149,6 +149,8 @@ class PySetuptools(Package, PythonExtension):
         # https://github.com/pypa/setuptools/issues/3661
         depends_on("python@:3.11", when="@:67")
 
+        depends_on("py-packaging@24.2:", when="@77:")
+
     depends_on("py-pip", type="build")
 
     conflicts(


RobertMaaskant:
Setuptools vendors its dependencies, so this should not be necessary. Only 80.6.0, 80.7.0 and 80.7.1 did not vendor some dependencies afaik.

Could you explain how you triggered this ImportError? Did you use a version that is not explicitly checksummed?

kmorel:
> Setuptools vendors its dependencies, so this should not be necessary. Only 80.6.0, 80.7.0 and 80.7.1 did not vendor some dependencies afaik.

I'm not sure what that means, but I note that this problem occurred for me with py-setuptools@80.9.0 (I think).

> Could you explain how you triggered this ImportError?

I was specifically attempting to compile ParaView on Frontier. (You can see my configuration and scripts [here](https://github.com/kmorel/frontier-spack-scripts).) There is a long chain of dependencies. I think I ran into this error installing py-pillow.

> Did you use a version that is not explicitly checksummed?

Although I think there is an older version of setuptools configured on the system, I can see that the concretizer installs its own newer version first and uses that when installing the other py packages.

Perhaps there is something installed on Frontier that is fouling up setuptool's automated dependencies, but there is no way for me to remove them. At any rate, making them explicit in spack fixes the problem.

kmorel:
@RobertMaaskant @alecbcs Any further thoughts on this change? This is still a necessary change for me to compile on Frontier.

alecbcs:
@adamjstewart this is a bit our of my wheelhouse. Can you speak to this situation a bit more?

I'm curious if this is a backwards compatibility issue or if we should be modeling the dependencies in setuptools somehow.

adamjstewart:
> Setuptools vendors its dependencies, so this should not be necessary.

I agree with this sentiment. The only other possible issue I can think of is that if packaging is present in `PYTHONPATH`, it could be picked up before the vendored copy of packaging. If that is the case, we can use:
```python
conflicts("py-packaging@:24.1", when="@77:")
```
But we should first understand more about the issue. @kmorel can you add more details about how to reproduce this issue? Maybe first open a bug report for the build error with the build logs and spec you are trying to install.

kmorel:
> But we should first understand more about the issue. @kmorel can you add more details about how to reproduce this issue? Maybe first open a bug report for the build error with the build logs and spec you are trying to install.

I spent several hours replicating the issue with a simpler configuration: #903.

> I agree with this sentiment. The only other possible issue I can think of is that if packaging is present in `PYTHONPATH`, it could be picked up before the vendored copy of packaging. If that is the case, we can use:
> 
> ```python
> conflicts("py-packaging@:24.1", when="@77:")
> ```

This does not work. The problem is that the dependency of py-packaging is happening outside of spack. Since py-packaging is not a dependency to py-setuptools (without my change), spack ignores this command. There is no dependency so there is no conflict.

I agree that the problem is most likely that the wrong version of packaging is present in `PYTHONPATH`. The problem is that necessary system packages are colocated in the same path.

adamjstewart:
I don't think there's much Spack can do about this. Adding a dependency on packaging creates a circular dependency for packaging 21 and older. Using externally-installed Python is known to cause issues, I always recommend building Python with Spack if you plan to actually use it.

kmorel:
I am closing this PR. The suggested change has been uniformly panned as a bad idea. I was able to get around the problem by disabling the configuration of an external `py-pip` package. The spack-installed pip worked with my system python (as opposed to the system pip causing dependency issues with spack).

PR #572, opened 2025-07-11
py-meson-python: Add dependency to ninja

kmorel:
Although meson requires few dependencies, ninja is sometimes one of them. See the following documentation of meson
(at https://mesonbuild.com/Use-of-Python.html).

> The only thing you need is Python 3 (and possibly Ninja).

I had some compile failures because ninja was not available.

### DIFF
--- a/repos/spack_repo/builtin/packages/py_meson_python/package.py
+++ b/repos/spack_repo/builtin/packages/py_meson_python/package.py
@@ -46,6 +46,7 @@ class PyMesonPython(PythonPackage):
     depends_on("py-tomli@1:", when="@0.11: ^python@:3.10", type=("build", "run"))
     depends_on("py-tomli@1:", when="@:0.10", type=("build", "run"))
     depends_on("py-setuptools@60:", when="@0.13 ^python@3.12:", type=("build", "run"))
+    depends_on("ninja", type=("build", "run"))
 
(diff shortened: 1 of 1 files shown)


adamjstewart:
I'll defer to @eli-schwartz @michaelkuhn @rgommers

rgommers (approved):
I'd say this is probably the right thing to do. `ninja` isn't in `build-system.requires` in `pyproject.toml` because it's a system dependency so best already installed and not gotten from PyPI, but it is dynamically injected into the build requirements when a system `ninja` isn't found:

https://github.com/mesonbuild/meson-python/blob/main/mesonpy/__init__.py#L1140-L1151

That will then pull it from PyPI for isolated builds, or fail for non-isolated builds. For Spack one would want to get the Spack-packaged rather than the PyPI-packaged `ninja`, which this PR achieves.

For completeness I checked how conda-forge does it, and there `ninja` is also added as a dependency of `meson-python`:

https://github.com/conda-forge/meson-python-feedstock/blob/4a0599da09d5911739fd46a025b783d995ed09cb/recipe/meta.yaml#L19

adamjstewart (approved):

kmorel:
@adamjstewart @rgommers Thanks for the quick review and merge!

PR #573, opened 2025-07-12
py-libclang: add v16.0.6, v17.0.6, v18.1.1

wdconinc:
This PR adds `py-libclang`, v16.0.6, v17.0.6, v18.1.1 (for the corresponding clang versions). This package is built in CI.


wdconinc:
@adamjstewart @aweits This is a dependency of `py-tensorflow` only.

Built and used correctly for the `py-tensorflow` build in https://gitlab.spack.io/spack/spack-packages/-/pipelines/1155434.

adamjstewart (approved):
Very happy we aren't forced to use an ancient version of llvm anymore!

PR #575, opened 2025-07-13
texinfo: add v7.2

michaelkuhn:
The C11 flag does not seem to be necessary anymore. Also, pass the flag using `flag_handler` instead of `build_targets`.

### DIFF
--- a/repos/spack_repo/builtin/packages/texinfo/package.py
+++ b/repos/spack_repo/builtin/packages/texinfo/package.py
@@ -41,6 +41,8 @@ class Texinfo(AutotoolsPackage, GNUMirrorPackage):
     version("5.1", sha256="50e8067f9758bb2bf175b69600082ac4a27c464cb4bcd48a578edd3127216600")
     version("5.0", sha256="2c579345a39a2a0bb4b8c28533f0b61356504a202da6a25d17d4d866af7f5803")
 
+    variant("xs", default=False, description="Enable Perl XS")
+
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/texinfo/package.py
```diff
@@ -63,12 +66,23 @@ class Texinfo(AutotoolsPackage, GNUMirrorPackage):
...
+        if spec.satisfies("+xs"):
+            config_args.append("--enable-perl-xs")
+        else:
+            config_args.append("--disable-perl-xs")
+
+        return config_args
+
+    def flag_handler(self, name, flags):
+        if self.spec.satisfies("@7.0:7.1"):
+            if name == "cflags":
```
alecbcs:
```suggestion
        if name == "cflags":
            if self.spec.satisfies("@7.0:7.1"):
```

I'd flip the conditionals here to make it easier to add additional flags in the future.

alecbcs (changes requested):
Suggestion to flip the flag handler conditionals. Otherwise looks good to me.

repos/spack_repo/builtin/packages/texinfo/package.py
michaelkuhn: Good point, done.

alecbcs (approved):
Looks good to me. Thanks!

alecbcs:
@spackbot run pipeline

PR #576, opened 2025-07-13
gmp: fix build with gcc@15

michaelkuhn:


manuelakuhn (approved):
Tested on rocky8 with gcc@15 and ubuntu 24.04 with gcc@13

PR #580, opened 2025-07-14
scorep: add version 9.2, continue adding variants

wrwilliams:
* Add release 9.2.
* Add +fortran and use it.
* Add +gotcha for 9.0+ and use it.
* Add +fortran dependency for +mpi_f08.
* Remove redundant enable/disable for +gcc-plugin.

### DIFF
--- a/repos/spack_repo/builtin/packages/scorep/package.py
+++ b/repos/spack_repo/builtin/packages/scorep/package.py
@@ -108,6 +108,11 @@ class Scorep(AutotoolsPackage):
         description="Enable debug info lookup via binutils",
         when="^binutils",
     )
+    variant(
+        "fortran",
+        default=True,
+        description="Enable fortran support",
+    )
     # Putting this in as preparation. F08 support exists in 9.0 but configure does not respect
     # --enable-mpi-f08 and will not until 9.1.
     variant("mpi_f08", default=True, description="Enable MPI F08 support", when="@9.1: +mpi")
@@ -116,10 +121,9 @@ class Scorep(AutotoolsPackage):
     # two components of cube -- cubew and cubelib.
 
     # Language dependencies
-    # TODO: we could allow a +fortran variant here.
     depends_on("c", type="build")  # generated
     depends_on("cxx", type="build")  # generated
-    depends_on("fortran", type="build")  # generated
+    depends_on("fortran", type="build", when="+fortran")  # generated
 
     # SCOREP 9
     depends_on("gotcha@1.0.8:", type="link", when="@9:")
--- a/repos/spack_repo/builtin/packages/scorep/package.py
+++ b/repos/spack_repo/builtin/packages/scorep/package.py
@@ -133,7 +133,7 @@ class Scorep(AutotoolsPackage):
     # two components of cube -- cubew and cubelib.
 
     # Language dependencies
-    depends_on("c", type="build") 
+    depends_on("c", type="build")
     depends_on("cxx", type="build")
     depends_on("fortran", type="build", when="+fortran")
 
--- a/repos/spack_repo/builtin/packages/scorep/package.py
+++ b/repos/spack_repo/builtin/packages/scorep/package.py
@@ -133,9 +133,9 @@ class Scorep(AutotoolsPackage):
     # two components of cube -- cubew and cubelib.
 
     # Language dependencies
-    depends_on("c", type="build")
-    depends_on("cxx", type="build")
-    depends_on("fortran", type="build", when="+fortran")
+    depends_on("c", type=("build", "run"))
+    depends_on("cxx", type=("build", "run"))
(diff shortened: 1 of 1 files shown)


wrwilliams:
@spackbot fix style

wrwilliams:
@spackbot fix style

wrwilliams:
@tldahlgren best practice question: 9.0 and 9.1 both have crash bugs in pthread instrumentation fixed by their successors. Should they both already be marked deprecated as part of this PR even if we're not deprecating Score-P 8?

tldahlgren:
> @tldahlgren best practice question: 9.0 and 9.1 both have crash bugs in pthread instrumentation fixed by their successors. Should they both already be marked deprecated as part of this PR even if we're not deprecating Score-P 8?

Good question. Buggy code isn't great to have around. Have they ever worked? 

I see `scalasca` has a dependency on `scorep@8:`, so that's covered. The other two packages (per `spack dependents`) are `chatterbug` and `tau` and neither are explicit on version.

Buggy software isn't explicitly mentioned in the guidelines at https://spack.readthedocs.io/en/latest/packaging_guide_creation.html#deprecating-old-versions. IMO perhaps a comment explaining why they are excluded is appropriate (so no-one else comes along and tries to add them later).

tldahlgren:
Should've mentioned I have run across other packages that skip buggy versions.

repos/spack_repo/builtin/packages/scorep/package.py
```diff
@@ -107,21 +108,26 @@ def url_for_version(self, version):
...
     # Dependencies for SCORE-P are quite tight. See the homepage for more
     # information. Starting with scorep 4.0 / cube 4.4, Score-P only depends on
     # two components of cube -- cubew and cubelib.
 
     # Language dependencies
-    # TODO: we could allow a +fortran variant here.
     depends_on("c", type="build")  # generated
     depends_on("cxx", type="build")  # generated
-    depends_on("fortran", type="build")  # generated
+    depends_on("fortran", type="build", when="+fortran")  # generated
```
tldahlgren: Please remove the comment (`# generated`) if the language dependencies are confirmed.

repos/spack_repo/builtin/packages/scorep/package.py
```diff
@@ -107,21 +108,26 @@ def url_for_version(self, version):
...
     )
+    variant("fortran", default=True, description="Enable fortran support")
     # Putting this in as preparation. F08 support exists in 9.0 but configure does not respect
     # --enable-mpi-f08 and will not until 9.1.
-    variant("mpi_f08", default=True, description="Enable MPI F08 support", when="@9.1: +mpi")
+    variant(
+        "mpi_f08", default=True, description="Enable MPI F08 support", when="@9.1: +mpi +fortran"
+    )
+    variant(
+        "gotcha", default=True, description="Enable library wrapping with libgotcha", when="@9.0:"
```
tldahlgren: You can simplify:

```suggestion
        "gotcha", default=True, description="Enable library wrapping with libgotcha", when="@9:"
```

tldahlgren (changes requested):
Confirmed the new sha256. Have a couple of minor comments/suggestions. Let me know what you plan to do about `9.0` and `9.1`.

repos/spack_repo/builtin/packages/scorep/package.py
wrwilliams: Done. Arguably could/should be "build, run" to ensure that compiler instrumentation works; that will force the same virtual provider for the language to be loaded as what we built with, right?

repos/spack_repo/builtin/packages/scorep/package.py
wrwilliams: Done.

repos/spack_repo/builtin/packages/scorep/package.py
tldahlgren: Good question. That is generally the case with dependencies but isn't the instrumentation added during the build?

wrwilliams: During the build of whatever _uses_ scorep is when compiler instrumentation will get added. And unless someone decides that we should be a virtual provider for everything we instrument, that will generally not be at build time for any spack-controlled package.

Concretely: I build an instrumented LULESH with scorep loaded and manually adding our compiler wrappers. This will add a bunch of our pre-cooked adapters at LULESH link time, and will inject compiler instrumentation (possibly via -finstrument-functions, possibly via plugin) as part of the LULESH compile process. From scorep’s perspective this is runtime IIUC.

So again, IIUC, for any dependency where RPATH is insufficient, ie we need the same executable in PATH that was in our build environment or we need to know that our user will not link a conflicting .so, we actually should be adding a “runtime” flag.

It’s late here but I will pull that change together tomorrow if the above makes sense.

tldahlgren (approved):
Thanks.

PR #597, opened 2025-07-15
motioncor2: add/remove multiple versions, fix cuda version, use scipion download server

Dreista:
Changes in this pull request:
- Use Scipion server so that manual downloading is no longer required
- Add multiple deprecated versions that is available on Scipion server
- Remove old versions without CUDA version tag since nobody knows which CUDA version they were compiled with
- Update deprecated versions according to MotionCor2 official download page
- Fix CUDA dependencies for all available versions
- Update install procedure for all versions


repos/spack_repo/builtin/packages/motioncor2/package.py
```diff
@@ -19,46 +17,74 @@ class Motioncor2(Package):
...
     homepage = "http://msg.ucsf.edu/em/software"
-    manual_download = True
+    # Use Scipion's download server as the official site no longer hosts some deprecated versions
+    url = "https://scipion.cnb.csic.es/downloads/scipion/software/em/motioncor2-1.6.4.tgz"
 
-    version("1.6.4", sha256="28bb3e6477abf34fe41a78bcb9da9d77d08e2e89ecd41240fab085a308e6c498")
-    version("1.4.7", sha256="8c33969b10916835b55f14f3c370f67ebe5c4b2a9df9ec487c5251710f038e6b")
+    version("1.4.5", sha256="2f04bceb6069cd4f91a51447aac07b887be46206e094555bc445e89944ed2851")
+    version("1.5.0", sha256="20cdb46efe3f3c578071ddbd9e9f45c8462b0cbb2fc5b3cfb2d0b66ded0d30e0")
+    version("1.6.4", sha256="bcb70ddde38daa7722e007f2509bae928c8603530524b6a36ed79f4ca7b3edb8")
```
alalazo: Can you put versions in the opposite order, like other packages?

repos/spack_repo/builtin/packages/motioncor2/package.py
```diff
@@ -19,46 +17,74 @@ class Motioncor2(Package):
     to keep up with automated data collection."""
 
     homepage = "http://msg.ucsf.edu/em/software"
-    manual_download = True
+    # Use Scipion's download server as the official site no longer hosts some deprecated versions
+    url = "https://scipion.cnb.csic.es/downloads/scipion/software/em/motioncor2-1.6.4.tgz"
 
-    version("1.6.4", sha256="28bb3e6477abf34fe41a78bcb9da9d77d08e2e89ecd41240fab085a308e6c498")
-    version("1.4.7", sha256="8c33969b10916835b55f14f3c370f67ebe5c4b2a9df9ec487c5251710f038e6b")
+    version("1.4.5", sha256="2f04bceb6069cd4f91a51447aac07b887be46206e094555bc445e89944ed2851")
```
alalazo: Is there a reason v1.4.5 is not deprecated, while v1.4.7 is deprecated?

repos/spack_repo/builtin/packages/motioncor2/package.py
```diff
@@ -19,46 +17,74 @@ class Motioncor2(Package):
...
+        version("1.2.3", sha256="49dfe25316f9f65f77861383a583e1cd9b466abd6dc0e9f446ffbdefdafeb370")
+        version("1.2.6", sha256="b16d512bd4a122563d73a568d44bb724fb8b96d0d20662ce1a94274fb39dfaa6")
+        version("1.3.0", sha256="7da654830ec64c4b66a965d184d008f310d90fa4594b9166da1d530381275e5f")
+        version("1.3.1", sha256="861fb9ab9fbe0ca72e538ef7c2f8e5d2ebd02d9b6c852c8d57a48782a3a89547")
+        version("1.3.2", sha256="d9fb3537278be179fd9bfa45f58e921649a893b169cbf8fa173ba95fdb831d10")
+        version("1.4.0", sha256="d68c4ab9c6a740a73aede7120854a1fb3aea94279afb5fe2ab50387b725e6329")
+        version("1.4.2", sha256="97735aa22d01135933c5b698d8990ed371a89aec90df5193a24400c6460eb276")
+        version("1.4.4", sha256="1ced27bbc46a437ea1e5af8c9c842c9789da100d860d3663fc8215266cb4678e")
+        version("1.4.7", sha256="1c6ee6b76413db2d044a0048957a23907ad3217225d5e1a3659df3c3947d0e2c")
+        version("1.6.3", sha256="16c97516992d638ea817a943185213874dfad7d87a6bb71e102598922bb35c75")
```
alalazo: No need to add new versions that start as deprecated. Just remove them.

repos/spack_repo/builtin/packages/motioncor2/package.py
```diff
@@ -19,46 +17,74 @@ class Motioncor2(Package):
...
+    depends_on("cuda@8.0,9.2,10.1", type="run", when="@1.2.6:1.3.0")
+    depends_on("cuda@9.2,10.1:10.2", type="run", when="@1.3.1")
+    depends_on("cuda@8.0,9.2,10.0:10.2", type="run", when="@1.3.2")
+    depends_on("cuda@9.2,10.0:10.2,11.0", type="run", when="@1.4.0")
+    depends_on("cuda@10.0:10.2,11.1", type="run", when="@1.4.2")
+    depends_on("cuda@10.0:10.2,11.1:11.3", type="run", when="@1.4.4:1.4.5")
+    depends_on("cuda@10.2,11.1:11.5", type="run", when="@1.4.7")
+    depends_on("cuda@10.1:10.2,11.1:11.6", type="run", when="@1.5.0")
+    depends_on("cuda@10.1:10.2,11.1:11.8", type="run", when="@1.6.3")
+    depends_on("cuda@9.2,10.1:10.2,11.1:11.8,12.1", type="run", when="@1.6.4")
```
alalazo: You can use:
```python
with default_args(type="run"):
    ....
```

alalazo (changes requested):

repos/spack_repo/builtin/packages/motioncor2/package.py
Dreista: MotionCor2 is known to make breaking changes in patch versions so I'd rather keep these deprecated versions just in case someone needs them.

See: https://discuss.cryosparc.com/t/motioncor2-error-no-such-file-or-directory/8591/6
Command-line arguments were changed during 1.4.5 -> 1.4.7 update.

repos/spack_repo/builtin/packages/motioncor2/package.py
Dreista: Done

repos/spack_repo/builtin/packages/motioncor2/package.py
Dreista: This is base on their [official Google Forms](https://docs.google.com/forms/d/e/1FAIpQLSfAQm5MA81qTx90W9JL6ClzSrM77tytsvyyHh1ZZWrFByhmfQ/viewform) listed [here](https://msg.ucsf.edu/software). Only 1.6.4, 1.5.0, and 1.4.5 are offered.

Command-line arguments were changed during 1.4.5 -> 1.4.7 update, and for certain Cryo-EM workflow (CryoSPARC) it has hard dependency on version 1.4.5. Using 1.4.7 simply won't work.
See: https://discuss.cryosparc.com/t/motioncor2-error-no-such-file-or-directory/8591/6

repos/spack_repo/builtin/packages/motioncor2/package.py
Dreista: Done

repos/spack_repo/builtin/packages/motioncor2/package.py
```diff
@@ -19,46 +17,76 @@ class Motioncor2(Package):
...
-    )
-    version(
-        "1.0.4",
-        sha256="c75738160ac18d3f27c33677e78e63313d8ec2b023b5a46173428c3fa0451a94",
-        url="http://msg.ucsf.edu/MotionCor2/MotionCor2-1.0.4.tar.gz",
-        deprecated=True,
-    )
+    # Deprecated versions, no longer available in official site
+    with default_args(deprecated=True):
+        version("1.6.3", sha256="16c97516992d638ea817a943185213874dfad7d87a6bb71e102598922bb35c75")
```
bernhardkaindl, line 29: As said, in genereal and despite breaking changes in minor versions like you said, there should be no reason to add new versions (like 1.6.3, AFAICS) as deprecated, unless you know a specific case were someone needs that version.
```suggestion
```

repos/spack_repo/builtin/packages/motioncor2/package.py
```diff
@@ -19,46 +17,76 @@ class Motioncor2(Package):
...
+        version("1.4.4", sha256="1ced27bbc46a437ea1e5af8c9c842c9789da100d860d3663fc8215266cb4678e")
+        version("1.4.2", sha256="97735aa22d01135933c5b698d8990ed371a89aec90df5193a24400c6460eb276")
+        version("1.4.0", sha256="d68c4ab9c6a740a73aede7120854a1fb3aea94279afb5fe2ab50387b725e6329")
+        version("1.3.2", sha256="d9fb3537278be179fd9bfa45f58e921649a893b169cbf8fa173ba95fdb831d10")
+        version("1.3.1", sha256="861fb9ab9fbe0ca72e538ef7c2f8e5d2ebd02d9b6c852c8d57a48782a3a89547")
+        version("1.3.0", sha256="7da654830ec64c4b66a965d184d008f310d90fa4594b9166da1d530381275e5f")
+        version("1.2.6", sha256="b16d512bd4a122563d73a568d44bb724fb8b96d0d20662ce1a94274fb39dfaa6")
+        version("1.2.3", sha256="49dfe25316f9f65f77861383a583e1cd9b466abd6dc0e9f446ffbdefdafeb370")
+        version("1.2.1", sha256="b68db6850d64fc56635ff968b53587c96bb42108ba58a415bef447f0810b9e57")
+        version("1.1.0", sha256="1cb7c486d7f575fefcc7dae5037e9bb645842e1e43b22fa3c18fd0e4d2a189db")
```
bernhardkaindl: > MotionCor2 is known to make breaking changes in patch versions so I'd rather keep these deprecated versions just in case someone needs them.

I think what alalazo said would still hold because I don't see that these versions existed before:

```suggestion
```

Dreista:
This should not be marked stale.

repos/spack_repo/builtin/packages/motioncor2/package.py
```diff
@@ -19,27 +17,54 @@ class Motioncor2(Package):
     to keep up with automated data collection."""
 
     homepage = "http://msg.ucsf.edu/em/software"
-    manual_download = True
+    # Use Scipion's download server as the official site no longer hosts some deprecated versions
+    url = "https://scipion.cnb.csic.es/downloads/scipion/software/em/motioncor2-1.6.4.tgz"
```
alalazo, line 21: I would still keep this download manual, given that:

> This software may only be downloaded and used for free by academic and/or non-profit users. All others are required to contact David Agard for licensing information prior to download.

Also, can we verify why the shasum changes for versions we already had?

bernhardkaindl: @Dreista, I'm not sure what's the status for this PR, can you sched some light on it?

If what alalazo found is true, he might be right in that tis would need at least a download with confirmation or a manual download as indicated.

PR #601, opened 2025-07-15
py_googleapis_common_protos: add version 1.70.0 and update repo

benoitmartin88:
Original PR (https://github.com/spack/spack-packages/pull/428) was closed due to owner transfer.

Needed for a py-ray update.

CC @viperML @theabm

### DIFF
--- a/repos/spack_repo/builtin/packages/py_googleapis_common_protos/package.py
+++ b/repos/spack_repo/builtin/packages/py_googleapis_common_protos/package.py
@@ -10,18 +10,16 @@ from spack.package import *
 class PyGoogleapisCommonProtos(PythonPackage):
     """Common protobufs used in Google APIs."""
 
-    homepage = "https://github.com/googleapis/python-api-common-protos"
-    pypi = "googleapis-common-protos/googleapis-common-protos-1.6.0.tar.gz"
+    homepage = "https://github.com/googleapis/google-cloud-python/tree/main/packages/googleapis-common-protos"
+    pypi = "googleapis_common_protos/googleapis_common_protos-1.70.0.tar.gz"
 
     license("Apache-2.0")
 
+    version("1.70.0", sha256="0e1b44e0ea153e6594f9f394fef15193a68aaaea2d843f83e2742717ca753257")
     version("1.63.0", sha256="17ad01b11d5f1d0171c06d3ba5c04c54474e883b66b949722b4938ee2694ef4e")
     version("1.58.0", sha256="c727251ec025947d545184ba17e3578840fc3a24a0516a020479edab660457df")
     version("1.56.4", sha256="c25873c47279387cfdcbdafa36149887901d36202cb645a0e4f29686bf6e4417")
     version("1.55.0", sha256="53eb313064738f45d5ac634155ae208e121c963659627b90dfcb61ef514c03e1")
-    version("1.6.0", sha256="e61b8ed5e36b976b487c6e7b15f31bb10c7a0ca7bd5c0e837f4afab64b53a0c6")
-
-    # https://github.com/googleapis/python-api-common-protos/blob/v1.58.0/setup.py
 
     variant(
         "grpc",
@@ -32,18 +30,13 @@ class PyGoogleapisCommonProtos(PythonPackage):
     depends_on("py-setuptools", type="build")
 
     with default_args(type=("build", "run")):
-        # https://github.com/googleapis/python-api-common-protos/blob/main/setup.py
-        # May be able to rais max version to :5 in next release
-        depends_on("py-protobuf@3.19.5:4", when="@1.58:")
+        depends_on("py-protobuf@3.22.2:6", when="@1.70:")
+        depends_on("py-protobuf@3.19.5:4", when="@1.58:1.63")
         depends_on("py-protobuf@3.15.0:4", when="@1.56:1.57")
         depends_on("py-protobuf@3.12.0:4", when="@1.55")
-        depends_on("py-protobuf@3.6.0:", when="@1.6.0:")
 
-    # Explicitly incompatibile versions per setup.py
-    # https://github.com/googleapis/python-api-common-protos/issues/128
-    conflicts("py-protobuf@3.20:3.20.1")
     conflicts("py-protobuf@4.21.1:4.21.5")
 
     with when("+grpc"), default_args(type="run"):
-        depends_on("py-grpcio@1.44:1", when="@1.57:")
-        depends_on("py-grpcio@1:")
+        depends_on("py-grpcio@1.44:1")
+


repos/spack_repo/builtin/packages/py_googleapis_common_protos/package.py
```diff
@@ -32,18 +30,13 @@ class PyGoogleapisCommonProtos(PythonPackage):
...
 
-    # Explicitly incompatibile versions per setup.py
-    # https://github.com/googleapis/python-api-common-protos/issues/128
-    conflicts("py-protobuf@3.20:3.20.1")
     conflicts("py-protobuf@4.21.1:4.21.5")
 
     with when("+grpc"), default_args(type="run"):
-        depends_on("py-grpcio@1.44:1", when="@1.57:")
-        depends_on("py-grpcio@1:")
+        depends_on("py-grpcio@1.44:1")
```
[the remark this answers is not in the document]
bernhardkaindl: Copilot suggested complicated version ranges, but we don't need them to be that complex. Simple looks better too.

repos/spack_repo/builtin/packages/py_googleapis_common_protos/package.py
```diff
@@ -10,18 +10,16 @@
...
+    pypi = "googleapis_common_protos/googleapis_common_protos-1.70.0.tar.gz"
 
     license("Apache-2.0")
 
+    version("1.70.0", sha256="0e1b44e0ea153e6594f9f394fef15193a68aaaea2d843f83e2742717ca753257")
     version("1.63.0", sha256="17ad01b11d5f1d0171c06d3ba5c04c54474e883b66b949722b4938ee2694ef4e")
     version("1.58.0", sha256="c727251ec025947d545184ba17e3578840fc3a24a0516a020479edab660457df")
     version("1.56.4", sha256="c25873c47279387cfdcbdafa36149887901d36202cb645a0e4f29686bf6e4417")
     version("1.55.0", sha256="53eb313064738f45d5ac634155ae208e121c963659627b90dfcb61ef514c03e1")
-    version("1.6.0", sha256="e61b8ed5e36b976b487c6e7b15f31bb10c7a0ca7bd5c0e837f4afab64b53a0c6")
```
w8jcik, line 22: As far as I know we don't delete versions, but rather deprecate them.

I would put it back and mark it with `deprecated=True`.

repos/spack_repo/builtin/packages/py_googleapis_common_protos/package.py
```diff
@@ -10,18 +10,16 @@
...
-
-    # https://github.com/googleapis/python-api-common-protos/blob/v1.58.0/setup.py
```
w8jcik, line 24: This URL was helpful to find older versions of package definitions. Sometimes old versions also need maintenance, due to incorrectly stated dependencies. How about moving it between `version` lines?

repos/spack_repo/builtin/packages/py_googleapis_common_protos/package.py
```diff
@@ -32,18 +30,13 @@ class PyGoogleapisCommonProtos(PythonPackage):
     depends_on("py-setuptools", type="build")
 
     with default_args(type=("build", "run")):
-        # https://github.com/googleapis/python-api-common-protos/blob/main/setup.py
-        # May be able to rais max version to :5 in next release
-        depends_on("py-protobuf@3.19.5:4", when="@1.58:")
+        depends_on("py-protobuf@3.22.2:6", when="@1.70:")
+        depends_on("py-protobuf@3.19.5:4", when="@1.58:1.63")
```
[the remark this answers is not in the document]
w8jcik: I don't mind AI support, but this is AI nonsense.

On the other hand it highlights another issue that this PR doesn't cover.

It doesn't describe dependencies for versions `1.64` to `1.69`. In case someone installs `py-googleapis-common-protos@=1.64` your specification will fail to deliver correct dependencies. We can skip version hashes in case only particular version was tested, but there is no reason to not describe the dependencies correctly.

repos/spack_repo/builtin/packages/py_googleapis_common_protos/package.py
```diff
@@ -32,18 +30,13 @@ class PyGoogleapisCommonProtos(PythonPackage):
...
         depends_on("py-protobuf@3.15.0:4", when="@1.56:1.57")
         depends_on("py-protobuf@3.12.0:4", when="@1.55")
-        depends_on("py-protobuf@3.6.0:", when="@1.6.0:")
 
-    # Explicitly incompatibile versions per setup.py
-    # https://github.com/googleapis/python-api-common-protos/issues/128
-    conflicts("py-protobuf@3.20:3.20.1")
```
w8jcik, line 44: This conflict is explicitly stated in `setup.py` from the authors of the package. What is the reason to delete these lines?

The conflict was still present in v1.66
https://github.com/googleapis/google-cloud-python/blob/googleapis-common-protos-v1.66.0/packages/googleapis-common-protos/setup.py

but absent in v1.67
https://github.com/googleapis/google-cloud-python/blob/googleapis-common-protos-v1.67.0/packages/googleapis-common-protos/setup.py

It means that you should bring this line back and add `when="@:1.66"`.

repos/spack_repo/builtin/packages/py_googleapis_common_protos/package.py
w8jcik, line 41: If we don't remove versions but deprecate them, then this constraints shouldn't be altered either.

w8jcik (changes requested):

repos/spack_repo/builtin/packages/py_googleapis_common_protos/package.py
```diff
@@ -32,18 +30,13 @@ class PyGoogleapisCommonProtos(PythonPackage):
     depends_on("py-setuptools", type="build")
 
     with default_args(type=("build", "run")):
-        # https://github.com/googleapis/python-api-common-protos/blob/main/setup.py
-        # May be able to rais max version to :5 in next release
-        depends_on("py-protobuf@3.19.5:4", when="@1.58:")
+        depends_on("py-protobuf@3.22.2:6", when="@1.70:")
```
w8jcik, line 33: The version range that you provided doesn't match the content of `pyproject.toml`
https://github.com/googleapis/google-cloud-python/blob/googleapis-common-protos-v1.70.0/packages/googleapis-common-protos/pyproject.toml#L44

repos/spack_repo/builtin/packages/py_googleapis_common_protos/package.py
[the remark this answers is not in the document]
w8jcik: I believe that this would be correct

```diff
-        depends_on("py-protobuf@3.19.5:4", when="@1.58:")
+        depends_on("py-protobuf@3.20.2:6", when="@1.70:")
+        depends_on("py-protobuf@3.20.2:5", when="@1.64:1.69")
+        depends_on("py-protobuf@3.19.5:4", when="@1.58:1.63")
```

w8jcik (changes requested):

repos/spack_repo/builtin/packages/py_googleapis_common_protos/package.py
```diff
@@ -32,18 +30,13 @@ class PyGoogleapisCommonProtos(PythonPackage):
...
+        depends_on("py-protobuf@3.19.5:4", when="@1.58:1.63")
         depends_on("py-protobuf@3.15.0:4", when="@1.56:1.57")
         depends_on("py-protobuf@3.12.0:4", when="@1.55")
-        depends_on("py-protobuf@3.6.0:", when="@1.6.0:")
```
w8jcik, line 40: This should be restored if we mark 1.6.0 as deprecated.

bernhardkaindl (changes requested):
@w8jcik is correct:

- `spack` don't remove versions but deprecate them, then this constraints shouldn't be altered either. After they have been deprecated for a spack release, then we can remove those deprecated verions.

- `spack` don't remove version constraints, it keeps them, even if those versions are deprecated or have no checksums. PRs don't create holes in the version depencies even if those versions are not listed, as they might be added by the used if need be.

- Only when a deprecated version range has been removed because they are below the version range spack can build, it makes sense to also remove the associated dependency versions.

PR #609, opened 2025-07-15
py-argcomplete: add v3.6.2

manuelakuhn:
https://github.com/kislyuk/argcomplete/tree/v3.6.2


alecbcs (approved):

(Side note: changing build backends in a patch release feels insane to me...but ah okay.)

PR #611, opened 2025-07-16
[Draft/WIP] meep dependents: Add missing compliler-depends

EMinsight:
The Meep depentent packages libctl, libgdsii and mpb are missing some compiler dependency.

### DIFF
--- a/repos/spack_repo/builtin/packages/libctl/package.py
+++ b/repos/spack_repo/builtin/packages/libctl/package.py
@@ -27,6 +27,7 @@ class Libctl(AutotoolsPackage):
     )
 
     depends_on("c", type="build")  # generated
+    depends_on("fortran", type="build")  # generated
 
     depends_on("guile")
 
--- a/repos/spack_repo/builtin/packages/mpb/package.py
+++ b/repos/spack_repo/builtin/packages/mpb/package.py
@@ -19,6 +19,7 @@ class Mpb(AutotoolsPackage):
     version("1.11.1", sha256="7311fc525214c1184cad3e0626b8540c0b53b3c31c28e61ce6ec2860088eca46")
 
     depends_on("c", type="build")  # generated
+    depends_on("fortran", type="build")  # generated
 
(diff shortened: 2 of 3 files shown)


alecbcs (changes requested):
@EMinsight if you've confirmed the language dependencies let's drop the `# generated` suffixes. Those were originally added by the automated tool @haampie wrote.

EMinsight:
I am confused about the compiler dependencies. This packages actually are not use this compiler which I have added. But they will failed during configure or compile stages. Maybe someone need to check the configure file.

repos/spack_repo/builtin/packages/mpb/package.py
```diff
@@ -19,6 +19,7 @@ class Mpb(AutotoolsPackage):
     version("1.11.1", sha256="7311fc525214c1184cad3e0626b8540c0b53b3c31c28e61ce6ec2860088eca46")
 
     depends_on("c", type="build")  # generated
+    depends_on("fortran", type="build")  # generated
```
alecbcs, line 22: `mpb` depends on fortran when it detects a BLAS implementation. What we should likely do here is to create a blas variant and update the fortran dependency to be conditional on that variant.

```
variant("blas", default=False, description="Enable BLAS support")

...
...
...

depends_on("fortran", type="build", when="+blas")

...
...

depends_on("blas", when="+blas")
```

repos/spack_repo/builtin/packages/libctl/package.py
```diff
@@ -27,6 +27,7 @@ class Libctl(AutotoolsPackage):
     )
 
     depends_on("c", type="build")  # generated
+    depends_on("fortran", type="build")  # generated
```
alecbcs, line 30: @EMinsight does `libctl` fail to build for you without `fortran` or does it fail later to link with `meep`?

EMinsight:
closed due to #2568 #2567 #2564 .

PR #614, opened 2025-07-16
Patch py-kornia-rs to pin fast_image_resize

kwryankrattiger:
Fixes [failure on develop](https://gitlab.spack.io/spack/spack-packages/-/jobs/17386285)

Newer versions of fast_image_resize require integer_sign_cast which was stabilized in rust 1.87. This was already done in newer releases of py-kornia-rs.

### DIFF
--- a/repos/spack_repo/builtin/packages/py_kornia_rs/package.py
+++ b/repos/spack_repo/builtin/packages/py_kornia_rs/package.py
@@ -43,6 +43,9 @@ class PyKorniaRs(PythonPackage):
     # dlpack-rs needs libclang
     depends_on("llvm+clang")
 
+    # Pin fast_image_resize to 5.1.0, >=5.2 requires rust >=1.87
+    patch("py-kornia-rs-pin-fast_image_resize.patch", when="@0.1.9")
+
     @property
     def build_directory(self):
         return "kornia-py" if self.spec.satisfies("@0.1.3:") else "py-kornia"
--- a/repos/spack_repo/builtin/packages/py_kornia_rs/package.py
+++ b/repos/spack_repo/builtin/packages/py_kornia_rs/package.py
@@ -43,7 +43,7 @@ class PyKorniaRs(PythonPackage):
     # dlpack-rs needs libclang
     depends_on("llvm+clang")
 
-    # Pin fast_image_resize to 5.1.0, >=5.2 requires rust >=1.87
+    # See https://github.com/kornia/kornia-rs/commit/93f768137814709b60767f7fc24a9b0184002aee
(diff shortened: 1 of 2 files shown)


repos/spack_repo/builtin/packages/py_kornia_rs/package.py
```diff
@@ -43,6 +43,9 @@ class PyKorniaRs(PythonPackage):
     # dlpack-rs needs libclang
     depends_on("llvm+clang")
 
+    # Pin fast_image_resize to 5.1.0, >=5.2 requires rust >=1.87
+    patch("py-kornia-rs-pin-fast_image_resize.patch", when="@0.1.9")
```
adamjstewart, line 47: Is this patch also needed to build older versions? Can you add a link to the PR where it was patched in future versions?

adamjstewart (commented):
Thanks for investigating and fixing!

repos/spack_repo/builtin/packages/py_kornia_rs/package.py
kwryankrattiger: It was fixed here, seems to be related to a similar issue.

https://github.com/kornia/kornia-rs/commit/93f768137814709b60767f7fc24a9b0184002aee

As far as older versions included in the spack package, they seemed to rely on older versions/different libraries.

adamjstewart (dismissed):
Could maybe use a comment with that link but it will be great to get this merged soon.

repos/spack_repo/builtin/packages/py_kornia_rs/py-kornia-rs-pin-fast_image_resize.patch
```diff
@@ -0,0 +1,13 @@
+diff --git a/crates/kornia-imgproc/Cargo.toml b/crates/kornia-imgproc/Cargo.toml
+index b501eb0..7950c30 100644
+--- a/crates/kornia-imgproc/Cargo.toml
++++ b/crates/kornia-imgproc/Cargo.toml
+@@ -12,7 +12,7 @@ rust-version.workspace = true
+ version.workspace = true
+ 
+ [dependencies]
+-fast_image_resize = "5.1.0"
++fast_image_resize = "=5.1.0"
```
haampie:
```suggestion
+fast_image_resize = "=5.1"
```

there are various patch releases, you shouldn't pin a patch version.

kwryankrattiger: fair, thanks

PR #618, opened 2025-07-16
py-asttokens: add v3.0.0

manuelakuhn:
https://github.com/gristlabs/asttokens/tree/v3.0.0

### DIFF
--- a/repos/spack_repo/builtin/packages/py_asttokens/package.py
+++ b/repos/spack_repo/builtin/packages/py_asttokens/package.py
@@ -15,6 +15,7 @@ class PyAsttokens(PythonPackage):
 
     license("Apache-2.0")
 
+    version("3.0.0", sha256="0dcd8baa8d62b0c1d118b399b2ddba3c4aff271d0d7a9e0d4c1681c79035bbc7")
     version("2.4.0", sha256="2e0171b991b2c959acc6c49318049236844a5da1d65ba2672c4880c1c894834e")
     version("2.2.1", sha256="4622110b2a6f30b77e1473affaa97e711bc2f07d3f10848420ff1898edbe94f3")
     version("2.0.8", sha256="c61e16246ecfb2cde2958406b4c8ebc043c9e6d73aaa83c941673b35e5d3a76b")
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/py_asttokens/package.py
```diff
@@ -23,5 +24,6 @@ class PyAsttokens(PythonPackage):
     depends_on("py-setuptools@44:", type="build")
     depends_on("py-setuptools-scm+toml@3.4.3:", type="build")
 
-    depends_on("py-six@1.12:", when="@2.3:", type=("build", "run"))
-    depends_on("py-six", type=("build", "run"))
+    with when("@:2"):
```
alecbcs, line 27: Was the previous constraint on `py-six@1.12` limited to `@2.3:` an error? Looks like it got dropped in the transition here.

manuelakuhn: Sorry, maybe I am misunderstanding what you mean, but why dropped? It is still there in 
```
    with when("@:2"):
        depends_on("py-six@1.12:", when="@2.3:", type=("build", "run"))
```
I did some tests, and as far as I can tell the two `when` restrictions are merged to `when="@2.3:2"`.

In case you mean the `py-asttokens` package itself. It had a `py-six@1.12:` restiction for versions 2.3.0, 2.4.0, 2.4.1
https://github.com/gristlabs/asttokens/blob/v2.3.0/setup.cfg#L38
They dropped the whole `py-six` dependency with version 3.0.0

bernhardkaindl: This review comment has been resoundingly answered and not applicable.

@alecbcs self-assigned this 3 weeks ago, but did not respond, so I guess his self-assgign is no longer in force?

bernhardkaindl (approved):
The review comment by @alecbcs, which was resolved 3 weeks ago, is now resolved.
He didn't respond to the answer, so I assume his self-assignment is no longer in effect.
```py
../spack/bin/spack find -v py-asttokens
-- linux-ubuntu24.04-x86_64_v4 / no compilers -------------------
py-asttokens@2.4.0 build_system=python_pip  py-asttokens@3.0.0 build_system=python_pip
==> 2 installed packages
```

bernhardkaindl (approved):

PR #622, opened 2025-07-16
py-async-lru: add 2.0.5

manuelakuhn:
Add version 2.0.5 and fix tar ball name that changed on pypi from async-lru back to async_lru
https://github.com/aio-libs/async-lru/tree/v2.0.5


tldahlgren (approved):
(Re)confirmed version sha256s and confirmed dependencies of latest version.

PR #619, opened 2025-07-16
Require newer version of llvm-openmp for xgboost and py_torch on macOS

zackgalbreath:
* Add v19.1.7 and v20.1.8 for llvm-openmp
* Depend on llvm-openmp v19 or newer when building xgboost or torch with OpenMP using AppleClang
* Remove py-torchtext from ml-darwin-aarch64-mps so we can re-enable the stack while debugging this build failure

PR #624, opened 2025-07-16
Disable ml mps in config

kwryankrattiger:
This will allow verifying the fixes

@zackgalbreath merge this and rebase, and then we can drop the hotfix from the gitlab variables.

This should unblock testing #619


eugeneswalker (approved):

kwryankrattiger:
This was fixed offline, will drop temporary variable from gitlab

PR #630, opened 2025-07-17
py-babel: add v2.17.0

manuelakuhn:
https://github.com/python-babel/babel/tree/v2.17.0


alecbcs (approved):

PR #634, opened 2025-07-17
py-bids-validator: add v1.14.7.post0 and new dependency packages

manuelakuhn:
Add `py-bids-validator` version 1.14.7.post0
https://pypi.org/project/bidsschematools/#files

Add new dependency packages `py-bidsschematools` and `py-acres`
https://pypi.org/project/bidsschematools/#files
https://github.com/nipreps/acres/tree/0.5.0

### DIFF
--- a/repos/spack_repo/builtin/packages/py_bids_validator/package.py
+++ b/repos/spack_repo/builtin/packages/py_bids_validator/package.py
@@ -11,8 +11,11 @@ class PyBidsValidator(PythonPackage):
     """Validator for the Brain Imaging Data Structure"""
 
     homepage = "https://github.com/bids-standard/bids-validator"
-    pypi = "bids-validator/bids-validator-1.7.2.tar.gz"
+    pypi = "bids-validator/bids_validator-1.14.6.tar.gz"
 
+    license("MIT")
+
+    version("1.14.7.post0", sha256="e6005a500b75f8a961593fb67d46085107dadb116f59a5c3b524aa0697945b66")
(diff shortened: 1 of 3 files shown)


repos/spack_repo/builtin/packages/py_bids_validator/package.py
```diff
@@ -11,8 +11,13 @@ class PyBidsValidator(PythonPackage):
     """Validator for the Brain Imaging Data Structure"""
 
     homepage = "https://github.com/bids-standard/bids-validator"
-    pypi = "bids-validator/bids-validator-1.7.2.tar.gz"
+    pypi = "bids-validator/bids_validator-1.14.6.tar.gz"
```
alecbcs: @manuelakuhn we should list the url as something after `v1.14.6` to allow `spack versions` to continue to dynamically find new versions. (`url_for_version()` unfortunately doesn't work for the crawler.)

manuelakuhn: I have exchanged the version in the url now with `1.14.7.post0` (the newest one) as requested. I didn't use it before because I did not know if it was a good idea to use a `post0` here  and a `1.14.7` does not exist. I used `1.14.6` because it already uses the new naming so that the crawler worked.

bernhardkaindl (approved):
The resolved review comment by @alecbcs is now 3 weeks old, and his self-assignment is resolved.
He didn't respond to the answer, so I assume his self-assignment is no longer in effect.
```py
-- linux-ubuntu22.04-skylake / no compilers ---------------------
py-bids-validator@1.14.7.post0 build_system=python_pip
```

PR #637, opened 2025-07-17
BLAS: normalize test symbols on Windows

johnwparent:
Ensure correct compiler definitions are utilized when compiling tests on Windows to ensure tests are compiled with the same symbol mangling as the rest of BLAS.


repos/spack_repo/builtin/packages/openblas/package.py
```diff
@@ -118,8 +118,17 @@ class Openblas(CMakePackage, MakefilePackage):
...
     patch("openblas-0.3.29-darwin-aarch64.patch", when="@0.3.29 platform=darwin")
 
-    # https://github.com/xianyi/OpenBLAS/pull/2519/files
-    patch("ifort-msvc.patch", when="%msvc")
+    # Add proper compiler args for symbol mangingling when using
+    # intel fortran compilers
+    patch(
+        "https://github.com/OpenMathLib/OpenBLAS/commit/6d54c9476056b13fb91bf90bbc803968e0743aef.patch?full_index=1",
+        sha256="735a51af451f5afd5d366ba88bbe2f6930a9b1f69f3b924aed2c951eee81ccae",
+        when="@:0.3.29",
```
scheibelp: I recommend making this just `@0.3.29` or the earliest version you have applied it to yourself (while it's probably true that it would be needed for older versions, there will likely be some point going back where the patch doesn't apply anymore).

repos/spack_repo/builtin/packages/openblas/package.py
```diff
@@ -118,8 +118,17 @@ class Openblas(CMakePackage, MakefilePackage):
     # https://github.com/OpenMathLib/OpenBLAS/issues/5202
     patch("openblas-0.3.29-darwin-aarch64.patch", when="@0.3.29 platform=darwin")
 
-    # https://github.com/xianyi/OpenBLAS/pull/2519/files
-    patch("ifort-msvc.patch", when="%msvc")
+    # Add proper compiler args for symbol mangingling when using
```
scheibelp: (typo) mangling

scheibelp (changes requested):
One typo, and I have a request to constraint range of specs the new patch applies for.

repos/spack_repo/builtin/packages/openblas/package.py
scheibelp: Also: should this be applied on just Windows? Or perhaps only when using `%msvc` or `%oneapi`? I thought our pipelines were already building this package with the Intel Fortran compilers on Linux (and succeeding without this patch).

johnwparent: This patch was merged into OpenBlas main as is, as such it should be ok for everyone. That being said, I actually think Jared's patch was more robust (he didn't hardcode the check for `ifort` the patch merged in upstream did, `intel` is a much stronger check). I'm going to revert this to use Jared's changes, and make another patch myself upstream to use the changes Jared made.

repos/spack_repo/builtin/packages/openblas/package.py
johnwparent: Yup, nice catch thanks. Beyond this PR, but.. should we add spellcheck to Spack?

repos/spack_repo/builtin/packages/openblas/package.py
johnwparent: The applicability of this patch extends back several versions (that I've tried), I'll update to stick to those constraints

repos/spack_repo/builtin/packages/openblas/package.py
scheibelp: I personally dislike spell check in code: it discourages use of made-up words. Generally I don't point out a typo unless there is something else wrong, and IMO that's always worked out for me. Fully resolving that debate goes beyond this comment (which I will now collapse).

johnwparent: That's a good point I had not thought of, I can always run one locally (I probably should given how many of these you catch on my PRs), thanks.

repos/spack_repo/builtin/packages/openblas/package.py
johnwparent: I've updated this, to revert back to Jared's patch and the version is now constrained to "versions ive tested on"

alecbcs: I'd really like to get away from storing patches in the packages repository whenever possible.

I'd vote to revert the update with the same patch that merged into mainline so that we don't diverge / need to store an additional patch.

johnwparent: Unfortunately the patch merged in upstream doesn't work for our use case. I'll update it once I PR my patches to openblas and they land upstream.

scheibelp (approved):

PR #651, opened 2025-07-18
rocprofiler-systems: add internal-dyninst variant

afzpatel:
Dyninst version 12 is unable to concretize with GCC version 13 and above due to restrictions in the `intel-tbb` recipe. This PR introduces an `internal-dyninst` option, allowing us to build the custom [ROCm fork of Dyninst](https://github.com/ROCm/dyninst)


afzpatel:
For some reason I can't add @srekolam as a reviewer.

renjithravindrankannath (approved):
LGTM

afzpatel:
@tldahlgren can you please review this pr?

repos/spack_repo/builtin/packages/rocprofiler_systems/package.py
```diff
@@ -106,6 +106,9 @@ class RocprofilerSystems(CMakePackage):
             "(target application can use any MPI installation)"
         ),
     )
+    variant("internal-dyninst", default=False, description="build internal dyninst")
+    # Can't concretize dyninst@12 with gcc@13: due to restrictions in intel-tbb
+    requires("+internal-dyninst", when="%gcc@13:")
```
tldahlgren: I'm not fond of interleaving directives as it can lead to some confusion. This comment is one in particular. My initial reaction to seeing it (out of the context of other changes here) was that you need a `conflicts` directive.

I'm wondering if moving the comment at line 110 and directive at line 111 to after the other dependencies and above the patch at line 149/150 would help.

tldahlgren (commented):
I have a suggestion to aid readability (IMO). Let me know if you want to address it.

repos/spack_repo/builtin/packages/rocprofiler_systems/package.py
afzpatel: I realized that it's actually useless to include that, so I've removed it. If dyninst@12 can't be concretized, then +internal-dyninst will automatically be set.

afzpatel: If anything, that line of code imposes unnecessary restrictions on the package.

tldahlgren (approved):

PR #659, opened 2025-07-19
eigen: add v3.4.0-44-g37248b26a

wdconinc:
This PR adds a newer version of `eigen` by commit since it is a requirement for `py-onnxruntime` as of 1.17.0 (https://github.com/microsoft/onnxruntime/commit/812532592e070823f0362af8c3219ca6b01eda77#diff-12c22e06cbb37ea0ed9f9eaf60cbe408dbeef04072df6a9f431c3290822ea835).

This version is added to avoid using the vendored `eigen` that is downloaded during the `py-onnxruntime` installation (and which is now broken due to changes in the checksums of the archive zip files provided on gitlab, see https://github.com/microsoft/onnxruntime/issues/24861). See also https://github.com/spack/spack-packages/pull/128.

### DIFF
--- a/repos/spack_repo/builtin/packages/eigen/package.py
+++ b/repos/spack_repo/builtin/packages/eigen/package.py
@@ -23,7 +23,8 @@ class Eigen(CMakePackage, ROCmPackage):
     license("MPL-2.0")
 
     version("master", branch="master")
-    version("3.4.0", sha256="8586084f71f9bde545ee7fa6d00288b264a2b7ac3607b974e54d13e7162c1c72")
+    version("3.4.0-44-ge7248b26a", commit="e7248b26a1ed53fa030c5c459f7ea095dfd276ac")
+    version("3.4.0", sha256="8586084f71f9bde545ee7fa6d00288b264a2b7ac3607b974e54d13e7162c1c72", preferred=True)
     version("3.3.9", sha256="7985975b787340124786f092b3a07d594b2e9cd53bbfe5f3d9b1daee7d55f56f")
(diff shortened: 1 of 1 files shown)


alecbcs (approved):
Looks good to me. With the addition of this can we drop the custom patch files added to the py-onnxruntime package? 
Thanks @wdconinc!

wdconinc:
> With the addition of this can we drop the custom patch files added to the py-onnxruntime package?

That's the intention. It requires addition of `onnxruntime_USE_PREINSTALLED_EIGEN` and `eigen_SOURCE_PATH`. I have not 100% tested yet that for `eigen_SOURCE_PATH` it is sufficient to point to the prefix, but based on a look at the CMake files it seems like it should be.

alecbcs:
Would your preference be to merge this and then test those additions? Or would you mind holding off and merging it all as one PR which might be a bit cleaner?

wdconinc:
> Would your preference be to merge this and then test those additions? Or would you mind holding off and merging it all as one PR which might be a bit cleaner?

Either way is fine with me. I had put this in as a separate PR to keep the commits to different packages separate, but I can add the other changes here too.

wdconinc:
There's another eigen version that I would probably add at some point (https://github.com/microsoft/onnxruntime/blob/main/cmake/deps.txt#L24, https://gitlab.com/libeigen/eigen/-/commit/1d8b82b0740839c0de7f1242a3585e3390ff5f33). But even though it's in the nightly/master branch, it's not linearly connected to 3.4 (not sure how they did that), so it won't have a clean git-describe tag like the version here.

bernhardkaindl (approved):

PR #661, opened 2025-07-20
py-nanobind: add v2.8.0

garth-wells:


alecbcs (approved):
Looks good to me. Thanks @garth-wells!

PR #662, opened 2025-07-20
Remove Python-related deprecations

adamjstewart:
This PR removes all Python-related deprecated versions and packages. Non-Python deprecations will be included in a separate PR.


adamjstewart:
@spackbot fix style

adamjstewart:
Could probably get us to 5K lines removed if you want:
```console
> spack pkg grep 'python.*3.8' | wc -l
     401
```

repos/spack_repo/builtin/packages/libproxy/package.py
```diff
@@ -23,32 +23,18 @@ class Libproxy(CMakePackage):
     version("0.4.13", sha256="d610bc0ef81a18ba418d759c5f4f87bf7102229a9153fb397d7d490987330ffd")
 
     variant("perl", default=False, description="Enable Perl bindings")
-    variant("python", default=False, description="Enable Python bindings", when="@0.4.16:")
```
alecbcs, line 26: Instead of dropping the python variant here, could we bump the version of `libproxy` to get support for newer versions of Python? It looks like they're testing with 3.12 in their GitHub workflows.

repos/spack_repo/builtin/packages/memsurfer/package.py
```diff
@@ -1,46 +0,0 @@
...
-#
-# SPDX-License-Identifier: (Apache-2.0 OR MIT)
-
-
-from spack_repo.builtin.build_systems.python import PythonPackage
-
-from spack.package import *
-
-
-class Memsurfer(PythonPackage):
```
alecbcs, line 11: Memsurfer has been tested with `v3.9.12` ([README.md](https://github.com/LLNL/MemSurfer?tab=readme-ov-file#dependencies)) which it looks like we're keeping around. We should keep this package and modify the Python dependency.

repos/spack_repo/builtin/packages/libproxy/package.py
adamjstewart: I unfortunately don't know anything about or care about libproxy. The package has no maintainer, so I'm not sure who else we could ask to do this.

repos/spack_repo/builtin/packages/memsurfer/package.py
adamjstewart: @bhatiaharsh would you be interested in updating this package recipe to add a newer version that supports Python 3.9+? This can be done in a separate PR either before or after this PR is merged.

adamjstewart:
@spackbot maintainers

adamjstewart:
@spackbot run pipeline

adamjstewart:
@spackbot run pipeline

repos/spack_repo/builtin/packages/ceed/package.py
adamjstewart, line 1: @jedbrown @v-dobrev @tzanio

repos/spack_repo/builtin/packages/gdb/package.py
adamjstewart, line 1: @robertu94

repos/spack_repo/builtin/packages/mongodb/package.py
adamjstewart, line 1: @DaxLynch

repos/spack_repo/builtin/packages/node_js/package.py
adamjstewart, line 1: @cosmicexplorer

repos/spack_repo/builtin/packages/paraview/package.py
adamjstewart, line 1: @danlipsa @vicentebolea @kwryankrattiger

repos/spack_repo/builtin/packages/petsc/package.py
adamjstewart, line 1: @balay @jczhang07 @joseeroman

repos/spack_repo/builtin/packages/py_cylc_uiserver/package.py
adamjstewart, line 1: @LydDeb @climbfuji

repos/spack_repo/builtin/packages/py_dlio_profiler_py/package.py
adamjstewart, line 1: @hariharan-devarajan

repos/spack_repo/builtin/packages/py_graphene_tornado/package.py
adamjstewart, line 1: @LydDeb

repos/spack_repo/builtin/packages/py_haphpipe/package.py
adamjstewart, line 1: @dorton21

repos/spack_repo/builtin/packages/py_sierrapy/package.py
adamjstewart, line 1: @dorton21

repos/spack_repo/builtin/packages/slepc/package.py
adamjstewart, line 1: @joseeroman @balay

repos/spack_repo/builtin/packages/tau/package.py
adamjstewart, line 1: @wspear @eugeneswalker @khuck @sameershende

repos/spack_repo/builtin/packages/vtk/package.py
adamjstewart, line 1: @chuckatkins @danlipsa @johnwparent

repos/spack_repo/builtin/packages/warpx/package.py
adamjstewart, line 1: @ax3l @dpgrote @EZoni @RemiLehe

adamjstewart (commented):
Manually pinging maintainers of all non-Python packages or Python packages that would be deleted by this PR. Please review the changes to this PR to make sure I didn't make any mistakes. If one of your packages is being deleted by this PR, feel free to add newer versions that support Python 3.9+ in a separate PR.

adamjstewart:
@spackbot run pipeline

adamjstewart:
@spackbot run pipeline

repos/spack_repo/builtin/packages/ceed/package.py
v-dobrev: The changes in the `ceed` package look fine to me.

alalazo (commented):
Can we test this against `spack/spack` CI, to check that bootstrapping keeps working?

adamjstewart:
How? And why aren't those pipelines in spack/spack-packages?

alalazo:
> And why aren't those pipelines in spack/spack-packages?

I guess for the same reason we don't have yet maintainers pinged automatically, or pipelines running in `spack/spack` etc. :slightly_smiling_face: 

I'll put on my list to add bootstrapping tests here, but the bottom line is that we probably want to check there are no issues there before merging this PR.

adamjstewart:
Happy to wait, but note that there are several thousand more non-Python deprecations that I'm waiting to remove. The longer we wait, the more accidental inclusions of things that were deprecated after 1.0.0.

alalazo:
> Happy to wait, but note that there are several thousand more non-Python deprecations that I'm waiting to remove. The longer we wait, the more accidental inclusions of things that were deprecated after 1.0.0.

Bootstrap tests should be in, if you update the PR branch

adamjstewart:
@spackbot run pipeline

adamjstewart:
@alalazo bootstrap tests are all passing!

repos/spack_repo/builtin/packages/libproxy/package.py
becker33: `python@3.10:` is supported as of `libproxy@0.4.18`.

See changelog at https://github.com/libproxy/libproxy/releases/tag/0.4.18

repos/spack_repo/builtin/packages/memsurfer/package.py
becker33: It looks like memsurfer only has a lower bound on the python version, not an upper bound -- I don't think this package should be included in this PR.

repos/spack_repo/builtin/packages/py_tensorflow/package.py
```diff
@@ -260,16 +183,13 @@ class PyTensorflow(Package, CudaPackage, ROCmPackage, PythonExtension):
         depends_on("python@3.8:3.11", when="@2.12:2.13")
         depends_on("python@:3.10", when="@2.8:2.11")
         depends_on("python@:3.9", when="@2.5:2.7")
```
becker33:
```suggestion
        depends_on("python@:3.9", when="@:2.7")
```

repos/spack_repo/builtin/packages/py_pydv/package.py
```diff
@@ -1,26 +0,0 @@
-# Copyright Spack Project Developers. See COPYRIGHT file for details.
-#
-# SPDX-License-Identifier: (Apache-2.0 OR MIT)
-
-from spack_repo.builtin.build_systems.python import PythonPackage
-
-from spack.package import *
-
-
-class PyPydv(PythonPackage):
```
becker33, line 10: This package is under active development, can we leave it in? It looks at first glance like the only change is to add a newer version and switch the `py-pyside` dependency to `py-pyside2`.

repos/spack_repo/builtin/packages/libproxy/package.py
becker33, line 26: Could you change this to
```
    variant("python", default=False, description="Enable Python bindings", when="@0.4.18:")
```
and revert the rest of the changes in this package besides removing the `extends` directive?

That seems a nice middleground that you don't have to add to the burden of this PR by updating libproxy but we don't make our lives more difficult when we do add newer versions.

repos/spack_repo/builtin/packages/py_graphene_tornado/package.py
```diff
@@ -1,31 +0,0 @@
-# Copyright Spack Project Developers. See COPYRIGHT file for details.
-#
-# SPDX-License-Identifier: (Apache-2.0 OR MIT)
-
-from spack_repo.builtin.build_systems.python import PythonPackage
-
-from spack.package import *
-
-
-class PyGrapheneTornado(PythonPackage):
```
becker33, line 10: Considering that more recent versions of py-graphene-tornado have removed all version constraints on werkzeug, I think it's best if we do the same, and we can deprecate this package/version later if it proves not to work with any now buildable versions of werkzeug (and hopefully add new versions, but if it turns out no one cares, we can delete it).

repos/spack_repo/builtin/packages/py_cylc_uiserver/package.py
```diff
@@ -1,46 +0,0 @@
-# Copyright Spack Project Developers. See COPYRIGHT file for details.
-#
-# SPDX-License-Identifier: (Apache-2.0 OR MIT)
-
-from spack_repo.builtin.build_systems.python import PythonPackage
-
-from spack.package import *
-
-
-class PyCylcUiserver(PythonPackage):
```
becker33, line 10: AFAICT the only reason this is removed is the dependency on py-graphene-tornado. Given that and my comment on py-graphene-tornado, I think we should keep this package for now.

becker33 (changes requested):
I think this partial review covers all packages except:

- py-pymongo
- mongodb
- py-motor

I am still working through my thoughts on those -- I am leaning towards adding a commit to update two of those packages to ensure none of the three are deleted.

repos/spack_repo/builtin/packages/memsurfer/package.py
adamjstewart: Memsurfer was removed because it relies on `vtk@8.1.2+python`, which requires Python 3.8, which was removed. We would need to relax the VTK version requirement to keep it.

repos/spack_repo/builtin/packages/py_pydv/package.py
adamjstewart: Volunteers are welcome to add new versions of the package.

repos/spack_repo/builtin/packages/py_graphene_tornado/package.py
adamjstewart: Volunteers are welcome to add new versions of the package.

repos/spack_repo/builtin/packages/py_cylc_uiserver/package.py
adamjstewart: Volunteers are welcome to add new versions of the package.

repos/spack_repo/builtin/packages/py_graphene_tornado/package.py
becker33: I don't think this requires adding any new version -- I think the version constraint is not required, based on what I see from newer versions. I'll test that out.

becker33 (approved):

repos/spack_repo/builtin/packages/py_cylc_uiserver/package.py
climbfuji: @adamjstewart In the process of updating our spack-packages fork from upstream, I found out that this package was removed. I must have missed your notification last year in July. 

I am curious to know why this package was removed. It's of vital importance to the Navy. Was it removed because one of the dependencies was removed (py-graphene-tornado)?

We definitely need to restore this package and its dependencies.

alecbcs: @climbfuji this package was removed because it would have been totally broken by the deprecation and removal of `py-graphene-tornado`. (Which itself was removed because it had no available versions that could be built with an up-to-date Python.) If `py-cylc-uiserver` is of vital importance to you all, you need to be updating it regularly and not forcing that maintenance burden onto others.

climbfuji: Sure, and "putting it back" means adding a more recent version that works with the available Python utilities. One of the problems with this particular package py-cylc-uiserver is or at least was tight version constraints that made it hard to update or integrate with other packages.

I am mostly confused about deprecation rules. I once tried to remove a version without deprecating it first and was told to never do that - which makes sense. Here, not just a few versions but entire packages got removed without ever being deprecated first. Should the rule be "in order to remove deprecated versions of a package X, all its dependencies need to be deprecated before they can be removed"?

I wonder if we have the right tooling to identify dependencies so that a package maintainer can say "I want to deprecate X@m.n.r in a pull request, go and find all dependencies and mark them as deprecated in the same PR"? That would ease the maintenance burden.

climbfuji: Before I spend a huge amount of time on this: there is an entire ecosystem of Python packages that I am not a maintainer of that need updating. Some of these packages require Python <= 3.10. Does it make sense at all to add these packages in? Do you Spack maintainers have deprecation plans for Python 3.10 itself?

climbfuji: I found a bug in py-graphql_{core,relay} - an unnecessary dependency on an older version of the full py-poetry package when py-poetry-core is sufficient. This solved a lot of concretization issues and allows us to build py-cylc-uiserver@1.7.1 with Python <=3.11 and 1.9.1 with Python >=3.12. Will create a PR in the next days.

adamjstewart: > Do you Spack maintainers have deprecation plans for Python 3.10 itself?

Python 3.10 reaches [EOL](https://devguide.python.org/versions/) in October, so yes.

Glad we found a way to support newer Python!

> I am mostly confused about deprecation rules. I once tried to remove a version without deprecating it first and was told to never do that - which makes sense. Here, not just a few versions but entire packages got removed without ever being deprecated first. Should the rule be "in order to remove deprecated versions of a package X, all its dependencies need to be deprecated before they can be removed"?

If someone wants to install this package, they would have to install Python 3.9 first, which would give them a deprecation warning message. We either add 1 line deprecating Python 3.9, or thousands of lines deprecating every package/version that requires Python 3.9. I opted for the former until there is a policy in place that forces me to do the latter.

climbfuji: Understood about Python 3.9. FWIW, the latest version of py-cylc-uiserver that got removed built and ran fine with Python 3.11 in Spack. But no longer a problem. I'll create a PR shortly that brings back two more recent versions that work with 3.11 and 3.12. Removing a package entirely and then bringing it back has its pros. No need to deprecate old versions, start with a clean package recipe :-)

PR #664, opened 2025-07-20
xorg-server: avoid implicit LibraryList to str conversion warning

wdconinc:
This PR fixes `xorg-server` by passing `spec.libs.ld_flags` instead of `spec.libs`, and avoids the following warning:
```
==> Warning: /home/wdconinc/git/spack-packages/repos/spack_repo/builtin/packages/xorg_server/package.py:96: when setting environment variable GL_LIBS=/usr/lib/x86_64-linux-gnu/libGL.so: value is of type `LibraryList`, but `str` was expected. This is deprecated and will be an error in Spack v1.0
```

This potentially affects `grads +hdf4 ^hdf ~shared`  too.


alecbcs (approved):
Looks good to me. Thanks!

PR #406, opened 2025-06-30
VMC: fix deprecated ROOT variant dependency, add version 2-1

jezwilkinson:
Adds version 2-1 to VMC recipe: https://github.com/vmc-project/vmc/releases/tag/v2-1
Also fixes a concretization error with newer versions of ROOT (the "vmc" variant of ROOT is no longer defined as of version 6.26, so Spack fails to concretize with newer versions when "~vmc" is requested)

PR #405, opened 2025-06-30
vgm: add version 5-3-1

jezwilkinson:
Adds version 5-3-1 (along with 5-2 and 5-3) to spec for VGM: 
- https://github.com/vmc-project/vgm/releases/tag/v5-3-1 
- https://github.com/vmc-project/vgm/releases/tag/v5-3
- https://github.com/vmc-project/vgm/releases/tag/v5-2

PR #673, opened 2025-07-21
FairROOT: new package

jezwilkinson:
Adds package recipe for FairROOT, along with a "FairSoft-bundle" bundle package for the default software environment for FAIR software + FairROOT dependencies.

Note: Depends on a few still-open pull requests for missing versions of dependencies and fixing a handful of build/concretisation errors that were encountered in dependency packages during testing: https://github.com/spack/spack-packages/pull/405 (VGM), https://github.com/spack/spack-packages/pull/406 (VMC), https://github.com/spack/spack-packages/pull/407 (Geant4-VMC), https://github.com/spack/spack-packages/pull/409 (Geant3).

### DIFF
--- a/repos/spack_repo/builtin/packages/fairroot/package.py
+++ b/repos/spack_repo/builtin/packages/fairroot/package.py
@@ -28,8 +28,11 @@ class Fairroot(CMakePackage):
     depends_on('faircmakemodules@0.2:', when='@18:')
     depends_on('fairlogger@1.4.0:')
     depends_on('fairmq@1.4.11:')
+    
     depends_on('fairsoft-bundle')
-   
+    # Version-specific fairsoft release dependencies
+    depends_on('fairsoft-bundle@may25', when='@18.8.2:')
+
     depends_on('flatbuffers')
     depends_on('geant3', when="+sim")
     depends_on('geant4', when="+sim")
@@ -40,14 +43,13 @@ class Fairroot(CMakePackage):
     depends_on('pythia6', when='+sim')
     depends_on('pythia8', when='+sim')
     depends_on('root+http+xml+gdml')
-    depends_on('root@6.18:', when='@19:')
     depends_on('vgm', when="+sim")
     depends_on('vmc', when='@18.4: ^root@6.18:')
     depends_on('yaml-cpp', when='@18.2:')
     for std in ('17','20'):
         for dep in ('root', 'fairmq'):
             depends_on('{0} cxxstd={1}'.format(dep, std),
-                       when='cxxstd={0} ^{1}'.format(std, dep))
+                       when='cxxstd={0}'.format(std))
 
     def setup_build_environment(self, env):
         super(Fairroot, self).setup_build_environment(env)
--- a/repos/spack_repo/builtin/packages/fairroot/package.py
+++ b/repos/spack_repo/builtin/packages/fairroot/package.py
@@ -15,6 +15,7 @@ class Fairroot(CMakePackage):
     git = "https://github.com/FairRootGroup/FairRoot.git"
     maintainers("dennisklein", "fuhlig1", "jezwilkinson")
 
+    tags = ["hep"]
     version("develop", branch="dev")
     version("18.8.2", sha256="0bc9bafd9583f8a4c92977647c1eb360d66f45fbc6c81a15c5a1613640934684")
 
--- a/repos/spack_repo/builtin/packages/fairroot/package.py
+++ b/repos/spack_repo/builtin/packages/fairroot/package.py
@@ -1,0 +1,88 @@
+# Copyright Spack Project Developers. See COPYRIGHT file for details.
+#
+# SPDX-License-Identifier: (Apache-2.0 OR MIT)
+
+from spack_repo.builtin.build_systems.cmake import CMakePackage
+from spack.package import *
+
+class Fairroot(CMakePackage):
+    """C++ simulation, reconstruction and analysis framework for particle physics experiments"""
+
+    homepage = "http://fairroot.gsi.de"
+    url = "https://github.com/FairRootGroup/FairRoot/archive/v18.8.2.tar.gz"
+    git = "https://github.com/FairRootGroup/FairRoot.git"
+    maintainers('dennisklein', 'fuhlig1', 'jezwilkinson')
+
+    version('develop', branch='dev')
+    version('18.8.2', sha256='0bc9bafd9583f8a4c92977647c1eb360d66f45fbc6c81a15c5a1613640934684')
+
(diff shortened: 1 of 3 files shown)


wdconinc:
If you want, we can add this to the standard HEP stack that is built in CI. It's under `.ci/gitlab/stacks/hep`.

jezwilkinson:
@wdconinc Thanks for mentioning this! At the moment we're probably not "stable" enough yet to include in the CI stack (there is still a bit of testing we need to do before officially switching fairroot from the legacy build to spack), but it's definitely something we'll consider for the future.

repos/spack_repo/builtin/packages/fairsoft_bundle/package.py
```diff
@@ -0,0 +1,56 @@
...
+
+
+class FairsoftBundle(BundlePackage):
+    """Bundle package providing default environment for FAIR software"""
+
+    homepage = "https://github.com/FairRootGroup/FairSoft"
+    maintainers("dennisklein", "fuhlig1", "jezwilkinson")
+
+    # Releases:
+    version("may25")
```
wdconinc: Please use year-month ordering (or calver). Versions like this (i.e. no pattern) are sorted lexicographically, so you'll end up with weird upgrade/downgrade effects.

repos/spack_repo/builtin/packages/fairsoft_config/package.py
```diff
@@ -0,0 +1,35 @@
...
+
+class FairsoftConfig(CMakePackage):
+    """Legacy fairsoft-config script"""
+
+    homepage = "https://github.com/FairRootGroup/fairsoft-config"
+    git = "https://github.com/FairRootGroup/fairsoft-config"
+    maintainers("dennisklein", "fuhlig1", "jezwilkinson")
+
+    version("develop")
+    version("may25")
```
wdconinc: What do these versions mean? There is no download url, and there is no branch that is called this.

repos/spack_repo/builtin/packages/fairsoft_config/package.py
```diff
@@ -0,0 +1,35 @@
...
+        description="C++ standard reported",
+    )
+
+    depends_on("cmake@3:", type="build")
+    depends_on("root", type=("build", "link", "run"))
+
+    def cmake_args(self):
+        args = []
+        args.append("-DFAIRSOFT_VERSION=%s" % self.version)
+        args.append("-DCMAKE_CXX_STANDARD=%s" % self.spec.variants["cxxstd"].value)
```
wdconinc: Please use `self.define(var, val)` (and please use python f-strings, if you ever need them).

wdconinc: There is also `self.define_from_variant`.

repos/spack_repo/builtin/packages/fairroot/package.py
```diff
@@ -0,0 +1,92 @@
...
+        options.append("-DBUILD_EXAMPLES:BOOL=%s" % ("ON" if "+examples" in self.spec else "OFF"))
+
+        if self.spec.satisfies("^boost@:1.69.99"):
+            options.append("-DBoost_NO_BOOST_CMAKE=ON")
+        options.append("-DBUILD_PROOF_SUPPORT=OFF")
+        return options
+
+    def common_env_setup(self, env):
+        # So that root finds the shared library / rootmap
+        env.prepend_path("LD_LIBRARY_PATH", self.prefix.lib)
```
wdconinc: Please avoid using `LD_LIBRARY_PATH`, but use the newer `ROOT_LIBRARY_PATH` (which has fewer side effects on other shared object library loads). In https://github.com/spack/spack-packages/blob/develop/repos/spack_repo/builtin/packages/root/package.py you can find a `root_library_path` function that returns the preferred variable for newer versions.

repos/spack_repo/builtin/packages/fairroot/package.py
```diff
@@ -0,0 +1,92 @@
...
+
+    def setup_run_environment(self, env):
+        self.common_env_setup(env)
+
+    def setup_dependent_build_environment(self, env, dependent_spec):
+        self.common_env_setup(env)
+
+    def setup_dependent_run_environment(self, env, dependent_spec):
+        self.common_env_setup(env)
```
wdconinc: There is some redundancy here, since some of these functions will always be called when the other is called. E.g. `setup_dependent_run_environment` is always called when `setup_dependent_build_environment` is called. 

See:
https://github.com/spack/spack/blob/2392f747559d3a69f795882728af4af4a8d8e795/lib/spack/spack/modules/common.py#L677-L680
https://github.com/spack/spack/blob/2392f747559d3a69f795882728af4af4a8d8e795/lib/spack/spack/build_environment.py#L1055-L1082

repos/spack_repo/builtin/packages/fairsoft_config/package.py
jezwilkinson: The version names are the release names of the FairSoft repo, and are passed as a CMake option to fairsoft-config so that it reports the installed version of the bundle (needed for legacy reasons by some dependent packages e.g. CbmRoot, in the form 'may25' ). But I can re-jig this to be a variant in fairsoft-config so that we avoid the version sorting issues mentioned above

repos/spack_repo/builtin/packages/fairsoft_bundle/package.py
jezwilkinson: switched to "2025-05" as the scheme (the ref to "may25" is still eventually needed by some dependents but this is handled in the variants for fairsoft-config now)

jezwilkinson:
Thanks for the comments, i've implemented all of them and hopefully this should be OK now.
Also thank you for following up on the dependency commits - I think it's only the version bump for Geant3 that's still missing the approval https://github.com/spack/spack-packages/pull/409

repos/spack_repo/builtin/packages/fairroot/package.py
```diff
@@ -0,0 +1,95 @@
...
+
+
+class Fairroot(CMakePackage):
+    """C++ simulation, reconstruction and analysis framework for particle physics experiments"""
+
+    homepage = "http://fairroot.gsi.de"
+    url = "https://github.com/FairRootGroup/FairRoot/archive/v18.8.2.tar.gz"
+    git = "https://github.com/FairRootGroup/FairRoot.git"
+    maintainers("dennisklein", "fuhlig1", "jezwilkinson")
+
```
vvolkl, line 17:
```suggestion

    tags = ["hep"]
```
Adding a tag is a nice way to keep an overview of HEP packages in spack.

vvolkl (approved):

PR #674, opened 2025-07-21
py-jaxlib: fix runtime error due to custom cuda location

AdhocMan:
Jaxlib with CUDA throws a runtime error if CUDA is installed at a custom location.
Setting `LD_LIBRARY_PATH` such that CUPTI can be found and `XLA_FLAGS` for XLA to find CUDA fixes the error.


adamjstewart:
A bit hesitant because I'm not sure if CUPTI is always in this location on all platforms and all external CUDA installations. What is XLA_FLAGS actually used for at run-time?

AdhocMan:
> A bit hesitant because I'm not sure if CUPTI is always in this location on all platforms and all external CUDA installations. What is XLA_FLAGS actually used for at run-time?

I took the `likwid` package as an example and made the CUPTI location detection more portable.
The XLA flag is required for the Jit compilation step. I think mainly for libdevice, but possibly for other libraries as well.

adamjstewart (approved):
Looks better now, thanks! As you can tell by the commented-out code, I tried getting external CUDA working during build time but couldn't make XLA happy. Feel free to investigate that if you get bored. Ideally, we would link to the CUDA libraries at build time so that these flags aren't needed.

PR #675, opened 2025-07-21
py-flash-attn: add v2.8.1

thomas-bouvier:


wdconinc (approved):
Looks good to me based on setup.py diff.

PR #677, opened 2025-07-21
py-numba: fix python dependency bounds

thomas-bouvier:
This patch corrects inconsistencies with Numba's Python dependencies as seen in:
- https://github.com/numba/numba/blob/release0.59/setup.py
- https://github.com/numba/numba/blob/release0.61/setup.py
- https://github.com/numba/llvmlite/blob/v0.40.0/setup.py
- https://github.com/numba/llvmlite/blob/v0.42.0/setup.py
- https://github.com/numba/llvmlite/blob/v0.44.0/setup.py


wdconinc (approved):
Thanks!

PR #682, opened 2025-07-21
pexsi: add pic option

mkrack:


mkrack:
@haampie Could you review this minor change which is needed to build CP2K with pexsi support using spack? This would allow to drop the [local spack recipe](https://github.com/cp2k/cp2k/tree/master/tools/spack/cp2k_dev_repo/packages/pexsi) in the CP2K project.

bernhardkaindl (approved):

PR #685, opened 2025-07-21
amrex: add new variant gpu_rdc

### LINKED ISSUE #421: Installation issue: WarpX w/ CUDA due to AMReX GPU RDC Variant

https://github.com/BLAST-WarpX/warpx/issues/5976

### Error message

<details><summary>Error message</summary>

<pre>
[ 94%] Linking CUDA device code CMakeFiles/lib_3d.dir/cmake_device_link.o
/spack/opt/spack/linux-cascadelake/cmake-3.31.6-seapj3dboyb5ezkbnivefkx7siviqvmb/bin/cmake -E cmake_link_script CMake     Files/lib_3d.dir/dlink.txt --verbose=1
nvlink error   : Entry function '_ZN5amrex13launch_globalILi256EZNS_14ParallelForRNGIiZNS_15filterParticlesINS_12ParticleTileINS_11So     AParticleILi7ELi0EEELi7ELi0ENS_20PinnedArenaAllocatorEEENS3_IS5_Li7ELi0ENS_14ArenaAllocatorEEEZN16WarpXOpenPMDPlot21WriteOpenPMDParti     clesERKNS_6VectorI12ParticleDiagSaISC_EEEdbbbEUlRKNS_21ConstParticleTileDataINS_15SoAParticleBaseELi7ELi0EEEiRKNS_12RandomEngineEE0_i     iLi0EEET2_RT_RKT0_RKT1_SQ_SQ_T3_EUliSO_E_vEENSt9enable_ifIXsr5amrex19MaybeDeviceRunnableIST_vEE5valueEvE4typeESR_SV_EUlvE_EEvST_' 
use     s too much data for compiler-generated constants; please recompile with -Xptxas --disable-optimizer-constants
</pre></details>

C++ compiler: GNU 13.3 with NVCC 12.6, V12.6.85

specs:
- `warpx+python compute=cuda ~eb %py-amrex+cuda cuda_arch=80 %amrex cuda_arch=80`
- `warpx+python compute=cuda dim=2,3,rz %py-amrex+cuda cuda_arch=80 %amrex cuda_arch=80`

### Additional information

We need to add a variant for gpu rdc control in the AMReX package.
I would even suggest to turn it off by default, because it can complicate compiles and is only needed by few codes.
https://github.com/BLAST-WarpX/warpx/issues/5976

In the warpx package, we should only require gpu rdc if the ascent or sensei variants are requested.

### General information

WeiqunZhang:
The default of the new gpu_rdc variant is true, because it is enabled by default for CUDA and HIP builds in AMReX. However, users can now disable it.

Close #421

### DIFF
--- a/repos/spack_repo/builtin/packages/amrex/package.py
+++ b/repos/spack_repo/builtin/packages/amrex/package.py
@@ -156,6 +156,10 @@ class Amrex(CMakePackage, CudaPackage, ROCmPackage):
     variant("sundials", default=False, description="Enable SUNDIALS interfaces")
     variant("pic", default=False, description="Enable PIC")
     variant("sycl", default=False, description="Enable SYCL backend")
+    variant("gpu_rdc", default=True, description="Enable relocatable GPU device code support",
+            when="+cuda")
+    variant("gpu_rdc", default=True, description="Enable relocatable GPU device code support",
+            when="+rocm")
 
     # Build dependencies
     depends_on("c", type="build")  # generated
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/amrex/package.py
```diff
@@ -156,6 +156,18 @@ class Amrex(CMakePackage, CudaPackage, ROCmPackage):
...
     variant("sycl", default=False, description="Enable SYCL backend")
+    variant(
+        "gpu_rdc",
+        default=True,
+        description="Enable relocatable GPU device code support",
+        when="+cuda",
+    )
+    variant(
+        "gpu_rdc",
+        default=True,
```
ax3l, line 167: I know it is on by default in AMReX configs, but I feel turning it off in Spack defaults would be better for the following reason:

RDC is needed, then a package can explicitly request it.
RDC definitely does not work, then a package can explicitly set it false.

RDC does not matter but generally is not needed, a package can keep the default and reuse what is there.

A `false` default will by default save compile & link-time and be more robust because we know RDC is often buggy.

ax3l: Maybe one can make the same argument the other way around...

WeiqunZhang: I would argue it's more robust to make it true by default, because we know some codes depend on it.

WeiqunZhang: Furthermore, if the default is true, we are pretty sure this PR will not cause any new issues. If the default is false, there is a chance of breaking someone's workflow.

ax3l (approved):

ax3l:
We should as a follow up update the WarpX package and set gpu rdc to off unless in situ Vis is used

Issue #328, opened 2025-06-20
Installation issue: tk


jhodrien:
### Steps to reproduce the issue

Tested on a system that doesn't have /usr/include/X11/X.h installed as it falls back to using the system installed header if it is.

```console
$ spack install tk
```

### Error message

<details><summary>Error message</summary>

<pre>
$ spack install tk
[+] /usr (external glibc-2.34-qsgdwegywjttuv6n75awwzuiow6lxnpw)
[+] /local/data/user/spack/opt/spack/linux-skylake/kbproto-1.0.7-r537h7ghq5z7itzya2szjznbamf7cn5m
[+] /local/data/user/spack/opt/spack/linux-skylake/libpthread-stubs-0.5-gaizo2g4me4tf5gplxscpiccea5k6myi
[+] /local/data/user/spack/opt/spack/linux-skylake/xproto-7.0.31-paargqgsp5nmxztz6ihk7oyq5rli7cxi
[+] /local/data/user/spack/opt/spack/linux-skylake/renderproto-0.11.1-lflwjcqcve7zfkngve6cbpfm7tsuzprk
[+] /local/data/user/spack/opt/spack/linux-skylake/xextproto-7.3.0-tcvvn7txo555ldtiqjo6clkr5psdtkkl
[+] /local/data/user/spack/opt/spack/linux-skylake/scrnsaverproto-1.2.2-agcsn4ngqhm7l45bio6fwc5cbwmimpeq
[+] /local/data/user/spack/opt/spack/linux-skylake/gcc-runtime-11.5.0-tbatj2kst2jtrbqrliy7tnhbrtrbie4y
[+] /local/data/user/spack/opt/spack/linux-skylake/font-util-1.4.1-wraadhgtoa3vazpsnnqwcyd3zecbp4rf
[+] /local/data/user/spack/opt/spack/linux-skylake/libiconv-1.17-ef2lb4m3o32tg4sbdyliujiiffesdhfi
[+] /local/data/user/spack/opt/spack/linux-skylake/libmd-1.1.0-d5crfxgnnh6lepafkjbgs3ffzdu5n2ye
[+] /local/data/user/spack/opt/spack/linux-skylake/xtrans-1.5.2-o2imgotdy6bkuj6zfwxr2oplypku7fa7
[+] /local/data/user/spack/opt/spack/linux-skylake/libxau-1.0.12-exmmji3z2nh4rfg6ryj2pbohnntetgfu
[...]

PR #692, opened 2025-07-21
tk: needs c compiler to build

vanderwb:
When compiler language dependencies were auto-added, the `tk` package did not get a C lang dependence, probably because of the source code layout. But this is required. Without it, the configure script picks up on the system GCC:

```
...
checking for gcc... gcc
...
```

and then since it isn't using the Spack wrapper, dependencies like libX11 aren't found properly.

This PR addresses this and should fix issue #328.


wdconinc (approved):
Thanks. Looks good to me.

PR #694, opened 2025-07-21
libdrm: version is property of spec

vanderwb:
Perhaps this worked at one point, but in Spack vlatest this gives an error as `version` is not a valid property of `self`.

This PR corrects it to pull the version from the `spec` object of `self`.


wdconinc (approved):
Thanks!

PR #561, opened 2025-07-11
GDAL: add v3.11.1-3, fix +arrow build

adamjstewart:
https://github.com/OSGeo/gdal/releases/tag/v3.11.2

PR #696, opened 2025-07-21
gdal: parquet support needs arrow with parquet and fs support

vanderwb:
The GDAL support for the parquet format seems to require not just `arrow`, but `arrow` with filesystem and parquet support:

```
-- Could NOT find Parquet (missing: Parquet_DIR)
CMake Error at cmake/helpers/CheckDependentLibrariesCommon.cmake:213 (message):
  Configured to use PARQUET, but not found
Call Stack (most recent call first):
  cmake/helpers/CheckDependentLibrariesArrowParquet.cmake:3 (gdal_check_package)
  cmake/helpers/CheckDependentLibraries.cmake:493 (include)
  gdal.cmake:76 (include)
  CMakeLists.txt:264 (include)
```

And the CMake logic in GDAL:

```
if (Arrow_FOUND)
    gdal_check_package(Parquet "Apache Parquet C++ library" CONFIG PATHS ${Arrow_DIR} CAN_DISABLE)
...
```

This PR ensures that parquet support is enabled in arrow, which resolves these errors.

### DIFF
--- a/repos/spack_repo/builtin/packages/gdal/package.py
+++ b/repos/spack_repo/builtin/packages/gdal/package.py
@@ -384,7 +384,7 @@ class Gdal(CMakePackage, AutotoolsPackage, PythonExtension):
     depends_on("openjpeg", when="+openjpeg")
     depends_on("openssl", when="+openssl")
     depends_on("oracle-instant-client", when="+oracle")
-    depends_on("arrow", when="+parquet")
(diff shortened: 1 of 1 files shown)


haampie (approved):

adamjstewart (approved):
Thanks for fixing this! I noticed a similar problem with `+arrow` and fixed it in #561 but forgot about `+parquet`. Confirmed that this works and both variants are needed to build correctly.

PR #697, opened 2025-07-21
 exmcutils: add v0.6.4, add maintainer, use autoreconf()

j-woz:


tldahlgren (approved):
Check confirmed the sha256.

PR #698, opened 2025-07-21
spherepack: fix to avoid flags nvhpc doesnt understand

vanderwb:
The current version of the package injects flags for GCC 10+ into the Makefile via environment variables, but a better approach is to use a flag handler to avoid conflicts with compilers that don't support those flags (nvhpc). This PR implements this.

### DIFF
--- a/repos/spack_repo/builtin/packages/spherepack/package.py
+++ b/repos/spack_repo/builtin/packages/spherepack/package.py
@@ -19,7 +19,18 @@ class Spherepack(Package):
     depends_on("fortran", type="build")
     depends_on("gmake", type="build")
 
+    def flag_handler(self, name, flags):
+        spec = self.spec
+
+        if name == "fflags":
+            flags.append("-O2")
+
+            if spec.satisfies("%gcc@10:"):
+                flags.append("-fallow-argument-mismatch")
+
+        return (flags, None, None)
+
     def install(self, spec, prefix):
-        make("MAKE=make", "F90=f90 -O2 -fallow-argument-mismatch", "AR=ar", "libspherepack")
-        make("MAKE=make", "F90=f90 -O2 -fallow-argument-mismatch", "AR=ar", "testspherepack")
+        make("MAKE=make", "F90=f90", "AR=ar", "libspherepack")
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/spherepack/package.py
```diff
@@ -19,7 +19,18 @@ class Spherepack(Package):
     depends_on("fortran", type="build")
     depends_on("gmake", type="build")
 
+    def flag_handler(self, name, flags):
+        spec = self.spec
+
+        if name == "fflags":
+            flags.append("-O2")
+
+            if spec.satisfies("%gcc@10:"):
```
haampie:
```suggestion
            if spec.satisfies("%fortran=gcc@10:"):
```

otherwise the branch is true too if only `%c=gcc`.

repos/spack_repo/builtin/packages/spherepack/package.py
```diff
@@ -19,7 +19,18 @@ class Spherepack(Package):
     depends_on("fortran", type="build")
     depends_on("gmake", type="build")
 
+    def flag_handler(self, name, flags):
+        spec = self.spec
+
+        if name == "fflags":
+            flags.append("-O2")
```
haampie:
```suggestion
```

is redundant if you inject flags via the compiler wrapper and the package has a sensible default.

repos/spack_repo/builtin/packages/spherepack/package.py
vanderwb: Resolved by commit [cefea84](https://github.com/spack/spack-packages/pull/698/commits/cefea84cab1f8007c20e42a826e3afc3b2769bc1)

repos/spack_repo/builtin/packages/spherepack/package.py
vanderwb: Resolved by [cefea84](https://github.com/spack/spack-packages/pull/698/commits/cefea84cab1f8007c20e42a826e3afc3b2769bc1)

haampie (approved):

PR #699, opened 2025-07-21
Generalize Intel logic to support package names too

vanderwb:
The ESMF package needs to determine compiler family based on the compiler name, but instead of giving `oneapi`, the compiler.name field now gives `intel-oneapi-compilers`. In the interest of future-proofing, I generalized the logic a bit to detect any compiler with `intel` or `oneapi` in the name to be of the Intel family, but specific new packages names could be used instead if desired.


haampie:
I would do literal comparisons instead of substring comparison, cause that is easier to grep.

vanderwb:
Changed in [7cd58f8](https://github.com/spack/spack-packages/pull/699/commits/7cd58f8297cac7ccc1db5ffb45dd9b14c051ba71)

haampie (approved):

PR #701, opened 2025-07-22
py-numpy: fix support for C++23 compilers

adamjstewart:
Discovered by #38 

TF 2.18 requires numpy 2.0 or older, so we have no choice but to backport this patch. TF 2.19+ adds support for newer numpy.


rgommers (approved):
Thanks @adamjstewart. The backport fix went into v2.1.2, so this looks right to me.

tldahlgren (approved):

PR #707, opened 2025-07-22
chapel: update `is_system_path` import

### LINKED ISSUE #702: Chapel produces warning due to use of deprecated feature.
Doesn't seem to fit any category of ticket so using blank issue.

I get the following warning when installing Chapel but seems to install fine.

```
==> Warning: ./repos/spack_repo/builtin/packages/chapel/package.py:681: spack.package.is_system_path is deprecated
==> Warning: ./repos/spack_repo/builtin/packages/chapel/package.py:685: spack.package.is_system_path is deprecated
==> Warning: ./repos/spack_repo/builtin/packages/chapel/package.py:681: spack.package.is_system_path is deprecated
==> Warning: ./repos/spack_repo/builtin/packages/chapel/package.py:685: spack.package.is_system_path is deprecated
==> Warning: ./repos/spack_repo/builtin/packages/chapel/package.py:681: spack.package.is_system_path is deprecated
==> Warning: ./repos/spack_repo/builtin/packages/chapel/package.py:685: spack.package.is_system_path is deprecated
```

Maintainers are listed as:
@arezaii @bonachea @arifthpe

arifthpe:
Switch chapel package.py from using deprecated `spack.package.is_system_path` to its new location `spack.util.environment.is_system_path` (moved in https://github.com/spack/spack/pull/50995).

Resolves https://github.com/spack/spack-packages/issues/702.

### DIFF
--- a/repos/spack_repo/builtin/packages/chapel/package.py
+++ b/repos/spack_repo/builtin/packages/chapel/package.py
@@ -13,6 +13,7 @@ from spack_repo.builtin.build_systems.rocm import ROCmPackage
 from spack.package import *
 from spack.util.environment import is_system_path
 
+
 def slingshot_network():
     return os.path.exists("/opt/cray/pe") and (
         os.path.exists("/lib64/libcxi.so") or os.path.exists("/usr/lib64/libcxi.so")
--- a/repos/spack_repo/builtin/packages/chapel/package.py
+++ b/repos/spack_repo/builtin/packages/chapel/package.py
@@ -11,7 +11,7 @@ from spack_repo.builtin.build_systems.cuda import CudaPackage
 from spack_repo.builtin.build_systems.rocm import ROCmPackage
 
 from spack.package import *
-
+from spack.util.environment import is_system_path
 
 def slingshot_network():
     return os.path.exists("/opt/cray/pe") and (


haampie:
Actually it's not moved, but I wanted to deprecate the functionality entirely because it's technically incorrect.

The deprecation warning is just in `spack.package` because Spack still uses the function internally and we didn't have time to fix that before the 1.0 release.

I can be convinced to remove the deprecation message again if it's too noisy.

The issue with `is_system_path` is that it's matching paths against a static list. What we should have instead is a way to query a certain application if a given path is among its default search paths. Something like `self["c"].is_default_include_path(...)` or `self["c"].is_default_link_path(...)`.

It's not an issue in most cases, but on CernVM-FS or Gentoo Prefix the `is_system_path` function is simply incorrect. There you have a `<sysroot>/usr/{bin,lib,include}` structure.

arifthpe:
Okay, thanks for the info. I'm not bothered by the deprecation message. I think it'd be fine to keep calling the deprecated function until a better way to check against default search paths exists, and switch to using that then.

My motivation for this PR was https://github.com/spack/spack-packages/issues/702 about the deprecation warning noise when building Chapel, though that isn't specific to just Chapel.

PR #708, opened 2025-07-22
R: A few package fixes for  gcc@14

Chrismarsh:
[In GCC 14:](https://gcc.gnu.org/gcc-14/porting_to.html)

```
[Implicit function declarations (-Werror=implicit-function-declaration)](https://gcc.gnu.org/gcc-14/porting_to.html#implicit-function-declaration)
It is no longer possible to call a function that has not been declared.
```

This change in default results in a variety of R packages failing with errors like:
```
rle.c:221:9: error: implicit declaration of function 'Calloc'; did you mean 'calloc'? [-Wimplicit-function-declaration]
  221 |   val = Calloc(n2, int);
      |         ^~~~~~
      |         calloc
```

This updates the affected packages to newer versions which have the changes required for gcc 14+. I don't think it's necessary to conflict directly on gcc@14

### DIFF
--- a/repos/spack_repo/builtin/packages/r_bit/package.py
+++ b/repos/spack_repo/builtin/packages/r_bit/package.py
@@ -19,6 +19,7 @@ class RBit(RPackage):
 
     license("GPL-2.0-only OR GPL-3.0-only")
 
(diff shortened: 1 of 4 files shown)


tldahlgren (approved):
Version sha256s have been confirmed.

PR #710, opened 2025-07-22
py-rpy2: add v3.6.1

Chrismarsh:
- Add a new version of rpy2
- Add patch to fix the dual specified license that violates PEP-0621
- This is reported upstream https://github.com/rpy2/rpy2/issues/1198


Chrismarsh:
@spackbot fix style

PR #712, opened 2025-07-22
unzip: fix built for gcc@15 and add security patches

manuelakuhn:
This fixes building with gcc@15.

While at it I also added security patches solving various CVEs.


manuelakuhn:
@spackbot run pipeline

bernhardkaindl (approved):
Built with gcc-13 on Ubuntu

PR #713, opened 2025-07-22
lammps: add v20250722, v20240829.4, deprecate patch releases

rbberger:
- adds new stable release `20250722`
- adds last update to previous stable release `20240829.4`
- the patch released between stable releases are marked deprecated


tldahlgren (approved):
Sha256s have been confirmed.

PR #720, opened 2025-07-22
neovim: fix build and update versions

keltonhalbert:
Building Neovim is broken with Spack 1.0. I has been a year or so since I built it, so I cannot pinpoint the exact causes, but this should address multiple issues. 

1. The way the `package.py` for Neovim handled dependencies was implicitly overriding "when" constraints in the "with" blocks -- resulting in the inclusion of dependencies that are either no longer required, or are vendored with Neovim itself, when concretizing the spec. For example, `libtermkey` was being included in `0.11.1` when it should not. The additions I made successfully prevent these from being required when installing modern versions, i.e. >= 0.11.0.
2. The unibilium dependency was broken, causing builds of Neovim to fail. It has something to do with autoreconf: ``` configure: error: cannot find install-sh, install.sh, or shtool in "." "./.." "./../.."```. I am by no means experienced in auto[re]conf, and I could not figure out how to patch it and get it to work. However, unibilium provides a CMakeLists.txt -- so I updated the `package.py` to accept cmake builds, and made it the default. I am happy to amend this PR if someone else can figure out how to fix the autotools build.

### DIFF
--- a/repos/spack_repo/builtin/packages/neovim/package.py
+++ b/repos/spack_repo/builtin/packages/neovim/package.py
@@ -125,7 +125,7 @@ class Neovim(CMakePackage):
     # versions
     with when("@0.4:"):
         depends_on("libuv@1.28:", type="link")
-        depends_on("libluv@1.30.0:", type="link", when="@:0.9")
+        depends_on("libluv@1.30.0:", type="link")
         depends_on("libtermkey@0.18:", type="link", when="@:0.9")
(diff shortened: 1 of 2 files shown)


keltonhalbert:
Sorry about that. Goofed on a few minor things caught by the CI. This should be good now, barring any changes needed to the autotools build of unibilium.

tldahlgren (commented):
The version check confirmed the new sha256s.

trws (approved):

PR #721, opened 2025-07-23
unittest_cpp: add c dependency

NadeemKever:
add compiler dependency for unittest_cpp


alecbcs (approved):
I can confirm that the autotools configure seems to check for a C compiler in addition to a C++ compiler. Looks good to me. Thanks @NadeemKever.

PR #731, opened 2025-07-23
python-venv: memoize sysconfig call

haampie:
This is called for every edge `py-*` to `python-venv` in `setup_dependent_package`, which is excessive.


alecbcs (approved):
Nice! Thanks @haampie

PR #735, opened 2025-07-23
py-bidskit: add v2025.1.30

manuelakuhn:
https://pypi.org/project/bidskit/2025.1.30/#files

### DIFF
--- a/repos/spack_repo/builtin/packages/py_bidskit/package.py
+++ b/repos/spack_repo/builtin/packages/py_bidskit/package.py
@@ -30,7 +30,7 @@ class PyBidskit(PythonPackage):
     depends_on("py-setuptools@72.1:", type=("build", "run"), when="@2025.1.30:")
     depends_on("py-setuptools", type="build")
 
-    depends_on("py-pydicom@2.3:", type=("build", "run"), when="@2025.1.30:")
+    depends_on("py-pydicom@2.3.1:", type=("build", "run"), when="@2025.1.30:")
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/py_bidskit/package.py
```diff
@@ -15,18 +15,28 @@ class PyBidskit(PythonPackage):
...
 
+    depends_on("python@3.10:3", type=("build", "run"), when="@2024.11.7:")
     depends_on("python@3.7:3", type=("build", "run"))
+
+    # requirements.txt
+    # ignore upper version limits, because it works with newer versions as well
+    depends_on("py-setuptools@72.1:", type=("build", "run"), when="@2025.1.30:")
     depends_on("py-setuptools", type="build")
 
+    depends_on("py-pydicom@2.3:", type=("build", "run"), when="@2025.1.30:")
```
tldahlgren: Shouldn't this be `@2.3.1:`?

tldahlgren (commented):
Confirmed the version sha256 and dependencies, though have a question about one and, judging from an `import bids` in `bitskit/fmaps.py` I can see why `py-pybids` is needed.

repos/spack_repo/builtin/packages/py_bidskit/package.py
manuelakuhn: Yes, you are right. This is fixed now. Thanks.

tldahlgren (approved):

PR #69, opened 2025-06-06
netcdf-c: add v4.9.3

wdconinc:
> [!NOTE]
> This PR was migrated from **https://github.com/spack/spack/pull/49331**.

This PR adds `netcdf-c`, v4.9.3 ([diff](https://github.com/Unidata/netcdf-c/compare/v4.9.2...v4.9.3)).

TODO:
- [x] rename CMake options

PR #736, opened 2025-07-23
netcdf-c: prepare the Autotools builder for version 4.9.3

skosukhin:
With these changes, we can add version 4.9.3 (see #69) and expect the package to build correctly when `build_system=autotools`.

### DIFF
--- a/repos/spack_repo/builtin/packages/netcdf_c/package.py
+++ b/repos/spack_repo/builtin/packages/netcdf_c/package.py
@@ -392,8 +392,12 @@ class AutotoolsBuilder(AnyBuilder, autotools.AutotoolsBuilder):
             "--enable-utilities",
             "--enable-static",
             "--enable-largefile",
-            "--enable-netcdf-4",
         ]
+
+        if self.spec.satisfies("@4.8.0:"):
+            config_args.append("--enable-hdf5")
+        else:
+            config_args.append("--enable-netcdf-4")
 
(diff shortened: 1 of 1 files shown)


wdconinc (approved):
This looks good to me. I'm assuming this will also get tested in the CI run (albeit without the 4.9.3).

repos/spack_repo/builtin/packages/netcdf_c/package.py
```diff
@@ -263,6 +263,12 @@ class NetcdfC(CMakePackage, AutotoolsPackage):
     # Byte-range I/O was added in version 4.7.0:
     conflicts("+byterange", when="@:4.6")
 
+    # Byte-range requires DAP starting version 4.9.3:
+    requires("+dap", when="@4.9.3:+byterange")
+
+    # JNA was added in 4.3.2 and removed in 4.9.3:
+    conflicts("+jna", when="@:4.3.1,4.9.3:")
```
alecbcs, line 270:
```suggestion
    # JNA was added in 4.3.2 and removed in 4.9.3:
    conflicts("+jna", when="@:4.3.2,4.9.2:")
```

Version ranges in Spack are inclusive, so we should modify the directive like so to mirror the comment above.

alecbcs (changes requested):
Edit: nevermind I missread one of the `conflicts()` directives. This looks good to me.

alecbcs (approved):

PR #739, opened 2025-07-23
curl: add 8.14.1 and 8.15.0

tgamblin:
Update to the latest 8.14.1 and 8.15.0 releases.

`cmake` has some issues with `curl` 8.14.1 and `8.15` -- specifically, `cmake` 3.11:3.31.6` need a patch, and older versions won't work with these `curl` versions.

`curl` 8.15 removes support for `secure_transport` TLS on macOS, which was our default TLS version. So, we need to make it rely on `openssl`. This is not too disruptive, as everyone is either installing `openssl` or pointing at an external one anyway.


haampie:
> We should probably put this in a 2015.07.1 release.

On the one hand good to have the fix backported, on the other hand people will probably complain about deprecated=True if they already have some spack.lock with that version.

tgamblin:
I feel like we have the flexibility to deal with this now.  If they need the exact thing they had, they can do:

```yaml
spack:
  repos:
    builtin:
      tag: v2025.07.0
```

If they need the fixes, they can keep what's in spack 1.0, which is basically:

```yaml
spack:
  repos:
    builtin:
      branch: releases/v2025.07
```

Which will pull in the fixes.  I think there is a case for non-ABI breaking changes to go into a release branch like this.  Curl is very stable version to version, so think this is fair game. If this was a bigger update (e.g. a major version change) I'd have more qualms.

This *will* cause some rebuilds, as there are a fair number of things that depend on curl, but I think that should be ok for CVEs.

tgamblin:
Also we have not historically done this but we probably should have better CVE scanning/updates for software in our mainline, so that future package releases would have more longevity than `v2025.07.0`. That is something we need to automate... we likely need the NVD numbers for these things embedded in packages to really make it work.

tgamblin:
8.15 removed secure transport support on macOS — working on that.

tgamblin:
ok should be ready.

tgamblin:
Screenshot 2025-07-24 at 11 21 20 PM

Sigh. Apparently cmake breaks with curl 8.15.0 due to a curl type it was using.  This is [fixed in 4.0.0 and backported to 3.31.7](https://www.mail-archive.com/debian-bugs-dist@lists.debian.org/msg2028391.html). However, we have a 3.25 in the pipeline. 

I could either put the correct constraint in to indicate where it's fixed, or I could see if the patch applies to 3.25.3.  The former seems better.

Thinking I should split this up into two PRs -- one to add 8.14.1 (which will build without issue and can more easily be back ported to 2025.07), and one that adds the various changes to variants and cmake required for curl 8.15.

tgamblin:
Ok bailing on this as a back port -- tried something less disruptive in #792, which should go into 2025.07.1 instead of this.

We can get 8.15 working separately, with the changes it requires in cmake and maybe other packages.

tgamblin:
@haampie this should be ready now.  See also: https://github.com/curl/curl/issues/18088.

Summary of integration issues:
*  8.15.0 depends on perl at build time (with autotools) due to a bug.
* 8.15.0 removes secure transport on macos.
* Both 8.14.1 and 8.15.0 require cmake patches.

repos/spack_repo/builtin/packages/curl/package.py
```diff
@@ -86,27 +92,33 @@ class Curl(NMakePackage, AutotoolsPackage):
...
+
+    # TODO: add dependencies for other possible TLS backends
+
+    variant("tls", default="openssl", description=tls_desc, values=tls_vals, multi=True)
+    with when("platform=windows"):
+        variant("tls", default="sspi", description=tls_desc, values=tls_vals, multi=True)
+    with when("platform=darwin @:8.14"):
+        variant(
+            "tls", default="secure_transport", description=tls_desc, values=tls_vals, multi=True
+        )
```
alalazo: If we set defaults here, in terms of conditional variants, we can remove the `default_tls` definition above

alalazo (commented):
LGTM, only one minor comment

repos/spack_repo/builtin/packages/curl/package.py
tgamblin: Yeah good point. I tried to simplify this further and make it easier to read -- see what you think.

alalazo (approved):

alalazo:
Restarted pipelines manually. They were :x: because of system failures.

tgamblin:
Finally 🎉

PR #740, opened 2025-07-23
metpy + pint: updates for numpy@2

Chrismarsh:
This updates the metpy and pint stack to work with numpy@2:
- Add `py-pint` 0.24.4
- Add `py-metpy` 1.7.0
- Fix numpy version constraints
- Add new variants to `py-pint` to correct capture enabling numpy/xarray/dask to ensure correct versions (implicit before)
- Add `py-flexcache` and `py-flexparser` which are required for the new `py-pint`
- Fix url naming change

I tested the urls via browser as I otherwise kept getting spack cache URLs. I think they are right but I would appreciate a quick sanity check on the old URLs.

### DIFF
--- a/repos/spack_repo/builtin/packages/py_pint/package.py
+++ b/repos/spack_repo/builtin/packages/py_pint/package.py
@@ -12,6 +12,7 @@ class PyPint(PythonPackage):
     quantities: the product of a numerical value and a unit of measurement.
     It allows arithmetic operations between them and conversions from and
     to different units."""
+
     homepage = "https://github.com/hgrecco/pint"
     pypi = "pint/pint-0.11.tar.gz"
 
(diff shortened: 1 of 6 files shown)


Chrismarsh:
@spackbot fix style

repos/spack_repo/builtin/packages/py_pint/package.py
```diff
@@ -31,9 +35,13 @@ class PyPint(PythonPackage):
     version("0.9", sha256="32d8a9a9d63f4f81194c0014b3b742679dce81a26d45127d9810a68a561fe4e2")
     version("0.8.1", sha256="afcf31443a478c32bbac4b00337ee9026a13d0e2ac83d30c79151462513bb0d4")
 
+    variant("numpy", default=False, description="Optional numpy support")
+    variant("xarray", default=False, description="Optional xarray support")
+    variant("dask", default=False, description="Optional dask support")
```
sethrj, line 40: These don't seem like heavy dependencies; is there a reason not to just require them?

sethrj (dismissed):
I successfully installed py-pint 0.24.4 with this PR. Thanks!

repos/spack_repo/builtin/packages/py_pint/package.py
Chrismarsh: Just that it "breaks" previous install assumptions. And not everyone likes the overhead as dask and xarray pull in a ton of dependencies. I was erring on the side of caution. My preference would be this variant is True by default.

Chrismarsh:
@spackbot fix style

sethrj (approved):

PR #744, opened 2025-07-23
ascent: add version 0.9.4

vicentebolea:


tldahlgren (dismissed):
Version commit was confirmed.

@cyrush Do you have any objections to this PR being merged?

cyrush:
this looks great thank you @vicentebolea !

cyrush (approved):

repos/spack_repo/builtin/packages/ascent/package.py
```diff
@@ -196,6 +201,7 @@ class Ascent(CMakePackage, CudaPackage, ROCmPackage):
         extends("python")
         depends_on("py-numpy", type=("build", "run"))
         depends_on("py-pip", type=("build", "run"))
+        depends_on("py-wheel", type=("build", "run"), when="@0.9.4:")
```
kwryankrattiger: Is this needed in the runtime environment? Same with pip

cyrush: For my spack builds - I keep them as run so they appear in spack views. 
This is b/c we like to develop outside of spack (after tpls are built) -- that makes sure they exist as we develop python modules.

kwryankrattiger: If that is the case, then you can put those as root specs in your environment as roots. I don't think packages should expose extra stuff in their runtime if it isn't expressly needed.

cyrush: Then we have to put this in every environment. I do forget and it burns time. 

Fine with whatever spack feels is best - but having basic python packaging infrastructure in a python view doesn't seem excessive to me.

kwryankrattiger: If people want a python development environment, they need to specify it. My issue is when I install ascent, or any package, into a production environment I don't want ascent loading python modules underneath and overriding my local python environment. This is, IMO, a package bug since it introduces runtime modifications that are unnecessary to the use of the package.

Packages should only load what they need at runtime, extra stuff is up to the environment to decide if and how it should be exposed.

vicentebolea: I have removed the run from `py-wheel` since I am not sure it will be needed for using Ascent python at runtime. How about we continue the discussion for pip in another PR?

tldahlgren (dismissed):

repos/spack_repo/builtin/packages/ascent/package.py
kwryankrattiger: From my perspective these are the same, I don't think it is controversial to not inject build-only requirements into a packages runtime that can potentially cause issues for users. I would prefer remove them now since we are here

vicentebolea: @cyrush in my last push I added the changes to change the python deps (numpy included) as default dependency type "build+link" (link is noop for pip and wheel) this going in line with the comment from @kwryankrattiger. To further add to this discussion, I understand the frustration of having to specify the Python dev packages in your environment when building a cpython app with Ascent, this is certainly something that we miss since spack do not offer `-devel|dev` flavor of the packages (Which is how this is managed in other package managers. We could perhaps add `devel` variant to this recipe?

In any case I propose to move forward with this PR so that we get the new version and we eliminate this issue with the python packages that is biting us in the dav sdk env.

cyrush: @vicentebolea  sounds good. If there is a convention for a variant like `devel|dev` that there is a precedence for in spack recipes that helps with this case, that that would be good to add the in future.  Thanks!

repos/spack_repo/builtin/packages/ascent/package.py
```diff
@@ -194,8 +199,9 @@ class Ascent(CMakePackage, CudaPackage, ROCmPackage):
     with when("+python"):
         depends_on("python+shared")
         extends("python")
-        depends_on("py-numpy", type=("build", "run"))
-        depends_on("py-pip", type=("build", "run"))
+        depends_on("py-numpy")
```
kwryankrattiger: should numpy still be a runtime dep here?

cyrush: yes, numpy is part of the python api

vicentebolea: My bad, it gets confusing since numpy is both c lib and python lib. Furthermore I added to be a link type of dep too since it is also a c lib.

repos/spack_repo/builtin/packages/ascent/package.py
vicentebolea: I am not sure, @kwryankrattiger, Do you know what is the best pattern here?

kwryankrattiger: So the `-devel` is usually for developing with a package, not for it, so I wouldn't expect that pattern to map to this case. There has been some thoughts in the past to _only_ install runtime components of libraries and then install headers as a  build dependency separately.

The state of the art for creating a development environment for a spack pacakge, using spack, is either use `spack develop` and/or to create an spack environment that lists the packages required for development.

vicentebolea: Fair point, I misunderstood that Cyrus wanted to build applications that uses Ascent python API, I had in mind to add a "dev" variant such in this case: https://github.com/spack/spack-packages/blob/develop/repos/spack_repo/builtin/packages/py_blight/package.py#L22. But for developing the package Ryan is right spack develop is the way to go.

vicentebolea: I am resolving this to move on with this PR. Please reopen if more comments needed

vicentebolea:
@kwryankrattiger Ready to merge

kwryankrattiger (approved):

PR #751, opened 2025-07-24
nvbandwidth: add v0.8; fix dependency

wu-kan:
Note that 1.66.0 is just the version of boost document in the README, but not the real dependency. Meanwhile, it requires cuda to build.


tldahlgren (commented):
Confirmed the sha256.

@rfbgo As the person who contributed the package a year ago, are you interested in reviewing this PR?

rfbgo:
> Confirmed the sha256.
> 
> @rfbgo As the person who contributed the package a year ago, are you interested in reviewing this PR?

I'm no authority on it, but would be willing to if a suitable option (possibly you?) cannot be found

Overall looks fine to me. I suspect specifying a minimum boost version would be reasonable

tldahlgren:
> > Confirmed the sha256.
> > @rfbgo As the person who contributed the package a year ago, are you interested in reviewing this PR?
> 
> I'm no authority on it, but would be willing to if a suitable option (possibly you?) cannot be found
> 
> Overall looks fine to me. I suspect specifying a minimum boost version would be reasonable

Thanks.

@wu-kan Is there a reason the minimum `boost` version was dropped?

wu-kan:
I have updated the code while retaining the fixed version as the minimum required version. This approach is beneficial for future compatibility. Specifically, if lower versions of Boost are deprecated, it ensures stability. Additionally, it prevents potential errors that could arise if other packages require nvbandwith but depend on different versions of Boost.

```python
depends_on("boost@1.66.0:+program_options")
```

tldahlgren (approved):

PR #753, opened 2025-07-24
libdwarf: remove url for version and add v2.1.0

pranav-sivaraman:
Built and passes all the test on M1 Air.


alecbcs (approved):
Looks good to me. Thanks @pranav-sivaraman!

PR #757, opened 2025-07-24
jujutsu: add latest version

pranav-sivaraman:
Built and passing all the tests on `darwin-macos-m1`

Edit: running this in a sandboxed image reveals a dependency on a `c` (or at least the linker)


tldahlgren (approved):
Confirmed the sha256, rest LGTM.

PR #759, opened 2025-07-24
iwyu: add patch for 0.23

wdconinc:
This PR adds a patch to `iwyu` for https://github.com/include-what-you-use/include-what-you-use/pull/1652, which fixes broken syntax in STL header mappings.


sethrj (approved):
Looks good, checksum valid!

PR #761, opened 2025-07-24
py-certifi: add v2025.7.14

manuelakuhn:
https://github.com/certifi/python-certifi/tree/2025.07.14


tldahlgren (approved):
Confirmed the sha256 and dependencies.

PR #775, opened 2025-07-25
py-scikit-build-core: add v0.11.5

garth-wells:


repos/spack_repo/builtin/packages/py_scikit_build_core/package.py
wdconinc, line 1:
```
    depends_on("python@3.8:", type=("build", "run"), when="@0.11:")
```

garth-wells: Done, thanks.

wdconinc (approved):

PR #779, opened 2025-07-25
libmesh: Fix VTK with +vtk, add poly2tri variant

capitalaslash:
`libmesh` would silently disable some optional libraries when they are set via a simple `--libname-enable` switch. Some dependencies offer a `*-required` variant that enforces their activation.

From `libmesh` `configure` script, as an example for `petsc`:
```
--enable-petsc-required Error if PETSc is not detected by configure
```

This uncovered an error in the `vtk` dependency that was not truly ever enabled until the `--vtk-required` was added, leading to the change to the dependency to `vtk+mpi`.


repos/spack_repo/builtin/packages/libmesh/package.py
```diff
@@ -325,6 +327,11 @@ def configure_args(self):
         else:
             options.append("--enable-tbb=no")
 
+        if "+poly2tri" in self.spec:
+            options.append("--enable-poly2tri=yes")
+        else:
+            options.append("--enable-poly2tri=no")
```
bernhardkaindl: @capitalaslash, is this regular `autotools` `configure`? If so, it would normally work with:
```suggestion
        options.append(self.enable_or_disable("poly2tri"))
```
That is, if it supports `--disable-poly2tri` as well as `--enable-poly2tri=no`.

Of course, please test both if you apply this suggestion.
Otherwise, you can replace it with a if expression, I think.

repos/spack_repo/builtin/packages/libmesh/package.py
```diff
@@ -156,7 +158,7 @@ class Libmesh(AutotoolsPackage):
     depends_on("slepc", when="+slepc")
     depends_on("petsc", when="+petsc")
     depends_on("tbb", when="threads=tbb")
-    depends_on("vtk", when="+vtk")
+    depends_on("vtk+mpi", when="+vtk")
```
bernhardkaindl, line 164: @capitalaslash, you should add a justification why this change is needed to the PR or the code.

repos/spack_repo/builtin/packages/libmesh/package.py
```diff
@@ -224,8 +226,8 @@ def configure_args(self):
             options.append("--enable-netcdf=no")
 
         if self.spec.satisfies("+vtk"):
-            options.append("--enable-vtk")
-            options.append("--with-vtk=%s" % self.spec["vtk"].prefix)
+            options.append("--enable-vtk-required")
+            options.append("VTK_DIR=%s" % self.spec["vtk"].prefix)
```
bernhardkaindl: @capitalaslash, you should add a justification why this change is needed to the PR or the code.

repos/spack_repo/builtin/packages/libmesh/package.py
```diff
@@ -126,6 +127,7 @@ class Libmesh(AutotoolsPackage):
         multi=False,
     )
     variant("shared", default=True, description="Enables the build of shared libraries")
+    variant("poly2tri", default=False, description="Compile the poly2tri library")
```
bernhardkaindl: @capitalaslash the description of this variant could be improved by changing it to describe that it changes from the viewpoint of a libmesh user/caller (which feature it enables). Simply a "Compile" could be done without the variant. A a minimum, it may say "compile libmesh with libpoli2tri" but, prefreably, it could better be "Enable poly2tri for ....".

repos/spack_repo/builtin/packages/libmesh/package.py
```diff
@@ -274,7 +276,7 @@ def configure_args(self):
             options.append("--disable-metis")
 
         if self.spec.satisfies("+petsc") or self.spec.satisfies("+slepc"):
-            options.append("--enable-petsc=yes")
+            options.append("--enable-petsc-required")
```
bernhardkaindl, line 282: @capitalaslash, you should add a justification why this change is needed to the PR or the code.

bernhardkaindl (changes requested):

repos/spack_repo/builtin/packages/libmesh/package.py
capitalaslash: Thanks for the review!
I added som explanation in the PR description as suggested.

rbberger, line 164: Did you test with `vtk~mpi` when `~mpi`? It might only be needed when `+vtk+mpi`.

repos/spack_repo/builtin/packages/libmesh/package.py
rbberger: @bernhardkaindl Are we enforcing f-strings?

bernhardkaindl: Not that I heard of. In this case, keeping the % is pretty harmless for me (either way).
Just `.format(...)` is really distracting and should be avoided.

> Did you test with vtk~mpi when ~mpi? It might only be needed when +vtk+mpi.

Thanks for this comment, @capitalaslash: Please address that other comment, thanks!

repos/spack_repo/builtin/packages/libmesh/package.py
bernhardkaindl, line 164:
```suggestion
    depends_on("vtk+mpi", when="+vtk+mpi")
```
@rbberger, I think this is your suggestion? I'll wait for @capitalaslash to respond or merge this incremental change.

repos/spack_repo/builtin/packages/libmesh/package.py
bernhardkaindl: I haven't found an explanation, but I assume it changes that configure shall fail if it does not find VTK.
Ah, found it in the description.

repos/spack_repo/builtin/packages/libmesh/package.py
bernhardkaindl: Ah, found it there.
> libmesh would silently disable some optional libraries when they are set via a simple --libname-enable switch. Some dependencies offer a *-required variant that enforces their activation.
From libmesh configure script, as an example for petsc:
--enable-petsc-required Error if PETSc is not detected by configure
>  This uncovered an error in the vtk dependency that was not truly ever enabled until the --vtk-required was added, leading to the change to the dependency to vtk+mpi

bernhardkaindl (approved):
As my requested changes were now resolved, I'm cancelling my requested change flag using this approve (the only way to do so).

However, the question by @rbberger on wether this would also work (only require `vtk+mpi` when `libmesh+vtk+mpi`) remains the open question:
```py
   depends_on("vtk+mpi", when="+vtk+mpi")
```

bernhardkaindl (commented):
@capitalaslash, is this working (as asked) instead?
```py
depends_on("vtk+mpi", when="+vtk+mpi")
```

repos/spack_repo/builtin/packages/libmesh/package.py
capitalaslash: from `libmesh` `configure`:

```
<<< libMesh requires VTK to be configured with -DVTK_Group_MPI:BOOL=ON >>>
```

capitalaslash:
@bernhardkaindl see https://github.com/spack/spack-packages/pull/779#discussion_r2293757087

I have been trying to have `libmesh ~mpi +vtk` to work without any success for now.

repos/spack_repo/builtin/packages/libmesh/package.py
capitalaslash: see https://github.com/spack/spack-packages/pull/779#discussion_r2293757087

capitalaslash:
let me know if you want to update all strings to f-strings

repos/spack_repo/builtin/packages/libmesh/package.py
```diff
@@ -325,6 +327,8 @@ def configure_args(self):
         else:
             options.append("--enable-tbb=no")
 
+        options.append(self.enable_or_disable("poly2tri"))
```
bernhardkaindl: @capitalaslash, please test the commits that you ask to be reviewed to ensure that they work, and even more importantly, that they don't break the build of the package due to simple type errors.

In this line added with your latest commit, `enable_or_disable()` returns a list, not a string, as you assume:
```py
        options.append(self.enable_or_disable("poly2tri"))
```

This breaks the calling function that require the options to be strings, e.g. `spack install libmesh@1.7.5` leads to

```py
==> libmesh: Executing phase: 'configure'
==> Error: TypeError: expected string or bytes-like object

/home/bkaindl/spack-packages/repos/spack_repo/builtin/build_systems/autotools.py:627, in configure:
        624        options += self.configure_args()
        625
        626        with working_dir(self.build_directory, create=True):
  >>    627            pkg.module.configure(*options)
```

You can fix this using `options.extend()` instead of `.append()`, but even shorter would be `options +=`:
```suggestion
        options += self.enable_or_disable("poly2tri"))
```

I suggest building the commits that you push to help the credibility of the PR.

bernhardkaindl (changes requested):

repos/spack_repo/builtin/packages/libmesh/package.py
bernhardkaindl: Two more changes, while you are at it, because now, you need to test the new commit(s) and `spack` also needs to rerun GitLab CI (which requires quite some compute in the Cloud):

1. As you now need to push a commit, also include changing the VTK= string to an f-string.

2. While at it, also apply this update to add "m4" as a build dependency for the configure (on a build host that does not have `m4` installed, configure failed, and I tested that it works with this line added):
```diff
--- a/repos/spack_repo/builtin/packages/libmesh/package.py
+++ b/repos/spack_repo/builtin/packages/libmesh/package.py
@@ -140,6 +140,7 @@ class Libmesh(AutotoolsPackage):
     depends_on("c", type="build")
     depends_on("cxx", type="build")
     depends_on("fortran", type="build")
+    depends_on("m4", type="build")

     depends_on("boost", when="+boost")
```

bernhardkaindl:
@capitalaslash, also take care that the PR title reflects the current changes in the PR and is specific to what the PR does, because not only reviewers read it, but it also becomes the commit message that is proposed when/if merging to the PR. I updated it now for these reasons.

repos/spack_repo/builtin/packages/libmesh/package.py
capitalaslash: Sorry about the build failure, I copy-pasted from your https://github.com/spack/spack-packages/pull/779#discussion_r2258699008. I test all my commits before pushing, I am not sure how I did not catch this.

repos/spack_repo/builtin/packages/libmesh/package.py
capitalaslash: I can confirm that the `vtk` dependency can only be enabled when `+mpi` is present in the spec. What is the best way to ensure `+mpi` is set when `+vtk` is required? @bernhardkaindl

capitalaslash: I did it with a `when="+mpi"` specification on the `vtk` variant definition

capitalaslash:
@bernhardkaindl any chance this can be reviewed again?

bernhardkaindl (approved):
Looks good with the updates.

rbberger (approved):

PR #782, opened 2025-07-25
py-coverage: add v7.10.0

manuelakuhn:
https://github.com/nedbat/coveragepy/tree/7.10.0


bernhardkaindl (approved):
Installs successfully on Ubuntu.

PR #783, opened 2025-07-25
gha: test package.py canonicalization across python versions

haampie:
Depends on #839.

Test that the canonicalized sources of builtin's package.py files are
identical for all supported Python versions

Container image is created in https://github.com/spack/all-pythons


alalazo (approved):
LGTM. Using `git` to compare canonicalizations saves us having to code more boilerplate.

PR #786, opened 2025-07-25
melissa: add v2.1.0

abhishek1297:
### Main changes:
- Removed Cmake RPATH from spack recipes
- Added two versions of `melissa` and `py-melissa-core` packages


repos/spack_repo/builtin/packages/melissa/package.py
```diff
@@ -14,10 +14,20 @@ class Melissa(CMakePackage):
 
     homepage = "https://gitlab.inria.fr/melissa/melissa"
     git = "https://gitlab.inria.fr/melissa/melissa.git"
-    url = "https://gitlab.inria.fr/melissa/melissa/-/archive/v2.0.0/melissa-v2.0.0.tar.gz"
+    url = "https://gitlab.inria.fr/melissa/melissa/-/archive/v2.1.0/melissa-v2.1.0.tar.gz"
```
alecbcs, line 17: Just a quick note that Spack [automatically handles URL extrapolation](https://spack.readthedocs.io/en/v1.0.0/packaging_guide_creation.html#versions-and-urls) to derive the correct download URLs from the existing URL without needing to update the URL each version. (But also this doesn't hurt anything so I'm not requiring changes, just might save you some keystrokes in the future.)

repos/spack_repo/builtin/packages/melissa/package.py
```diff
@@ -14,10 +14,20 @@ class Melissa(CMakePackage):
...
+    version(
+        "2.1.0",
+        sha256="cf0f105ed5b1da260cc7476aec23df084470b50a61df997c0e457c38948bed93",
+        preferred=True,
+    )
+    version(
+        "2.0.1",
+        sha256="a7ff4df75ea09af435b0c28c3fa3cab9335c1c76e1c48757facce36786b4962c",
+        preferred=True,
+    )
```
alecbcs: Let's drop the `preferred=True` property from all of these `version()` directives. The preferred property is used to tell Spack to pick a specific version of a package that is not the latest as the default. Since all of the versions here are preferred, this flag makes no difference to concretization.

(Versions like `develop`, `main`, etc... are not preferred by default if that was the original intention.)

repos/spack_repo/builtin/packages/melissa/package.py
```diff
@@ -48,17 +58,17 @@ class Melissa(CMakePackage):
...
+    #     rpaths = [self.spec["libzmq"].prefix.lib, self.spec["mpi"].prefix.lib]
+    #     joined_rpaths = ";".join(rpaths)
 
-        args.append(f"-DCMAKE_INSTALL_RPATH={joined_rpaths}")
-        args.append("-DCMAKE_INSTALL_RPATH_USE_LINK_PATH=ON")
+    #     args.append(f"-DCMAKE_INSTALL_RPATH={joined_rpaths}")
+    #     args.append("-DCMAKE_INSTALL_RPATH_USE_LINK_PATH=ON")
 
-        return args
+    #     return args
```
alecbcs: Are these meant to be commented out? Or was this leftover from debugging the build.

repos/spack_repo/builtin/packages/py_melissa_core/package.py
```diff
@@ -17,11 +17,21 @@ class PyMelissaCore(PythonPackage, CudaPackage):
 
     homepage = "https://gitlab.inria.fr/melissa/melissa"
     git = "https://gitlab.inria.fr/melissa/melissa.git"
-    url = "https://gitlab.inria.fr/melissa/melissa/-/archive/v2.0.0/melissa-v2.0.0.tar.gz"
+    url = "https://gitlab.inria.fr/melissa/melissa/-/archive/v2.1.0/melissa-v2.1.0.tar.gz"
```
alecbcs, line 20: Same feedback as on the main melissa package.

repos/spack_repo/builtin/packages/py_melissa_core/package.py
```diff
@@ -17,11 +17,21 @@ class PyMelissaCore(PythonPackage, CudaPackage):
...
+    version(
+        "2.1.0",
+        sha256="cf0f105ed5b1da260cc7476aec23df084470b50a61df997c0e457c38948bed93",
+        preferred=True,
+    )
+    version(
+        "2.0.1",
+        sha256="a7ff4df75ea09af435b0c28c3fa3cab9335c1c76e1c48757facce36786b4962c",
+        preferred=True,
+    )
```
alecbcs: Same feedback as on the main melissa package.

alecbcs (changes requested):
@abhishek1297 thanks for the added versions! Just a couple small notes regarding the formatting of the directives. Otherwise this looks good to me.

repos/spack_repo/builtin/packages/melissa/package.py
abhishek1297: Ah, Okay. Noted. Thanks

repos/spack_repo/builtin/packages/melissa/package.py
abhishek1297: Fixed in new commit

repos/spack_repo/builtin/packages/melissa/package.py
abhishek1297: Yes. I moved the RPATH logic inside the project cmake now. I still keep it commented in spack. Will remove it later

alecbcs: So we don't want to comment this out / remove it from the package since that'll break backwards compatibility for people trying to build older versions of melissa. Instead I'd recommend putting in some conditional logic to add the flags when the version of melissa is less than the version in which you moved the logic.

abhishek1297: My bad. Fixed it now.

repos/spack_repo/builtin/packages/melissa/package.py
```diff
@@ -14,16 +14,16 @@ class Melissa(CMakePackage):
 
     homepage = "https://gitlab.inria.fr/melissa/melissa"
     git = "https://gitlab.inria.fr/melissa/melissa.git"
-    url = "https://gitlab.inria.fr/melissa/melissa/-/archive/v2.0.0/melissa-v2.0.0.tar.gz"
+    url = "https://gitlab.inria.fr/melissa/melissa/-/archive/v2.1.0/melissa-v2.1.0.tar.gz"
     # attention: Git**Hub**.com accounts
     maintainers("abhishek1297", "viperML", "raffino")
```
viperML: Remove viperML from maintainers.

viperML (changes requested):

raffino (approved):

abhishek1297:
Hi @alecbcs, Please review this PR.

abhishek1297:
closing because of some conflicts with the upstream

PR #787, opened 2025-07-25
sherpa: filter_file to fix failing texinfo file

wdconinc:
This PR fixes `sherpa`, all versions `@:2`, by changing an incorrect `R@end` into `@end` in a texinfo file. See https://gitlab.com/sherpa-team/sherpa/-/blob/v2.2.16/Manual/Starting.texi?ref_type=tags#L469.

The v2 series is in maintenance mode. If this ends up getting fixed upstream anyway, and if a new version is released and added to spack, then the filter_file will become a no-op, so this is a future-proof fix.

This should fix the failing HEP GitLab CI stack.

Root cause analysis: The upgrade in #575 of `texinfo` to 7.2 (which is what now causes an error where previously it didn't cause an error, as verified for 7.1) did not trigger this issue in the PR pipeline since `texinfo` is a build dependency, and packages with updated build dependencies are not rebuilt in GitLab CI.

### DIFF
--- a/repos/spack_repo/builtin/packages/sherpa/package.py
+++ b/repos/spack_repo/builtin/packages/sherpa/package.py
@@ -95,6 +95,8 @@ class Sherpa(CMakePackage, AutotoolsPackage):
     depends_on("libtool", when="@:2")
     depends_on("m4", when="@:2")
     depends_on("texinfo", type="build")
+    # limit texinfo to 7.1 to avoid errors due to missing references
+    depends_on("texinfo@:7.1", type="build", when="@:2")
     depends_on("sqlite")
 
     depends_on("mpi", when="+mpi")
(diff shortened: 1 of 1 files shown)


wdconinc:
Rebased onto 34e144695e8d045d0f2f68b439e1799b7643ad92 for faster CI checks.

adamjstewart:
I would also be okay with a `depends_on("texinfo@:7.1", when="@:2", type="build")`

wdconinc:
> I would also be okay with a `depends_on("texinfo@:7.1", when="@:2", type="build")`

If this simple fix doesn't work, I'll revert to that (see 2ebf29f).

wdconinc:
Succeeded `sherpa@2` build at https://gitlab.spack.io/spack/spack-packages/-/jobs/17611782.

adamjstewart (approved):

PR #788, opened 2025-07-25
scorep: fix `gotcha` `prefix`

wrwilliams:
`--with-libgotcha` needs to look at the prefix for `gotcha`, which exists, not the one for `libgotcha`, which does not.


wrwilliams:
@spackbot fix style

Datawarrior85:
I tested this update and the package builds without erroring out on the libgotcha configure error. Thanks Bill

tldahlgren (approved):

PR #792, opened 2025-07-25
curl: add 8.14.1

tgamblin:
Trying a simpler change for 2025.07.1 instead of the jump to 8.15.

8.11.1 has CVEs and triggers scanners. Update to 8.14.1 to solve that issue, but not (yet) 8.15.0.

8.15.0 removes support for secure transport and `cmake` requires a patch to work with it, so hold off on that update. This PR is meant to be backported to `v2025.07.1`. I think the 8.15.0 update will need to come later because it is more disruptive.


tgamblin:
8.14.1 apparently requires the same patch as 8.15.0, so closing in favor of doing both at once.

PR #688, opened 2025-07-21
micromamba: fix build

Pandapip1:
Fixes a few issues that cause micromamba not to build

PR #797, opened 2025-07-25
(micro)mamba: Fix build, deps, rename, and add v2.3

Chrismarsh:
The current package.py for mamba is 2-3 years out of date and doesn't compile. 

This PR fixes these issues:
- Rename from `micromamba` to `mamba`. `micromamba` is the statically linked version of `mamba` ([as per](https://github.com/mamba-org/mamba)). This package's static builds are busted, and micromamba is really a subset of mamba. Thus, this has been renamed for clarity.
- Add version `mamba` version 2.3
- Deprecated the older versions of `mamba` as I couldn't get them to compile, and I'm skeptical of much value in 3 year old versions of package managers
- Add a patch to support the new fmt and tighten up all the depends versions
- Added new versions of dependent libraries required to build with gcc14+ compilers
- Fix a few missing `depends_on("c")`
- Fix `tl_expected` to produce the other missing half of the includes. As a result I've added the `+precompiled` variant to return to the previous behaviour, however based on the docs I don't think the precompiled is being used as intended when compile in Spack.

The static build of `mamba` is still broken, and I am not entirely sure how it should be built. I've kept the template in the file such that this can be fixed in the future.


Chrismarsh:
@spackbot fix style

Chrismarsh:
I note that the chapel and cmake changes were added by spackbot fixing style and not me!

Chrismarsh:
@nightlark: `cli11`
@haampie: `cpp_termcolor`
@charmoniumQ: `libreproc`, `libsolv`, `mamba`, `tl_expected`

wdconinc:
See also #688 @Pandapip1.

Chrismarsh:
@wdconinc urghhh, I searched in issues not PR when checking if this was being worked on!

wdconinc:
These efforts are complementary.

Chrismarsh:
I agree.

 @Pandapip1 are you ok if I take a few of your changes from https://github.com/spack/spack-packages/pull/688 into here?

- `cli11` I think the precompiled variant is the correct way but if your fix fixes the imp/ then having both is worth imo
- Where is the curl version determined from?

Chrismarsh:
I added the cli11 `imp/` fix and it works correctly in my testing

Pandapip1:
Yea, feel free to cherry-pick the changes from my PR (I'd prefer git coauthorship for attribution but I don't particularly mind). I don't remember where I found the curl version, but it cleared one set of errors after it was applied so :shrug:. If you want to do a bisection, well that's the best option I can think of. Do we want to go ahead and merge #688 and rebase this one?

Chrismarsh:
Ok I'll revert [my commit](https://github.com/spack/spack-packages/pull/797#issuecomment-3121042166) and cherry pick [commit1](https://github.com/spack/spack-packages/pull/688/commits/7470ef81048cb4de031b2fd6b174a374f076e2b3) and [commit2](https://github.com/spack/spack-packages/pull/688/commits/2fb20feb7aa83206f1282fe75b24d2e185134f9a) to ensure the co authorship. 

IMO closing #688 and having all the commits here makes the most sense?

Pandapip1:
Sure; that works too. Closing https://github.com/spack/spack-packages/pull/688.

Chrismarsh:
@Pandapip1 please let me know if you're happy with how the commit attribution is showing up for you. I cherry picked your cli11 commits off your PR, but I had to `git --amend --author` for the curl commit because a) the file was renamed and b) the package.py has diverged significantly from the original, so the patch wouldn't cleanly apply.

Pandapip1 (approved):
LGTM :shipit:

bernhardkaindl (approved):

PR #807, opened 2025-07-27
py-libclang: setup.py renamed

adamjstewart:
Extracted out of #39 

@rbberger @wdconinc @haampie @skosukhin @trws

Next problem is that the package calls `spec['llvm'].libs`, which calls `llvm-config`, which is not installed in the CI image because we only installed clang/flang. https://github.com/spack/gitlab-runners/pull/67 is one potential fix, but not very robust. Ideally, there would be some way to say that `py-libclang` requires `llvm+dev` and Spack would install it itself. Is there any variant or CMake flag that controls whether `llvm-config` or the actual library files get installed?

### DIFF
--- a/repos/spack_repo/builtin/packages/py_libclang/package.py
+++ b/repos/spack_repo/builtin/packages/py_libclang/package.py
@@ -38,9 +38,14 @@ class PyLibclang(PythonPackage):
         depends_on("llvm+clang@" + ver, when="@" + ver, type="build")
 
     def patch(self):
+        if self.version >= Version("14"):
+            setup = "setup_ext.py"
+        else:
+            setup = "setup.py"
+
         filter_file(
             "source_dir = './native/'",
             "source_dir = '{0}'".format(self.spec["llvm"].libs.directories[0]),
-            "setup.py",
+            setup,
             string=True,
         )


wdconinc (commented):
Looks good to me. Sorry to have missed this in the recent upgrade of libclang.

wdconinc (approved):

wdconinc:
> Next problem is that the package calls `spec['llvm'].libs`, which calls `llvm-config`, which is not installed in the CI image because we only installed clang/flang. [spack/gitlab-runners#67](https://github.com/spack/gitlab-runners/pull/67) is one potential fix, but not very robust. Ideally, there would be some way to say that `py-libclang` requires `llvm+dev` and Spack would install it itself. Is there any variant or CMake flag that controls whether `llvm-config` or the actual library files get installed?

Maybe another approach is to prevent detecting llvm as external when llvm-config is missing. [This isn't the only case](https://github.com/search?q=repo%3Aspack%2Fspack-packages%20llvm-config&type=code) where llvm-config is used. And it's not the only case of a spack package requiring multiple OS packages to satisfy the assumptions (rust = rustc + cargo).

adamjstewart:
I think we still have to detect it in order to use clang/flang.

PR #813, opened 2025-07-27
genie: add v3.6.2

wdconinc:
This PR adds `genie`, v3.6.2 ([diff](https://github.com/GENIE-MC/Generator/compare/R-3_04_02...R-3_06_02)). No build system changes that require adapting our build recipe (we have not exposed all build options as variants, so the addition of pythia6 as a new build option is not exposed here either and true remains the default).


alecbcs (approved):

PR #814, opened 2025-07-27
py-hatch: add v1.14.1

wdconinc:
This PR adds `py-hatch`, v1.14.1 ([diff](https://github.com/pypa/hatch/compare/hatch-v1.13.0...hatch-v1.14.1)). Some updated dependencies and two mistakes fixed: py-hatchling is a run dependency as well, and py-zstandard must be < 1 per pyproject.toml.

### DIFF
--- a/repos/spack_repo/builtin/packages/py_hatch/package.py
+++ b/repos/spack_repo/builtin/packages/py_hatch/package.py
@@ -15,14 +15,16 @@ class PyHatch(PythonPackage):
 
     license("MIT")
 
+    version("1.14.1", sha256="ca1aff788f8596b0dd1f8f8dfe776443d2724a86b1976fabaf087406ba3d0713")
     version("1.13.0", sha256="5e1a75770cfe8f3ebae3abfded3a976238b0acefd19cdabc5245597525b8066f")
     version("1.12.0", sha256="ae80478d10312df2b44d659c93bc2ed4d33aecddce4b76378231bdf81c8bf6ad")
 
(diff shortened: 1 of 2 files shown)


repos/spack_repo/builtin/packages/py_hatch/package.py
```diff
@@ -37,5 +39,7 @@ class PyHatch(PythonPackage):
     depends_on("py-tomlkit@0.11.1:", type=("build", "run"))
     depends_on("py-userpath@1.7:1.7", type=("build", "run"))
```
manuelakuhn: While you are at it can you fix the dependency restriction here?
According to the `pyproject.toml` it is `userpath~=1.7`
```suggestion
    depends_on("py-userpath@1.7:1", type=("build", "run"))
```

repos/spack_repo/builtin/packages/py_virtualenv/package.py
manuelakuhn, line 1: Although also not a problem of your PR, but while you are on it, can you also fix the `platformdirs` dependency here?
```suggestion
    depends_on("py-platformdirs@3.9.1:4", when="@20.24.7:", type=("build", "run"))
    depends_on("py-platformdirs@3.9.1:3", when="@20.24.1:20.24.6", type=("build", "run"))
```

wdconinc: Took two weeks, but done now.

alalazo (approved):
Feel free to merge, if you're not waiting for further reviews

PR #817, opened 2025-07-28
py-uvloop: add v0.21.0

wu-kan:


bnikolic:
Not related to this update, but by the way  I think the cythonized files will not be regenerated unless cython-always flag is passed. (And so we'll be using a very large auto-generated file)

bernhardkaindl (approved):

PR #825, opened 2025-07-28
e4s ci: require ~hipblaslt for rocblas

afzpatel:
This pr adds a requirement for rocblas~hipblaslt since hipblaslt hangs in ci.
Also attempts to add back the specs that were turned off here: https://github.com/spack/spack-packages/pull/646


eugeneswalker (approved):
Thank you!

PR #826, opened 2025-07-28
py-cycler: add v0.12.1

manuelakuhn:
https://github.com/matplotlib/cycler/tree/v0.12.1

Add a `py-setuptools@61:` dependency because older versions don't work and the package is not found when imported.


alecbcs (approved):

PR #828, opened 2025-07-28
paraview: add viskores in newer versions

kmorel:
Starting with ParaView 6.0, the VTK-m accelerator library changed to Viskores. Update the options to use the new accelerator library.

### DIFF
--- a/repos/spack_repo/builtin/packages/paraview/package.py
+++ b/repos/spack_repo/builtin/packages/paraview/package.py
@@ -39,6 +39,9 @@ supported_cuda_archs = {
     "86": "ampere",
 }
 
+# This is a list of paraview variants that require the viskores library.
+viskores_dependency_variants = ["+cuda", "+fides", "+rocm"]
+
 
 class Paraview(CMakePackage, CudaPackage, ROCmPackage):
     """ParaView is an open-source, multi-platform data analysis and
@@ -146,6 +149,7 @@ class Paraview(CMakePackage, CudaPackage, ROCmPackage):
     variant(
         "use_vtkm",
         default="default",
+        when="@5.3.0:5.13",
         multi=False,
         values=("default", "on", "off"),
         description="Build VTK-m with ParaView by setting PARAVIEW_USE_VTKM=ON,OFF."
@@ -173,8 +177,6 @@ class Paraview(CMakePackage, CudaPackage, ROCmPackage):
     conflicts("build_edition=catalyst", when="@:5.7")
     conflicts("build_edition=rendering", when="@:5.7")
     conflicts("build_edition=core", when="@:5.7")
-    # before 5.3.0, ParaView didn't have VTK-m
-    conflicts("use_vtkm=on", when="@:5.3")
     # paraview@5.9.0 is recommended when using the xl compiler
     # See https://gitlab.kitware.com/paraview/paraview/-/merge_requests/4433
     conflicts(
@@ -311,14 +313,30 @@ class Paraview(CMakePackage, CudaPackage, ROCmPackage):
     depends_on("xz")
     depends_on("zlib-api")
     depends_on("libcatalyst@2:", when="+libcatalyst")
-    depends_on("hip@5.2:", when="+rocm")
-    # CUDA thrust is already include in the CUDA pkg
-    depends_on("rocthrust", when="@5.13: +rocm ^cmake@3.24:")
-    for target in ROCmPackage.amdgpu_targets:
-        depends_on(
-            "kokkos@:3.7 +rocm amdgpu_target={0}".format(target),
-            when="+rocm amdgpu_target={0}".format(target),
-        )
+
+    # For @6.0:, depend on external Viskores module when needed.
+    with when("@6:"):
+        for variant in viskores_dependency_variants:
+            depends_on("viskores +vtktypes +64bitids", when=f"{variant}")
+        depends_on("viskores +cuda", when="+cuda")
+        for arch in CudaPackage.cuda_arch_values:
+            depends_on(f"viskores +cuda cuda_arch={arch}", when=f"+cuda cuda_arch={arch}")
+        depends_on("viskores +rocm", when="+rocm")
+        for target in ROCmPackage.amdgpu_targets:
+            depends_on(
+                f"viskores +rocm amdgpu_target={target}", when=f"+rocm amdgpu_target={target}"
+            )
+
+    # For @:5.13, VTK-m is built internally, so set up hip internally.
+    with when("@:5"):
+        depends_on("hip@5.2:", when="+rocm")
+        # CUDA thrust is already include in the CUDA pkg
+        depends_on("rocthrust", when="+rocm ^cmake@3.24:")
+        for target in ROCmPackage.amdgpu_targets:
+            depends_on(
+                "kokkos@:3.7 +rocm amdgpu_target={0}".format(target),
+                when="+rocm amdgpu_target={0}".format(target),
+            )
 
     # Older builds of pugi export their symbols differently,
     # and pre-5.9 is unable to handle that.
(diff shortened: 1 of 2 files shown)


kmorel:
@kwryankrattiger Can you please review this?

repos/spack_repo/builtin/packages/paraview/package.py
```diff
@@ -152,18 +152,31 @@ class Paraview(CMakePackage, CudaPackage, ROCmPackage):
         ' "default" lets the build_edition make the decision.'
         ' "on" or "off" will always override the build_edition.',
     )
+    variant(
+        "use_viskores",
+        default="default",
+        multi=False,
+        values=("default", "on", "off"),
+        description="Build Viskores with ParaView by setting PARAVIEW_USE_VISKORES=ON,OFF."
```
kwryankrattiger: Here and `when="@:5"` for the `use_vtkm` variant. This should make it possible to drop the added version constraint everywhere these are used.

```suggestion
        when="@6:",
        description="Build Viskores with ParaView by setting PARAVIEW_USE_VISKORES=ON,OFF."
```

repos/spack_repo/builtin/packages/paraview/package.py
```diff
@@ -152,18 +152,31 @@ class Paraview(CMakePackage, CudaPackage, ROCmPackage):
         ' "default" lets the build_edition make the decision.'
         ' "on" or "off" will always override the build_edition.',
     )
+    variant(
+        "use_viskores",
```
kwryankrattiger: I am torn about adding this variant vs just renaming it to something like `accelerators` or making it implicit with `+cuda`, `+rocm`, `+fides` rather than having all of those be conflicts.

repos/spack_repo/builtin/packages/paraview/package.py
```diff
@@ -661,10 +677,17 @@ def use_x11():
         cmake_args.append("-DPARAVIEW_BUILD_SHARED_LIBS:BOOL=%s" % variant_bool("+shared"))
```
kwryankrattiger, line 686: See previous suggestion to just drop the variant.

```suggestion
use_viskores = spec.satisfies("+cuda") or spec.satisfies("+rocm") or spec.satisfies("+fides")
```

repos/spack_repo/builtin/packages/paraview/package.py
```diff
@@ -175,6 +188,9 @@ class Paraview(CMakePackage, CudaPackage, ROCmPackage):
     conflicts("build_edition=core", when="@:5.7")
     # before 5.3.0, ParaView didn't have VTK-m
     conflicts("use_vtkm=on", when="@:5.3")
+    # After 5.13, ParaView switched from VTK-m to Viskores
+    conflicts("use_vtkm=on", when="@6.0:")
+    conflicts("use_viskores=on", when="@:5.13")
```
kwryankrattiger: For ParaView 6.0 and later, we can use external Viskores. I would rather do that here if we can too. We will come back through on this as well and do external VTK, but that requires some TLC to the VTK package first.

kwryankrattiger (commented):
Sorry for taking so long on this, I have been back and forth on how best to handle this going forward.

It may be better to just prune this variant at enable the accelerator implicitly. And then external Viskores would be really nice to not have to rebuild it from scratch if we don't have to, especially for the GPU stuff.

repos/spack_repo/builtin/packages/paraview/package.py
kmorel: I like the idea of using an external package. That is more in line with Spack and can simplify several things.

kmorel:
@kwryankrattiger I've updated this package to use an external viskores package. It also removes the `use_viskores` variant and instead implicitly adds the dependency when needed. I think that generally addresses your comments.

I have not extensively tested this, but the spec at least works.

repos/spack_repo/builtin/packages/paraview/package.py
```diff
@@ -308,13 +307,25 @@ class Paraview(CMakePackage, CudaPackage, ROCmPackage):
...
+
+    # For @6.0:, depend on external Viskores module when needed.
+    depends_on("viskores", when="@6: +fides")
+    depends_on("viskores +cuda", when="@6: +cuda")
+    for arch in CudaPackage.cuda_arch_values:
+        depends_on(f"viskores +cuda cuda_arch={arch}", when=f"@6: +cuda cuda_arch={arch}")
+    depends_on("viskores +rocm", when="@6: +rocm")
+    for target in ROCmPackage.amdgpu_targets:
+        depends_on(
+            f"viskores +rocm amdgpu_target={target}", when=f"@6: +rocm amdgpu_target={target}"
```
kwryankrattiger: This can go inside the same loop as kokkkos.

repos/spack_repo/builtin/packages/paraview/package.py
```diff
@@ -308,13 +307,25 @@ class Paraview(CMakePackage, CudaPackage, ROCmPackage):
...
+
+    depends_on("hip@5.2:", when="@:5.13 +rocm")
     # CUDA thrust is already include in the CUDA pkg
-    depends_on("rocthrust", when="@5.13: +rocm ^cmake@3.24:")
+    depends_on("rocthrust", when="@5.13 +rocm ^cmake@3.24:")
     for target in ROCmPackage.amdgpu_targets:
         depends_on(
             "kokkos@:3.7 +rocm amdgpu_target={0}".format(target),
-            when="+rocm amdgpu_target={0}".format(target),
+            when="@:5.13 +rocm amdgpu_target={0}".format(target),
```
kwryankrattiger: You should be able to just use `@:5` here. I think that is more clear to read "All versions of 5 and older`

```suggestion
            when="@:5 +rocm amdgpu_target={0}".format(target),
```

repos/spack_repo/builtin/packages/paraview/package.py
```diff
@@ -661,11 +672,21 @@ def use_x11():
         cmake_args.append("-DPARAVIEW_BUILD_SHARED_LIBS:BOOL=%s" % variant_bool("+shared"))
 
         # VTK-m added to ParaView in 5.3.0 and up
-        if spec.satisfies("@5.3.0:") and spec.variants["use_vtkm"].value != "default":
+        if spec.satisfies("@5.3.0:5.13") and spec.variants["use_vtkm"].value != "default":
```
kwryankrattiger:
```suggestion
        if spec.satisfies("@5.3.0:5") and spec.variants["use_vtkm"].value != "default":
```

repos/spack_repo/builtin/packages/paraview/package.py
kmorel: > This can go inside the same loop as kokkkos.

I would rather keep these as separate loops. That keeps a separation of the code that is pre-v6 using an internal VTK-m and post-6 using Viskores and an external package.

repos/spack_repo/builtin/packages/paraview/package.py
```diff
@@ -308,13 +310,27 @@ class Paraview(CMakePackage, CudaPackage, ROCmPackage):
     depends_on("xz")
     depends_on("zlib-api")
     depends_on("libcatalyst@2:", when="+libcatalyst")
-    depends_on("hip@5.2:", when="+rocm")
+
+    # For @6.0:, depend on external Viskores module when needed.
+    for variant in viskores_dependency_variants:
+        depends_on("viskores +vtktypes +64bitids", when="@6: +fides")
```
kwryankrattiger:
```suggestion
        depends_on("viskores +vtktypes +64bitids", when=f"@6: {variant}")
```

repos/spack_repo/builtin/packages/paraview/package.py
```diff
@@ -308,13 +310,27 @@ class Paraview(CMakePackage, CudaPackage, ROCmPackage):
     depends_on("xz")
     depends_on("zlib-api")
     depends_on("libcatalyst@2:", when="+libcatalyst")
-    depends_on("hip@5.2:", when="+rocm")
+
```
kwryankrattiger, line 316: I think this entire section could be in a 

```
with when("@6:"):
    ....
```

repos/spack_repo/builtin/packages/paraview/package.py
kwryankrattiger: Now that I think about it a bit more I agree, see my other comment to put the 6: stuff in a `when` block

repos/spack_repo/builtin/packages/paraview/package.py
```diff
@@ -743,7 +770,12 @@ def use_x11():
             cmake_args.append("-DVTK_DEBUG_LEAKS:BOOL=ON")
 
         if spec.satisfies("@5.11:"):
-            cmake_args.append("-DPARAVIEW_USE_HIP:BOOL=%s" % variant_bool("+rocm"))
+            if spec.satisfies("@6.0:"):
```
kwryankrattiger: I think this would be easier to read as

```python
if spec.satisfies("@5.11:5"):
    cmake_args.append("-DPARAVIEW_USE_HIP:BOOL=%s" % variant_bool("+rocm"))
else:
    cmake_args.append("-DPARAVIEW_USE_KOKKOS:BOOL=%s" % variant_bool("+rocm"))

if "+rocm" in spec:
    if spec.satisfies("@6:"):
            cmake_args.append("-DPARAVIEW_KOKKOS_BACKEND:STRING=HIP")
    ...
```

repos/spack_repo/builtin/packages/paraview/package.py
kmorel: Good suggestion. I implemented that.

repos/spack_repo/builtin/packages/paraview/package.py
kmorel: Done.

kwryankrattiger:
Ported these changes to https://github.com/spack/spack-packages/pull/192 with other ParaView 6 Updates

kmorel:
Closing this PR. Changes are superseded by #192

bernhardkaindl:
kmorel wrote:
> Closing this PR. Changes are superseded by https://github.com/spack/spack-packages/pull/192

PR #830, opened 2025-07-28
gdrcopy: use `cuda_arch` information

albestro:
Before this PR `gdrcopy` relied on an internal default, and more recently on an [internal custom script](https://github.com/NVIDIA/gdrcopy/blob/master/scripts/get_cuda_gencode.sh), to decide for which set of GPU architectures to build for.

With this change, the spack package now respects the value specified for the variant `cuda_arch`.

Not tested for older versions, but according to https://github.com/NVIDIA/gdrcopy/commit/43305764b41fe7edeb389f8a7c1d5d920085fdac the name of the variable has changed over time.

### DIFF
--- a/repos/spack_repo/builtin/packages/gdrcopy/package.py
+++ b/repos/spack_repo/builtin/packages/gdrcopy/package.py
@@ -49,6 +49,10 @@ class Gdrcopy(MakefilePackage, CudaPackage):
 
     def setup_build_environment(self, env: EnvironmentModifications) -> None:
         env.set("CUDA", self.spec["cuda"].prefix)
+        env.set(
+            "NVCCFLAGS" if self.spec.satisfies("@2.4:") else "CUFLAGS",
+            "".join(self.cuda_flags(self.spec.variants["cuda_arch"].values)),
+        )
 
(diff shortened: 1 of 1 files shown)


msimberg (dismissed):
Looks good to me, but giving package maintainer a chance to comment.

albestro:
@RMeli did a quick search about the old versions, and apparently `CUFLAGS` was actually never used. On his input, I investigated a bit more and it seems that `tests/*.cu` appeared starting from 2.4, so I'd say the problem of passing `NVCCFLAGS` did not exist before then.

This seems confirmed by looking also at `Makefile` across 2.3 and 2.4 versions, where in the older one there is neither a target addressing `cu` files nor a reference to `nvcc`.

https://github.com/NVIDIA/gdrcopy/blob/v2.3/tests/Makefile
https://github.com/NVIDIA/gdrcopy/blob/v2.4/tests/Makefile

Therefore, I'm going to make this change conditional and applied just starting from 2.4 exclusively.

msimberg (approved):

RMeli (approved):

scothalverson:
Looks good to me!

PR #831, opened 2025-07-28
CI: Run release tags as noop jobs

kwryankrattiger:
Release tag copy can be done using other infrastructure tricks that are not subject to non-determinism in concretization. Due to the requirement for protected publish, something has to run and complete in CI for the protected publish job, and the subsequent cache indexing jobs, to run. This forces all generate steps to run noop jobs and nothing else.


bernhardkaindl (commented):
@zackgalbreath, as you are a code owner of one of files this would require your approval for it to be merged.

.ci/gitlab/.gitlab-ci.yml
```diff
@@ -46,14 +46,10 @@ default:
...
-    - if: $CI_COMMIT_TAG =~ /^develop-[\d]{4}-[\d]{2}-[\d]{2}$/ || $CI_COMMIT_TAG =~ /^v.*/
-      # Pipelines on tags (release or dev snapshots) only copy binaries from one mirror to another
+    - if: $CI_COMMIT_TAG =~ /^v.*/ || $CI_COMMIT_TAG =~ /^develop-[\d]{4}-[\d]{2}-[\d]{2}$/
       when: always
       variables:
-        SPACK_PIPELINE_TYPE: "spack_copy_only"
-        SPACK_COPY_BUILDCACHE: "${PROTECTED_MIRROR_PUSH_DOMAIN}/${CI_COMMIT_REF_NAME}"
-        PIPELINE_MIRROR_TEMPLATE: "copy-only-protected-mirrors.yaml.in"
-        OIDC_TOKEN_AUDIENCE: "protected_binary_mirror"
+        NOOP_JOB: 1
```
zackgalbreath, line 76: this should probably default to 0, right?

zackgalbreath (changes requested):

.ci/gitlab/.gitlab-ci.yml
bernhardkaindl: @kwryankrattiger mentioning you to notify you of the review (with a request to change)

kwryankrattiger: The checks that use this only check if it is set, it isn't defined in any of the other pipeline cases.

zackgalbreath (dismissed):

zackgalbreath (approved):

PR #832, opened 2025-07-28
zlib-ng: disable avxvnni for gcc@:11 

Chrismarsh:
With `gcc@:11`, `binutils@2.35` is too old to support the avx instructions emitted

```
  >> 202    /tmp/chris/374840/ccUPKedH.s:212: Error: unsupported instruction `vpdpbusd'
  >> 203    /tmp/chris/374840/ccUPKedH.s:216: Error: unsupported instruction `vpdpbusd'
  >> 204    /tmp/chris/374840/ccUPKedH.s:268: Error: unsupported instruction `vpdpbusd'
```

and so zlib-ng needs to be build with `CFLAGS="-mno-avxvnni"` (e.g., https://access.redhat.com/solutions/7049696)

### DIFF
--- a/repos/spack_repo/builtin/packages/zlib_ng/package.py
+++ b/repos/spack_repo/builtin/packages/zlib_ng/package.py
@@ -76,6 +76,12 @@ class ZlibNg(AutotoolsPackage, CMakePackage):
     def flag_handler(self, name, flags):
         if name == "cflags" and self.spec.satisfies("+pic build_system=autotools"):
             flags.append(self["c"].pic_flag)
+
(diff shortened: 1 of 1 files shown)


Chrismarsh:
@haampie follow up from the slack conversation

haampie (changes requested):
This is not a good idea, it's the combination of versions of `gcc` and `binutils` that happens to be incompatible on your system. This would make other users unhappy who have a working toolchain. What needs to be fixed is the zlib-ng detection of these compiler/assembler features.

PR #844, opened 2025-07-29
Fixes to cray* packages

alalazo:
This makes all `cray-*` packages consistent on:
- Having the same install method that raises
- depending on `c` to be able to specify a compiler on the externals
- requiring `linux`


repos/spack_repo/builtin/packages/cray_mpich/package.py
```diff
@@ -41,6 +41,8 @@ class CrayMpich(MpichEnvironmentModifications, Package, CudaPackage, ROCmPackage
     depends_on("cray-pmi")
     depends_on("libfabric")
 
+    depends_on("c", type="build")
```
haampie: leave a comment?

alalazo:
@spackbot fix style

repos/spack_repo/builtin/build_systems/cray.py
```diff
@@ -0,0 +1,23 @@
+# Copyright Spack Project Developers. See COPYRIGHT file for details.
+#
+# SPDX-License-Identifier: (Apache-2.0 OR MIT)
+from spack.package import InstallError, Package, depends_on, requires
+
+
+class CrayPackage(Package):
```
haampie: Maybe call it `CrayExternal`?

repos/spack_repo/builtin/build_systems/cray.py
```diff
@@ -0,0 +1,23 @@
...
+    """Specialization of Package that takes care of setting common metadata for Cray software"""
+
+    has_code = False  # Skip attempts to fetch a source that is not available
+
+    # Allows attaching compilers to externals in packages.yaml
+    depends_on("c", type="build")
```
haampie: I'm not sure if I like this mixin class. It's a bit of an arbitrary collection of features which are the common denominator of a weird concept of a "cray package".

* Maybe not all should depend on `c`?
* It assumes a `Package` base class, so requires a particular `build_system`.
* Cray packages exist that don't fit the pattern: `libcxi` is open source.

I don't mind if (a) there's a bit of repetition in the handful of packages and (b) cray is less special.

haampie (approved):

haampie:
Thanks for removing the base class. My issue with it is mostly that it's something we wish wasn't necessary, which then becomes impossible to remove even after it's redundant in builtin, because of "backward compat" reasons.

PR #361, opened 2025-06-25
py-gpaw: new maintainers and package versions

alikhamze:
This PR modifies the `py-gpaw` package with the following changes:
- Adds myself and @Chronum94 as maintainers.
- Adds the last 2 years of GPAW versions (including the most recent) and deprecates old versions.
- Cleaner dependency handling for specific versions.
- Adds an elpa variant for increased performance.
- Adds an openmp variant.
- Updates the license information.
- Removes v1.3 and the associated patch since Python 2 is no longer supported/distributed by spack.

This is a manually replicated PR of  https://github.com/spack/spack/pull/50563 , which the migrate tool failed to handle.

PR #849, opened 2025-07-29
py_ase: new versions and maintainer

alikhamze:
Updates:
- Adds versions 3.24.0 and 3.25.0
- Adds myself as a maintainer.
- Use the `when` context manager for dependencies for improved maintainability and readability.
- Raises minimum python version to 3.4 since python 2 is no longer supported (only relevant for the oldest versions)

### DIFF
--- a/repos/spack_repo/builtin/packages/py_ase/package.py
+++ b/repos/spack_repo/builtin/packages/py_ase/package.py
@@ -17,6 +17,8 @@ class PyAse(PythonPackage):
 
     license("LGPL-2.1-or-later")
 
+    version("3.25.0", sha256="374cf8ca9fe588f05d6e856da3c9c17ef262dc968027b231d449334140c962c2")
+    version("3.24.0", sha256="9acc93d6daaf48cd27b844c56f8bf49428b9db0542faa3cc30d9d5b8e1842195")
     version("3.23.0", sha256="91a2aa31d89bd90b0efdfe4a7e84264f32828b2abfc9f38e65e041ad76fec8ae")
     version("3.21.1", sha256="78b01d88529d5f604e76bc64be102d48f058ca50faad72ac740d717545711c7b")
     version("3.21.0", sha256="2c561e9b767cf16fc8ce198ea9326d77c6b67d33a85f44b68455e23466a64608")
@@ -34,7 +36,8 @@ class PyAse(PythonPackage):
     depends_on("python@2.6:", type=("build", "run"), when="@:3.15.0")
     depends_on("python@3.5:", type=("build", "run"), when="@3.18.0:")
     depends_on("python@3.6:", type=("build", "run"), when="@3.20.0:")
-    depends_on("python@3.8:", type=("build", "run"), when="@3.23.0:")
+    depends_on("python@3.8:", type=("build", "run"), when="@3.23.0")
+    depends_on("python@3.9:", type=("build", "run"), when="@3.24.0:")
     depends_on("py-numpy@1.11.3:", type=("build", "run"))
     depends_on("py-numpy@1.18.5:", type=("build", "run"), when="@3.23.0:")
     depends_on("py-matplotlib@2.0.0:", type=("build", "run"))
--- a/repos/spack_repo/builtin/packages/py_ase/package.py
+++ b/repos/spack_repo/builtin/packages/py_ase/package.py
@@ -42,13 +42,6 @@ class PyAse(PythonPackage):
         depends_on("py-matplotlib@3.3.4:", type=("build", "run"))
         depends_on("py-setuptools@61.0.0:", type="build")
 
-    with when("@3.24.0"):
-        depends_on("python@3.9:", type=("build", "run"))
-        depends_on("py-numpy@1.19.5:", type=("build", "run"))
-        depends_on("py-scipy@1.6.0:", type=("build", "run"))
-        depends_on("py-matplotlib@3.3.4:", type=("build", "run"))
-        depends_on("py-setuptools@61.0.0:", type="build")
-
     with when("@3.23.0"):
         depends_on("python@3.8:", type=("build", "run"))
         depends_on("py-numpy@1.18.5:", type=("build", "run"))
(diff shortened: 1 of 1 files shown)


alikhamze:
This will fix the failing CI in #361 after the latest commits to add GPAW@25.7.0.

repos/spack_repo/builtin/packages/py_ase/package.py
```diff
@@ -34,7 +38,8 @@ class PyAse(PythonPackage):
     depends_on("python@2.6:", type=("build", "run"), when="@:3.15.0")
     depends_on("python@3.5:", type=("build", "run"), when="@3.18.0:")
     depends_on("python@3.6:", type=("build", "run"), when="@3.20.0:")
-    depends_on("python@3.8:", type=("build", "run"), when="@3.23.0:")
+    depends_on("python@3.8:", type=("build", "run"), when="@3.23.0")
+    depends_on("python@3.9:", type=("build", "run"), when="@3.24.0:")
```
tldahlgren: A couple of the dependencies either need lower bounds or updated lower bounds (https://spack.readthedocs.io/en/latest/packaging_guide_creation.html#conditional-dependencies) for the newer version(s) per `pyproject.toml`:

For`3.24.0` and `3.25.0`:

```
[build-system]
requires = ['setuptools >= 61.0.0']
...
dependencies = [
    'numpy>=1.19.5',  # January 2021
...
```

tldahlgren (changes requested):
Additional dependencies, conditioned on the version, are needed starting with `3.24.0`. See comment(s).

tldahlgren:
Very odd that `pypi.org`'s search doesn't appear to yield `ase` though they are hosting the package's files.

repos/spack_repo/builtin/packages/py_ase/package.py
alikhamze: Ah, I was just looking at the release notes and missed those, thanks for catching them! Pushed a fix and switched to using the when context manager for dependencies.

alikhamze:
I'm also not able to find it with pypi's search, that is very odd.

tldahlgren:
> Very odd that `pypi.org`'s search doesn't appear to yield `ase` though they are hosting the package's files.

Turns out there's a reindex going on:  https://status.python.org.

repos/spack_repo/builtin/packages/py_ase/package.py
```diff
@@ -31,15 +35,53 @@ class PyAse(PythonPackage):
...
+        depends_on("py-matplotlib@3.3.4:", type=("build", "run"))
+        depends_on("py-setuptools@61.0.0:", type="build")
+
+    with when("@3.24.0"):
+        depends_on("python@3.9:", type=("build", "run"))
+        depends_on("py-numpy@1.19.5:", type=("build", "run"))
+        depends_on("py-scipy@1.6.0:", type=("build", "run"))
+        depends_on("py-matplotlib@3.3.4:", type=("build", "run"))
+        depends_on("py-setuptools@61.0.0:", type="build")
+
```
tldahlgren: There is no reason to have this when the dependencies in the context above (lines 38-43) cover both new versions that have the same dependencies thanks to the use of the version range specified at line 38.

repos/spack_repo/builtin/packages/py_ase/package.py
```diff
@@ -31,15 +35,53 @@ class PyAse(PythonPackage):
...
-    depends_on("py-matplotlib@2.0.0:", type=("build", "run"))
-    depends_on("py-matplotlib@3.3.4:", type=("build", "run"), when="@3.23.0:")
-    depends_on("py-scipy@0.18.1:", type=("build", "run"))
-    depends_on("py-scipy@1.6.0:", type=("build", "run"), when="@3.23.0:")
-    depends_on("py-flask", type=("build", "run"), when="@:3.18.0")
-    depends_on("py-setuptools", type="build")
+    with when("@3.24.0:"):
+        depends_on("python@3.9:", type=("build", "run"))
+        depends_on("py-numpy@1.19.5:", type=("build", "run"))
+        depends_on("py-scipy@1.6.0:", type=("build", "run"))
```
tldahlgren: FWIW. You could drop the `.0` patch release and have the same results (https://spack.readthedocs.io/en/latest/packaging_guide_creation.html#specifying-versions-in-other-directives):

```suggestion
        depends_on("py-scipy@1.6:", type=("build", "run"))
```

repos/spack_repo/builtin/packages/py_ase/package.py
```diff
@@ -31,15 +35,53 @@ class PyAse(PythonPackage):
...
+        depends_on("py-matplotlib@3.3.4:", type=("build", "run"))
+        depends_on("py-setuptools@61.0.0:", type="build")
```
tldahlgren: Same here for minor and patch versions:

```suggestion
        depends_on("py-setuptools@61:", type="build")
```

repos/spack_repo/builtin/packages/py_ase/package.py
```diff
@@ -31,15 +35,53 @@ class PyAse(PythonPackage):
...
+
+    with when("@3.23.0"):
+        depends_on("python@3.8:", type=("build", "run"))
+        depends_on("py-numpy@1.18.5:", type=("build", "run"))
+        depends_on("py-scipy@1.6.0:", type=("build", "run"))
+        depends_on("py-matplotlib@3.3.4:", type=("build", "run"))
+        depends_on("py-setuptools@61.0.0:", type="build")
+
+    with when("@3.21.0:3.21.1"):
```
tldahlgren: FWIW. *Probably* safe to use the simpler range:

```suggestion
    with when("@3.21"):
```

tldahlgren (changes requested):
There some duplication and optional simplification. See comments for details.

repos/spack_repo/builtin/packages/py_ase/package.py
alikhamze: Ack, my mistake!

repos/spack_repo/builtin/packages/py_ase/package.py
alikhamze: Done, will keep this in mind for the future!

tldahlgren (approved):
Thanks.

alikhamze:
Thank you for the feedback!

PR #855, opened 2025-07-29
verible: add v0.0.4013

davekeeshan:
add v0.0.4013

### DIFF
--- a/repos/spack_repo/builtin/packages/verible/package.py
+++ b/repos/spack_repo/builtin/packages/verible/package.py
@@ -118,8 +118,6 @@ class Verible(Package):
         url="https://github.com/chipsalliance/verible/archive/refs/tags/v0.0-3428-gcfcbb82b.tar.gz",
     )
 
-    depends_on("cxx", type="build")  # generated
-
     depends_on("flex", type="build")
     depends_on("bison", type="build")
     depends_on("bazel", type="build")
--- a/repos/spack_repo/builtin/packages/verible/package.py
+++ b/repos/spack_repo/builtin/packages/verible/package.py
@@ -39,7 +39,7 @@ class Verible(Package):
 
     version(
         "0.0.4013",
-        sha256="1b282233b91fc6884708f76232dcc3b38d47a14c783c5a0b8035bb23bcb8dbc7",
+        sha256="703766302083af76eeb3e545683017643ff24e4184e3195296623992eb55e4b8.",
         url="https://github.com/chipsalliance/verible/archive/refs/tags/v0.0-4013-gba3dc371.tar.gz",
     )
     version(
--- a/repos/spack_repo/builtin/packages/verible/package.py
+++ b/repos/spack_repo/builtin/packages/verible/package.py
@@ -38,6 +38,11 @@ class Verible(Package):
     version("master", branch="master")
 
     version(
+        "0.0.4013",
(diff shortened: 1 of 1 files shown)


alalazo (approved):

repos/spack_repo/builtin/packages/verible/package.py
```diff
@@ -118,8 +123,6 @@ class Verible(Package):
         url="https://github.com/chipsalliance/verible/archive/refs/tags/v0.0-3428-gcfcbb82b.tar.gz",
     )
 
-    depends_on("cxx", type="build")  # generated
```
alalazo, line 121: Why was this removed? Doesn't this package need a compiler?

alalazo (changes requested):

repos/spack_repo/builtin/packages/verible/package.py
davekeeshan: I was getting this complaint:

…gomez/.cache/bazel/_bazel_gomez/8f60070842ad6e764e3148c914321bd5/external/rules_bison~~default_toolchain_ext~bison_v3.3.2/BUILD.bazel:57:11: Compiling src/bazel_runfiles.cc [for tool] failed: (Exit 2): gcc failed: error executing CppCompile command (from target @@@@rules_bison~~default_toolchain_ext~bison_v3.3.2//:bison_bazel_runfiles) /cadtools/spack-2025/opt/linux-ubuntu24.04-skylake/none-none/compiler-wrapper/1.0-7jwusgfgu3whgccg6col46kp7qmvxsho/libexec/spack/gcc/gcc -U_FORTIFY_SOURCE -fstack-protector -Wall -Wthread-safety ... (remaining 43 arguments skipped)


Which disappear and compiled successfully when I removed that line.

alalazo: If you compile C++ sources you need to declare:
```
```
for Spack to keep track of runtime requirements. If you compile C sources:
```
depends_on("c", type="build")
```
etc.

davekeeshan: How is that related to the problem I saw, and the fix I made?

alalazo: You removed any dependency on a compiled language from a package that needs to compile things. That means the package is modeled wrongly (and Spack compiler wrapper is not injected).

davekeeshan: But the bit I am lost on is that it failed to compile with the line, and successfully compiled without it, what am I missing?

davekeeshan: OK I looked into this, with the help of some AI, and the problem is not the cxx, it was the absense of depends_on('c'). When the bulk add of all the depends on happened, only cxx was added to the verible package file:

https://github.com/spack/spack/commit/7b9f8abce5cee74546a6f588f88c6f353170d52b#diff-ed47cd496dd3d7abd8a235df9c9f8810d1d79bd54ff779fad126f6096094764b

At the time I didn't understand why.  But with some research is that bazel appears to have its own internal c compiler is one isn't defined, `abseil-cpp`, and it was falling back to this.  Why it all worked without cxx too is unclear, probably falling back to the bazel one too, so arguably it is safe to have none, or both, but only having one of the 2 is where problems surface.

Any way fix is in.

alalazo (approved):

PR #861, opened 2025-07-29
podman: Also install bin/rootlessport

scothalverson:


bernhardkaindl (approved):

bernhardkaindl (commented):
@scothalverson, rootlessport is already installed, even without this patch, but to libexec/podman/rootlessport:

```py
ls -l `find * -name rootlessport`
-rwxr-xr-x 1 5963800 Aug  8 12:46 podman-4.9.3-3q7olprgfcihwyx6hjgpzfz7hjn2vf2z/bin/rootlessport
-rwxr-xr-x 1 5963800 Aug  8 12:46 podman-4.9.3-3q7olprgfcihwyx6hjgpzfz7hjn2vf2z/libexec/podman/rootlessport
-rwxr-xr-x 1 4629922 Aug  8 12:43 podman-5.5.1-h27dnid27hcprmyjthljas3sqrkq3mae/libexec/podman/rootlessport
```
If tooling needs as bin/rootlessport, why not add it as a relative symlink to libexec/podman/rootlessport, which you can do in python as well as `shutil.copy2` currently.

scothalverson:
> @scothalverson, rootlessport is already installed, even without this patch, but to libexec/podman/rootlessport:
> 
> ```python
> ls -l `find * -name rootlessport`
> -rwxr-xr-x 1 5963800 Aug  8 12:46 podman-4.9.3-3q7olprgfcihwyx6hjgpzfz7hjn2vf2z/bin/rootlessport
> -rwxr-xr-x 1 5963800 Aug  8 12:46 podman-4.9.3-3q7olprgfcihwyx6hjgpzfz7hjn2vf2z/libexec/podman/rootlessport
> -rwxr-xr-x 1 4629922 Aug  8 12:43 podman-5.5.1-h27dnid27hcprmyjthljas3sqrkq3mae/libexec/podman/rootlessport
> ```
> 
> If tooling needs as bin/rootlessport, why not add it as a relative symlink to libexec/podman/rootlessport, which you can do in python as well as `shutil.copy2` currently.

Ahh, I swear I looked for it but didn't find it.  I think it doesn't end up in PATH when its in libexec, though.  Will test further and either amend or get rid of this PR.

bernhardkaindl (commented):
@scothalverson Yes, `libexec` would be correct if it should NOT be in the PATH for regular users, but should only be called internally by podman or podman scripts/tools.

Installing (or moving) it to the bin directory would be justified if it is (only) called by users and(or not by podman or podman scripts that know where to find it).

PR #863, opened 2025-07-29
MsMpi: Support stage and prefix with spaces

johnwparent:
Patches the `networkdirect` and `msmpi` build systems to be compatible with being built and installed in stages and prefixes whose paths contain spaces

Adds mpi to the Windows gitlab CI by enabling the `+mpi` variant of `vtk`.
Adds a version of the `win-wdk` to the GL CI `packages.yaml` that corresponds to the existing `win-sdk` version to ensure we don't get broken concretizations like https://gitlab.spack.io/spack/spack/-/jobs/16911750#L135


repos/spack_repo/builtin/packages/seacas/package.py
```diff
@@ -334,6 +337,9 @@ class Seacas(CMakePackage):
         when="@:2023-10-24",
     )
 
+    patch("2022_ioss_win_defines.patch", when="@:2022-10-14")
```
kwryankrattiger: windows only patches?

johnwparent: Yes, they're just preliminary at this point though, this almost certainly needs a lower bound, and presumably there's some upstream patch that also does this, I'm just trying to get _a_ passing build before cleaning up.

repos/spack_repo/builtin/packages/networkdirect/package.py
```diff
@@ -35,6 +35,14 @@ class Networkdirect(msbuild.MSBuildPackage):
     # CBT is entirely deprecated, and fully incompatible with modern dotnet versions
     # so we disable the CBT system and drive the underlying MSBuild system directly
     patch("no_cbt.patch")
+    # MSBuild's native handling of OutputPath does not handle spaces in paths well
```
kwryankrattiger, line 38:
```suggestion
    with when("platform=windows"):
        # MSBuild's native handling of OutputPath does not handle spaces in paths well
```

johnwparent: This package can't build anywhere else it's a Windows driver package, `platform=windows` is implied. 

It should have a requires statement if it doesn't.

tldahlgren: FYI. It does have the `requires` at line 26.

tldahlgren (commented):
Any chance of adding the fact that `networkdirect` is a  `Windows` package to the docstring?

repos/spack_repo/builtin/packages/networkdirect/package.py
johnwparent: Thanks @tldahlgren! (I would have been surprised if I had neglected to prevent this from being selected on other platforms)

johnwparent:
> Any chance of adding the fact that `networkdirect` is a `Windows` package to the docstring?

No problem, looking like I'm going to have to keep hammering at `seacas+mpi` for a bit longer anyway, so any/all suggested changes are still more than welcome!

johnwparent:
@spackbot fix style

repos/spack_repo/builtin/packages/seacas/package.py
johnwparent: Turns out I didn't need them, the `cgns` seacas vendors just doesn't work w/ parallel enabled on Windows, so attempting to patch it was never the right approach.

johnwparent:
@kwryankrattiger @scheibelp gitlab passed after the last commit, ready for review, PTAL at your convenience.

alecbcs:
Apparently out of the loop, what is PTAL?

kwryankrattiger:
> What is PTAL?

Please Take A Look

.ci/gitlab/configs/win64/packages.yaml
```diff
@@ -23,3 +23,8 @@ packages:
     - spec: win-sdk@10.0.22621 plat=x64
       prefix: "C:\\Program Files (x86)\\Windows Kits\\10"
     buildable: False
+  win-wdk:
+    externals:
+    - spec: win-wdk@10.0.22621 plat=x64
+      prefix: "C:\\Program Files (x86)\\Windows Kits\\10"
+    buildable: False
```
scheibelp, line 30: Is this external still required if win-wdk adds a new version (to match whatever new version exists in the CI image)?

repos/spack_repo/builtin/packages/seacas/package.py
```diff
@@ -269,7 +274,10 @@ class Seacas(CMakePackage):
     depends_on("trilinos~exodus+mpi+pamgen", when="+mpi+pamgen")
     depends_on("trilinos~exodus~mpi+pamgen", when="~mpi+pamgen")
     # Always depends on netcdf-c
-    depends_on("netcdf-c@4.8.0:+mpi+parallel-netcdf", when="+mpi")
+    depends_on("netcdf-c@4.8.0:+mpi", when="+mpi")
+    depends_on("netcdf-c+parallel-netcdf", when="+mpi platform=linux")
+    depends_on("netcdf-c+parallel-netcdf", when="+mpi platform=darwin")
+    depends_on("netcdf-c+parallel-netcdf", when="+mpi platform=freebsd")
```
scheibelp, line 280: (minor) loop would be good

zackgalbreath (dismissed):

repos/spack_repo/builtin/packages/msmpi/package.py
```diff
@@ -114,5 +123,21 @@ def install(self, pkg, spec, prefix):
             install_path = x.replace(build_configuration, prefix)
             mkdirp(os.path.dirname(install_path))
             install(x, install_path)
+
+        # construct more traditional install layout
+        # grab all .lib .exp and .pdb files from their components specific layouts
+        # glob statement matches any of the above.
+        # symlink all linking related components (libs, .exp, etc)
+        # to lib
+        def install_to_lib(file_type):
```
scheibelp: Can you rename this? This isn't installing things, it is only symlinking things that have already been installed.

repos/spack_repo/builtin/packages/netcdf_c/strip_csd_from_mpi_inc.patch
```diff
@@ -0,0 +1,14 @@
...
+index de95010c6..3edd8a520 100644
+--- a/CMakeLists.txt
++++ b/CMakeLists.txt
+@@ -1492,6 +1492,9 @@ IF(ENABLE_PARALLEL4 AND ENABLE_HDF5)
+     SET(USE_PARALLEL ON CACHE BOOL "")
+     SET(USE_PARALLEL4 ON CACHE BOOL "")
+     SET(STATUS_PARALLEL "ON")
++    if(MPI_C_INCLUDE_PATH)
++      list(REMOVE_ITEM MPI_C_INCLUDE_PATH ${CMAKE_CURRENT_SOURCE_DIR})
++    endif()
```
scheibelp, line 11: (question - no change requested) what does this do?

scheibelp (changes requested):
I have a couple requests and a question.

.ci/gitlab/configs/win64/packages.yaml
johnwparent: Yes, for the same reason the SDK entry above is required, to avoid a call to `spack external find` during CI runs and to ensure we target the MSVC, SDK, and WDK that allows us to ensure compatibility with the most MSVC versions.

repos/spack_repo/builtin/packages/seacas/package.py
johnwparent: I personally find the loops less readable/obvious. I don't feel particularly strongly, but IMO we should decide on one consistent approach and now's as good a time as any to discuss.

repos/spack_repo/builtin/packages/msmpi/package.py
johnwparent: Sure

repos/spack_repo/builtin/packages/netcdf_c/strip_csd_from_mpi_inc.patch
johnwparent: It removes a path to the current source directory from what eventually becomes the INTERFACE_INCLUDE_DIRECTORIES of anything that uses that list of includes. AFAICT the current source directory gets conditionally included by the FindMPI module, so we just have to remove it here, otherwise CMake generate will throw an error.

.ci/gitlab/configs/win64/packages.yaml
scheibelp: Apologies, I'm confused because:

* there was no `win-wdk` entry before
* I recall speaking w/ you about a concretization error that occurs without this and
* I think the constraints in the package suggested to me that adding `version("10.0.22621",` would fix that concretization error

Does the concretization pass fine without this though?

repos/spack_repo/builtin/packages/seacas/package.py
scheibelp: Interesting, I thought we were using loops over platforms in most places (and had discussed a way to wrap it into some sort of helper function); something like

```
with when("+mpi"):
  for x in ["linux"...]:
    depends_on(..., when="platform={x}")
```

seems worse to you? As it is, when long constraints like this are duplicated, I feel like I have to read each line carefully (e.g. to make sure it's `netcdf-c+parallel-netcdf` in all cases instead of `netcdf-c~parallel-netcdf`)

.ci/gitlab/configs/win64/packages.yaml
johnwparent: > there was no win-wdk entry before

There is a `win-sdk` entry

> I think the constraints in the package suggested to me that adding version("10.0.22621", would fix that concretization error

yes, it would, however we still want to use the external WDK that's setup on the runner image rather than install our own, and so rather than run an expensive find package, we do the same thing we've been doing for the SDK.

repos/spack_repo/builtin/packages/seacas/package.py
johnwparent: > when long constraints like this are duplicated, I feel like I have to read each line carefully

When reviewing I agree, it is more painstaking when written like this, but when reading as a consumer of the package, I find not having to unroll a loop more directly overt.

> I thought we were using loops over platforms in most places 

I have been taking that approach previously, but I think those have been undone in some places and I've seen other folks take this approach and I just found it more readable. IMO an abstraction function would be as readable as this more verbose approach (something semantically obvious like `depends_on_without_windows` or similar) and would be easier to review. Perhaps we chose one approach here and then work on the abstraction?

scheibelp: OK in this case it's fine to leave this as-is.

johnwparent: Sounds good, will start the abstraction work today.

repos/spack_repo/builtin/packages/netcdf_c/strip_csd_from_mpi_inc.patch
scheibelp: Sorry I guess what I'm curious about is what the error is: I assume that adding this directory is a default CMake behavior; would it be expected to occur (and cause problems) for other packages?

johnwparent: Adding that directory seems to be a quirk of using HDF5+mpi and MSMPI w/ CMake's `FindMPI`. I've never observed previously and googling (admittedly a pretty quick pass) shows the only presentation of this scenario where it's not user error is VCPKG reporting this exact issue with netcdf-c when using Hdf5 w/ msmpi.

It's not really generalizable as (I believe) we'd have to patch CMake, but since a lot of CMakes are system packages, that's not really viable. That said this could potentially come from hdf5, if you want I can investigate more. I basically just evaluated this until I got to this obvious solution, I could press the matter further.

scheibelp: I think putting some of these details in a comment on the patch declaration might be useful (or maybe a short description with a reference to a comment in the main thread, since this will need to collapse).

johnwparent: I've added that

repos/spack_repo/builtin/packages/msmpi/package.py
johnwparent: Renamed to `link_to_lib`

scheibelp (dismissed):

johnwparent:
Spurious Gitlab pipeline failures (e4s-rocm execution timeout): manually restarted only failing jobs just now

scheibelp (approved):

PR #864, opened 2025-07-30
PRRTE: Q to @tgambin: How to don't use /usr for external packages

nhanford:
Hello @haampie and @tgamblin,

Edit by @bernhardkaindl: The first paragraph was already merged with #917:

First, according to the PRRTE [docs](https://docs.prrte.org/en/v4.0.0/install.html#minimum-pmix-version) and the configure file itself, PRRTE 4+ requires PMIx 6+. I have added the relevant dependency requirement to avoid build failures when an older PMIx is on the system.

Second, for a variety of reasons, we really prefer to build our own hwloc and libevent as externals. Unfortunately, due to Spack's external logic, the `/usr` path gets inserted for these by default. This leads to an autotools race condition where the general `/usr/lib64` path is put on the link line first, where the compiler also finds and links to an older system default PMIx. Therefore, I have added some logic to remove these spurious paths when they are not necessary. However, this is part of a broader discussion on how this should be handled in Spack as the method I use to do this is deprecated.

This has been tested as part of a broader Open MPI installation on LLNL [Matrix](https://hpc.llnl.gov/hardware/compute-platforms/matrix) and appears to work correctly with some benchmarks and smoke tests.

Thanks,
Nate

@spackbot fix style

### DIFF
--- a/repos/spack_repo/builtin/packages/prrte/package.py
+++ b/repos/spack_repo/builtin/packages/prrte/package.py
@@ -83,11 +83,8 @@ class Prrte(AutotoolsPackage):
         spec = self.spec
         config_args = ["--enable-shared", "--enable-static", "--disable-sphinx"]
 
-        # libevent
-        config_args.append("--with-libevent={0}".format(spec["libevent"].prefix))
-        # hwloc
-        config_args.append("--with-hwloc={0}".format(spec["hwloc"].prefix))
-        # pmix
-        config_args.append("--with-pmix={0}".format(spec["pmix"].prefix))
+        for dep in ["libevent", "hwloc", "pmix"]:
+            prefix = spec[dep].prefix
+            config_args.append(f"--with-{dep}" if is_system_path(prefix) else f"--with-{dep}={prefix}")
 
         return config_args
--- a/repos/spack_repo/builtin/packages/prrte/package.py
+++ b/repos/spack_repo/builtin/packages/prrte/package.py
@@ -85,6 +85,8 @@ class Prrte(AutotoolsPackage):
 
         for dep in ["libevent", "hwloc", "pmix"]:
             prefix = spec[dep].prefix
-            config_args.append(f"--with-{dep}" if is_system_path(prefix) else f"--with-{dep}={prefix}")
+            config_args.append(
+                f"--with-{dep}" if is_system_path(prefix) else f"--with-{dep}={prefix}"
+            )
 
         return config_args


nhanford:
@spackbot fix style

repos/spack_repo/builtin/packages/prrte/package.py
```diff
@@ -83,11 +83,10 @@ def configure_args(self):
...
-        config_args.append("--with-libevent={0}".format(spec["libevent"].prefix))
-        # hwloc
-        config_args.append("--with-hwloc={0}".format(spec["hwloc"].prefix))
-        # pmix
-        config_args.append("--with-pmix={0}".format(spec["pmix"].prefix))
+        for dep in ["libevent", "hwloc", "pmix"]:
+            prefix = spec[dep].prefix
+            config_args.append(
+                f"--with-{dep}" if is_system_path(prefix) else f"--with-{dep}={prefix}"
+            )
```
bernhardkaindl, line 90: @nhanford, did you have a chance to talk with Todd about this?

I tried to understand what you want, and while less Pythonic, writing it in a more traditional C-like style would clarify the workaround you like to use:
```suggestion
            prefix = spec[dep].prefix
            # Workaround for spack adding /usr with externals, we want non-system externals:
            if is_system_path(prefix):
                config_args.append(f"--with-{dep}")  # Don't set the /usr prefix
            else
                config_args.append(f"--with-{dep}={prefix}")
```

I think the point that this raises is that, as far as I understand, `spack external find` looks in the system for externals, and adds the prefix it found to your spack packages.yaml:
```yaml
  perl:
    externals:
    - spec: perl@5.34.0~cpanm+opcode+open+shared+threads
      prefix: /usr
```
How do you add your external custom-compiled hwloc and other libraries to the build?
I'd assume that spack should provide you the `prefix:` from `packages.yaml` as the `spec[dep].prefix`, and if you could edit `packages.yaml` to define the prefix you need, you should get this prefix into your build.

Does `spack` add `/usr` to the search path even then?
Could you add a more detailed example of what the problem is, if you like to keep this PR open for the public so it could be closer understood?

Or could you contact Todd directly on this if this does not lead you closer to solving the issue in general?

bernhardkaindl (commented):
.

repos/spack_repo/builtin/packages/prrte/package.py
nhanford: Hello @bernhardkaindl Yes I would like to keep this open for comment and tracking. The issue is as follows:
- I have an hwloc in a non-standard directory that I would like to use.
- I also have an hwloc in `/usr` that I **do not** want to use.
- I have a libevent in `/usr` that I **would** like to use.

Spack superfluously adds `/usr` to the linker search path because of libevent, but this causes the linker to see the wrong hwloc. The "natural" autotools behavior if `/usr` wasn't specified would be to add it at the end so that the correct (specified) hwloc gets linked, but because we overspecified `/usr` in configure for libevent, we break this behavior, causing linking errors.

bernhardkaindl (commented):
@nhanford, just a comment as reminder for this PR for you.

I'm unsubscribing from this PR as I'm out available time to watch it.
I'm hoping that's fine for you, as I understood you have other contacts.

repos/spack_repo/builtin/packages/prrte/package.py
bernhardkaindl: @nhanford, any updates?

nhanford: I think @tgamblin and @becker33 would probably want to weigh in on a possible solution to this issue...

PR #496, opened 2025-07-07
MFEM: add new version v4.8

v-dobrev:
Migrated from https://github.com/spack/spack/pull/49871.

TODO:
- [x] After MFEM v4.8 is released, update the version number (`4.8.0-dev` -> `4.8.0`) and its source
- [x] Migrate to https://github.com/spack/spack-packages, e.g. using this tool: https://github.com/spack/migrate-package-prs
- [x] Add support for building `mfem@develop`
- [x] Add `enzyme` variant
  - [x] CPU builds
  - [x] CUDA builds
  - [x] HIP builds

OPTIONAL:
- [x] [POSTPONED] ~~Add `magma` variant~~

PR #869, opened 2025-07-30
Update toolchain for radiuss stack

adrienbernede:
An old compiler is used in radiuss stack. We update it by changing the container used.
In particular, this allows us to activate MFEM again. (see https://github.com/spack/spack-packages/pull/496)


adrienbernede:
@v-dobrev for the records.

repos/spack_repo/builtin/packages/papi/package.py
```diff
@@ -74,6 +74,7 @@ class Papi(AutotoolsPackage, ROCmPackage):
     depends_on("cxx", type="build")
     depends_on("fortran", type="build")
 
+    depends_on("perl", type="build")
```
adrienbernede: @G-Ragghianti this dependency was added after failing to build PAPI on a centos7 container that did not have perl installed.
The dependency should only apply for versions 7.1.0 and later, which I will fix right now.
This relates to the fix I pushed in PAPI, which has been merged:
https://github.com/icl-utk-edu/papi/pull/464

repos/spack_repo/builtin/packages/papi/package.py
```diff
@@ -74,6 +74,7 @@ class Papi(AutotoolsPackage, ROCmPackage):
     depends_on("cxx", type="build")
     depends_on("fortran", type="build")
 
+    depends_on("perl", when="@7.1.0:", type="build")
```
adrienbernede, line 77: @G-Ragghianti this dependency was added after failing to build PAPI on a centos7 container that did not have perl installed.
This relates to the fix I pushed in PAPI, which has been merged:
https://github.com/icl-utk-edu/papi/pull/464

G-Ragghianti: OK, looks good.  Thanks.

stacks/radiuss/spack.yaml
```diff
@@ -26,11 +31,12 @@ spack:
   - hypre
   - lbann
   - lvarray ~tests  # per Spack issue #23192  # ~examples
-  # - mfem  # does not work with gcc 7
+  - mfem  # does not work with gcc 7
```
v-dobrev, line 34: Thanks @adrienbernede!

Just FYI: `mfem` is also disabled in the stacks `radiuss-aws-aarch64` and `radiuss-aws` for the same reason -- it does not compile with gcc 7 (compiler crash).

becker33 (approved):

PR #879, opened 2025-07-30
trilinos: update for latest kokkos ecosystem release

lucbv:
master and develop need to pull kokkos 4.6.02 not 4.6.01


cgcgcg (approved):

psakievich (approved):

psakievich:
Musings: It would be cool if this could be `@4.6` and then the concretizer could absorb the latest patch release, but I guess that can't work because then `@4.6.01` would still be valid.

lucbv:
Yeah, the other thing is that Tpetra checks for an exact version of Kokkos and we need to match that otherwise it would fail when CMake is configuring the build. We have a Trilinos CI build which spack that failed in one of my PR so I noticed and created this PR...

PR #883, opened 2025-07-31
py-debugpy: add v1.8.15

manuelakuhn:
https://github.com/microsoft/debugpy/tree/v1.8.15

Also the packages on pypi switched from zip to tar balls.


bernhardkaindl (approved):
Builds on Ubuntu

PR #885, opened 2025-07-31
ci: use `python -m pip install` instead of `pip install`

tgamblin:
The windows audits (at least) has started to fail with an error if we invoke pip as `pip install` and not with `python -m pip install`.

- [x] Update CI to use `python -m pip install`

### DIFF
--- a/.github/workflows/audit.yaml
+++ b/.github/workflows/audit.yaml
@@ -31,7 +31,7 @@ jobs:
         python-version: ${{inputs.python_version}}
     - name: Install Python packages
       run: |
-        pip install --upgrade pip setuptools pytest coverage[toml]
+        python -m pip install --upgrade pip setuptools pytest coverage[toml]
     - name: Setup for Windows run
       if: runner.os == 'Windows'
       run: |


.github/workflows/audit.yaml
```diff
@@ -31,7 +31,7 @@ jobs:
         python-version: ${{inputs.python_version}}
     - name: Install Python packages
       run: |
-        pip install --upgrade pip setuptools pytest coverage[toml]
+        python -m pip install --upgrade pip setuptools pytest coverage[toml]
```
haampie, line 34:
```suggestion
        python3 -m pip install --upgrade pip setuptools pytest coverage[toml]
```

haampie:
do you know why `pip` fails?

.github/workflows/audit.yaml
adamjstewart: I recently learned that `python3` is not guaranteed to exist on Windows (while `python` is not guaranteed to exist on UNIX).

haampie: curious...

haampie (approved):

tgamblin:
> do you know why pip fails?

I initially thought this was a stricter warning in newer python, but the python version is the same on ubuntu and windows. 

I don't know why it only started failing now, but apparently it's failing specifically on windows because using:

```
pip install 
```

instead of:

```
python -m pip install
```

means `pip.exe` can't be upgraded because `pip.exe` is already running (something Linux is fine with). Running `python -m pip` circumvents this and allows pip to modify "itself".

.github/workflows/audit.yaml
tgamblin: We could fix this but I *believe* the GitHub runners symlink python on unix for this reason. 
image

Sigh.

manuelakuhn:
The pipeline fails because the `bash` packages was re-uploaded yesterday. PR #887 fixes `bash` but runs into the problem this PR here tries to fix.

tgamblin:
Since the thing this is trying to fix passes already, we can force-merge this and let @manuelakuhn rebase #887 so it can hopefully fix the pipeline.

becker33:
I think we should we backport this to v1.0.1. It may come up in release CI for future 1.0.x releases

manuelakuhn:
I rebased but the pipeline fails because the checks still use the old checksum

tgamblin:
> I think we should we backport this to v1.0.1.

`v2025.07`? added the label.

PR #887, opened 2025-07-31
bash: fix checksum and add patches

manuelakuhn:
The bash checksum changed because of a reupload see
 https://lists.gnu.org/archive/html/bug-bash/2025-07/msg00145.html

While on it, I also added the new patches.


michaelkuhn (approved):
Thanks!

PR #898, opened 2025-07-31
py-xhistogram: fix python 3.12

Chrismarsh:
Fixes compatibility with python 3.12. 
Patch is from a PR I submitted. The project doesn't look very active so I don't think it's worth waiting for this to be merged in

### DIFF
--- a/repos/spack_repo/builtin/packages/py_xhistogram/package.py
+++ b/repos/spack_repo/builtin/packages/py_xhistogram/package.py
@@ -26,8 +26,8 @@ class PyXhistogram(PythonPackage):
     depends_on("py-numpy@1.17:", type=("build", "run"))
 
     # compatibility with python 3.12
+    # https://github.com/xgcm/xhistogram/pull/90
     patch(
-        "https://github.com/xgcm/xhistogram/commit/6eef6c697d95ea70883a5ff6ee2f3e7188eaa4c5.patch?full_index=1",
-        sha256="2a508aded3ac57dfe7c87ff3e27124400daf04391759009fea209ce5cb0067fe",
+        "patch_py312_versioneer.patch",
         when="^python@3.12:",
     )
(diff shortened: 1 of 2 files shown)


Chrismarsh:
@spackbot fix style

repos/spack_repo/builtin/packages/py_xhistogram/package.py
```diff
@@ -18,6 +18,16 @@ class PyXhistogram(PythonPackage):
...
+
     depends_on("py-xarray@0.12:", type=("build", "run"))
     depends_on("py-dask@2.3:+array", type=("build", "run"))
     depends_on("py-numpy@1.17:", type=("build", "run"))
+
+    # compatibility with python 3.12
+    patch(
+        "https://github.com/xgcm/xhistogram/commit/6eef6c697d95ea70883a5ff6ee2f3e7188eaa4c5.patch?full_index=1",
+        sha256="2a508aded3ac57dfe7c87ff3e27124400daf04391759009fea209ce5cb0067fe",
+        when="^python@3.12",
```
wdconinc: We should avoid commits on forks. We should probably version guard patches so they don't break on the next package upgrade. Does this only apply to python 3.12, or also higher?

Chrismarsh: What is best approach on this? standalone patch file? It is `3.12:` afaik -- good point!

wdconinc: Depends. What's the timescale when this might get merged?

Chrismarsh: The last code update was 3 years ago and there are 2 year old open PRs. So, I don't know, but I worry "not any time soon"

wdconinc: In that case I think a local copy seems appropriate (it's small anyway).

Chrismarsh: I switched it to a local patch file

Chrismarsh:
@spackbot fix style

wdconinc (approved):
Looks good to me.

PR #904, opened 2025-07-31
petsc, py-petsc4py: add v3.23.5

balay:


alecbcs:
@balay do we not need to update the `petsc` dependency in `py-petsc4py` anymore to match versions?

balay:
Right now its set so that all v3.23.x of petsc4py matches v3.23.x of petsc.

But I guess its ok to tighten this dependency (for exact version match) - but thats for a different PR.

alecbcs (approved):
Sounds good to me. Thanks @balay!

PR #905, opened 2025-07-31
msvc: refactor env setup to use env module from sourcing

johnwparent:
Refactor msvc env setup approach


johnwparent:
@spackbot fix style

johnwparent:
@spackbot fix style

PR #907, opened 2025-08-01
py-prometheus-client: add v0.22.1

wu-kan:


repos/spack_repo/builtin/packages/py_prometheus_client/package.py
wdconinc, line 1: Note the new minimum version of setuptools required. This should be reflected in this update.

https://github.com/prometheus/client_python/compare/v0.17.1...v0.22.1#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711R2

wdconinc (commented):
Thanks for the update. One comment.

repos/spack_repo/builtin/packages/py_prometheus_client/package.py
wu-kan: Done

wdconinc (approved):
Thanks. Looks good to me.

PR #909, opened 2025-08-01
py-transformers: add v4.54.1

wu-kan:


adamjstewart (changes requested):
Need to update dependencies too, see `setup.py`

PR #918, opened 2025-08-01
OpenBlas: add constraint for required versions of CMake when compiling with MSVC

johnwparent:
CMake 4.1 adds https://gitlab.kitware.com/cmake/cmake/-/merge_requests/10836 which is required to build openblas with MSVC + oneAPI

Presumably more packages will have this issue, it's not an issue with openblas' CMake, but openblas is the package that allowed us to create the fix for CMake, so it gets the constraint first as its the only one I'm currently aware of

### DIFF
--- a/repos/spack_repo/builtin/build_systems/cmake.py
+++ b/repos/spack_repo/builtin/build_systems/cmake.py
@@ -228,7 +228,7 @@ class CMakePackage(PackageBase):
         # must conflict.
         # this should be updated to reflect a oneapi fortran provider
         # once oneapi is usable with fortran on Windows
-        conflicts("cmake@:4.0", when="%fortran=msvc")
+        conflicts("cmake@:4.0", when="%cxx=msvc %fortran=msvc")
 
     def flags_to_build_system_args(self, flags):
         """Return a list of all command line arguments to pass the specified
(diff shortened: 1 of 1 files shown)


mathomp4 (approved):
Trust in @johnwparent as I have no idea how to test this :) Approve

johnwparent:
> Trust in @johnwparent as I have no idea how to test this :) Approve

As much as I appreciate that, if you did want to test:

On Spack Develop on Windows:
- build this package using oneAPI and MSVC using CMakes before 4.1 and 4.1, and verify it works for 4.1 (something like `spack install "openblas ^^cmake@4.0"` and then `spack install "openblas ^^cmake@4.1"`)

It should fail on the first and pass on the second. Basically we found a bug in CMake trying to build this on Windows, so the real "test" is actually in the CMake codebase (see linked CMake issue)

If you're unable to get your hands on a Windows machine, I can provide a number of suggestions on how to do that, from spinning up a VM to SSH-ing into a GHA runner.

mathomp4 (approved):

johnwparent:
@spackbot fix style

johnwparent:
@spackbot fix style

tldahlgren (approved):

PR #921, opened 2025-08-01
labels: add ci and tests labels

alecbcs:
Adds a `ci` and a `tests` label to PRs modifying CI/CD or testing files.


wdconinc:
Note that this has a conflict to be resolved.

wdconinc (approved):
Looks reasonable to me.

Only nitpick one could have is that a slash after `.ci` or `.github` might be required to avoid matching something like `.circleci/config.yml` (but even then the label `ci` would be appropriate).

PR #928, opened 2025-08-03
apfel: fix cmake version typo; add type=build

wdconinc:
This PR fixes a typo in the `apfel` build dependency on `cmake`. This causes the concretization errors in https://gitlab.spack.io/spack/spack-packages/-/jobs/17247852 (which don't actually print the offending version typo).


alecbcs (approved):

PR #949, opened 2025-08-05
py-horovod: fix non-determinism issue

haampie:
The package defines `tensor_ops=nccl`, `~cuda`, `~rocm` as variant
defaults, but also adds a conflict `tensor=nccl ~cuda ~rocm`.

This forces the concretizer to pick a solution like `tensor_ops=mpi` or
`tensor_ops=gloo`, which have an identical score, leading to
non-determinism.

### DIFF
--- a/repos/spack_repo/builtin/packages/py_horovod/package.py
+++ b/repos/spack_repo/builtin/packages/py_horovod/package.py
@@ -76,7 +76,7 @@ class PyHorovod(PythonPackage, CudaPackage):
     )
     variant(
         "tensor_ops",
-        default="nccl",
+        default="mpi",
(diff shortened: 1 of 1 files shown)


haampie:
ping @adamjstewart @aweits @tgaddair @thomas-bouvier

adamjstewart:
A bit confused because this package defines `+cuda` as a default, not `~cuda`. I would rather prefer `tensor_ops=nccl +cuda`.

haampie:
Hm, good point, I didn't see the redefinition. The environment in CI probably sets `~cuda`, leading to this non-determinism.

Then this should be closed, and it should be considered a bug: https://github.com/spack/spack/issues/51112

PR #950, opened 2025-08-05
netcdf-fortran: Make sure -fallow-argument-mismatch is added only for gfortran

msimberg:
With mixed compiler toolchains, e.g. `%c,cxx=gcc %fortran=nvfortran` the current setup would wrongly add `-fallow-argument-mismatch` to the nvfortran compiler flags even though it doesn't support the flag. This PR restricts the condition to only allow `-fallow-argument-mismatch` when the fortran compiler is gcc.


RMeli (approved):
LGTM. Leaving @skosukhin and @WardF the chance to comment before merging.

skosukhin (approved):

PR #961, opened 2025-08-05
Palace: update package.py for 0.13.0 and include 0.12.0 patch.

cameronrutherford:
This includes the local development that we have been testing for 0.13.0. 0.14.0 is on the way soon.

### DIFF
--- a/repos/spack_repo/builtin/packages/palace/package.py
+++ b/repos/spack_repo/builtin/packages/palace/package.py
@@ -157,12 +157,8 @@ class Palace(CMakePackage, CudaPackage, ROCmPackage):
         depends_on("sundials~cuda", when="~cuda")
         depends_on("sundials~rocm", when="~rocm")
 
-    conflicts(
-        "+cuda", when="@:0.13", msg="CUDA is only supported for Palace versions after 0.13"
-    )
-    conflicts(
-        "+rocm", when="@:0.13", msg="ROCm is only supported for Palace versions after 0.13"
-    )
+    conflicts("+cuda", when="@:0.13", msg="CUDA is only supported for Palace versions after 0.13")
+    conflicts("+rocm", when="@:0.13", msg="ROCm is only supported for Palace versions after 0.13")
     conflicts("+cuda+rocm", msg="PALACE_WITH_CUDA is not compatible with PALACE_WITH_HIP")
     conflicts(
         "cuda_arch=none", when="+cuda", msg="palace: Please specify a CUDA arch value / values"
--- a/repos/spack_repo/builtin/packages/palace/package.py
+++ b/repos/spack_repo/builtin/packages/palace/package.py
@@ -158,10 +158,10 @@ class Palace(CMakePackage, CudaPackage, ROCmPackage):
         depends_on("sundials~rocm", when="~rocm")
 
     conflicts(
-        "+cuda", when="@:0.12", msg="CUDA is only supported for Palace versions 0.13 and above"
+        "+cuda", when="@:0.13", msg="CUDA is only supported for Palace versions after 0.13"
     )
     conflicts(
-        "+rocm", when="@:0.12", msg="ROCm is only supported for Palace versions 0.13 and above"
+        "+rocm", when="@:0.13", msg="ROCm is only supported for Palace versions after 0.13"
     )
     conflicts("+cuda+rocm", msg="PALACE_WITH_CUDA is not compatible with PALACE_WITH_HIP")
     conflicts(
--- a/repos/spack_repo/builtin/packages/palace/package.py
+++ b/repos/spack_repo/builtin/packages/palace/package.py
@@ -49,11 +49,9 @@ class Palace(CMakePackage, CudaPackage, ROCmPackage):
         description="Build with GSLIB library for high-order field interpolation",
     )
 
-    # -- Patches for Spack Build System --
     # Fix API mismatch between libxsmm@main and internal libceed build
     patch("palace-0.12.0.patch", when="@0.12")
 
-    # -- Core Dependencies --
     # NOTE: We can't depend on git tagged versions here
     #       https://github.com/spack/spack/issues/50171
     #       Instead, version in environment / spec
@@ -68,10 +66,8 @@ class Palace(CMakePackage, CudaPackage, ROCmPackage):
     depends_on("fmt~shared", when="~shared")
     depends_on("eigen")
 
-    # -- Sparse Direct Solvers --
     conflicts("~superlu-dist~strumpack~mumps", msg="Need at least one sparse direct solver")
 
-    # -- MUMPS --
     conflicts("^mumps+int64", msg="Palace requires MUMPS without 64 bit integers")
     with when("+mumps"):
         depends_on("mumps+metis+parmetis")
@@ -80,7 +76,6 @@ class Palace(CMakePackage, CudaPackage, ROCmPackage):
         depends_on("mumps+openmp", when="+openmp")
         depends_on("mumps~openmp", when="~openmp")
 
-    # -- SuperLU-Dist --
     with when("+superlu-dist"):
         depends_on("superlu-dist+parmetis~cuda~rocm")
         depends_on("superlu-dist+shared", when="+shared")
@@ -90,7 +85,6 @@ class Palace(CMakePackage, CudaPackage, ROCmPackage):
         depends_on("superlu-dist+openmp", when="+openmp")
         depends_on("superlu-dist~openmp", when="~openmp")
 
-    # -- Strumpack --
     with when("+strumpack"):
         depends_on("strumpack+butterflypack+zfp+parmetis~cuda~rocm")
         depends_on("strumpack+shared", when="+shared")
@@ -98,10 +92,8 @@ class Palace(CMakePackage, CudaPackage, ROCmPackage):
         depends_on("strumpack+openmp", when="+openmp")
         depends_on("strumpack~openmp", when="~openmp")
 
-    # -- Eigenvalue Solvers --
     conflicts("~arpack~slepc", msg="At least one eigenvalue solver is required")
 
-    # -- SLEPc --
     with when("+slepc"):
         depends_on("slepc~arpack")
         depends_on("slepc~cuda", when="~cuda")
@@ -116,26 +108,22 @@ class Palace(CMakePackage, CudaPackage, ROCmPackage):
         depends_on("petsc~cuda", when="~cuda")
         depends_on("petsc~rocm", when="~rocm")
 
-    # -- Arpack --
     with when("+arpack"):
         depends_on("arpack-ng+mpi+icb")
         depends_on("arpack-ng+shared", when="+shared")
         depends_on("arpack-ng~shared", when="~shared")
 
-    # -- gslib --
     with when("+gslib @0.14:"):
         depends_on("gslib+mpi")
         depends_on("gslib+shared", when="+shared")
         depends_on("gslib~shared", when="~shared")
 
-    # -- METIS --
     depends_on("metis@5:")
     depends_on("metis+shared", when="+shared")
     depends_on("metis~shared", when="~shared")
     depends_on("metis+int64", when="+int64")
     depends_on("metis~int64", when="~int64")
 
-    # -- HYPRE --
     conflicts("^hypre+int64", msg="Palace uses HYPRE's mixedint option for 64 bit integers")
     depends_on("hypre~complex")
     depends_on("hypre+shared", when="+shared")
@@ -147,7 +135,6 @@ class Palace(CMakePackage, CudaPackage, ROCmPackage):
     depends_on("hypre~cuda", when="~cuda")
     depends_on("hypre~rocm", when="~rocm")
 
-    # -- libxsmm --
     with when("+libxsmm"):
         # NOTE: @=main != @main since libxsmm has a version main-2023-22
         depends_on("libxsmm@=main blas=0")
@@ -157,10 +144,8 @@ class Palace(CMakePackage, CudaPackage, ROCmPackage):
         # https://github.com/libxsmm/libxsmm/issues/883
         depends_on("libxsmm+shared")
 
-    # -- libCEED --
     depends_on("libceed@0.13:", when="@0.14:")
 
-    # -- Sundials --
     with when("+sundials @0.14:"):
         depends_on("sundials")
         depends_on("sundials+shared", when="+shared")
@@ -170,7 +155,6 @@ class Palace(CMakePackage, CudaPackage, ROCmPackage):
         depends_on("sundials~cuda", when="~cuda")
         depends_on("sundials~rocm", when="~rocm")
 
-    # -- GPU --
     conflicts(
         "+cuda", when="@:0.12", msg="CUDA is only supported for Palace versions 0.13 and above"
     )
@@ -187,7 +171,6 @@ class Palace(CMakePackage, CudaPackage, ROCmPackage):
         msg="palace: Please specify an AMD GPU target / targets",
     )
 
-    # -- Magma --
     with when("+cuda"):
         depends_on("magma+shared", when="+shared")
         depends_on("magma~shared", when="~shared")
@@ -198,7 +181,6 @@ class Palace(CMakePackage, CudaPackage, ROCmPackage):
         depends_on("magma~shared", when="~shared")
         depends_on("libceed+magma", when="@0.14:")
 
-    # -- CUDA --
     with when("+cuda"):
         for arch in CudaPackage.cuda_arch_values:
             cuda_variant = f"+cuda cuda_arch={arch}"
@@ -209,7 +191,6 @@ class Palace(CMakePackage, CudaPackage, ROCmPackage):
             depends_on(f"slepc{cuda_variant}", when=f"+slepc{cuda_variant}")
             depends_on(f"petsc{cuda_variant}", when=f"+slepc{cuda_variant}")
 
-    # -- ROCm --
     with when("+rocm"):
         for arch in ROCmPackage.amdgpu_targets:
             rocm_variant = f"+rocm amdgpu_target={arch}"
--- a/repos/spack_repo/builtin/packages/palace/package.py
+++ b/repos/spack_repo/builtin/packages/palace/package.py
@@ -239,13 +239,13 @@ class Palace(CMakePackage, CudaPackage, ROCmPackage):
         ]
 
         # We guarantee that there are arch specs with conflicts above
-        if "+cuda" in self.spec:
+        if self.spec.satisfies("+cuda"):
             args.append(
                 self.define(
                     "CMAKE_CUDA_ARCHITECTURES", ";".join(self.spec.variants["cuda_arch"].value)
                 )
             )
-        if "+rocm" in self.spec:
+        if self.spec.satisfies("+rocm"):
             args.append(
                 self.define(
                     "CMAKE_HIP_ARCHITECTURES", ";".join(self.spec.variants["amdgpu_target"].value)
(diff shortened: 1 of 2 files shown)


cameronrutherford:
Maybe a maintainer can clarify this - I ran `spack style --fix` locally, and that passes, but then CI fails? What gives?

```
> spack style --fix
==> Running style checks on spack
  selected: import, isort, black, flake8, mypy
==> Modified files
==> Running import checks
  import checks were clean
==> Running isort checks
  isort checks were clean
==> Running black checks
  black checks were clean
==> Running flake8 checks
  flake8 checks were clean
==> Running mypy checks
Success: no issues found in 617 source files
  mypy checks were clean
==> spack style checks were clean
> spack debug report
* **Spack:** 1.1.0.dev0 (https://github.com/spack/spack/commit/982e400c0a74a20a00f71008cc3b4a03c218cf09)
* **Builtin repo:** https://github.com/spack/spack-packages/commit/e374d581a464648286485f3ac8325f7749028250
* **Python:** 3.13.3
* **Platform:** darwin-sequoia-m1
```

repos/spack_repo/builtin/packages/palace/package.py
```diff
@@ -3,92 +3,107 @@
 # SPDX-License-Identifier: (Apache-2.0 OR MIT)
 
 from spack_repo.builtin.build_systems.cmake import CMakePackage
+from spack_repo.builtin.build_systems.cuda import CudaPackage
+from spack_repo.builtin.build_systems.rocm import ROCmPackage
```
cameronrutherford, line 7: In addition to the spack style questions, I don't quite understand why the older package API is being used in our local testing, and that we didn't catch these required changes sooner.

See [slack thread](https://spackpm.slack.com/archives/C08Q62S7XEX/p1754419018258109)

cameronrutherford:
> Maybe a maintainer can clarify this - I ran `spack style --fix` locally, and that passes, but then CI fails? What gives?
> 
> ```
> > spack style --fix
> ==> Running style checks on spack
>   selected: import, isort, black, flake8, mypy
> ==> Modified files
> ==> Running import checks
>   import checks were clean
> ==> Running isort checks
>   isort checks were clean
> ==> Running black checks
>   black checks were clean
> ==> Running flake8 checks
>   flake8 checks were clean
> ==> Running mypy checks
> Success: no issues found in 617 source files
>   mypy checks were clean
> ==> spack style checks were clean
> > spack debug report
> * **Spack:** 1.1.0.dev0 (https://github.com/spack/spack/commit/982e400c0a74a20a00f71008cc3b4a03c218cf09)
> * **Builtin repo:** https://github.com/spack/spack-packages/commit/e374d581a464648286485f3ac8325f7749028250
> * **Python:** 3.13.3
> * **Platform:** darwin-sequoia-m1
> ```

From Slack, `spack style --fix` now only applies to spack core, and doesn't lint packages.

hughcars:
Hi @alalazo,

@sebastiangrimberg is no longer a maintainer of the Palace project, part of this pr is actually updating the maintainer list too. Not sure if that alters how the review process should work here.

bernhardkaindl (commented):
Asking Copilot to review as well just in case this surfaces something that might have been overlooked (sometimes, even if there can be many inaccuracies in the review, the comments can lead to finding an issue).

repos/spack_repo/builtin/packages/palace/package.py
```diff
@@ -97,68 +112,179 @@ class Palace(CMakePackage):
...
+        depends_on("sundials~shared", when="~shared")
+        depends_on("sundials+openmp", when="+openmp")
+        depends_on("sundials~openmp", when="~openmp")
+
+    # -- arpack --
+    with when("+arpack"):
+        depends_on("arpack-ng+mpi+icb")
+        depends_on("arpack-ng+shared", when="+shared")
+        depends_on("arpack-ng~shared", when="~shared")
+
```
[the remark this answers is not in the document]
cameronrutherford: Yes copilot seems to have caught something here! I will fix soon

repos/spack_repo/builtin/packages/palace/package.py
```diff
@@ -97,68 +112,179 @@ class Palace(CMakePackage):
...
+        msg="palace: Please specify an AMD GPU target / targets",
+    )
 
-    with when("+magma"):
+    # -- Magma --
+    with when("+cuda") or when("+rocm"):
         depends_on("magma")
         depends_on("magma+shared", when="+shared")
         depends_on("magma~shared", when="~shared")
+        depends_on("libceed+magma", when="@0.14:")
```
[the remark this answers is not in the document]
cameronrutherford: This seems like speculation? I would sure hope a logical or would work here but maybe it doesn't?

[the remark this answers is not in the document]
bernhardkaindl: @cameronrutherford, `with when()` is a context manager that would be expected to append +cuda / +rocm to the `depends_on(when="...")` argument in the context.

I'm not sure how it behaves. I think it may work as intended, though, but it might as well break.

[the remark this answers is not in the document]
cameronrutherford: Hmm I guess I am also unsure about the behaviour, so it seems safest to just go with the separate `when` blocks.

alalazo:
> ... is no longer a maintainer of the Palace project, part of this pr is actually updating the maintainer list too. Not sure if that alters how the review process should work here.

We usually ask the old maintainer for confirmation, when somebody else removes them :wink: I see a :+1: from @sebastiangrimberg so everything's fine on that front.

repos/spack_repo/builtin/packages/palace/package.py
```diff
@@ -3,92 +3,107 @@
...
+    # -- MUMPS --
+    conflicts("^mumps+int64", msg="Palace requires MUMPS without 64 bit integers")
+    with when("+mumps"):
+        depends_on("mumps+metis+parmetis")
+        depends_on("mumps+shared", when="+shared")
+        depends_on("mumps~shared", when="~shared")
+        depends_on("mumps+openmp", when="+openmp")
+        depends_on("mumps~openmp", when="~openmp")
 
+    # -- SuperLU-Dist --
```
alalazo: Can we leave out these kind of comments (also for other dependencies)?
```suggestion
```

repos/spack_repo/builtin/packages/palace/package.py
```diff
@@ -97,68 +112,177 @@ class Palace(CMakePackage):
...
+        for arch in CudaPackage.cuda_arch_values:
+            cuda_variant = f"+cuda cuda_arch={arch}"
+            depends_on(f"hypre{cuda_variant}", when=f"{cuda_variant}")
+            depends_on(f"magma{cuda_variant}", when=f"{cuda_variant}")
+            depends_on(f"superlu-dist{cuda_variant}", when=f"+superlu-dist{cuda_variant}")
+            depends_on(f"libceed{cuda_variant}", when=f"{cuda_variant} @0.14:")
+            depends_on(f"strumpack{cuda_variant}", when=f"+strumpack{cuda_variant}")
+            depends_on(f"sundials{cuda_variant}", when=f"+sundials{cuda_variant} @0.14:")
+            depends_on(f"slepc{cuda_variant}", when=f"+slepc{cuda_variant}")
+            depends_on(f"petsc{cuda_variant}", when=f"+slepc{cuda_variant}")
```
alalazo: Hmm, forwarding `cuda_arch` is really adding a lot of constraints, which are better dealt with with requirements. 

Is this strictly necessary, or it's just to limit the valid configurations to the "sanest" defaults?

repos/spack_repo/builtin/packages/palace/package.py
```diff
@@ -97,68 +112,177 @@ class Palace(CMakePackage):
...
+        for arch in ROCmPackage.amdgpu_targets:
+            rocm_variant = f"+rocm amdgpu_target={arch}"
+            depends_on(f"hypre{rocm_variant}", when=f"{rocm_variant}")
+            depends_on(f"magma{rocm_variant}", when=f"{rocm_variant}")
+            depends_on(f"superlu-dist{rocm_variant}", when=f"+superlu-dist{rocm_variant}")
+            depends_on(f"libceed{rocm_variant}", when=f"{rocm_variant} @0.14:")
+            depends_on(f"strumpack{rocm_variant}", when=f"+strumpack{rocm_variant}")
+            depends_on(f"sundials{rocm_variant}", when=f"+sundials{rocm_variant} @0.14:")
+            depends_on(f"slepc{rocm_variant}", when=f"+slepc{rocm_variant}")
+            depends_on(f"petsc{rocm_variant}", when=f"+slepc{rocm_variant}")
```
alalazo: Same consideration here

repos/spack_repo/builtin/packages/palace/package.py
```diff
@@ -97,68 +112,177 @@ class Palace(CMakePackage):
...
+            self.define_from_variant("PALACE_WITH_OPENMP", "openmp"),
+            self.define_from_variant("PALACE_WITH_SLEPC", "slepc"),
+            self.define_from_variant("PALACE_WITH_STRUMPACK", "strumpack"),
+            self.define_from_variant("PALACE_WITH_SUNDIALS", "sundials"),
+            self.define_from_variant("PALACE_WITH_SUPERLU", "superlu-dist"),
             self.define("PALACE_BUILD_EXTERNAL_DEPS", False),
         ]
 
+        # We guarantee that there are arch specs with conflicts above
+        if "+cuda" in self.spec:
```
alalazo: (nitpick) This is considered more idiomatic:
```suggestion
        if self.spec.satisfies("+cuda"):
```

repos/spack_repo/builtin/packages/palace/package.py
```diff
@@ -97,68 +112,177 @@ class Palace(CMakePackage):
...
+            args.append(
+                self.define(
+                    "CMAKE_CUDA_ARCHITECTURES", ";".join(self.spec.variants["cuda_arch"].value)
+                )
+            )
+        if "+rocm" in self.spec:
```
alalazo:
```suggestion
        if self.spec.satisfies("+rocm"):
```

repos/spack_repo/builtin/packages/palace/package.py
cameronrutherford: > Hmm, forwarding cuda_arch is really adding a lot of constraints, which are better dealt with with requirements.

So you mean moving to using [`requires`](https://spack.readthedocs.io/en/latest/packaging_guide_creation.html#requires) instead of `depends_on` for this would be a best practice (it's better for the concretizer?)? (if it is necessary?)

> Is this strictly necessary, or it's just to limit the valid configurations to the "sanest" defaults?

This is a bug [elsewhere in Spack](https://github.com/spack/spack-packages/blob/develop/repos/spack_repo/builtin/packages/exago/package.py#L180) that I have commited, and so perhaps that should be fixed there too 😄 

I suppose you are right though that this is attempting to enforce sane defaults, and had not considered that this would be a thing that should be solved at the `spack.yaml` sort of level?

alalazo: > So you mean moving to using requires instead of depends_on for this would be a best practice (it's better for the concretizer?)? (if it is necessary?)

Sorry for the confusion, I mean requirements outside of a package recipe, so in environments etc.

alalazo: > I suppose you are right though that this is attempting to enforce sane defaults, and had not considered that this would be a thing that should be solved at the spack.yaml sort of level?

Yeah, that's exactly what I meant.

hughcars: The aim with these `depends_on` is to propagate the constraints such that the installation consist of `spack add palace@develop+cuda cuda_arch=80`, then `spack install`, so the constraints are propagated to all the dependencies. We have run into issues previously with under constraining dependencies causing weird runtime errors, so we opt for keeping all of them aligned with each other. Most of our users don't have the inclination to precisely setup their environment, and are looking to spack for a 1 liner to start using `palace` (with or without gpus).

Were these removed, the user would need to check that all the dependencies are constrained in their environment consistently yes?

repos/spack_repo/builtin/packages/palace/package.py
```diff
@@ -3,92 +3,107 @@
...
+    with when("+mumps"):
+        depends_on("mumps+metis+parmetis")
+        depends_on("mumps+shared", when="+shared")
+        depends_on("mumps~shared", when="~shared")
+        depends_on("mumps+openmp", when="+openmp")
+        depends_on("mumps~openmp", when="~openmp")
 
+    # -- SuperLU-Dist --
     with when("+superlu-dist"):
+        depends_on("superlu-dist+parmetis")
```
hughcars: Needs `~cuda~rocm` see https://github.com/awslabs/palace/pull/436

repos/spack_repo/builtin/packages/palace/package.py
```diff
@@ -3,92 +3,107 @@
...
         depends_on("superlu-dist+shared", when="+shared")
         depends_on("superlu-dist~shared", when="~shared")
         depends_on("superlu-dist+int64", when="+int64")
         depends_on("superlu-dist~int64", when="~int64")
         depends_on("superlu-dist+openmp", when="+openmp")
         depends_on("superlu-dist~openmp", when="~openmp")
 
+    # -- Strumpack --
     with when("+strumpack"):
         depends_on("strumpack+butterflypack+zfp+parmetis")
```
hughcars: Needs `~cuda~rocm` see https://github.com/awslabs/palace/pull/436

repos/spack_repo/builtin/packages/palace/package.py
```diff
@@ -97,68 +112,177 @@ class Palace(CMakePackage):
...
-    conflicts("mfem")
-    conflicts("libceed")
-    conflicts("gslib")
+    # -- CUDA --
+    with when("+cuda"):
+        for arch in CudaPackage.cuda_arch_values:
+            cuda_variant = f"+cuda cuda_arch={arch}"
+            depends_on(f"hypre{cuda_variant}", when=f"{cuda_variant}")
+            depends_on(f"magma{cuda_variant}", when=f"{cuda_variant}")
+            depends_on(f"superlu-dist{cuda_variant}", when=f"+superlu-dist{cuda_variant}")
```
hughcars: Delete `superlu-dist` and `strumpack` from this list, their gpu builds don't work with palace presently.

repos/spack_repo/builtin/packages/palace/package.py
```diff
@@ -97,68 +112,177 @@ class Palace(CMakePackage):
...
-    conflicts("^hypre+int64", msg="Palace uses HYPRE's mixedint option for 64 bit integers")
-    conflicts("^mumps+int64", msg="Palace requires MUMPS without 64 bit integers")
-    conflicts("^slepc+arpack", msg="Palace requires SLEPc without ARPACK")
+    # -- ROCm --
+    with when("+rocm"):
+        for arch in ROCmPackage.amdgpu_targets:
+            rocm_variant = f"+rocm amdgpu_target={arch}"
+            depends_on(f"hypre{rocm_variant}", when=f"{rocm_variant}")
+            depends_on(f"magma{rocm_variant}", when=f"{rocm_variant}")
+            depends_on(f"superlu-dist{rocm_variant}", when=f"+superlu-dist{rocm_variant}")
```
hughcars: As with `cuda` above, remove `superlu-dist` and `strumpack` from this list.

hughcars (changes requested):
Small fix in light of https://github.com/awslabs/palace/pull/436

repos/spack_repo/builtin/packages/palace/package.py
cameronrutherford: This seems to be fixed by explicitly setting the package API version.

repos/spack_repo/builtin/packages/palace/package.py
cameronrutherford: I did this in a single commit in case @hughcars wants to revert.

repos/spack_repo/builtin/packages/palace/package.py
cameronrutherford: @alalazo I think that since Palace doesn't have any downstream dependencies (at least not for the conceivable future), I think we want these constraints for our package for ease of use for end users.

However, I could be in favor of removing this if there's a way to somehow forward this requirement to all dependencies at the spec level, or even say something like:

```bash
spack install palace@develop ++cuda %%cuda_arch=80`
```

Where somehow `++cuda` forward `+cuda` to all dependencies that we don't have a `depends_on("foo~cuda")`, and then there is some syntax (I chose `%%` arbitrarily) to forward this.

I know that we can do `packages:all:prefer:"+cuda cuda_arch=80"`, but again our preference is to make this as easy as possible for end users.

Again I see both sides here, but I am leaning towards leaving this as-is since it seems logical for "sane defaults".

cameronrutherford:
I rebased on develop, as well as pulling in the internal Palace GPU fixes in this PR. Ready for final review from @hughcars and @alalazo to hopefully resolve the one final thread.

repos/spack_repo/builtin/packages/palace/package.py
```diff
@@ -74,91 +86,186 @@ class Palace(CMakePackage):
...
+                self.define("HYPRE_REQUIRED_PACKAGES", "LAPACK;BLAS"),
+                self.define("BLAS_LIBRARIES", self.spec["blas"].libs.joined(";")),
+                self.define("LAPACK_LIBRARIES", self.spec["lapack"].libs.joined(";")),
+            ]
+        )
 
         # MPI compiler wrappers are not required, but MFEM test builds need to know to link
-        # against MPI libraries
+        # against MPI libraries.
         if "+superlu-dist" in self.spec:
```
hughcars: I think these `in self.spec` can be changed to the previously suggested `self.spec.satisfies("XXX"):` approach.

repos/spack_repo/builtin/packages/palace/package.py
alalazo: > However, I could be in favor of removing this if there's a way to somehow forward this requirement to all dependencies at the spec level [ ... ]

@cameronrutherford You can either force it in requirements:
```yaml
packages:
  all:
    require:
    - "+cuda"
    - "cuda_arch=80"
```
in which case it will be applied to all the packages that have the variant, or propagate it from the input spec. The syntax is:
```console
$ spack install palace@develop ++cuda cuda_arch==80`
```

hughcars (approved):
Tested develop, develop + cuda, 0.13, all work. 0.13 with cuda did not work, so added the block until a possible patch can be introduced. Will probably just be superseded by the upcoming 0.14.0.

repos/spack_repo/builtin/packages/palace/package.py
cameronrutherford: This is an outdated comment - I have updated this in the latest commit(s).

repos/spack_repo/builtin/packages/palace/package.py
cameronrutherford: @hughcars this is the last open thread - what do you think? I am on board with the recommended requirements here, but perhaps this would necessitate someone like @sbozzolo adding this information to our docs for users.

hughcars: I'm happy with the current variant approach here. If we want to change to the `++cuda` model, that will propagate to every dependency as I understand it (unless we explicitly say `~cuda`). We've had issues with dependencies breaking palace with that in the past, but maybe that's just an overabundance of caution. The current approach most closely models the cmake build process, rather than leaving the cuda specification implicit. I'm very open to revisiting this in a future PR, not least the upcoming 0.14.0, but for now I think it works as is.

cameronrutherford: Agreed. I think as well once we have GPU testing for Palace, we can be have much more confidence about these approaches. @alalazo if this isn't blocking the PR, I think we can merge!

alalazo: Since:
```console
$ spack dependents palace
No dependents
```
I'm fine with dealing with it in a following PR.

alalazo (approved):

PR #962, opened 2025-08-05
iwyu,g2,r-curl: requires(pkg) -> requires(^pkg)

wdconinc:
This PR fixes issues in three packages, `iwyu`, `g2`, `r-curl`, where the `requires` clause was used incorrectly, without caret to refer to dependencies.


AlexanderRichert-NOAA (approved):
D'oh-- Thanks for this. Approving for g2.

alalazo (approved):

alalazo (approved):

PR #973, opened 2025-08-06
libpressio: add blosc2 depends_on

mathomp4:
In [trying to build libpressio today](https://github.com/robertu94/libpressio/issues/42), I found that because I'd added `+blosc2` to my variants, the CMake failed because it didn't find blosc2.

So, we add a `depends_on`


robertu94 (approved):
Good catch.

alalazo (approved):

PR #263, opened 2025-06-16
relion: add v5 runtime deps

snehring:
model-angelo: add new package
py-fair-esm: add v1.0.3
py-pyhmmer: add v0.7.1
py-relion-blush: add new package
py-relion-classranker: add new package
topaz-3dem: add new package
py-deepspeed: add new versions
relion: adding whens to new deps, rework for repo changes

migrated from spack repo

PR #977, opened 2025-08-06
relion: Fix the build to with libx11 and deps fltk and ctffind

bernhardkaindl:
Fixes for building `relion` and some of its build-time dependencies:

- relion: Use spack's libx11 headers, the X11 headers of the host (if they exist).
- ctffind: Fix build failure: Add missing depends on cxx
- fltk@1.3.7: Fix build when pkgconfig is not installed on the host

Commit messages of the new commits:

- relion: Fix build failure: Use spack's libx11 includes
  To find the spack libx11 includes, we need to pass -I<path> for libx11
  and with fltk@1.3, we also need to pass -I<path> for xproto.
  `-DCMAKE_C_FLAGS=-g and -DCMAKE_CXX_FLAGS=-g`, which are not needed, are
  replaced by passing the include directories when relion is using +gui.
- ctffind: Fix build failure: Add missing depends on cxx
- fltk: Fix build when pkgconfig is not installed on the host
- fltk: Add new version fltk-1.4.4

I developed these while trying to build #263 by @snehring, who also added these to his PR now.

### DIFF
--- a/repos/spack_repo/builtin/packages/ctffind/package.py
+++ b/repos/spack_repo/builtin/packages/ctffind/package.py
@@ -26,7 +26,8 @@ class Ctffind(AutotoolsPackage):
         extension="tar.gz",
     )
 
-    depends_on("cxx", type="build")  # generated
+    depends_on("c", type="build")
+    depends_on("cxx", type="build")
 
     def url_for_version(self, version):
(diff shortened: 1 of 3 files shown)


bernhardkaindl:
Merged with #263

PR #992, opened 2025-08-07
doxygen: added support for doxygen@1.14.0

mcmehrtens:
* added a `conflicts` field for a compilation error that occurs when compiling Doxygen 1.13.2 or prior when using LLVM 19 or higher
  * [Description of error](https://github.com/doxygen/doxygen/issues/10928)
* added support for Doxygen 1.14.0
  * added the `use_sys_fmt` option to prevent linker errors between Doxygen's built-in `fmt` dependency and the version built for `spdlog`
* checked license
* updated the package description based on [Doxygen's GitHub](https://github.com/doxygen/doxygen#doxygen)
* ~added variants to optionally compile with system versions of SQLite3 and spdlog~
  * ~Doxygen bundles its own versions of SQLite3 and spdlog; previously this package forced the user to build SQLite3 and spdlog which was causing errors for (compiling on macOS with clang)~
* ~restricts Python versions to Python 2.7 or higher per the [Doxygen build documentation](https://www.doxygen.nl/manual/install.html)~
* ~restricts SQLite versions to SQLite 3 or higher~


repos/spack_repo/builtin/packages/doxygen/package.py
```diff
@@ -92,7 +97,7 @@ def determine_variants(cls, exes, version_str):
     depends_on("cmake@3.12:", type="build", when="@1.9.8:")
     depends_on("cmake@3.14:", type="build", when="@1.10:")
 
-    depends_on("python", type="build")  # 2 or 3 OK; used in CMake build
+    depends_on("python@2.7:", type="build")  # 2 or 3 OK; used in CMake build
```
mcmehrtens: [Doxygen build documentation](https://www.doxygen.nl/manual/install.html) specifies Python 2.7 or higher

repos/spack_repo/builtin/packages/doxygen/package.py
```diff
@@ -102,8 +107,8 @@ def determine_variants(cls, exes, version_str):
     depends_on("bison@2.7:", type="build", when="@1.8.10:")
 
     # originally bundled dependencies
-    depends_on("spdlog", when="@1.9.8:")
-    depends_on("sqlite", when="@1.10:")
+    depends_on("spdlog", when="+system_spdlog")
+    depends_on("sqlite@3", when="+system_sqlite3")
```
mcmehrtens: Doxygen requires requires SQLite3 or higher

alecbcs:
@mcmehrtens while we do occasionally allow packages to bundle dependencies in their own builds, we try really hard in Spack to use Spack built dependencies so that they can be reused for other builds across your system and not lead to hard to debug errors in loader race conditions. My preference would be to leave `doxygen` using the Spack built dependencies and debug why those builds aren't compiling for MacOS.

mcmehrtens:
@alecbcs Roger that, I'll do some debugging! It was probably a me issue. Setting this to draft until I do some investigation.

mcmehrtens:
@alecbcs @wdconinc I finally had time to circle back to this. I updated the PR description, fixed the Doxygen 1.13.2 compilation error that drove me to add 1.14.0 support in the first place, and added proper 1.14.0 support without using Doxygen-bundled dependencies.

mcmehrtens:
Shoot, I forgot to tag the current maintainer: @sethrj

Sorry, about that...

repos/spack_repo/builtin/packages/doxygen/llvm19-fix-1.13.2.patch
```diff
@@ -0,0 +1,56 @@
+diff --git a/CMakeLists.txt b/CMakeLists.txt
+index a8f977c75..0e5b81fbd 100644
```
sethrj: Rather than including patch files with spack (which is effectively "forking" the repository if the patch is customized) it's much preferred to reference an existing git commit or pull request online. Is this issue fixed upstream and can you link to a patch?

repos/spack_repo/builtin/packages/doxygen/package.py
```diff
@@ -10,17 +10,19 @@
 
 
 class Doxygen(CMakePackage):
-    """Doxygen is the de facto standard tool for generating documentation
-    from annotated C++ sources, but it also supports other popular programming
-    languages such as C, Objective-C, C#, PHP, Java, Python, IDL (Corba,
-    Microsoft, and UNO/OpenOffice flavors), Fortran, VHDL, Tcl, and to some
-    extent D.."""
+    """Doxygen is the de facto standard tool for generating
```
sethrj: The rewrapping is a little strange and should ideally be reverted (I think spack's formatter is set up to something like 100 chars wide?)

repos/spack_repo/builtin/packages/doxygen/package.py
```diff
@@ -10,17 +10,19 @@
...
+    documentation from annotated C++ sources, but it also supports other
+    popular programming languages such as C, Objective-C, C#, PHP, Java,
+    Python, IDL (Corba, Microsoft, and UNO/OpenOffice flavors), Fortran,
+    and to some extent D. Doxygen also supports the hardware description
+    language VHDL."""
 
     homepage = "https://www.doxygen.org"
-    url = "https://github.com/doxygen/doxygen/archive/refs/tags/Release_1_13_2.tar.gz"
+    url = "https://github.com/doxygen/doxygen/archive/refs/tags/Release_1_14_0.tar.gz"
```
sethrj: The URLs don't need to be updated at every version change; they're strictly for interpreting the format of the download URL.

sethrj (commented):
Thanks @mcmehrtens ! Can you please address these minor comments?

repos/spack_repo/builtin/packages/doxygen/package.py
mcmehrtens: Should be addressed in 78158b1f861a82e61a5ad6e0ee7e0a45ee90f3c9.

repos/spack_repo/builtin/packages/doxygen/package.py
mcmehrtens: Reverted this change. See 78158b1f861a82e61a5ad6e0ee7e0a45ee90f3c9.

repos/spack_repo/builtin/packages/doxygen/llvm19-fix-1.13.2.patch
mcmehrtens: @sethrj This proved to be a bit tricky. I tried using a variety of different commits from the PR that fixed this and the merge commit with no success. Main issue was that there were significant (enough) changes between release 1.13.2 and the commits that fixed the bug that prevented the auto-generated GitHub patches from applying cleanly.

I'm happy to keep trying to find a solution using auto-generated patches from GitHub, but I'll also offer 78158b1f861a82e61a5ad6e0ee7e0a45ee90f3c9 as a potential solution. Can we just restrict LLVM to < 19 for versions prior to Doxygen 1.14?

sethrj: I think that's the best solution 😄  If someone really needs to install an older doxygen with a very new compiler, we can let them submit a patch 😉

sethrj (approved):
Checksum valid, extra annotations look good! Thanks for the update @mcmehrtens .

repos/spack_repo/builtin/packages/doxygen/llvm19-fix-1.13.2.patch
mcmehrtens: That works for me :) Thank you for reviewing—I appreciate your diligence. Can you resolve this thread so the PR can merge?

PR #1000, opened 2025-08-07
gnupg: add v2.5.11, fix conflicts for libassuan on macos

alecbcs:
Add gnupg v2.5.11 and update conflicts for libassuan on MacOS.


alalazo:
We should test this one in the `spack/spack` repo to check whether bootstrapping from sources continues to work. Otherwise LGTM.

alalazo (approved):

alalazo:
Bootstrapping tested

PR #1003, opened 2025-08-07
Python iconv flexibility

marcmengel:
> This is a redo of  https://github.com/spack/spack/pull/39992

Currently python **always** depends on gettext, regardless what you have set for iconv preference.

This PR allows building python depending on  glibc  or libiconv instead of gettext for iconv functionality, while still allowing avoiding the libxml2 +python / python / gettext +libxml2 dependency cycle. 

With this change you can:
```
spack spec python                          # gives python +libxml2 -> gettext +libxml2 -> libxml2
spack spec python ~libxml2                 # gives python ~lixml2  -> gettext ~libxml2
spack spec python ~libxml2 ^iconv=glibc    # gives python ~libxml2 -> glibc
spack spec python ~libxml2 ^iconv=libiconv # gives python ~libxml2 -> libiconv

```
and specifying your preferred iconv in packages.yaml works for python ~libxml2

### DIFF
--- a/repos/spack_repo/builtin/packages/python/package.py
+++ b/repos/spack_repo/builtin/packages/python/package.py
@@ -316,7 +316,7 @@ class Python(Package):
         depends_on("pkgconfig@0.9.0:", type="build")
         depends_on("gettext +libxml2", when="+libxml2")
         depends_on("iconv", when="~libxml2")
-        depends_on("gettext ~libxml2", when="~libxml2 ^gettext")
+        depends_on("gettext ~libxml2", when="~libxml2 ^[virtuals=iconv]gettext")
 
         # Optional dependencies
         # See detect_modules() in setup.py for details
--- a/repos/spack_repo/builtin/packages/python/package.py
+++ b/repos/spack_repo/builtin/packages/python/package.py
@@ -315,7 +315,8 @@ class Python(Package):
         depends_on("gmake", type="build")
         depends_on("pkgconfig@0.9.0:", type="build")
         depends_on("gettext +libxml2", when="+libxml2")
-        depends_on("gettext ~libxml2", when="~libxml2")
+        depends_on("iconv", when="~libxml2")
+        depends_on("gettext ~libxml2", when="~libxml2 ^gettext")
 
         # Optional dependencies
         # See detect_modules() in setup.py for details


bernhardkaindl (approved):
@becker33, this looks good, could you review it? Thanks!

repos/spack_repo/builtin/packages/python/package.py
```diff
@@ -315,7 +315,8 @@ class Python(Package):
         depends_on("gmake", type="build")
         depends_on("pkgconfig@0.9.0:", type="build")
         depends_on("gettext +libxml2", when="+libxml2")
-        depends_on("gettext ~libxml2", when="~libxml2")
+        depends_on("iconv", when="~libxml2")
```
becker33, line 318: Does it only depend on iconv when ~libxml, or should it still be providing `iconv` on the gettext dependency edge?

repos/spack_repo/builtin/packages/python/package.py
```diff
@@ -315,7 +315,8 @@ class Python(Package):
...
+        depends_on("gettext ~libxml2", when="~libxml2 ^[virtuals=iconv]gettext")
```
becker33, line 319:
```suggestion
        depends_on("gettext ~libxml2", when="~libxml2 %iconv=gettext")
```
some syntactic sugar

becker33 (changes requested):

repos/spack_repo/builtin/packages/python/package.py
marcmengel: I would think it *should* always depend on iconv, and when="+libxml2"  iconv should be implemented by gettext+libxml2. 
But whoever previously made the recipe made it directly depend on gettext for that case, so I wasn't sure if there was a reason for that.

becker33: I think the best way to accomplish that directly is

```
depends_on("iconv")
requires("%iconv=gettext+libxml2", when="+libxml2")
```
But the better thing to do overall is probably to remove the `libxml2` variant. I added it years ago to work around a cyclic dependency (back with the old non-backtracking concretizer that couldn't work around cycles itself). Then we can safely change `depends_on("gettext")` to `depends_on("iconv")`.

However, it looks from what I'm seeing on cpython's github that python has additional capabilities for internationalization when it depends on gettext vs any other iconv implementation. It makes me think that the best option is:

1. Remove the `libxml2` variant and rely on the concretizer to avoid the cycle.
2. Add a boolean variant for `gettext`, default `True`.
3. `depends_on("iconv")`
4. `depends_on("gettext", when="+gettext")`
5. `requires("%iconv=gettext", when="+gettext")` so that Spack recognizes that gettext has to be the iconv provider when it is present.
6. `conflicts("%iconv=gettext", when="~gettext") so that Spack doesn't use gettext when it's not being used.

I'm only moderately confident in all of this -- I haven't spent a ton of time looking at the internals of the python package.

@adamjstewart or @scheibelp may want to chime in on this as python package maintainers as well.

adamjstewart: Completely fine with removing the libxml2 variant, that was only added as a workaround for the same issue we're trying to fix here.

Other ideas, well we'll just have to try and see. I remain curious how other package managers resolve this dependency cycle. It really shouldn't be that hard to break this cycle.

scheibelp: I don`t see `gettext having anything like `provides(iconv)` (and I don't see changes to gettext here)

Removing `+libxml2` sounds good. If `gettext` is only present for `iconv` (what it looks like to me based on 4/5), then it seems like you could remove 4/6. It might be worth naming the variant based on

> However, it looks from what I'm seeing on cpython's github that python has additional capabilities for internationalization when it depends on gettext 

vs. `gettext` though.

`gettext` package: https://github.com/spack/spack-packages/blob/develop/repos/spack_repo/builtin/packages/gettext/package.py

becker33: I'm guessing that it provides iconv as a pass-through and we don't model it because it also depends on iconv?

becker33 (approved):
Ok having seen how much more complicated this might get, I think this PR is a good improvement as-is and more work should go into this later to make further improvements

Issue #1024, opened 2025-08-08
Installation issue: zsh


wspear:
### Steps to reproduce the issue

```console
$ spack spec -I zsh
 -   zsh@5.9~etcdir~lmod+skip-tcsetpgrp-test build_system=autotools arch=linux-ubuntu24.04-westmere %c=gcc@14.2.0
[+]      ^compiler-wrapper@1.0 build_system=generic arch=linux-ubuntu24.04-westmere 
[e]      ^gcc@14.2.0~binutils+bootstrap~graphite~mold~nvptx~piclibs~profiled~strip build_system=autotools build_type=RelWithDebInfo languages:='c,c++,fortran' arch=linux-ubuntu24.04-westmere 
[+]      ^gcc-runtime@14.2.0 build_system=generic arch=linux-ubuntu24.04-westmere 
[e]      ^glibc@2.39 build_system=autotools arch=linux-ubuntu24.04-westmere 
[+]      ^gmake@4.4.1~guile build_system=generic arch=linux-ubuntu24.04-westmere %c=gcc@14.2.0
[+]          ^compiler-wrapper@1.0 build_system=generic arch=linux-ubuntu24.04-westmere 
[e]          ^gcc@14.2.0~binutils+bootstrap~graphite~mold~nvptx~piclibs~profiled~strip build_system=autotools build_type=RelWithDebInfo languages:='c,c++,fortran' arch=linux-ubuntu24.04-westmere 
[+]          ^gcc-runtime@14.2.0 build_system=generic arch=linux-ubuntu24.04-westmere 
[+]      ^ncurses@6.5~symlinks+termlib abi=none build_system=autotools patches:=7a351bc arch=linux-ubuntu24.04-westmere %c,cxx=gcc@14.2.0
[+]          ^pkgconf@2.3.0 build_system=autotools arch=linux-ubuntu24.04-westmere %c=gcc@14.2.0
[+]      ^pcre@8.45~jit+multibyte+pic+shared+static+utf build_system=autotools arch=linux-ubuntu24.04-westmere %c,cxx=gcc@14.2.0

```

### Error message

<details><summary>Error message</summary>

<pre>
==> zsh: Executing phase: 'build'
==> Error: ProcessError: Command exited with status 2:
    '/home/wspear/bin/SPACK/spack/opt/spack/linux-westmere/gmake-4.4.1-ea3z3vn3mqbyqw6a7qxc42inqnxcl2h3/bin/make' '-j16' 'V=1'

[...]

PR #1025, opened 2025-08-08
zsh doesn't build with 6 series ncurses

wspear:
See: https://github.com/spack/spack-packages/issues/1024

### DIFF
--- a/repos/spack_repo/builtin/packages/zsh/package.py
+++ b/repos/spack_repo/builtin/packages/zsh/package.py
@@ -41,9 +41,11 @@ class Zsh(AutotoolsPackage):
     depends_on("c", type="build")  # generated
 
     depends_on("pcre")
-    depends_on("ncurses@:5.9")
+    depends_on("ncurses")
 
     conflicts("+lmod", when="~etcdir", msg="local etc required to setup env for lmod")
+
+    patch("pointer-types.patch", when="@5.6.2:")
 
     def configure_args(self):
         args = []
--- a/repos/spack_repo/builtin/packages/zsh/package.py
+++ b/repos/spack_repo/builtin/packages/zsh/package.py
@@ -41,7 +41,7 @@ class Zsh(AutotoolsPackage):
     depends_on("c", type="build")  # generated
 
     depends_on("pcre")
-    depends_on("ncurses")
+    depends_on("ncurses@:5.9")
 
     conflicts("+lmod", when="~etcdir", msg="local etc required to setup env for lmod")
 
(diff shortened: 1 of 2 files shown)


alecbcs:
We should probably investigate this further as neither [nix](https://github.com/NixOS/nixpkgs/blob/nixos-25.05/pkgs/by-name/zs/zsh/package.nix) or [brew](https://github.com/Homebrew/homebrew-core/blob/7606217f5605c3db14cf0934e0ea5d8b576a6eed/Formula/z/zsh.rb) limit the version of ncurses used in their builds.

michaelkuhn:
I'm guessing https://github.com/zsh-users/zsh/commit/4c89849c98172c951a9def3690e8647dae76308f.patch is the patch we need (used in the Nix and Fedora recipes).

alecbcs:
@wspear can you try relaxing dependency limit on `ncurses@5` and add the linked patch and see if that also fixes the build issue for you?

alecbcs:
@wspear do you have time to try the linked patch and update the PR?

wspear:
The patch operation failed and I'm not sure why. I downloaded it to the zsh directory and added patch("4c89849c98172c951a9def3690e8647dae76308f.patch") on line 47 of package.py and got:

```
==> Installing zsh-5.9-gmazlahrrlb7gxvdyqbnr2t7lmmtq5ng [8/8]
==> Using cached archive: /home/wspear/bin/SPACK/spack/var/spack/cache/_source-cache/archive/9b/9b8d1ecedd5b5e81fbf1918e876752a7dd948e05c1a0dba10ab863842d45acd5.tar.xz
1 out of 1 hunk FAILED -- saving rejects to file ChangeLog.rej
==> Patch /home/wspear/.spack/package_repos/fncqgg4/repos/spack_repo/builtin/packages/zsh/4c89849c98172c951a9def3690e8647dae76308f.patch failed.
    '/usr/bin/patch' '-s' '-p' '1' '-i' '/home/wspear/.spack/package_repos/fncqgg4/repos/spack_repo/builtin/packages/zsh/4c89849c98172c951a9def3690e8647dae76308f.patch' '-d' '.'
```

michaelkuhn:
This happens due to the `ChangeLog` changes, just remove that hunk.

wspear:
That worked and zsh built. But I also had to remove the commit id (changed the filename and removed the line internally, not sure which worked) to prevent spack from rejecting it due to the hash difference. Is there a best practice for committing someone else's edited patch?

wspear:
Thanks for tracking that patch down. This pr works for me.

eugeneswalker:
Any objections to merging this? Can we proceed @michaelkuhn @alecbcs ?

alecbcs (approved):
Looks good to me. Thanks @wspear for digging into this patch.

PR #1026, opened 2025-08-09
affinity: Added better ROCm support, updated cmake variables

nhanford:
Hello @bcumming,

I updated the cmake arguments to reflect your variable changes and tested this successfully with ROCm on the Corona system at LLNL. Please let me know what you think.

Thanks,
Nate

### DIFF
--- a/repos/spack_repo/builtin/packages/affinity/package.py
+++ b/repos/spack_repo/builtin/packages/affinity/package.py
@@ -4,11 +4,12 @@
 
 from spack_repo.builtin.build_systems.cmake import CMakePackage
 from spack_repo.builtin.build_systems.cuda import CudaPackage
+from spack_repo.builtin.build_systems.rocm import ROCmPackage
 
 from spack.package import *
 
 
-class Affinity(CMakePackage, CudaPackage):
+class Affinity(CMakePackage, CudaPackage, ROCmPackage):
     """Simple applications for determining Linux thread and gpu affinity."""
 
     homepage = "https://github.com/bcumming/affinity"
@@ -21,8 +22,15 @@ class Affinity(CMakePackage, CudaPackage):
 
     variant("mpi", default=False, description="Build MPI support")
     variant("rocm", default=False, description="Build ROCm Support")
+    variant("cuda", default=False, description="Build CUDA Support")
 
+    depends_on("cmake@3.21:", type="build")
     depends_on("mpi", when="+mpi")
     depends_on("hip", when="+rocm")
     depends_on("mpi", when="+mpi")
-    depends_on("mpi", when="+cuda")
+    depends_on("cuda", when="+cuda")
+
+    def cmake_args(self):
+        args = [self.define_from_variant("AFFINITY_MPI", "mpi")]
+        args += [self.define("AFFINITY_GPU_BACKEND", v) for v in ["cuda", "rocm"] if self.spec.satisfies(f"+{v}")]
+        return args
--- a/repos/spack_repo/builtin/packages/affinity/package.py
+++ b/repos/spack_repo/builtin/packages/affinity/package.py
@@ -21,7 +21,6 @@ class Affinity(CMakePackage, CudaPackage, ROCmPackage):
     license("BSD-3-Clause", checked_by="nhanford")
 
     variant("mpi", default=False, description="Build MPI support")
-    variant("rocm", default=False, description="Build ROCm Support")
 
     depends_on("cmake@3.21:", type="build")
     depends_on("mpi", when="+mpi")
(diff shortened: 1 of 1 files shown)


nhanford:
@spackbot fix style

bcumming (approved):

repos/spack_repo/builtin/packages/affinity/package.py
```diff
@@ -21,8 +22,19 @@ class Affinity(CMakePackage, CudaPackage):
 
     variant("mpi", default=False, description="Build MPI support")
     variant("rocm", default=False, description="Build ROCm Support")
+    variant("cuda", default=False, description="Build CUDA Support")
```
alalazo: The variants `rocm` and `cuda` are inherited from the base classes. Since the default values are the same, we can remove them from the definition here.

repos/spack_repo/builtin/packages/affinity/package.py
```diff
@@ -21,8 +22,19 @@ class Affinity(CMakePackage, CudaPackage):
...
 
+    depends_on("cmake@3.21:", type="build")
     depends_on("mpi", when="+mpi")
     depends_on("hip", when="+rocm")
     depends_on("mpi", when="+mpi")
-    depends_on("mpi", when="+cuda")
+    depends_on("cuda", when="+cuda")
```
alalazo: This is also coming from `CudaPackage`
```suggestion
```

repos/spack_repo/builtin/packages/affinity/package.py
```diff
@@ -21,8 +22,19 @@ class Affinity(CMakePackage, CudaPackage):
 
     variant("mpi", default=False, description="Build MPI support")
     variant("rocm", default=False, description="Build ROCm Support")
+    variant("cuda", default=False, description="Build CUDA Support")
 
+    depends_on("cmake@3.21:", type="build")
     depends_on("mpi", when="+mpi")
     depends_on("hip", when="+rocm")
     depends_on("mpi", when="+mpi")
```
alalazo:
```suggestion
```

repos/spack_repo/builtin/packages/affinity/package.py
alalazo:
```suggestion
```

repos/spack_repo/builtin/packages/affinity/package.py
```diff
@@ -21,8 +22,19 @@ class Affinity(CMakePackage, CudaPackage):
 
     variant("mpi", default=False, description="Build MPI support")
     variant("rocm", default=False, description="Build ROCm Support")
```
alalazo:
```suggestion
```

alalazo (changes requested):
I think some cleanup is still possible, but otherwise LGTM.

nhanford:
@spackbot fix style

nhanford:
@alalazo Thanks for the helpful comments and suggestions. I finally got a chance to get this tested here and it works for us, so it should finally be ready.

repos/spack_repo/builtin/packages/affinity/package.py
```diff
@@ -20,9 +21,15 @@ class Affinity(CMakePackage, CudaPackage):
...
-    depends_on("mpi", when="+mpi")
-    depends_on("mpi", when="+cuda")
+
+    def cmake_args(self):
+        args = [self.define_from_variant("AFFINITY_MPI", "mpi")]
+        args += [
+            self.define("AFFINITY_GPU_BACKEND", v)
+            for v in ["cuda", "rocm"]
+            if self.spec.satisfies(f"+{v}")
+        ]
```
alalazo, line 37: Does this work if `+rocm+cuda`, or does it just use ROCm?

nhanford: It should work for `+rocm+cuda`, either variant, or none.

alalazo: Sorry, what I mean is the options above translates to:
```console
$ cmake -DAFFINITY_GPU_BACKEND=cuda -DAFFINITY_GPU_BACKEND=rocm [ ... ]
```
if `+cuda+rocm`, so I think only ROCm would be active, despite the `+rocm+cuda` in the spec.

nhanford: Oh yeah, good catch. The current `CMakeLists.txt` makes these exclusive, but that may not be the case forever. Perhaps I should make the two variants conflicts for now?

bcumming (approved):

alalazo (approved):

PR #1027, opened 2025-08-09
h5py: add 3.14 and fix numpy version bounds

Chrismarsh:
Fixes the numpy version bounds introduced in https://github.com/spack/spack-packages/commit/4aaf3853c3b6e4937101c7c3a684faaab949e733#r161610115

Despite the confusing notes, 3.11 doesn't require numpy2 to build

### DIFF
--- a/repos/spack_repo/builtin/packages/py_h5py/package.py
+++ b/repos/spack_repo/builtin/packages/py_h5py/package.py
@@ -19,6 +19,7 @@ class PyH5py(PythonPackage):
     license("BSD-3-Clause")
 
     version("master", branch="master")
+    version("3.14.0", sha256="2372116b2e0d5d3e5e705b7f663f7c8d96fa79a4052d250484ef91d24d6a08f4")
     version("3.13.0", sha256="1870e46518720023da85d0895a1960ff2ce398c5671eac3b1a41ec696b7105c3")
     version("3.12.1", sha256="326d70b53d31baa61f00b8aa5f95c2fcb9621a3ee8365d770c551a13dbbcbfdf")
     # Yanked
@@ -61,8 +62,10 @@ class PyH5py(PythonPackage):
     depends_on("py-cython@0.29.14:0", type=("build"), when="@3:3.7 ^python@3.8.0:3.8")
     depends_on("py-cython@0.29:0", type=("build"), when="@3.0:3.10")
     depends_on("py-cython@0.23:0", type="build", when="@:2")
+
     depends_on("py-pkgconfig", type="build")
     depends_on("py-setuptools@61:", type="build", when="@3.8.0:")
+    depends_on("py-setuptools@77:", type="build", when="@3.14:")
     depends_on("py-setuptools", type="build")
 
     # pre-3.11 is numpy@1 only
--- a/repos/spack_repo/builtin/packages/py_h5py/package.py
+++ b/repos/spack_repo/builtin/packages/py_h5py/package.py
@@ -63,8 +63,8 @@ class PyH5py(PythonPackage):
     depends_on("py-cython@0.23:0", type="build", when="@:2")
 
     depends_on("py-pkgconfig", type="build")
-    depends_on("py-setuptools@61:", type="build", when="@3.8.0:")
     depends_on("py-setuptools@77:", type="build", when="@3.14:")
+    depends_on("py-setuptools@61:", type="build", when="@3.8.0:")
     depends_on("py-setuptools", type="build")
 
     # pre-3.11 is numpy@1 only
--- a/repos/spack_repo/builtin/packages/py_h5py/package.py
+++ b/repos/spack_repo/builtin/packages/py_h5py/package.py
@@ -59,7 +59,6 @@ class PyH5py(PythonPackage):
     depends_on("py-cython@0.29.31:3", type="build", when="@3.11:")
     depends_on("py-cython@0.29.31:0", type="build", when="@3.9:3.10")
     depends_on("py-cython@0.29.15:0", type=("build"), when="@3:3.7")
-    depends_on("py-cython@0.29.14:0", type=("build"), when="@3:3.7 ^python@3.8.0:3.8")
     depends_on("py-cython@0.29:0", type=("build"), when="@3.0:3.10")
     depends_on("py-cython@0.23:0", type="build", when="@:2")
 
@@ -94,7 +93,6 @@ class PyH5py(PythonPackage):
     depends_on("py-mpi4py", when="@:2 +mpi", type=("build", "run"))
 
     # Historical dependencies
-    depends_on("py-cached-property@1.5:", type=("build", "run"), when="@:3.6 ^python@:3.7")
     depends_on("py-six", type=("build", "run"), when="@:2")
 
     def flag_handler(self, name, flags):
--- a/repos/spack_repo/builtin/packages/py_h5py/package.py
+++ b/repos/spack_repo/builtin/packages/py_h5py/package.py
@@ -53,12 +53,7 @@ class PyH5py(PythonPackage):
 
     depends_on("c", type="build")
 
-    # Python versions
-    # python@3.9+ is required in 3.12+, but 3.11 needs python@3.9+ for numpy@2
-    # as python@3.8 is deprecated in spack, harmonized on python@3.9+
-    # https://github.com/h5py/h5py/blob/3.11.0/pyproject.toml#L5
-    depends_on("python@3.9:", type=("build", "run"), when="@3.11:")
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/py_h5py/package.py
```diff
@@ -53,38 +53,33 @@ class PyH5py(PythonPackage):
 
     depends_on("c", type="build")
 
-    # Python versions
-    # python@3.9+ is required in 3.12+, but 3.11 needs python@3.9+ for numpy@2
-    # as python@3.8 is deprecated in spack, harmonized on python@3.9+
-    # https://github.com/h5py/h5py/blob/3.11.0/pyproject.toml#L5
-    depends_on("python@3.9:", type=("build", "run"), when="@3.11:")
-    depends_on("python@:3.9", type=("build", "run"), when="@:2.8")
+    depends_on("python@3.9:", type=("build", "run"))
```
adamjstewart: This is not true for older versions, which support older Python. However, Spack itself no longer supports installing Python 3.8 and older. So you could probably just delete this line.

repos/spack_repo/builtin/packages/py_h5py/package.py
```diff
@@ -53,38 +53,33 @@ class PyH5py(PythonPackage):
...
     # until 3.11.0, uses "oldest-supported-numpy" and doesn't support numpy2
     # https://github.com/scipy/oldest-supported-numpy/blob/main/setup.cfg
     # pre-3.11 is numpy@1 only
     # https://github.com/h5py/h5py/issues/2353
+    # https://github.com/h5py/h5py/blob/3.11.0/docs/whatsnew/3.11.rst#new-features
     depends_on("py-numpy@:1", when="@:3.10", type=("build", "run"))
-    depends_on("py-numpy@1.19.3:", type=("build", "run"), when="@3:3.10 ^python@3.9:")
-    depends_on("py-numpy@1.17.5:", type=("build", "run"), when="@3:3.5 ^python@3.8.0:3.8")
+    depends_on("py-numpy@1.19.3:", type=("build", "run"), when="@3:3.10")
     depends_on("py-numpy@1.7:", type=("build", "run"), when="@:2")
```
adamjstewart: All of this is wrong. Based on `setup.py`, I believe it should be:
```python
depends_on("py-numpy@1.19.3:", type=("build", "run"), when="@3.12:")
depends_on("py-numpy@1.17.5:", type=("build", "run"), when="@3.9:")
depends_on("py-numpy@1.14.5:", type=("build", "run"), when="@3.2:")
depends_on("py-numpy@1.12:", type=("build", "run"), when="@3.0:")
depends_on("py-numpy@1.7:", type=("build", "run"), when="@2.7:")
depends_on("py-numpy@1.6.1:", type=("build", "run"))
```
However again, note that Spack no longer supports any of these older versions of numpy, and they will be removed in #662 anyway.

adamjstewart (changes requested):
Thanks for catching this. You are correct that numpy 2+ is not strictly required at run-time. I'm guessing the mistake was because numpy 2+ is requested at build-time by pyproject.toml, but this is only needed for ABI compatibility. Since Spack builds everything from source and ensures a single version of numpy will be in the DAG, this doesn't affect us.

Chrismarsh:
I was overly cautious interpreting the project.toml 😢

repos/spack_repo/builtin/packages/py_h5py/package.py
Chrismarsh: I'll double check but that numpy@:1 needs to exist as it doesn't work until 3.11 with nunpy@2. If those numpy versions are removed in Spack I think we should remove these here?

adamjstewart: Yes, the upper bound on 1 does need to exist. 

We typically don't remove lower bounds even if they don't exist in Spack (except for Python, and even that is controversial).

Chrismarsh: I don't think this is still quite right, e.g., 
3.11: needs `@1.17.3:` https://github.com/h5py/h5py/blob/3.11.0/setup.py#L34

I've gone through this again and I think it's right now?

I find this
https://github.com/h5py/h5py/blob/3.0.0/setup.py#L28
confusing. Does one just target the python 3.6 like you did and let py-numpy handle any python version requirements?

Chrismarsh:
I have also added the new version 3.14

repos/spack_repo/builtin/packages/py_h5py/package.py
adamjstewart: Yeah, I would just target the lower bound of h5py, not the lower bound of Python.

repos/spack_repo/builtin/packages/py_h5py/package.py
```diff
@@ -53,39 +54,29 @@ class PyH5py(PythonPackage):
...
-    depends_on("python@3.9:", type=("build", "run"), when="@3.11:")
-    depends_on("python@:3.9", type=("build", "run"), when="@:2.8")
-
     # Build dependencies
     # h5py@3.11 can build with cython@3.x
     depends_on("py-cython@0.29.31:3", type="build", when="@3.11:")
     depends_on("py-cython@0.29.31:0", type="build", when="@3.9:3.10")
-    depends_on("py-cython@0.29.15:0", type=("build"), when="@3:3.7 ^python@3.9.0:")
+    depends_on("py-cython@0.29.15:0", type=("build"), when="@3:3.7")
     depends_on("py-cython@0.29.14:0", type=("build"), when="@3:3.7 ^python@3.8.0:3.8")
```
adamjstewart: Could prob delete this line, Python 3.8 is no longer supported by Spack

repos/spack_repo/builtin/packages/py_h5py/package.py
```diff
@@ -53,39 +54,29 @@ class PyH5py(PythonPackage):
...
     depends_on("py-cython@0.29:0", type=("build"), when="@3.0:3.10")
     depends_on("py-cython@0.23:0", type="build", when="@:2")
+
     depends_on("py-pkgconfig", type="build")
     depends_on("py-setuptools@61:", type="build", when="@3.8.0:")
+    depends_on("py-setuptools@77:", type="build", when="@3.14:")
```
adamjstewart: I like to sort depends_on() from newest to oldest, just like with version()

adamjstewart (dismissed):

adamjstewart:
Failing tests are due to this PR, it looks like the numpy 2 lower bound during build is now being enforced as of h5py 3.14: https://github.com/h5py/h5py/blob/3.14.0/setup_build.py#L130

This is quite unfortunate, as package managers like Spack do not allow different versions of dependency like numpy at build-time and run-time like pip does.

repos/spack_repo/builtin/packages/py_h5py/package.py
Chrismarsh: removed another at the bottom too

Chrismarsh:
I've enforced `numpy@2:` and asked if this is intended 
https://github.com/h5py/h5py/issues/2644

repos/spack_repo/builtin/packages/py_h5py/package.py
```diff
@@ -53,39 +54,31 @@ class PyH5py(PythonPackage):
...
-    depends_on("python@3.9:", type=("build", "run"), when="@3.11:")
-    depends_on("python@:3.9", type=("build", "run"), when="@:2.8")
-
     # Build dependencies
     # h5py@3.11 can build with cython@3.x
     depends_on("py-cython@0.29.31:3", type="build", when="@3.11:")
     depends_on("py-cython@0.29.31:0", type="build", when="@3.9:3.10")
-    depends_on("py-cython@0.29.15:0", type=("build"), when="@3:3.7 ^python@3.9.0:")
-    depends_on("py-cython@0.29.14:0", type=("build"), when="@3:3.7 ^python@3.8.0:3.8")
+    depends_on("py-cython@0.29.15:0", type=("build"), when="@3:3.7")
```
adamjstewart, line 61: Technically this could be `0.29.14:0` if someone really wanted to build with older Python. It isn't supported by Spack, but it is supported by older h5py.

adamjstewart (approved):
I think this looks good now, but would like to give our h5py maintainers time to review: @bryanherman @takluyver

repos/spack_repo/builtin/packages/py_h5py/package.py
Chrismarsh: Since it wasn't supported by spack that's why I did it like this. I think just leave it unless the maintainers want it

PR #1029, opened 2025-08-09
aom: add v3.12.1

michaelkuhn:
This also makes the following changes:
- Use nasm instead of yasm, since the latter seems to be unmaintained
- Update v1.0.0-errata1's checksum, which seems to have been wrong all along (see https://aomedia.googlesource.com/aom/+/refs/tags/v1.0.0-errata1)
- Deprecate v1.0.0-errata1

### DIFF
--- a/repos/spack_repo/builtin/packages/aom/package.py
+++ b/repos/spack_repo/builtin/packages/aom/package.py
@@ -15,11 +15,17 @@ class Aom(CMakePackage):
 
     license("BSD-2-Clause AND AOM-Patent-License-1.0", checked_by="tgamblin")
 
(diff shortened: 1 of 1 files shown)


bernhardkaindl (approved):
Built on x86-linux:
```
spack find -v aom
-- linux-ubuntu22.04-skylake / %c,cxx=gcc@13.1.0 ----------------
aom@3.12.1~ipo build_system=cmake build_type=Release commit=10aece4157eb79315da205f39e19bf6ab3ee30d0 generator=make
aom@3.12.1+ipo build_system=cmake build_type=Release commit=10aece4157eb79315da205f39e19bf6ab3ee30d0 generator=make
==> 2 installed packages
```

PR #1038, opened 2025-08-11
py-lit: add v12.0.1, v13.0.1, v14.0.6, v15.0.7, v16.0.6, v17.0.6, v18.1.8

pearzt:


alalazo (approved):
It still seem to depend on `setuptools` only: https://github.com/llvm/llvm-project/blob/main/llvm/utils/lit/pyproject.toml

PR #1048, opened 2025-08-11
bgen: update package

teaguesterling:
This improves the bgen package to use Spack-provided dependencies instead of the dependencies vendored into the package. It also adds variants that install headers and libraries, allowing it to be used as a dependency for other packages.


EbiArnie (approved):
Looks good to me.
Thanks!

PR #1054, opened 2025-08-11
curl: add NMake deprecation acknowledgement

johnwparent:
Curl deprecated their NMake build system with 8.12, which requires a special argument to continue supporting. Add that argument.


johnwparent:
@Chrismarsh

alalazo (approved):

PR #1055, opened 2025-08-11
Curl: add CMake build system

johnwparent:
Curl added a CMake build system a while back, finally port it to Spack's recipe.

Adds appropriate cutover from NMake (being removed in a month (sept 2025) to CMake on Windows

### DIFF
--- a/repos/spack_repo/builtin/packages/curl/package.py
+++ b/repos/spack_repo/builtin/packages/curl/package.py
@@ -431,9 +431,9 @@ class CMakeBuilder(CMakeBuilder):
             )
             if self.spec.satisfies("+ldap"):
                 args.append(self.define("USE_WIN32_LDAP", True))
-
-        if "shared" in self.spec.variants["libs"]:
+            
+        if self.spec.satisfies("libs=shared"):
             args.append(self.define("BUILD_SHARED_LIBS", True))
-        if "static" in self.spec.variants["libs"]:
+        if self.spec.satisfies("libs=static"):
             args.append(self.define("BUILD_STATIC_LIBS", True))
         return args
--- a/repos/spack_repo/builtin/packages/curl/package.py
+++ b/repos/spack_repo/builtin/packages/curl/package.py
@@ -8,8 +8,8 @@ import re
 import sys
 
 from spack_repo.builtin.build_systems.autotools import AutotoolsBuilder, AutotoolsPackage
-from spack_repo.builtin.build_systems.nmake import NMakeBuilder, NMakePackage
 from spack_repo.builtin.build_systems.cmake import CMakeBuilder, CMakePackage
+from spack_repo.builtin.build_systems.nmake import NMakeBuilder, NMakePackage
 
 from spack.package import *
 
@@ -146,7 +146,6 @@ class Curl(NMakePackage, AutotoolsPackage, CMakePackage):
     # 3.0, which curl@7.63 requires
     depends_on("cmake@:3", when="build_system=cmake @:7.63")
 
-
     depends_on("gnutls", when="tls=gnutls")
 
     with when("tls=mbedtls"):
@@ -182,7 +181,9 @@ class Curl(NMakePackage, AutotoolsPackage, CMakePackage):
     # https://github.com/curl/curl/pull/9054
     patch("easy-lock-sched-header.patch", when="@7.84.0")
 
-    build_system("autotools", "cmake", conditional("nmake", when="@:8.11 platform=windows"), default="cmake")
+    build_system(
+        "autotools", "cmake", conditional("nmake", when="@:8.11 platform=windows"), default="cmake"
+    )
 
     @classmethod
     def determine_version(cls, exe):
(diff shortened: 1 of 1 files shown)


johnwparent:
@spackbot fix style

repos/spack_repo/builtin/packages/curl/package.py
```diff
@@ -172,7 +181,9 @@ class Curl(NMakePackage, AutotoolsPackage):
     # https://github.com/curl/curl/pull/9054
     patch("easy-lock-sched-header.patch", when="@7.84.0")
 
-    build_system("autotools", conditional("nmake", when="platform=windows"), default="autotools")
+    build_system(
+        "autotools", "cmake", conditional("nmake", when="@:8.11 platform=windows"), default="cmake"
```
haampie: `default=autotools` please, cmake depends on curl.

haampie (dismissed):
defaults

repos/spack_repo/builtin/packages/curl/package.py
```diff
@@ -374,3 +385,55 @@ def install(self, pkg, spec, prefix):
...
+            self.define_from_variant("CURL_USE_OPENLDAP", "ldap"),
+            self.define_from_variant("CURL_DISABLE_LDAP", "ldap"),
+            self.define_from_variant("USE_NGHTTP2", "nghttp2"),
+            self.define_from_variant("CURL_USE_GSSAPI", "gssapi"),
+            self.define_from_variant("USE_LIBRTMP", "librtmp"),
+            self.define_from_variant("USE_LIBIDN2", "libidn2"),
+        ]
+
+        if self.spec.satisfies("tls=sspi"):
+            self.define("CURL_WINDOWS_SSPI", True)
```
haampie:
```suggestion
        self.define("CURL_WINDOWS_SSPI", self.spec.satisfies("tls=sspi"))
```
as part of  `args = [...]`.

Right now you're producing an unused string.

repos/spack_repo/builtin/packages/curl/package.py
```diff
@@ -374,3 +385,55 @@ def install(self, pkg, spec, prefix):
...
+            args.extend(
+                [
+                    self.define_from_variant("ENABLE_UNICODE", "unicode"),
+                    self.define_from_variant("CURL_STATIC_CRT", "static-crt"),
+                ]
+            )
+            if self.spec.satisfies("+ldap"):
+                args.append(self.define("USE_WIN32_LDAP", True))
+
+        if "shared" in self.spec.variants["libs"]:
```
haampie: use `satisfies` because it's easier to grep.

repos/spack_repo/builtin/packages/curl/package.py
johnwparent: doh!

repos/spack_repo/builtin/packages/curl/package.py
johnwparent: Ahh, for some reason I thought the concretizer was able to handle resolving this too, I don't know why I thought that.

It seems like Curl is trying to deprecate their non CMake build systems, do we maybe want to pin some versions that should be built with autotools/nmake and let everything else build w/ CMake?

repos/spack_repo/builtin/packages/curl/package.py
johnwparent: Good suggestion, thanks

repos/spack_repo/builtin/packages/curl/package.py
haampie: Yeah, the concretizer will be able to handle it, but it will always result in a penalty for that variant, and potentially there can be multiple ways to work around it, leading to non-determinism in the solver (I did not check this, but it could :p)

haampie: How about `default="cmake" if sys.platform == "win32" else "autotools"` or something?

johnwparent:
@spackbot fix style

repos/spack_repo/builtin/packages/curl/package.py
johnwparent: I went with that, was hoping to have some sort of per-version scheme, but this is fine for the time being. If/when this becomes an issue on Windows I'll tackle it then.

repos/spack_repo/builtin/packages/curl/package.py
```diff
@@ -181,9 +181,12 @@ class Curl(NMakePackage, AutotoolsPackage, CMakePackage):
     # https://github.com/curl/curl/pull/9054
     patch("easy-lock-sched-header.patch", when="@7.84.0")
 
-    platform_default = "cmake" if sys.platform =="win32" else "autotools"
+    platform_default = "cmake" if sys.platform == "win32" else "autotools"
```
haampie: this defines a class attribute, I would inline it

johnwparent: I personally don't love inlining logic like this in method calls this way, but polluting the class api is also/more gross, so fair enough. 

I realize we inline a lot of the Windows logic like this, really wish there was a better way.

johnwparent: Updated.

kwryankrattiger (approved):

PR #1060, opened 2025-08-12
regenie: new pacakge

teaguesterling:
This adds the regenie package, which depends on bgen (#1048). This package uses spack dependencies but leaves a variant to allow building with the vendored deps if desired.


EbiArnie (approved):
Wow, quite a bit of work for this one.
I think this looks fine.
I also validated the checksum.

Thanks!

PR #1067, opened 2025-08-12
eigen: Use Release build type by default

jcortial-safran:

### DIFF
--- a/repos/spack_repo/builtin/packages/eigen/package.py
+++ b/repos/spack_repo/builtin/packages/eigen/package.py
@@ -76,7 +76,7 @@ class Eigen(CMakePackage, ROCmPackage):
     # library."
     variant(
         "build_type",
-        default="RelWithDebInfo",
+        default="Release",
         description="The build type to build",
         values=("Debug", "Release", "RelWithDebInfo"),
     )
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/eigen/package.py
```diff
@@ -76,7 +76,7 @@ class Eigen(CMakePackage, ROCmPackage):
     # library."
     variant(
         "build_type",
-        default="RelWithDebInfo",
+        default="Release",
         description="The build type to build",
         values=("Debug", "Release", "RelWithDebInfo"),
     )
```
wdconinc: Can we just get rid of this entirely? It's provided by `CMakePackage` and may just be here because it predates that functionality.

https://github.com/spack/spack-packages/blob/3902c3da2e160ec366268a669516b481a441a76c/repos/spack_repo/builtin/build_systems/cmake.py#L197-L202

(and there is no reason or record in their repo of `eigen` not supporting MinSizeRel)

jcortial-safran: Ok, it's even better.

HaoZeke (approved):
LTGTM. Thanks @jcortial-safran

wdconinc (approved):

PR #1071, opened 2025-08-12
add newest zsh version

PDoakORNL:
Zsh 5.9 has build issues with newer gcc's, newer distros and/or ncurses.
see issue #1024
This should supercede PR's #1025, #567

This PR adds version 5.9.0.2 a "test" version which fixes these issues for me.

Since the zsh release schedule seems quite conservative and many distros just apply a large number of patches, I  don't really anticipate this becoming an official "release" in a timely manner. I have found no issues with zsh 5.9.0.2 on ubuntu or the SUSE derive cray os on Frontier. 

Spack does provide git branch or commit based versions of some packages but here I think just providing the newest test tarball is sufficient.


repos/spack_repo/builtin/packages/zsh/package.py
```diff
@@ -16,8 +16,16 @@ class Zsh(AutotoolsPackage):
     homepage = "https://www.zsh.org"
     url = "https://downloads.sourceforge.net/project/zsh/zsh/5.4.2/zsh-5.4.2.tar.xz"
 
+    def url_for_version(self, version):
```
tldahlgren: Please move this below the directives (i.e., around line 56) so it is the first method in the class.

tldahlgren (changes requested):
Confirmed all of the version sha256s (since `url_for_version` is introduced).

Please move the method below the directives.

tldahlgren (dismissed):
(Re)Confirmed all of the version sha256s since `url_for_version` is introduced.

tldahlgren:
Looks like there are several CI build failures related to `zsh` (e.g., https://gitlab.spack.io/spack/spack-packages/-/jobs/20203696).

repos/spack_repo/builtin/packages/zsh/package.py
```diff
@@ -47,6 +49,13 @@ class Zsh(AutotoolsPackage):
 
     patch("pointer-types.patch", when="@5.6.2:")
 
+    def url_for_version(self, version):
+        if version <= Version("5.9.0.0"):
+            url = "https://downloads.sourceforge.net/project/zsh/zsh/{0}/zsh-{0}.tar.xz"
+        if version > Version("5.9.0.1"):
```
becker33:
```suggestion
        else:
```

becker33 (changes requested):

becker33 (dismissed):

pdsc-limited:
I've not had time to follow up on this till now and an official 5.9.2 release has been made.  Updating to reflect that, for my use case the test versions are no longer needed so I've dropped them.  The need for the patch seems to have finally been dealt with in the configure.ac for 5.9.2.

PDoakORNL:
I've not had time to follow up on this till now and an official 5.9.2 release has been made. Updating to reflect that, for my use case the test versions are no longer needed so I've dropped them. The need for the patch seems to have finally been dealt with in the configure.ac for 5.9.2.

I cannot replicate #4896's `timeout 3 zsh -c '$(echo echo success)' || echo failure` issue with llvm 21.1.4 and 5.9.2 so this does not address that PR's concern one way or another. Nor do I add the build time autoconf which seems implied by this being an AutotoolsProject subtype.

PR #1072, opened 2025-08-12
Win: Numpy in CI

johnwparent:
Adds Numpy to Gitlab CI

Numpy allows us to validate a blas provider and python with one root package, and is a high value/traffic package

Puts us one step closer to testing AI stacks on Windows

Removes:

* b4s
* netlib-lapack

Adds:

* py-numpy

Modifies `netlib-lapack` to search for non runtime libraries in it's `libs` methods on Windows (sets runtime for all platforms but its a no-op everywhere else)


johnwparent:
@spackbot fix style

scheibelp (approved):

PR #1075, opened 2025-08-13
seacas: allow to compile older versions with recent GCC

jcortial-safran:
Allows seacas 2022-01, 2022-02 and 2022-03 to compile correctly with recent GCC compilers. Tested with GCC 13.2. The flag exists since GCC 10 so it should be safe to add it as soon as GCC is sufficiently recent.


gsjaardema (approved):
That looks good to me. Thanks.

alalazo (approved):

PR #1077, opened 2025-08-13
vtk: rework missing standard include management

jcortial-safran:
The build of vtk@9.1 is currently broken on gcc@13:. The recipe tries to apply two conflicting patches related to missing standard includes.

The PR reworks and simplifies the logic as follows:
- The redundant patch is eliminated
- The contraint on compiler version is removed since it was overly restrictive (only activated for gcc, it should be applied for any standard-compliant compilers as well) and the patch should be safe to apply even for older compilers
- The logic is gathered in a single place in the recipe with reworked comments

### DIFF
--- a/repos/spack_repo/builtin/packages/vtk/package.py
+++ b/repos/spack_repo/builtin/packages/vtk/package.py
@@ -74,17 +74,19 @@ class Vtk(CMakePackage):
 
     patch("gcc.patch", when="@6.1.0")
 
-    # patches to fix some missing stl includes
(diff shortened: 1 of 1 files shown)


vicentebolea (commented):
Thanks for getting rid of this redundancy with the patches, I am testing the vtk@9.1 build in my computer I will approve once it passes

vicentebolea (approved):

johnwparent (approved):

PR #1078, opened 2025-08-13
py-jupyterlab-server: add the "build-tools" tag

alalazo:
refers https://github.com/spack/spack/issues/51134

This identifies the package as possibly needing more than one node in a DAG, due to conflicting build constraints of other packages


alalazo:
@spackbot fix style

adamjstewart (approved):
So this will now apply to build/run deps of build deps?

alalazo:
This potentially allows the package to have its own unification set, and to have multiple nodes. It depends on the configuration values of the `concretizer` section.

PR #1084, opened 2025-08-13
py-dm-tree: add v0.1.9; depends_on abseil-cpp cxxstd=14

wdconinc:
This PR adds `py-dm-tree`, v0.1.9, and enforces the c++14 standard as [hardcoded](https://github.com/google-deepmind/tree/blob/0.1.8/tree/CMakeLists.txt#L15) in the CMakeLists.txt. This package will be built in CI (but the requirement of `cxxstd=14` on `abseil-cpp` may cause the concretizer to pick 0.1.7).


wdconinc:
> but the requirement of cxxstd=14 on abseil-cpp may cause the concretizer to pick 0.1.7

The new version was built in CI.

alalazo (approved):

PR #1088, opened 2025-08-14
podio: Add latest tag 1.4 and 1.4.1

tmadlener:
Add the latest tag of podio: https://github.com/AIDASoft/podio/releases/tag/v01-04

### DIFF
--- a/repos/spack_repo/builtin/packages/podio/package.py
+++ b/repos/spack_repo/builtin/packages/podio/package.py
@@ -21,6 +21,7 @@ class Podio(CMakePackage):
     tags = ["hep", "key4hep"]
 
     version("master", branch="master")
+    version("1.4", sha256="f8b7f5ba965ff58270d617f50f168a4683a98fbcd643b66f1852bec960e02bbd")
     version("1.3", sha256="7efdf049822f171f4da5e83a7101096c066679904e59e741f3c2833ccda5e363")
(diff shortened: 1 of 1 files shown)


wdconinc:
@spackbot run pipeline

wdconinc (approved):
Looks good to me.

Curious whether cxxstd=23 is supported since it was added to CMakeLists.txt in https://github.com/AIDASoft/podio/commit/2e232bc972ab5d2660339c4a160c804ac26926f0 (v1.3).

tmadlener:
Oh yeah, that should work, but that is just changing the CMake check to also allow c++23. There is no use of c++23 yet in podio (or generated EDMs). Shall I add the corresponding changes to the `cxxstd` variant in this PR?

wdconinc:
> Oh yeah, that should work, but that is just changing the CMake check to also allow c++23. There is no use of c++23 yet in podio (or generated EDMs). Shall I add the corresponding changes to the `cxxstd` variant in this PR?

Up to you. Sometimes these changes get through the cracks if there's no explicit entry in the release notes to call it out. I just wanted to flag it in case `cxxstd=23` is now supported (i.e. tested in podio CI and expected to work).

tmadlener:
I will have to confirm that we actually have a test in CI that runs on top of a `root cxxstd=23` for podio. Let's do that in a followup PR once that is all done properly.

PR #1091, opened 2025-08-14
Restrict ispc dependency in openvkl to make it build successfully

paulmelis:
Disabled ispc 1.19.0 as well, as it does not build without error


repos/spack_repo/builtin/packages/ispc/package.py
```diff
@@ -34,7 +34,8 @@ class Ispc(CMakePackage):
     version("1.21.1", sha256="99bbb1d1f15bc4433d6a63b5bb35b321af3e3af753c3b28a61850d1748e8a89f")
     version("1.21.0", sha256="023782f721bfb5893bac24bc2153a8214c916be82c290bf63a3ec6678949b5ef")
     version("1.20.0", sha256="8bd30ded7f96859451ead1cecf6f58ac8e937288fe0e5b98c56f6eba4be370b4")
-    version("1.19.0", sha256="c1aeae4bdfb28004a6949394ea1b3daa3fdf12f646e17fcc0614861077dc8b6a")
+    # Does not compile, error related to missing stdint.h/cstdint include
+    # version("1.19.0", sha256="c1aeae4bdfb28004a6949394ea1b3daa3fdf12f646e17fcc0614861077dc8b6a")
```
alalazo: We usually deprecate versions before deleting them. Might be worth considering a patch if it's easy to fix (or deprecating more versions, if they are outdated and not used).

alalazo (changes requested):

alalazo:
@aumuell

aumuell (commented):
On which OS and compiler does this build error related to <cstdint> occur?
This sounds like something that could be patched easily.

Or a `conflicts` together with a deprecation of this version could resolve the build error?

paulmelis:
I screwed up the merge/rebase of this PR. Opening new PR #1192, which also only concerns ispc and not openvkl (the latter does not need patching)

PR #1092, opened 2025-08-14
Add new version for all AOCL libs and update aocl_libmem and amdlibm

amd-toolchain-support:
This PR includes below main changes in addition to few minor supporting changes:
- Adding new version 5.1 for AOCL libraries: 
`amdblis` `amdfftw` `amdlibflame` `amdlibm` `amdscalapack` `aocl-compression` `aocl-crypto` `aocl-da` `aocl-libmem` `aocl-sparse` `aocl-utils` and corresponding bundle package `amd-aocl`
- Adding `CMake` support for `amdlibm` with version `5.1`
- Update compiler dependency for `amdlibm`
- Update `aocl-libmem` recipe to include new variants `dynamic-dispatch` and update existing variant `vectorization` specific to version `5.1`
- Align the CMake version dependency for all libraries.

PR #1093, opened 2025-08-14
Add new version for amduprof 5.1.701

amd-toolchain-support:
This PR adds a new version `5.1.701` for `amduprof`

### DIFF
--- a/repos/spack_repo/builtin/packages/amduprof/package.py
+++ b/repos/spack_repo/builtin/packages/amduprof/package.py
@@ -23,10 +23,15 @@ class Amduprof(Package):
     maintainers("amd-toolchain-support")
 
     version(
+        "5.1.701",
+        sha256="8fd83170170883a6617391609545dffd557d6ca4a8f8f00a7a8a2d6cdee08189",
+        url="file://{0}/AMDuProf_Linux_x64_5.1.701.tar.bz2".format(os.getcwd()),
+        preferred=True,
+    )
+    version(
         "5.0.1479",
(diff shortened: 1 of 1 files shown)


amd-toolchain-support:
@zzzoom , adding you as reviewer based on suggestion.

repos/spack_repo/builtin/packages/amduprof/package.py
```diff
@@ -22,11 +22,16 @@ class Amduprof(Package):
 
     maintainers("amd-toolchain-support")
 
+    version(
+        "5.1.701",
+        sha256="8fd83170170883a6617391609545dffd557d6ca4a8f8f00a7a8a2d6cdee08189",
+        url="file://{0}/AMDuProf_Linux_x64_5.1.701.tar.bz2".format(os.getcwd()),
+        preferred=True,
+    )
```
wdconinc: Preferred doesn't do anything when it's on the most recent version.

wdconinc (commented):
Same comment here as in #1092 :-)

repos/spack_repo/builtin/packages/amduprof/package.py
amd-toolchain-support: updated as suggested.

amd-toolchain-support:
> Same comment here as in #1092 :-)

Thank you @wdconinc for your time and attention.
I have addressed the review and comments.

wdconinc (approved):

zzzoom:
For what it's worth I still think that if the package is so adamant about its EULA it should require a `+license-agreed` variant set like AOCC does, instead of forcing users to download the package by hand. This is just pestering people that want to make your tools available to their users.

PR #1123, opened 2025-08-19
intel-oneapi-compilers-classic updates

becker33:
This PR addresses three issues:

1. external packages for `intel-oneapi-compilers-classic` currently fail because they need to query their `intel-oneapi-compilers` dependency. This is fixed by changing which paths are used in `setup_run_environment`.
2. fortran compilers from `intel-oneapi-compilers-classic` are not currently usable with C/C++ compilers from a version of `intel-oneapi-compilers` other than the one the fortran was built with. This is fixed by changing the dependency type from `run` to `build`, and by changing from symlinks to copies to ensure that the "build" dependency type is accurate.
3. newer fortran compilers for `intel-oneapi-compilers-classic` need to be added, but their locations within `intel-oneapi-compilers` are different from previous versions. This is fixed by adding special logic for versions `2021.11.0:` as those versions only include `ifort`

### DIFF
--- a/repos/spack_repo/builtin/packages/intel_oneapi_compilers_classic/package.py
+++ b/repos/spack_repo/builtin/packages/intel_oneapi_compilers_classic/package.py
@@ -2,6 +2,7 @@
 #
 # SPDX-License-Identifier: (Apache-2.0 OR MIT)
 import os
+import pathlib
 import sys
 
 from spack_repo.builtin.build_systems.compiler import CompilerPackage
--- a/repos/spack_repo/builtin/packages/intel_oneapi_compilers_classic/package.py
+++ b/repos/spack_repo/builtin/packages/intel_oneapi_compilers_classic/package.py
@@ -93,10 +93,6 @@ class IntelOneapiCompilersClassic(Package, CompilerPackage):
            $ source {prefix}/{component}/{version}/env/vars.sh
         and from setting CC/CXX/F77/FC
         """
-        _link_path = pathlib.Path(os.readlink(self.prefix.bin.icc))
-        vars_path = _link_path.parent.parent / "env" / "vars.sh"
-        env.extend(EnvironmentModifications.from_sourcing_file(vars_path))
-
         env.set("CC", self.prefix.bin.icc)
         env.set("CXX", self.prefix.bin.icpc)
         env.set("F77", self.prefix.bin.ifort)
--- a/repos/spack_repo/builtin/packages/intel_oneapi_compilers_classic/package.py
+++ b/repos/spack_repo/builtin/packages/intel_oneapi_compilers_classic/package.py
@@ -117,12 +117,30 @@ class IntelOneapiCompilersClassic(Package, CompilerPackage):
             env.append_flags("SPACK_ALWAYS_FFLAGS", "-diag-disable=10448")
 
     def install(self, spec, prefix):
-        # If we use symlinks, we create a run dependency on oneapi compilers
-        # preventing ifort use with a different version of icx
+        # List of all binaries from intel-oneapi-compilers related to ifort
+        binaries = [
+            "codecov",
+            "fortcom",
+            "fpp",
+            "ifort",
+            "ifort.cfg",
+            "map_opts",
+            "profdcg",
+            "profmerge",
+            "profmergesampling",
+            "proforder",
+            "tselect",
+            "xiar",
+            "xiar.cfg",
+            "xild",
+            "xild.cfg",
+        ]
+
+        # We do a full copy (not symlinks) to avoid a run dependency on intel-oneapi-compilers
+        # which would prevent mixing versions in a DAG
         mkdirp(prefix.bin)
-        install(self.oneapi_compiler_prefix.bin.ifort, prefix.bin)
-        install(self.oneapi_compiler_prefix.bin.fortcom, prefix.bin)
-        install(self.oneapi_compiler_prefix.bin.join("ifort.cfg"), prefix.bin)
+        for binary in binaries:
+            install(self.oneapi_compiler_prefix.bin.join(binary), prefix.bin)
         install_tree(self.oneapi_compiler_prefix.lib, prefix.lib)
         install_tree(self.oneapi_compiler_prefix.opt, prefix.opt)
         install_tree(self.oneapi_compiler_prefix.etc, prefix.etc)
@@ -130,8 +148,8 @@ class IntelOneapiCompilersClassic(Package, CompilerPackage):
 
     @when("@:2021.10")
     def install(self, spec, prefix):
-        # If we use symlinks, we create a run dependency on oneapi compilers
-        # preventing ifort use with a different version of icx
+        # We do a full copy (not symlinks) to avoid a run dependency on intel-oneapi-compilers
+        # which would prevent mixing versions in a DAG
         install_tree(self.oneapi_compiler_prefix.linux.bin.intel64, prefix.bin)
         install_tree(self.oneapi_compiler_prefix.linux.lib, prefix.lib)
         install_tree(self.oneapi_compiler_prefix.linux.include, prefix.include)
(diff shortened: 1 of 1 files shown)


becker33:
@rscohn2 you're listed as the maintainer for this package, can you let me know what you think of this?

repos/spack_repo/builtin/packages/intel_oneapi_compilers_classic/package.py
```diff
@@ -126,27 +117,26 @@ def setup_dependent_build_environment(
...
+        install(self.oneapi_compiler_prefix.bin.fortcom, prefix.bin)
+        install(self.oneapi_compiler_prefix.bin.join("ifort.cfg"), prefix.bin)
+        install_tree(self.oneapi_compiler_prefix.lib, prefix.lib)
+        install_tree(self.oneapi_compiler_prefix.opt, prefix.opt)
+        install_tree(self.oneapi_compiler_prefix.etc, prefix.etc)
+        install_tree(self.oneapi_compiler_prefix.share, prefix.share)
+
+    @when("@:2021.10")
+    def install(self, spec, prefix):
+        # If we use symlinks, we create a run dependency on oneapi compilers
```
kwryankrattiger:
```suggestion
        # We don't use symlinks to avoid creating a run dependency on oneapi compilers
```

kwryankrattiger: I read this a couple times and without the context of "it used to be symlinks" it isn't really clear what you are saying.

Suggested change to make it say what we _are_ doing and why.

repos/spack_repo/builtin/packages/intel_oneapi_compilers_classic/package.py
```diff
@@ -126,27 +117,26 @@ def setup_dependent_build_environment(
...
-        # Symlink all files in src to dest keeping directories as dirs
-        for entry in os.listdir(src):
-            src_path = os.path.join(src, entry)
-            dest_path = os.path.join(dest, entry)
-            if os.path.isdir(src_path) and os.access(src_path, os.X_OK):
-                link_tree = LinkTree(src_path)
-                link_tree.merge(dest_path)
-            else:
-                os.symlink(src_path, dest_path)
+        # If we use symlinks, we create a run dependency on oneapi compilers
```
kwryankrattiger: Same here

kwryankrattiger (approved):
LGTM, thanks!

repos/spack_repo/builtin/packages/intel_oneapi_compilers_classic/package.py
```diff
@@ -68,19 +66,26 @@ def compiler_version_regex(self):
         "2021.8.0": "2023.0.0",
         "2021.9.0": "2023.1.0",
         "2021.10.0": "2023.2.4",
-        "2021.11.1": "2024.0.2",
-    }.items():
+        "2021.11.0": "2024.0.0",
+        "2021.11.1": "2024.0.1:2024.0.2",
+        "2021.12.0": "2024.1.0",
+        "2021.13.0": "2024.2.0",
+        "2021.13.1": "2024.2.1",
```
rscohn2, line 73: I thought we had to stop updating this package when they stopped shipping icc because a toolchain without a C compiler would not be useful. I am unable to try it out, but do you know what happens?

hfp (commented):
Code wise the PR looks good to me with clear intent to improve things (which is good). I started testing it locally trying to run some specs with no luck to concretize (which might be followup work). I mostly tested with Spack 1.0.1 and tested, e.g., `spack spec dbcsr@develop %oneapi %fortran=intel@2021.12.0 ^intel-oneapi-mpi ^[virtuals=lapack] intel-oneapi-mkl@2025 ^[virtuals=blas] intel-oneapi-mkl@2025` (no local config aka externals hacked in just relying on what can be pulled). I will start a more formal review asap/quickly.

rscohn2 (approved):
No objections If you know using intel@2021.13.1 is usable as a fortran compiler even though there isn't a c compiler.

repos/spack_repo/builtin/packages/intel_oneapi_compilers_classic/package.py
becker33: Yeah, the `ifort` is still usable and I work with a few code teams that are stuck with `ifort` due to `ifx` either still not compiling their code or being sufficiently weak on performance for their code that it's non-viable.

becker33:
@hfp I changed the spec slightly to make it more legible for Spack 1.0, but I had success with:
```
spack spec dbcsr@develop %c,cxx=oneapi@2025 %fortran=intel@2021.12.0 ^intel-oneapi-mpi ^blas,lapack=intel-oneapi-mkl@2025
```
I think it's likely your issue was that you need to install the `intel-oneapi-compilers` and `intel-oneapi-compilers-classic` packages before doing this, since Spack only uses compilers that are either already installed or external

hfp:
> @hfp I changed the spec slightly to make it more legible for Spack 1.0, but I had success with:
> 
> ```
> spack spec dbcsr@develop %c,cxx=oneapi@2025 %fortran=intel@2021.12.0 ^intel-oneapi-mpi ^blas,lapack=intel-oneapi-mkl@2025
> ```
> 
> I think it's likely your issue was that you need to install the `intel-oneapi-compilers` and `intel-oneapi-compilers-classic` packages before doing this, since Spack only uses compilers that are either already installed or external

I have/had the following list but no luck:

> $> spack compiler list
==> Available compilers
-- gcc sles15-x86_64 --------------------------------------------
[e]  gcc@7.5.0  [e]  gcc@14.2.0  [e]  gcc@12.2.1  [e]  gcc@11.3.0
-- intel-oneapi-compilers sles15-x86_64 -------------------------
[e]  intel-oneapi-compilers@2025.2.0  [e]  intel-oneapi-compilers@2025.1.0  [e]  intel-oneapi-compilers@2024.1.0
-- intel-oneapi-compilers-classic sles15-x86_64 -----------------
[e]  intel-oneapi-compilers-classic@2021.12.0
-- llvm sles15-x86_64 -------------------------------------------
[e]  llvm@7.0.1

I have no objections on this PR, and it's approved. I still think the transition to Spack 1.0 caused breakages with external bits.

Let's sync internally on how to improve test coverage (or education/documentation on how to get the machinery under control). It can certainly help to test/fix things not working rather than enjoying tests that work.

hfp (approved):

becker33:
I'll take another stab at trying to reproduce the error using externals instead of spack-built compilers... I do want to catch any issues before putting this into develop if possible.

becker33:
@hfp any chance you could paste in the error message you're seeing? Could you also post the `packages.yaml` section for the intel packages you're using as externals?

I've copied the exact spec you used, modified it to take the `intel@2021.10.0` compiler because that's the one I have available as an external, and concretized it successfully.

I've also tried your exact spec without modifications using a spack-built `intel@2021.12.0`. That concretized successfully as well.

repos/spack_repo/builtin/packages/intel_oneapi_compilers_classic/package.py
```diff
@@ -90,24 +95,10 @@ def setup_run_environment(self, env: EnvironmentModifications) -> None:
...
-            # is external (or check extra_attributes if a user wants to supply it
-            # manually)
-        env.set("CC", bin_prefix.icc)
-        env.set("CXX", bin_prefix.icpc)
-        env.set("F77", bin_prefix.ifort)
-        env.set("FC", bin_prefix.ifort)
+        env.set("CC", self.prefix.bin.icc)
+        env.set("CXX", self.prefix.bin.icpc)
+        env.set("F77", self.prefix.bin.ifort)
+        env.set("FC", self.prefix.bin.ifort)
```
alalazo, line 101: This is in general wrong, if `self.spec` is not used for those languages. I'll submit a fix, after checking if we call this function to setup the run environment of a build dependency.

alalazo: Here is the part that concerns me:

https://github.com/spack/spack/blob/887ae03b6b2b6a36f7742b57446e23676f84ff98/lib/spack/spack/build_environment.py#L995-L997

We setup the run dependencies of direct build dependencies.

PR #1124, opened 2025-08-19
yafyaml: conflict ifx 2025.2

mathomp4:
Intel oneAPI has an `fpp` bug that is exposed in yafyaml, see https://community.intel.com/t5/Intel-Fortran-Compiler/Regression-with-fpp-2025-2-0/m-p/1703735

Until Intel can fix this (or we can somehow find a workaround), we have to conflict out ifx 2025.2.

NOTE: I have tested the newly released 2025.2.1 and it still has the bug.

cc: @climbfuji @rickgrubin-noaa

PR #1126, opened 2025-08-19
mapl: Allow ifx 2025

mathomp4:
At some point this:
```python
conflicts("^[virtuals=fortran] intel-oneapi-compilers")
```
was added to the MAPL package. However, MAPL does work with ifx 2025.0 and 2025.1 (I just tested them).

This PR updates it to use:
```python
    conflicts("^[virtuals=fortran] intel-oneapi-compilers@2025:", when="@:2.50")
```
which seems to be the spack v1 way of doing this for ifx.

NOTE: I am not sure if MAPL builds with 2025.2 yet. The main issue is the default way we run (with `+pflogger`) is broken at the moment as ifx (or rather fpp) 2025.2 does not compile a library needed by `pflogger` (see https://github.com/spack/spack-packages/pull/1124). I'm going to try and test to see if the fpp bug hits the MAPL code.

### DIFF
--- a/repos/spack_repo/builtin/packages/mapl/package.py
+++ b/repos/spack_repo/builtin/packages/mapl/package.py
@@ -277,7 +277,6 @@ class Mapl(CMakePackage):
     # MAPL can use ifx only from MAPL 2.51 onwards and only supports
     # ifx 2025.0 and newer due to bugs in ifx.
     conflicts("%oneapi@2025:", when="@:2.50")
-    conflicts("^[virtuals=fortran] intel-oneapi-compilers")
 
     variant("flap", default=False, description="Build with FLAP support", when="@:2.39")
     variant("pflogger", default=True, description="Build with pFlogger support")
--- a/repos/spack_repo/builtin/packages/mapl/package.py
+++ b/repos/spack_repo/builtin/packages/mapl/package.py
@@ -276,7 +276,7 @@ class Mapl(CMakePackage):
 
     # MAPL can use ifx only from MAPL 2.51 onwards and only supports
     # ifx 2025.0 and newer due to bugs in ifx.
-    conflicts("%oneapi@2025:", when="@:2.50")
+    conflicts("^[virtuals=fortran] intel-oneapi-compilers@2025:", when="@:2.50")
 
     variant("flap", default=False, description="Build with FLAP support", when="@:2.39")
     variant("pflogger", default=True, description="Build with pFlogger support")


mathomp4:
cc: @climbfuji for opinions

climbfuji:
I don't think this will work. `mapl` is totally fine with Intel oneAPI 2024 as long as you use `ifort` instead of `ifx` (and both compilers are shipped with Intel oneAPI 2024). In spack-stack, the agreement is that any oneAPI 2024 uses `ifort` because `ifx` has serious performance issues (in addition to the problems leading to this discussion). Any oneAPI 2025+ uses `ifx` because there is no more `ifort` and we need to move on.

mathomp4:
> I don't think this will work. `mapl` is totally fine with Intel oneAPI 2024 as long as you use `ifort` instead of `ifx` (and both compilers are shipped with Intel oneAPI 2024). In spack-stack, the agreement is that any oneAPI 2024 uses `ifort` because `ifx` has serious performance issues (in addition to the problems leading to this discussion). Any oneAPI 2025+ uses `ifx` because there is no more `ifort` and we need to move on.

@climbfuji So what would be the right spack way to say:

1. This works with all versions of `ifort`
2. MAPL 2.50+ work only with ifx 2025.0+

?

climbfuji:
> > I don't think this will work. `mapl` is totally fine with Intel oneAPI 2024 as long as you use `ifort` instead of `ifx` (and both compilers are shipped with Intel oneAPI 2024). In spack-stack, the agreement is that any oneAPI 2024 uses `ifort` because `ifx` has serious performance issues (in addition to the problems leading to this discussion). Any oneAPI 2025+ uses `ifx` because there is no more `ifort` and we need to move on.
> 
> @climbfuji So what would be the right spack way to say:
> 
> 1. This works with all versions of `ifort`
> 2. MAPL 2.50+ work only with ifx 2025.0+
> 
> ?

I don't know in spack v1. Trial and error I am afraid

mathomp4:
Okay. From what I see, what I have here now should be good. I think? 

Doing just:
```python
    conflicts("^[virtuals=fortran] intel-oneapi-compilers@2025:", when="@:2.50")
```

seems to work like I want. For example this works:
```console
$ spack spec mapl@2.59 %[virtuals=c,cxx,fortran] oneapi@2025.1.1
 -   mapl@2.59.0~debug+extdata2g~f2py+fargparse~ipo+pflogger+pfunit+shared+zstd build_system=cmake build_type=Release generator=make arch=linux-ubuntu24.04-icelake %c,cxx,fortran=oneapi@2025.1.1
 -       ^cmake@3.31.8~doc+ncurses+ownlibs~qtgui build_system=generic build_type=Release arch=linux-ubuntu24.04-icelake %c,cxx=oneapi@2025.1.1
...
```
and this fails:
```console
$ spack spec mapl@2.50 %[virtuals=c,cxx,fortran] oneapi@2025.1.1
==> Error: failed to concretize `mapl@2.50 %c,cxx,fortran=oneapi@2025.1.1` for the following reasons:
     1. Cannot satisfy 'mapl@2.50'
        required because mapl@2.50 %c,cxx,fortran=oneapi@2025.1.1 requested explicitly
     2. Cannot satisfy 'mapl@2.50'
```

And it does seem to target just ifx because this "works" (i.e., doesn't fail due to conflict, not sure if it would actually work!):
```console
$ spack spec mapl@2.50 %[virtuals=c,cxx] oneapi@2025.1.1 %[virtuals=fortran] gcc@14.2.0
 -   mapl@2.50.3~debug+extdata2g~f2py+fargparse~ipo+pflogger+pfunit+shared+zstd build_system=cmake build_type=Release generator=make arch=linux-ubuntu24.04-icelake %fortran=gcc@14.2.0 %c,cxx=oneapi@2025.1.1
 -       ^cmake@3.31.8~doc+ncurses+ownlibs~qtgui build_system=generic build_type=Release arch=linux-ubuntu24.04-icelake %c,cxx=oneapi@2025.1.1
```

NOTE: I'm keeping this draft until I can assure myself that MAPL 2.59 will build with oneAPI 2025.1.1. The build is currently ongoing.

mathomp4:
Undrafting as `spack install mapl@2.59 %[virtuals=c,cxx,fortran] oneapi@2025.1.1` built for me.

climbfuji (approved):

climbfuji (approved):
Thanks @mathomp4

mathomp4:
@climbfuji Should I try and encode the "yafyaml doesn't work with ifx 2025.2" in this and the `yafyaml` module? Such that: ifx 2025.2 + `mapl +pflogger` is not allowed?

climbfuji:
> @climbfuji Should I try and encode the "yafyaml doesn't work with ifx 2025.2" in this and the `yafyaml` module? Such that: ifx 2025.2 + `mapl +pflogger` is not allowed?

Does it make sense to do this in this PR, or as a separate/follow-up?

AlexanderRichert-NOAA:
> @climbfuji Should I try and encode the "yafyaml doesn't work with ifx 2025.2" in this and the `yafyaml` module? Such that: ifx 2025.2 + `mapl +pflogger` is not allowed?

Can that be encapsulated by modifying the yafyaml recipe to have a conflict (which would then automatically force mapl to go disable that variant when using ifx 2025.2)?

mathomp4:
> > @climbfuji Should I try and encode the "yafyaml doesn't work with ifx 2025.2" in this and the `yafyaml` module? Such that: ifx 2025.2 + `mapl +pflogger` is not allowed?
> 
> Can that be encapsulated by modifying the yafyaml recipe to have a conflict (which would then automatically force mapl to go disable that variant when using ifx 2025.2)?

True. I'll do the `yafyaml` one first as that is explicit. That should fail the mapl.

mathomp4:
Oh:

https://github.com/spack/spack-packages/pull/1124

I already did it.

PR #1128, opened 2025-08-19
Deprecation removals: Singularity

adamjstewart:


alalazo:
@hppritcha Can you check `openmpi`, and suggest alternatives if the option removed might still be relevant?

becker33:
I think `singularityce` should be a drop-in replacement but I'm not certain, would be good if he could chime in.

hppritcha:
we don't support singularity in Open MPI so i'm fine with this change.

alalazo:
@spackbot run pipeline :sob:

alalazo (dismissed):
We need to update a couple of pipelines:
Screenshot from 2025-08-20 07-10-00

becker33 (approved):

adamjstewart:
Needs fix in #1140

becker33:
The failure here is a pod failure -- I'll rerun it

becker33:
@spackbot rerun pipeline

PR #1137, opened 2025-08-20
Update omega h

jacobmerson:
@cwsmith @Sichao25 if we have documentation of what kokkos versions work with what versions of omega-h we should be more explicit.


cwsmith (changes requested):
The minimum kokkos version with omegah `master` @ [c62cc6d](https://github.com/SCOREC/omega_h/commit/c62cc6d51cbc82a13b073d9ad471779cd46e8820) and scorec-10.9.0 is 4.0.00 according to my testing (using GCC 12 and CUDA 12).  More details are here: https://github.com/SCOREC/omega_h/issues/157.

jacobmerson:
Thanks. I will update the pull request with the minimum.

jacobmerson:
Can't replicate the issue again. Closing for now.

PR #1138, opened 2025-08-20
Swap "if ... in self.spec" with "self.spec.satisfies(...)" in ICON package

stelliom:
There was still a line in the ICON package using the syntax `if ... in self.spec` instead of `self.spec.satisfies(...)`.

Judging from old commits, there was quite an effort to change all `if ... in self.spec` to `self.spec.satisfies(...)` some months ago, so I thought it would be good to change [line 219](https://github.com/spack/spack-packages/blob/develop/repos/spack_repo/builtin/packages/icon/package.py#L219) in *repos/spack_repo/builtin/packages/icon/package.py* accordingly.

### DIFF
--- a/repos/spack_repo/builtin/packages/icon/package.py
+++ b/repos/spack_repo/builtin/packages/icon/package.py
@@ -216,7 +216,7 @@ class Icon(AutotoolsPackage):
         else:
             args.append("--disable-gpu")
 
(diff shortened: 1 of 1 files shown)


stelliom:
Hi @alalazo,

As suggested, I opened a PR for my change. Not sure if you can review this, otherwise let me know who I can contact.

Cheers,
Mikael

alalazo (approved):

PR #1139, opened 2025-08-20
otf2: Lift requirement on Python <= 3.11 for newer versions

pearzt:
If I'm not mistaken, `otf2` no longer depends on Python's `imp` module since version 3.1. Maybe @wrwilliams can confirm?

### DIFF
--- a/repos/spack_repo/builtin/packages/otf2/package.py
+++ b/repos/spack_repo/builtin/packages/otf2/package.py
@@ -74,7 +74,8 @@ class Otf2(AutotoolsPackage):
     extends("python")
 
     # `imp` module required
-    depends_on("python@:3.11", type=("build", "run"))
+    depends_on("python@3", type=("build", "run"), when="@3.1:")
+    depends_on("python@:3.11", type=("build", "run"), when="@:3.0")
 
     with when("@2.2 %cce"):
         depends_on("autoconf", type="build")


wrwilliams:
> If I'm not mistaken, `otf2` no longer depends on Python's `imp` module since version 3.1. Maybe @wrwilliams can confirm?

Correct. But 3.1 does still want a Python interpreter at configure/build time if it is to build the python bindings. So a bit more surgery is required to enforce (build-time) python 3.x @ OTF2 3.1 and python <= 3.11 at older versions.

I would not be sad to deprecate older OTF2s so that we're not dealing with weird EOLed python versions as requirements.

pearzt:
Ah right, I didn't mean to remove the dependency on Python entirely, although just having the `extend` clause would probably be fine in most cases, right?

> So a bit more surgery is required to enforce (build-time) python 3.x @ OTF2 3.1 and python <= 3.11 at older versions.

This should now be reflected, or am I missing something?

> I would not be sad to deprecate older OTF2s so that we're not dealing with weird EOLed python versions as requirements.

No disagreement from my side.

wrwilliams:
> Ah right, I didn't mean to remove the dependency on Python entirely, although just having the `extend` clause would probably be fine in most cases, right?
> 
...I don't actually know what `extend` does internally. Probably?

> > So a bit more surgery is required to enforce (build-time) python 3.x @ OTF2 3.1 and python <= 3.11 at older versions.
> 
> This should now be reflected, or am I missing something?
> 
`depends_on("python@3", type=("build", "run"), when="@3.1:")`

is safer IME. Overlapping version ranges in `depends_on` have caused me no small amount of grief.

I think modulo that point, this is ready for a global maintainer to take a look at though.

pearzt:
@wspear @eugeneswalker @khuck @sameershende (maintainers of TAU) Could you have a look the CI failure here? It seems to me that the relaxed Python version requirement uncovers a build problem with TAU that is unrelated to `otf2`.

pearzt:
@wspear @eugeneswalker @khuck @sameershende Friendly ping:)

wspear:
Thanks for the heads-up. I see the problem. The spack-side PR should be pretty trivial and can probably be in later today. I just need to figure out if we're adjusting anything on the TAU side.

wspear:
This should do the trick: https://github.com/spack/spack-packages/pull/1486

pearzt:
@wspear Thanks for taking care of this! I have rebased this PR, so let's see if the CI is happy now

pearzt:
I think this is ready to merge. @alecbcs Maybe you can have a look and merge this?

alecbcs (approved):
Looks good to me. Thanks @pearzt!

PR #1143, opened 2025-08-20
poppler: add v25.08.0

rbberger:


cessenat (changes requested):
I get an error on Ubuntu 20 compiler gnu 9.4:
```
In file included from /tmp/cessenat/spack-stage/spack-stage-poppler-25.08.0-drng475e52ilymt65z7tuzhrjjp3w3sp/spack-src/fofi/FoFiBase.cc:35:
/tmp/cessenat/spack-stage/spack-stage-poppler-25.08.0-drng475e52ilymt65z7tuzhrjjp3w3sp/spack-src/fofi/FoFiBase.h:33:10: fatal error: span: No such file or directory
   33 | #include <span>
      |          ^~~~~~
compilation terminated.
```
So I suggest you add somethink like:
```
    conflicts("%gcc@:9.9.99", when="@25.08.0:",
              msg="poppler@25.08.0: requires at least gcc 10")
```
because following https://stackoverflow.com/questions/76453853/unable-to-compile-c20-code-using-g-9-3-0

> g++ 9.3.0 does not support span, g++ 10 or above supports span.


see https://en.cppreference.com/w/cpp/compiler_support/20.

Of course, the are other C++ compilers, it would be nice to set a conflict for those as well (clang, oneapi...).

cessenat (approved):
Thanks for the conflict warning. Indeed we ought to indicate poppler needs C++ std20 rather than gcc 10. I do not know if there is something in spack that would bring up the compiler C++ standard for every compiler - instead of indicating a conflict for every compiler on earth...

kwryankrattiger (approved):

PR #241, opened 2025-06-15
pmix: add v5.0.[6-8], v6.0.0; prrte: add v[2-4]; opempi ~internal-pmix

wdconinc:
This PR adds the most recent `pmix` versions in the v5 family, and the first of the v6 family.

[Release notes](https://github.com/openpmix/openpmix/releases/tag/v6.0.0) and [diff](https://github.com/openpmix/openpmix/compare/v5.0.8...v6.0.0) indicate that there are no major changes required in `configure.ac`, but "PRRTE < v4.0 is not compatible with PMIx >= v6.0" so this may break some things in dependents (i.e. `flux-pmix`, `mpich`, `mvapich`, `mvapich2`, `mvapich2-gdr`, `mvapich2x`, `openmpi`, `prrte`, `slurm`). From those, it turns out only `openmpi` uses `prrte` and should depend on it.

Test build of `openmpi` with external (new) `pmix` and `prrte`:
```
[+]  openmpi@5.0.7+atomics~cuda~debug~gpfs~internal-hwloc~internal-libevent~internal-pmix~ipv6~java~lustre~memchecker~openshmem~romio+rsh~static~two_level_namespace+vt+wrapper-rpath build_system=autotools fabrics:=none patches:=38529b5 romio-filesystem:=none schedulers:=none arch=linux-ubuntu25.04-skylake %c,cxx,fortran=gcc@14.2.0
[+]      ^autoconf@2.72 build_system=autotools arch=linux-ubuntu25.04-skylake 
[+]          ^m4@1.4.20+sigsegv build_system=autotools arch=linux-ubuntu25.04-skylake %c,cxx=gcc@14.2.0
[+]              ^diffutils@3.12 build_system=autotools arch=linux-ubuntu25.04-skylake %c=gcc@14.2.0
[+]              ^libsigsegv@2.15 build_system=autotools arch=linux-ubuntu25.04-skylake %c=gcc@14.2.0
[+]      ^automake@1.18 build_system=autotools arch=linux-ubuntu25.04-skylake %c=gcc@14.2.0
[+]      ^compiler-wrapper@1.0 build_system=generic arch=linux-ubuntu25.04-skylake 
[e]      ^gcc@14.2.0~binutils+bootstrap~graphite~mold~nvptx~piclibs~profiled~strip build_system=autotools build_type=RelWithDebInfo languages:='c,c++,fortran' arch=linux-ubuntu25.04-skylake 
[...]

PR #1145, opened 2025-08-20
openmpi: fix schedulers= variant propagation to external prrte

### LINKED ISSUE #670: Add option to build with external prrte for openmpi
### Summary

The prrte commit in 3rdParty/prrte is obsolete and bugged (see https://github.com/open-mpi/ompi/issues/13337)
It would be great to have an option (like there is for pmix) to build with external prrte.
@hppritcha @naughtont3 

### Rationale

Openmpi is now bound to 3rdParty prrte, which is bugged as to version 5.0.8.
prrte v3 fixes the bug but there's no easy way to pick an external prrte software now.

### Description

add `internal-prrte` variant and have the same mechanism that is now implemented for `internal-pmix` to use an external prrte implementation.

### Additional information

_No response_

### General information

### LINKED ISSUE #1144: openmpi: schedulers broken with external prrte
Building `openmpi@5.0.8 schedulers=tm` succeeds, but ultimately can't run jobs through the PBS scheduler. Deep diving it shows that even though openmpi was configured with `--with-tm=/opt/pbs` as per my system pbs listed as an external, there are no `mca:plm:tm` components reported by `prte_info`  and `prte_info` doesn't show `--with-tm` in how it was configured.

Checking the spack configure log confirms this omission.

The changes in https://github.com/spack/spack-packages/pull/241 to enforce the external prrte somehow aren't working with `schedulers=tm` (and I assume slurm).

Both pmix (https://docs.openpmix.org/en/latest/installing-pmix/configure-cli-options/runtime.html) and prrte ( https://github.com/openpmix/prrte/blob/master/config/prte_check_tm.m4) seem to expect `--with-tm` to be passed down to them (and slurm). This does not seem to occur and thus no support.

I have openmpi 5.0.7 built prior to https://github.com/spack/spack-packages/pull/241 which works and it shows the `--with-tm` has properly propagated down to `prte_info`. In this case, the at-the-time default behaviour of using an internal prrte was used.

/cc @wdconinc

Chrismarsh:
Fixes: https://github.com/spack/spack-packages/issues/1144
Closes: https://github.com/spack/spack-packages/issues/670

As described in https://github.com/spack/spack-packages/issues/1144, the changes in https://github.com/spack/spack-packages/pull/241 around `~internal-pmix` does not correctly propagate scheduler selection down to `pmix` or `prrte`. As a result, running mpi jobs would fail (for me) after trying to fall back to ssh (which may mask this issue).

With these changes:
```
prte_info | grep tm
                          '--with-tm=/opt/pbs/'
                 MCA ess: tm (MCA v2.1.0, API v3.0.0, Component v4.0.0)
                 MCA plm: tm (MCA v2.1.0, API v2.0.0, Component v4.0.0)
```

It is not 100% clear to me that `--with-alps` is the correct way to [configure `prrte`](https://github.com/search?q=repo%3Aopenpmix%2Fprrte%20--with-alps&type=code). 
I am hoping @rhc54 might have a moment to lend his expertise.

**maintainers:**
@wdconinc 
`openmpi`: @naughtont3 @hppritcha

### DIFF
--- a/repos/spack_repo/builtin/packages/openmpi/package.py
+++ b/repos/spack_repo/builtin/packages/openmpi/package.py
@@ -621,7 +621,7 @@ class Openmpi(AutotoolsPackage, CudaPackage, ROCmPackage):
     )
     # Variants to use internal packages
     variant("internal-hwloc", default=False, description="Use internal hwloc")
-    variant("internal-pmix", default=False, description="Use internal pmix")
+    variant("internal-pmix", default=False, description="Use internal pmix and prrte")
     variant("internal-libevent", default=False, description="Use internal libevent")
     variant("openshmem", default=False, description="Enable building OpenSHMEM")
     variant("debug", default=False, description="Make debug build", when="build_system=autotools")
@@ -737,6 +737,11 @@ with '-Wl,-commons,use_dylibs' and without
         # https://github.com/open-mpi/ompi/issues/13275#issuecomment-2907903468
         depends_on("prrte")
 
+        for scheduler in [s for s in SCHEDULERS if s not in ("none", "auto", "loadleveler")]:
+            depends_on(f"prrte schedulers={scheduler}", when=f"schedulers={scheduler}")
+            depends_on(f"pmix schedulers={scheduler}", when=f"schedulers={scheduler}")
+
+
     # Libevent is required when *vendored* PMIx is used
     depends_on("libevent@2:", when="~internal-libevent")
 
@@ -776,6 +781,12 @@ with '-Wl,-commons,use_dylibs' and without
         "schedulers=loadleveler",
         when="@3:",
         msg="The loadleveler scheduler is not supported with openmpi(>=3).",
+    )
+
+    conflicts(
+        "schedulers=auto",
+        when="~internal-pmix",
+        msg="External pmix and prrte requires specifying schedulers explicitly.",
     )
 
     # According to this comment on github:
--- a/repos/spack_repo/builtin/packages/openmpi/package.py
+++ b/repos/spack_repo/builtin/packages/openmpi/package.py
@@ -737,7 +737,7 @@ with '-Wl,-commons,use_dylibs' and without
         # https://github.com/open-mpi/ompi/issues/13275#issuecomment-2907903468
         depends_on("prrte")
 
-        for scheduler in [s for s in SCHEDULERS if s not in ("none", "auto", "loadleveler")]:
+        for scheduler in [s for s in SCHEDULERS if s not in ("loadleveler")] + ["none"]:
             depends_on(f"prrte schedulers={scheduler}", when=f"schedulers={scheduler}")
             depends_on(f"pmix schedulers={scheduler}", when=f"schedulers={scheduler}")
 
@@ -785,7 +785,7 @@ with '-Wl,-commons,use_dylibs' and without
     conflicts(
         "schedulers=auto",
         when="~internal-pmix",
-        msg="External pmix and prrte requires specifying schedulers explicitly.",
+        msg="External pmix and prrte requires specifying schedulers explicitly (including 'none').",
     )
 
     # According to this comment on github:
--- a/repos/spack_repo/builtin/packages/openmpi/package.py
+++ b/repos/spack_repo/builtin/packages/openmpi/package.py
@@ -733,13 +733,17 @@ with '-Wl,-commons,use_dylibs' and without
         # See https://www.mail-archive.com/announce@lists.open-mpi.org//msg00158.html
         depends_on("pmix@:4.2.2", when="@:4.1.5")
 
-        # When an external PMIx is used, also an external PRRTE should be used
-        # https://github.com/open-mpi/ompi/issues/13275#issuecomment-2907903468
-        depends_on("prrte")
+        # @:4 does not depend on prrte and used orte
+        with when("@5"):
 
-        for scheduler in [s for s in SCHEDULERS if s not in ("loadleveler")] + ["none"]:
-            depends_on(f"prrte schedulers={scheduler}", when=f"schedulers={scheduler}")
-            depends_on(f"pmix schedulers={scheduler}", when=f"schedulers={scheduler}")
+            # When an external PMIx is used, also an external PRRTE should be used
+            # https://github.com/open-mpi/ompi/issues/13275#issuecomment-2907903468
+            depends_on("prrte")
+
+            # only prrte knows about schedulers
+            # https://github.com/spack/spack-packages/pull/1145#issuecomment-3208378366
+            for scheduler in [s for s in SCHEDULERS if s not in ("loadleveler")] + ["none"]:
+                depends_on(f"prrte schedulers={scheduler}", when=f"schedulers={scheduler}")
 
     # Libevent is required when *vendored* PMIx is used
     depends_on("libevent@2:", when="~internal-libevent")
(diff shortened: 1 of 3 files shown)


Chrismarsh:
@spackbot fix style

repos/spack_repo/builtin/packages/openmpi/package.py
```diff
@@ -737,6 +737,10 @@ def patch(self):
         # https://github.com/open-mpi/ompi/issues/13275#issuecomment-2907903468
         depends_on("prrte")
 
+        for scheduler in [s for s in SCHEDULERS if s not in ("none", "auto", "loadleveler")]:
```
wdconinc: I think `none` and `auto` are only added at the variant level, but aren't in `SCHEDULERS` itself.
```suggestion
        for scheduler in [s for s in SCHEDULERS if s not in ("loadleveler")]:
```

repos/spack_repo/builtin/packages/pmix/package.py
```diff
@@ -200,6 +200,17 @@ class Pmix(AutotoolsPackage):
     depends_on("py-setuptools", when="+python")
     depends_on("munge", when="+munge")
 
+    SCHEDULERS = ("alps", "lsf", "tm", "slurm", "sge")
+
+    variant(
+        "schedulers",
+        values=disjoint_sets(SCHEDULERS),
+        description="List of schedulers for which support is enabled",
+    )
```
wdconinc: Is there a 'cheapest' default here that we can specify explicitly? For `openmpi ~internal-pmix scheduler=none` this will now have an undefined default leading to potentially some unexpected concretizations.

Same comment applies to `prrte` as well.

Chrismarsh: afaik: only `openmpi` depends on `prrte` but a few others depend on `pmix`:
```
ag 'depends_on\(\"prrte\"'
repos/spack_repo/builtin/packages/openmpi/package.py
738:        depends_on("prrte")

$
❯ ag 'depends_on\(\"pmix\"'
repos/spack_repo/builtin/packages/padicotm/package.py
77:    depends_on("pmix", when="+pmix")

repos/spack_repo/builtin/packages/mpich/package.py
336:    depends_on("pmix", when="pmi=pmix")

repos/spack_repo/builtin/packages/nmad/package.py
78:    depends_on("pmix", when="+pmix")

repos/spack_repo/builtin/packages/mvapich2/package.py
151:    depends_on("pmix", when="pmi_version=pmix")

repos/spack_repo/builtin/packages/prrte/package.py
47:    depends_on("pmix")

repos/spack_repo/builtin/packages/mvapich/package.py
118:    depends_on("pmix", when="pmi_version=pmix")
```

so I think any scheduler default in `pmix` (other than none?) has the potential to disrupt any assumptions of those packages. I think adding back the `'none'` variant will resolve your concern. Whether or not it produces an openmpi that behaves in a useful manner, I am not 100% sure (I think it does).

Chrismarsh: I have added `none` back and made it's propagation default

Chrismarsh:
@spackbot fix style

repos/spack_repo/builtin/packages/openmpi/package.py
Chrismarsh: fixed

repos/spack_repo/builtin/packages/pmix/package.py
```diff
@@ -271,4 +282,23 @@ def configure_args(self):
         if spec.satisfies("@:2.1.0 target=aarch64:"):
             config_args.append("--without-tests-examples")
 
+        # schedulers
+        # listed in https://github.com/openpmix/prrte/tree/master/config
+        # in the prte_check_X.m4 files
+        if spec.satisfies("schedulers=lsf"):
+            config_args.append(f"--with-lsf={self.spec['lsf'].prefix}")
+            config_args.append("--with-lsf-libdir={0}".format(spec["lsf"].libs.directories[0]))
```
wdconinc:
```suggestion
            config_args.append(f"--with-lsf-libdir={spec['lsf'].libs.directories[0]}")
```

repos/spack_repo/builtin/packages/prrte/package.py
```diff
@@ -90,4 +102,22 @@ def configure_args(self):
...
         config_args.append("--with-pmix={0}".format(spec["pmix"].prefix))
 
+        # schedulers
+        # https://github.com/openpmix/openpmix/blob/master/docs/installing-pmix/configure-cli-options/runtime.rst
+        if spec.satisfies("schedulers=alps"):
+            config_args.append("--with-alps")
+
+        if spec.satisfies("schedulers=lsf"):
+            config_args.append(f"--with-lsf={self.spec['lsf'].prefix}")
+            config_args.append("--with-lsf-libdir={0}".format(spec["lsf"].libs.directories[0]))
```
wdconinc:
```suggestion
            config_args.append(f"--with-lsf-libdir={spec['lsf'].libs.directories[0]}")
```

wdconinc (commented):
Minor suggestions for consistent f-string usage. (or is this one of those cases where it doesn't work because we rely on an implicit conversion?)

rhc54:
I think things have gotten a little confused here (or maybe it's just my confusion) so let me try to clarify. The "scheduler" pieces of PMIx were never completed nor active, so you can just leave those completely out of your recipes. I have removed the relevant code pieces from PMIx so they will no longer appear anywhere, even by accident. I actually never intended them to be in releases, but they apparently slipped through until the more recent releases.

PRRTE, on the other hand, _requires_ those bits to be there in order to properly operate under a given scheduler. It sounds like the issue here was that the recipe wasn't propagating a specification to an _external_ PRRTE. I cannot speak to this PR in terms of fixing that as I don't understand Spack. However, I can say that PRRTE will, in the absence of a specific configuration directive, properly build support for an environment _if_ it sees the required headers/libraries present during configure. So you don't really need to include all the `--with-foo` directives so long as you include the "foo" dependency in a standard location in the build environment. On the other hand, it doesn't hurt anything to do so, and you might require some of it to handle your dependency locations (e.g., in the case where things are not in a standard location and/or the libdir isn't in an expected location).

HTH
Ralph

wdconinc:
> So you don't really need to include all the `--with-foo` directives so long as you include the "foo" dependency in a standard location in the build environment.

Thanks. This may indeed be specific to spack:
- we attempt to avoid depending on "system" installs because that adds functionality that is not manifest in the spack environment,
- spack installs packages on unique hashed directory prefixes, not in standard paths like /use/local (this allows multiple versions of the same package to be installed on the same system),
- we can rely on "automatic" finding of dependencies, but typically this requires pkgconfig for autotools packages (or some config helper scripts).

So, in summary, I think we probably do have to specify those options explicitly (and with paths in some cases).

rhc54:
> spack installs packages on unique hashed directory prefixes, not in standard paths like /use/local (this allows multiple versions of the same package to be installed on the same system),

Good!! I've been advocating that to other distros for years - with zero success

> typically this requires pkgconfig for autotools package

If it helps, both PMIx and PRRTE install pkgconfig files

> we probably do have to specify those options explicitly (and with paths in some cases).

Understood. Just trying to explain what is in the code and how it works.

repos/spack_repo/builtin/packages/prrte/package.py
Chrismarsh: I just copy-pasted from OpenMPI's to keep it consistent. I am not 100% sure what you mean about the implicit conversion, but the fstring quote reuse I ran into isn't an issue here.

hppritcha (approved):

alalazo (approved):

PR #1151, opened 2025-08-21
add mumps4py package to spack

williampiat3:
[Python wrapper for the MUMPS solvers](https://github.com/imadki/mumps4py), allow using all solver types with a friendly python interface


fbordeu:
This looks good to me. Nice work

williampiat3:
@spackbot request review

williampiat3:
@spackbot hello!

williampiat3:
@spackbot help

williampiat3:
@spackbot maintainers

repos/spack_repo/builtin/packages/py_mumps4py/package.py
```diff
@@ -0,0 +1,63 @@
...
+        args = [
+            "--inplace",
+            "MUMPS_INC={0}".format(mumps.prefix.include),
+            "MUMPS_LIB={0}".format(mumps.prefix.lib),
+        ]
+        return args
+
+    def setup_build_environment(self, env):
+        # Required by mumps4py to specify which MUMPS solvers to use
+        env.set("MUMPS_SOLVERS", "dmumps,cmumps,zmumps,smumps")
```
becker33, line 45: Does this list not depend at all on how mumps itself was built?

repos/spack_repo/builtin/packages/py_mumps4py/package.py
```diff
@@ -0,0 +1,63 @@
...
+    depends_on("py-numpy", type=("build", "run"))
+    depends_on("py-cython", type="build")
+    depends_on("py-mpi4py", type=("build", "run"))
+
+    # Optional/test deps
+    depends_on("py-pytest", type="test")
+
+    # External solver
+    depends_on("mumps")
+    depends_on("cmake", type="build")
```
becker33: This can be cleaned up using default arguments, and I think the structure makes more sense that way

```suggestion
    # build dependencies
    with default_args(type="build"):
        depends_on("cmake")
        depends_on("py-setuptools")
        depends_on("py-cython")

    # Python dependencies
    with default_args(type=("build", "run")):
        depends_on("py-numpy")
        depends_on("py-mpi4py")

    # Optional/test deps
    depends_on("py-pytest", type="test")

    # External solver
    depends_on("mumps")
```
Notes:
1. Removed dependency on `py-pip`, as that is implied by inheriting from `PythonPackage`
2. Removed `python` dependency because a python dependency is implied by inheriting from `PythonPackage`, and Spack has fully removed all python versions that are too old for this package.

repos/spack_repo/builtin/packages/py_mumps4py/package.py
williampiat3: By default, mumps builds all 4 solvers, this env variable allows to build the python bindings for each mumps solver. This usually allow users to adapt the build for custom install, but through spack we want to install all solver bindings.

becker33: I think there needs to be a query here to check how `mumps` was built, and set the solvers based on how mumps was built.

I'm not a mumps expert, but something like this is what I would expect:

```suggestion
    def setup_build_environment(self, env):
        solvers = []
        mumps = self.spec["mumps"]
        if mumps.satisfies("+float"):
            solvers.append("smumps")
        if mumps.satisfies("+double"):
            solvers.append("dmumps")
        if mumps.satisfies("+float+complex"):
            solvers.append("cmumps")
        if mumps.satisfies("+double+complex"):
            solvers.append("zmumps")

        # Required by mumps4py to specify which MUMPS solvers to use
        env.set("MUMPS_SOLVERS", ",".join(solvers))
```
@jcortial-safran you're listed as the maintainer for `mumps`, do you think that's the right way to figure out which solvers are available?

becker33 (changes requested):

repos/spack_repo/builtin/packages/py_mumps4py/package.py
jcortial-safran: IMHO we should not query mumps to know what solvers are available. Rather, mumps4py should mandate which solvers must be available in MUMPS.

By default, all 4 solvers are built in the [MUMPS spack recipe](https://github.com/spack/spack-packages/blob/develop/repos/spack_repo/builtin/packages/mumps/package.py) so `requires("mumps")` is already enough to ensure that the prerequisite for `env.set("MUMPS_SOLVERS", "dmumps,cmumps,zmumps,smumps")` is fulfilled. And if you really want to be sure, you can explicitly specify `requires("mumps+float+complex+double")`.

If you *really* want to allow mumps4py to work with a subset of the MUMPS solvers (an approach that I would not recommend btw), you could  for instance mirror the `float`, `double` and `complex` variants of MUMPS in the mumps4py recipe, and add the corresponding prerequisites, e.g. `requires("mumps+double", when="+double")` and so on.

williampiat3: @becker33, do you think I should complexify the recipe to allow working with subsets of MUMPS solver?

becker33: For now I think it's fine to just say `depends_on("mumps+float+complex+double")` and keep the solver list simple.

becker33 (approved):

PR #1152, opened 2025-08-21
Always disable xpu to avoid torch build from auto-finding it on system

scheibelp:
A build of `py-torch` failed for me because it auto-determined `USE_XPU=ON` based on the presence of system artifacts, then failed.

Right now there are no variants/dependencies for Intel GPU support. Until those are added, it is safest to turn it off.


kwryankrattiger:
Looking at why CI status not updating.

adamjstewart (approved):
I'm fine with disabling or adding a variant. Didn't even know it was that easy to enable. We don't yet have any Intel ML pipelines, only CUDA and ROCm.

PR #1156, opened 2025-08-21
Added mariadb 11.8.3 and 12.1.1

emwjacobson:
Added mariadb versions 11.8.3 (latest 11.x version) and 12.1.1 (latest 12.x).

I also updated the URL to not be a (seemingly) arbitrary mirror and to use https.

Update was motivated by the inability to compile mariadb 11.3.2 using gcc@15.2.0

### DIFF
--- a/repos/spack_repo/builtin/packages/mariadb/package.py
+++ b/repos/spack_repo/builtin/packages/mariadb/package.py
@@ -73,6 +73,8 @@ class Mariadb(CMakePackage):
     depends_on("openssl")
     depends_on("openssl@:1.0", when="@:10.1")
     depends_on("krb5")
+    depends_on("snappy", when="@11.8.3: platform=darwin")
+    depends_on("pcre2", when="@11.8.3: platform=darwin")
 
     conflicts("%gcc@9.1.0:", when="@:5.5")
     conflicts("%gcc@13:", when="@:10.8.7")  # https://github.com/spack/spack/issues/41377
--- a/repos/spack_repo/builtin/packages/mariadb/package.py
+++ b/repos/spack_repo/builtin/packages/mariadb/package.py
@@ -73,8 +73,9 @@ class Mariadb(CMakePackage):
     depends_on("openssl")
     depends_on("openssl@:1.0", when="@:10.1")
     depends_on("krb5")
-    depends_on("snappy", when="@11.8.3: platform=darwin")
-    depends_on("pcre2", when="@11.8.3: platform=darwin")
+    depends_on("snappy", when="@11.8.3:")
+    depends_on("pcre2", when="@11.8.3:")
+    depends_on("fmt", when="@11.8.3:")
 
     conflicts("%gcc@9.1.0:", when="@:5.5")
     conflicts("%gcc@13:", when="@:10.8.7")  # https://github.com/spack/spack/issues/41377
--- a/repos/spack_repo/builtin/packages/mariadb/package.py
+++ b/repos/spack_repo/builtin/packages/mariadb/package.py
@@ -20,10 +20,12 @@ class Mariadb(CMakePackage):
     """
 
     homepage = "https://mariadb.org/about/"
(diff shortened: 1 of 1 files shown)


rbberger:
On MacOS I ran into build errors due to missing dependencies: `snappy` and `pcre2`.

emwjacobson:
> On MacOS I ran into build errors due to missing dependencies: `snappy` and `pcre2`.

With both 11.8.3 and 12.1.1, or just one of the two?

I don't have a Mac to test with, but maybe adding the following resolve the errors?
```
depends_on("snappy")
depends_on("pcre2")
```

If it works, then maybe it's as easy as adding them as dependencies?

rbberger:
@emwjacobson yes for both versions. And yes adding the two libraries does fix the build. As far as I can tell it has to do with the build of the RocksDB storage plugin. Overall this recipe is pretty simple and doesn't expose that many variants. If there is no appetite to explore this further, the easiest fix is adding those two libraries as dependencies.

emwjacobson:
Just pushed a commit that will add snappy and pcre2 as requirements if building on a Mac (darwin). Let me know if this helps with things.

rbberger:
Looks like sqlite actually downloads PCRE2 and fmtlib and compiles it in the Linux case. Something we want to avoid. 

```
-- Downloading...
   dst='/tmp/rberger/spack-stage/spack-stage-mariadb-12.1.1-7v5bbzap4pnvbh3qqex5oi4f73xrxigs/spack-build-7v5bbza/extra/pcre2/src/pcre2-10.45.zip'
   timeout='none' 
   inactivity timeout='none'
-- Using src='https://github.com/PCRE2Project/pcre2/releases/download/pcre2-10.45/pcre2-10.45.zip'
-- Downloading...
   dst='/tmp/rberger/spack-stage/spack-stage-mariadb-12.1.1-7v5bbzap4pnvbh3qqex5oi4f73xrxigs/spack-build-7v5bbza/extra/libfmt/src/fmt-11.1.4.zip'
   timeout='none'
   inactivity timeout='none'
-- Using src='https://github.com/fmtlib/fmt/releases/download/11.1.4/fmt-11.1.4.zip'

```

Something also doesn't add up on the Linux build without `snappy` (it completes though, unlike on my MacOS system):
```
-- Could NOT find LZO (missing: LZO_LIBRARIES LZO_INCLUDE_DIRS)
-- Could NOT find Snappy (missing: SNAPPY_LIBRARIES SNAPPY_INCLUDE_DIRS)

...

The following features have been enabled:

...
 * ROCKSDB, Storage Engine MODULE
 * ROCKSDB_LZ4, LZ4 Compression in the RocksDB storage engine
 * ROCKSDB_BZip2, BZip2 Compression in the RocksDB storage engine
 * ROCKSDB_Snappy, Snappy Compression in the RocksDB storage engine
 * ROCKSDB_ZSTD, ZSTD Compression in the RocksDB storage engine
 * ROCKSDB_ZLIB, zlib Compression in the RocksDB storage engine

...

-- The following OPTIONAL packages have not been found:

 * Java (required version >= 1.6)
   Required for the CONNECT_JDBC feature
 * JNI
   Required for the CONNECT_JDBC feature
 * Judy
   Required for the OQGraph storage engine
 * LZO
 * Snappy
```

So I wouldn't make them conditional on MacOS. It's needed in all cases, but the CMake seems to be more tolerant in the Linux case. I would also add a `fmt` dependency to avoid the downloads during CMake. This makes it better installable in offline environments.

emwjacobson:
Latest commit:
- Added Snappy as a requirement `@11.8.3:`
```console
-- Found Snappy: /path/to/spack/opt/spack/linux-icelake/snappy-1.2.1-hcp4v35iw6hgionfi6vcsxg3pyaplzjc/lib/libsnappy.so
```
- Added pcre2 as a requirement `@11.8.3:`
  - I didn't personally have a problem compiling without this, but adding it as a dependency builds using it:
```console
--   Found libpcre2-8, version 10.44
-- Looking for pcre2_match_8 in pcre2-8
```
- Added fmt as a requirement `@11.8.3:`
  - No more downloading of fmt

rbberger (approved):

PR #1157, opened 2025-08-21
netcdfc: ensure patchs are applied to only appropriate versions post 3.9.3 update

johnwparent:


johnwparent:
@jrood-nrel I've verified these patch boundaries work for me locally, I think for non Windows the second patch I change here was the only unbounded patch applied.

jrood-nrel:
Thanks!

johnwparent:
Looks like we'll need to wait for the Paraview timeouts to be resolved.

repos/spack_repo/builtin/packages/netcdf_c/package.py
```diff
@@ -55,14 +55,14 @@ class NetcdfC(CMakePackage, AutotoolsPackage):
     with when("build_system=cmake"):
         # TODO: document why we need to revert https://github.com/Unidata/netcdf-c/pull/1731
         #  with the following patch:
-        patch("4.8.1-win-hdf5-with-zlib.patch", when="@4.8.1: platform=windows")
+        patch("4.8.1-win-hdf5-with-zlib.patch", when="@4.8.1:4.9.2 platform=windows")
 
         # TODO: https://github.com/Unidata/netcdf-c/pull/2595 contains some of the changes
         # made in this patch but is not sufficent to replace the patch. There is currently
         # no upstream PR (or set of PRs) covering all changes in this path.
         # When #2595 lands, this patch should be updated to include only
```
Chrismarsh, line 63: I've always found this line's comment confusing -- has this been done? are we still waiting for it? Given the bound on this patch, does that mean that 4.9.3+ includes the full set of changes in this patch?

johnwparent: The MPI specific parts of this patch are redundant as of 4.9.3, the hdf5 may still be required but I need to test that further with the MPI behavior and newer versions of hdf5. Since this only effects Windows, lets let this land and I'll figure out what new patch may need to be added. It has since been pointed out to me a few ways this patch can be improved anyway, so I'll likely be upstreaming something that refactors both the MPI and HDF5 handling.

Chrismarsh:
I just tested this again, and the patch applies correctly, and it builds correctly 

```
$ spack spec /cn6pf3uo
[+]  netcdf-c@4.9.3+blosc~byterange~dap~fsync~hdf4~ipo~jna~logging+mpi~nczarr_zip+optimize~parallel-netcdf+pic+shared+szip+zstd build_system=cmake build_type=Release generator=make patches:=1858258 platform=linux os=rhel8 target=icelake %c,cxx=gcc@14.3.0

```

jrood-nrel:
I would love to see this go in.

johnwparent:
@kwryankrattiger this is one of the PRs I was talking about

kwryankrattiger (approved):

PR #1162, opened 2025-08-22
kokkos: remove unnecessary language dependencies

tpadioleau:
Rationale:
- Kokkos: the C dependency is not necessary, the only C file, `core/unit_test/tools/TestCInterface.c`, is not compiled by the cmake project.
- Kokkos_nvcc_wrapper: the recipe does not compile any source file but uses `self.compiler.cxx`.


tpadioleau:
I have a doubt though, does calling `self.compiler.cxx` require to depend on the C++ language ?

repos/spack_repo/builtin/packages/kokkos_nvcc_wrapper/package.py
```diff
@@ -159,10 +159,6 @@ class KokkosNvccWrapper(Package):
         deprecated=True,
     )
 
-    depends_on("c", type="build")  # generated
-    depends_on("cxx", type="build")  # generated
```
dalg24, line 163: do we want to keep the depends on cxx?
we do have a depends on Cuda below

rbberger: you need the `depends_on("cxx", type="build")` for `self.compiler.cxx` to exist.

rbberger (changes requested):
`cxx` dependency is needed `kokkos-nvcc-wrapper`.

repos/spack_repo/builtin/packages/kokkos_nvcc_wrapper/package.py
tpadioleau: I agree, I could test without `depends_on("cxx", type="build")` and it indeed fails. Adding it makes it work again.

tpadioleau: fixed in 88bd7e2f733a95bd680b93f2c0a4911d6a60fcc3

rbberger:
@tpadioleau actually, you do inject a `C` dependency at the very end. See https://gitlab.spack.io/spack/spack-packages/-/jobs/17906906#L1013 and my fix in https://github.com/kokkos/kokkos/pull/8357. So the `depends_on("c", type="build")` has to be guarded with a `when` version check once the above fix is merged.

tpadioleau:
> @tpadioleau actually, you do inject a `C` dependency at the very end. See https://gitlab.spack.io/spack/spack-packages/-/jobs/17906906#L1013 and my fix in [kokkos/kokkos#8357](https://github.com/kokkos/kokkos/pull/8357). So the `depends_on("c", type="build")` has to be guarded with a `when` version check once the above fix is merged.

Good catch, thanks. Would it be worth patching this cmake file ? Last update was 4 years ago.

rbberger:
@tpadioleau now that it is merged, we can use this patch file: https://github.com/kokkos/kokkos/commit/05d4901538251fff7ae6e58c84db670ad326b5c8.patch?full_index=1

tpadioleau:
Back to draft, the CI issue https://gitlab.spack.io/spack/spack-packages/-/jobs/17906906#L1013 should be fixed with the patch but since a few releases, the directory `scripts` is not anymore part of the tarball making the patch fail for recent releases like 4.7.

repos/spack_repo/builtin/packages/kokkos/package.py
```diff
@@ -398,6 +397,8 @@ class Kokkos(CMakePackage, CudaPackage, ROCmPackage):
         sha256="145619e87dbf26b66ea23e76906576e2a854a3b09f2a2dd70363e61419fa6a6e",
         when="@4.2.00",
     )
+    # Remove unnecessary C and C++ languages dependency in scripts/spack_test/CMakeLists.txt
```
rbberger: See above patch for example on how to download and document a patch that was upstreamed. Even if the folder is not part of release tarballs, there is git versions like `develop` that still contain the folder.

rbberger (changes requested):

repos/spack_repo/builtin/packages/kokkos/package.py
tpadioleau: I agree that the `develop` branch contains the directory but we should not need to apply the patch as the fixed was merged right ?

tpadioleau: > See above patch for example on how to download and document a patch that was upstreamed.

see 7686857c5bc27af95cf64f3ef783f3275667171c

rbberger (approved):

PR #1164, opened 2025-08-22
osu-micro-benchmarks: Add cray-mpich GTL support

vanderwb:
This PR adds GTL support to the `osu-micro-benchmarks` package, so that the `cray-mpich` MPI provided by the Cray Programming Environment can be used to build the package with either `+cuda` or `+rocm` support. Without this change, the "gtl" library is not linked into the OSU executables and they fail at runtime.

The code was borrowed and slightly modified from the `lammps` package.

### DIFF
--- a/repos/spack_repo/builtin/packages/osu_micro_benchmarks/package.py
+++ b/repos/spack_repo/builtin/packages/osu_micro_benchmarks/package.py
@@ -59,6 +59,22 @@ class OsuMicroBenchmarks(AutotoolsPackage, CudaPackage, ROCmPackage):
     depends_on("gnuplot", when="+graphing")
     depends_on("imagemagick", when="+graphing")
 
+    def flag_handler(self, name, flags):
+        wrapper_flags = []
+        build_system_flags = []
+
+        if self.spec.satisfies("+cuda") or self.spec.satisfies("+rocm"):
+            if self.spec.satisfies("^[virtuals=mpi] cray-mpich"):
+                gtl_lib = self.spec["cray-mpich"].package.gtl_lib
+                build_system_flags.extend(gtl_lib.get(name) or [])
+            # hipcc is not wrapped, we need to pass the flags via the build
+            # system.
+            build_system_flags.extend(flags)
+        else:
+            wrapper_flags.extend(flags)
+
+        return (wrapper_flags, [], build_system_flags)
+
     def configure_args(self):
         spec = self.spec
         config_args = ["CC=%s" % spec["mpi"].mpicc, "CXX=%s" % spec["mpi"].mpicxx]
--- a/repos/spack_repo/builtin/packages/osu_micro_benchmarks/package.py
+++ b/repos/spack_repo/builtin/packages/osu_micro_benchmarks/package.py
@@ -63,17 +63,20 @@ class OsuMicroBenchmarks(AutotoolsPackage, CudaPackage, ROCmPackage):
         wrapper_flags = []
         build_system_flags = []
 
-        if self.spec.satisfies("+cuda") or self.spec.satisfies("+rocm"):
-            if self.spec.satisfies("^[virtuals=mpi] cray-mpich"):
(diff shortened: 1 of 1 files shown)


etiennemlb:
I have an issue where the configure steps, while it does try to add the gtl, does not know where to find it. it seems that the configure does not read `LDFLAGS=-L/opt/cray/pe/mpich/8.1.30/gtl/lib -Wl,-rpath,/opt/cray/pe/mpich/8.1.30/gtl/lib'` but it does read `'LIBS=-lmpi_gtl_hsa'`.

The configure line is `'...spack-src/configure' 'LDFLAGS=-L/opt/cray/pe/mpich/8.1.30/gtl/lib -Wl,-rpath,/opt/cray/pe/mpich/8.1.30/gtl/lib' 'LIBS=-lmpi_gtl_hsa' '--prefix=.../osu-micro-benchmarks-7.5-ry5s' 'CC=/opt/cray/pe/mpich/8.1.30/ofi/gnu/11.2/bin/mpicc' 'CXX=/opt/cray/pe/mpich/8.1.30/ofi/gnu/11.2/bin/mpicxx' '--enable-rocm' '--with-rocm=/opt/rocm-6.1.2' 'HCC_AMDGPU_TARGET=--amdgpu-target=gfx90a' 'LDFLAGS=-lrt'`

Indeed, looking at the configure logs, the values of LDFLAGS is set to 'LDFLAGS=-lrt'. So the fact that osu micro bench modify the flags in a non "spack standard way", breaks this compilation.

etiennemlb:
You could probably modify it like so:

```python
        wrapper_flags = []
        build_system_flags = []

        if name == "ldlibs":
            if not sys.platform == "darwin":
                # librt not available on darwin (and not required)
                flags.append("-lrt")

        if self.spec.satisfies("+cuda") or self.spec.satisfies("+rocm"):
            if self.spec.satisfies("^[virtuals=mpi] cray-mpich"):
                gtl_lib = self.spec["cray-mpich"].package.gtl_lib
                build_system_flags.extend(gtl_lib.get(name) or [])
            # hipcc is not wrapped, we need to pass the flags via the build
            # system.
            build_system_flags.extend(flags)
        else:
            wrapper_flags.extend(flags)

        return (wrapper_flags, [], build_system_flags)
```

and remove the lrt flags at the bottom of the package.py

etiennemlb:
Now we have: 
```
'/configure' 'LDFLAGS=-L/opt/cray/pe/mpich/8.1.30/gtl/lib -Wl,-rpath,/opt/cray/pe/mpich/8.1.30/gtl/lib' 'LIBS=-lmpi_gtl_hsa -lrt' '--prefix=/linux-zen3/osu-micro-benchmarks-7.5-h5nj' 'CC=/opt/cray/pe/mpich/8.1.30/ofi/gnu/11.2/bin/mpicc' 'CXX=/opt/cray/pe/mpich/8.1.30/ofi/gnu/11.2/bin/mpicxx' '--enable-rocm' '--with-rocm=/opt/rocm-6.1.2' 'HCC_AMDGPU_TARGET=--amdgpu-target=gfx90a'
```
which is correct

MatthewLieber (approved):
lgtm

etiennemlb:
> lgtm

I showed earlier, and as is, it does not build on the Cray system I use.

vanderwb:
Thanks @etiennemlb - let me see why I am getting a different result. I will likely incorporate your suggested change, in any case.

vanderwb:
@etiennemlb - while I still was not able to reproduce the failure you mentioned with the latest CPE, when I tried with older versions (namely `cray-mpich@8.1.29` from CPE 24.03), I did see the linking issue manifest.

I've incorporated your suggestion to move the `-ldl` addition to the flag handler, and I've also made it so that non-GPU flags are still injected via the Spack wrappers even if CUDA or ROCm support are requested, as this is what I suspect users will expect. Let me know if this doesn't resolve the build failure for you.

PR #1165, opened 2025-08-22
Don't retry `script_failure` on build jobs

jjnesbitt:
This will prevent legitimate build failures from being needlessly retried two extra times. Since we see very little "false positives" for these failures lately (network errors, etc.), this should be a big improvement short-term. After this is merged, our goal is to then address any infrastructure related errors that may be classified as `script_failure`, ensuring they are retried in an appropriate manner.

@zackgalbreath @kwryankrattiger Since we're just targeting build failures for the time being, I think this change is all that's required. As far as I can tell, these are the only two places that build job configurations inherit from. However, I could be missing something.


.ci/gitlab/.gitlab-ci.yml
```diff
@@ -175,6 +175,21 @@ default:
 .build:
   extends: [ ".base-job" ]
   stage: build
+  # Retry for everything but "legitimate build failures" (script_failure)
+  retry:
+    max: 2
```
kwryankrattiger: Move this under `.generate-common`, this job template is only used for the triggered jobs.

jjnesbitt: My recent commit removed this retry config and just specified it under `.ci/gitlab/configs/ci.yaml`. Isn't that what we want? I.e. we don't want to retry generate jobs, just build jobs?

kwryankrattiger: we are already retrying generate jobs which can have failures for the same reasons. Putting this in there would probably also be an okay idea.

jjnesbitt:
@spackbot re-run pipeline

zackgalbreath (approved):

kwryankrattiger (approved):

PR #1167, opened 2025-08-22
Remove false fortran dependencies

sethrj:
The 'generated' compiler languages have a number of false positives, especially for Fortran, which is bad because macOS doesn't include a fortran compiler by default, leading to concretization errors like:
```
==> Error: failed to concretize `py-sphinxcontrib-bibtex`, `python@3.6:`, `git`, `geant4@11`, ..., `py-breathe` for the following reasons:
     1. Only external, or concrete, compilers are allowed for the fortran language. You could consider setting `concretizer:unify` to `when_possible` or `false` to allow multiple versions of some packages.
```

Many packages have Fortran as an optional target, or use Fortran as example files, etc. I've manually verified a few packages that definitely do not use Fortran. I also removed the other languages from "spack/package.yaml" since those are obviously fake too...

### DIFF
--- a/repos/spack_repo/builtin/packages/hepmc3/package.py
+++ b/repos/spack_repo/builtin/packages/hepmc3/package.py
@@ -46,6 +46,7 @@ class Hepmc3(CMakePackage):
         description="Install interfaces for some Monte-Carlo Event Gens",
     )
 
+    depends_on("c", type="build", when="@:3.2")
     depends_on("cxx", type="build")
 
     depends_on("cmake@2.8.9:", type="build")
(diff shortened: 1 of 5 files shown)


repos/spack_repo/builtin/packages/hepmc3/package.py
```diff
@@ -46,9 +46,7 @@ class Hepmc3(CMakePackage):
         description="Install interfaces for some Monte-Carlo Event Gens",
     )
 
-    depends_on("c", type="build")  # generated
-    depends_on("cxx", type="build")  # generated
-    depends_on("fortran", type="build")  # generated
+    depends_on("cxx", type="build")
```
kwryankrattiger, line 50: It looks like `@3.1:3.2` require C implicitly for configure.

https://gitlab.cern.ch/hepmc/HepMC3/-/blob/3.2.7/CMakeLists.txt?ref_type=heads#L3

wdconinc:
```suggestion
    depends_on("c", type="build", when="@3.1:3.2")
    depends_on("cxx", type="build")
```

Fortran is only used for `examples/Pythia6Example` (which is not supported in spack; no `+examples` variant).

wdconinc (dismissed):
Approving for `hepmc3` and for `spack` :-)

wdconinc (approved):

sethrj:
@mrguilima This hopefully fixes the spack installation error we were seeing friday.

PR #1168, opened 2025-08-22
[gloo] updating c++ standard

qwertos:

### DIFF
--- a/repos/spack_repo/builtin/packages/gloo/package.py
+++ b/repos/spack_repo/builtin/packages/gloo/package.py
@@ -62,6 +62,13 @@ class Gloo(CMakePackage, CudaPackage):
     depends_on("cmake@2.8.12:", type="build")
     depends_on("libuv", when="platform=windows")
 
+    def patch(self):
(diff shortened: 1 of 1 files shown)


qwertos:
bump

wdconinc:
Note that this fails style checks.

repos/spack_repo/builtin/packages/gloo/package.py
```diff
@@ -62,6 +62,14 @@ class Gloo(CMakePackage, CudaPackage):
     depends_on("cmake@2.8.12:", type="build")
     depends_on("libuv", when="platform=windows")
 
+    def patch(self):
+        if self.compiler.name == "gcc" and self.compiler.version >= Version("14.0.0"):
```
becker33:
```suggestion
        if self.spec.satisfies("%cxx=gcc@14:"):`
```

becker33 (changes requested):
Also please fix the style check

becker33 (approved):

PR #1169, opened 2025-08-22
MFEM: fix `+strumpack` error

white238:
This fixes the following error:

```
==> mfem: Executing phase: 'configure'
==> Error: KeyError: 'FC'

package_repos/fncqgg4/repos/spack_repo/builtin/packages/mfem/package.py:807, in get_make_config_options:
        804            # Parts of STRUMPACK use fortran, so we need to link with the
        805            # fortran library and also the MPI fortran library:
        806            if "~shared" in strumpack:
  >>    807                if os.path.basename(env["FC"]) == "gfortran":
        808                    gfortran = Executable(env["FC"])
        809                    libext = "dylib" if sys.platform == "darwin" else "so"
        810                    libfile = os.path.abspath(
```


white238:
@alecbcs

alecbcs (approved):
Looks good to me. Thanks @white238!

PR #1170, opened 2025-08-22
mcl: update and addition of cimfomfa dependency

emwjacobson:
Was working on updating mcl, prompted due to a problematic import, and found that newer versions depend on [cimfomfa](https://github.com/micans/cimfomfa).

One concern/question about this would be the license on cimfomfa. It has a GPLv2 license in the repo, and I'm not sure if this would qualify for a `GPL-2.0-only` or `GPL-2.0-or-later` license.

In the cimfomfa repo there is a "configure.ac.in" file to use for configuration. This isn't auto-detected by `autoreconf`, but renaming it resolves things. I'm not sure if a `shutil.move` is standard practice, though I found it also used in some other packages. Any recommendations here would be appreciated!

### DIFF
--- a/repos/spack_repo/builtin/packages/cimfomfa/package.py
+++ b/repos/spack_repo/builtin/packages/cimfomfa/package.py
@@ -18,7 +18,7 @@ class Cimfomfa(AutotoolsPackage):
     homepage = "https://github.com/micans/cimfomfa"
     url = "https://github.com/micans/cimfomfa/archive/refs/tags/21-361.tar.gz"
 
-    license("GPL-2.0-only", checked_by="emwjacobson")
+    license("GPL-2.0-or-later", checked_by="emwjacobson")
 
     version("21-361", sha256="e554f7838a16dfee79999b28133abf58dce01ac9a18f99c38c4183805b5b19d4")
 
--- a/repos/spack_repo/builtin/packages/cimfomfa/package.py
+++ b/repos/spack_repo/builtin/packages/cimfomfa/package.py
@@ -2,27 +2,9 @@
 #
 # SPDX-License-Identifier: (Apache-2.0 OR MIT)
 
-# ----------------------------------------------------------------------------
-# If you submit this package back to Spack as a pull request,
-# please first remove this boilerplate and all FIXME comments.
-#
-# This is a template package file for Spack.  We've put "FIXME"
-# next to all the things you'll want to change. Once you've handled
-# them, you can save this file and test your package like this:
-#
-#     spack install cimfomfa
-#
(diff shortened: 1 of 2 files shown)


repos/spack_repo/builtin/packages/cimfomfa/package.py
```diff
@@ -0,0 +1,35 @@
...
+class Cimfomfa(AutotoolsPackage):
+    """This library supports both MCL, a cluster algorithm for graphs, and zoem, a macro/DSL
+    language. It supplies abstractions for memory management, I/O, associative arrays, strings,
+    heaps, and a few other things.
+    """
+
+    homepage = "https://github.com/micans/cimfomfa"
+    url = "https://github.com/micans/cimfomfa/archive/refs/tags/21-361.tar.gz"
+
+    license("GPL-2.0-only", checked_by="emwjacobson")
```
bernhardkaindl: Hi @emwjacobson,

> One concern/question about this would be the license on cimfomfa. It has a GPLv2 license in the repo, and I'm not sure if this would qualify for a GPL-2.0-only or GPL-2.0-or-later license.

The content of https://github.com/micans/cimfomfa/blob/main/LICENSE saying

> These programs are free software; you can redistribute and/or modify them
> under the terms of the GNU General Public License as published by the Free
> Software Foundation; either version 2 of the License, or (at your option)
> any later version.

is a good indication that the author intented to use `GPL-2.0-or-later` for this project. What actually counts is origin of the repository content and the license headers in the files themselfes, for example:

https://github.com/micans/cimfomfa/blob/main/src/trumpet/cff_edit.c

This also contains the "or any later version" clause, so you could expect that all are the same and there should be no other license (hopefully). So this change would apply:

```suggestion
    license("GPL-2.0-or-later", checked_by="emwjacobson")
```

bernhardkaindl (commented):
Hi @emwjacobson,

> In the cimfomfa repo there is a "configure.ac.in" file to use for configuration. This isn't auto-detected by autoreconf, but renaming it resolves things. I'm not sure if a shutil.move is standard practice, though I found it also used in some other packages.
> Any recommendations here would be appreciated

As other packages use it and it is a perfectly valid way to make things compile, that looks fine.

Apart from the change to "or-later" for the license field, I think I'd be fine with approving and merging after this change.

emwjacobson:
Sounds good! Thanks for the assurance as well as the tip for the license change, I appreciate it :)

hainest (approved):
Confirmed SHAs.

hainest:
@spackbot re-run pipeline

PR #1179, opened 2025-08-25
podio: Add the conditional value 23 for cxxstd

tmadlener:
As discussed in https://github.com/spack/spack-packages/pull/1088#pullrequestreview-3136616166 adding `23` as a possible value for `cxxstd` is done in this separate PR. The podio PR that adds CI tests for this is https://github.com/AIDASoft/podio/pull/822.

### DIFF
--- a/repos/spack_repo/builtin/packages/podio/package.py
+++ b/repos/spack_repo/builtin/packages/podio/package.py
@@ -84,6 +84,7 @@ class Podio(CMakePackage):
     depends_on("catch2@3.4:", type=("test"), when="cxxstd=20")
     depends_on("catch2@3.3:", type=("test"), when="@1.2: cxxstd=17")
     depends_on("catch2@3.5:", type=("test"), when="@1.5:")
+    depends_on("catch2@3.5:", type=("test"), when="@1.3: cxxstd=23")
     depends_on("py-graphviz", type=("run"))
     depends_on("py-tabulate", type=("run", "test"))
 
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/podio/package.py
```diff
@@ -109,8 +113,10 @@ class Podio(CMakePackage):
     depends_on("root@6.28.04: +root7", when="+rntuple")
     depends_on("root@6.28:", when="@0.17:")
     depends_on("root@6.32: +root7", when="@1.3: +rntuple")
-    for cxxstd in ("17", "20"):
-        depends_on("root cxxstd={}".format(cxxstd), when="cxxstd={}".format(cxxstd))
+    for cxxstd in _cxxstd_values:
+        depends_on(
+            "root cxxstd={}".format(cxxstd[0].value), when="cxxstd={}".format(cxxstd[0].value)
+        )
```
tmadlener, line 75: I am not entirely sure if there is a more elegant solution than this, but the previous list of hardcoded values seemed more brittle than this.

wdconinc: Not sure it's more elegant, but @sethrj put some clever stuff in `vecgeom` IIRC:
https://github.com/spack/spack-packages/blob/0c164b846058d9c57c9adb6b17ef86ead9f4b8c4/repos/spack_repo/builtin/packages/vecgeom/package.py#L12-L18
https://github.com/spack/spack-packages/blob/0c164b846058d9c57c9adb6b17ef86ead9f4b8c4/repos/spack_repo/builtin/packages/vecgeom/package.py#L141-L144

I don't know if you actually need to propagate the `when`s...

sethrj: I have gotten many catastrophic errors from ROOT trying to build dictionaries with a different standard, so I think that one is basically required. For the other ones, I think most standard library implementations are good enough to correctly link across c++ standards?

P.S. I may have borrowed/adapted the `_std_when` technique from another package, so it's surely a collective effort that deserves credit  ;)

jmcarcell (approved):
Version ranges look OK

tmadlener:
rebased to resolve conflict.

wdconinc (approved):
Looks good to me. Apologies for delay.

PR #1180, opened 2025-08-25
Update arkouda/package.py to inculde recent versions up to v2025.08.20

ajpotts:
## Summary

Adds Arkouda and Py-Arkouda versions **2025.07.03** and **2025.08.20**.
Deprecates **2024.10.02** and **2024.06.21**, which are no longer actively supported.

Arkouda and Py-Arkouda function as a server/client pair and must be updated and tested together, so both packages are updated in this PR.

## Testing

The new versions were validated end-to-end using a unified Spack environment:

```
source /opt/spack-develop/share/spack/setup-env.sh
spack env create ark-env
spack env activate -p ark-env

# Unify dependencies across both roots
spack config add concretizer:unify:true

# Add matching server/client roots
spack add arkouda@2025.08.20
spack add py-arkouda@2025.08.20

# Concretize and verify the solve
spack concretize -f
spack spec -Il
spack install

```
Runtime validation was performed by cloning the Arkouda GitHub repo and running the full unit test suite:
```
python3 -m pytest -c pytest.ini --size=100 --skip_doctest=True
```
## Changes

- **Branch rename**: Updated default branch from `master` → `main` to match the upstream project.
- **Chapel dependency**:  Updated upper bound to 2.4, consistent with Arkouda’s current support for 2025.07.03–2025.08.20.
`depends_on("chapel@2.0:2.4 +hdf5 +zmq",when="@2025.07.03:2025.08.20", type=("build","link","run","test"))`
- **Arrow / LZ4**: Added lz4 explicitly to ensure Parquet I/O support.  I confirmed that building Arrow (and its compression stack, including LZ4) using the CMake build system works correctly with Arkouda’s Parquet I/O path.
```
depends_on("arrow@:19+brotli+bz2+lz4+parquet+snappy+zlib+zstd", type=("build", "link", "run"))
depends_on("lz4 build_system=cmake", type="build")
```
- **Python versions**: Updated to match currently supported versions in the Arkouda project.
```
depends_on("python@3.9:3.12.3", when="@:2025.01.13", type=("build","run"))
depends_on("python@3.9:3.13", when="@2025.07.03:", type=("build","run"))
```
- **Updated dependency ranges**:  Adjusted for better alignment with Arkouda’s supported stack:
`py-pyarrow`, `py-scipy`, `py-tables`, and `py-h5py`.
- **New dependencies (required since 2025.07.03)**:  `py-cloudpickle`, `py-contourpy`.

### DIFF
--- a/repos/spack_repo/builtin/packages/arkouda/package.py
+++ b/repos/spack_repo/builtin/packages/arkouda/package.py
@@ -16,10 +16,8 @@ class Arkouda(MakefilePackage):
 
     # Arkouda does not have a current PyPI package, so we use the GitHub tarball
     list_url = "https://github.com/Bears-R-Us/arkouda/tags"
+    url = "https://github.com/Bears-R-Us/arkouda/archive/refs/tags/v2025.08.20.tar.gz"
     git = "https://github.com/Bears-R-Us/arkouda.git"
-
-    def url_for_version(self, version):
-        return f"https://github.com/Bears-R-Us/arkouda/archive/refs/tags/v{version}.tar.gz"
 
     # See https://spdx.org/licenses/ for a list.
     license("MIT")
--- a/repos/spack_repo/builtin/packages/arkouda/package.py
+++ b/repos/spack_repo/builtin/packages/arkouda/package.py
@@ -15,23 +15,35 @@ class Arkouda(MakefilePackage):
     homepage = "https://github.com/Bears-R-Us/arkouda"
 
     # Arkouda does not have a current PyPI package, so we use the GitHub tarball
-    url = "https://github.com/Bears-R-Us/arkouda/archive/refs/tags/v2024.10.02.tar.gz"
+    list_url = "https://github.com/Bears-R-Us/arkouda/tags"
     git = "https://github.com/Bears-R-Us/arkouda.git"
+
+    def url_for_version(self, version):
+        return f"https://github.com/Bears-R-Us/arkouda/archive/refs/tags/v{version}.tar.gz"
 
     # See https://spdx.org/licenses/ for a list.
     license("MIT")
 
     # A list of GitHub accounts to notify when the package is updated.
-    # TODO: add arkouda devs github account
-    maintainers("ajpotts", "arezaii")
+    maintainers("1RyanK", "ajpotts", "arezaii", "drculhane", "jaketrookman")
 
-    version("master", branch="master")
+    version("main", branch="main")
 
     version(
-        "2024.10.02", sha256="00671a89a08be57ff90a94052f69bfc6fe793f7b50cf9195dd7ee794d6d13f23"
+        "2025.08.20", sha256="3e305930905397ff3a7a28a5d8cc2c9adca4194ca7f6ee51f749f427a2dea92c"
     )
     version(
-        "2024.06.21", sha256="ab7f753befb3a0b8e27a3d28f3c83332d2c6ae49678877a7456f0fcfe42df51c"
+        "2025.07.03", sha256="eb888fac7b0eec6b4f3bfa0bfe14e5c8f15b449286e84c45ba95c44d8cd3917a"
+    )
+    version(
+        "2024.10.02",
+        sha256="00671a89a08be57ff90a94052f69bfc6fe793f7b50cf9195dd7ee794d6d13f23",
+        deprecated=True,
+    )
+    version(
+        "2024.06.21",
+        sha256="ab7f753befb3a0b8e27a3d28f3c83332d2c6ae49678877a7456f0fcfe42df51c",
+        deprecated=True,
     )
 
     variant(
@@ -40,17 +52,25 @@ class Arkouda(MakefilePackage):
         description="Build Arkouda for multi-locale execution on a cluster or supercomputer",
     )
 
-    depends_on("chapel@2.1: +hdf5 +zmq", type=("build", "link", "run", "test"))
+    depends_on(
+        "chapel@2.0:2.4 +hdf5 +zmq",
+        when="@2025.07.03:2025.08.20",
+        type=("build", "link", "run", "test"),
+    )
+    depends_on(
+        "chapel@2.1: +hdf5 +zmq", when="@:2025.01.13", type=("build", "link", "run", "test")
+    )
+
     depends_on("cmake@3.13.4:", type="build")
     depends_on("python@3.9:", type=("build", "link", "run", "test"))
     depends_on("libzmq@4.2.5:", type=("build", "link", "run", "test"))
     depends_on("hdf5+hl~mpi", type=("build", "link", "run", "test"))
     depends_on("libiconv", type=("build", "link", "run", "test"))
     depends_on("libidn2", type=("build", "link", "run", "test"))
-    depends_on(
-        "arrow +parquet +snappy +zlib +brotli +bz2 +lz4 +zstd",
-        type=("build", "link", "run", "test"),
-    )
+    depends_on("arrow@:19+brotli+bz2+lz4+parquet+snappy+zlib+zstd", type=("build", "link", "run"))
+
+    # force lz4 to use cmake (add as a direct dep to control its variant)
+    depends_on("lz4 build_system=cmake", type="build")
 
     requires("^chapel comm=none", when="~distributed")
     requires("^chapel +python-bindings", when="@2024.10.02:")
--- a/repos/spack_repo/builtin/packages/py_arkouda/package.py
+++ b/repos/spack_repo/builtin/packages/py_arkouda/package.py
@@ -14,10 +14,8 @@ class PyArkouda(PythonPackage):
 
     # Updating the arkouda PyPI package is future work
     list_url = "https://github.com/Bears-R-Us/arkouda/tags"
+    url = "https://github.com/Bears-R-Us/arkouda/archive/refs/tags/v2025.08.20.tar.gz"
     git = "https://github.com/Bears-R-Us/arkouda.git"
-
-    def url_for_version(self, version):
-        return f"https://github.com/Bears-R-Us/arkouda/archive/refs/tags/v{version}.tar.gz"
 
     # See https://spdx.org/licenses/ for a list.
     license("MIT")
--- a/repos/spack_repo/builtin/packages/py_arkouda/package.py
+++ b/repos/spack_repo/builtin/packages/py_arkouda/package.py
@@ -13,8 +13,11 @@ class PyArkouda(PythonPackage):
     homepage = "https://github.com/Bears-R-Us/arkouda"
 
     # Updating the arkouda PyPI package is future work
-    url = "https://github.com/Bears-R-Us/arkouda/archive/refs/tags/v2024.10.02.tar.gz"
+    list_url = "https://github.com/Bears-R-Us/arkouda/tags"
     git = "https://github.com/Bears-R-Us/arkouda.git"
+
+    def url_for_version(self, version):
+        return f"https://github.com/Bears-R-Us/arkouda/archive/refs/tags/v{version}.tar.gz"
 
     # See https://spdx.org/licenses/ for a list.
     license("MIT")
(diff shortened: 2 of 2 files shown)


repos/spack_repo/builtin/packages/arkouda/package.py
```diff
@@ -40,15 +52,27 @@ class Arkouda(MakefilePackage):
...
+    )
     depends_on("cmake@3.13.4:", type="build")
     depends_on("python@3.9:", type=("build", "link", "run", "test"))
     depends_on("libzmq@4.2.5:", type=("build", "link", "run", "test"))
     depends_on("hdf5+hl~mpi", type=("build", "link", "run", "test"))
     depends_on("libiconv", type=("build", "link", "run", "test"))
     depends_on("libidn2", type=("build", "link", "run", "test"))
     depends_on(
-        "arrow +parquet +snappy +zlib +brotli +bz2 +lz4 +zstd",
+        "arrow +parquet +snappy +zlib +brotli +bz2 ~lz4 +zstd",
```
arezaii: Do all the listed versions of Arkouda in this package work with Arrow without lz4? 

If any previous versions do require it, we may need to modify this `depends_on` a little.

Also, is `lz4` explicitly not to be used, or is it more like Arkouda can make do with any Arrow that may exist, whether or not it has `lz4` support?

If it's the latter, I might suggest just removing the `lz4` component from here entirely, which will allow the spack concretizer to choose an install of arrow that has or does not have `lz4`, so long as it does have the other required variants.

repos/spack_repo/builtin/packages/arkouda/package.py
```diff
@@ -40,15 +52,27 @@ class Arkouda(MakefilePackage):
         description="Build Arkouda for multi-locale execution on a cluster or supercomputer",
     )
 
-    depends_on("chapel@2.1: +hdf5 +zmq", type=("build", "link", "run", "test"))
+    # For Arkouda releases before July 2025, support up to Chapel 2.4.x
+    depends_on(
+        "chapel@2.0.0:2.4.99 +hdf5 +zmq",
```
arifthpe: There is no need for the `.99`, the spack version specifier `2.4` will default to the latest patch release under 2.4. Same for 2.5.99 below

arifthpe: ~~Also, I'm assuming the use of 2.0.0 specifically is intentional, but if you're okay with newer patch versions there you could just specify 2.0. That would actually make a difference in this case as there is a Chapel 2.0.1. Ditto below~~

edit: never mind, that's incorrect, it's a lower bound so you're not excluding chapel 2.0.1 here

repos/spack_repo/builtin/packages/arkouda/package.py
ajpotts: Arkouda does need lz4 to write Parquet with compression="lz4".

arezaii: Ok, then adding `lz4` back is probably necessary, although there are a couple things we could do here. 

1. We could make an `lz4` variant of the arkouda spackage, such that `lz4` for `arrow` is only required when the `lz4` variant of arkouda was specified.
2. Alternatively we could require `lz4` as before, but put a restriction on the version number to work around the issue where `lz4` versions above `1.9` cause build failures. 
This is slightly less preferred because it encodes this restriction in the `arkouda` spackage, when the problem is really with another spackage (maybe `arrow`, maybe `lz4`).
3. Debug the issue with `lz4@1.10:` and submit a spackage PR for the bug.
4. Don't add `lz4` back and don't offer support for `compression=lz4` with the spack install. (probably the least preferred option)
5. something else/other ideas?

repos/spack_repo/builtin/packages/arkouda/package.py
```diff
@@ -15,42 +15,64 @@ class Arkouda(MakefilePackage):
...
 
     version(
-        "2024.10.02", sha256="00671a89a08be57ff90a94052f69bfc6fe793f7b50cf9195dd7ee794d6d13f23"
+        "2025.08.20", sha256="3e305930905397ff3a7a28a5d8cc2c9adca4194ca7f6ee51f749f427a2dea92c"
     )
     version(
-        "2024.06.21", sha256="ab7f753befb3a0b8e27a3d28f3c83332d2c6ae49678877a7456f0fcfe42df51c"
+        "2025.07.03", sha256="eb888fac7b0eec6b4f3bfa0bfe14e5c8f15b449286e84c45ba95c44d8cd3917a"
     )
+    # version(
```
arifthpe: It's generally expected to deprecate packages for a `spack-packages` release cycle first before removing them (https://spack.readthedocs.io/en/latest/packaging_guide_creation.html#deprecating-old-versions), so that should probably be done here

repos/spack_repo/builtin/packages/arkouda/package.py
```diff
@@ -15,42 +15,64 @@ class Arkouda(MakefilePackage):
...
         description="Build Arkouda for multi-locale execution on a cluster or supercomputer",
     )
 
-    depends_on("chapel@2.1: +hdf5 +zmq", type=("build", "link", "run", "test"))
+    depends_on(
+        "chapel@2.0:2.4 +hdf5 +zmq",
+        when="@2025.07.03:2025.08.20",
+        type=("build", "link", "run", "test"),
+    )
+    # depends_on(
```
arifthpe: Is there a reason this is commented instead of just removed? There's always the git history if archaelogy needs to be done later

repos/spack_repo/builtin/packages/arkouda/package.py
```diff
@@ -63,7 +85,6 @@ class Arkouda(MakefilePackage):
     )
 
     # Some systems need explicit -fPIC flag when building the Arrow functions
-    patch("makefile-fpic-2024.06.21.patch", when="@2024.06.21")
```
arifthpe, line 65: Will need to be restored if this version is brought back as deprecated, and if it _doesn't_ end up coming back the actual patch file in this package dir should also be removed

repos/spack_repo/builtin/packages/arkouda/package.py
```diff
@@ -15,42 +15,64 @@ class Arkouda(MakefilePackage):
     homepage = "https://github.com/Bears-R-Us/arkouda"
 
     # Arkouda does not have a current PyPI package, so we use the GitHub tarball
-    url = "https://github.com/Bears-R-Us/arkouda/archive/refs/tags/v2024.10.02.tar.gz"
+    url = "https://github.com/Bears-R-Us/arkouda/archive/refs/tags/v2025.08.20.tar.gz"
+    list_url = "https://github.com/Bears-R-Us/arkouda/tags"
     git = "https://github.com/Bears-R-Us/arkouda.git"
 
+    def url_for_version(self, version):
```
arifthpe: Out of curiosity, was this added because the default `url_for_version` was picking things up incorrectly?

repos/spack_repo/builtin/packages/arkouda/package.py
```diff
@@ -15,42 +15,64 @@ class Arkouda(MakefilePackage):
     homepage = "https://github.com/Bears-R-Us/arkouda"
 
     # Arkouda does not have a current PyPI package, so we use the GitHub tarball
-    url = "https://github.com/Bears-R-Us/arkouda/archive/refs/tags/v2024.10.02.tar.gz"
+    url = "https://github.com/Bears-R-Us/arkouda/archive/refs/tags/v2025.08.20.tar.gz"
```
arifthpe: You may already know this, but there's no need to update this URL to match the latest, since that's `url_for_version`'s job. But also no harm in doing so

repos/spack_repo/builtin/packages/py_arkouda/package.py
```diff
@@ -13,45 +13,69 @@ class PyArkouda(PythonPackage):
...
 
     version(
-        "2024.10.02", sha256="00671a89a08be57ff90a94052f69bfc6fe793f7b50cf9195dd7ee794d6d13f23"
+        "2025.08.20", sha256="3e305930905397ff3a7a28a5d8cc2c9adca4194ca7f6ee51f749f427a2dea92c"
     )
     version(
-        "2024.06.21", sha256="ab7f753befb3a0b8e27a3d28f3c83332d2c6ae49678877a7456f0fcfe42df51c"
+        "2025.07.03", sha256="eb888fac7b0eec6b4f3bfa0bfe14e5c8f15b449286e84c45ba95c44d8cd3917a"
     )
+    # version(
```
arifthpe: Ditto about needing to deprecate old versions. That also means the commented/removed `depends_on`s for those old versions below would need to be restored, if those versions needed them.

repos/spack_repo/builtin/packages/arkouda/package.py
ajpotts: I was just being thorough.

repos/spack_repo/builtin/packages/arkouda/package.py
ajpotts: OK.

repos/spack_repo/builtin/packages/arkouda/package.py
ajpotts: I could deprecated it but as far as I could tell it doesn't actually work.  At least I couldn't get it to install after considerable effort.

repos/spack_repo/builtin/packages/arkouda/package.py
ajpotts: I'm hoping to add some of the older versions if I can get them to work.  However, I think that needs to be a separate PR because of the time involved.  However, I could delete it and add it back in later if you think that is best.

repos/spack_repo/builtin/packages/arkouda/package.py
ajpotts: Removed the url.

repos/spack_repo/builtin/packages/arkouda/package.py
ajpotts: This has been fixed.

repos/spack_repo/builtin/packages/arkouda/package.py
ajpotts: I added the packages back in and marked them as deprecated.

repos/spack_repo/builtin/packages/arkouda/package.py
ajpotts: I'm just deleting these for now.

repos/spack_repo/builtin/packages/arkouda/package.py
ajpotts: I left it in since you wanted to deprecate the package.

repos/spack_repo/builtin/packages/arkouda/package.py
ajpotts: It's been removed.

repos/spack_repo/builtin/packages/py_arkouda/package.py
ajpotts: I marked it as deprecated.

repos/spack_repo/builtin/packages/arkouda/package.py
arifthpe: I see, in that case either way seems fine to me, not sure if there's a spack convention about this

repos/spack_repo/builtin/packages/arkouda/package.py
ajpotts: Can you take a look and see if your concern is resolved?

repos/spack_repo/builtin/packages/arkouda/package.py
arifthpe: I'm not sure if you're supposed to deprecate or remove for versions that are broken. If they're completely unusably broken, so no one might be using them in the wild, IMO just deleting is fine. But a Spack maintainer may still disagree, if you want to I think it's reasonable to leave it removed until one comes along to review

repos/spack_repo/builtin/packages/arkouda/package.py
arifthpe: TIL `url` field isn't needed if you have `url_for_version`. I had meant just that you don't need to update `url`, not that you don't need it entirely, but turns out that's true per the version precedence list at https://spack.readthedocs.io/en/latest/packaging_guide_creation.html#versions-and-urls.

arifthpe (approved):

arezaii (approved):

drculhane (approved):

jaketrookman (approved):
Looks good

1RyanK (approved):

jaketrookman (approved):
Looks good

drculhane (approved):
Looks straightforward.

becker33 (approved):

PR #1190, opened 2025-08-26
charliecloud: add v0.39 and v0.40 with CDI support

alstar555:
* v0.39+: Use GitLab archive URLs
* v0.39+: Add py-lark as build dependency for autogen.sh because GitLab tarballs lack pre-built configure scripts
* v0.39+: Add `force_autoreconf` method to run autogen.sh for GitLab tarballs
* v0.40+: Add CDI with cjson dependency
* v0.40+: Update squashfuse flag syntax


repos/spack_repo/builtin/packages/charliecloud/package.py
```diff
@@ -107,6 +109,7 @@ class Charliecloud(AutotoolsPackage):
 
     variant("docs", default=False, description="Build man pages and html docs")
     variant("squashfuse", default=True, description="Build with squashfuse support", when="@0.32:")
+    variant("cdi", default=False, description="Build with CDI support", when="@0.40:")
```
reidpr: CDI is really nice, so my guess would be default this true?

repos/spack_repo/builtin/packages/charliecloud/package.py
```diff
@@ -124,6 +127,7 @@ class Charliecloud(AutotoolsPackage):
     depends_on("py-requests", type="run")
     depends_on("git@2.28.1:", type="run", when="@0.29:")  # build cache
     depends_on("py-lark", type="run", when="@:0.24")  # 0.25+ bundles lark
+    depends_on("py-lark", type="build", when="@0.39:")
```
reidpr: I’m uncertain this is the right boundary for depending on Lark; did something change in 0.39 that I’ve forgotten about? My guess is that we should either depend on the correct version of Lark or use the bundled Lark, for all Charliecloud versions.

If we do use Spack’s Lark, there are arguments to `autogen.sh` to not download the bundled Lark.

repos/spack_repo/builtin/packages/charliecloud/package.py
```diff
@@ -170,9 +189,23 @@ def configure_args(self):
             args.append("--disable-html")
 
         if self.spec.satisfies("+squashfuse"):
-            args.append(f"--with-libsquashfuse={self.spec['squashfuse'].prefix}")
+            if self.spec.satisfies("@0.40:"):
+                # Version 0.40 uses a new syntax for squashfuse
+                args.append(f"--with-squashfuse-include={self.spec['squashfuse'].prefix}/include")
+                args.append(f"--with-squashfuse-lib={self.spec['squashfuse'].prefix}/lib")
+            else:
+                args.append(f"--with-libsquashfuse={self.spec['squashfuse'].prefix}")
```
reidpr: Random style suggestion. I would reverse the sense of the conditional here, i.e.:
```suggestion
            if self.spec.satisfies("@:0.40"):
                args.append(f"--with-libsquashfuse={self.spec['squashfuse'].prefix}")                
            else:
                # Version 0.40 uses a new syntax for squashfuse
                args.append(f"--with-squashfuse-include={self.spec['squashfuse'].prefix}/include")
                args.append(f"--with-squashfuse-lib={self.spec['squashfuse'].prefix}/lib")
```

because then the shorter branch is first, bringing the `if` and `else` closer together, the reasoning being that the reader would remember the condition better if they get to `else` sooner.

reidpr (commented):
I don’t know the Spack code, but this looks pretty solid. Some suggestions in-line.

reidpr (approved):

j-ogas (approved):
LGTM.

tgamblin (approved):

PR #1191, opened 2025-08-26
snakemake: prefer most recent 8-series version

wdconinc:
This PR marks `snakemake`, v8.25.2 as preferred as a quick temporary fix to avoid that an incomplete v9.6.3 is used. In parallel, we can go ahead and make the necessary additions for v9 (which requires adding updated versions for several  of the dependencies, and adding those version requirements to the `snakemake` package dependencies).


wdconinc:
For review, @markusboden @w8jcik.

alalazo (approved):

w8jcik (approved):

w8jcik:
> In parallel, we can go ahead and make the necessary additions for v9

@wdconinc are you working on this, or should I create a PR?

Update: I just noticed that it is [already in progress](https://github.com/spack/spack-packages/pull/482#issuecomment-3225285783) so all good.

PR #1192, opened 2025-08-26
Patch ispc 1.19 to make it compile

paulmelis:
I screwed up the merge/rebase on https://github.com/spack/spack-packages/pull/1091. And rather than try to figure out if it can still be salvaged, I made this new clean PR, which only concerns ispc (and not openvkl)

### DIFF
--- a/repos/spack_repo/builtin/packages/ispc/package.py
+++ b/repos/spack_repo/builtin/packages/ispc/package.py
@@ -81,12 +81,17 @@ class Ispc(CMakePackage):
         sha256="d3ccf547d3ba59779fd375e10417a436318f2200d160febb9f830a26f0daefdc",
     )
 
+    # Turn a few regexp literals into raw strings, as otherwise some escape sequence fail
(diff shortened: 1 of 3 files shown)


repos/spack_repo/builtin/packages/ispc/package.py
```diff
@@ -81,6 +81,18 @@ class Ispc(CMakePackage):
...
+        "ispc-1.19-regexp.patch",
+        when="@1.19",
+        sha256="dc5e5492442df91ba17877d84f062bc364140ee5620ed55b7beef87257030b74",
+    )
+
+    patch(
+        "stdint-fix.patch",
+        when="@1.19",
+        sha256="03303b2407718e340bfc574511f736ae0f6678e60db33f68147db29bdd806a1a",
+    )
```
alalazo: Can you add a brief comment above the `patch` directives explaining:
2. Why it is needed
3. Where it was taken from

?

paulmelis: Done

alalazo (approved):

PR #1196, opened 2025-08-26
elastix: add new package elastix for medical image registration

Markus92:
This PR adds [elastix](https://elastix.dev/), a medical image registration tool to spack.

I only checksummed 5.2.0 and 5.1.0, older versions have some wonky CMakelists and aren't always the easiest to compile, these versions are fine though. It's not a package known for having major regressions or backwards-incompatible changes over the years.

It requires a modern version of ITK (see #1195 for PR) to build.

There is some compiler options for using OpenCL but that functionality is not that well maintained AFAIK, so focusing on getting this in first.


Markus92:
@spackbot fix style

repos/spack_repo/builtin/packages/elastix/package.py
```diff
@@ -0,0 +1,42 @@
...
+
+    version("5.2.0", sha256="7267d7f2efccc3ddd9529aa83e4d10eeea3707a972de06b2e020fc95d4bec6c1")
+    version("5.1.0", sha256="f2e3e33359d1e35cb986bc1eb7a1b0179cdb20a67e410ac8423102a160bfc29e")
+
+    depends_on("c", type="build")
+    depends_on("cxx", type="build")
+
+    depends_on("itk@5.4:", when="@5.2:")
+    depends_on("itk@5.3:", when="@5.1:")
+    depends_on("itk@5.1.1:", when="@5.0.1")
```
becker33: To pass the audits, you'll need to remove this if you aren't adding version 5.0.1.

becker33 (changes requested):

repos/spack_repo/builtin/packages/elastix/package.py
Markus92: Thanks for the suggestion!

becker33 (approved):

PR #1205, opened 2025-08-27
add-rucio-client-38.0.0

panta-123:
Added rucio client 38.0.0 .

### DIFF
--- a/repos/spack_repo/builtin/packages/py_rucio_clients/package.py
+++ b/repos/spack_repo/builtin/packages/py_rucio_clients/package.py
@@ -17,6 +17,7 @@ class PyRucioClients(PythonPackage):
 
     license("Apache-2.0", checked_by="wdconinc")
 
+    version("38.0.0",  sha256="d49f912f2f98870cab2227e0464129ba0954e99b975d0225126cca1b9d9c983c")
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/py_rucio_clients/package.py
```diff
@@ -51,7 +52,8 @@ class PyRucioClients(PythonPackage):
     depends_on("py-packaging@24.2:", type=("build", "run"), when="@37:")
     depends_on("py-rich@13.7.1:", type=("build", "run"), when="@36:")
     depends_on("py-rich@13.9.4:", type=("build", "run"), when="@37:")
-    depends_on("py-typing-extensions@4.12.2:", type=("build", "run"))
+    depends_on("py-typing-extensions@4.12.2:", type=("build", "run"), when="@:37")
+    depends_on("py-typing-extensions@4.14.0:", type=("build", "run"), when="@38:")
```
wdconinc: Rather than modifying previous lines with end ranges, it is more common to just add new lines (this reduces diff and blame noise):

```suggestion
    depends_on("py-typing-extensions@4.12.2:", type=("build", "run"))
    depends_on("py-typing-extensions@4.14.0:", type=("build", "run"), when="@38:")
```

This has the same effect of having a new version requirement for v38 and newer.

wdconinc (approved):
Looks good. Thanks!

PR #1212, opened 2025-08-28
llvm: add v21.1.0, v21.1.1, v21.1.2, v21.1.3, v21.1.4

pearzt:
CC maintainers @trws @haampie @skosukhin

### DIFF
--- a/repos/spack_repo/builtin/packages/llvm/package.py
+++ b/repos/spack_repo/builtin/packages/llvm/package.py
@@ -630,7 +630,7 @@ class Llvm(CMakePackage, CudaPackage, LlvmDetection, CompilerPackage):
     patch(
         "https://github.com/llvm/llvm-project/commit/cd24d108a2c19c23c4ac80b501fa7361963cca3d.patch?full_index=1",
         sha256="0dc6e0bf66edf260b56c088dfbf37abb8417e210f256abe4ee11c395a2665ed8",
-        when="@21"
+        when="@21",
     )
 
     @when("@14:17")
--- a/repos/spack_repo/builtin/packages/llvm/package.py
+++ b/repos/spack_repo/builtin/packages/llvm/package.py
@@ -54,6 +54,7 @@ class Llvm(CMakePackage, CudaPackage, LlvmDetection, CompilerPackage):
     version("main", branch="main")
 
     # Latest stable
+    version("21.1.4", sha256="3a0921d78be74302cb054da1dad59e706814d8fed3a6ac9b532e935825a0715c")
     version("21.1.3", sha256="5bc91fe86bafebc64189465faca1ff35626dcb1b8539a14ae2ec07834c3e8e95")
     version("21.1.2", sha256="eced3dd78186621f4df8a1accbcd1ecf2ee399571e62d052c21e9bf363af2166")
     version("21.1.1", sha256="5f048351ee63050d7fa45b6a1160768fb222a8d306a89e1344515ef7a4bcd278")
@@ -630,7 +631,7 @@ class Llvm(CMakePackage, CudaPackage, LlvmDetection, CompilerPackage):
     patch(
         "https://github.com/llvm/llvm-project/commit/cd24d108a2c19c23c4ac80b501fa7361963cca3d.patch?full_index=1",
         sha256="0dc6e0bf66edf260b56c088dfbf37abb8417e210f256abe4ee11c395a2665ed8",
-        when="@21",
+        when="@21.1.0:21.1.4",
     )
 
     @when("@14:17")
--- a/repos/spack_repo/builtin/packages/llvm/package.py
+++ b/repos/spack_repo/builtin/packages/llvm/package.py
@@ -72,13 +72,27 @@ class Llvm(CMakePackage, CudaPackage, LlvmDetection, CompilerPackage):
     version("20.1.0", sha256="08bc382733777dda3c96259e3732ff96c1df98d0470c4f85b163274eae687f4f")
 
     with default_args(deprecated=True):
-        version("19.1.6", sha256="f07fdcbb27b2b67aa95e5ddadf45406b33228481c250e65175066d36536a1ee2")
-        version("19.1.5", sha256="e2204b9903cd9d7ee833a2f56a18bef40a33df4793e31cc090906b32cbd8a1f5")
-        version("19.1.4", sha256="010e1fd3cabee8799bd2f8a6fbc68f28207494f315cf9da7057a2820f79fd531")
-        version("19.1.3", sha256="e5106e2bef341b3f5e41340e4b6c6a58259f4021ad801acf14e88f1a84567b05")
-        version("19.1.2", sha256="622cb6c5e95a3bb7e9876c4696a65671f235bd836cfd0c096b272f6c2ada41e7")
-        version("19.1.1", sha256="115dfd98a353d05bffdab3f80db22f159da48aca0124e8c416f437adcd54b77f")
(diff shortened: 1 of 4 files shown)


rbberger:
Looks like there is a CI failure in multiple jobs due to a missing `ld.lld` linker. It appears that in the `offload` project they are now forcing the `lld` linker, although it supposedly was always required for AMDGPU.

https://github.com/llvm/llvm-project/blob/3623fe661ae35c6c80ac221f14d85be76aa870f1/offload/DeviceRTL/CMakeLists.txt#L129-L160

More info can be found in this commit: https://github.com/llvm/llvm-project/commit/346792aafb483a53fb5e3274298d85bc2dde4a35 and the linked issues.

At minimum there should be
```
conflicts("~lld", when="+offload")
```
assuming that will make it visible for that part of the build. Since the affected CI stacks explicitly do `~lld`, this will have to be changed too or `~offload` needs to be added.

pearzt:
@rbberger Thanks for looking into this! I added `conflicts("~lld", when="+offload")` to express `offloads` dependecy on `lld`. So far, I have not limited this to `@21:`, since `offload` has always required `lld`, according to the commit message that you referenced.

I'm not really familiar with the CI stacks. Why would we need to change those and not just let the concretizer disable `offload`?

pearzt:
@eugeneswalker I have increased the job timeout on Darwin in an attempt to fix the job timeout [(link)](https://gitlab.spack.io/spack/spack-packages/-/jobs/17973233), but have now realized that there is another problem with the build on Darwin:

```
kmp_alloc.cpp:1549:46: error: unknown type name 'hwloc_membind_policy_t'
 1549 | static bool __kmp_is_hwloc_membind_supported(hwloc_membind_policy_t policy) {
```

The hwloc version in the build environment (2.11) seems okay to me, so I'm not sure what the problem is here. Maybe it picks up an old version of hwloc installed in the system? Can someone with more experience building LLVM on Macs look into this?

rbberger:
@pearzt no, you're right, the concretizer will disable it. I just wanted it to be explicit. that way you don't have to reason about it afterwards. But it's not absolutely necessary.

rbberger:
> The hwloc version in the build environment (2.11) seems okay to me, so I'm not sure what the problem is here. Maybe it picks up an old version of hwloc installed in the system? Can someone with more experience building LLVM on Macs look into this?

Seems to me there is an actual bug in the LLVM openmp build for Darwin that was introduced in the past 3-6 months. `kmp_alloc.cpp` has [code parts](https://github.com/llvm/llvm-project/blob/3623fe661ae35c6c80ac221f14d85be76aa870f1/openmp/runtime/src/kmp_alloc.cpp#L1548-L1564) that use symbols that are only defined if `KMP_AFFINITY_SUPPORTED`, which [isn't the case for Darwin](https://github.com/llvm/llvm-project/blob/llvmorg-21.1.0/openmp/runtime/src/kmp.h#L665). The `hwloc.h` header isn't even [included in that case](https://github.com/llvm/llvm-project/blob/llvmorg-21.1.0/openmp/runtime/src/kmp.h#L108).

So we either patch it or turn off hwloc completely. I'll reach out to LLVM devs, maybe a fix is changing the preprocessor conditions from `KMP_USE_HWLOC` to `KMP_USE_HWLOC && KMP_AFFINITY_SUPPORTED` in those parts.

pearzt:
> @pearzt no, you're right, the concretizer will disable it. I just wanted it to be explicit. that way you don't have to reason about it afterwards. But it's not absolutely necessary.

Ah, got it. I agree that making it explicit it's probably better to make it explicit, so I've added `~offload` to all stacks that have `~lld`.

rbberger:
@pearzt I'm not sure how long it's gonna take for the LLVM folks to fix their issue. It might be worth to disable hwloc for libomp on Darwin for now and add a link to the LLVM issue in a comment. I don't think this is hard to fix, but I just didn't have any spare cycles myself to tackle it.

pearzt:
@rbberger Ah, thanks for reminding me of this. I have opened a PR in LLVM to fix this, although I don't work on a Mac and can't really test whether it fixes our issue on Darwin. Could you have a look? https://github.com/llvm/llvm-project/pull/158349

I agree that, should fixing this be more complicated than what I'm proposing in my PR, we should just disable hwloc on Darwin, since LLVM doesn't seem to use it on that platform anyway.

rbberger:
@pearzt sorry, I'll be out on vacation the next few weeks and that includes being without any of my devices.

rbberger:
@pearzt looks like you MR got in. now you can use https://github.com/llvm/llvm-project/commit/cd24d108a2c19c23c4ac80b501fa7361963cca3d.patch?full_index=1 to patch the versions here. Looks like there have been more 21.x releases in the meantime, but we missed last week's. So at the moment the patch would have to apply for `@21.1.0:21.1.3`.

pearzt:
@spackbot fix style

pearzt:
@spackbot fix style

pearzt:
@rbberger To get patch in to the next LLVM-v21.1.x release, I would need to open another PR to get the commits into the `release/21.x` branch (https://llvm.org/docs/GitHub.html#backporting-fixes-to-the-release-branches). Otherwise, it will only be released with LLVM 22. Let's wait if this actually fixes our CI problem first.

I have removed the increased CI timeout for now as I'm not sure if it's actually necessary.

pearzt:
LLVM PR to backport the fix for the next 21.x release: https://github.com/llvm/llvm-project/pull/163768

pearzt:
As the [PR to backport](https://github.com/llvm/llvm-project/pull/163768) the fix to LLVM's `release/21.x` branch did not get merged in time for the [release of 21.1.4](https://github.com/llvm/llvm-project/releases/tag/llvmorg-21.1.4), I suggest we merge this (assuming green CI) so LLVM 21 is finally available via Spack. @trws @haampie @skosukhin

rbberger (dismissed):

repos/spack_repo/builtin/packages/llvm/package.py
```diff
@@ -64,17 +71,8 @@ class Llvm(CMakePackage, CudaPackage, LlvmDetection, CompilerPackage):
...
 
-    # Previous stable series releases
-    version("19.1.7", sha256="59abea1c22e64933fad4de1671a61cdb934098793c7a31b333ff58dc41bff36c")
-    version("19.1.6", sha256="f07fdcbb27b2b67aa95e5ddadf45406b33228481c250e65175066d36536a1ee2")
-    version("19.1.5", sha256="e2204b9903cd9d7ee833a2f56a18bef40a33df4793e31cc090906b32cbd8a1f5")
-    version("19.1.4", sha256="010e1fd3cabee8799bd2f8a6fbc68f28207494f315cf9da7057a2820f79fd531")
-    version("19.1.3", sha256="e5106e2bef341b3f5e41340e4b6c6a58259f4021ad801acf14e88f1a84567b05")
-    version("19.1.2", sha256="622cb6c5e95a3bb7e9876c4696a65671f235bd836cfd0c096b272f6c2ada41e7")
-    version("19.1.1", sha256="115dfd98a353d05bffdab3f80db22f159da48aca0124e8c416f437adcd54b77f")
-    version("19.1.0", sha256="0a08341036ca99a106786f50f9c5cb3fbe458b3b74cab6089fd368d0edb2edfe")
```
alalazo, line 75: LGTM, but I think we decided to deprecate these versions first, before removing them completely.

pearzt: Fixed. I was unaware of this decision, so maybe we should documented it in a comment?

pearzt:
@spackbot fix style

alalazo (approved):

pearzt:
CI failure seems to be due to some kind of infrastructure problem.

@spackbot re-run pipeline

PR #1218, opened 2025-08-28
Deprecation removals: A-D

adamjstewart:


adamjstewart:
@spackbot fix style

adamjstewart:
Requires #1236

adamjstewart:
@cyrush can you investigate the ascent build failure?

cyrush:
@adamjstewart  It looks like it it is trying to build with a newer umpire than ascent 0.8.0 supports. 

Ascent 0.8.0 is from 2022, so it's quite old at this point.

adamjstewart:
@adrienbernede  @davidbeckingsale  @kab163 can you look over the changes to umpire? Many of the patches didn't work for 2022.10.0.

adamjstewart:
Seems like umpire is still too new. Any idea why it's trying to build such an old version of ascent?

repos/spack_repo/builtin/packages/umpire/package.py
```diff
@@ -178,13 +178,6 @@ class Umpire(CachedCMakePackage, CudaPackage, ROCmPackage):
         when="@:5.0.1 ^blt@0.4:",
     )
 
-    # https://github.com/LLNL/Umpire/pull/805
-    patch(
```
becker33, line 182: Why is this patch no longer needed? The relevant version is still in the package?

repos/spack_repo/builtin/packages/ascent/package.py
```diff
@@ -761,25 +741,6 @@ def hostconfig(self):
...
-            else:
-                cfg.write("# devil ray from spack \n")
-                cfg.write(cmake_cache_entry("DRAY_DIR", spec["dray"].prefix))
-        else:
-            if self.spec.satisfies("@0.8.1:"):
-                cfg.write("# devil ray\n")
-                cfg.write(cmake_cache_entry("ENABLE_DRAY", "OFF"))
-                cfg.write(cmake_cache_entry("ENABLE_APCOMP", "OFF"))
-            else:
-                cfg.write("# devil ray not build by spack\n")
```
becker33: I'm not certain that the variant isn't still effective to control whether the `dray` portion of the source is built (can't tell from the packages). @cyrush can you confirm whether the variant should still be present absent supporting old external versions of dray?

repos/spack_repo/builtin/packages/cupla/package.py
```diff
@@ -1,47 +0,0 @@
...
-
-
-class Cupla(Package):
-    """C++ User interface for the Platform independent Library Alpaka"""
-
-    homepage = "https://github.com/alpaka-group/cupla"
-    git = "https://github.com/alpaka-group/cupla.git"
-    url = "https://github.com/alpaka-group/cupla/archive/refs/tags/0.3.0.tar.gz"
-
-    maintainers("vvolkl")
```
becker33, line 18: @vvolkl Are you intending to add newer versions of cupla that build with non-deprecated versions of `alpaka`?

repos/spack_repo/builtin/packages/umpire/package.py
adamjstewart: Run `spack patch umpire@2022.10.0` on develop and you'll see why. I have no idea how these versions ever worked.

adamjstewart:
@spackbot run pipeline

repos/spack_repo/builtin/packages/ascent/package.py
cyrush: A while back, Devil Ray was moved to be developed inside of Ascent (separate repo was frozen). That said, it is still an option in Ascent - it can be turned on or off for a build. For newer ascent variant was used to select if it should be on or off for a given build.

repos/spack_repo/builtin/packages/ascent/package.py
```diff
@@ -126,7 +126,6 @@ class Ascent(CMakePackage, CudaPackage, ROCmPackage):
     variant("raja", default=True, description="Build with RAJA support")
     variant("umpire", default=True, description="Build with Umpire support")
     variant("mfem", default=False, description="Build MFEM filter support")
-    variant("dray", default=False, description="Build with Devil Ray support")
```
becker33, line 129: This variant cannot be removed yet, since the option wasn't removed the package was brought into ascent proper. It should be

```
variant("dray", default=False, when="@0.8.1:", description="Build with Devil Ray support")
```
The older versions required the (now removed) external package, but the newer versions do not.

repos/spack_repo/builtin/packages/ascent/package.py
```diff
@@ -761,25 +741,6 @@ def hostconfig(self):
         else:
             cfg.write("# occa not built by spack \n")
 
-        #######################
-        # Devil Ray
-        #######################
-        if spec.satisfies("+dray"):
```
becker33, line 767: This should still respect the dray variant for newer versions that have it internally

```
# dray variant is not guaranteed for all versions
# only apply ON/OFF when the variant is set one way or the other
if self.spec.satisfies("+dray"):
    cfg.write(cmake_cache_entry("ENABLE_DRAY", "ON"))
    cfg.write(cmake_cache_entry("ENABLE_APCOMP", "ON"))
elif self.spec.satisfies("~dray"):
    cfg.write(cmake_cache_entry("ENABLE_DRAY", "OFF"))
    cfg.write(cmake_cache_entry("ENABLE_APCOMP", "OFF"))

File: repos/spack_repo/builtin/packages/cupla/package.py (hunk 3 above)
Comment (becker33) [reply repos/spack_repo/builtin/packages/cupla/package.py:18]: Resolving as the maintainer has not responded

**Review by becker33 — CHANGES_REQUESTED** [review]
Everything looks good except for the `dray` variant in `ascent` -- see comments on that package.

File: repos/spack_repo/builtin/packages/ascent/package.py (hunk 4 above)
Comment (adamjstewart) [reply repos/spack_repo/builtin/packages/ascent/package.py:129]: Thanks, now I understand the request. Done.

File: repos/spack_repo/builtin/packages/ascent/package.py [hunk 6]
```diff
@@ -766,19 +747,12 @@ def hostconfig(self):
...
             cfg.write("# devil ray\n")
-            if self.spec.satisfies("@0.8.1:"):
-                cfg.write(cmake_cache_entry("ENABLE_DRAY", "ON"))
-                cfg.write(cmake_cache_entry("ENABLE_APCOMP", "ON"))
-            else:
-                cfg.write("# devil ray from spack \n")
-                cfg.write(cmake_cache_entry("DRAY_DIR", spec["dray"].prefix))
+            cfg.write(cmake_cache_entry("ENABLE_DRAY", "ON"))
+            cfg.write(cmake_cache_entry("ENABLE_APCOMP", "ON"))
         else:
```
becker33: This needs to be `elif spec.satisfies("~dray"):`

becker33 (changes requested):
Otherwise LGTM

jrood-nrel (approved):

becker33 (approved):

cyrush:
I am fine with removing external dray and having folks use newer Ascent to use these features. 
But what is the driver / process for marking packages deprecated and removing them?  I don't understand the context from this PR.

adamjstewart:
Motivation is to delete ~4K lines of code that no longer need to be maintained.

Process is documented here: https://spack.readthedocs.io/en/latest/packaging_guide_creation.html#deprecating-old-versions

cyrush:
@adamjstewart  thanks for the link. 

That shares reasons for removing a package, and says all versions should be deprecated for one release cycle of the package before the package is fully removed. 

Sounds like this is cleanup before that process kicks in?

I am fine with dray being removed, just want to understand how things like this will work.

adamjstewart:
We've always had the policy, but no one ever actually bothered to remove these deprecated versions before. I'm trying to do this every 6 months (I started when 1.0 was released and it's taking forever). @becker33 and I are trying to figure out ways to better document when a version is deprecated and when it will be removed (more than just a boolean).

cyrush:
Got it - dray versions were never deprecated in the package.

I don't want to be pain - cleanup is a good thing and a lot of effort on your part. 

Just feedback on how it seemed a bit out of the blue.

PR #1221, opened 2025-08-28
Deprecation removals: GDAL

adamjstewart:


adamjstewart:
@spackbot fix style

vanderwb (approved):
Given the decision to remove older GDAL versions, there doesn't seem to be much choice other than to remove GDAL support in the NCL recipe. So that LGTM.

Truthfully, NCL itself is deprecated as it is no longer receiving any development or financial support from NSF NCAR. Is there some deprecation policy for Spack packages dictated somewhere for reference? E.g., how was this GDAL deprecation-to-removal process determined?

adamjstewart:
Deprecation policy is documented here: https://spack.readthedocs.io/en/latest/packaging_guide_creation.html#deprecating-old-versions

Typically, if a version is deprecated in one release, it isn't removed until the next minor release (~6 months later). Of course, if you rely on develop, versions can be deprecated the day before a release and removed the day after a release.

In terms of deciding what and when to deprecate, this is mostly up to package recipe maintainers. If there is a good reason to keep GDAL 2, I could be convinced. I'm hoping to slowly deprecate things before the CMake build system was added (3.4 and older), but that will likely not be for another couple years.

alalazo (dismissed):
Feel free to merge, if you're not waiting on further reviews

repos/spack_repo/builtin/packages/libemos/package.py
```diff
@@ -25,12 +25,7 @@ class Libemos(CMakePackage):
     version("4.4.7", sha256="669fb070c1ce655812882140a92b100233f065829868d9374bad2fcbb6b356e5")
     version("4.4.2", sha256="e2d20ad71e3beb398916f98a35a3c56ee0141d5bc9b3adff15095ff3b6dccea8")
 
-    variant(
-        "grib",
-        default="eccodes",
-        values=("eccodes", "grib-api"),
-        description="Specify GRIB backend",
-    )
+    variant("grib", default="eccodes", values=("eccodes",), description="Specify GRIB backend")
```
becker33: This variant should have a version guard so that we don't propagate the variant with only one value to newer versions of libemos.

```suggestion
    variant("grib", default="eccodes", values=("eccodes",), when="@:4.5.1", description="Specify GRIB backend")
```

repos/spack_repo/builtin/packages/magics/package.py
```diff
@@ -35,12 +35,7 @@ class Magics(CMakePackage):
 
     conflicts("%gcc@11:", when="@:4.4", msg="missing #include <limits>")
 
-    variant(
-        "grib",
-        default="eccodes",
-        values=("eccodes", "grib-api"),
-        description="Specify GRIB backend",
-    )
+    variant("grib", default="eccodes", values=("eccodes",), description="Specify GRIB backend")
```
becker33:
```suggestion
    variant("grib", default="eccodes", values=("eccodes",), when="@:4.15.3", description="Specify GRIB backend")
```
same here

becker33 (changes requested):

repos/spack_repo/builtin/packages/libemos/package.py
adamjstewart: We don't have a formal variant deprecation procedure. I would rather just delete the variant.

becker33 (approved):

PR #1224, opened 2025-08-28
Deprecation removals: GCC

adamjstewart:
@becker33 had specific concerns about this one. We should decide whether to remove or undeprecate. If anyone wants to undeprecate, I suggest helping to maintain this recipe.


michaelkuhn:
I don't have a strong opinion about this, but wouldn't we also need to remove the conflicts/patches/etc. that only apply to the to-be-removed versions to have an actual maintenance benefit from this?

adamjstewart:
@spackbot fix style

adamjstewart:
@spackbot run pipeline

becker33 (changes requested):
I will be adding a PR to add myself as a maintainer and slow down the deprecation policy on this package. Requesting changes to ensure this doesn't go in before I do so.

adamjstewart:
Closing in favor of #1373

PR #1230, opened 2025-08-28
llvm: also patch for lldb python detection in newer llvm versions

jrood-nrel:
I continue to get this known error https://lists.freebsd.org/archives/freebsd-ports-bugs/2024-January/047644.html by default. So I used the exact same patch done for :15 and applied it to 16: and it solves the issue.


jrood-nrel:
I'm going to reopen this because I still think it is necessary.

pearzt:
Let's ping the package maintainers @haampie @trws @skosukhin so that they can have a look at this:)

Is the patch really necessary for all LLVM versions starting with 16?

bernhardkaindl (approved):

PR #1232, opened 2025-08-28
lammps: remove deprecated versions prior to 2025

rbberger:
This removes all deprecated versions of LAMMPS prior to 2025. Older stable versions are kept for now.


repos/spack_repo/builtin/packages/lammps/package.py
```diff
@@ -53,390 +53,20 @@ class Lammps(CMakePackage, CudaPackage, ROCmPackage, PythonExtension):
         sha256="a4cb0a58451d47ac31ee3e1f148d92f445298d6e27f2d06f161b9b4168d79eb1",
         deprecated=True,
```
adamjstewart, line 54: What about these deprecated versions?

rbberger: They got deprecated in June with the new stable release. It would seem premature to remove them right away.

adamjstewart: Before or after the stable release? We typically wait to remove versions that were deprecated in a previous stable release so that users have time to adapt. So you don't need to wait for two stable releases if you don't want to.

adamjstewart (approved):

repos/spack_repo/builtin/packages/lammps/package.py
rbberger: @adamjstewart A LAMMPS "stable" release only happens once per year. The releases between stable releases are also known as "patch" releases. All patch releases become deprecated the moment there is a new stable release. So in June all patch releases between the stable of 2024 to the stable in 2025 were marked deprecated. Given this it would make sense to only prune the deprecated versions of the previous year.

PR #1236, opened 2025-08-29
vtk-h: add missing C compiler dependency

adamjstewart:
The package later tries to use `SPACK_CC`, which isn't defined unless a compiler dependency is added.

### DIFF
--- a/repos/spack_repo/builtin/packages/vtk_h/package.py
+++ b/repos/spack_repo/builtin/packages/vtk_h/package.py
@@ -68,6 +68,7 @@ class VtkH(CMakePackage, CudaPackage):
     variant("logging", default=False, description="Build vtk-h with logging enabled")
     variant("contourtree", default=False, description="Enable contour tree support")
 
+    depends_on("c", type="build")
     depends_on("cxx", type="build")  # generated
     depends_on("fortran", type="build")  # generated
 


adamjstewart:
Very strange. I discovered this missing dependency in #1218, where this build failure is causing the `data-vis-sdk-build` pipeline to fail. However, in this PR, `data-vis-sdk-build` says "no specs to rebuild". Perhaps `depends_on("c")` doesn't cause the hash to change because `depends_on("cxx")` pulls in the same compiler?

haampie:
> Perhaps `depends_on("c")` doesn't cause the hash to change because `depends_on("cxx")` pulls in the same compiler?

It shouldn't:

```pycon
>>> spack.spec.Spec("foo %[virtuals=cxx] gcc").dag_hash()
'cfunoq4p66bnk2mapxvpi5bsmxquqolm'
>>> spack.spec.Spec("foo %[virtuals=c,cxx] gcc").dag_hash()
'7m3zv4fhs6yuflccsllwz6sqrman44zl'
```

Edit: the difference is that in your previous concretization `ascent@0.8.0` was picked, and here `ascent@0.9.4`, which no longer depends on `vtk-h`.

haampie (approved):

PR #1238, opened 2025-08-29
py-pyamg: dependency update and add v5.3.0

garth-wells:
- pyamg version <= 5.2 fails with scipy version >=1.16 due to Scipy change. Update dependency to reflect this.
- Add pyamg v5.3.0

### DIFF
--- a/repos/spack_repo/builtin/packages/py_pyamg/package.py
+++ b/repos/spack_repo/builtin/packages/py_pyamg/package.py
@@ -31,7 +31,6 @@ class PyPyamg(PythonPackage):
     depends_on("py-setuptools@42:", type="build", when="@4.2.0:")
     # Later version required due to
     # https://github.com/pypa/setuptools_scm/issues/758
-    # depends_on("py-setuptools-scm@8.3.0+toml", type="build", when="@5.3.0:")
     depends_on("py-setuptools-scm@7.1:+toml", type="build", when="@5.0.0:")
     depends_on("py-setuptools-scm@5:+toml", type="build", when="@4.2.0:")
     depends_on("py-pybind11@2.8.0:", type=("build", "link"), when="@4.2.0:")
--- a/repos/spack_repo/builtin/packages/py_pyamg/package.py
+++ b/repos/spack_repo/builtin/packages/py_pyamg/package.py
@@ -18,6 +18,7 @@ class PyPyamg(PythonPackage):
 
     license("MIT")
 
+    version("5.3.0", sha256="c83e745f93e60a3fc1598ca6369049387403be1db4f6ac887866259c853f6364")
     version("5.0.0", sha256="088be4b38203e708905fa45295593c1336b127a28391486d4f5917cf0b96f5f2")
     version("4.2.3", sha256="dcf23808e0e8edf177fc4f71a6b36e0823ffb117137a33a9eee14b391ddbb733")
     version("4.1.0", sha256="9e340aef5da11280a1e28f28deeaac390f408e38ee0357d0fdbb77503747bbc4")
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/py_pyamg/package.py
```diff
@@ -26,6 +27,7 @@ class PyPyamg(PythonPackage):
     depends_on("python", type=("build", "link", "run"))
     depends_on("py-numpy@1.7:", type=("build", "run"))
     depends_on("py-scipy@0.12:", type=("build", "run"))
+    depends_on("py-scipy@0.12:1.15", type=("build", "run"), when="@:5.2")
     depends_on("py-setuptools@42:", type="build", when="@4.2.0:")
     # Later version required due to
     # https://github.com/pypa/setuptools_scm/issues/758
```
tldahlgren, line 33: You need to add a dependency on `py-setuptools-scm@8.3.0+toml` based on https://github.com/pyamg/pyamg/blob/v5.3.0/pyproject.toml.

repos/spack_repo/builtin/packages/py_pyamg/package.py
```diff
@@ -26,6 +27,7 @@ class PyPyamg(PythonPackage):
     depends_on("python", type=("build", "link", "run"))
```
tldahlgren, line 27: There should be a conditional dependency (https://spack.readthedocs.io/en/latest/packaging_guide_creation.html#conditional-dependencies) on `python@3.9:` for at least `5.3.0` based on https://github.com/pyamg/pyamg/blob/v5.3.0/setup.cfg.

repos/spack_repo/builtin/packages/py_pyamg/package.py
```diff
@@ -26,6 +27,7 @@ class PyPyamg(PythonPackage):
...
     depends_on("py-numpy@1.7:", type=("build", "run"))
     depends_on("py-scipy@0.12:", type=("build", "run"))
```
tldahlgren, line 29: I don't see a version bound in https://github.com/pyamg/pyamg/blob/v5.3.0/setup.cfg for `py-numpy` but there is one for spicy: `py-scipy@1.11:`.

repos/spack_repo/builtin/packages/py_pyamg/package.py
tldahlgren: Looks like `5.0.0` had a minimum of `python@3.7:` https://github.com/pyamg/pyamg/blob/v5.0.0/setup.cfg so perhaps the version constraints should be considered for at least python in older versions of this package.

repos/spack_repo/builtin/packages/py_pyamg/package.py
tldahlgren: Ah, the minimum version for `py-numpy` is shown in the setup for `5.0.0`.

repos/spack_repo/builtin/packages/py_pyamg/package.py
tldahlgren: Looks like it should be `py-setuptools-scm@7:+toml` for `5.0.0` based on https://github.com/pyamg/pyamg/blob/v5.0.0/pyproject.toml.

tldahlgren (changes requested):
Confirmed the sha256 but there are some additions/adjustments needed to the dependency version constraints.

repos/spack_repo/builtin/packages/py_pyamg/package.py
garth-wells: I doubt very much that the hard version dependency is required; I can install and run just fine with the looser versioning. I've asked the question at https://github.com/pyamg/pyamg/issues/454.

The hard version dependency can make concretising a pain.

repos/spack_repo/builtin/packages/py_pyamg/package.py
tldahlgren: Will defer the recommendation for someone to put in a follow-on PR.

repos/spack_repo/builtin/packages/py_pyamg/package.py
tldahlgren: Will ignore this recommendation as well for the time being.

tldahlgren (approved):
IMO Needs to be consistent but will approve anyway since 1) comment says builds fine and 2) this has been open without maintainer feedback for a long time.

PR #1241, opened 2025-08-29
Added gcc conflict for py-numpy

emwjacobson:
Numpy >=2.3.0 now requires GCC >= 9.3.0
https://numpy.org/doc/stable/release/2.3.0-notes.html#the-minimum-supported-gcc-version-is-now-9-3-0

Also mentioning @rgommers and @adamjstewart as they are the listed maintainers.


rgommers (approved):
That looks right, thanks @emwjacobson!

One of these days we'll capture this in metadata rather than release notes, so we get these requirements first-time right ....

adamjstewart (approved):

PR #1245, opened 2025-08-30
go-sh: add sh package (includes shfmt and gosh)

mcmehrtens:
* adds a package for [sh](https://github.com/mvdan/sh), the software package housing `shfmt` and `gosh`
* package is named `go-sh` to avoid potential confusion with the Bourne shell
* `go-sh` creates two possible binaries: `shfmt` and `gosh`. I've dealt with this by adding variants for each.
  * To try and set sensible defaults, I've set the `shfmt` variant to default to true, since this seems to be the more widely used tool

### DIFF
--- a/repos/spack_repo/builtin/packages/go_sh/package.py
+++ b/repos/spack_repo/builtin/packages/go_sh/package.py
@@ -7,7 +7,7 @@ from spack_repo.builtin.build_systems.go import GoPackage
 from spack.package import *
 
 
-class Sh(GoPackage):
+class GoSh(GoPackage):
     """A shell parser, formatter, and interpreter. Supports POSIX
     Shell, Bash, and mksh. Requires Go 1.23 or later."""
 
--- a/repos/spack_repo/builtin/packages/go_sh/package.py
+++ b/repos/spack_repo/builtin/packages/go_sh/package.py
@@ -25,11 +25,14 @@ class GoSh(GoPackage):
     variant("gosh", default=False, description="Build and install gosh")
     conflicts("~shfmt~gosh", msg="One of shfmt or gosh must be specified")
 
-    sanity_check_is_file = []
-    with when("+shfmt"):
-        sanity_check_is_file.append(join_path("bin", "shfmt"))
-    with when("+gosh"):
-        sanity_check_is_file.append(join_path("bin", "gosh"))
+    @property
+    def sanity_check_is_file(self):
+        files = []
+        if self.spec.satisfies("+shfmt"):
+            files.append(join_path("bin", "shfmt"))
+        if self.spec.satisfies("+gosh"):
+            files.append(join_path("bin", "gosh"))
+        return files
 
     def build(self, spec: Spec, prefix: Prefix) -> None:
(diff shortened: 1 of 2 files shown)


haampie:
I don't think this is a good name, given `/bin/sh`. Can you add some prefix, maybe `go-sh`? Or call it `shfmt`?

mcmehrtens:
> I don't think this is a good name, given `/bin/sh`. Can you add some prefix, maybe `go-sh`? Or call it `shfmt`?

@haampie, I addressed your comment in 2a356b28348ced3af7a75cd123fa92284cd0cccb.

I would vote on `go-sh`. If we went with `shfmt` as the package name and the user installed with `shfmt~shfmt+gosh`, it might seem a bit strange. It reads like, "`shfmt` is installed without `shfmt`."

Let me know if I should make any other changes :)

repos/spack_repo/builtin/packages/go_sh/package.py
```diff
@@ -0,0 +1,57 @@
...
+
+    variant("shfmt", default=True, description="Build and install shfmt")
+    variant("gosh", default=False, description="Build and install gosh")
+    conflicts("~shfmt~gosh", msg="One of shfmt or gosh must be specified")
+
+    sanity_check_is_file = []
+    with when("+shfmt"):
+        sanity_check_is_file.append(join_path("bin", "shfmt"))
+    with when("+gosh"):
+        sanity_check_is_file.append(join_path("bin", "gosh"))
```
haampie: You cannot make class level code run "lazily" like that, the with body will always execute.

You'll have to do something like

```suggestion
    @property
    def sanity_check_is_file(self):
        files = []
        if self.spec.satisfies("+shfmt"):
            files.append(join_path("bin", "shfmt"))
        if self.spec.satisfies("+gosh"):
            files.append(join_path("bin", "gosh"))
        return files
```

repos/spack_repo/builtin/packages/go_sh/package.py
```diff
@@ -0,0 +1,57 @@
...
+    maintainers("mcmehrtens")
+    license("BSD-3-Clause", checked_by="mcmehrtens")
+
+    version("3.12.0", sha256="ac15f42feeba55af29bd07698a881deebed1cd07e937effe140d9300e79d5ceb")
+
+    depends_on("go@1.23:", type="build", when="@3.12.0:")
+
+    variant("shfmt", default=True, description="Build and install shfmt")
+    variant("gosh", default=False, description="Build and install gosh")
+    conflicts("~shfmt~gosh", msg="One of shfmt or gosh must be specified")
```
haampie, line 26: why not both?

repos/spack_repo/builtin/packages/go_sh/package.py
mcmehrtens: Shoot, I don't know how I missed that. Thank you! I applied your suggestion (1b60cd8f89c22e9899ec63744fce1718214dedf9).

haampie (approved):

PR #1246, opened 2025-08-31
hpctoolkit: add 2025.0.0 and update docs dependencies

blue42u:
- Add `@2025.0.0`: https://gitlab.com/hpctoolkit/hpctoolkit/-/releases/2025.0.0
- Update `+docs` dependencies to match accepted versions and add `py-sphinx-book-theme`

CC maintainer: @mwkrentel

### DIFF
--- a/repos/spack_repo/builtin/packages/hpctoolkit/package.py
+++ b/repos/spack_repo/builtin/packages/hpctoolkit/package.py
@@ -34,6 +34,7 @@ class Hpctoolkit(AutotoolsPackage, MesonPackage):
 
     version("develop", branch="develop")
     version("2025.0.stable", branch="release/2025.0")
+    version("2025.0.0", tag="2025.0.0", commit="9d0878a37ff3273ebebd1559780557de7a8a05cb")
     version("2024.01.stable", branch="release/2024.01")
     version("2024.01.1", tag="2024.01.1", commit="0672b9a9a2a1e3846c5e2059fb73a07a129f22cd")
     version("2023.08.stable", branch="release/2023.08")
--- a/repos/spack_repo/builtin/packages/hpctoolkit/package.py
+++ b/repos/spack_repo/builtin/packages/hpctoolkit/package.py
@@ -203,8 +203,9 @@ class Hpctoolkit(AutotoolsPackage, MesonPackage):
     depends_on("zlib+shared", when="^[virtuals=zlib-api] zlib")
 
     depends_on("py-docutils", type="build", when="@2025:")
-    depends_on("py-sphinx", type="build", when="+docs")
-    depends_on("py-myst-parser@0.19:", type="build", when="+docs")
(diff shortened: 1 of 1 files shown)


mwkrentel (approved):
I approve, but one of the CI tests fails on detecting hip.

mwkrentel:
I'm trying to figure out if the CI failure is real or spurious.
I suspect real.

I can load a `rocm` module, run `spack external find` to create an
`externals` entry for `hip`, etc, and then `hpctoolkit +rocm` builds
cleanly.  That's using the AMD supplied ROCm module.

The other test is using a spack-built `hip`, `hsa-rocr-dev`, etc.
This is where CI is failing.  Unfortunately, I can't build `hip`, etc
due to dependency failures.  For example, `mesa` fails with:
```
==> Error: ProcessError: /usr/bin/llvm-config: No such file or directory
    Command: '/usr/bin/llvm-config' '--has-rtti'
```
So, I can't test if `hpctoolkit +rocm` builds with spack-built rocm
pieces due to dependency failures.  Sigh.  I guess I'll have to set up
a build cache for them.

We had one theory that the rocm packages needed a dependency on
`rocm-device-libs`.  I think it would be worth trying to add that, see
if the CI succeeds.

blue42u:
@mwkrentel FYI `hip` et al. build just fine in the Spack containers, I've confirmed in `ghcr.io/spack/ubuntu-noble:1.0.1`. Do note it takes a while to build from scratch, it has to build 3 different copies of LLVM (`llvm`, `llvm-amdgpu`, and `rust`). After building the stack, I was able to confirm that the missing CMake script that fails the build in CI is available in `llvm-amdgpu +rocm-device-libs`, which is a link-type dependency of `hip`.

The root problem is that our dependency on `hip` is of build-type. Spack doesn't expose indirect dependencies from `hip` to our build, only link-type indirect dependencies are exposed. In `2024.01.1` and before this dependency type is necessary to keep Spack from embedding RPATHs that may conflict with the application's library paths. As of `@2025.0.0` we now directly link against the ROCm libraries and use `dlopen` to avoid conflicting library paths, *but* there are a few libraries (unrelated to ROCm support) that can't use the `dlopen` trick and can cause problems if RPATHs are embedded in.

I think the best approach here is: (1) upgrade `hip` et al. to link-type dependencies, and (2) fix our build system to strip/trim RPATHs for the afflicted libraries, which will likely come out as `@2025.0.1` (to avoid a re-release). I believe the RPATHs issue could extend to non-Spack builds as well, so IMHO delegating the responsibility to our build system is appropriate here, even if it means this PR (may) need to be on hold for a bit.\
@mwkrentel does this plan seem reasonable?

blue42u:
This PR has been superceded by https://github.com/spack/spack-packages/pull/1390. That PR implements the changes briefly discussed in my comment above.

PR #1252, opened 2025-09-01
py-gosam: Depend on setuptools since distutils is removed in python 3.13

tmadlener:
By default this builds with distutils, but that has been removed in python 3.12. If that doesn't work it falls back to setuptools, but that needs to be in the build dependencies for that to work.

Without these changes, I run into the following build error:

```
==> py-gosam: Executing phase: 'install'
==> [2025-08-29-21:18:05.277776] '/home/madlener/work/.spack/spackages/skylake-ubuntu24.04-nonenone/python-venv/1.0-hkozj5/bin/python3' '-s' 'setup.py' '--no-user-cfg' 'build'
  File "/home/madlener/work/.spack/spack-stage/spack-stage-py-gosam-2.1.2-oxngbfo426op5vg3knokjv3oht4gmemp/spack-src/setup.py", line 5, in <module>
    from distutils.core import setup
ModuleNotFoundError: No module named 'distutils'

During handling of the above exception, another exception occurred:

  File "/home/madlener/work/.spack/spack-stage/spack-stage-py-gosam-2.1.2-oxngbfo426op5vg3knokjv3oht4gmemp/spack-src/setup.py", line 14, in <module>
    from setuptools import setup
ModuleNotFoundError: No module named 'setuptools'
```

### DIFF
--- a/repos/spack_repo/builtin/packages/py_gosam/package.py
+++ b/repos/spack_repo/builtin/packages/py_gosam/package.py
@@ -40,6 +40,7 @@ class PyGosam(Package):
     depends_on("qgraf", type="run")
     depends_on("gosam-contrib", type="link")
     depends_on("python@3:", type=("build", "run"))
+    depends_on("py-setuptools", type="build", when="^python@3.12:")
 
     def setup_run_environment(self, env: EnvironmentModifications) -> None:
         gosam_contrib_lib_dir = self.spec["gosam-contrib"].prefix.lib
--- a/repos/spack_repo/builtin/packages/py_gosam/package.py
+++ b/repos/spack_repo/builtin/packages/py_gosam/package.py
@@ -40,7 +40,7 @@ class PyGosam(PythonPackage):
     depends_on("qgraf", type="run")
     depends_on("gosam-contrib", type="link")
     depends_on("python@3:", type=("build", "run"))
-    depends_on("py-setuptools", type="build", when="^python@3.12:")
+    depends_on("py-setuptools", type="build")
 
     def setup_run_environment(self, env: EnvironmentModifications) -> None:
         gosam_contrib_lib_dir = self.spec["gosam-contrib"].prefix.lib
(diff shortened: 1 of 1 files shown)


tmadlener:
@jmcarcell, I just realized you had the exact same change for Key4hep already. Did you ever try to build with an older version of python for this version to verify that this doesn't break there?

jmcarcell:
No, but also I don't think they are completely equivalent since the installation folders are different. The `share` folder is installed in a weird path now.

wdconinc:
Still appears to fail at https://gitlab.spack.io/spack/spack-packages/-/jobs/17971401#377, in particular:
```
pip._vendor.pyproject_hooks._impl.BackendUnavailable: Cannot import 'setuptools.build_meta'
```
with `python@3.11.11` and (therefore) no `py-setuptools`.

repos/spack_repo/builtin/packages/py_gosam/package.py
```diff
@@ -40,11 +40,8 @@ class PyGosam(Package):
     depends_on("qgraf", type="run")
     depends_on("gosam-contrib", type="link")
     depends_on("python@3:", type=("build", "run"))
+    depends_on("py-setuptools", type="build", when="^python@3.12:")
```
tmadlener:
```suggestion
    depends_on("py-setuptools", type="build")
```

So this then @wdconinc?

wdconinc: I think so? There might be some version limits. I think py-setuptools only started to provide distutils with something like v74 or so. Can't find it from my phone though.

adamjstewart: Pip has always silently replaced distutils imports with setuptools. We should always depend on setuptools.

wdconinc: Merged on the expert recommendation of @adamjstewart :-)

wdconinc (approved):
Ignore the RFR @adamjstewart, this isn't my PR so I can approve.

tmadlener:
Thanks :)

PR #1253, opened 2025-09-01
Add library `conversion` to boost

fbarbari:
@hainest

### DIFF
--- a/repos/spack_repo/builtin/packages/boost/package.py
+++ b/repos/spack_repo/builtin/packages/boost/package.py
@@ -123,7 +123,7 @@ class Boost(Package):
         "container",
         "context",
         "contract",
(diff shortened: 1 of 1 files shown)


hainest (changes requested):
`numeric` has only been buildable since `@1.87.0:`, so it also needs an entry in `all_libs_opts`.

hainest (approved):
LGTM. Have you tried building against the latest version?

fbarbari:
Yes, I just did it with spack `1.1.0.dev0 (a4c6825ef2575c751e32d5936ae5657fea7876b0)` and spec `boost@1.88.0 +conversion` and also in #1255.

tldahlgren (approved):

PR #1254, opened 2025-09-01
oneapi: only download resources for variants when variant enabled

becker33:
Currently, `intel-oneapi-compilers` unconditionally requires resources for the nvidia and amd plugins, despite the fact that those are variants on the package.

This PR fixes this by making the resource specifications conditional on the variants.


becker33:
@hfp this might be of interest to you.

alalazo (approved):

PR #1256, opened 2025-09-01
BWA: new versions and a patch fix

amd-toolchain-support:
Change in this request includes following updates to BWA spack recipe:
- Adding two latest versions: `0.7.19` and `0.7.18`
- Redo a patch by using the original fix from application source https://github.com/lh3/bwa/pull/267:
      - Using a github link from bwa source repo instead of change using filter_file
      - Extending to compilers other than gcc
      - Restrict the patch for applicable versions only.


tldahlgren (approved):
Confirmed the sha256s. Nice to see the filter file replaced with a patch.

PR #1259, opened 2025-09-02
Add m4 as build-time dependency for flint (required by configure)

d-torrance:
Currently, the flint build fails with the following error:

```
  >> 133    checking for suitable m4... configure: error: No usable m4 in $PA
            TH or /usr/5bin (see config.log for reasons).
```

Adding m4 as a build dependency fixes the issue.


tldahlgren (approved):
Confirmed dependency in configure.

PR #1260, opened 2025-09-02
fplll: Add C compiler as a dependency (otherwise configure fails)

d-torrance:
Without this dependency, we get (from `config.log`):

```
configure:4631: checking whether the C compiler works
configure:4653: gcc    conftest.c  >&5
[spack cc]: Error: SPACK_CC_* variables not set: this usually means a missing `depends_on("c", type="build")` in package.py
configure:4657: $? = 1
configure:4697: result: no
```

Also add latest version (5.5.0).

### DIFF
--- a/repos/spack_repo/builtin/packages/fplll/package.py
+++ b/repos/spack_repo/builtin/packages/fplll/package.py
@@ -17,6 +17,7 @@ class Fplll(AutotoolsPackage):
 
     license("LGPL-2.1-or-later")
 
+    version("5.5.0", sha256="f0af6bdd0ebd5871e87ff3ef7737cb5360b1e38181a4e5a8c1236f3476fec3b2")
     version("5.4.4", sha256="0fd9d378f04ff886d8864728baf5d90b8b0b82c1e541e92550644fb54f75691d")
     version("5.4.1", sha256="7bd887957173aa592091772c1c36f6aa606b3b2ace0d14e2c26c7463dcf2deb7")
     version("5.4.0", sha256="fe192a65a56439b098e26e3b7ee224dda7c2c73a58f36ef2cc6f9185ae8c482b")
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/fplll/package.py
```diff
@@ -29,6 +30,7 @@ class Fplll(AutotoolsPackage):
     version("5.1.0", sha256="58175c54cc92752576a64361c73e4ea7797fc18fb703b3f22c7570a09075486f")
     version("5.0.3", sha256="d2b11b7dcb26c30ac1aab9ff75aca9b3dd6e0b0b40c382af16017a717dfe05c2")
 
+    depends_on("c", type="build")
     depends_on("cxx", type="build")  # generated
```
tldahlgren: Can you confirm `cxx` is a dependency and, if so, remove `# generated` comment?

tldahlgren (changes requested):
Confirmed the sha256. Have  minor request w.r.t. the `cxx` dependency.

repos/spack_repo/builtin/packages/fplll/package.py
d-torrance: Yes, it's definitely a dependency!  Without it, we get:

```
     102    checking whether g++ supports C++11 features with -std=c++11... no
     103    checking whether g++ supports C++11 features with +std=c++11... no
     104    checking whether g++ supports C++11 features with -h std=c++11... n
            o
     105    checking whether g++ supports C++11 features with -std=c++0x... no
     106    checking whether g++ supports C++11 features with +std=c++0x... no
     107    checking whether g++ supports C++11 features with -h std=c++0x... n
            o
  >> 108    configure: error: *** A compiler with support for C++11 language fe
            atures is required.
```

Comment removed

tldahlgren (approved):
Thanks!

PR #1261, opened 2025-09-02
(py-)fenics-dolfinx: build and dependency updates

garth-wells:
- Add dependency on C compiler to py-fenics-dolfinx. This is because HDF5 and ADIOS2 use a C compiler in their CMake config files. Fixes a configure stage failure.
- Expose custom (py)-fenics-dolfinx CMake build type variants.
- Parallel build of py-fenics-dolfinx using `make_jobs` processes.
- Verbose build of C++ bindings for py-fenics-dolfinx.


tldahlgren (approved):
LGTM. Do wonder why the conflicts are being removed from `fenics-dolfinx` but will approve anyway.

PR #1262, opened 2025-09-02
(py-)-fenics-*: build updates

garth-wells:
- Expose custom CMake build type variants.
- Add support for parallel builds of the C++/Python binding code.
- Make builds of C++ code managed by scikit-core-build verbose.
- Refactor some tight dependencies into loops over versions.


tldahlgren (approved):
LGTM

PR #1263, opened 2025-09-02
rocm: add missing rocm tags and improve external detection

afzpatel:
Add `rocm` tags wherever missing.
Add `determine_version` methods to various packages so they can be found externally.

### DIFF
--- a/repos/spack_repo/builtin/packages/rocal/package.py
+++ b/repos/spack_repo/builtin/packages/rocal/package.py
@@ -13,7 +13,7 @@ class Rocal(CMakePackage):
 
     homepage = "https://github.com/ROCm/rocAL"
     url = "https://github.com/ROCm/rocAL/archive/refs/tags/rocm-6.4.3.tar.gz"
-    tag=["rocm"]
+    tag = ["rocm"]
 
     maintainers("afzpatel", "srekolam", "renjithravindrankannath")
 
--- a/repos/spack_repo/builtin/packages/rocal/package.py
+++ b/repos/spack_repo/builtin/packages/rocal/package.py
@@ -13,7 +13,7 @@ class Rocal(CMakePackage):
 
     homepage = "https://github.com/ROCm/rocAL"
     url = "https://github.com/ROCm/rocAL/archive/refs/tags/rocm-6.4.3.tar.gz"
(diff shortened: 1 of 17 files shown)


repos/spack_repo/builtin/packages/rocal/package.py
```diff
@@ -13,6 +13,7 @@ class Rocal(CMakePackage):
 
     homepage = "https://github.com/ROCm/rocAL"
     url = "https://github.com/ROCm/rocAL/archive/refs/tags/rocm-6.4.3.tar.gz"
+    tag = ["rocm"]
```
srekolam: i think rocal installs something like librocal.so.2.2.0 . can you check ?

repos/spack_repo/builtin/packages/rocalution/package.py
```diff
@@ -25,7 +25,7 @@ class Rocalution(CMakePackage):
     tags = ["rocm"]
 
     maintainers("cgmb", "srekolam", "renjithravindrankannath", "afzpatel")
-    libraries = ["librocalution_hip"]
+    libraries = ["librocalution"]
```
srekolam, line 28: just see determine_version(cls, lib):  not implemented for this .

srekolam (changes requested):
for rocal, i think there is librocal.so.* . rocalution too, missing one fn

repos/spack_repo/builtin/packages/rocalution/package.py
afzpatel: When I tried `spack external find rocalution`, I wasn't able to get the version information using the `librocalution_hip` library but when I switched to `librocalution` it worked

repos/spack_repo/builtin/packages/rocal/package.py
afzpatel: Yes it does install librocal.so.2.2.0 but I didn't add it because the library doesn't display the version number.

renjithravindrankannath (approved):
LGTM

srekolam (approved):

afzpatel:
@tldahlgren could you please review this pr?

tldahlgren (approved):

PR #1264, opened 2025-09-02
new package: py-textual-plotext

afzpatel:
add package py_textual_plotext that will be a dependency for rocprofiler-compute v7.0. Also add new versions py_textual.

### DIFF
--- a/repos/spack_repo/builtin/packages/py_textual/package.py
+++ b/repos/spack_repo/builtin/packages/py_textual/package.py
@@ -14,6 +14,12 @@ class PyTextual(PythonPackage):
     homepage = "https://github.com/Textualize/textual"
     pypi = "textual/textual-0.47.1.tar.gz"
 
+    version("5.3.0", sha256="1b6128b339adef2e298cc23ab4777180443240ece5c232f29b22960efd658d4d")
+    version("5.2.0", sha256="6d4a77fddde7aa659acc25c3e14ed27ff75e125c6730c6f57263cf91b60e0d50")
+    version("5.1.1", sha256="8df8830155046660592c8926e18d43c712dfe4e8a1351a98b1138f9b64ec14dc")
+    version("5.1.0", sha256="28e3833c8bef2319ab09483077c9812da422cecdeaa673319ba00f0523921e67")
+    version("5.0.1", sha256="c6e20489ee585ec3fa43b011aa575f52e4fafad550e040bff9f53a464897feb6")
+    version("5.0.0", sha256="44f507a1e264bab753b436c55245d8a957c348b3cada37f12a7782b5ccfbb1da")
     version("0.47.1", sha256="4b82e317884bb1092f693f474c319ceb068b5a0b128b121f1aa53a2d48b4b80c")
 
     depends_on("python@3.8:3", type=("build", "run"))
--- a/repos/spack_repo/builtin/packages/py_textual/package.py
+++ b/repos/spack_repo/builtin/packages/py_textual/package.py
@@ -15,11 +15,6 @@ class PyTextual(PythonPackage):
     pypi = "textual/textual-0.47.1.tar.gz"
 
     version("5.3.0", sha256="1b6128b339adef2e298cc23ab4777180443240ece5c232f29b22960efd658d4d")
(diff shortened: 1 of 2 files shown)


srekolam (approved):

repos/spack_repo/builtin/packages/py_textual/package.py
```diff
@@ -14,6 +14,12 @@ class PyTextual(PythonPackage):
     homepage = "https://github.com/Textualize/textual"
     pypi = "textual/textual-0.47.1.tar.gz"
 
+    version("5.3.0", sha256="1b6128b339adef2e298cc23ab4777180443240ece5c232f29b22960efd658d4d")
```
tldahlgren, line 17: If I'm reading the `pyproject.toml` correctly for this version, `py-platformdirs@3.6:4` is a missing dependency (for at least `5.2.0`, `5.3.0`).

repos/spack_repo/builtin/packages/py_textual_plotext/package.py
```diff
@@ -0,0 +1,24 @@
...
+
+    homepage = "https://github.com/Textualize/textual-plotext"
+    pypi = "textual-plotext/textual_plotext-1.0.1.tar.gz"
+
+    version("1.0.1", sha256="836f53a3316756609e194129a35c2875638e7958c261f541e0a794f7c98011be")
+    version("1.0.0", sha256="6e8e608ca9da52b4f171fa4adb0e80cbcd8e668f1334e9e4d7fbfd064d393a4f")
+
+    depends_on("python@3.8:3")
+    depends_on("py-poetry-core")
+    depends_on("py-textual@5:")
```
tldahlgren, line 22: How do you know you can use newer versions?

According to the `pyproject.toml` and what I've ready about `"^0.86.2"`, you can theoretically only count on `py-textual@0.86.2:0.86` (for `1.0.0` though that package doesn't include that version). This range appears to be fine for `1.0.1` so perhaps it's a moot question.

repos/spack_repo/builtin/packages/py_textual_plotext/package.py
```diff
@@ -0,0 +1,24 @@
...
+
+class PyTextualPlotext(PythonPackage):
+    """A Textual widget wrapper library for Plotext"""
+
+    homepage = "https://github.com/Textualize/textual-plotext"
+    pypi = "textual-plotext/textual_plotext-1.0.1.tar.gz"
+
+    version("1.0.1", sha256="836f53a3316756609e194129a35c2875638e7958c261f541e0a794f7c98011be")
+    version("1.0.0", sha256="6e8e608ca9da52b4f171fa4adb0e80cbcd8e668f1334e9e4d7fbfd064d393a4f")
+
```
tldahlgren, line 19: If I'm reading the `pyproject.toml` files correctly, there's a missing dependency on `py-plotext@5.2.8:5`.

tldahlgren (changes requested):
Confirmed all of the checksums and nearly all dependencies. I think there's a missing dependency for each of the packages.

Odd that `textual-plotext` doesn't have official tags or releases for the two new versions on `pypi.org`.

Are there meaningful differences in the `py-textual` versions that warrant including so many new ones? Also, please confirm you have at least successfully built each version being added to each package.

repos/spack_repo/builtin/packages/py_textual/package.py
afzpatel: Added dependency for py-platformdirs@3.6:4 and removed additional versions. I've only left 5.3.0.

repos/spack_repo/builtin/packages/py_textual_plotext/package.py
afzpatel: When I checked the dependency for py-textual, it was >=0.86.2. Since I've only added version 5, I made the dependency `py-textual@5:"`

repos/spack_repo/builtin/packages/py_textual_plotext/package.py
afzpatel: Thanks, I've added the dependency

tldahlgren (approved):
Thanks.

PR #1265, opened 2025-09-02
ci: add spack-reviewers to add maintainers to PRs

alecbcs:
Building on the labeler work, this PR adds an extensible script to automatically add package maintainers as reviewers to pull requests. 

This PR is left very minimal to allow this to merge quickly before adding additional advanced functionality. 
(E.g. adding the users to the org on merge to develop, adding build system maintainers on `new-package` pull requests, etc...)

### DIFF
--- a/.github/workflows/bin/spack-reviewers.py
+++ b/.github/workflows/bin/spack-reviewers.py
@@ -8,7 +8,7 @@ import subprocess
 
 import requests
 
-IS_PACKAGE = re.compile(r"repos/spack_repo/builtin/packages/([^/]+)/package.py$")
+IS_PACKAGE_CHANGE = re.compile(r"repos/spack_repo/builtin/packages/([^/]+)/.*$")
 
 
 def main():
@@ -29,7 +29,7 @@ def main():
     maintainers = set()
 
     for file in pull_request_files.json():
-        if match := IS_PACKAGE.match(file["filename"]):
+        if match := IS_PACKAGE_CHANGE.match(file["filename"]):
             package = match.group(1)
 
             # add maintainers from packages here
@@ -47,7 +47,9 @@ def main():
 
     added_reviewers = reviewers - existing_reviewers
     if added_reviewers:
-        print(f"[PR #{pr_number}]: adding reviewer(s): @{' @'.join(added_reviewers)}")
+        print(f"[PR #{pr_number}]: adding reviewer(s):")
+        for reviewer in sorted(added_reviewers):
+            print(f"    @{reviewer}")
 
     if token:
         resp = requests.post(
--- a/.github/workflows/bin/spack-reviewers.py
+++ b/.github/workflows/bin/spack-reviewers.py
@@ -42,12 +42,12 @@ def main():
     reviewers = (maintainers | existing_reviewers) - {author}
 
     if existing_reviewers == reviewers:
-        print(f"[PR #{pr_number}]: labels already up-to-date")
+        print(f"[PR #{pr_number}]: reviewers already up-to-date")
         return
 
     added_reviewers = reviewers - existing_reviewers
     if added_reviewers:
-        print(f"[PR #{pr_number}]: Adding reviewer(s): @{' @'.join(added_reviewers)}")
+        print(f"[PR #{pr_number}]: adding reviewer(s): @{' @'.join(added_reviewers)}")
 
     if token:
         resp = requests.post(
--- a/.github/workflows/bin/spack-reviewers.py
+++ b/.github/workflows/bin/spack-reviewers.py
@@ -51,9 +51,7 @@ def main():
 
     if token:
         resp = requests.post(
-            f"{base_url}/requested_reviewers",
-            json={"reviewers": list(reviewers)},
-            headers=headers,
+            f"{base_url}/requested_reviewers", json={"reviewers": list(reviewers)}, headers=headers
         )
         # https://docs.github.com/en/rest/pulls/review-requests
         # endpoint may return status code 422 for users who are not collaborators
--- a/.github/workflows/bin/spack-reviewers.py
+++ b/.github/workflows/bin/spack-reviewers.py
@@ -1,0 +1,66 @@
+# Copyright Spack Project Developers. See COPYRIGHT file for details.
+#
+# SPDX-License-Identifier: (Apache-2.0 OR MIT)
+
+import os
+import re
+import subprocess
+
+import requests
+
+IS_PACKAGE = re.compile(r"repos/spack_repo/builtin/packages/([^/]+)/package.py$")
+
+
+def main():
+    repository = os.environ["GH_REPO"]
+    pr_number = os.environ["GH_PR_NUMBER"]
+    token = os.environ["GH_TOKEN"]
+
+    headers = {"Accept": "application/vnd.github+json"}
+    if token:
+        headers["Authorization"] = f"Bearer {token}"
+
+    base_url = f"https://api.github.com/repos/{repository}/pulls/{pr_number}"
+    pull_request = requests.get(base_url, headers=headers).json()
+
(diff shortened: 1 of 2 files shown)


.github/workflows/bin/spack-reviewers.py
```diff
@@ -0,0 +1,64 @@
...
+            result = subprocess.run(
+                ["spack", "maintainers", package], capture_output=True, text=True, check=False
+            )
+            maintainers.update(result.stdout.split())
+
+    author = pull_request["user"]["login"]
+    reviewers = (maintainers | existing_reviewers) - {author}
+
+    if existing_reviewers == reviewers:
+        print(f"[PR #{pr_number}]: labels already up-to-date")
```
kwryankrattiger:
```suggestion
        print(f"[PR #{pr_number}]: reviewers already up-to-date")
```

.github/workflows/bin/spack-reviewers.py
```diff
@@ -0,0 +1,64 @@
...
+        return
+
+    added_reviewers = reviewers - existing_reviewers
+    if added_reviewers:
+        print(f"[PR #{pr_number}]: Adding reviewer(s): @{' @'.join(added_reviewers)}")
```
kwryankrattiger: probably easier to read if this prints per-line. For spanning deprecation type PRs this list will become unsearchable.

.github/workflows/triage.yml
```diff
@@ -24,11 +24,14 @@ jobs:
           ref: develop # DO NOT CHANGE
 
       - uses: actions/setup-python@a26af69be951a213d495a4c3e4e4022e16d87065
+      - uses: ./.github/actions/checkout-spack
 
       - name: Install Python dependencies
         run: pip install -r .github/workflows/requirements/triage/requirements.txt
 
-      - run: python3 .github/workflows/bin/spack-labeler.py
```
kwryankrattiger, line 31: Maybe split these into two steps with different names. Easier to search the logs when each function is  its own section.

kwryankrattiger (changes requested):

.github/workflows/triage.yml
```diff
@@ -24,11 +24,14 @@ jobs:
           ref: develop # DO NOT CHANGE
 
       - uses: actions/setup-python@a26af69be951a213d495a4c3e4e4022e16d87065
+      - uses: ./.github/actions/checkout-spack
```
kwryankrattiger, line 27: It may be good to explicitly configure this to point at the `spack/spack-packages@develop` rather than do the clone inside of the scripts.

.github/workflows/bin/spack-reviewers.py
```diff
@@ -0,0 +1,64 @@
...
+#
+# SPDX-License-Identifier: (Apache-2.0 OR MIT)
+
+import os
+import re
+import subprocess
+
+import requests
+
+IS_PACKAGE = re.compile(r"repos/spack_repo/builtin/packages/([^/]+)/package.py$")
```
adamjstewart: Not sure if we want/need to limit ourselves to package.py, if any files like patches are modified, we should notify the maintainers.

adamjstewart (commented):
Still missing a lot of basic features from the old spackbot implementation:

* ask people to maintain unmaintained packages
* use comments for people without reviewer access
* invite people to get reviewer access
* get maintainers from both develop and the PR branch in case of addition/removal/new package

I would rather get this merged than hold this up any longer, but these are all very high priority features we should add in the near future. I might be able to help add some of these if I'm ever less busy.

adamjstewart:
P.S. Once these features are added, there are a ton of "maintainer"-related bugs/feature requests in the spackbot repo. We should go through those and either move them to this repo or close them.

.github/workflows/bin/spack-reviewers.py
alecbcs: Fixed, I think right while you were doing your review haha

.github/workflows/bin/spack-reviewers.py
alecbcs: Modified this to print each reviewer on a new line and sort the reviewers so it's easier to inspect for large deprecation PRs.

.github/workflows/triage.yml
alecbcs: Good idea. I've split them into two steps.

.github/workflows/triage.yml
alecbcs: I agree it'd be good to be more explicit here, but we can't always clone develop since the latest Spack might not be compatible with the packages repository for a short time when we modify the API. This existing action reads the max Spack version from `.ci/`. I'd like to change this later on to use `spack/setup-spack` but we'll need to extract the commit from the config file and we'll need to create a new release of the action that supports v1.0 repositories.

.github/workflows/bin/spack-reviewers.py
alecbcs: Good idea, I've updated the code so we notify a maintainer if anything in the package's directory changes.

.github/workflows/triage.yml
kwryankrattiger: Right, the part I am talking about is for the packages repo, not spack. I agree we should not be grabbing the head of spack/spack develop.

This line where
```
${{ inputs.path }}/bin/spack repo set --destination "$REPOSITORY_PATH" --scope site builtin
```

For this check we should point at the develop of packages, not the PR, when querying maintainers. This ensures that people can't remove a maintainer without them at least getting a ping.

Then change woiuld be
```suggestion
      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683
        with:
            ref: develop
            fetch-depth: 1
            path: develop/
      - uses: ./.github/actions/checkout-spack
        with:
            builtin: develop/
```

alecbcs:
Just want to quickly follow up on a bit of @adamjstewart's points here even though they're not blockers for this PR to get people's ideas.

> * ask people to maintain unmaintained packages

This is easy I'll write this up as a simple follow on PR.

> * use comments for people without reviewer access

This is somewhat difficult because the `$GITHUB_TOKEN` doesn't have access to the private Spack organization data so for a lot of people who don't make their membership public it's hard to tell if they have triage permissions or not. I have an idea on how to get around this with a GitHub app, but I want to take more time to test out the security of that plan.

> * invite people to get reviewer access

Same issue as above, `$GITHUB_TOKEN` can't write to the Spack org but we might be able to get around this.

> * get maintainers from both develop and the PR branch in case of addition/removal/new package

This is tricky to get right on the security side. We absolutely *cannot* checkout the PR code in this workflow since that could give PRs the ability to run in a privileged context and modify core branches in the main Spack repository. We might be able to use the GitHub API to gather a diff of the changes in a PR and then add newly added maintainers by parsing the directive from a regular expression.

.github/workflows/triage.yml
alecbcs: What you're describing already happens in lines 22-24. We don't need to specify the path for either since they both default to the GitHub runner workspace.

kwryankrattiger: Oh duh, yea okay then I am fine on this point :)

kwryankrattiger:
> * get maintainers from both develop and the PR branch in case of addition/removal/new package

For this point, we are already getting the pull request files, which have a `patch` field that list the PRs changes to the file. This is what we used previously to extract added maintainers in the PR.

This was also used to determine if a new package was missing a maintainer and post a comment about the author adding themselves as a maintainer for that package.

kwryankrattiger (dismissed):

kwryankrattiger (approved):

kwryankrattiger:
Failure in gitlab-ci is on develop

alalazo:
This failure when adding a new package was reported on Slack: https://github.com/spack/spack-packages/actions/runs/17510261816/job/49740860323?pr=1231

PR #1269, opened 2025-09-03
py-executing: add v2.2.1

manuelakuhn:
https://github.com/alexmojaki/executing/tree/v2.2.1


tldahlgren (approved):
Confirmed sha256 and dependencies.

PR #1273, opened 2025-09-03
py-expandvars: add v1.1.1

manuelakuhn:
https://github.com/sayanarijit/expandvars/tree/v1.1.1

### DIFF
--- a/repos/spack_repo/builtin/packages/py_expandvars/package.py
+++ b/repos/spack_repo/builtin/packages/py_expandvars/package.py
@@ -15,6 +15,7 @@ class PyExpandvars(PythonPackage):
 
     license("MIT")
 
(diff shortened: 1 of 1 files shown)


tldahlgren (changes requested):
Confirmed the checksum.

I prefer to be explicit on the version of python (i.e., `python@3.6.2:` for the latest version and `python@3:` for the initial one) BUT I realize you would generally get it indirectly through `py-hatchling`. 

Let me know if you want to add the python dependencies.

manuelakuhn:
I also prefer to have explicit python versions as stated in the `pyproject.toml` because it makes maintenance much clearer but in the past it was often requested to remove them.
I have added them now.

tldahlgren (approved):
Thanks.

PR #1276, opened 2025-09-03
py-filelock: add v3.19.1 and py-hatch-vcs: add v0.5

manuelakuhn:
https://github.com/tox-dev/filelock/tree/3.19.1

`py-filelock` needs a newer version of `py-hatch-vcs`, so this updated as well.
`py-hatch-vcs` had a dependency to `py-hatchling@1.24.2:` in the recipe but no information where this came from. According to the [py-project.toml](https://github.com/ofek/hatch-vcs/blob/v0.5.0/pyproject.toml) it only depends on `py-hatchling@1.1`, so I removed the other one.

### DIFF
--- a/repos/spack_repo/builtin/packages/py_filelock/package.py
+++ b/repos/spack_repo/builtin/packages/py_filelock/package.py
@@ -19,6 +19,7 @@ class PyFilelock(PythonPackage):
 
     license("Unlicense")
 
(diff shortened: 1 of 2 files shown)


manuelakuhn:
@spackbot re-run pipeline

tldahlgren (approved):
Confirmed checksums and dependencies. 

FYI. `py-` could be dropped from the `py-filelock` homepage (https://github.com/tox-dev/filelock) but it does redirect correctly.

PR #1279, opened 2025-09-03
amd-aocl: remove virtual providers

G-Ragghianti:
The amd-aocl package requests the virtual providers blas, lapack, and scalapack, but the providers are hard-coded to be satisfied by the amd implementations of these libraries.  If the provider is hard-coded to a specific package, then the virtual provider (e.g. "blas") shouldn't be used.  One reason to remove the virtuals is that is prevents the use of some other virtual provider for blas by other packages within a unified environment.

### DIFF
--- a/repos/spack_repo/builtin/packages/amd_aocl/package.py
+++ b/repos/spack_repo/builtin/packages/amd_aocl/package.py
@@ -40,13 +40,10 @@ class AmdAocl(BundlePackage):
     depends_on("c", type="build")
     depends_on("cxx", type="build")
     depends_on("fortran", type="build")
-    depends_on("scalapack")
-    depends_on("lapack")
-    depends_on("blas")
 
-    requires("^[virtuals=scalapack] amdscalapack")
-    requires("^[virtuals=lapack] amdlibflame")
-    requires("^[virtuals=blas] amdblis")
+    requires("amdscalapack")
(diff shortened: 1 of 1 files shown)


tldahlgren:
@amd-toolchain-support Any objections to this PR being merged?

tldahlgren (approved):

alalazo:
> One reason to remove the virtuals is that is prevents the use of some other virtual provider for blas by other packages within a unified environment.

That might also be a reason to keep it, depending on the package. Would we ever be in a situation where using two `lapack` providers causes issues with `amd-aocl`? Also, is `amd-aocl` really depending on the `lapack` API?

G-Ragghianti:
> That might also be a reason to keep it, depending on the package. Would we ever be in a situation where using two `lapack` providers causes issues with `amd-aocl`? Also, is `amd-aocl` really depending on the `lapack` API?

That is a worthwhile perspective/consideration, but I can't speak directly to how the internals of aocl work. For my use case, it is less a matter of "using" two providers and more of just having two providers available in the environment.  The problem/conflict occurs when one of the providers is "used" by another package in the environment.  Otherwise, it is fine to have multiple provider packages. Maybe I'm slightly misusing unified environments, but there is no other way that I've found to avoid having multiple versions of the same package within an environment.

PR #1281, opened 2025-09-03
ack: add v2.28 and v3.9.0

KineticTheory:
Adding 2 new versions.  
* Added the last in the 2.X series.
* Added 3.9.  Starting with 3.x the install process is slightly different.


KineticTheory:
@rbberger For review when you have time...

rbberger (approved):

PR #1283, opened 2025-09-03
CMake: deprecate non latest patch versions pre 1.1

johnwparent:
I try to deprecate non latest CMake patch versions before minor releases. Trying to get ahead of 1.1.

Also getting a couple I apparently missed pre 1.0.


johnwparent:
@spackbot fix style

alalazo (approved):

alalazo:
`vtk` is failing on `develop` too.

PR #1287, opened 2025-09-04
gettext: provide additional linking information

kuberry:
- Adds `setup_dependent_build_environment` for gettext, which does not produce a pkgconfig file. 
- Adds to `ldflags` when using `+libxml2` in gettext, as gettext ignores libxml2's pkgconfig.
- Make python's dependency on xz consistent.

With these changes, it is possible to build `gettext` and all dependencies statically (`spack install gettext~~shared`)

### DIFF
--- a/repos/spack_repo/builtin/packages/gettext/package.py
+++ b/repos/spack_repo/builtin/packages/gettext/package.py
@@ -170,10 +170,32 @@ class Gettext(AutotoolsPackage, GNUMirrorPackage):
         )
         return libs
 
-    # gettext doesn't produce a pkg-config
     def setup_dependent_build_environment(self, env, dependent_spec):
-        env.append_flags(
-            "LDFLAGS",
-            "-L{0} ".format(dependent_spec["gettext"].prefix.lib)
-            + dependent_spec["gettext"].libs.link_flags,
-        )
+        # gettext doesn't produce a pkg-config, which means that static dependencies are not
+        # propagated to dependent packages
+
+        # either gettext is built static (~shared), in which case we gather all dependencies
+        # otherwise, we are building shared and only require static dependencies
+        if dependent_spec.satisfies("^gettext+shared"):
+            link_deps = dependent_spec["gettext"].dependencies(deptype="link")
+            for link_dep in link_deps:
+                if not (dependent_spec["gettext"].satisfies(fr"^{link_dep.name}+shared") or \
+                        dependent_spec["gettext"].satisfies(fr"^{link_dep.name} libs=shared")):
+                    env.append_flags(
+                        "LDFLAGS",
+                        "-L{0} ".format(dependent_spec[link_dep.name].prefix.lib)
+                        + dependent_spec[link_dep.name].libs.link_flags,
+                    )
+        else:
+            env.append_flags(
+                "LDFLAGS",
+                "-L{0} ".format(dependent_spec["gettext"].prefix.lib)
+                + dependent_spec["gettext"].libs.link_flags,
+            )
+            link_deps = dependent_spec["gettext"].dependencies(deptype="link")
+            for link_dep in link_deps:
+                env.append_flags(
+                    "LDFLAGS",
+                    "-L{0} ".format(dependent_spec[link_dep.name].prefix.lib)
+                    + dependent_spec[link_dep.name].libs.link_flags,
+                    )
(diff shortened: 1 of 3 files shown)


kuberry:
@spackbot fix style

repos/spack_repo/builtin/packages/gettext/package.py
```diff
@@ -166,3 +169,11 @@ def libs(self):
...
         )
         return libs
+
+    # gettext doesn't produce a pkg-config
+    def setup_dependent_build_environment(self, env, dependent_spec):
+        env.append_flags(
+            "LDFLAGS",
+            "-L{0} ".format(dependent_spec["gettext"].prefix.lib)
+            + dependent_spec["gettext"].libs.link_flags,
+        )
```
michaelkuhn: I'm not sure about these, should we really unconditionally link against these libraries if they are a dependency?
Also, should these be limited to `self.spec.satisfies("~shared")`? Doesn't Spack's compiler wrapper add `-L...` already?

michaelkuhn (changes requested):
When trying to build `gettext~shared ^libxml2~shared ^bzip2~shared ^zlib-ng~shared ^zstd libs:=static ^xz libs:=static ^libiconv libs:=static`, the build still fails with undefined references to `libiconv_*` (in `libxml2.a`).

kuberry:
Hello @michaelkuhn, thank you for having a look at this PR.

>>I'm not sure about these, should we really unconditionally link against these libraries if they are a dependency?
Also, should these be limited to self.spec.satisfies("~shared")? Doesn't Spack's compiler wrapper add -L... already?

I agree that we should guard with something like self.spec.satisfies("~shared"). Spack does add a `-L/path/to/gettext` and `-lasprintf -lgettextpo -lintl` to the `spec["gettext"].libs.link_flags`, however this is not sufficient to successfully build against if gettext is a dependency.

For instance, with libxml2 depending on libiconv, libxml2 creates a `lib/pkgconfig/libxml-2.0.pc` containing:
```
Libs: -L${libdir} -lxml2   -L/Users/user/spack_separate/opt/spack/darwin-m1/xz-5.6.3-4wjbmahzlxjh3jn5gla5ysvmzjsal6t7/lib -llzma  -L/Users/user/spack_separate/opt/spack/darwin-m1/libiconv-1.18-64udvsut4isv2tcyszncwo2tlb4uzpe5/lib -liconv 
```
This gives information on dependencies of libxml2 (lzma and iconv) as well as libxml2 itself.

Since `libiconv` and `gettext` do not produce pkgconfig files, I'm suggesting we make these additions to `setup_dependent_build_environment`. Without them, nothing can successfully build off of gettext if it has static dependencies. I came across a patch elsewhere to add a intl.pc.in https://lists.gnu.org/archive/html/bug-gettext/2019-10/msg00003.html, but I don't think that upstream packages would expect to find gettext with a pkgconfig file. I'll amend the PR with that `static` guard. Apologies for the delay, but I've been trying to find ways around using `setup_dependent_build_environment`.

repos/spack_repo/builtin/packages/python/package.py
```diff
@@ -144,7 +144,7 @@ class Python(Package):
         depends_on("gdbm", when="+dbm")  # alternatively ndbm or berkeley-db
         depends_on("zlib-api", when="+zlib")
         depends_on("bzip2", when="+bz2")
-        depends_on("xz libs=shared", when="+lzma")
+        depends_on("xz", when="+lzma")
```
kuberry, line 179: When `+lzma` is requested, `xz` is required regardless of whether it is shared. 

If this remained as-is and `xz libs=static` is built as a dependency of `gettext`, then `xz libs=static` symbols will be in `gettext` and a vendored `xz` will be built inside of `python`, potentially with different symbols and versions.  Removing `libs=shared` here ensures that these symbols and versions of `xz` are kept consistent.

repos/spack_repo/builtin/packages/gettext/package.py
kuberry: I've added a commit to handle both the `+shared` and `~shared` case for `gettext`. For `+shared`, only flags for static dependencies of `gettext` appended it `LDFLAGS` (which are needed by `gettext`'s dependents). For `~shared`, all dependencies of `gettext` are appended to `LDFLAGS`. This is because if `gettext` is built static, it no longer carries that dependency information otherwise. 

For `gettext +shared` with shared library dependencies, there will be no change to `LDFLAGS`.

kuberry:
@spackbot fix style

PR #1288, opened 2025-09-04
csdp: add 6.2.0 and generalize blas/lapack implementation

d-torrance:
Note that the license also changed to the Eclipse Public License 2.0.

### DIFF
--- a/repos/spack_repo/builtin/packages/csdp/package.py
+++ b/repos/spack_repo/builtin/packages/csdp/package.py
@@ -15,15 +15,31 @@ class Csdp(MakefilePackage):
     homepage = "https://projects.coin-or.org/Csdp"
     url = "https://www.coin-or.org/download/source/Csdp/Csdp-6.1.1.tgz"
 
-    license("CPL-1.0")
+    license("CPL-1.0", when="@:6.2.0", checked_by="d-torrance")
+    # license has been updated in git for future releases to EPL-2.0
 
+    version("6.2.0", sha256="7f202a15f33483ee205dcfbd0573fdbd74911604bb739a04f8baa35f8a055c5b")
     version("6.1.1", sha256="0558a46ac534e846bf866b76a9a44e8a854d84558efa50988ffc092f99a138b9")
 
-    depends_on("c", type="build")  # generated
+    depends_on("c", type="build")
 
-    depends_on("atlas")
+    depends_on("blas")
+    depends_on("lapack")
 
     def edit(self, spec, prefix):
         mkdirp(prefix.bin)
         makefile = FileFilter("Makefile")
         makefile.filter("/usr/local/bin", prefix.bin)
+        makefile.filter(r"^export LIBS.*$", "")  # use flag_handler instead
+
+    def flag_handler(self, name: str, flags: List[str]):
+        if name == "ldflags":
+            flags.extend(
+                [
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/csdp/package.py
```diff
@@ -15,15 +15,26 @@ class Csdp(MakefilePackage):
     homepage = "https://projects.coin-or.org/Csdp"
     url = "https://www.coin-or.org/download/source/Csdp/Csdp-6.1.1.tgz"
 
-    license("CPL-1.0")
+    license("EPL-2.0", checked_by="d-torrance")
```
alalazo: We should use the `when=` argument and a new directive for a change of license, unless it applies to all versions in Spack.

repos/spack_repo/builtin/packages/csdp/package.py
```diff
@@ -15,15 +15,26 @@ class Csdp(MakefilePackage):
...
         makefile.filter("/usr/local/bin", prefix.bin)
+
+    @property
+    def build_targets(self):
+        return [
+            "LIBS=-L../lib -lsdp "
+            f"{self.spec['blas'].libs.ld_flags} "
+            f"{self.spec['lapack'].libs.ld_flags} "
+            "-lm"
+        ]
```
alalazo: These are not build targets. We should use [flag_handler](https://spack.readthedocs.io/en/latest/packaging_guide_build.html#compiler-flags) for these kind of things.

alalazo (changes requested):

repos/spack_repo/builtin/packages/csdp/package.py
d-torrance: Makes sense!  My first attempt was actually incorrect anyway -- they've switched to EPL in git, but the 6.2.0 release was still CPL.

repos/spack_repo/builtin/packages/csdp/package.py
d-torrance: Ok cool!  I've updated it to use `flag_handler` instead (and also clear out the hardcoded `LIBS` in the Makefile).

repos/spack_repo/builtin/packages/csdp/package.py
```diff
@@ -15,15 +15,31 @@ class Csdp(MakefilePackage):
...
         mkdirp(prefix.bin)
         makefile = FileFilter("Makefile")
         makefile.filter("/usr/local/bin", prefix.bin)
+        makefile.filter(r"^export LIBS.*$", "")  # use flag_handler instead
+
+    def flag_handler(self, name: str, flags: List[str]):
+        if name == "ldflags":
+            flags.extend(
+                [
+                    "-L../lib -lsdp",
```
alalazo: Hmm, this line doesn't seem very portable. Why do we have a relative path, and which library is `sdp`?

alalazo (changes requested):

repos/spack_repo/builtin/packages/csdp/package.py
d-torrance: I've updated it to use an absolute path.  `sdp` is csdp's own static library that is built and then linked in with the executables.

alalazo (approved):

PR #1292, opened 2025-09-04
libtirpc: softlink include/tirpc/* back to include/

kftse-ust-hk:
Fix the issue that import `<rpc/*.h>` does not compile
because the include file is in `./include/tirpc/rpc/*`.


repos/spack_repo/builtin/packages/libtirpc/package.py
```diff
@@ -55,3 +57,15 @@ def configure_args(self):
         if self.spec.satisfies("@1.3.3 platform=darwin"):
             return ["--disable-gssapi"]
         return []
+
+    @run_after("install")
+    def softlink_includes(self):
+        """Libtirpc installs its headers in a subdirectory of include.
+        Softlink them to the include directory so that packages that
+        expect to find rpc/rpc.h in include can find it."""
+        mkdirp(self.prefix.include)
```
tldahlgren: This directory already exists so why run `mkdirp()`?

tldahlgren (commented):
Have a question.

repos/spack_repo/builtin/packages/libtirpc/package.py
kftse-ust-hk: right it is unnecessary to `mkdir -p`, just softlink `include/tirpc/*` to `include/*` should do.

bernhardkaindl:
@kftse-ust-hk, could you review https://github.com/spack/spack-packages/pull/1292#discussion_r3486190642?

PR #1297, opened 2025-09-04
py-formulaic: add v1.2.0 and py-wrapt: add v1.17.3

manuelakuhn:
https://github.com/matthewwardrop/formulaic/tree/v1.2.0

`py-formulaic` needs a newer versions of `py-wrapt`, so this was updated as well
https://github.com/GrahamDumpleton/wrapt/tree/1.17.3


alecbcs (approved):
Looks good to me. Thanks!

PR #1298, opened 2025-09-04
py-narwhals: add v2.3.0

manuelakuhn:
https://github.com/narwhals-dev/narwhals/tree/v2.3.0


alecbcs (approved):
Looks good to me. Thanks!

PR #1301, opened 2025-09-04
llvm: add conflict for gcc@15 and older llvm versions

manuelakuhn:
Add conflicts because older llvm versions do not build with `gcc@15`


haampie:
I believe you, but can you add the type of error or a link for context as a comment? Could help with workarounds in the future.

manuelakuhn:
I have added a line with error examples that I get when building `llvm@15` (needed for `py-numba`).

manuelakuhn:
ping @haampie

michaelkuhn (approved):
I was able to reproduce the problem on Rocky 9 with GCC 15.

PR #1304, opened 2025-09-04
ci: use internal Spack SDK to satisfy @haampie's need for speed

alecbcs:
I'd originally opted to use Spack's CLI since it's more stable and unlikely to change. However, @haampie brought up a decent point regarding PRs with bulk package updates and the overhead of each being in a python process.

### DIFF
--- a/.github/workflows/bin/spack-reviewers.py
+++ b/.github/workflows/bin/spack-reviewers.py
@@ -8,7 +8,7 @@ import subprocess
 
 import requests
 
-import spack
+import spack.repo
 
 IS_PACKAGE_CHANGE = re.compile(r"repos/spack_repo/builtin/packages/([^/]+)/.*$")
 
--- a/.github/workflows/bin/spack-reviewers.py
+++ b/.github/workflows/bin/spack-reviewers.py
@@ -4,7 +4,6 @@
 
 import os
 import re
-import subprocess
 
 import requests
 
--- a/.github/workflows/bin/spack-reviewers.py
+++ b/.github/workflows/bin/spack-reviewers.py
@@ -8,6 +8,8 @@ import subprocess
 
 import requests
 
+import spack
+
 IS_PACKAGE_CHANGE = re.compile(r"repos/spack_repo/builtin/packages/([^/]+)/.*$")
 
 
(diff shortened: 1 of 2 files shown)


.github/workflows/bin/spack-reviewers.py
```diff
@@ -8,6 +8,8 @@
 
 import requests
 
+import spack
```
haampie: import the module you're using

kwryankrattiger: Also, if we are going to be using spack modules, maybe run this under `spack-python` in CI as well.

alecbcs: Added both.

kwryankrattiger (approved):

.github/workflows/bin/spack-reviewers.py
```diff
@@ -33,10 +34,8 @@ def main():
             package = match.group(1)
 
             # add maintainers from packages here
-            result = subprocess.run(
-                ["spack", "maintainers", package], capture_output=True, text=True, check=False
-            )
-            maintainers.update(result.stdout.split())
+            pkg_maintainers = spack.repo.PATH.get_pkg_class(package).maintainers
```
tldahlgren, line 37: ~Unfortunately, you won't pick up maintainers from base classes anymore.~

Granted, there are so few in the base classes anymore (now only `RacketPackage` (9) and `RubyPackage` (18)).

~But I still think we should be providing the complete list.~

And after a bit of investigating/testing, I guess it still does. 🎊

tldahlgren (changes requested):
See comment.

.github/workflows/bin/spack-reviewers.py
tldahlgren: Should've tested this first. It looks like we actually DO get the inherited maintainers.

tldahlgren (approved):

.github/workflows/bin/spack-reviewers.py
tldahlgren: I *suppose* it's a moot point.  All packages that inherit from `RubyPackage` don't add maintainers to what is inherited. And all packages that inherit from `RacketPackage` override the inherited maintainer (with the same account).

haampie:
I would still convert the directory name to the package name, since nothing guarantees that passing `"_7zip"` and `"py_foo_bar"` will work in the future.

alecbcs:
@haampie if you can follow up in another PR with those pkg name modifications I'm happy to review :)

PR #1305, opened 2025-09-04
ci: limit the maximum number of reviewers to ping

alecbcs:
For very large bulk PRs, we should not ping more than 100 people.

### DIFF
--- a/.github/workflows/bin/spack-reviewers.py
+++ b/.github/workflows/bin/spack-reviewers.py
@@ -9,6 +9,7 @@ import subprocess
 import requests
 
 IS_PACKAGE_CHANGE = re.compile(r"repos/spack_repo/builtin/packages/([^/]+)/.*$")
+NUM_MAX_REVIEWERS = 100
 
 
 def main():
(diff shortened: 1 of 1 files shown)


adamjstewart:
I think I'm okay with this? I'm pretty sure GitHub has limits on both the maximum number of reviewers and the maximum number of people that can be mentioned by comment. Wish we had those numbers in mind, but can't find them from a quick search.

If we really want to get pedantic, we should do something like @tgamblin did for black where we scrape all historical PRs, run the bot on all of them to see how many people would have been pinged, and find if there's a distinct cut off between accidental pings and wanted pings. But I'm not volunteering to do that and won't make anyone else do it either. It might actually not be that hard though...

Would be cool to have some kind of `@spackbot ping --force` to override though. But as long as the script exists, I could theoretically run it myself and manually ping people?

.github/workflows/bin/spack-reviewers.py
```diff
@@ -47,6 +48,11 @@ def main():
 
     added_reviewers = reviewers - existing_reviewers
     if added_reviewers:
+        # exit early if the PR would notify a very large number of people
+        if len(added_reviewers) > NUM_MAX_REVIEWERS:
+            print(f"[PR #{pr_number}]: skipping PR that would ping >{NUM_MAX_REVIEWERS} people")
```
kwryankrattiger, line 53: I think this should fail the check and post a comment along the lines of

```
Your pull request exceeds the package change limit ({NUM_MAX_REVIEWERS}). Please consider breaking this up into smaller PRs.
```

And then request review from `@spack-releasers` or something similar since that is a group that can merge with bypass for cases like license header changes or large deprecation/removal situations.

kwryankrattiger:
>  But as long as the script exists, I could theoretically run it myself and manually ping people?

you would have to modify the script to override the limit.

Maybe adding an option for that in this PR would be good as well so we could do so via spackbot with options.

PR #1306, opened 2025-09-04
NCCL: remove maintainer

adamjstewart:
I'm really not qualified to maintain this package. I only added myself because it was needed by much of the ML ecosystem. Any other volunteers want to take over?


adamjstewart:
Actually, I should probably just do one big PR with all of the packages I want to add/remove myself from.

PR #1310, opened 2025-09-04
covfie: add v0.15.3

wdconinc:
This PR adds `covfie`, v0.15.3.


wdconinc:
The examples get built but don't get installed... sob. I'll just reduce this to only add the new version: 463764643a25ef3a2728ad8a7733fe9c2e981c8b.

sethrj (approved):
Checksum is good, no major changes to cmake found.

PR #1312, opened 2025-09-04
Proj 7:9.1: add find_dependency on TIFF

johnwparent:
Previous patches adding linkage to the TIFF::TIFF CMake target in the interface of an exported proj target failed to add a corresponding `find_dependency` call in the exported CMake config.

This PR adds that call (and supporting logic).

Applies to `Proj@7:9.1`

I'm not 100% sure how we've been building against these proj versions for this long without issue, I think it's probably some combination of autotools builds and getting lucky that a lot of packages that use Proj are also importing a TIFF::TIFF target.

Tested locally on Windows


johnwparent:
@alecbcs tentative fix for VTK issue in CI, let's see how Gitlab does.

Either way this should probably be merged so older proj versions don't have broken CMake configs.

johnwparent:
Trying to re-trigger gitlab, it didn't like the un-drafting

adamjstewart:
I obviously trust you (and Kitware) to know what CMake config is required. Just wondering if the config is correct now in the latest versions of PROJ or if we need to upstream this.

johnwparent:
Welp we're now building a newer VTK than we were observing for the failures, and there's an unrelated paraview failure... Not a super useful test to determine if we've fixed the CI regression.

johnwparent:
> I obviously trust you (and Kitware) to know what CMake config is required. Just wondering if the config is correct now in the latest versions of PROJ or if we need to upstream this.

I checked to see how far into Proj my patch would apply when writing it. They start using the `TIFF::TIFF` target (and subsequently export it correctly) starting in 9.2, so no need for us to upstream. 

Realistically this patch is only necessary because Spack is changing the exported interface via a patch and those patches don't do enough to handle that change. If we don't apply the `TIFF::TIFF` target usage patches, Proj actually exports its tiff interface correctly for all versions, albeit using the "old CMake" conventions (listing the absolute path to the linked library vs a target name).

I generally try to upstream my patches when relevant, if only to have some url to point to above the patch for some context. 
I think I'll actually make a note for me to go through all my patches next week and see if I missed upstreaming some, thanks for making me think about this a bit more.

adamjstewart:
I'm CMake illiterate so have no idea what you just said, but it sounds like this fix isn't needed in newer versions because we don't patch other parts in newer versions. If so, looks good to me!

adamjstewart (dismissed):

alalazo:
https://gitlab.spack.io/spack/spack-packages/-/jobs/18016169

johnwparent:
> https://gitlab.spack.io/spack/spack-packages/-/jobs/18016169

```
CMake Error at /home/software/spack/[padded-to-256-chars]/linux-x86_64_v3/proj-8.1.0-rstkrzusbsrfbvgrb5nmxetephveu6b6/lib/cmake/proj4/proj4-config.cmake:27 (include):
  include could not find requested file:
    /proj4-targets.cmake
```

Well it's a new error at least... ugh

johnwparent:
> > https://gitlab.spack.io/spack/spack-packages/-/jobs/18016169
> 
> ```
> CMake Error at /home/software/spack/[padded-to-256-chars]/linux-x86_64_v3/proj-8.1.0-rstkrzusbsrfbvgrb5nmxetephveu6b6/lib/cmake/proj4/proj4-config.cmake:27 (include):
>   include could not find requested file:
>     /proj4-targets.cmake
> ```
> 
> Well it's a new error at least... ugh

Got it I think.

johnwparent:
@adamjstewart @alalazo green CI lgtm

adamjstewart (approved):

johnwparent:
Thanks! Hopefully this fixes the CI

PR #1318, opened 2025-09-05
remove unused `D` provider

haampie:
~Depends on https://github.com/spack/spack/pull/51274~: no longer the case after removal.

`D` is not a valid package name, so I don't see why it should be used as a
virtual name. And `d` is probably too short, so use `d-lang`?

The `ldc` package hasn't been updated in half a decade, so I guess this is
not controversial. Nobody uses it :)


haampie:
Actually, given this provider is unused, I would suggest to just remove it entirely.

michaelkuhn (approved):

PR #1319, opened 2025-09-05
ELPA: new version 2025.06.001

yizeyi18:
Add new version 2025.06.001 of package elpa.
<del>
Previous commit 621ec78 moved configure cmdline options to flag_handler, which caused conflicts against using mpi's compiler wrappers: </del>
<del> - flag_handler produces `SPACK_*FLAGS` </del>
<del>   -> which cannot be used by mpi's wrappers
<del>   -> causing configure test about BLAS fails for not founding dynlib at run time(i.e. missing rpath) on some machines. 

<del>This commit turns to use spack's compiler wrappers and add mpi's flags manually.

<del>A special case is about openmpi: openmpi's `libs()` method only returns the C-side dynlib `libmpi.so`, and elpa requires both C-side and Fortran-side. To avoid interference of other packages, openmpi stays unchanged and only elpa was edited. Should this be reported in a seperate issue?
</del>

**EDIT**: Due to unable to reproduce the configure problem, changes about MPI are reverted.

The range to apply `hipcc.patch` was set to `:2025.06.001` due to marekandreas/elpa#68.

### DIFF
--- a/repos/spack_repo/builtin/packages/elpa/package.py
+++ b/repos/spack_repo/builtin/packages/elpa/package.py
@@ -185,7 +185,7 @@ class Elpa(AutotoolsPackage, CudaPackage, ROCmPackage):
                 flags.append(spec["mpi"].headers.include_flags)
             if name == "fflags":
                 flags.append(f'-I{spec["mpi"].prefix.lib}')
-            if name == "ldflags":# or name == "fflags":
+            if name == "ldflags":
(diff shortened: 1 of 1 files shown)


yizeyi18:
@spackbot fix style

yizeyi18:
@spackbot fix style

RMeli (approved):

PR #1325, opened 2025-09-05
Updated r packages, depends_on, broken packages

emwjacobson:
A fairly large number of packages touched here, most of them for the same reason, all for the purpose of getting `trinity` installed while also using the newest version of R (`4.5.1`)

A large number of R packages just needed c, cxx, or fortran depends.

**express** - this relies on protobuf, but was never added as an explicit `depends_on`. A patch to the CMake file is also required in order to not depend on a hard-coded library file. Finally, if using `boost@1.82:`, patching the CXX standard is also required as boost >= 1.82 requires C++14. If there's a more elegant way of doing this, I'm all ears.

**trinity** - part of the build process creates a symlink that points to itself. This would cause the install to fail with `Error: OSError: [Errno 40] Too many levels of symbolic links`. Deleting this file from the installation directory allows the program to install.


alalazo:
> A large number of R packages just needed c, cxx, or fortran depends.

Would you be willing to extract this in a PR of its own?

emwjacobson:
> > A large number of R packages just needed c, cxx, or fortran depends.
> 
> Would you be willing to extract this in a PR of its own?

Done! #1381

emwjacobson:
Updated this PR with the latest R contributions. Realized I missed a couple dependencies for rcurl and limma on the other R commit (sorry...). Otherwise should be good to go! @alalazo :)

repos/spack_repo/builtin/packages/express/package.py
```diff
@@ -56,5 +57,15 @@ def patch(self):
                 "%s" % self.spec["bamtools"].libs,
             )
 
+            # Dont try and compile against specific file
+            edit.filter(
+                r"\${PROTOBUF_LIB_DIR}/" "libprotobuf.a", "%s" % self.spec["protobuf"].libs
```
becker33:
```suggestion
                r"\${PROTOBUF_LIB_DIR}/libprotobuf.a", "%s" % self.spec["protobuf"].libs
```
The string concatenation in this line is really difficult to read as-is

You could also use the `string=True` argument to avoid needing to escape regex symbols.

becker33 (changes requested):

becker33 (approved):

PR #1326, opened 2025-09-05
py-mdanalysis: add v2.8.0, v2.9.0, v2.10.0 and related packages

LydDeb:


repos/spack_repo/builtin/packages/py_mdanalysis/package.py
```diff
@@ -61,8 +66,12 @@ class PyMdanalysis(PythonPackage):
     depends_on("py-tqdm@4.43.0:", type=("build", "run"))
     depends_on("py-threadpoolctl", type=("build", "run"))
     depends_on("py-packaging", type=("build", "run"))
-    depends_on("py-fasteners", type=("build", "run"))
+    depends_on("py-fasteners", when="@:2.8.0", type=("build", "run"))
     depends_on("py-mda-xdrlib", when="@2.7.0:", type=("build", "run"))
+    depends_on("py-waterdynamics", when="@2.8.0", type=("build", "run"))
+    depends_on("py-pathsimanalysis", when="@2.8.0", type=("build", "run"))
+    depends_on("py-mdahole2", when="@2.8.0", type=("build", "run"))
```
RMeli: This is a circular dependency. Since the different MDAKits are optional within MDAnalysis, I think it best to drop them from here and just keep the MDAKit package themselves.

adamwitmer:
This package should now be updated to include version `2.10.0` (released 10/26/25), i.e.:

```
version("2.10.0", sha256="5cef1f1eba4b25cbf9dd4045f08eeab293bb9577b970aa96132f14491312d877")
```

as is noted at https://www.mdanalysis.org/2025/10/26/release-2.10.0/:

> This release supports NumPy 2.0+ and offers backwards compatibility through to NumPy 1.26.0.

> Supported Python versions: 3.11, 3.12, 3.13, 3.14. Note: not all optional dependencies currently work with Python 3.14.

repos/spack_repo/builtin/packages/py_mdanalysis/package.py
LydDeb: Thank you, I have removed those dependencies.

LydDeb:
> This package should now be updated to include version `2.10.0` (released 10/26/25), i.e.:
> 
> ```
> version("2.10.0", sha256="5cef1f1eba4b25cbf9dd4045f08eeab293bb9577b970aa96132f14491312d877")
> ```
> 
> as is noted at https://www.mdanalysis.org/2025/10/26/release-2.10.0/:
> 
> > This release supports NumPy 2.0+ and offers backwards compatibility through to NumPy 1.26.0.
> 
> > Supported Python versions: 3.11, 3.12, 3.13, 3.14. Note: not all optional dependencies currently work with Python 3.14.

Thank you, I've added this version.

repos/spack_repo/builtin/packages/py_mdanalysis/package.py
```diff
@@ -89,3 +98,7 @@ class PyMdanalysis(PythonPackage):
     depends_on("py-gsd@1.9.3:", when="@:2.5.0", type=("build", "run"))
     depends_on("py-biopython@1.80:", when="@:2.6.1", type=("build", "run"))
     depends_on("py-networkx@2.0:", when="@:2.6.1", type=("build", "run"))
+
+    def url_for_version(self, version):
+        if version >= Version("2.8.0"):
+            return f"https://files.pythonhosted.org/packages/source/m/mdanalysis/mdanalysis-{version}.tar.gz"
```
RMeli, line 107: Why was this necessary?

repos/spack_repo/builtin/packages/py_mdanalysis/package.py
RMeli, line 1: Would you mind updating `py-mdanalysistests` too, since they are released at the same time? Otherwise I can take care of it in a follow-up PR.

repos/spack_repo/builtin/packages/py_mdanalysis/package.py
```diff
@@ -38,19 +41,24 @@ class PyMdanalysis(PythonPackage):
     depends_on("c", type="build")  # generated
     depends_on("cxx", type="build")  # generated
 
+    depends_on("python@3.11:", type=("build", "run"), when="@2.10.0:")
+    depends_on("python@3.10:", type=("build", "run"), when="@2.8.0:")
     depends_on("python@3.9:", type=("build", "run"), when="@2.5.0:")
     depends_on("python@3.8:", type=("build", "run"))
 
     depends_on("py-setuptools", type="build")
+    depends_on("py-setuptools@10.9.0:", type="build", when="@2.8.0:")
```
RMeli:
```suggestion
```
For `py-mdanalysis@2.8.0` it is already `setuptools@40.9.0:`.

RMeli (commented):
Thanks for looking into it. I left just a few comments, otherwise LGTM.

repos/spack_repo/builtin/packages/py_mdanalysis/package.py
LydDeb: Before 2.8.0 the name of the archive is `MDAnalysis-<version>.tar.gz`

repos/spack_repo/builtin/packages/py_mdanalysis/package.py
LydDeb: Okay, I'll take care of it.

repos/spack_repo/builtin/packages/py_mdanalysis/package.py
LydDeb: Good catch, it might have been a typo.

repos/spack_repo/builtin/packages/py_mdanalysis/package.py
RMeli: Ah! I did not realize this. Thanks for pointing it out.

```suggestion
    def url_for_version(self, version):
        """
        From version 2.8.0 onwards, the archive named changed
        to 'mdanalysis-{version}.tar.gz from 'MDAnalysis-{version}.tar.gz
        """
        if version >= Version("2.8.0"):
            return f"https://files.pythonhosted.org/packages/source/m/mdanalysis/mdanalysis-{version}.tar.gz"
```
Might be worth adding a small comment about it, so it's clearer what is going on.

repos/spack_repo/builtin/packages/py_mdanalysistests/package.py
```diff
@@ -25,17 +28,30 @@ class PyMdanalysistests(PythonPackage):
     version("2.4.2", sha256="6e8fb210a4268691c77717ea5157e82d85874a4f7ee0f8f177718451a44ee793")
 
     # Version need to match MDAnalysis'
+    depends_on("py-mdanalysis@2.9.0", when="@2.9.0", type=("build", "run"))
```
RMeli: The corresponding `py-mdnalysis@2.10.0` dependency is missing for `2.10.0`.

Maybe one could have a dictionary of version and sha256 and loop over it to avoid mismatehces, but this can be done in a separate pr.

LydDeb: > The corresponding py-mdnalysis@2.10.0 dependency is missing for 2.10.0.

I added it.

> Maybe one could have a dictionary of version and sha256 and loop over it to avoid mismatehces, but this can be done in a separate pr.

I don't understand what you want to do with this dictionary. Could you elaborate on your idea?

RMeli: To avoid this problem in the future, we could have something like the following dictionary

https://github.com/spack/spack-packages/blob/92789f0bbd082f52996a0ccbc869ed352915c23b/repos/spack_repo/builtin/packages/py_metatensor_core/package.py#L10-L13

and then have a loop that adds the `version` and the corresponding `depends_on` as

https://github.com/spack/spack-packages/blob/92789f0bbd082f52996a0ccbc869ed352915c23b/repos/spack_repo/builtin/packages/py_metatensor_core/package.py#L30-L32

RMeli (approved):
LGTM, thanks @LydDeb for all the work!

PR #1329, opened 2025-09-05
py-llvmlite: add versions and ncurses link dependency

jrood-nrel:
Without this I see missing `-ltinfo`


jrood-nrel:
Actually these versions don't build. I see this same error: https://github.com/numba/llvmlite/issues/1294

PR #1330, opened 2025-09-05
ccls: add versions and update llvm constraints

jrood-nrel:
Also `0.20201025` doesn't appear to exist so I removed it.


jacobmerson (approved):
I checked the version hashes. Looks good.

alalazo (approved):

PR #1331, opened 2025-09-05
amrvis: add c and fortran dependency

jrood-nrel:

### DIFF
--- a/repos/spack_repo/builtin/packages/amrvis/package.py
+++ b/repos/spack_repo/builtin/packages/amrvis/package.py
@@ -39,7 +39,9 @@ class Amrvis(MakefilePackage):
     variant("debug", default=False, description="Enable debugging features")
     variant("profiling", default=False, description="Enable AMReX profiling features")
 
-    depends_on("cxx", type="build")  # generated
+    depends_on("cxx", type="build")
+    depends_on("c", type="build")
+    depends_on("fortran", type="build")
 
     depends_on("gmake", type="build")
     depends_on("mpi", when="+mpi")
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/amrvis/package.py
```diff
@@ -39,7 +39,9 @@ class Amrvis(MakefilePackage):
     variant("debug", default=False, description="Enable debugging features")
     variant("profiling", default=False, description="Enable AMReX profiling features")
 
-    depends_on("cxx", type="build")  # generated
+    depends_on("cxx", type="build")
+    depends_on("c", type="build")
+    depends_on("fortran", type="build")
```
alalazo: Why do we need fortran in this package?

jrood-nrel: @WeiqunZhang Is this only the case if `USE_PROFPARSER=TRUE`? I can maybe reduce when fortran is necessary.

jrood-nrel: https://github.com/AMReX-Codes/Amrvis/blob/5a915cc5396d083d89f6a0aec0f3739f316e06c1/GNUmakefile#L46-L52

WeiqunZhang: Yes, Fortran is needed only if `USE_PROFPARSER=TRUE`.

bernhardkaindl (approved):
@jrood-nrel, you might want to re-request a review from alalazo by clicking the "arrow-in-circles" icon left of his avatar in the list of reviewers. In case you can't find it, I can do so too.

If there are no further review questions in due time, I think this can be merged. Cc: @tldahlgren

tldahlgren (approved):

tldahlgren:
Committed since @alalazo's feedback was addressed and none from the maintainer.

PR #1332, opened 2025-09-05
py-imageio: add v2.37.0

manuelakuhn:
https://github.com/imageio/imageio/tree/v2.37.0


alecbcs (approved):
Looks good to me. Thanks!

PR #1333, opened 2025-09-05
xterm: add versions and c and ncurses dependency

jrood-nrel:


tldahlgren (commented):
Confirmed the 8 new version sha256s. 

Are there substantial improvements in each of the versions to make it worth including all of them?

jrood-nrel:
I have no idea. I definitely needed the `depends_on` statements for a machine that didn't have ncurses and the `c` dependency is obviously missing. I just added all the versions I could find as a public service.

tldahlgren (approved):

PR #1334, opened 2025-09-05
py-wandb: add v0.21.0

thomas-bouvier:

### DIFF
--- a/repos/spack_repo/builtin/packages/py_sentry_sdk/package.py
+++ b/repos/spack_repo/builtin/packages/py_sentry_sdk/package.py
@@ -36,12 +36,13 @@ class PySentrySdk(PythonPackage):
     depends_on("python@3.6:", type=("build", "run"), when="@2:")
     depends_on("python@2.7,3.4:", type=("build", "run"))
     depends_on("py-setuptools", type="build")
-    depends_on("pu-urllib3@1.26.11:", type=("build", "run"), when="@2.34.1:")
+    depends_on("py-urllib3@1.26.11:", type=("build", "run"), when="@2.34.1:")
     depends_on("py-urllib3@1.10.0:", type=("build", "run"))
     depends_on("py-certifi", type=("build", "run"))
 
     depends_on("py-flask@0.11:", type=("build", "run"), when="+flask")
     depends_on("py-blinker@1.1:", type=("build", "run"), when="+flask")
+    depends_on("py-markupsafe", type=("build", "run"), when="@1.23: +flask")
     depends_on("py-quart@0.16.1:", type=("build", "run"), when="+quart")
     depends_on("py-blinker@1.1:", type=("build", "run"), when="+quart")
     depends_on("py-bottle@0.12.13:", type=("build", "run"), when="+bottle")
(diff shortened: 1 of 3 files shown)


thomas-bouvier:
I can't understand why the following error is thrown:

```
==> Error: Package 'spack_repo.builtin.packages.py_eval_type_backport.package' not found in repository '/home/runner/work/spack-packages/spack-packages/repos/spack_repo/builtin'
```

This package is added by the current PR.

thomas-bouvier:
@spackbot re-run pipeline

thomas-bouvier:
Hey @alalazo, it seems that the pipeline failed because of an internal error.

tldahlgren (commented):
Confirmed the new package homepage and all of the new version sha256s.

repos/spack_repo/builtin/packages/py_sentry_sdk/package.py
```diff
@@ -32,20 +33,24 @@ class PySentrySdk(PythonPackage):
     variant("chalice", default=False, description="Builts with chalice")
     variant("httpx", default=False, when="@1.5.5:", description="Builts with httpx")
 
+    depends_on("python@3.6:", type=("build", "run"), when="@2:")
     depends_on("python@2.7,3.4:", type=("build", "run"))
```
tldahlgren: You probably want to add the appropriate condition here, e.g.


```suggestion
    depends_on("python@2.7,3.4:", type=("build", "run"), when="@:1")
```

tldahlgren: Same for the other dependencies whose versions have changed for the newer release(s).

repos/spack_repo/builtin/packages/py_wandb/package.py
```diff
@@ -18,27 +18,55 @@ class PyWandb(PythonPackage):
...
 
-    depends_on("py-setproctitle", type=("build", "run"))
-    depends_on("py-appdirs@1.4.3:", type=("build", "run"))
-    depends_on("py-protobuf@3.19:4", type=("build", "run"))
-    conflicts("^py-protobuf@4.21.0")
+    depends_on("py-pyyaml", type=("build", "run"))
+
+    depends_on("py-protobuf@3.19:6", type=("build", "run"), when="@0.19.10: ^python@3.10:")
+    depends_on("py-protobuf@3.15:6", type=("build", "run"), when="@0.19.10: ^python@3.9")
+    depends_on("py-protobuf@3.12:6", type=("build", "run"), when="@0.19.10: ^python@:3.8")
```
tldahlgren: Is it not possible to also include `platform=linux` in the condition (e.g. `when="@0.19.10: platform=linux ^python@:3.8"`)?

tldahlgren (changes requested):
Confirmed (new) homepage, version sha256s, and dependencies though there are some conditions in `py-wandb` that should probably be included/updated.

repos/spack_repo/builtin/packages/py_sentry_sdk/package.py
thomas-bouvier: I'm sorry, I don't understand why this is necessary. Ideally, the exact upper version bound should be specified, which adds a lot of work. Usually they are not even documented. Why is the current solution not satisfactory?

tldahlgren: Being explicit when a dependency doesn't apply should reduce the numbers of dependencies that have to be considered during concretization.

tldahlgren (dismissed):
Thanks.

tldahlgren:
@spackbot fix style

tldahlgren (approved):

thomas-bouvier:
Thank you @tldahlgren !

PR #1340, opened 2025-09-05
py_faststructure: Removed due to python 2

emwjacobson:
Faststructure is written in python 2, which has been deprecated from spack. This PR removed py_faststructure due to it not being able to be installed anymore.

Also pinging @wdconinc for review


wdconinc (approved):
Thanks for identifying and fixing this obsolete package. Since we don't package python 2 anymore, we can remove the package immediately since it is not buildable.

PR #1346, opened 2025-09-07
gtkplus: add v3.24.43, v3.24.50 (last old stable); default meson; fix url

wdconinc:
This PR adds `gtkplus`, v3.24.43, v3.24.50, the last two versions of the so-called "old stable" series. [Diff 3.24.41...3.24.50](https://gitlab.gnome.org/GNOME/gtk/-/compare/3.24.41...3.24.50)

v3.24.43 is the last version in the current url scheme (with `gtk+`), and  v3.24.50 is the last version in the new url scheme (with `gtk`), and the versions from v3.24.44 to v3.24.47 are nowhere to be found... (Note that version up to 4.20.0 are available, but that's for a different PR.)

The newer versions, with the `gtk` url scheme also defines the meson project name as `gtk` instead of `gtk+`. I am not sure if this affects the ability of dependents to find the package if they use `gtk+` as name.

This package is build in CI, but it falls back to the `autotools` version, so the most recent versions are not built. That's something I'll look into just to make sure we build newer versions as they are added, but may become a different PR.
> EDIT: Even after changing the default to `meson`, this doesn't trigger a rebuild. It may be that this package used to get built in CI, but doesn't anymore.

### DIFF
--- a/repos/spack_repo/builtin/packages/gtkplus/package.py
+++ b/repos/spack_repo/builtin/packages/gtkplus/package.py
@@ -21,7 +21,7 @@ class Gtkplus(AutotoolsPackage, MesonPackage):
     build_system(
         conditional("autotools", when="@:3.24.35"),
         conditional("meson", when="@3.24.9:"),
-        default="autotools",
+        default="meson",
     )
 
     version("3.24.50", sha256="399118a5699314622165a11b769ea9b6ed68e037b6d46d57cfcf4851dec07529")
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/gtkplus/package.py
```diff
@@ -24,6 +24,8 @@ class Gtkplus(AutotoolsPackage, MesonPackage):
         default="autotools",
```
wdconinc: Let's just use the default `meson` and see if that gets (I think) `e4s` stacks to use the newer versions.
```suggestion
        default="meson",
```

alalazo (approved):
Recipe changes LGTM. Feel free to merge if you're not waiting on further reviews.

tldahlgren:
Merging since this has been in "approved" status for over 2 months.

PR #1347, opened 2025-09-07
librsvg: add v2.59.2, v2.60.0, v2.61.0; switch to meson

wdconinc:
This PR adds `librsvg`, v2.59.2, v2.60.0, v2.61.0, which is now using meson as the build system. No major version changes, per https://gitlab.gnome.org/GNOME/librsvg/-/blob/main/meson.build?ref_type=heads#L51-58.

This package is built in CI.
> EDIT: It doesn't seem like this package was rebuilt in CI, even though it is listed on cache.spack.io...


alecbcs (approved):

PR #1350, opened 2025-09-08
py-coherent-licensed: add new package

manuelakuhn:
https://pypi.org/project/coherent.licensed/#files
(The files in the pypi tar file differ from them in github)


michaelkuhn (approved):

PR #1351, opened 2025-09-08
py-importlib-metadata: add v8.7.0 and py-zipp: add v3.23.0

manuelakuhn:
https://github.com/python/importlib_metadata/tree/v8.7.0
https://github.com/jaraco/zipp/tree/v3.21.0

`py-zipp@2.23.0` would need a dependency on `py-coherent-licensed` (#1350) but as soon as I add this dependency the concretizer throws an (not helpful) error:
```
==> Error: failed to concretize `py-zipp@3.23` for the following reasons:
     1. Cannot satisfy 'py-zipp@3.23'
     2. Cannot satisfy 'py-zipp@3.23'
        required because py-zipp@3.23 requested explicitly
```


manuelakuhn:
@alalazo any idea how to solve the problem with concretization? I have now another package were I have the same issue: as soon as I add a dependency to a new package that builds with `py-flit-core` (no other deps)  whereas the root package builds with `py-setuptools` , the concretization fails.

manuelakuhn:
I found the issue for the newer `py-zipp` versions and it has nothing to do with `py-flit-core` vs `py-setuptools` but was actually a mistake on my side. The reason was a typo and because of the really bad error handling and checking of the concretizer it took me ages to find it. See https://github.com/spack/spack/issues/51364

manuelakuhn:
So once this PR and #1350 are merged, I can update `py-zipp` as well to the newest version.

michaelkuhn (approved):

PR #1352, opened 2025-09-08
py-httpcore: add v1.0.9

manuelakuhn:
https://github.com/encode/httpcore/tree/1.0.9

### DIFF
--- a/repos/spack_repo/builtin/packages/py_httpcore/package.py
+++ b/repos/spack_repo/builtin/packages/py_httpcore/package.py
@@ -40,4 +40,4 @@ class PyHttpcore(PythonPackage):
         depends_on("py-anyio@3", when="@0.14.7")
         depends_on("py-anyio@3:4", when="@0.16.3:0.18")
 
-    depends_on("py-setuptools", when="@:1.16.3", type="build")
+    depends_on("py-setuptools", when="@:0.16.3", type="build")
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/py_httpcore/package.py
```diff
@@ -16,24 +16,28 @@ class PyHttpcore(PythonPackage):
...
-        depends_on("py-h11@0.13:0.14", when="@0.16.3:")
+        depends_on("py-h11@0.8:0.9", when="@0.11.0")
 
+        # Historical dependencies
         depends_on("py-sniffio@1", when="@0")
 
         depends_on("py-anyio@3", when="@0.14.7")
         depends_on("py-anyio@3:4", when="@0.16.3:0.18")
+
+    depends_on("py-setuptools", when="@:1.16.3", type="build")
```
alecbcs, line 43:
```suggestion
    depends_on("py-setuptools", when="@:0.16.3", type="build")
```

I think there's a typo here and should be `0.16` not `1.16`

alecbcs (changes requested):

repos/spack_repo/builtin/packages/py_httpcore/package.py
manuelakuhn: Yes, you are right, the version does not make sense. Thanks!

alecbcs (approved):
Looks good to me now. Thanks!

PR #1357, opened 2025-09-08
abinit: add GPU support, bump versions & dependencies, fix GCC15

marcsrt:
This PR upgrades recipe for ABINIT, a DFT simulation code, to add features, fix compilation bugs and add new versions 10.4.3 and 10.4.7.

The main addition is GPU support. ABINIT features has been partly ported on GPU using OpenMP GPU offload and only works with NVHPC & CUDA or Cray & ROCm for now. It is enabled using new `+gpu_openmp_offload` variant, being implied by added `+cuda` or `+rocm` variants.

This PR also adds two variants for tuning GPU compilation, regarding CUDA-aware MPI, `+gpu_aware_mpi`, and NVTX/ROCtx markers, `+gpu_markers`.

In 10.4.x, build system of ABINIT changed handling of GPU so this recipe will only enable GPU from this version point.

Some other changes concerns existing dependencies : 
- AtomPAW is no longer required on recent ABINIT releases
- Recent HDF5 version are supported on recent ABINIT releases
- Recent libXC is required to be compiled with `+kxc` variant

Finally, a bug concerning GCC 15 is fixed by a patch for versions <10.4.

### DIFF
--- a/repos/spack_repo/builtin/packages/abinit/package.py
+++ b/repos/spack_repo/builtin/packages/abinit/package.py
@@ -3,11 +3,13 @@
 # SPDX-License-Identifier: (Apache-2.0 OR MIT)
 
 from spack_repo.builtin.build_systems.autotools import AutotoolsPackage
+from spack_repo.builtin.build_systems.cuda import CudaPackage
+from spack_repo.builtin.build_systems.rocm import ROCmPackage
 
 from spack.package import *
 
 
-class Abinit(AutotoolsPackage):
+class Abinit(AutotoolsPackage, CudaPackage, ROCmPackage):
     """ABINIT is a package whose main program allows one to find the total
     energy, charge density and electronic structure of systems made of
     electrons and nuclei (molecules and periodic solids) within
@@ -53,6 +55,12 @@ class Abinit(AutotoolsPackage):
 
     variant("wannier90", default=False, description="Enables the Wannier90 library")
     variant("libxml2", default=False, description="Enable libxml2 support, used by multibinit")
+    variant(
+        "gpu_openmp_offload",
+        when="@10.4: +openmp",
+        default=False,
+        description="Enable OpenMP offload support. Requires CUDA & NVHPC or ROCM & Cray CE",
+    )
 
     variant(
         "optimization-flavor",
@@ -95,6 +103,11 @@ class Abinit(AutotoolsPackage):
         # Cannot ask for +wannier90 if it does not depend on MPI
         conflicts("+wannier90")
 
+    with when("+rocm"):
+        depends_on("hipblas+rocm")
+        depends_on("hipfft+rocm")
+        depends_on("hipsolver+rocm")
+
     # constrain version of hdf5
     depends_on("hdf5@:1.8", when="@9:9.8")
     depends_on("hdf5@:1.14", when="@10:")
@@ -116,6 +129,58 @@ class Abinit(AutotoolsPackage):
 
     conflicts("%gcc@7:", when="@:8.8")
     conflicts("%gcc@9:", when="@:8.10")
+
+    conflicts("%nvhpc", when="@:9.8")
+    conflicts("%cce", when="@:9.8")
+
+    conflicts(
+        "+gpu_openmp_offload",
+        when="~mpi",
+        msg="ABINIT builds with GPU support require MPI support enabled as well",
+    )
+    conflicts(
+        "+cuda",
+        when="~gpu_openmp_offload",
+        msg="ABINIT builds with GPU support require OpenMP offload support enabled as well",
+    )
+    conflicts(
+        "+rocm",
+        when="~gpu_openmp_offload",
+        msg="ABINIT builds with GPU support require OpenMP offload support enabled as well",
+    )
+    conflicts(
+        "+cuda",
+        when="~openmp",
+        msg="ABINIT builds with GPU support require OpenMP support enabled as well",
+    )
+    conflicts(
+        "+rocm",
+        when="~openmp",
+        msg="ABINIT builds with GPU support require OpenMP support enabled as well",
+    )
+    conflicts("cuda_arch=none", when="+cuda", msg="CUDA architecture is required")
+    conflicts("amdgpu_target=none", when="+rocm", msg="ROCM architecture is required")
+
+    requires(
+        "%nvhpc",
+        "%cce",
+        when="+gpu_openmp_offload",
+        policy="one_of",
+        msg="ABINIT with OpenMP offload builds only with NVHPC or Cray CE",
+    )
+    requires(
+        "+cuda",
+        "+rocm",
+        when="+gpu_openmp_offload",
+        policy="one_of",
+        msg="ABINIT with OpenMP offload builds only with CUDA or ROCM",
+    )
+
+    # propagate ROCm architecture when +rocm
+    for arch in ROCmPackage.amdgpu_targets:
+        depends_on("hipblas+rocm amdgpu_target=%s" % arch, when="+rocm amdgpu_target=%s" % arch)
+        depends_on("hipfft+rocm amdgpu_target=%s" % arch, when="+rocm amdgpu_target=%s" % arch)
+        depends_on("hipsolver+rocm amdgpu_target=%s" % arch, when="+rocm amdgpu_target=%s" % arch)
 
     # need openmp threading for abinit+openmp
     # TODO: The logic here can be reversed with the new concretizer. Instead of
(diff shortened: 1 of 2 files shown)


downloadico (approved):
Looks good to me!  Thanks for taking this on.

marcsrt:
There is an SSL certificate issue with the URL where ABINIT releases are fetched from, preventing checksums validation.
Unfortunately, releases from Github are shaped differently, with no `configure` and different checksums so I can't just simply change the URL to validate.
ABINIT devs are aware of the issue and it should be fixed soon...

marcsrt:
After reaching out to the devs of ABINIT, it seems they can't fix the certificate issue right away so I ended up using the official Github repo for recent versions.

The difference with official release tarballs is that a `autoreconf` step is performed using a Python script `./config/scripts/makemake`, which basically calls autoconf for `configure` and generate a few Fortran files.
Hence, the addition of `python` and `automake` as build dependencies.

However, this doesn't work on versions 8 and 9 : 
- `config/scripts/makemake` from 9.8 and earlier has obsoleted regexes that throw errors with recent Python versions
- there are v8:9 patches that fix flags in `configure` from release tarballs, which won't apply since that file doesn't exist in Git sources

Therefore, I decided to keep the old URL for version 8 and 9, that can still work by skipping certificate check, and only fix versions 10.x, as the recent 10.4.x are the one being initially targeted by this MR.

downloadico (commented):
Looks good to me.  TY!

repos/spack_repo/builtin/packages/abinit/package.py
```diff
@@ -51,6 +63,24 @@ class Abinit(AutotoolsPackage):
 
     variant("wannier90", default=False, description="Enables the Wannier90 library")
     variant("libxml2", default=False, description="Enable libxml2 support, used by multibinit")
+    variant(
+        "gpu_openmp_offload",
+        when="@10.4: +openmp",
+        default=False,
+        description="Enable OpenMP offload support. Requires CUDA & NVHPC or ROCM & Cray CE",
+    )
+    variant(
```
kwryankrattiger, line 64: Instead of having a variant for this, it may be better to just detect if the MPI is GPU aware in the configuration step. Is there a reason _not_ to build with that option if the MPI is GPU aware?

marcsrt: For benchmarks purposes or to check for regressions with this flag.

I wasn't sure about how to proceed for that one. I just thought that a user may not be aware of GPU-enablement in MPI nor ABINIT and overlook it.
Having that variant allows to advertise it and ensures that MPI has GPU enabled through the `requires` on line 147.

kwryankrattiger: So my point is `~gpu_aware_mpi` when MPI is GPU aware is going to build without that feature. So now this is a dimension of the package the user needs to be aware if in order to take advantage of the MPI/GPU optimizations (I am guessing that is what the flag is for).

The default here is false, so the out of the box spec that _most_ people will use going to be the "unoptimized" version.

I would argue that for benchmarks it would probably be better to use an MPI that is _not_ GPU aware vs one that is. If you are using a system MPI you can accomplish this same thing by tricking the external spec

```
packages:
    cray-mpich:
        externals:
        - spec: cray-mpich ~cuda # Use this version without cuda support
          prefix: /path/to/cray-mpich/
          modules:
          - cray-mpich
        - spec: cray-mpich +cuda # Use this version with cuda support
          prefix: /path/to/cray-mpich/
          modules:
          - cray-mpich
          - cuda
```

Then your benchmark spec would be

```
abinit ... ^[virtuals=mpi] cray-mpich +cuda
abinit ... ^[virtuals=mpi] cray-mpich ~cuda
```

marcsrt: You're right that most people would overlook it, while my use-cases for this variant are quite niche.

But if we remove this variant, we would rely on user awareness to enable GPU in MPI to leverage that optimization in ABINIT.
And it would be an implicit optimization, with no trace in the specs, which would we confuse a seasoned ABINIT user whether GPU-aware MPI optimization is actually on.

Perhaps I can just set `gpu_aware_mpi` to `true` by default, so most people that target GPU enable it as well ?
I set it to `false` because in ABINIT, the matching flag is `false` by default... but I think it's due of the lack of any auto-detection of GPU-awareness in MPI by `configure` script.

kwryankrattiger: > But if we remove this variant, we would rely on user awareness to enable GPU in MPI to leverage that optimization in ABINIT.

That is already a requirement with the variant, the variant here is just another place it needs to show up.

> And it would be an implicit optimization, with no trace in the specs

It would be in the dependency MPI spec, which is still part of the package spec.

> but I think it's due of the lack of any auto-detection of GPU-awareness in MPI by configure script.

In general I try to advocate not to expose _every_ feature flag a package has, the spec looking like the configure command is not great IMO. The more variants there are the more the solver needs to chew on which has real costs in concretization as well. Spack can carry a lot of the weight of feature detection that may be lacking in a packages build scripts. Spack knows _all_ of the dependencies and _all_ of the features that are exposed in the entire stack.

> Perhaps I can just set gpu_aware_mpi to true by default

That is probably fine. If the variant on the package is really important to your use case then I won't gate keep it, but the default behavior should be taking advantage of the fact that Spack knows the features enabled in MPI, and enable the optimizations available for it.

marcsrt: > That is already a requirement with the variant, the variant here is just another place it needs to show up.

Sure, but user will likely use `spack info abinit` once to display the list of variants and learn about this setting.

> In general I try to advocate not to expose every feature flag a package has, the spec looking like the configure command is not great IMO. The more variants there are the more the solver needs to chew on which has real costs in concretization as well. Spack can carry a lot of the weight of feature detection that may be lacking in a packages build scripts. Spack knows all of the dependencies and all of the features that are exposed in the entire stack.

I understand your points. I see this optimization more like enabling OpenMP on CPU-only builds. You would it want on most of the time but might need it off for select reasons, hence the need for control.

I can think of another reason against relying on MPI specs: when user import MPI through `spack external find`, which may be the case on a supercalculator, as user usually doesn't want to compile MPI themselves.

I pushed the change to have `gpu_aware_mpi` enabled by default.

...

Regarding your recommendations, I now feel that `+gpu_openmp_offload` variant might be offending. I added it because ABINIT has 2 other legacy GPU implementations based on CUDA kernels and Kokkos, but... there is no plan in enabling those in Spack right now and `+cuda` or `+rocm` simply enables `+gpu_openmp_offload` implicitly.

Having that variant mostly allows to : 
- group constraints and variants common to `+cuda` and `+rocm` in the recipe
- signaling reliance on OpenMP offload rather than CUDA/ROCm kernels to justify the constraints on supported compilers
- explicit tell an ABINIT developer `--enable-openmp-offload=yes` is being used

User is never expected to have `~gpu_openmp_offload` along `+cuda`. I'm not sure if it was clear that `+gpu_openmp_offload` is mostly implicitly used but given our discussion, I now feel this one might be too much.
Your thoughts ?

marcsrt:
I removed the Git workaround now that original source for ABINIT tarballs works again.

repos/spack_repo/builtin/packages/abinit/package.py
```diff
@@ -93,13 +126,34 @@ class Abinit(AutotoolsPackage):
...
+        requires("^mpich +cuda", when="^[virtuals=mpi] mpich")
+        requires("^mvapich +cuda", when="^[virtuals=mpi] mvapich")
+        requires("^mvapich2 +cuda", when="^[virtuals=mpi] mvapich2")
+        requires("^mvapich2-gdr +cuda", when="^[virtuals=mpi] mvapich2-gdr")
+        requires("^openmpi+cuda", when="^[virtuals=mpi] openmpi")
+
+    with when("+gpu_aware_mpi+rocm"):
+        requires("^cray-mpich +rocm", when="^[virtuals=mpi] cray-mpich")
+        requires("^mpich +rocm", when="^[virtuals=mpi] mpich")
+        requires("^mvapich2-gdr +rocm", when="^[virtuals=mpi] mvapich2-gdr")
```
becker33: This probably needs logic to require that one of these supported mpi implementations is used when the variant is on. Analogous to the requirements on the compiler when offload is turned on.

becker33 (changes requested):

repos/spack_repo/builtin/packages/abinit/package.py
marcsrt: Sorry for the delay !
I've just pushed a change refactoring this section with the following below:

```
    # Need a MPI provider built with either CUDA or ROCM when configured
    # with GPU-aware optimizations
    requires(
        "^cray-mpich +cuda",
        "^mpich +cuda",
        "^mvapich +cuda",
        "^mvapich2 +cuda",
        "^mvapich2-gdr +cuda",
        "^openmpi+cuda",
        when="+gpu_aware_mpi+cuda",
        policy="one_of",
        msg="'+gpu_aware_mpi +cuda' requires an MPI implementation with CUDA awareness enabled.",
    )

    requires(
        "^cray-mpich +rocm",
        "^mpich +rocm",
        "^mvapich2-gdr +rocm",
        "^openmpi +rocm",
        when="+gpu_aware_mpi+rocm",
        policy="one_of",
        msg="'+gpu_aware_mpi +rocm' requires an MPI implementation with ROCm awareness enabled.",
    )
```

marcsrt: And removed `mvapich2-gdr` from the list, the package seems gone.

downloadico (approved):
looks good to me

becker33 (approved):

PR #1361, opened 2025-09-08
py-zarr: add new version, fix deps

adamjstewart:
Previous versions of py-zarr and py-numcodecs had a bunch of missing or incorrect dependencies that prevented either package from being imported.


alalazo:
@spackbot fix style

adamjstewart:
~@alecbcs bug in spackbot maintainers~

Looks like this has already been fixed in develop.

becker33 (approved):

PR #1366, opened 2025-09-08
py-kornia-rs: pin another dependency

kwryankrattiger:

### DIFF
--- a/repos/spack_repo/builtin/packages/py_kornia_rs/package.py
+++ b/repos/spack_repo/builtin/packages/py_kornia_rs/package.py
@@ -45,6 +45,7 @@ class PyKorniaRs(PythonPackage):
 
     # See https://github.com/kornia/kornia-rs/commit/93f768137814709b60767f7fc24a9b0184002aee
     patch("py-kornia-rs-pin-fast_image_resize.patch", when="@0.1.9")
+    patch("py-kornia-rs-pin-circular_buffer.patch", when="@0.1.9")
 
     @property
     def build_directory(self):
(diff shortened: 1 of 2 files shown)


alalazo:
@alecbcs In this PR we didn't ping maintainers, even though the action seems to identify them correctly: https://github.com/spack/spack-packages/actions/runs/17564713655/job/49889079846?pr=1366

alalazo:
@adamjstewart  @cjpurackal @ducha-aiki @edgarriba @johnnv1 @lferraz @shijianjian

edgarriba:
not sure why is this raising an error and in the repo for kornia-rs-py we do not expose this feature in principle

adamjstewart:
I would also like to know which bug is being fixed.

kwryankrattiger:
The bug is a dependency (`circular-buffer`) got updated in the cargo repo. `py-kornia-rs@0.1.9` is configured to assume that future versions of that dependency were going to continue to be compatible. That is not the case, newer versions of `circular-buffer` require a newer rust compiler.

This was also the case for the other patch that was applied for this project.

adamjstewart:
Thanks, I now see the error message in https://gitlab.spack.io/spack/spack-packages/-/jobs/18062302. LGTM.

adamjstewart:
I guess another solution would be to declare a dependency on rust 1.87+. Technically, rust packages should be maintained just like any other package where we have a separate recipe for circular-buffer and pin the versions used. Right now, none of our rust builds are reproducible.

I'm fine with merging this as is, but let's keep pushing on better rust support in the future.

adamjstewart (approved):

kwryankrattiger:
> I'm fine with merging this as is, but let's keep pushing on better rust support in the future.

There has been some conversion on this after the last time this happened :) 

https://spackpm.slack.com/archives/C5VL7V81G/p1755190838823679

PR #1371, opened 2025-09-09
py-jaraco-classes: add v3.4.0

manuelakuhn:
https://github.com/jaraco/jaraco.classes/tree/v3.4.0


alecbcs (approved):
Looks good to me. Thanks!

PR #1373, opened 2025-09-09
Deprecations: GCCs older than 8

becker33:
Closes #1224 

@adamjstewart


adamjstewart (approved):
I'm obviously fine with this assuming the other GCC maintainers are fine with this new deprecation policy.

michaelkuhn:
I don't really get why we would want to undeprecate the older minor releases, but if someone takes care of them, that's obviously also fine for me. I'm only actively using the latest release of any branch.

PR #1374, opened 2025-09-09
armpl-gcc: add v25.07

paolotricerri:
This patch introduces the possibility of installing ArmPL 25.07.

It also removes the libstdc++ from the list of required libraries as the necessary functions should already be included in the installed ArmPL.


paolotricerri:
@spackbot run pipeline

tldahlgren (approved):

PR #1382, opened 2025-09-09
kokkos: change the default value of the variant serial

tpadioleau:
The PR changes the default value of the variant `serial` to better match the cmake behavior of Kokkos and adds a constraint to enable at least the serial host backend when no host is enabled.


cedricchevalier19 (approved):
The main effect is to disable `serial` and enable it if no other host backend is enabled.

cedricchevalier19 (approved):

nmm0 (approved):

tpadioleau:
Hi @rbberger, would you have time to look at it ?

rbberger (approved):
LGTM

PR #1384, opened 2025-09-09
tcl/tk: Add versions 8.6.17 and 9.0.2.

KineticTheory:
Adding 2 newer versions of `tcl` and `tk`.

### DIFF
--- a/repos/spack_repo/builtin/packages/tcl/package.py
+++ b/repos/spack_repo/builtin/packages/tcl/package.py
@@ -41,6 +41,8 @@ class Tcl(AutotoolsPackage, NMakePackage, SourceforgePackage):
 
     license("TCL")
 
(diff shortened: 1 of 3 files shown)


KineticTheory:
* `turbine` expects `tcl@8.6`.  Update the recipe to make this explicit.
* There are newer versions of `turbine` that might be compatible.  However, such changes are outside the scope of this MR.

alalazo (changes requested):
There are legitimate failures in pipelines due to the major version bump of `tcl`

KineticTheory:
> There are legitimate failures in pipelines due to the major version bump of `tcl`

I modified the recipes to make 1.8.17 the default version for now.  It seems that many downstream packages aren't aware of the 9.x versions yet.

alalazo (approved):

PR #1386, opened 2025-09-09
ci: fix reviewers workflow for added packages and drafts

### LINKED ISSUE #1369: `triage` action fails on new packages
The step added in #1265 to tag maintainers on PRs fails when a new package is added to the repository.

Examples of failing runs:
- https://github.com/spack/spack-packages/actions/runs/17510261816/job/49740860323?pr=1231
- https://github.com/spack/spack-packages/actions/runs/17567622529/job/49897569449?pr=1367

alecbcs:
Fixes #1369.

### DIFF
--- a/.github/workflows/bin/spack-reviewers.py
+++ b/.github/workflows/bin/spack-reviewers.py
@@ -24,6 +24,10 @@ def main():
     base_url = f"https://api.github.com/repos/{repository}/pulls/{pr_number}"
     pull_request = requests.get(base_url, headers=headers).json()
 
+    if pull_request["draft"]:
+        print(f"[PR #{pr_number}]: skipping draft pull request")
+        return
+
     pull_request_files = requests.get(f"{base_url}/files", headers=headers)
 
     existing_reviewers = {reviewer["login"] for reviewer in pull_request["requested_reviewers"]}
@@ -34,8 +38,11 @@ def main():
             package = match.group(1)
 
             # add maintainers from packages here
-            pkg_maintainers = spack.repo.PATH.get_pkg_class(package).maintainers
-            maintainers.update(pkg_maintainers)
+            try:
+                pkg_maintainers = spack.repo.PATH.get_pkg_class(package).maintainers
+                maintainers.update(pkg_maintainers)
+            except spack.repo.UnknownPackageError:
+                pass
 
     author = pull_request["user"]["login"]
     reviewers = (maintainers | existing_reviewers) - {author}


haampie:
Can you add more defensive checks, and at least print the exception? In particular GitHub API requests may fail.

.github/workflows/bin/spack-reviewers.py
```diff
@@ -34,8 +38,11 @@ def main():
             package = match.group(1)
 
             # add maintainers from packages here
-            pkg_maintainers = spack.repo.PATH.get_pkg_class(package).maintainers
-            maintainers.update(pkg_maintainers)
+            try:
+                pkg_maintainers = spack.repo.PATH.get_pkg_class(package).maintainers
+                maintainers.update(pkg_maintainers)
+            except spack.repo.UnknownPackageError:
```
haampie, line 44:
```suggestion
            except Exception:
```

alecbcs:
> Can you add more defensive checks, and at least print the exception? In particular GitHub API requests may fail.

Yes, but I think that's outside the scope of this PR which is a hot patch to fix the failing checks for new package PRs.

.github/workflows/bin/spack-reviewers.py
alecbcs: Why should we backoff to the exception catchall here? Are there other possible errors in Spack here that we're likely to hit? My preference would be to not catch all so that we're made aware of changes to the core codebase instead of having this check always pass.

kwryankrattiger: I think printing something about the error here would be good. But I agree wtih @alecbcs that catching the specific expected error mode is better here, if this fails for some other unexpected reason it should make some noise.

haampie: Then catch RepoError or SpackError, which can be raised when code loading fails. I'm suggesting a generic exception because it's private api and lately a few exceptions were moved between modules.

haampie:
> Yes, but I think that's outside the scope of this PR which is a hot patch to fix the failing checks for new package PRs.

It's so little work.

.github/workflows/bin/spack-reviewers.py
kwryankrattiger: If it does break, it will only break on a bump PR and it can be fixed there as well. The point is we won't suppress unexpected errors, so if spack updates and creates a new edge case it will be more apparent when it is hit so the check can be appropriately updated.

kwryankrattiger (approved):

PR #1400, opened 2025-09-10
shellcheck: Add versions 0.10.0 and 0.11.0

fbarbari:


alalazo:
Why is this package not building from sources?

alalazo (approved):
This fits the current way of doing things, but it would be better if we built this package from sources.

fbarbari:
Hi @alalazo, thanks for the quick response. Unfortunately, shellcheck is written in haskell and AFAIK even the basic infrastructure for haskell packages is missing in spack. I never used it, but I tried hacking something together to have the GHC compiler and cabal (the build system, i think). One of the problems is that GHC needs another GHC to be built, like many compilers, therefore it needs to exist in the system and spack needs to recognize it (which it doesn't and I don't know how much work it could be).

alalazo:
@fbarbari Could be a nice project for Spack v1.2 to add Haskell as a language :slightly_smiling_face:

PR #1401, opened 2025-09-10
Fix build error with `rdkit@2022_09_5` and python>=3.12

fbarbari:


repos/spack_repo/builtin/packages/rdkit/package.py
```diff
@@ -82,7 +84,7 @@ class Rdkit(CMakePackage):
 
         conflicts("+xyz2mol", when="~yaehmop", msg="XY2MOL requires YAeHMOP")
 
-    depends_on("boost@1.53.0: +python +serialization +iostreams +system")
+    depends_on("boost@1.53.0: +python +serialization +iostreams +system +conversion")
```
tldahlgren, line 87: Looks like `conversion` is only available for version `@1.87.0:`.

tldahlgren:
@bvanessen @RMeli Any objections to the PR being merged as is?

tldahlgren (commented):
Have a slight concern about the reliance on a dependency's variant that is not available at the stated lower bounds. So perhaps another dependency is in order with the variant enabled and the proper lower bound?

repos/spack_repo/builtin/packages/rdkit/package.py
RMeli: I struggle to understand how adding a variant only available for a fairly recent version of boost (released Dec 12, 2024) is the right fix to compile a version of RDKit from 2022. @fbarbari do you have more information about why this variant is needed?

fbarbari: Ah yes sorry, some context is due here. The problem occurred with some compilation errors regarding `boost::numeric_cast` when trying to install `rdkit@2022_09_5` which I think came from spack wanting to install `boost@1.88` since there were no conflicts otherwise. To fix this, I could add the requirement of `boost +conversion` only on recent versions of boost like:
```
depends_on("boost@1.53.0:1.86.0 +python +serialization +iostreams +system")
depends_on("boost@1.87.0: +python +serialization +iostreams +system +conversion")
```

tldahlgren:
@RMeli Are you okay with this PR being merged as is?

RMeli:
@tldahlgren and @fbarbari, sorry, I forgot about this PR. I think the solution is not right, as it stands `rdkit ^boost@1.80` does not concretize:
```
==> Error: failed to concretize `rdkit ^boost@1.80` for the following reasons:
     1. Cannot satisfy 'boost@1.80'
     2. Cannot satisfy 'boost@1.80'
        required because rdkit ^boost@1.80 requested explicitly
     3. Cannot satisfy 'boost@1.53.0:' and 'boost@1.80'
        required because rdkit depends on boost@1.53.0:+conversion+iostreams+python+serialization+system
          required because rdkit ^boost@1.80 requested explicitly
        required because rdkit ^boost@1.80 requested explicitly
```
while removing the `+conversions` variant from boost (as a test) allows concretization again.

PR #1402, opened 2025-09-10
radiuss packages update from radiuss-spack-configs

adrienbernede:
This PR updates radiuss packages with some changes:
- Caliper, RAJA, Umpire, CHAI: Add latest versions
- Care: Update version compatibility
- Camp: Rocm handling fix
- RAJA, RAJAPerf: update c++17 activation


alecbcs:
@adrienbernede would it be possible to break these changes out into a few single package PRs? That would greatly help in reviewing and diagnosing what's breaking in CI.

adrienbernede:
@alecbcs Gitlab CI was passing, waiting for new run. Style fixed.

repos/spack_repo/builtin/packages/umpire/package.py
alecbcs:
```suggestion
    depends_on("fortran", type="build", when="+fortran")  # generated
```

I believe we should only require fortran when building fortran support.

repos/spack_repo/builtin/packages/umpire/package.py
```diff
@@ -341,7 +349,7 @@ def initconfig_compiler_entries(self):
 
         option_prefix = "UMPIRE_" if spec.satisfies("@2022.03.0:") else ""
 
-        if spec.satisfies("+fortran") and compiler.fc is not None:
+        if spec.satisfies("+fortran") and "fortran" in spec:
```
alecbcs:
```suggestion
        if spec.satisfies("+fortran"):
```

When `+fortran` there will always be a fortran compiler so the second statement will always be a noop.

repos/spack_repo/builtin/packages/camp/package.py
```diff
@@ -117,7 +120,8 @@ def cmake_args(self):
 
         options.append(self.define_from_variant("ENABLE_HIP", "rocm"))
         if spec.satisfies("+rocm"):
-            options.append("-DHIP_ROOT_DIR={0}".format(spec["hip"].prefix))
+            rocm_root = dirname(spec["llvm-amdgpu"].prefix)
+            options.append("-DROCM_PATH={0}".format(rocm_root))
```
alecbcs:
```suggestion
            options.append(self.define("ROCM_PATH", rocm_root))
```

Since this is a `CMakePackage` we can use `self.define()` here to generate a `-D` option.

repos/spack_repo/builtin/packages/camp/package.py
```diff
@@ -117,7 +120,8 @@ def cmake_args(self):
 
         options.append(self.define_from_variant("ENABLE_HIP", "rocm"))
         if spec.satisfies("+rocm"):
-            options.append("-DHIP_ROOT_DIR={0}".format(spec["hip"].prefix))
+            rocm_root = dirname(spec["llvm-amdgpu"].prefix)
```
alecbcs: I don't think this'll work when `llvm-amdgpu` is installed by Spack and not an external since `dirname()` will return one level above the Spack prefix which will get you into `$spack/opt/spack/os-arch`.

Can you explain why `spec["llvm-amdgpu"].prefix` doesn't work for your usecase?

alecbcs (changes requested):
Couple of quick suggestions. Otherwise this PR looks mostly good to me.

repos/spack_repo/builtin/packages/umpire/package.py
adrienbernede: I agree. This very change is coming in a follow-up PR (you asked me to split :D).

repos/spack_repo/builtin/packages/camp/package.py
adrienbernede: Here is how we define the external install of `llvm-amdgpu`:
```
    - spec: llvm-amdgpu@6.4.1
      prefix: /opt/rocm-6.4.1/llvm
      extra_attributes:
        compilers:
          c: /opt/rocm-6.4.1/llvm/bin/amdclang
          cxx: /opt/rocm-6.4.1/llvm/bin/amdclang++
          fortran: /opt/rocm-6.4.1/llvm/bin/amdflang
        flags: {}
        environment: {}
        extra_rpaths: [/opt/rocm-6.4.1/lib, /opt/rocm-6.4.1/llvm/lib, /opt/cray/pe/mpich/8.1.32/gtl/lib]
```
As you can see, the prefix is such that we need to get the parent directory to obtain the rocm root directory.
I don't have a clear memory of trying something else, so I am trying now with your proposal. However, it looks like Spack would need to be robust enough to look into the `<prefix>`+`llvm` directory.
Is this directory layout generic for rocm, or specific to LC, I couldn't tell.

adrienbernede: @alecbcs after investigations, the reason why I am doing this change here is simply because... that's what we already have in cached-cmake build system... So I was simply unifying the configs across the radiuss packages.

Your comment is of course still relevant. I’ll run some tests to change the rocm path every where (in cached-cmake as well).

adrienbernede:
```suggestion
            rocm_root = spec["llvm-amdgpu"].prefix
```

adrienbernede:
@alecbcs I think this is ready. Just need to wait for you to be back from SC ;)

alecbcs (approved):
Looks good to me. Thanks @adrienbernede!

PR #1403, opened 2025-09-10
Kokkos: Add conflict for debug_dualview_modify_check

cedricchevalier19:
Newer versions of Kokkos always check dual view usage.

Should fix Kokkos nightly tests.

### DIFF
--- a/repos/spack_repo/builtin/packages/kokkos/package.py
+++ b/repos/spack_repo/builtin/packages/kokkos/package.py
@@ -217,7 +217,7 @@ class Kokkos(CMakePackage, CudaPackage, ROCmPackage):
     }
 
     conflicts("~debug_dualview_modify_check", when="@4.7:")  # always enable from 4.7.00
-    
+
     spack_micro_arch_map = {
         "thunderx2": "THUNDERX2",
         "zen": "ZEN",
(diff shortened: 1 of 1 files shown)


cedricchevalier19:
@spackbot style fix

cedricchevalier19:
@spackbot fix style

nmm0 (approved):

tpadioleau (approved):
LGTM, see also https://github.com/kokkos/kokkos/blob/4.7.01/CHANGELOG.md#incompatibilities-ie-breaking-changes

repos/spack_repo/builtin/packages/kokkos/package.py
```diff
@@ -216,6 +216,8 @@ class Kokkos(CMakePackage, CudaPackage, ROCmPackage):
         "tests": [False, None, "Build for tests"],
     }
 
+    conflicts("~debug_dualview_modify_check", when="@4.7:")  # always enable from 4.7.00
```
alalazo, line 219: Request for a following PR. Can we unroll variant definitions? Currently they are obfuscating the code. We have a dictionary at line 198 for "options" that is not immediately readable. The code using it is at L340 🙁

tpadioleau: We can have a look but one downside I see is that we will have to unroll twice, once for declaring the Spack variants and once for their counterpart of CMake. If you have further suggestion or a better example, please share it with us.

cedricchevalier19: I think it will be a lot more readable to unroll this code. I would rather have a duplication than this automatic translation that is quite difficult to maintain when we want to add conditionals or to not map exactly the name of the CMake option.

tpadioleau: ok we can add this in the todo list

PR #1406, opened 2025-09-10
add versions 0.20.0 and 0.21.0 of mochi-margo

carns:
* adds two new versions of mochi-margo
* adds an hwloc dependency that can be optionally disabled by setting the `hwloc` variant to false


carns:
deprecated by #1676

PR #1409, opened 2025-09-10
gasnet: add more variants to spack package

rbberger:
Changes moved out of https://github.com/spack/spack-packages/pull/1231. @bonachea @PHHargrove please review


bonachea (changes requested):
This PR greatly expands the variant space of the GASNet spackage.
As such, I'm very interested in thoroughly testing and reviewing this PR. 
Unfortunately I'm under water this week and next with the INCITS/J3 Fortran standards meeting, so realistically I might not get to until the first week of November.

In lieu of a "real" review, here are some requests I'd have based on a quick skim:

1. GASNet's documented configure defaults were chosen carefully over time and experience to balance the needs of multiple clients and target systems. In order to satisfy the principle of least surprise for Spack users and package authors, any Spack variants corresponding to a GASNet configure option should have matching default behavior (unless there's a very compelling reason to diverge). Offhand the following variants currently appear to fail this criteria: `pic`, `par`, `pthreads`.
2. Some GASNet configure options default to "smart" detection of available dependencies, which improves the install experience. Corresponding Spack variants should default to *using* that smart detection unless the client asked for a specific variant (again, unless there's a very compelling reason to diverge). IOW the Spack default for these should be `auto` (or equivalent) with no corresponding configure argument. Offhand the following variants currently appear to fail this criteria: `pmi`, `pthreads`.
3. The dependency on `cray-pmi` package is pointless and degrades the casual user experience. That package doesn't actually install anything (so it's just forcing manual addition of an external), and GASNet configure is already capable of finding the system Cray PMI installation without Spack's help.
4. The `segment` variant doesn't appear to be "wired-up" at all?

@PHHargrove may have more available time to help audit places where the Spack variant default behavior doesn't match the GASNet configure defaults.

rbberger:
@bonachea thanks for the initial comments. I'll look at these points in the coming days. I'm expecting we can accommodate some of the automatic detection that should stay. Others will require some duplication in logic within the Spack package. The reason for this is a difference in responsibilities between Spack and the configure script. Some decisions in the Spack concretization we will not be able to postpone until the configure script runs. As carefully crafted that configure script is, some of the problems it solves are done by Spack and its recipes before the script even runs. Therefore its capabilities become less relevant.

rbberger:
@bonachea @PHHargrove here is a new iteration based on your initial feedback.

1 & 2. Many variants now default to `"auto"` where configure has meaningful auto-detection, preserving GASNet's carefully tuned behavior. Some were removed and conflicts were added.
3. The `cray-pmi` dependency for `pmi=cray` is kept because multiple versions may exist. With Spack we can't rely on the process environment or modules. Users will need to specify which via externals or have it auto-detected via `spack external find` (that might be its own PR just for that package, unrelated to this one). `pmi=auto` does not have the dependency.
4. Hooked up segment variant.

I've also added external detection.

bonachea:
@rbberger Thanks for the updates!

This is a very busy week for us, so realistically I won't have time for an in-depth review until next week.

repos/spack_repo/builtin/packages/gasnet/package.py
```diff
@@ -72,8 +84,89 @@ class Gasnet(Package, CudaPackage, ROCmPackage):
...
+    )
+
+    variant(
+        "parsync",
+        default="auto",
+        values=("auto", "on", "off"),
+        description="support PARSYNC-mode pthreaded GASNet clients",
+    )
+
+    variant("seq", default=True, description="support SEQ-mode single-threaded GASNet clients")
```
bonachea: The asymmetry here between the `seq` and `par`/`parsync` variants is a bit inelegant.

The "auto" logic for `seq` is a trivial default to "on", but for symmetry it would be nice if `seq=on`/`seq=off` was accepted in addition to `+seq`/`~seq`.

A possibly better design here would be to expose a single multi-valued `threadmode` variant that defaults to `auto` with values that include `seq,par,parsync`.

repos/spack_repo/builtin/packages/gasnet/package.py
```diff
@@ -122,58 +233,136 @@ def install(self, spec, prefix):
...
-            options.append("--disable-auto-conduit-detect")
-            for c in spec.variants["conduits"].value:
-                options.append("--enable-" + c)
-
-            options.append("--enable-rpath")
-
-            configure(*options)
-            make()
-            make("install")
+    @run_after("install")
```
bonachea: Please restore some version of the comment that was deleted above, which explains the motivation for this very important method.
```suggestion
    # The GASNet-EX library has a highly multi-dimensional configure space,
    # to accomodate the varying behavioral requirements of each client runtime.
    # The library's ABI/link compatibility is strongly dependent on these
    # client-specific build-time settings, and that variability not fully
    # encoded in the variants of this package. The recommended way to build/deploy
    # GASNet is as an EMBEDDED library within the build of the client package
    # (e.g., UPC++, Chapel, etc), some of which provide build-time
    # selection of the GASNet library sources. This spack package provides
    # the GASNet-EX sources, for use by appropriate client packages.
    @run_after("install")
```

repos/spack_repo/builtin/packages/gasnet/package.py
```diff
@@ -4,15 +4,16 @@
...
+from spack_repo.builtin.build_systems.autotools import AutotoolsPackage
 from spack_repo.builtin.build_systems.cuda import CudaPackage
-from spack_repo.builtin.build_systems.generic import Package
 from spack_repo.builtin.build_systems.rocm import ROCmPackage
 
 from spack.package import *
 
 
-class Gasnet(Package, CudaPackage, ROCmPackage):
+class Gasnet(AutotoolsPackage, CudaPackage, ROCmPackage):
```
bonachea: Was there a specific motivation for this change?

I've noticed one side-effect of inheriting from `AutotoolsPackage` is that Spack installs with `--test` now run `make check` which builds ALL the GASNet tests, whereas previously we only built a small subset of the tests for `on_package_attributes(run_tests=True)`.

I don't have a specific objection to this change, but I'm wondering what other hidden side-effects there might be here..

repos/spack_repo/builtin/packages/gasnet/package.py
```diff
@@ -111,8 +211,19 @@ class Gasnet(Package, CudaPackage, ROCmPackage):
...
+
+    conflicts("par=on", when="pthreads=off", msg="PAR mode requires pthreads")
+    conflicts("parsync=on", when="pthreads=off", msg="PARSYNC mode requires pthreads")
+    conflicts(
+        "pshm=on", when="segment=everything", msg="PSHM not compatible with SEGMENT_EVERYTHING"
+    )
+
+    @run_before("configure")
+    def bootstrap(self):
+        if self.spec.satisfies(self.bootstrap_version):
```
bonachea: # Regression

With the inheritance change above from `Package` to `AutotoolsPackage`, this Bootstrap logic now appears to be broken for `bootstrap_version` versions.

In particular, the package change has apparently broken the ability to install GASNet versions from git (e.g. `spack install gasnet@develop`) which now fail the (new) `autoreconf` build step with an error about a missing libtool dependency.  

I'd prefer to avoid adding a dependency on `libtool` that GASNet does not actually require. More importantly, I'd like to preserve the use of GASNet's supported/documented `Bootstrap` script that has a similar effect to `autoreconf` but additionally performs some patching on the Makefile.in and configure output files generated by automake/autoconf.

I _think_ it may be sufficient to make the following change, but I've only done limited testing:

```suggestion
    @run_before("autoreconf")
    def bootstrap(self):
        if self.spec.satisfies(self.bootstrap_version):
```

repos/spack_repo/builtin/packages/gasnet/package.py
```diff
@@ -72,8 +84,89 @@ class Gasnet(Package, CudaPackage, ROCmPackage):
...
+    variant(
+        "pmi",
+        default="auto",
+        values=("auto", "none", "x", "1", "2", "cray"),
+        description="PMI version (auto-detected when needed by conduit)",
+    )
+
+    # ============================================================
+    # True binary choice variants
+    # ============================================================
```
bonachea: This comment is misleading because only the next two variants are boolean.

We should either drop this comment, or add a comment introducing the remaining (non-boolean) variants.

```suggestion
```

repos/spack_repo/builtin/packages/gasnet/package.py
```diff
@@ -122,58 +233,136 @@ def install(self, spec, prefix):
...
-        # encoded in the variants of this package. The recommended way to build/deploy
-        # GASNet is as an EMBEDDED library within the build of the client package
-        # (eg. Berkeley UPC, UPC++, Legion, etc), some of which provide build-time
-        # selection of the GASNet library sources. This spack package provides
-        # the GASNet-EX sources, for use by appropriate client packages.
-        install_tree(".", prefix + "/src")
-
-        # Library build is provided for unit-testing purposes only (see notice above)
-        if "conduits=none" not in spec:
-            options = ["--prefix=%s" % prefix]
```
bonachea, line 138: # Regression

The conditional above is a very important piece of logic that has been deleted, and is not currently reflected in the proposed changes. By design the logic in the current `gasnet` spackage deliberately skips the `configure`/`make`/`make install` phases when `conduits=none`, because there are no binaries to build.

As a result of this change, `spack install gasnet conduits=none` is now broken, and causes `configure` to drop dead with a fatal error. Client packages including `chapel` and `upcxx` rely upon this variant, here are a few examples that have regressed:

1. `spack install gasnet conduits=none`
1. `spack install upcxx +gasnet`
2. `spack install chapel llvm=none comm=gasnet comm_substrate=smp gasnet=spack %c,,cxx=gcc ^gasnet@develop`

**Fixing this regression is blocker for merge.** I have a strong preference for a solution that maintains the existing behavior where `conduits=none` skips the `configure`/`make` steps, because this greatly accelerates the install (relative to installing even a single conduit).

I believe this is a primary reason we we've been using `Package` instead of `AutotoolsPackage`, to ensure we have full (conditional) control over the `configure`/`build`/`install` phases. Reverting that build system change is probably the simplest solution.

repos/spack_repo/builtin/packages/gasnet/package.py
bonachea: Turns out there is a more serious problem with this change, in that it breaks `conduits=none` installs (see comment thread below). This build system change should probably be reverted.

repos/spack_repo/builtin/packages/gasnet/package.py
```diff
@@ -72,8 +84,89 @@ class Gasnet(Package, CudaPackage, ROCmPackage):
...
 
+    variant(
+        "mpi_compat", default=False, description="Enable/disable MPI compatibility in all conduits"
+    )
+    variant(
+        "segment",
+        default="off",
+        values=("off", "fast", "large", "everything"),
+        description="Build GASNet in the FAST/LARGE/EVERYTHING shared segment configuration",
+    )
```
bonachea, line 139: GASNet's segment configuration is a ternary choice (defaulting to "fast"), there is no "off". 

```suggestion
    variant(
        "segment",
        default="fast",
        values=("fast", "large", "everything"),
        description="Build GASNet in the FAST/LARGE/EVERYTHING shared segment configuration",
    )
```

repos/spack_repo/builtin/packages/gasnet/package.py
```diff
@@ -122,58 +233,136 @@ def install(self, spec, prefix):
...
+            elif pmi_val == "cray":
+                options.append(f"--with-pmi-home={spec['cray-pmi'].prefix}")
+        # else: "auto" - let configure and conduits auto-detect PMI needs
+
+        # ============================================================
+        # Segment configuration
+        # ============================================================
+
+        if not spec.satisfies("segment=off"):
+            options.append(f"--enable-segment-{spec.variants['segment'].value}")
```
bonachea:
```suggestion
        options.append(f"--enable-segment-{spec.variants['segment'].value}")
```

repos/spack_repo/builtin/packages/gasnet/package.py
```diff
@@ -72,8 +84,89 @@ class Gasnet(Package, CudaPackage, ROCmPackage):
...
+            default="auto",
+            values=("auto", "ssh", "mpi", "pmi"),
+            description="ofi job spawner",
+        )
+
+    with when("conduits=ibv"):
+        variant(
+            "ibv_max_hcas",
+            default="1",
+            values=(str(x) for x in range(9)),
```
bonachea: The `values` expression above accepts `ibv_max_hcas=0`, which is a meaningless/forbidden value (leading to a `configure` error).

It also forbids `ibv_max_hcas=9`, which is unlikely in practice but not fundamentally invalid. I've personally seen real systems with up to 8 physical HCAs per node. Some IB installations expose multiple "virtual" HCAs for each physical HCA, which can further multiply the number of software HCAs that ibv-conduit needs to support. 

Ideally this variant should accept any positive integer, but Spack seems to only support boolean and string variants, thus might lack a mechanism to correctly express that.

I suggest we compromise by disallowing 0 and raising the max value to 16, which is hopefully sufficient to encompass most production systems:

```suggestion
            values=(str(x) for x in range(1,17)),
```

repos/spack_repo/builtin/packages/gasnet/package.py
```diff
@@ -72,8 +84,89 @@ class Gasnet(Package, CudaPackage, ROCmPackage):
...
+            default="auto",
+            values=("auto", "cxi"),
+            description="Statically configure ofi-conduit for the given provider",
+        )
+        variant(
+            "ofi_spawner",
+            default="auto",
+            values=("auto", "ssh", "mpi", "pmi"),
+            description="ofi job spawner",
+        )
```
bonachea, line 153: There's an asymmetry here. 

If we're going to provide `ofi_spawner` to set `--with-ofi-spawner`, then we should also provide `ibv_spawner` for `--with-ibv-spawner` and `ucx_spawner` for `--with-ucx-spawner` which have the same values/defaults and are equally important to the operation of those conduits.

There's also a `--with-smp-spawner` option, but it has different values and default (and it's less commonly needed), so I would be fine omitting that one for now.

repos/spack_repo/builtin/packages/gasnet/package.py
```diff
@@ -72,8 +84,89 @@ class Gasnet(Package, CudaPackage, ROCmPackage):
...
+        default="off",
+        values=("off", "fast", "large", "everything"),
+        description="Build GASNet in the FAST/LARGE/EVERYTHING shared segment configuration",
+    )
+
+    with when("conduits=ofi"):
+        variant(
+            "ofi_provider",
+            default="auto",
+            values=("auto", "cxi"),
```
bonachea: Please include other valid OFI provider values here that have [documented GASNet support](https://gasnet.lbl.gov/dist-ex/ofi-conduit/README) (or experimental support):

```suggestion
            values=("auto", "generic", "cxi", "psm2", "verbs", "gni", "efa", "sockets", "udp"),
```

repos/spack_repo/builtin/packages/gasnet/package.py
```diff
@@ -235,3 +424,70 @@ def test_testgasnet(self):
...
+                            flags["pshm"] = True
+                        elif line.startswith("#define HAVE_PTHREAD_H 1"):
+                            flags["pthreads"] = True
+                        elif line.startswith("#define GASNET_DEBUG 1"):
+                            flags["debug"] = True
+
+                # Detect threading modes by checking for library files
+                # Check any conduit (use first one found) for mode-specific libraries
+                if lib_dir.exists() and conduits:
+                    conduit = conduits.split(",")
```
bonachea: Thanks for adding this external detection logic, unfortunately it's not working at all for me in testing. 

After manually installing GASNet on my x86_64/Linux system and placing `gasnet_trace` in my `PATH`, I get output like the following:

```
$ spack external find gasnet       
==> Warning: error detecting "gasnet" from prefix [redacted]/bin: [AttributeError: 'list' object has no attribute 'split']
==> No new external packages detected
```

I'm not sure which `split` call is the problem, but I suspect it's this one.

repos/spack_repo/builtin/packages/gasnet/package.py
```diff
@@ -111,8 +211,19 @@ class Gasnet(Package, CudaPackage, ROCmPackage):
...
-        if spec.satisfies(self.bootstrap_version):
+    # ============================================================
+    # Conflicts
+    # ============================================================
+
+    conflicts("par=on", when="pthreads=off", msg="PAR mode requires pthreads")
+    conflicts("parsync=on", when="pthreads=off", msg="PARSYNC mode requires pthreads")
+    conflicts(
+        "pshm=on", when="segment=everything", msg="PSHM not compatible with SEGMENT_EVERYTHING"
+    )
```
bonachea, line 231: There are probably additional semantic conflicts possible with the new variants that are not yet expressed above.

Off the top of my head, `conduits=ucx` conflicts with `segment=everything`. 

There may be others.

bonachea (changes requested):
Hi @rbberger -

Thanks for all the improvements in response to prior feedback!

I finally found some time for in-depth testing/review. Most importantly, I've identified two serious regressions that need to be addressed to avoid breaking client packages. Also, the new external variant detection logic doesn't appear to work in testing. I've also noted several suggestions for improvement and a few more cosmetic concerns detailed below.

Please let me know if any of my feedback is unclear and I'd be happy to provide further elaboration.

rbberger:
@bonachea thanks for the detailed testing and feedback. I think they are all good points and it's important to fix the regressions.

repos/spack_repo/builtin/packages/gasnet/package.py
rbberger: @bonachea one thing I notice with these source installs is that they don't put these files into a project subdirectory, but directly into `<prefix>/src`.

It's probably a good idea to change this to `<prefix>/src/gasnet` because with Spack environment views this might cause collisions. I'm guessing though that this would require changing some other packages that rely on this "simpler" folder structure.

rbberger: (this might be its own PR)

repos/spack_repo/builtin/packages/gasnet/package.py
rbberger: The primary motivation was reusing functionality. But I agree the cons outweigh the benefits. Reverting.

repos/spack_repo/builtin/packages/gasnet/package.py
rbberger: Going for the combined `threadmode` variant, which seems much better.

repos/spack_repo/builtin/packages/gasnet/package.py
rbberger: Adding `ibv_spawner` and `ucx_spawner`.

repos/spack_repo/builtin/packages/gasnet/package.py
rbberger: Instead of a range I'll add a validator that makes sure it's a positive integer.

repos/spack_repo/builtin/packages/gasnet/package.py
rbberger: This should go away with the revert.

rbberger:
@bonachea pushed a new iteration, but I still need to do testing on my end.

repos/spack_repo/builtin/packages/gasnet/package.py
bonachea: > It's probably a good idea to change this to `<prefix>/src/gasnet` because with Spack environment views this might cause collisions. I'm guessing though that this would require changing some other packages that rely on this "simpler" folder structure.

This is an interesting observation, and I see your point. 

Such a change would definitely require corresponding changes to other client packages. However it also seems orthogonal to the goals of this PR, so I'd strongly prefer to defer exploration of that idea to a separate PR.

rbberger:
@bonachea looking good on my end.

repos/spack_repo/builtin/packages/gasnet/package.py
bonachea: Thanks.
I see the new variant for ibv_spawner, but it looks like perhaps you forgot to commit the new `variant` declaration for `ucx_spawner`?

repos/spack_repo/builtin/packages/gasnet/package.py
```diff
@@ -131,14 +271,55 @@ def install(self, spec, prefix):
...
+                flags["cxxflags"].append(self.compiler.cxx_pic_flag)
+
+            for key, value in sorted(flags.items()):
+                if value:
+                    options.append(f"--with-{key}={' '.join(value)}")
+
+            options += self.enable_or_disable("debug", spec.satisfies("+debug"))
+            options += self.enable_or_disable(
+                "seq", spec.satisfies("threadmode=seq") or spec.satisfies("threadmode=auto")
+            )
```
bonachea: I don't understand why this condition includes ` or spec.satisfies("threadmode=auto")`. 

`auto` means no configure argument needs to be passed, which activates GASNet's configure logic that implicitly enables whatever threadmodes are available (which incidentally always includes `seq`).

That being said I don't object to this line, because behaviorally I _think_ it makes no difference.

repos/spack_repo/builtin/packages/gasnet/package.py
```diff
@@ -131,14 +271,55 @@ def install(self, spec, prefix):
...
+            options += self.enable_or_disable("par", spec.satisfies("threadmode=par"))
+            options += self.enable_or_disable("parsync", spec.satisfies("threadmode=parsync"))
```
bonachea: These two lines seem not quite right.

When `threadmode=auto` these lines appear to result in `configure --disable-par --disable-parsync`, which is a prohibition, not automatic detection. For `threadmode=auto` we want these threadmode variants to add nothing to the configure line (preserving auto-detection).

repos/spack_repo/builtin/packages/gasnet/package.py
```diff
@@ -235,3 +430,68 @@ def test_testgasnet(self):
...
+                # Detect threading modes by checking for library files
+                # Check any conduit (use first one found) for mode-specific libraries
+                if lib_dir.exists() and conduits:
+                    for conduit in conduits:
+                        if (lib_dir / f"libgasnet-{conduit}-seq.a").exists():
+                            flags["threadmode"].add("seq")
+                        if (lib_dir / f"libgasnet-{conduit}-par.a").exists():
+                            flags["threadmode"].add("par")
+                        if (lib_dir / f"libgasnet-{conduit}-parsync.a").exists():
+                            flags["threadmode"].add("parsync")
```
bonachea, line 502: This external variant detection logic unfortunately is still not working for me in testing.

After manually installing GASNet on my x86_64/Linux system and placing `gasnet_trace` in my `PATH`, I get output like the following:

```
$ spack external find gasnet 
==> Warning: error detecting "gasnet" from prefix <redacted>/inst-llvm22.1-debug/bin: [AttributeError: 'dict' object has no attribute 'add']
==> No new external packages detected
```
I'm not sure which `add` call is the problem, but I suspect it's these.

bonachea (changes requested):
Hi @rbberger - 

Thanks for the updates, and apologies again for the slow review cycle.
I finally found some time for in-depth testing on this, and I've confirmed the significant regressions identified in my prior review appear to be resolved.

I've added a few additional minor requests on the updated code, which hopefully won't take long to finish up. Otherwise, I think this is close to ready!

repos/spack_repo/builtin/packages/gasnet/package.py
rbberger: The mistake was that it was supposed to be a set.

repos/spack_repo/builtin/packages/gasnet/package.py
rbberger: The options are now not set if `threadmode=auto`.

repos/spack_repo/builtin/packages/gasnet/package.py
rbberger: See above.

repos/spack_repo/builtin/packages/gasnet/package.py
bonachea: Thanks, the error message is resolved in commit b1bb64aee9.

However now I just get the unhelpful output below, and nothing is added to my packages.yaml.
```
$ spack external find gasnet
==> No new external packages detected
```

Does this logic work for you?

What the best way to debug my problem?

rbberger: I'll test the external logic on my end.

repos/spack_repo/builtin/packages/gasnet/package.py
bonachea: Thanks, resolved in b1bb64aee9

repos/spack_repo/builtin/packages/gasnet/package.py
bonachea: Thanks, resolved in b1bb64aee9

repos/spack_repo/builtin/packages/gasnet/package.py
```diff
@@ -170,10 +372,10 @@ def install(self, spec, prefix):
             for c in spec.variants["conduits"].value:
                 testdir = join_path(self.prefix.tests, c)
                 mkdirp(testdir)
-                make("-C", c + "-conduit", "testgasnet-par")
-                install(c + "-conduit/testgasnet", testdir)
-            make("-C", c + "-conduit", "testtools-par")
-            install(c + "-conduit/testtools", self.prefix.tests)
+                make("-C", f"{c}-conduit", "testgasnet-par")
+                make("-C", f"{c}-conduit", "testtools-par")
```
bonachea: Another defect I just noticed: 

If you build with `pthreads=off` these two tests will be unavailable (because PAR mode is configured out) and the attempt to `make` them drops dead with an error message of the form:
```
make[3]: Entering directory '/dev/shm/bonachea/spack-stage/spack-stage-gasnet-2025.8.0-d5vlnlb4x3f3vtet3psc4dszwl5v2xko/spack-src/ofi-conduit'
ERROR: pthreads support was not detected at configure time
       try re-running configure with --enable-pthreads
```

The `testtools` target actually already defaults to building the "most threaded config available", so for that one this patch is sufficient:

```suggestion
                make("-C", f"{c}-conduit", "testgasnet-par")
                make("-C", f"{c}-conduit", "testtools")
```

However the `testgasnet` target defaults to the `-seq` threadmode regardless of pthread availability.

Ideally we'd build target `testgasnet-par` whenever pthreads are available (or an a reasonable approximation when `pthreads` is not `off`), and otherwise build the `-seq` version. As a fallback, we could instead just always build target `testgasnet` and accept a small loss in test coverage:

```suggestion
                make("-C", f"{c}-conduit", "testgasnet")
                make("-C", f"{c}-conduit", "testtools")
```
I'd be fine with this second (simpler) option, since this is only intended as a smoke test and is nothing like a comprehensive test of GASNet anyhow.

bonachea (changes requested):
Just noticed one additional problem.

repos/spack_repo/builtin/packages/gasnet/package.py
rbberger: Found a bug in the external detection. The conduits weren't properly detected. (you can see the error with `spack -d external find ...`)

repos/spack_repo/builtin/packages/gasnet/package.py
rbberger: Applied the simpler option.

repos/spack_repo/builtin/packages/gasnet/package.py
bonachea: Thanks, fix confirmed!
I'm now able to successfully `spack external find gasnet`, and the variant detection appears to be correct (at least for the handful of cases I tested).

For example, here are some auto-detect results from NERSC Perlmutter:
```
  gasnet:
    externals:
    - spec: gasnet@2024.5.0+debug~mpi_compat conduits=mpi,ofi,smp,udp pshm=on pthreads=on segment=fast threadmode=par,seq
      prefix: /global/common/software/m2878/perlmutter/upcxx/2023.9.0-gex-2024.5.0/gnu/PrgEnv-gnu-8.5.0-13.2/gasnet.debug
    - spec: gasnet@2024.5.0~debug~mpi_compat conduits=mpi,ofi,smp,udp pshm=on pthreads=on segment=fast threadmode=par,seq
      prefix: /global/common/software/m2878/perlmutter/upcxx/2023.9.0-gex-2024.5.0/gnu/PrgEnv-gnu-8.5.0-13.2/gasnet.opt
    - spec: gasnet@2025.8.0~debug~mpi_compat conduits=mpi,ofi,smp,udp pshm=on pthreads=on segment=fast threadmode=par,seq
      prefix: /global/common/software/m2878/perlmutter/upcxx/2023.9.0-gex-2025.8.0/gnu/PrgEnv-gnu-8.5.0-13.2/gasnet.opt
    - spec: gasnet@2025.8.0+debug~mpi_compat conduits=mpi,ofi,smp,udp pshm=on pthreads=on segment=fast threadmode=par,seq
      prefix: /global/common/software/m2878/perlmutter/upcxx/2023.9.0-gex-2025.8.0/gnu/PrgEnv-gnu-8.5.0-13.2/gasnet.debug
    - spec: gasnet@2025.8.0+debug~mpi_compat conduits=mpi,ofi,smp,udp pshm=on pthreads=on segment=fast threadmode=par,seq
      prefix: /global/common/software/m2878/perlmutter/upcxx/2025.10.0/gnu/PrgEnv-gnu-8.6.0-14/gasnet.debug
    - spec: gasnet@2025.8.0~debug~mpi_compat conduits=mpi,ofi,smp,udp pshm=on pthreads=on segment=fast threadmode=par,seq
      prefix: /global/common/software/m2878/perlmutter/upcxx/2025.10.0/gnu/PrgEnv-gnu-8.6.0-14/gasnet.opt
    - spec: gasnet@2025.8.1+debug~mpi_compat conduits=mpi,ofi,smp,udp pshm=on pthreads=on segment=fast threadmode=par,seq
      prefix: /global/common/software/m2878/perlmutter/upcxx-cuda-native/bleeding-edge/gnu/PrgEnv-gnu-8.6.0-14/gasnet.debug
    - spec: gasnet@2025.8.1~debug~mpi_compat conduits=mpi,ofi,smp,udp pshm=on pthreads=on segment=fast threadmode=par,seq
      prefix: /global/common/software/m2878/perlmutter/upcxx-cuda-native/bleeding-edge/gnu/PrgEnv-gnu-8.6.0-14/gasnet.opt
```

repos/spack_repo/builtin/packages/gasnet/package.py
```diff
@@ -235,3 +438,69 @@ def test_testgasnet(self):
...
+                        elif line.startswith("#define GASNET_SEGMENT_LARGE 1"):
+                            variants.append("segment=large")
+                        elif line.startswith("#define GASNET_SEGMENT_EVERYTHING 1"):
+                            variants.append("segment=everything")
+                        elif line.startswith("#define GASNETI_PSHM_ENABLED 1"):
+                            flags["pshm"] = True
+                        elif line.startswith("#define HAVE_PTHREAD_H 1"):
+                            flags["pthreads"] = True
+                        elif line.startswith("#define GASNET_DEBUG 1"):
+                            flags["debug"] = True
```
bonachea, line 491: This lacks support for auto-detecting some variants, most notably the GPU variants: `+cuda`,  `+rocm` and `+level_zero`, which also appear in the same gasnet_config.h file (e.g. `GASNETI_MK_CLASS_CUDA_UVA_ENABLED`).

However I do _not_ consider this a blocker for merge, as this could always be added later.

repos/spack_repo/builtin/packages/gasnet/package.py
```diff
@@ -235,3 +438,69 @@ def test_testgasnet(self):
...
+                            flags["threadmode"].add("parsync")
+
+                # Add auto/on/off variants
+                variants.append("pshm=on" if flags.get("pshm") else "pshm=off")
+                variants.append("pthreads=on" if flags.get("pthreads") else "pthreads=off")
+                variants.append(f"threadmode={','.join(flags['threadmode'])}")
+
+                # Add boolean variants
+                variants.append("+debug" if flags.get("debug") else "~debug")
+                variants.append("+mpi_compat" if flags.get("mpi_compat") else "~mpi_compat")
```
bonachea: Where is the `mpi_compat` flag set?

I suspect this is currently missing from the detection, especially as the feature is not advertised in gasnet_config.h. 

We should either:
1. add detection (which requires looking in either a pkgconfig file or the conduit makefile fragment for `^GASNET_SPAWNER_MPI=1`) , or 
2. stop advertising a variant value we did not actually detect.

repos/spack_repo/builtin/packages/gasnet/package.py
```diff
@@ -22,20 +24,31 @@ class Gasnet(Package, CudaPackage, ROCmPackage):
...
 
     homepage = "https://gasnet.lbl.gov"
     url = "https://gasnet.lbl.gov/EX/GASNet-2024.5.0.tar.gz"
     git = "https://bitbucket.org/berkeleylab/gasnet.git"
 
-    maintainers("PHHargrove", "bonachea")
+    maintainers("PHHargrove", "bonachea", "rbberger")
 
     tags = ["e4s", "ecp"]
+    executables = ["^gasnet_trace$"]
```
bonachea, line 54: Tangential to this PR, but I just noticed we're missing a `license` identifier.

Please add the following, consistent with the [GASNet license](https://gasnet.lbl.gov/dist-ex/license.txt).

```suggestion

    license("BSD-3-Clause-LBNL")
```

bonachea (changes requested):
Thanks for the fixes! I've added a few other minor comments below, and I think this should be ready to merge very soon.

Github won't allow me to resolve comment threads on this repo, even comment threads I opened myself..  🚱  I don't suppose you know why?

repos/spack_repo/builtin/packages/gasnet/package.py
rbberger: Added the detection of additional variants.

repos/spack_repo/builtin/packages/gasnet/package.py
rbberger: Add missing detection by looking at configure args.

repos/spack_repo/builtin/packages/gasnet/package.py
rbberger: Added.

repos/spack_repo/builtin/packages/gasnet/package.py
bonachea: The `configure` line is unfortunately not a reliable indicator for detecting `mpi_compat` from (non-Spack) installs of GASNet, because the MPI compatibility setting is often auto-detected. I've just confirmed that currently the autodetect reports `~mpi_compat` for a build where MPI compatibility was implicitly auto-enabled. 

If we really want this variant to be auto-detected, then I think robust detection requires looking elsewhere in the install tree, as mentioned above. The best indicator to look for is whether the line `GASNET_SPAWNER_MPI=1` appears in one or more `lib/pkgconfig/gasnet-*-seq.pc` files (it's only available in some conduits).

However unless you have a use case in mind then it might not be crucial to auto-detect this external variant; it might be simpler to just delete this variant from auto-detect? That seems slightly preferable to often reporting incorrect variant values.

repos/spack_repo/builtin/packages/gasnet/package.py
```diff
@@ -235,3 +440,77 @@ def test_testgasnet(self):
...
+                        elif line.startswith("#define GASNET_SEGMENT_LARGE 1"):
+                            variants.append("segment=large")
+                        elif line.startswith("#define GASNET_SEGMENT_EVERYTHING 1"):
+                            variants.append("segment=everything")
+                        elif line.startswith("#define GASNETI_MK_CLASS_CUDA_UVA_ENABLED 1"):
+                            variants.append("+cuda")
+                        elif line.startswith("#define GASNETI_MK_CLASS_HIP_ENABLED 1"):
+                            variants.append("+rocm")
+                        elif line.startswith("#define GASNETI_MK_CLASS_ZE_ENABLED 1"):
+                            variants.append("+level_zero")
```
bonachea: Thanks! I've manually confirmed the `+cuda` variant is now correctly detected when present.

Unfortunately this logic does not report the absence of CUDA support (`~cuda`) when it's missing.

Perhaps the GPU support booleans should instead be handled in `flags` like the other boolean variants?

bonachea (changes requested):
Thanks for the additional updates, some feedback on the new logic below.

repos/spack_repo/builtin/packages/gasnet/package.py
rbberger: They're default off.

repos/spack_repo/builtin/packages/gasnet/package.py
rbberger: Ok, I'll remove it from detection

repos/spack_repo/builtin/packages/gasnet/package.py
bonachea: Sure, but so is `debug` and variant detection is reporting `~debug` but not `~cuda`.

As an example, here's are two results I just got from auto-detection on NERSC Perlmutter:

```
    - spec: gasnet@2025.8.0+cuda+debug~mpi_compat conduits=mpi,ofi,smp,udp pshm=on pthreads=on segment=fast threadmode=par,seq

    - spec: gasnet@2025.8.0~debug~mpi_compat conduits=mpi,ofi,smp,udp pshm=on pthreads=on segment=fast threadmode=par,seq
```
The first tells me this is a CUDA-enabled debug build.

The second tells me this is a non-debug build, but IIUC tells me nothing about whether or not CUDA support is available.

Am I missing something here?

rbberger: Well, to be consistent I would also have to remove `~debug`, that is true. But I'm not opposed to be more explicit either. I'll add ~cuda, ~rocm, ~level_zero when necessary.

bonachea (approved):
LGTM!

Github annoyingly won't allow me to resolve comment threads, but I think you've satisfactorily resolved all my comments to-date.

Thanks for all your great work on this @rbberger!

junghans (approved):

PR #1412, opened 2025-09-11
fflas-ffpack: new package

d-torrance:
[FFLAS-FFPACK](https://linbox-team.github.io/fflas-ffpack/) is a header-only library for linear algebra over finite fields.  It's a dependency of software such as LinBox and Macaulay2.


tldahlgren (approved):
Confirmed homepage and version sha256.

PR #1414, opened 2025-09-11
qgis: fix building

Chrismarsh:
Fixes building qgis due to a few issues:
-  `libxml` had a missing a define for `+http` when building with cmake
- `libxml` has deprecated the `http` feature, so this tidies up the `libxml` depend in `libspatiallite`
- `qscintilla`'s complex post-install patching fails when paths are internally symlinked but are written to the file as realpaths. This now tries to patch out both
- there is a complex interaction of `py-sip` and `py-pyqt5-sip` and `py-pyqt5` such that with the versions in spack currently, `qscintialla` does not import cleanly due to an ABI mismatch. Error below for anyone searching:

```
 python -c "from PyQt5.Qsci import QSCINTILLA_VERSION_STR; print(QSCINTILLA_VERSION_STR)"
  File "<string>", line 1, in <module>
    from PyQt5.Qsci import QSCINTILLA_VERSION_STR; print(QSCINTILLA_VERSION_STR)
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
RuntimeError: the sip module implements API v12.0 to v12.13 but the PyQt5.Qsci module requires API v12.15
```
Updating all the various `*-sip` dependencies resolves this problem.
- qgis needs a patch to work with newish `py-sip` versions and this applies the back-ported patch

Maintainers:
@adamjstewart @Sinan81 @AlexanderRichert-NOAA

### DIFF
--- a/repos/spack_repo/builtin/packages/libspatialite/package.py
+++ b/repos/spack_repo/builtin/packages/libspatialite/package.py
@@ -45,10 +45,11 @@ class Libspatialite(AutotoolsPackage):
     depends_on("iconv")
     depends_on("librttopo", when="@5.0.1:")
 
-    # in libxml2 2.15+ http support is completely removed
+    # in libxml2 2.15+ http support is completely removed, so this will need
+    # to be refined when libspatiallite is updated
     # https://www.gaia-gis.it/fossil/libspatialite/tktview?name=e8f33aa9d8
     # https://www.gaia-gis.it/fossil/libspatialite/tktview?name=ac85f0fca3
-    depends_on("libxml2@:2.14 +http", when="@:5.1.0")
+    depends_on("libxml2 +http")
     depends_on("minizip", when="@5.0.0:")
     depends_on("proj")
     depends_on("proj@:5", when="@:4")
--- a/repos/spack_repo/builtin/packages/libspatialite/package.py
+++ b/repos/spack_repo/builtin/packages/libspatialite/package.py
@@ -44,7 +44,11 @@ class Libspatialite(AutotoolsPackage):
     depends_on("geos@:3.9", when="@:5.0.0")
     depends_on("iconv")
     depends_on("librttopo", when="@5.0.1:")
-    depends_on("libxml2+http")
+
+    # in libxml2 2.15+ http support is completely removed
+    # https://www.gaia-gis.it/fossil/libspatialite/tktview?name=e8f33aa9d8
+    # https://www.gaia-gis.it/fossil/libspatialite/tktview?name=ac85f0fca3
+    depends_on("libxml2@:2.14 +http", when="@:5.1.0")
     depends_on("minizip", when="@5.0.0:")
     depends_on("proj")
     depends_on("proj@:5", when="@:4")
--- a/repos/spack_repo/builtin/packages/qscintilla/package.py
+++ b/repos/spack_repo/builtin/packages/qscintilla/package.py
@@ -78,9 +78,10 @@ class Qscintilla(QMakePackage):
         )
         # if prefix includes symlinks, the realpath is what was written to the make file
         makefile.filter(
-            "$(INSTALL_ROOT)" + os.path.realpath(self.spec["qmake"].prefix), "$(INSTALL_ROOT)", string=True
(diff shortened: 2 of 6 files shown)


Chrismarsh:
@spackbot fix style

repos/spack_repo/builtin/packages/libspatialite/package.py
```diff
@@ -44,7 +44,11 @@ class Libspatialite(AutotoolsPackage):
     depends_on("geos@:3.9", when="@:5.0.0")
     depends_on("iconv")
     depends_on("librttopo", when="@5.0.1:")
-    depends_on("libxml2+http")
+
+    # in libxml2 2.15+ http support is completely removed
+    # https://www.gaia-gis.it/fossil/libspatialite/tktview?name=e8f33aa9d8
+    # https://www.gaia-gis.it/fossil/libspatialite/tktview?name=ac85f0fca3
+    depends_on("libxml2@:2.14 +http", when="@:5.1.0")
```
adamjstewart: Why the upper bound on libspatialite version? Has this been fixed upstream? Is libxml2 no longer a dependency?

repos/spack_repo/builtin/packages/qscintilla/package.py
```diff
@@ -76,6 +76,12 @@ def fix_install_path(self):
         makefile.filter(
             "$(INSTALL_ROOT)" + self.spec["qmake"].prefix, "$(INSTALL_ROOT)", string=True
         )
+        # if prefix includes symlinks, the realpath is what was written to the make file
```
adamjstewart, line 79: Is the previous line still needed? E.g., if there are no symlinks, are both lines identical?

adamjstewart (commented):
@Sinan81 is the expert on these packages, I've only dabbled in most of them.

Sinan81:
I'll provide a review sometime next week.

repos/spack_repo/builtin/packages/libspatialite/package.py
Chrismarsh: Based on the issue tracker, it sounds like next release of `libspatiallite` will have the http requirement dropped. I see what you mean though, I've changed it to let libxml2 control what version it surfaces for that variant

repos/spack_repo/builtin/packages/qscintilla/package.py
Chrismarsh: I don't know. I had previously installed qgis a while ago on this machine, so spack's behaviour has changed in the past ~year regarding how the prefix is resolved. It felt safer to try both.

Sinan81: is there an example where qmake prefix is a symlink dir?
I'm ok with keeping both filter statements.
In a future MR it can be useful to filter *.pro files directly to replace the variables like $$QT_INSTALL...
```
$ grep path qscintilla.pro
    QMAKE_POST_LINK += install_name_tool -id @rpath/$(TARGET1) $(TARGET)
INCLUDEPATH += . ../scintilla/include ../scintilla/lexlib ../scintilla/src
target.path = $$[QT_INSTALL_LIBS]
header.path = $$[QT_INSTALL_HEADERS]
trans.path = $$[QT_INSTALL_TRANSLATIONS]
qsci.path = $$[QT_INSTALL_DATA]
features.path = $$[QT_HOST_DATA]/mkspecs/features
```
alternatively these paths could be defined during `qmake` phase of the build I think.

Sinan81 (approved):
looks good to me. built qgis and tested that it works.

Sinan81:
@Chrismarsh I would recommend adding yourself to the maintainers list in the package.

Chrismarsh:
@Sinan81 great to hear this worked for you! I've added myself.

Sinan81 (approved):

adamjstewart (approved):

repos/spack_repo/builtin/packages/qscintilla/package.py
Chrismarsh: In this case my project dir is symlinks but the realpath was written. It strikes me as a non obvious case and I don't have it in me to test. I will try to look at it again in the future and see if it's needed or do as you propose with the qmake

PR #1418, opened 2025-09-11
py-keyrings-alt: add v5.0.2 and add new dep packages

manuelakuhn:
https://github.com/jaraco/keyrings.alt/tree/v5.0.2

`py-keyrings-alt` has a new dependency on `py-jaraco-context` which again depends on `py-backports-tarfile`. So these 2 new packages were added as well.
https://github.com/jaraco/jaraco.context/tree/v6.0.1
https://github.com/jaraco/backports.tarfile/tree/v1.2.0


michaelkuhn (approved):

PR #1426, opened 2025-09-11
py-mne: add v1.10.1 and py-h5io: add v0.2.5

manuelakuhn:
https://github.com/mne-tools/mne-python/tree/v1.10.1
Reordered dependency to match `pyproject.toml` for easier reviewing

`py-mne+hdf5` also needs a newer version of `py-h5io`, so this is updated as well.
https://github.com/h5io/h5io/tree/h5io-0.2.5


ChristopherChristofi (approved):
Looks good to me and builds successfully

alalazo (approved):

PR #1429, opened 2025-09-11
kokkos: support clang as CMake CUDA/HIP compiler

rbberger:
So far `nvcc`/`amdclang++` have been set as `CMAKE_CUDA_COMPILER`and `CMAKE_HIP_COMPILER` if CMake Language support is enabled (+cuda_lang).

However, some projects require both CMake language support and need to use Clang as CUDA/HIP compiler. This still doesn't cover all the cases (such as clang+nvcc), which will hopefully be possible once Spack learns about CUDA and HIP language requirements.

~Due to how the CMake CUDA compiler detection logic operates, we need to add CMAKE_CUDA_FLAGS to make sure we are building for a new enough C++ and for the right target device to pass the initial checks.~

### DIFF
--- a/repos/spack_repo/builtin/packages/kokkos/package.py
+++ b/repos/spack_repo/builtin/packages/kokkos/package.py
@@ -423,21 +423,39 @@ class Kokkos(CMakePackage, CudaPackage, ROCmPackage):
             options.append(self.define("CMAKE_CXX_COMPILER", self.kokkos_cxx))
         elif "+rocm" in self.spec:
             if "+cmake_lang" in self.spec:
+                if self.spec.satisfies("%cxx=clang"):
+                    options.append(self.define("CMAKE_HIP_COMPILER", self.compiler.cxx))
+                else:
+                    options.append(
+                        self.define(
+                            "CMAKE_HIP_COMPILER",
+                            join_path(self.spec["llvm-amdgpu"].prefix.bin, "amdclang++"),
+                        )
+                    )
+                options.append(from_variant("CMAKE_HIP_STANDARD", "cxxstd"))
                 options.append(
                     self.define(
-                        "CMAKE_HIP_COMPILER",
-                        join_path(self.spec["llvm-amdgpu"].prefix.bin, "amdclang++"),
+                        "CMAKE_HIP_ARCHITECTURES", self.spec.variants["amdgpu_target"].value
                     )
                 )
-                options.append(from_variant("CMAKE_HIP_STANDARD", "cxxstd"))
+                options.append(self.define("CMAKE_HIP_EXTENSIONS", False))
             else:
                 options.append(self.define("CMAKE_CXX_COMPILER", self.spec["hip"].hipcc))
             options.append(self.define("Kokkos_ENABLE_ROCTHRUST", True))
         elif "+cuda" in self.spec and "+cmake_lang" in self.spec:
+            if self.spec.satisfies("%cxx=clang"):
+                options.append(self.define("CMAKE_CUDA_COMPILER", self.compiler.cxx))
+            else:
+                options.append(
+                    self.define(
+                        "CMAKE_CUDA_COMPILER", join_path(self.spec["cuda"].prefix.bin, "nvcc")
+                    )
+                )
             options.append(
-                self.define("CMAKE_CUDA_COMPILER", join_path(self.spec["cuda"].prefix.bin, "nvcc"))
+                self.define("CMAKE_CUDA_ARCHITECTURES", self.spec.variants["cuda_arch"].value)
             )
             options.append(from_variant("CMAKE_CUDA_STANDARD", "cxxstd"))
+            options.append(self.define("CMAKE_CUDA_EXTENSIONS", False))
 
         if self.spec.satisfies("%oneapi") or self.spec.satisfies("%intel"):
             options.append(self.define("CMAKE_CXX_FLAGS", "-fp-model=precise"))


repos/spack_repo/builtin/packages/kokkos/package.py
```diff
@@ -521,21 +521,47 @@ def cmake_args(self):
...
             options.append(self.define("Kokkos_ENABLE_ROCTHRUST", True))
         elif "+cuda" in self.spec and "+cmake_lang" in self.spec:
+            if self.spec.satisfies("%cxx=clang"):
+                # make CMake CUDA compiler detection happy
+                options.append(self.define("CMAKE_CUDA_COMPILER", self.compiler.cxx))
+                options.append(
+                    self.define(
+                        "CMAKE_CUDA_FLAGS",
+                        f"-std=c++{self.spec.variants['cxxstd'].value} "
+                        f"--cuda-gpu-arch=sm_{self.spec.variants['cuda_arch'].value}",
```
tpadioleau: The CMake options `CMAKE_CUDA_ARCHITECTURES` and `CMAKE_CUDA_STANDARD` are not enough ?

rbberger: Yes. At least with Cmake 3.30 I ran into the situation where CUDA compiler detection fails, even if I set those two. But it might be worth looking into if that was a bug that got fixed in later versions.

repos/spack_repo/builtin/packages/kokkos/package.py
```diff
@@ -521,21 +521,47 @@ def cmake_args(self):
...
+                        "CMAKE_HIP_ARCHITECTURES", self.spec.variants["amdgpu_target"].value
                     )
                 )
-                options.append(from_variant("CMAKE_HIP_STANDARD", "cxxstd"))
+                options.append(self.define("CMAKE_HIP_EXTENSIONS", False))
             else:
                 options.append(self.define("CMAKE_CXX_COMPILER", self.spec["hip"].hipcc))
             options.append(self.define("Kokkos_ENABLE_ROCTHRUST", True))
         elif "+cuda" in self.spec and "+cmake_lang" in self.spec:
+            if self.spec.satisfies("%cxx=clang"):
```
cedricchevalier19, line 446: I am ok for now as it is protected by the `+cmake_lang`

repos/spack_repo/builtin/packages/kokkos/package.py
cedricchevalier19: Can we confirm that these options are needed?

cedricchevalier19 (approved):
Except the strange usefulness duplication of cuda arch (that may be necessary), I think it is a great addition.

repos/spack_repo/builtin/packages/kokkos/package.py
tpadioleau: As mentionned on Slack, I could make it work without the extra flags, I have used Spack 1.0.0 with your branch of packages and the following packages:
- clang 20
- cuda 12.3 (latest fully supported)
- cmake 3.30.9, cuda_arch=80

rbberger: I'll take another look

rbberger: @tpadioleau did another take on this. and yes it seems to work without the extra flags. I'll remove them to get this PR in. If I come across my failure again I'll investigate it again.

repos/spack_repo/builtin/packages/kokkos/package.py
rbberger: @cedricchevalier19 See other thread. I've removed them.

repos/spack_repo/builtin/packages/kokkos/package.py
```diff
@@ -423,21 +423,39 @@ def cmake_args(self):
...
+                options.append(from_variant("CMAKE_HIP_STANDARD", "cxxstd"))
                 options.append(
                     self.define(
-                        "CMAKE_HIP_COMPILER",
-                        join_path(self.spec["llvm-amdgpu"].prefix.bin, "amdclang++"),
+                        "CMAKE_HIP_ARCHITECTURES", self.spec.variants["amdgpu_target"].value
                     )
                 )
-                options.append(from_variant("CMAKE_HIP_STANDARD", "cxxstd"))
+                options.append(self.define("CMAKE_HIP_EXTENSIONS", False))
```
tpadioleau, line 441: Still curious, did you encounter any issue without setting this variable ?

repos/spack_repo/builtin/packages/kokkos/package.py
```diff
@@ -423,21 +423,39 @@ def cmake_args(self):
...
+                            "CMAKE_HIP_COMPILER",
+                            join_path(self.spec["llvm-amdgpu"].prefix.bin, "amdclang++"),
+                        )
+                    )
+                options.append(from_variant("CMAKE_HIP_STANDARD", "cxxstd"))
                 options.append(
                     self.define(
-                        "CMAKE_HIP_COMPILER",
-                        join_path(self.spec["llvm-amdgpu"].prefix.bin, "amdclang++"),
+                        "CMAKE_HIP_ARCHITECTURES", self.spec.variants["amdgpu_target"].value
```
tpadioleau, line 438: Out of curiosity, did you encounter any issue without setting this variable ?

tpadioleau (approved):
lgtm

repos/spack_repo/builtin/packages/kokkos/package.py
rbberger: No. This was mainly to make it consistent and follow the recommended way of building HIP code with CMake.

repos/spack_repo/builtin/packages/kokkos/package.py
rbberger: Same. Not really. I'm just generally opposed to allowing GNU extensions if the regular standard gives us what we need. Let me know if you want to stay conservative on this and I'll remove it again. If I actually encounter a real issue on it would then reintroduce it.

repos/spack_repo/builtin/packages/kokkos/package.py
tpadioleau: Fine with me, thanks

repos/spack_repo/builtin/packages/kokkos/package.py
tpadioleau: > Let me know if you want to stay conservative on this and I'll remove it again.

I am fine with it, I am not aware of any performance gain related to extensions of compilers.

alalazo (approved):

PR #1430, opened 2025-09-11
mpt: fix for build environment

mathomp4:
Testing at NASA NAS (which uses MPT) found that the `mpt` module needed an update for `setup_dependent_build_environment`. This PR "borrows" what is being done in `openmpi` and applies to `mpt`. It seems to allow builds


mathomp4:
Note: This was needed for spack v1. The old package worked fine in the old spack v0 days, but apparently something changed

alalazo (approved):

PR #1438, opened 2025-09-12
py_cftime: Add build dependency on 'c'

AlexanderRichert-NOAA:
py-cftime uses a C compiler. This PR adds that dependency so $CC is set properly.


AlexanderRichert-NOAA:
@spackbot fix style

alalazo (approved):

PR #1444, opened 2025-09-12
Add cray mpich GTL by default if GPT variant specified

etiennemlb:
Add the GTL as a library to be linked against, by default IIF an accelerator variant has been specified.

The current implementation puts the burden of supporting the GTL on top of the package maintainer. Now, the GTL should trickle down to cray mpich consumers, if they propagate the +rocm or +cuda variant to cray mpich.

@spackbot fix style


etiennemlb:
@spackbot fix style

etiennemlb:
This does not resolve the fact that cmake or configure files are not looking for the gtl, as such, we need the spack package to inject the gtl library themselves.

@haampie Do you know of a way of injecting flags from a dependent package into a parent ? Such as specifying or "populating" the flag_handler of a parent package ?

Closing this for now.

PR #1450, opened 2025-09-12
Casacore support for v 3.7.1

preshanth:
- Bumped up the casacore version number to 3.7.1. 
- Removed the boost dependency for dysco as it is not needed. 
- Patch is run optionally i.e. the cmake file is removed only if found. Recent casacore versions have removed that file in favor of standard cmake find.

These are general improvements. I am able to build casacore using `spack` with the updated versions. I have tested it in a local repo with the current changes.


repos/spack_repo/builtin/packages/casacore/package.py
```diff
@@ -74,7 +75,7 @@ class Casacore(CMakePackage):
     depends_on("python@2.6:", when="@:3.5.0 +python")
     depends_on("python@3:", when="@3.6.0: +python")
     depends_on("boost +python", when="+python")
-    depends_on("boost +system +filesystem", when="+dysco")
+    # depends_on("boost +system +filesystem", when="+dysco")
```
mpokorny: You need a more descriptive `when` clause, as the dependency still exists when building tests, afaict.

Also, this dependency still exists for older versions, so an additional `when` term is needed.

Perhaps this works
```python
depends_on("boost +system +filesystem", type=("build","link"), when="@:3.6 +dysco")
depends_on("boost +system +filesystem", type="test", when="@3.7: +dysco")
```

repos/spack_repo/builtin/packages/casacore/package.py
```diff
@@ -19,6 +19,7 @@ class Casacore(CMakePackage):
 
     license("LGPL-2.0-only")
 
+    version("3.7.1", sha256="ad7ad865d5d8517524bdde71e0083a69442648b693850660dadb0ce6109d91c7")
```
mpokorny, line 22: Might as well add 3.7.0 as well while you're at it.

mpokorny (changes requested):
A few simple changes so that the file works for all casacore versions.

repos/spack_repo/builtin/packages/casacore/package.py
preshanth: Yeah I put that in. Thanks for pointing that out.

repos/spack_repo/builtin/packages/casacore/package.py
preshanth: I am not sure how to build with testing on to check if boost is still needed. I need casacore measures for testing right ? Would it make sense to put in a download for it from here should `BUILD_TESTING=ON`. It is turned off by default. Let me know your thoughts. I think its a valid suggestion but I can't test it out.

mpokorny: Last time that I tried it, you can at least run the tests that don't depend on casacore measures. Try `spack install casacore --test root` -- although it will probably fail on the tests that do require measures, you can confirm whether or not the build completes, which is good enough since it's no worse than the current support for running tests.

Anyway, https://github.com/casacore/casacore/blob/83ef98a360d164f985b9085d084c1639eae4017a/tables/Dysco/CMakeLists.txt#L43-L58 indicates where the dependency comes in. If you have a look v3.6 too, you'll see how it changed.

There is a `casacore-measures` package, btw. However, that package doesn't set up a run, build or test environment that configures things to point to the installed measures; that would be nice to have, but I haven't implemented it yet.

preshanth: The cmake makes it obvious. Have made the change as you suggested. Will test and then resolve the conversation here. I saw that you have a casacore-measure package. I can attempt a modified version that can be installed and found here if thats something you are interested in. Will put in a different PR for that.

mpokorny: OK, once you resolve this, I will approve the PR.

I'm happy to help with the casacore-measures work so we can add it as a test dependency here. But you should decide whether it's worth the effort. It's not a big priority, so there's no need to spend a lot of time on it.

preshanth: I have checked that it installs with the suggested fix for both 3.7.1 and 3.6. I pushed in the change too.

mpokorny (approved):
lgtm

mpokorny (approved):

alalazo (approved):

PR #1454, opened 2025-09-12
mapl: add commit hashes for esma_cmake resources

AlexanderRichert-NOAA:
This PR adds commit hashes for the esma_cmake resources in the mapl recipe so that they can be prefetched and mapl can be built offline.


mathomp4:
@AlexanderRichert-NOAA Huh. I didn't know this was needed. My apologies. I'll pull your changes here into my other PR #1782 in case that gets in first.

AlexanderRichert-NOAA:
Thanks-- It's needed for Spack to automatically download it when running `spack fetch`, which we want in cases where we're installing on non-internet connected nodes.

mathomp4 (approved):

PR #1467, opened 2025-09-14
Celeritas: new versions 0.6.1, 0.6.2

sethrj:
Minor updates to Celeritas, no compatibility changes.


wdconinc (commented):
Only comment is the new CMake minimum versions when +covfie.

sethrj:
@wdconinc I incorporated your suggestion and also added another patch release to Celeritas. Thanks for your review!

wdconinc (approved):
Looks good to me.

PR #1474, opened 2025-09-15
Revert "lzo: ensure #include <lzo/lzo1x.h> works with pkgconf cflags `-I${includedir}/lzo`"

adamjstewart:
Reverts spack/spack-packages#75

Symlink results in infinite filesystem recursion, resulting in failures in our gitlab-runners repo: https://github.com/spack/gitlab-runners/actions/runs/17732188924/job/50385553346?pr=67

A better solution would be to either patch lzo/cairo/squashfuse or to symlink all `*.h` files instead of symlinking the directory. This can be done in a follow-up PR.


wdconinc (approved):
Looks good to me. I'll see what is depending on this and fix there.

PR #1481, opened 2025-09-15
at_spi2_atk: Remove max version allowed for depends_on(at-spi2-core)

KineticTheory:
# Summary

* I'm trying to develop some new spack packages that require a newer version of `at-spi2-core`.  This requires the removal of the upper limit in the command `depends_on("at-spi2-core@2.28.0:2.45.1")`.
* Does anyone know why the upper limit was listed previously?
* I didn't experience any negative side effects of removing this upper limit, but my testing is far from exhaustive.

### DIFF
--- a/repos/spack_repo/builtin/packages/at_spi2_atk/package.py
+++ b/repos/spack_repo/builtin/packages/at_spi2_atk/package.py
@@ -26,7 +26,7 @@ class AtSpi2Atk(MesonPackage):
     depends_on("c", type="build")  # generated
 
     depends_on("pkgconfig", type="build")
-    depends_on("at-spi2-core@2.28.0:2.45.1")
+    depends_on("at-spi2-core@2.28.0:")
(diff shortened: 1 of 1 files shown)


KineticTheory:
@michaelkuhn If you can find time, I would like your thoughts on this change as well.

michaelkuhn:
Did you see https://github.com/spack/spack/pull/47026? This is the PR that introduced the version constraint, apparently due to issues in GTK and its dependencies.

KineticTheory:
> Did you see [spack/spack#47026](https://github.com/spack/spack/pull/47026)? This is the PR that introduced the version constraint, apparently due to issues in GTK and its dependencies.

Thanks for pointing this out.  Let me see if I can find a work around for the version issue I'm hitting.  That is, one that doesn't require a newer version of `at_spi2_core`.

KineticTheory:
I found another way to fix my version issue. Closing...

PR #1482, opened 2025-09-15
py-aiohttp: add v3.12.15

cmelone:
adds new constraints found in https://github.com/aio-libs/aiohttp/blob/v3.12.15/setup.cfg and fixes a couple others that were missed in previous recipes.

also adds a new version of dependency py-aiosignal

### DIFF
--- a/repos/spack_repo/builtin/packages/py_aiohttp/package.py
+++ b/repos/spack_repo/builtin/packages/py_aiohttp/package.py
@@ -19,6 +19,7 @@ class PyAiohttp(PythonPackage):
 
     license("Apache-2.0")
 
+    version("3.12.15", sha256="4fc61385e9c98d72fcdf47e6dd81833f47b2f77c114c29cd64a361be57a763a2")
     version("3.11.16", sha256="16f8a2c9538c14a557b4d309ed4d0a7c60f0253e8ed7b6c9a2859a7582f8b1b8")
     version("3.9.5", sha256="edea7d15772ceeb29db4aff55e482d4bcfb6ae160ce144f2682de02f6d693551")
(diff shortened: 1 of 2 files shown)


repos/spack_repo/builtin/packages/py_aiohttp/package.py
```diff
@@ -35,20 +36,22 @@ class PyAiohttp(PythonPackage):
...
         depends_on("py-attrs@17.3.0:")
         depends_on("py-charset-normalizer@2:3", when="@3.8.4:")
         depends_on("py-charset-normalizer@2", when="@3.8.0:3.8.3")
         depends_on("py-multidict@4.5:6", when="@3.6.3:")
-        depends_on("py-async-timeout@4:", when="@3.8.0 ^python@:3.10")
+        depends_on("py-async-timeout@4:5", when="@3.8: ^python@:3.10")
         depends_on("py-async-timeout@3", when="@:3.7.4 ^python@:3.10")
         depends_on("py-asynctest@0.13.0", when="@3.8.0: ^python@:3.7")
-        depends_on("py-yarl@1.17:", when="@3.11:")
+        depends_on("py-yarl@1.17:1", when="@3.11:")
```
kwryankrattiger: I am curious why this isn't covered by the constraint below this.

cmelone: good point, I guess I was trying to write `≥1.17 < 2.0`...what's the best way to express that?

becker33: You expressed it correctly, it's just that the `<2 constraint is already enforced by the following line (for all versions) so there's no need to separately enforce it for `when=@3.11:`

cmelone: right, thanks

kwryankrattiger (approved):

PR #1483, opened 2025-09-15
chapel: add 2.6

arifthpe:
Add `chapel` version 2.6.

Also includes:
- Require `@2.2:` and `llvm=bundled` for ROCm support, to work around fallout from https://github.com/spack/spack-packages/commit/2bf4ab9585c8d483cc8581d65912703d3f020393 and Chapel's requirement to use patched bundled LLVM for ROCm 6.
- Deprecate versions `@2.0:2.3`

### DIFF
--- a/repos/spack_repo/builtin/packages/chapel/package.py
+++ b/repos/spack_repo/builtin/packages/chapel/package.py
@@ -605,6 +605,9 @@ class Chapel(AutotoolsPackage, CudaPackage, ROCmPackage):
 
     def build(self, spec, prefix):
         with set_env(CHPL_MAKE_THIRD_PARTY=join_path(self.build_directory, "third-party")):
+            # print all the explicitly set CHPL_* config variables to assist in debugging
+            for var in self.chpl_env_vars:
+                tty.info(var + '=' + os.getenv(var, '<unset>'))
             make()
             with set_env(CHPL_HOME=self.build_directory):
                 if spec.satisfies("+chpldoc"):
--- a/repos/spack_repo/builtin/packages/chapel/package.py
+++ b/repos/spack_repo/builtin/packages/chapel/package.py
@@ -607,7 +607,7 @@ class Chapel(AutotoolsPackage, CudaPackage, ROCmPackage):
         with set_env(CHPL_MAKE_THIRD_PARTY=join_path(self.build_directory, "third-party")):
             # print all the explicitly set CHPL_* config variables to assist in debugging
             for var in self.chpl_env_vars:
-                tty.info(var + '=' + os.getenv(var, '<unset>'))
+                tty.info(var + "=" + os.getenv(var, "<unset>"))
             make()
             with set_env(CHPL_HOME=self.build_directory):
                 if spec.satisfies("+chpldoc"):
--- a/repos/spack_repo/builtin/packages/chapel/package.py
+++ b/repos/spack_repo/builtin/packages/chapel/package.py
@@ -118,7 +118,7 @@ class Chapel(AutotoolsPackage, CudaPackage, ROCmPackage):
         "dpcpp": "intel",
         "gcc": "gnu",
         "intel": "intel",
-        "llvm": "llvm",
+        "llvm": "clang",
         "oneapi": "intel",
         "rocmcc": "clang",
         "unset": "unset",
--- a/repos/spack_repo/builtin/packages/chapel/package.py
+++ b/repos/spack_repo/builtin/packages/chapel/package.py
@@ -489,8 +489,10 @@ class Chapel(AutotoolsPackage, CudaPackage, ROCmPackage):
     with when("@2.2: +rocm"):
         depends_on("hsa-rocr-dev@6.0:6.2")
         depends_on("hip@6.0:6.2")
-    # This is the case because the package only supports ROCm 6, and Chapel requires bundled LLVM for that version.
-    # TODO: Modify this constrant and message if/when Chapel supports an additional ROCm version without that requirement.
+    # This is the case because the package only supports ROCm 6, and Chapel
+    # requires bundled LLVM for that version.
+    # TODO: Modify this constrant and message if/when Chapel supports an
+    # additional ROCm version without that requirement.
     requires("llvm=bundled", when="+rocm", msg="Chapel ROCm support requires llvm=bundled")
 
     conflicts(
--- a/repos/spack_repo/builtin/packages/chapel/package.py
+++ b/repos/spack_repo/builtin/packages/chapel/package.py
@@ -66,11 +66,13 @@ class Chapel(AutotoolsPackage, CudaPackage, ROCmPackage):
     version("2.6.0", sha256="e469c35be601cf1f59af542ab885e8a14aa2b087b79af0d5372a4421976c74b6")
     version("2.5.0", sha256="020220ca9bf52b9f416e9a029bdc465bb1f635c1e274c6ca3c18d1f83e41fce1")
     version("2.4.0", sha256="a51a472488290df12d1657db2e7118ab519743094f33650f910d92b54c56f315")
-    version("2.3.0", sha256="0185970388aef1f1fae2a031edf060d5eac4eb6e6b1089e7e3b15a130edd8a31")
-    version("2.2.0", sha256="bb16952a87127028031fd2b56781bea01ab4de7c3466f7b6a378c4d8895754b6")
-    version("2.1.0", sha256="72593c037505dd76e8b5989358b7580a3fdb213051a406adb26a487d26c68c60")
-    version("2.0.1", sha256="19ebcd88d829712468cfef10c634c3e975acdf78dd1a57671d11657574636053")
-    version("2.0.0", sha256="b5387e9d37b214328f422961e2249f2687453c2702b2633b7d6a678e544b9a02")
+
+    with default_args(deprecated=True):
+        version("2.3.0", sha256="0185970388aef1f1fae2a031edf060d5eac4eb6e6b1089e7e3b15a130edd8a31")
+        version("2.2.0", sha256="bb16952a87127028031fd2b56781bea01ab4de7c3466f7b6a378c4d8895754b6")
+        version("2.1.0", sha256="72593c037505dd76e8b5989358b7580a3fdb213051a406adb26a487d26c68c60")
+        version("2.0.1", sha256="19ebcd88d829712468cfef10c634c3e975acdf78dd1a57671d11657574636053")
+        version("2.0.0", sha256="b5387e9d37b214328f422961e2249f2687453c2702b2633b7d6a678e544b9a02")
 
     sanity_check_is_dir = ["bin", join_path("lib", "chapel"), join_path("share", "chapel")]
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/chapel/package.py
```diff
@@ -479,7 +484,7 @@ class Chapel(AutotoolsPackage, CudaPackage, ROCmPackage):
         "https://github.com/chapel-lang/chapel/issues/27273",
     )
 
-    conflicts("+rocm", when="@:2.1", msg="ROCm support in spack requires Chapel 2.0.0 or later")
+    conflicts("+rocm", when="@:1", msg="ROCm support in spack requires Chapel 2.0.0 or later")
```
bonachea: Looking at the changes made in 2bf4ab9585c8d483cc8581d65912703d3f020393, I see other changes that also seem wrong.

In particular, that commit removed all reference to ROCm versions `4:5.4` (presumably because those ROCm versions are now deprecated). However ROCm deprecation does not change the fact that Chapel versions `2:2.1` still exist in the list of supported Chapel versions, and with `+rocm` those older Chapel versions have a _hard_ dependency on ROCm `4:5.4`. By removing the lines expressing those restrictions, one can now attempt to build (say) `chapel@2.1 +rocm` without any spack conflict, which due to inheritance from `ROCmPackage` will default to pulling in the latest `hip`. That will of course fail at Chapel build time, but IMO we should not allow it to even concretize.

I didn't notice those changes when that ROCm PR went by, but those changes seem Just Plain Wrong for the Chapel spackage and we should probably fix this broken state. 

One minimalist approach would be to simply disable ROCm support entirely for chapel versions earlier that 2.2. For example:


```suggestion
    conflicts("+rocm", when="@:2.1", msg="ROCm support in spack requires Chapel 2.2.0 or later")
```

bonachea: Another change made in 2bf4ab9585c8d483cc8581d65912703d3f020393 that looks incorrect is the removal of this line:

```
depends_on("llvm-amdgpu@4:5.4", when="+rocm llvm=spack")
```

There was definitely rationale for originally adding this line, which probably remains true regardless of the deprecation of those `llvm-amdgpu` package versions in spack. In particular, there is now logic below (around line 644) that references the (now missing)  `llvm-amdgpu` dependence that will presumably cause Spack to fatal error if it's ever reached.

[Chapel docs](https://chapel-lang.org/docs/main/technotes/gpu.html#readme-gpu) state:
```
Specifically for targeting AMD GPUs:

ROCm version between 5.0 and 5.4 or between ROCm 6.0 and 6.2 must be installed.

For ROCm 5.x, CHPL_LLVM must be set to system. Note that, ROCm installations come with LLVM. Setting CHPL_LLVM=system will allow you to use that LLVM. Note that ROCm 5.x is not actively tested and we recommend using ROCm 6.x.

For ROCm 6.x, only CHPL_LLVM=bundled is supported.
```

It seems like the deprecation of ROCm packages in spack are forcing the Chapel spackage into a position of *only* supporting ROCm 6.x, despite the fact that very recent Chapel versions have lagged behind in that goal. If that's the case then IMO we really need to encode that final requirement as something like:

```
conflicts("llvm=spack", when="+rocm", msg="Chapel ROCm support through Spack requires llvm=bundled")
conflicts("llvm=none",  when="+rocm", msg="Chapel ROCm support through Spack requires llvm=bundled")
```

bonachea (changes requested):
@arifthpe Thanks for the draft PR with the new forthcoming version!

I made a few observations about other defects that appear to have recently been introduced into the ROCm support. These are only tangentially relevant to the new release, but likely render the `+rocm` variant largely unusable. Even after repairing these fairly obvious defects I discovered through source inspection, there could easily be additional, more subtle issues "lurking" in the ROCm support of this package.

I'm not sure what priority the Chapel effort currently places on ROCm support. However if this matters to anyone then after fixing the current defects we should ideally perform some manual testing with `+rocm` to establish confidence it works at all.

repos/spack_repo/builtin/packages/chapel/package.py
arifthpe: > One minimalist approach would be to simply disable ROCm support entirely for chapel versions earlier that 2.2.

That seems fine to me. Is there even another option, short of just removing those Chapel versions?

bonachea: @arifthpe said:
> Is there even another option, short of just removing those Chapel versions?

I'm not sure what you're suggesting here.

In any case, we should definitely consider marking some of the older Chapel versions in the spackage as `deprecated` (see [Spack docs](https://spack.readthedocs.io/en/latest/packaging_guide_creation.html#deprecating-old-versions)), so they can eventually be removed entirely after about 6 months. Currently the oldest version in the spackage is Chapel 2.0.0 which is now ~18 months and 7 releases behind the most recent.

Is there any formal Chapel support/life-cycle policy for ancient releases? 
For example, in `gasnet` we only support GASNet releases from the past 2 years (although that spackage is also much simpler due to fewer dependencies).

In practice, I'd say we should consider deprecating any outdated versions that nobody is willing test whenever the spackage logic changes.

arifthpe: > It seems like the deprecation of ROCm packages in spack are forcing the Chapel spackage into a position of only supporting ROCm 6.x, despite the fact that very recent Chapel versions have lagged behind in that goal. If that's the case then IMO we really need to encode that final requirement

I agree with that assessment, and will go ahead with requiring bundled LLVM for rocm support.

repos/spack_repo/builtin/packages/chapel/package.py
```diff
@@ -479,12 +484,14 @@ class Chapel(AutotoolsPackage, CudaPackage, ROCmPackage):
         "https://github.com/chapel-lang/chapel/issues/27273",
     )
 
-    conflicts("+rocm", when="@:2.1", msg="ROCm support in spack requires Chapel 2.0.0 or later")
+    conflicts("+rocm", when="@:2.1", msg="ROCm support in spack requires Chapel 2.2.0 or later")
     # Chapel restricts the allowable ROCm versions
     with when("@2.2: +rocm"):
         depends_on("hsa-rocr-dev@6.0:6.2")
         depends_on("hip@6.0:6.2")
     requires("llvm=bundled", when="+rocm ^hip@6.0:6.2", msg="ROCm 6 support requires llvm=bundled")
```
bonachea: I believe the `requires` line above can now be removed, as it is subsumed by the more general `conflicts` lines below.

Alternatively we could keep this `requires` line and remove the version 6 dependence, then the `conflicts` lines would be redundant.

Sorry I did not notice these partially overlapping constraints when writing my prior comment.

All of this assumes that (as documented) Chapel still cannot work with external, recent versions of llvm-amdgpu. I recall that for awhile there was an upstream bug preventing this from working (and thus requiring the bundled/patched LLVM), but I don't know the current status of that.

repos/spack_repo/builtin/packages/chapel/package.py
arifthpe: > I'm not sure what you're suggesting here.

I meant dropping support for them. I forgot about the requirement to deprecate first.

 > In any case, we should definitely consider marking some of the older Chapel versions in the spackage as deprecated (see [Spack docs](https://spack.readthedocs.io/en/latest/packaging_guide_creation.html#deprecating-old-versions)), so they can eventually be removed entirely after about 6 months. Currently the oldest version in the spackage is Chapel 2.0.0 which is now ~18 months and 7 releases behind the most recent.
> Is there any formal Chapel support/life-cycle policy for ancient releases?
> For example, in gasnet we only support GASNet releases from the past 2 years (although that spackage is also much simpler due to fewer dependencies).
> In practice, I'd say we should consider deprecating any outdated versions that nobody is willing test whenever the spackage logic changes.

We don't have a general lifecycle policy that I know of. My understanding is we essentially don't support old releases, except for occasionally making point releases on the latest version if a particularly bad flaw is found.

For Spack, I would like us to deprecate all versions older than the current release, and have made issue https://github.com/chapel-lang/chapel/issues/27818 for that. Will gather input from my team and then act on it, but probably in a separate PR after this release so there's a chance for discussion.

repos/spack_repo/builtin/packages/chapel/package.py
arifthpe: > I believe the requires line above can now be removed, as it is subsumed by the more general conflicts lines below.
> Alternatively we could keep this requires line and remove the version 6 dependence, then the conflicts lines would be redundant.

I think I'll go with the latter, seems like a more explicit way to encode the requirement.

> All of this assumes that (as documented) Chapel still cannot work with external, recent versions of llvm-amdgpu. I recall that for awhile there was an upstream bug preventing this from working (and thus requiring the bundled/patched LLVM), but I don't know the current status of that.

I checked with our GPU people and this is still the case.

arifthpe: > Alternatively we could keep this requires line and remove the version 6 dependence, then the conflicts lines would be redundant.

Actually, why not keep the version 6 dependence (assuming you mean the `^hip@6.0:6.2` in this `requires`)? The limitation for Chapel is specifically that using ROCm _6_ with Chapel requires using the bundled LLVM, but that wouldn't inherently be the case for future versions. I get it boils down to the same logical conditions at the moment, I just prefer that phrasing.

bonachea: Well now that we've removed spackage support for ROCm 5, the lines above ensure the only available ROCm support is for `hip@6.0:6.2`. So as you observed the condition is redundant, and it means the same thing whether we leave `^hip@6.0:6.2` in the `requires` or not. 

Wrt the message: `ROCm 6 support requires llvm=bundled` is a _little_ misleading, because it's currently _all_ ROCm support in this package (regardless of ROCm version) that requires `llvm=bundled`, but that's at best a minor aesthetic quibble.

Hopefully/presumably some future Chapel version will add support for a newer ROCm, and at that point we'll definitely need to update this entire group of constraints accordingly.

arifthpe: I could see someone feeling annoyed at the message specifying ROCm 6 when that's now the only supported version. I'll change it back and add a comment to update this spot if future Chapel changes things

arifthpe:
@bonachea should be ready now

bonachea (approved):
LGTM!

Sha256 hash manually verified.

I've tested basic operation of `chapel@2.6` in this branch on four different platforms (NERSC Perlmutter, OLCF Frontier and two private clusters), using several C/C++ toolchains and various relevant combinations of variants (including `+cuda` and `+rocm`).

tldahlgren (approved):
(Also) confirmed the version sha256(s).

tldahlgren:
> LGTM!
> 
> Sha256 hash manually verified.

FWIW. There's a check for that now called `ci / prechecks / verify-checksums`, which will check sha256s as well as commit[+tag] values.  See https://github.com/spack/spack-packages/actions/runs/18043507389/job/51348455423?pr=1483.

bradcray:
Thanks, @tldahlgren !

PR #1485, opened 2025-09-15
intel-oneapi-runtime: remove when constraint for providing fortran-rt and libifcore

climbfuji:
## Description

In spack-packages, the known `intel-oneapi-compilers` are 2021.1.2 and newer. Therefore, the constraint
```
provides("fortran-rt", "libifcore@5", when="%oneapi@2021:")
```
in the `intel-oneapi-runtime` package can be simplified to
```
provides("fortran-rt", "libifcore@5")
```

Notes:
1. This change does not affect the classic Intel compilers (`intel-oneapi-compilers-classic`), since these do not impose any dependency on the oneAPI runtime library. For the Intel LLVM compilers (`intel-oneapi-compilers`), this dependency is added automatically via the `runtime_constraints` option in the `intel-oneapi-compilers` package. For `intel-oneapi-compilers-classic`, this is not the case. Also, the `intel-oneapi-runtime` library explicitly depends on the LLVM compilers `intel-oneapi-compilers`.
2. The first release of Intel oneAPI **by Intel** (package intel-oneapi-compilers) ever was 2021.2, hence this PR is just removing unnecessary code.


haampie:
There can still be older, external versions of oneapi compilers on user systems. I don't get the obsession with trying to make external `*-runtime` packages a thing. We do want to express constraints like these, and likely even more on an LLVM runtime package when we get to it. I would not be inclined to merge this PR.

climbfuji:
> There can still be older, external versions of oneapi compilers on user systems. I don't get the obsession with trying to make external `*-runtime` packages a thing. We do want to express constraints like these, and likely even more on an LLVM runtime package when we get to it. I would not be inclined to merge this PR.

1. The first ever released Intel oneAPI compiler was 2021.1, which was released in December 2020 and replaced the previously released Intel Parallel Studio XE. There has never been an intel-oneapi-compilers@2020 or earlier.
2. Please do not make this personal. I have very valid reasons that I laid out in https://github.com/spack/spack/issues/51268#issuecomment-3258146470, especially bullet 1, why this is needed. If you don't have to worry about FISMA high systems, then you can call yourself lucky.
3. I am not asking for anyone else and in particular you to do the same, just to restore the option to go back to the spack v0.23 behavior where external runtimes were possible for both GCC and oneAPI (now it's only possible for GCC, but not oneAPI).

I spent many hours debugging the concretizer and I showed in https://github.com/spack/spack/issues/51268#issuecomment-3283406636 and https://github.com/spack/spack/issues/51268#issuecomment-3285191379 how to reproduce and avoid the bug.   Given what you said, if you want to express constraints like these [in the context of compilers and runtimes], then hopefully my debugging efforts are useful for the spack developers. Who knows how this bug may materialize in the future.

haampie:
As for https://github.com/spack/spack/issues/51268#issuecomment-3258146470, I didn't consider "compilers are installed by the system administrators only" problematic, since Spack runs basically `cp /usr/lib/libx.so <spack prefix>/lib/libx.so`, that's it. We can discuss further in that topic.

> Please do not make this personal
> I spent many hours debugging the concretizer and I showed in spack/spack#51268 (comment) and spack/spack#51268 (comment) how to reproduce and avoid the bug

I did not mean to make it personal. It's just that these runtime packages were not meant to be externals, and the fact that it happened to work in Spack v0 does not mean it should work in Spack v1 -- it's not a bug, and it's not a regression.

climbfuji:
Sorry, I just saw your comment in https://github.com/spack/spack-packages/pull/1485#issuecomment-3302436767. Yep, all good.

The reasoning behind allowing external runtime libraries has to do with security concerns. The argument you made for being able to hand a build off and allow it to continue to work makes sense for many users. But on high security platforms, the admins want to have as much control as possible. Since the compiler runtime libraries are the ones that are linked to literally all libraries and executables, any security issues discovered in those have the largest consequences. Admins want to be able to pull these vulnerabilities from the systems if needed in an emergency. If we copy the runtime libraries, we proliferate them on the system and take away that tool from the admins (noted: this works only for shared linking of runtime libraries, but that is far more common nowadays).

These are rare occurrences, but I would like to be able to address these concerns by having the option of an external runtime library. By no means am I implying that these should be external by default. If we give system administrators on these high security platforms time and options to work with spack inside their rules and comfort zone, then we will make the adoption of spack easier and faster, or even possible.

But if that's not an option, then that's ok. We may be able to work around this in our own fork of spack and spack-packages to facilitate the transition on these FISMA HIGH platforms. Or we can find a way to label external runtime libraries as a not recommended or deprecated option that will be removed in spack 1.x/2?

Regardless, I think that:
1. This PR should still be merged because it removes a constraint on a version range that always evaluates to true.
2. The underlying bug in the concretization logic should still be addressed. I still haven't understood fully what is going on, but it looks like a circular dependency that the concretizer can't resolve (it needs a concrete compiler spec to satisfy the runtime library, but it can't concretize the compiler spec yet because of the runtime constrains logic in the compiler package?). FWIW, I ran into the same error with the GNU compilers on a Cray machine (this time: `1. Only external, or concrete, compilers are allowed for the c, c++, fortran language`). The GNU compiler does not have the `when=` condition that the Intel oneAPI compiler has. I am happy to jump on a call if that helps clarifying the situation.

climbfuji:
Pinging the reviewers to take another look at this pull request, please. Thanks very much in advance.

climbfuji:
Pinging all reviewers again.

I recognize that I've lost the battle for non-buildable compiler runtime packages, that doesn't work anymore with the spack develop concretizer, regardless of the changes here. But that's ok, I choose to die on a different hill today.

Nevertheless, this PR still makes sense, because the version constraint specified here has no effect (the first-ever oneapi release was `2021.1` in December 2020).

climbfuji:
@haampie I still think this change makes sense by itself, but I no longer need it. Do you want me to close the PR? Thanks.

climbfuji:
No longer needed

PR #1486, opened 2025-09-15
Add cmake dependency for Python 3.12+ support

wspear:
TAU needs to build its internal perfstubs library to support python instrumentation for python@3.12:. This requires cmake.


eugeneswalker (approved):
Thank you!

PR #1487, opened 2025-09-15
libfabric: Add ROCm support

nhanford:
Adds ROCm support to libfabric.

### DIFF
--- a/repos/spack_repo/builtin/packages/libfabric/package.py
+++ b/repos/spack_repo/builtin/packages/libfabric/package.py
@@ -212,7 +212,7 @@ class Libfabric(AutotoolsPackage, CudaPackage, ROCmPackage):
             *self.with_or_without("uring"),
             *self.with_or_without("cuda", activation_value="prefix"),
             *self.with_or_without("ze", variant="level_zero"),
-            *self.with_or_without("gdrcopy"),
+            *self.with_or_without("gdrcopy", activation_value="prefix"),
             *self.with_or_without(
                 "rocr", variant="rocm", activation_value=lambda _: self.spec["hip"].prefix
                 ),
--- a/repos/spack_repo/builtin/packages/libfabric/package.py
+++ b/repos/spack_repo/builtin/packages/libfabric/package.py
@@ -212,6 +212,10 @@ class Libfabric(AutotoolsPackage, CudaPackage, ROCmPackage):
             *self.with_or_without("uring"),
             *self.with_or_without("cuda", activation_value="prefix"),
             *self.with_or_without("ze", variant="level_zero"),
+            *self.with_or_without("gdrcopy"),
+            *self.with_or_without(
+                "rocr", variant="rocm", activation_value=lambda _: self.spec["hip"].prefix
+                ),
         ]
 
         if self.spec.satisfies("+kdreg"):
@@ -232,12 +236,6 @@ class Libfabric(AutotoolsPackage, CudaPackage, ROCmPackage):
 
         if self.spec.satisfies("fabrics=xpmem"):
             args.append(f"--enable-xpmem={self.spec['xpmem'].prefix}")
-
-        if self.spec.satisfies("+gdrcopy"):
-            args.append(f"--with-gdrcopy={self.spec['gdrcopy'].prefix}")
-
-        if self.spec.satisfies("+rocm"):
-            args.append(f"--with-rocr={self.spec['hip'].prefix}")
 
         return args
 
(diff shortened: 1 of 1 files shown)


nhanford:
@spackbot fix style

nhanford:
@spackbot fix style

nhanford:
@rajachan @msimberg Please let me know if anything here looks awry. I tried to copy the same style as the CUDA support.

msimberg:
Thanks @nhanford for pinging me. I think this looks good in principle but I'd like to try a build of this in the next few days before approving. I have a few questions on the implementation though:

The main question I have is regarding `--enable-rocr-dlopen`: is that necessary? Especially if the build is set up with `--with-rocr` in the first place? In general I prefer to avoid any dynamic loading in spack environments as it can lead to unpredictable versions of libraries being loaded. There is for cuda also a `--enable-cuda-dlopen` flag in libfabric but it's not set in the spack package (I'm not sure if this was a conscious choice or not when cuda support was added).

Second question is if you could use `self.with_or_without` like for cuda and ze, specifying the `activation_value` and `variant` params? I'm not an expert on the `Autotools` build system and haven't used those options before, but from a quick look it seems possibly useful (to make the support more uniform with cuda and ze).

nhanford:
@msimberg Thanks for the reply. Yes I agree about `dlopen` and have removed it. I had been blindly following our system LLNL builds. @alalazo For the second point, am I correct that `with_or_without` can only be used when the name of the configure argument matches the variant name? In this case, the variant name `+rocm` is inherited from `ROCmPackage` and the configure argument name is `rocr`, so my understanding is that we unfortunately need the logic, but I'd love to be proven wrong.

msimberg:
@nhanford from a bit of testing it seems like
```
*self.with_or_without("rocr", variant="rocm", activation_value=lambda _: self.spec["hip"].prefix),
```
does the right thing. It's arguable whether that's easier to read than your current version so I'll leave it up to you to decide whether you apply that or not (please test it yourself as well, it seems to work in my environment though).

Thanks for removing the dlopen option as well.

And finally, sorry for the conflict from my PR (https://github.com/spack/spack-packages/pull/1641)! While commenting here, I realize that I could probably do the same `with_or_without` for the gdrcopy variant as well.

nhanford:
@spackbot fix style

nhanford:
@msimberg No worries! Sorry for the noise--I saw an opportunity to consolidate a bunch of these things. I have successfully tested everything except gdrcopy at the moment. I'll see if I can test that too.

repos/spack_repo/builtin/packages/libfabric/package.py
```diff
@@ -211,30 +212,32 @@ def configure_args(self):
             *self.with_or_without("uring"),
             *self.with_or_without("cuda", activation_value="prefix"),
             *self.with_or_without("ze", variant="level_zero"),
+            *self.with_or_without("kdreg2", variant="kdreg"),
```
msimberg: I'm not quite sure why libfabric is doing what it's doing, but I guess with the previous version `--with-kdreg2` was emitted only when `+kdreg2` and otherwise no flag was set. Now it's emitting `--without-kdreg2` and for some reason in CI it's now looking for kdreg2 headers?

Unless you feel like further debugging might be safest to just revert the kdreg2 change but keep the rest?

msimberg: Weirdly enough on my previous PR (https://github.com/spack/spack-packages/pull/1641) there's no kdreg2 flag, kdreg2 headers are not found, but the build doesn't fail. On this PR there's `--without-kdreg2`, headers are not found, and the build fails.

It seems like kdreg2 is enabled by default (https://github.com/ofiwg/libfabric/blob/7d60749967a425fff75dd36df03757ab871b0b4e/configure.ac#L617), and oddly the flag to disable it is `--disable-kdreg2` (https://github.com/ofiwg/libfabric/blob/7d60749967a425fff75dd36df03757ab871b0b4e/configure.ac#L622).

Internally the `AutotoolsPackage` seems to have support setting the "on" and "off" keywords, but I think that's not exposed in the API (only `with_or_without` and `enable_or_disable`).

All that to say: I don't really understand what's going on, but conclusion is probably the same as in my previous comment (i.e. revert the kdreg2 part).

What do you think?

nhanford: Thanks for tracking down the root cause on this! I agree that's the best approach and I have reverted it accordingly.

msimberg (approved):

PR #1492, opened 2025-09-15
libfabric: kdreg variant

nhanford:
Spack's configure output shows that the `kdreg` variant is not influential:

```
configure: WARNING: unrecognized options: --with-kdreg
```

I can't find a history of a `--with-kdreg` variant in the libfabric repo. AFAICT it has always
been `--with-kdreg2`. I have changed the variant name and therfore the `with_or_without` so that it is
influential.

This PR includes the fixes from #1478. Therefore, I can close that one and we can merge this, or I can rebase this one on develop if we want to do them separately.

### DIFF
--- a/repos/spack_repo/builtin/packages/libfabric/package.py
+++ b/repos/spack_repo/builtin/packages/libfabric/package.py
@@ -101,11 +101,11 @@ class Libfabric(AutotoolsPackage, CudaPackage):
         multi=True,
     )
 
-    # NOTE: the 'kdreg2' variant enables use of the special /dev/kdreg2 file to
+    # NOTE: the 'kdreg' variant enables use of the special /dev/kdreg2 file to
     #   assist in memory registration caching in the GNI provider.  This
     #   device file can only be opened once per process, however, and thus it
     #   frequently conflicts with MPI.
-    variant("kdreg2", default=False, description="Enable kdreg2 on supported Cray platforms")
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/libfabric/package.py
```diff
@@ -101,11 +101,11 @@ class Libfabric(AutotoolsPackage, CudaPackage):
         multi=True,
     )
 
-    # NOTE: the 'kdreg' variant enables use of the special /dev/kdreg file to
+    # NOTE: the 'kdreg2' variant enables use of the special /dev/kdreg2 file to
     #   assist in memory registration caching in the GNI provider.  This
     #   device file can only be opened once per process, however, and thus it
     #   frequently conflicts with MPI.
-    variant("kdreg", default=False, description="Enable kdreg on supported Cray platforms")
+    variant("kdreg2", default=False, description="Enable kdreg2 on supported Cray platforms")
```
alalazo: Can we keep the variant name, to avoid unnecessary backward incompatibilities?

alalazo (changes requested):
There are legitimate failures in pipelines

nhanford:
@alalazo I went ahead and added the logic to keep the backwards compatibility. Please let me know what you think. Thanks!

alalazo (approved):

PR #1500, opened 2025-09-16
py-biosppy: add new package + new dep packages

manuelakuhn:
https://github.com/scientisst/BioSPPy/tree/v2.2.3

needs new packages as dependencies:
https://github.com/jab/bidict/tree/v0.23.1
https://github.com/opencv/opencv-python/tree/88
https://pypi.org/project/PeakUtils/#files (no tags on bitbucket)

for me it only concretizes and builds fine with a newer `py-pywavelets` version (see #1488)

### DIFF
--- a/repos/spack_repo/builtin/packages/py_bidict/package.py
+++ b/repos/spack_repo/builtin/packages/py_bidict/package.py
@@ -1,0 +1,21 @@
+# Copyright Spack Project Developers. See COPYRIGHT file for details.
+#
+# SPDX-License-Identifier: (Apache-2.0 OR MIT)
+
+from spack_repo.builtin.build_systems.python import PythonPackage
+
+from spack.package import *
+
+
+class PyBidict(PythonPackage):
+    """The bidirectional mapping library for Python."""
+
+    homepage = "https://github.com/jab/bidict"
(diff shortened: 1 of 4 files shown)


repos/spack_repo/builtin/packages/py_bidict/package.py
```diff
@@ -0,0 +1,21 @@
...
+from spack.package import *
+
+
+class PyBidict(PythonPackage):
+    """The bidirectional mapping library for Python."""
+
+    homepage = "https://github.com/jab/bidict"
+    pypi = "bidict/bidict-0.23.1.tar.gz"
+
+    license("MPL-2.0", checked_by="github_user1")
```
alecbcs:
```suggestion
    license("MPL-2.0")
```

Accidental placeholder left in package. Suggest to remove or add yourself.

alecbcs (changes requested):
Quick suggestion, otherwise looks good to me.

repos/spack_repo/builtin/packages/py_bidict/package.py
manuelakuhn: Sorry that was indeed an overlook on. Thank you very much.

alecbcs (approved):
No worries. Looks good to me. Thanks!

PR #1509, opened 2025-09-16
py-patool: add v4.0.1 + new dep package

manuelakuhn:
https://github.com/wummel/patool/tree/4.0.1

`py-patool` needs the new `py-setuptool-reproducible`, so this was added as well
https://github.com/wimglenn/setuptools-reproducible/tree/v0.1


alecbcs (approved):

PR #1511, opened 2025-09-16
gcovr: add v8.2, v8.3

tvoskui:
Add the latest releases to the gcovr package.

@psakievich

### DIFF
--- a/repos/spack_repo/builtin/packages/py_gcovr/package.py
+++ b/repos/spack_repo/builtin/packages/py_gcovr/package.py
@@ -14,11 +14,18 @@ class PyGcovr(PythonPackage):
     Python."""
 
     homepage = "https://gcovr.com/"
-    pypi = "gcovr/gcovr-4.2.tar.gz"
+    pypi = "gcovr/gcovr-8.3.tar.gz"
 
+    version("develop", branch="main")
+    version("8.3", sha256="faa371f9c4a7f78c9800da655107d4f99f04b718d1c0d9f48cafdcbef0049079")
+    version("8.2", sha256="9a1dddd4585d13ec77555db5d6b6a31ee81587ea6fc604ff9fcd232cb0782df5")
     version("7.2", sha256="e3e95cb56ca88dbbe741cb5d69aa2be494eb2fc2a09ee4f651644a670ee5aeb3")
     version("5.2", sha256="217195085ec94346291a87b7b1e6d9cfdeeee562b3e0f9a32b25c9530b3bce8f")
     version("4.2", sha256="5aae34dc81e51600cfecbbbce3c3a80ce3f7548bc0aa1faa4b74ecd18f6fca3f")
+
+    depends_on("py-hatchling", when="@8.3:")
+    depends_on("py-hatch-vcs", when="@8.3:")
+    depends_on("py-hatch-fancy-pypi-readme", when="@8.3:")
 
     depends_on("python@3.8:", when="@7.2:", type=("build", "run"))
     depends_on("python@3.7:", when="@5.1:", type=("build", "run"))
--- a/repos/spack_repo/builtin/packages/py_gcovr/package.py
+++ b/repos/spack_repo/builtin/packages/py_gcovr/package.py
@@ -27,10 +27,9 @@ class PyGcovr(PythonPackage):
     depends_on("python@3.6:", when="@5.0", type=("build", "run"))
     depends_on("python@2.7:2,3.5:", when="@:4", type=("build", "run"))
     depends_on("py-setuptools", type=("build", "run"), when="@:8.2")
-
-    depends_on("py-hatchling", when="@8.3:")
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/py_gcovr/package.py
```diff
@@ -14,12 +14,18 @@ class PyGcovr(PythonPackage):
     Python."""
 
     homepage = "https://gcovr.com/"
-    pypi = "gcovr/gcovr-4.2.tar.gz"
+    pypi = "gcovr/gcovr-8.3.tar.gz"
```
alecbcs: Spack performs version inference in the URLs already (by detecting and replacing the version in a URL automatically) so this is isn't necessary, but doesn't hurt to bump it. Just a note for future PRs.

repos/spack_repo/builtin/packages/py_gcovr/package.py
```diff
@@ -14,12 +14,18 @@ class PyGcovr(PythonPackage):
...
+    version("8.3", sha256="faa371f9c4a7f78c9800da655107d4f99f04b718d1c0d9f48cafdcbef0049079")
+    version("8.2", sha256="9a1dddd4585d13ec77555db5d6b6a31ee81587ea6fc604ff9fcd232cb0782df5")
     version("7.2", sha256="e3e95cb56ca88dbbe741cb5d69aa2be494eb2fc2a09ee4f651644a670ee5aeb3")
     version("5.2", sha256="217195085ec94346291a87b7b1e6d9cfdeeee562b3e0f9a32b25c9530b3bce8f")
     version("4.2", sha256="5aae34dc81e51600cfecbbbce3c3a80ce3f7548bc0aa1faa4b74ecd18f6fca3f")
 
+    depends_on("py-hatchling", when="@8.3:")
+    depends_on("py-hatch-vcs", when="@8.3:")
+    depends_on("py-hatch-fancy-pypi-readme", when="@8.3:")
+
```
alecbcs: These should likely be placed after `depends_on("py-setuptools")` in keeping with the existing format.

alecbcs (commented):
Two very small notes, otherwise this looks good to me :)

Thanks @tvoskui and welcome to the Spack community!

repos/spack_repo/builtin/packages/py_gcovr/package.py
tvoskui: Ok, moved them down after py-setuptools

repos/spack_repo/builtin/packages/py_gcovr/package.py
tvoskui: Noted. Changed it back to 4.2 here.

psakievich (dismissed):
Looks good to me. Thanks @tvoskui

repos/spack_repo/builtin/packages/py_gcovr/package.py
```diff
@@ -26,6 +28,10 @@ class PyGcovr(PythonPackage):
     depends_on("python@2.7:2,3.5:", when="@:4", type=("build", "run"))
     depends_on("py-setuptools", type=("build", "run"))
 
+    depends_on("py-hatchling", when="@8.3:")
+    depends_on("py-hatch-vcs", when="@8.3:")
+    depends_on("py-hatch-fancy-pypi-readme", when="@8.3:")
+
```
alecbcs:
```suggestion
    depends_on("py-hatchling", type="build", when="@8.3:")
    depends_on("py-hatch-vcs", type="build", when="@8.3:")
    depends_on("py-hatch-fancy-pypi-readme", when="@8.3:", type="build")

```

Let's remove the empty line to join these with existing build dependencies. Also by default dependencies end up as both build and link types and here we only want hatching to be a build dependency.

repos/spack_repo/builtin/packages/py_gcovr/package.py
```diff
@@ -26,6 +28,10 @@ class PyGcovr(PythonPackage):
     depends_on("python@2.7:2,3.5:", when="@:4", type=("build", "run"))
     depends_on("py-setuptools", type=("build", "run"))
```
alecbcs:
```suggestion
    depends_on("py-setuptools", type=("build", "run"), when="@:8.2")
```

We'll also want to constrain the past dependency on setuptools so that we don't keep pulling it in for newer builds.

alecbcs (changes requested):
Two additional notes of feedback. One those are fixed, I think this looks good to me.

repos/spack_repo/builtin/packages/py_gcovr/package.py
tvoskui: Done

alecbcs (approved):
Looks good to me. Thanks @tvoskui!

psakievich (approved):

PR #1513, opened 2025-09-16
libcerf: Add missing build dependency

michaelkuhn:


alecbcs:
`pod2html` is built into the perl5 standard library. We should perhaps add perl as a build dependency.

michaelkuhn:
Thanks for the hint! I've pushed an update to add perl as a build dependency.

alalazo (approved):

Issue #1451, opened 2025-09-12
Installation issue: paving the way to Octave packages build with Octave 10.2.0


cessenat:
### Steps to reproduce the issue

```console
$ spack spec -I octave-io ^octave@10.2.0
...
 -   octave-io@2.7.0 build_system=octave platform=linux os=ubuntu20.04 target=zen2 %cxx=gcc@9.4.0
[+]      ^compiler-wrapper@1.0 build_system=generic platform=linux os=ubuntu20.04 target=zen2 
[e]      ^gcc@9.4.0~binutils+bootstrap~graphite~nvptx~piclibs~profiled~strip build_system=autotools build_type=RelWithDebInfo languages:='c,c++,fortran' patches:=d4919d6 platform=linux os=ubuntu20.04 target=zen2 
[+]      ^gcc-runtime@9.4.0 build_system=generic platform=linux os=ubuntu20.04 target=zen2 
[e]      ^glibc@2.31 build_system=autotools platform=linux os=ubuntu20.04 target=zen2 
[+]      ^octave@10.2.0~arpack+bz2+curl+fftw+fltk+fontconfig+freetype+gl2ps+glpk+gnuplot+hdf5~jdk~llvm+magick~opengl+pcre2+qhull+qrupdate+qscintilla+qt+readline~suitesparse+zlib build_system=autotools platform=linux os=ubuntu20.04 target=zen2 %c,cxx,fortran=gcc@9.4.0
[+]          ^bzip2@1.0.8~debug~pic+shared build_system=generic platform=linux os=ubuntu20.04 target=zen2 %c=gcc@9.4.0
[+]              ^diffutils@3.12 build_system=autotools platform=linux os=ubuntu20.04 target=zen2 %c=gcc@9.4.0
[+]          ^curl@8.15.0~gssapi~ldap~libidn2~librtmp~libssh~libssh2+nghttp2 build_system=autotools libs:=shared,static tls:=openssl platform=linux os=ubuntu20.04 target=zen2 %c,cxx=gcc@9.4.0
[+]              ^nghttp2@1.65.0 build_system=autotools platform=linux os=ubuntu20.04 target=zen2 %c,cxx=gcc@9.4.0
[+]              ^openssl@3.4.1~docs+shared build_system=generic certs=mozilla platform=linux os=ubuntu20.04 target=zen2 %c,cxx=gcc@9.4.0
[...]

PR #1514, opened 2025-09-16
octave new version 10.2.0 demands stdc++17

cessenat:
There are many changes:
 - removed maintainer @mtmiller that unfortunately is no longer active,
 - added maintainer @mmuetzel and myself
 - added version 10.2.0 that demands stdc++17 with GNU extensions,
 - OCTAVE_CONF_MKOCTFILE_CXX is set to `quote(self.compiler.cxx + " -std=gnu++17")` when 10.2.0
 - run environment sets CXX to `self.compiler.cxx + " -std=gnu++17"` when 10.2.0

I would like to have advise from many skilled spack packagers, such as @alalazo, @bernhardkaindl, @michaelkuhn, @alecbcs, @davydden since the way I made Octave 10.2.0 compile properly, with extensions using spack packages or using Octave "pkg install -forge extension" where extension may be "io", "gsl", "aduino", ..., well, this way may be naive or irrelevant.
I have been greatly helped by Markus Mützel @mmuetzel from Savannah who indicated that within Octave package installation for 10.2.0 may require, on some platforms, the setting:
```
setenv('CXX', '/usr/bin/clang++ -std=gnu++17');
pkg install -forge io

```
This is why I added a `setup_run_environment` function to set CXX value.
I have tested these modifications with gcc 10.5.0 and octave 10.2.0 or 9.4.0 on Ubuntu 20.04 with no external dep apart from compiler on
$ spack debug report
* **Spack:** 1.1.0.dev0 (https://github.com/spack/spack/commit/4a3df09f488e4475692b7512df64318e068fe4ab)
* **Builtin repo:** https://github.com/spack/spack-packages/commit/07dbc55cea8ae969762a8824fbe62b77d0adb6d3
* **Python:** 3.8.10
* **Platform:** linux-ubuntu20.04-zen2

This PR is related to discussion/issue
https://github.com/spack/spack-packages/issues/1451

Octave is such a thing, and so is spack, it is really worth.

### DIFF
--- a/repos/spack_repo/builtin/packages/octave/package.py
+++ b/repos/spack_repo/builtin/packages/octave/package.py
@@ -25,7 +25,7 @@ class Octave(AutotoolsPackage, GNUMirrorPackage):
 
     homepage = "https://www.gnu.org/software/octave/"
     gnu_mirror_path = "octave/octave-4.0.0.tar.gz"
-    maintainers("cessenat", "mmuetzel")
+    maintainers("cessenat")
 
     extendable = True
 
--- a/repos/spack_repo/builtin/packages/octave/package.py
+++ b/repos/spack_repo/builtin/packages/octave/package.py
@@ -87,7 +87,6 @@ class Octave(AutotoolsPackage, GNUMirrorPackage):
     # Required dependencies
     depends_on("c", type="build")
     depends_on("cxx", type="build")
-    depends_on("cxx@17:", when="@10:")
     depends_on("fortran", type="build")
 
     depends_on("blas")
--- a/repos/spack_repo/builtin/packages/octave/package.py
+++ b/repos/spack_repo/builtin/packages/octave/package.py
@@ -400,4 +400,3 @@ class Octave(AutotoolsPackage, GNUMirrorPackage):
             env.set("CXX", self.compiler.cxx + " -std=gnu++17")
         else:
             env.set("CXX", self.compiler.cxx)
-
--- a/repos/spack_repo/builtin/packages/octave/package.py
+++ b/repos/spack_repo/builtin/packages/octave/package.py
@@ -25,12 +25,13 @@ class Octave(AutotoolsPackage, GNUMirrorPackage):
 
     homepage = "https://www.gnu.org/software/octave/"
     gnu_mirror_path = "octave/octave-4.0.0.tar.gz"
-    maintainers("mtmiller")
+    maintainers("cessenat", "mmuetzel")
 
     extendable = True
 
     license("GPL-3.0-or-later")
 
+    version("10.2.0", sha256="07fb6d9339d2f350735c91671be8e874d160018cc6b688f9efd9d558d237f69f")
     version("9.4.0", sha256="da9481205bfa717660b7d4a16732d8b2d58aadceab4993d41242a8e2848ea6c1")
     version("9.3.0", sha256="809fa39a7acc84815bf4dc4d2d7e6b228ce75a07f3b2413f3313aa8e0aaa3287")
     version("9.1.0", sha256="3f8c6c6ecfa249a47c97e18e651be4db8499be2f5de1a095a3eea53efc01d6a1")
@@ -132,9 +133,13 @@ class Octave(AutotoolsPackage, GNUMirrorPackage):
         # of the build.
         mkoctfile_in = os.path.join(self.stage.source_path, "src", "mkoctfile.in.cc")
         quote = lambda s: '"' + s + '"'
+        if self.spec.satisfies("@10.2.0:"):
+            cxx = quote(self.compiler.cxx + " -std=gnu++17")
+        else:
+            cxx = quote(self.compiler.cxx)
         entries_to_patch = {
             r"%OCTAVE_CONF_MKOCTFILE_CC%": quote(self.compiler.cc),
-            r"%OCTAVE_CONF_MKOCTFILE_CXX%": quote(self.compiler.cxx),
+            r"%OCTAVE_CONF_MKOCTFILE_CXX%": cxx,
             r"%OCTAVE_CONF_MKOCTFILE_F77%": quote(self.compiler.f77),
             r"%OCTAVE_CONF_MKOCTFILE_DL_LD%": quote(self.compiler.cxx),
             r"%OCTAVE_CONF_MKOCTFILE_LD_CXX%": quote(self.compiler.cxx),
@@ -389,3 +394,6 @@ class Octave(AutotoolsPackage, GNUMirrorPackage):
         """
         # Octave extension builds can have a global Octave executable function
         module.octave = Executable(join_path(self.spec.prefix.bin, "octave"))
+
+    def setup_run_environment(self, env):
+        env.set("CXX", self.compiler.cxx + " -std=gnu++17")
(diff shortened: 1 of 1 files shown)


alecbcs:
I believe the "correct" way to do this now that compilers are nodes that implement language dependencies is
```python3
depends_on("cxx@17:", when="@10:")
```

I'd try that as an alternative to your modifications here and let me know if it doesn't work.

cessenat:
> ```python
> depends_on("cxx@17:", when="@10:")
> ```

Good thing to explicitly set the dependency to C++ 17. Unluckily it happened to be unsufficient when I removed the other explicit `-std=gnu++17` modifications.

repos/spack_repo/builtin/packages/octave/package.py
```diff
@@ -86,6 +87,7 @@ class Octave(AutotoolsPackage, GNUMirrorPackage):
     # Required dependencies
     depends_on("c", type="build")
     depends_on("cxx", type="build")
+    depends_on("cxx@17:", when="@10:")
```
alalazo: Let's not add this, since this is bound to fail when other packages will require e.g. `cxx@14` and are mixed in the same DAG. We discussed sometimes ago that the best way to deal with this is to extend virtual packages with variants, but this is not implemented yet.

cessenat: OK, removed.

alalazo: Thanks!

mmuetzel:
I've never used spack. So, I don't think I can act as a maintainer for this package. 
Feel free to tag me though if there are questions regarding Octave itself.

repos/spack_repo/builtin/packages/octave/package.py
```diff
@@ -132,9 +133,13 @@ def patch(self):
...
         mkoctfile_in = os.path.join(self.stage.source_path, "src", "mkoctfile.in.cc")
         quote = lambda s: '"' + s + '"'
+        if self.spec.satisfies("@10.2.0:"):
+            cxx = quote(self.compiler.cxx + " -std=gnu++17")
+        else:
+            cxx = quote(self.compiler.cxx)
         entries_to_patch = {
             r"%OCTAVE_CONF_MKOCTFILE_CC%": quote(self.compiler.cc),
-            r"%OCTAVE_CONF_MKOCTFILE_CXX%": quote(self.compiler.cxx),
+            r"%OCTAVE_CONF_MKOCTFILE_CXX%": cxx,
```
mmuetzel, line 142: I'm not sure this is the best solution in general.
During `configure` Octave checks if the used compiler defaults to a new enough version of C++. If it doesn't, it adds flags that enable support for C++17 (preferably with GNU extensions).
If the compiler defaults to a newer version of C++ (e.g., C++23), it doesn't add any flags.

Whichever flags were necessary are stored with the compiler in `OCTAVE_CONF_MKOCTFILE_CXX` and used when a user builds .oct file binaries with `mkoctfile`.
These .oct file binaries should be built using the same compiler and C++ version as Octave itself.

Using the current approach, Octave could have been built using C++23 (using a new enough compiler), but .oct file sources would be built using C++17. That probably shouldn't be the default.

I still don't understand why the `mkoctfile.in.cc` file needs to be patched like that in the first place. What would happen if you didn't patch it at all?

repos/spack_repo/builtin/packages/octave/package.py
```diff
@@ -389,3 +394,9 @@ def setup_dependent_package(self, module, dependent_spec):
         """
         # Octave extension builds can have a global Octave executable function
         module.octave = Executable(join_path(self.spec.prefix.bin, "octave"))
+
+    def setup_run_environment(self, env):
+        if self.spec.satisfies("@10.2.0:"):
+            env.set("CXX", self.compiler.cxx + " -std=gnu++17")
```
mmuetzel, line 400: Not sure if this is a good default in general. See above.

cessenat: > Not sure if this is a good default in general. See above.

Well, without that pkg install -forge package fails.
Octave compilation is fine without setting -std=gnu++17, but the extended packages fails without that.

repos/spack_repo/builtin/packages/octave/package.py
cessenat: Thus is done from
https://github.com/spack/spack-packages/commit/b0ae12637e6a423cebc4c98b173916aeb2b397a7
commit by @alalazo.
Setting " -std=gnu++17" is a workaround to be able to compile packages. For Octave it is not necessary.
Also, @alalazo explained "We discussed sometimes ago that the best way to deal with this is to extend virtual packages with variants, but this is not implemented yet.", so you are right this is probably not the best solution to keep. So either we wait for a smart solution and stick down to 9.4.3, or we do that ugly stuff for the moment to allow 10.2.0. What do spack experts think about that ?

mmuetzel: I've read "compiler wrappers" being mentioned repeatedly. But what are these compiler wrappers? And why is it an issue if they are used?

In any case, this is only an issue because your Clang is so old (version 10) and doesn't default to C++17 or newer (like LLVM version 16 or newer).
Are all spack packages built with such an old version of Clang?

repos/spack_repo/builtin/packages/octave/package.py
mmuetzel: Like I wrote above, that is because of the old Clang that you are using.

repos/spack_repo/builtin/packages/octave/package.py
mmuetzel: Alternative idea (assuming these compiler wrappers cannot be used when building .oct files while running Octave for some reason):
Instead of patching `mkoctfile.in.cc`, run the `configure` script and use `sed` to replace the compiler wrappers in the generated `Makefile` (in the build tree) with the "actual" compilers.
E.g., something like this:
```
sed -i "s/MKOCTFILE_CC= ${cc_wrapper}/MKOCTFILE_CC= ${actual_cc}/g" Makefile
sed -i "s/MKOCTFILE_CXX = ${cxx_wrapper}/MKOCTFILE_CXX = ${actual_cxx}/g" Makefile
sed -i "s/MKOCTFILE_F77 = ${f77_wrapper}/MKOCTFILE_F77 = ${actual_f77}/g" Makefile
```
That should be retaining potential compiler flags in these Makefile variables.
After that, proceed building the project with `make`.

I don't know how to integrate that into this build script though. I also don't know how to query the paths of the compiler wrappers. (The latter are potentially passed to the `configure` script as command line arguments or they are set as environment variables when the `configure` script is being run.)

alalazo: > I've read "compiler wrappers" being mentioned repeatedly. But what are these compiler wrappers? And why is it an issue if they are used?

https://spack.readthedocs.io/en/latest/packaging_guide_build.html#compiler-wrappers-and-flags

mmuetzel: Thank you for the link.

I'm still not sure if I understand what these compiler wrappers are. Are they only "valid" while packaging and are no longer "available" after a user installs a spack package? Or is there something more to it why `mkoctfile` can't use them?

Could the Octave package have an explicit dependency on the compiler wrapper package, so the wrappers are installed when Octave is installed?

cessenat:
This PR is not good enough, so it is closed.

PR #1515, opened 2025-09-16
legion: fix kokkos + cuda + openmp dependency variants

H030m:
This PR fixes the Kokkos dependency variants in the Legion package.

### Changes
- Added explicit kokkos dependency combinations:
  * `+kokkos +cuda ~openmp` → `kokkos@3.3.01:+cuda+cuda_lambda+wrapper`
  * `+kokkos +cuda +openmp` → `kokkos@3.3.01:+cuda+cuda_lambda+openmp+wrapper`
  * `+kokkos ~cuda +openmp` → `kokkos@3.3.01:~cuda+openmp`
  * `+kokkos ~cuda ~openmp` → `kokkos@3.3.01:~cuda`

### Testing
The following concretization tests now succeed:
spack spec -I legion %gcc
spack spec -I legion +kokkos %gcc
spack spec -I legion +kokkos +openmp %gcc
spack spec -I legion +cuda cuda_arch=80 %gcc
spack spec -I legion +kokkos +cuda cuda_arch=80 %gcc
spack spec -I legion +kokkos +cuda +openmp cuda_arch=80 %gcc

### Issue
Fixes #50103


alalazo:
@elliottslaughter  @pmccormick  @streichler

Not sure why maintainers were not pinged @alecbcs fyi

repos/spack_repo/builtin/packages/legion/package.py
```diff
@@ -142,8 +142,10 @@ class Legion(CMakePackage, ROCmPackage):
             when=f"+kokkos+cuda cuda_arch={arch} %clang",
         )
 
-    depends_on("kokkos@3.3.01:~cuda", when="+kokkos~cuda")
-    depends_on("kokkos@3.3.01:~cuda+openmp", when="+kokkos+openmp")
+    depends_on("kokkos@3.3.01:+cuda+cuda_lambda+wrapper", when="+kokkos+cuda~openmp")
+    depends_on("kokkos@3.3.01:+cuda+cuda_lambda+openmp+wrapper", when="+kokkos+cuda+openmp")
```
elliottslaughter: I'm not a Spack expert, but should this be merged with the `depends_on` lines up above (around 136-143)? Not sure if it matters to have these duplicated.

mpokorny: The Kokkos `+cuda` variant does not require the Kokkos `+wrapper` variant. It is only required under the conditions shown here
https://github.com/spack/spack-packages/blob/b6639d23bab5e27189e6223c14dc0d4f6309f714/repos/spack_repo/builtin/packages/kokkos/package.py#L363-L368
I think the right way to handle this is to leave `+wrapper` out of the Legion package, and require the user to provide it, as needed.

mpokorny: Alternatively, add `when` clauses that name the compiler, as in the lines @elliottslaughter referenced. In that case, you might still miss the other conditions under which `+wrapper` isn't needed, unless you expand the alternatives beyond gcc and clang.

H030m: You're right. I removed the unnecessary wrapper from the Legion package.
  * `-kokkos +cuda ~openmp` → `kokkos@3.3.01:+cuda+cuda_lambda+wrapper`
  * `-kokkos +cuda +openmp` → `kokkos@3.3.01:+cuda+cuda_lambda+openmp+wrapper`

I also checked what I did in the previous test.

I have updated the branch. Sorry to bother you again.

repos/spack_repo/builtin/packages/legion/package.py
```diff
@@ -142,8 +142,8 @@ class Legion(CMakePackage, ROCmPackage):
             when=f"+kokkos+cuda cuda_arch={arch} %clang",
         )
 
-    depends_on("kokkos@3.3.01:~cuda", when="+kokkos~cuda")
-    depends_on("kokkos@3.3.01:~cuda+openmp", when="+kokkos+openmp")
+    depends_on("kokkos@3.3.01:~cuda+openmp", when="+kokkos~cuda+openmp")
+    depends_on("kokkos@3.3.01:~cuda", when="+kokkos~cuda~openmp")
```
mpokorny: I don't see why the `cuda` variant is even mentioned here, at all. Isn't the point simply to propagate the Legion package `openmp` variant to Kokkos? The propagation (&c) of the `cuda` variant is already done in the preceding block, and the handling of `openmp` is orthogonal to that of `cuda`. Something like the following, modulo Kokkos version numbers, should suffice
```python
depends_on("kokkos +openmp", when="+openmp")
depends_on("kokkos ~openmp", when="~openmp")
```
We could discuss whether it should be required that Kokkos is configured without a backend that Legion doesn't use, but that's a different issue.
[Another side note: I think that it would be cleaner to set the Kokkos version dependency once, before any of the backend-specific stuff, and then you could avoid repeating it over and over.]

H030m: My original intention was to keep the program closer to the original, but I understand your point. CUDA is indeed a bit redundant here.

I tested removing `depends_on("kokkos~openmp", when="+kokkos~openmp")` and it works fine.  

Then, I think removing backends that the Legion doesn't need isn't necessary. Users have a good chance to test different combinations of kokkos. 

I also thought it would be worthwhile to fix the kokkos version, so I modified other kokkos dependencies.
```python
# Kokkos base version
depends_on("kokkos@3.3.01:", when="+kokkos")

# OpenMP backend
depends_on("kokkos+openmp", when="+kokkos+openmp")

#cuda-centric
cuda_arch_list = CudaPackage.cuda_arch_values
for arch in CudaPackage.cuda_arch_values:
#UCX transport dependency when using CUDA
depends_on(f"ucc cuda_arch={arch}", when=f"@25.03.0: network=ucx +cuda cuda_arch={arch}")

# Kokkos CUDA + compiler-specific wrapper
depends_on(
f"kokkos+cuda+cuda_lambda+wrapper cuda_arch={arch}",
when=f"+kokkos+cuda cuda_arch={arch} %gcc",
)
depends_on(
f"kokkos+cuda+cuda_lambda~wrapper cuda_arch={arch}",
when=f"+kokkos+cuda cuda_arch={arch} %clang",
)
...
for arch in ROCmPackage.amdgpu_targets:
depends_on(
f"ucc amdgpu_target={arch}",
when=f"@25.03.0: network=ucx +rocm amdgpu_target={arch}",
)
depends_on(
f"kokkos+rocm amdgpu_target={arch}",
when=f"+kokkos+rocm amdgpu_target={arch}",
)

depends_on("kokkos+rocm", when="+kokkos+rocm")
```

mpokorny: I only suggested not explicitly removing backends from the Kokkos requirement that are not used by Legion as something to consider, not anything to implement immediately. I don't know what will happen, for example, if Legion is built without CUDA, but then tries to link to a Kokkos library that has the CUDA backend enabled, as the choice of compiler might prove difficult. I've not tried that myself, and I would like to see some verification that it works both for Legion itself and applications that use Legion+Kokkos. A more conservative approach is to continue to propagate the exact Legion backend variants to Kokkos, and maybe try out the looser constraints in another PR.

I suspect, too, that there are might be changes coming in the Legion CMake build system support for Kokkos. It's probably best not to try to do too much with this PR.

repos/spack_repo/builtin/packages/legion/package.py
```diff
@@ -129,22 +129,28 @@ class Legion(CMakePackage, ROCmPackage):
     depends_on("hwloc", when="+hwloc")
     depends_on("libfabric", when="network=gasnet conduit=ofi-slingshot11")
 
+    # Kokkos base version
+    depends_on("kokkos@3.3.01:", when="+kokkos")
```
cedricchevalier19: Kokkos lowest version available in Spack is for now `3.7.02` so I think you can omit the version requirement here.

rbberger: Seconded.

repos/spack_repo/builtin/packages/legion/package.py
rbberger: This will conflict a bit with my ongoing work for #1231, but that's ok. Note that with the Legion/Realm repo split the Kokkos dependency basically moves to Realm. I would be ok with merging this one once the version restriction is removed and the Legion folks gave their blessing.

repos/spack_repo/builtin/packages/legion/package.py
H030m: I’ve pushed a new version for edit kokkos version.
Thanks for the suggestions and guidance!

Changes:
* depends_on("kokkos@3.3.01:", when="+kokkos") → depends_on("kokkos", when="+kokkos")

rbberger: You restored deprecated versions that were just removed. Please undo.

H030m: Thanks, I've removed wrong version in the lastest commit.

repos/spack_repo/builtin/packages/legion/package.py
rbberger: @elliottslaughter ping. merge?

elliottslaughter (approved):
LGTM

rbberger (approved):

PR #1518, opened 2025-09-16
intel-xed: add version 2025.06.08

mwkrentel:
Also versions 2024.08.15, 2024.11.04 and 2025.03.02. Update version for mbuild resource.
Update python to 3.9.

### DIFF
--- a/repos/spack_repo/builtin/packages/intel_xed/package.py
+++ b/repos/spack_repo/builtin/packages/intel_xed/package.py
@@ -24,6 +24,10 @@ class IntelXed(Package):
 
     # Current versions now have actual releases and tags.
     version("main", branch="main")
+    version("2025.06.08", tag="v2025.06.08", commit="dc6bdbe036515c7ef55d3b2d50bafc5232bfe534")
+    version("2025.03.02", tag="v2025.03.02", commit="1bdc793f5f64cf207f6776f4c0e442e39fa47903")
+    version("2024.11.04", tag="v2024.11.04", commit="d4d502003bfff51c55c2808804301a62878d7cc8")
+    version("2024.08.15", tag="v2024.08.15", commit="ea330dbdf42837029ddd8c690729fceb6e7f5e42")
     version("2024.05.20", tag="v2024.05.20", commit="7e88c3e00274a10daa6b9d053decc057f65aa0ec")
     version("2024.04.01", tag="v2024.04.01", commit="6d87b5481aa53b5ab1fc2b5a5622759c46746bf9")
     version("2024.02.22", tag="v2024.02.22", commit="d08a6f66f780a685f26322960cd3ae297dbad931")
@@ -51,9 +55,9 @@ class IntelXed(Package):
         name="mbuild",
         placement=mdir,
         git=mbuild_git,
-        tag="v2022.07.28",
-        commit="75cb46e6536758f1a3cdb3d6bd83a4a9fd0338bb",
-        when="@2022.07:9999",
+        tag="v2024.11.04",
+        commit="7c4497f41f576b43a80bb0f8d8452bbcfd58b6e2",
+        when="@2022.07:",
     )
 
     resource(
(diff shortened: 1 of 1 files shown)


mwkrentel:
@tldahlgren Sorry to nag again, would you like to review and merge?
It's really just adding a few recent versions.   Thanks

repos/spack_repo/builtin/packages/intel_xed/package.py
```diff
@@ -85,7 +89,7 @@ class IntelXed(Package):
     depends_on("cxx", type="build")
 
     # The current mfile uses python3 by name.
-    depends_on("python@3.7:", type="build")
+    depends_on("python@3.9:", type="build")
```
tldahlgren, line 92: Is it worth adding a conditional dependency for the older versions (https://spack.readthedocs.io/en/latest/packaging_guide_creation.html#conditional-dependencies) so they can still use `3.7:3.8`?

tldahlgren (commented):
Confirmed the version commit+tags but have a quick question about a dependency change.

tldahlgren:
> @tldahlgren Sorry to nag again, would you like to review and merge? It's really just adding a few recent versions. Thanks

No worries. Have a quick question.

mwkrentel:
Are you asking me that question literally?  Do `**I**` think it's worth
allowing `3.7:3.8` for older xed?  :-)

I know what you mean, but eh, whatever, I don't care.

In this case, I'll suggest no, because python 3.7 is
currently marked deprecated.  That tips the scales for me.

It's funny, I wrote what you suggested, but made the mistake of
actually testing it.  I got this odd failure:
```
$ ./spack spec intel-xed@2024.04.01 ^python@3.7  
==> Error: failed to concretize `intel-xed@2024.04.01 ^python@3.7` for the following reasons:
     3. Cannot satisfy 'python@3.7:' and 'python@3.7'
        required because intel-xed@2024.04.01 ^python@3.7 requested explicitly 
        required because intel-xed depends on python@3.7: when @:2024.7 
          required because intel-xed@2024.04.01 ^python@3.7 requested explicitly 
```

Huh?  `Cannot satisfy 'python@3.7:' and 'python@3.7'` ??

'Cuz 3.7 is deprecated.  Arrrrgh!
The error message really should include that.

Anyway, I'll stick with no, just so no one falls into that pit.
But if you really prefer, then Ok, I'll add it.
:-)  Thanks!

tldahlgren:
> Are you asking me that question literally? Do `**I**` think it's worth allowing `3.7:3.8` for older xed? :-)
> 
> I know what you mean, but eh, whatever, I don't care.
> 
> In this case, I'll suggest no, because python 3.7 is currently marked deprecated. That tips the scales for me.
> 
> It's funny, I wrote what you suggested, but made the mistake of actually testing it. I got this odd failure:
> 
> ```
> $ ./spack spec intel-xed@2024.04.01 ^python@3.7  
> ==> Error: failed to concretize `intel-xed@2024.04.01 ^python@3.7` for the following reasons:
>      3. Cannot satisfy 'python@3.7:' and 'python@3.7'
>         required because intel-xed@2024.04.01 ^python@3.7 requested explicitly 
>         required because intel-xed depends on python@3.7: when @:2024.7 
>           required because intel-xed@2024.04.01 ^python@3.7 requested explicitly 
> ```
> 
> Huh? `Cannot satisfy 'python@3.7:' and 'python@3.7'` ??
> 
> 'Cuz 3.7 is deprecated. Arrrrgh! The error message really should include that.
> 
> Anyway, I'll stick with no, just so no one falls into that pit. But if you really prefer, then Ok, I'll add it. :-) Thanks!

😆 Thanks for checking *and* letting me know.

tldahlgren (approved):

repos/spack_repo/builtin/packages/intel_xed/package.py
mwkrentel: It's complaining that not all conversations are resolved.
So, maybe by typing something here will take care of that.

PR #1524, opened 2025-09-17
Add missing `depends_on("cxx", type="build")` directive for adms

soruh:
The adms package builds C++ code but does not declare this in `package.py` which makes the build fail.
On my machine, adding the directive (as suggested [here)](https://github.com/spack/spack/discussions/51159) makes the build complete without any problems (using gcc14).

@cessenat


cessenat (approved):
Very good.
I suggest you remove "# generated" line 27 since you (and I) checked the actual dependence on both c and cxx.

cessenat (approved):
Perfect. Thank you.

alalazo (approved):

PR #1526, opened 2025-09-17
[package: py-setuptools-scm@7.1.0:] fixed build issue on dependency py-packaging <22.0

addy419:
@RobertMaaskant py-setuptools-scm fails with an 'unexpected keyword argument strip_trailing_zero' error when py-packaging <22 is used with py-setuptools-scm@7.1.0+. This is intended behaviour according to this issue https://github.com/pypa/setuptools/issues/4483#issuecomment-2236528158

I've managed to fix it in the runtime dependency for now. Since I haven't built py-setuptools-scm from scratch yet, you might want to fix the dependency issue there as well.

### DIFF
--- a/repos/spack_repo/builtin/packages/py_setuptools_scm/package.py
+++ b/repos/spack_repo/builtin/packages/py_setuptools_scm/package.py
@@ -74,6 +74,7 @@ class PySetuptoolsScm(PythonPackage):
         depends_on("py-typing-extensions", when="@7:8.0.4")
 
         depends_on("py-packaging@20.0:", when="@6.3:")
+        depends_on("py-packaging@22.0:", when="@7.1:")
 
         depends_on("py-importlib-metadata", when="@7: ^python@:3.7")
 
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/py_setuptools_scm/package.py
```diff
@@ -74,6 +74,7 @@ class PySetuptoolsScm(PythonPackage):
         depends_on("py-typing-extensions", when="@7:8.0.4")
 
         depends_on("py-packaging@20.0:", when="@6.3:")
+        depends_on("py-packaging@22.0:", when="@7.1:")
```
kwryankrattiger: In the linked comment, the `package` dependency it is asking for 24 and later

```
	"packaging>=24",
	"ordered-set>=3.1.1",
	"more_itertools>=8.8",
	"jaraco.text>=3.7",
	"importlib_resources>=5.10.2",
	"importlib_metadata>=6",
	"tomli>=2.0.1",
	"wheel>=0.43.0",
```

addy419: @kwryankrattiger I didn't notice that. Thanks for checking. I've updated both the build and run dependencies.

kwryankrattiger (approved):

PR #1527, opened 2025-09-17
ci: fix inconsistent behavior for packages with maintainers not in org

alecbcs:
TLDR: Fixed a bug where we wouldn't ping maintainers when one of them wasn't in the Spack org.

The GitHub API appears to be nondeterministic in how it responds to reviewer requests when a subset of the reviewers don't have triage permissions for a repository. Sometimes it'll add all of the users with permissions and respond `422` and sometimes it'll just with the error status code without adding anyone.

This PR modifies the existing approach so that we first verify the permissions for all maintainers before we attempt to request their review.

### DIFF
--- a/.github/workflows/bin/spack-reviewers.py
+++ b/.github/workflows/bin/spack-reviewers.py
@@ -21,14 +21,15 @@ def main():
     if token:
         headers["Authorization"] = f"Bearer {token}"
 
-    base_url = f"https://api.github.com/repos/{repository}/pulls/{pr_number}"
-    pull_request = requests.get(base_url, headers=headers).json()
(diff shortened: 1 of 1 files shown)


.github/workflows/bin/spack-reviewers.py
```diff
@@ -44,8 +45,16 @@ def main():
             except spack.repo.UnknownPackageError:
                 pass
 
+    # filter maintianers to those who have triage permissions in the repo
```
alalazo, line 48:
```suggestion
    # filter maintainers to those who have triage permissions in the repo
```

alalazo (dismissed):
LGTM, only a typo

alalazo (approved):

PR #726, opened 2025-07-23
py-ruff: add v0.12.4, v0.13.0

adamjstewart:

PR #1528, opened 2025-09-17
py-maturin: add v1.9.4

manuelakuhn:
https://github.com/PyO3/maturin/tree/v1.9.4


adamjstewart (approved):
Also added in #726 but that PR is taking too long.

manuelakuhn:
Sorry, did not see that. Thanks!

manuelakuhn:
The built for "e4sbuild" fails because it tries to build `py-maturin@1.9.4` with `py-setuptools@63.4.3` although it is clearly written in the `package.py` that it needs `py-setuptools@77:`.
All other builds and also my local build seem to work fine.

Any idea what is going on?

adamjstewart:
Ah, I had the same bug on my PR and a couple others. @alalazo is looking into it now.

manuelakuhn:
The problem that the concretizer still uses a wrong `py-setuptools` version against the package recipe is still happening. @alalazo is there any progress on that?

alalazo:
Closed this one since #2098 added a newer version. If v1.9.4 was specifically needed, please reopen.

PR #1535, opened 2025-09-17
aspell: add patch to support gcc@15: and clang

alecbcs:
Aspell hasn't had a release since 2023 and compilers have started enforcing stricter templating standards. 
Patch is also found in Homebrew.

https://github.com/Homebrew/homebrew-core/blob/c3308ba52e699e076bd24f7eb2fe4ef94ef6445f/Formula/a/aspell.rb#L559


manuelakuhn (dismissed):
LGTM and builds successfully on rocky 8 with gcc@15 (without the patch it fails there).

alecbcs:
@manuelakuhn should be ready for you to review again. Decided last minute I wanted to become a maintainer.

manuelakuhn (approved):

PR #1546, opened 2025-09-17
Deprecation removals: P

adamjstewart:
Not sure if OpenMPI changes are correct...


adamjstewart:
@spackbot fix style

jbigot:
On PDI side, we deprecate aggressively and remove the deprecated versions when new ones come in. Is it really required to remove them? If so, I would un-deprecate instead because I know they're still used.

adamjstewart:
Definitely not required. I'm trying to get us in the habit of removing deprecated versions once in a while. If they are still in use, maybe we should undeprecate them? Spack will still choose the newest version by default.

jbigot:
Our idea was to warn users that they are relying on a version that is not supported anymore upstream, but I admit I don't know what the exact semantic of spack deprecation is

adamjstewart:
Policy is defined here: https://spack.readthedocs.io/en/latest/packaging_guide_creation.html#deprecating-old-versions

We try to remove deprecated versions at least one minor release after deprecation, although we haven't historically been good at this.

jbigot:
Given this policy, I'd keep the PDI deprecated versions for now and wait for our normal cleanup that seems well aligned with the policy

jbigot:
Thank you very much!

repos/spack_repo/builtin/packages/openmpi/package.py
```diff
@@ -717,7 +717,6 @@ def patch(self):
     # PMIx is unavailable for @1, and required for @2:
     # OpenMPI @2: includes a vendored version:
     with when("~internal-pmix"):
-        depends_on("pmix@1", when="@2")
```
becker33, line 720: I think this needs some additional handling -- IIUC this means that `@:4 ~internal-pmix` is not possible to build anymore.

I think we need a `depends_on("pmix")` within this `with when` block, in addition to the more specific ones.

repos/spack_repo/builtin/packages/dd4hep/package.py
```diff
@@ -154,7 +154,6 @@ class Dd4hep(CMakePackage):
 
         # Specific version requirements
         depends_on("edm4hep@0.10.5:", when="@1.31:")
-        depends_on("podio@:0.16.03", when="@:1.23")
```
becker33, line 157: It looks like the last three lines of this block are vestigial at this point, and can be removed

becker33 (changes requested):
@natshineman am I right in understanding the `mvapich2x` and `mvapich2-gdr` packages should be fine to deprecate/remove things from because it's only very old versions in those?

becker33:
@jbigot please do not deprecate versions that you know are still broadly in use.

jbigot:
You recommend not to deprecate versions that upstream has decided to stop supporting?

becker33:
I would not deprecate a version simply because it is no longer receiving bugfixes. So much of the HPC and general open source ecosystem relies on things built once-upon-a-time against some old version of {perl or insert other package here} that still work because the interfaces are slow-moving.

If the package build system is changing such that maintaining the old version in Spack is cumbersome, definitely feel free to deprecate them. If the requirements are changing so that the whole dependency list has to be duplicated, and it's difficult to support both versions? Yeah go ahead and deprecate. If the package has a CVE for the old version, ABSOLUTELY deprecate it.

But if the only difference with the deprecation is you add a `deprecated=True` and eventually remove a single line of code, I don't think we're serving the community well by doing that.

adamjstewart:
While I agree with @becker33 on when to deprecate and when not to deprecate, I also defer to @jbigot as the maintainer of the PDI recipe to make all final decisions.

repos/spack_repo/builtin/packages/dd4hep/package.py
adamjstewart: Not sure I understand. Is this because older versions of podio don't exist? I don't usually remove those (except for Python) in case someone is using older external packages.

becker33 (approved):

PR #1553, opened 2025-09-17
gopls: add v0.20.0

alecbcs:
https://github.com/golang/tools/releases/tag/gopls%2Fv0.20.0

### DIFF
--- a/repos/spack_repo/builtin/packages/gopls/package.py
+++ b/repos/spack_repo/builtin/packages/gopls/package.py
@@ -17,10 +17,13 @@ class Gopls(GoPackage):
 
     license("BSD-3-Clause", checked_by="alecbcs")
 
+    version("0.20.0", sha256="1ff2a83be8be5a61b97fc5d72eab66f368ec20b52c513cc6656fc2e502e46f19")
     version("0.18.1", sha256="e49fae5dd964432a0ea1661868e858acd2aa66aaf7e1c1d646fb8506f15c8e52")
     version("0.17.1", sha256="5794ebd3302ef4fd08de284834b22810dbb17b7e08efeeaa9b96d5c94eb90d6d")
     version("0.16.2", sha256="be68b3159fcb8cde9ebb8b468f67f03531c58be2de33edbac69e5599f2d4a2c1")
 
+    depends_on("go@1.24.2:", type="build", when="@0.19:")
+    depends_on("go@1.23.4:", type="build", when="@0.18:")
     depends_on("go@1.23.1:", type="build", when="@0.17:")
 
     build_directory = "gopls"


manuelakuhn:
I can review this if you like, but unfortunately I am not that familiar with `go`. Can you tell me where the `go` dependency restrictions come from?
I though they would come from `go.mod`, but the version there are different from the spack `package.py`.

alecbcs:
Nice job checking on the dependencies! Really happy to see some other reviewers consistency go deps :)

`golang/tools` is a bit of a special case, it is a monorepo containing multiple sub projects each with their own `go.mod`, gopls is separately versioned from the rest of the tools and is located in `gopls/` inside there you'll notice a [go module file](https://github.com/golang/tools/blob/gopls/v0.20.0/gopls/go.mod) that *should* be consistent with the `gopls/v0.20.0` release.

repos/spack_repo/builtin/packages/gopls/package.py
```diff
@@ -17,10 +17,13 @@ class Gopls(GoPackage):
...
     license("BSD-3-Clause", checked_by="alecbcs")
 
+    version("0.20.0", sha256="1ff2a83be8be5a61b97fc5d72eab66f368ec20b52c513cc6656fc2e502e46f19")
     version("0.18.1", sha256="e49fae5dd964432a0ea1661868e858acd2aa66aaf7e1c1d646fb8506f15c8e52")
     version("0.17.1", sha256="5794ebd3302ef4fd08de284834b22810dbb17b7e08efeeaa9b96d5c94eb90d6d")
     version("0.16.2", sha256="be68b3159fcb8cde9ebb8b468f67f03531c58be2de33edbac69e5599f2d4a2c1")
 
+    depends_on("go@1.24.2:", type="build", when="@0.19:")
+    depends_on("go@1.23.4:", type="build", when="@0.18:")
     depends_on("go@1.23.1:", type="build", when="@0.17:")
```
manuelakuhn, line 28: Optional: Just in case you want to add it to be consistent
```
depends_on("go@1.19:", type="build")
```
(for version 0.16.2)

manuelakuhn (approved):

If you want to add the dependency just ping me, then I will approve it again, otherwise feel free to merge it if you like.

manuelakuhn:
Also thank you very much for the explanation :-)

repos/spack_repo/builtin/packages/gopls/package.py
alecbcs: I'm always split on adding dependency constraints like this for versions of the dependency we don't have in Spack. (The oldest version of Go that we have in Spack is 1.22.6, and I doubt we'll ever add something less than 1.19 due to CVEs....but technically you are correct.)

I'll merge this as is for now, but may decide to start being more constraint complete in the future.

PR #1554, opened 2025-09-17
liburing: add v2.12

alecbcs:
https://github.com/axboe/liburing/releases/tag/liburing-2.12


manuelakuhn (approved):

PR #1555, opened 2025-09-17
bfs: add v4.1

alecbcs:
https://github.com/tavianator/bfs/releases/tag/4.1


manuelakuhn (approved):

PR #1559, opened 2025-09-18
PythonPackage: empty libs/headers by default

haampie:
The current `PythonPackage.libs` and `PythonPackage.headers` default implementations are generally redundant, as evidenced by this pipeline which built a large number of Python packages including `numpy`, `tensorflow` and `torch` and their dependents: https://gitlab.spack.io/spack/spack-packages/-/pipelines/1235620.

From my understanding:

* There is no use of `spec["py-*"].libs` and `spec["py-*"].headers` in builtin
* You typically cannot link to the libraries in the usual way (`-L... -lfoo`) because they are Python modules with names of the form `_foo.cpython-<version>-<platform>.so`. They're meant to be opened by the Python interpreter.
* For packages like `numpy`, the typical use is to ask the module itself for its headers, using `numpy.get_include()`, when defining an extension. The headers is all you need, no linking to numpy libs is necessary.
* For `numpy` there are *some* static libraries (`numpy/random/lib/libnpyrandom.a`, `numpy/core/lib/libnpymath.a`), but those don't require our rpaths, and people are moving away from them, apparently.

So, I would be in favor of defaulting to an empty library and header list, because:

1. Spack's build system always calls `.libs` and `.headers`, which can be slow when packages are marked external
2. The heuristic can be return false positives (vendored/private headers) or too many results, leading to issues like https://github.com/spack/spack/issues/33237.
3. It's not much work to add it back on per-package basis if needed, and I'm happy to do that if needed

### DIFF
--- a/repos/spack_repo/builtin/build_systems/python.py
+++ b/repos/spack_repo/builtin/build_systems/python.py
@@ -2,8 +2,6 @@
 #
 # SPDX-License-Identifier: (Apache-2.0 OR MIT)
 
-import functools
-import operator
 import os
 import re
 import shutil
@@ -15,8 +13,6 @@ from spack.package import (
     ClassProperty,
     HeaderList,
     LibraryList,
-    NoHeadersError,
-    NoLibrariesError,
     PackageBase,
     Prefix,
     Spec,
@@ -27,8 +23,6 @@ from spack.package import (
     extends,
     filter_file,
     find,
-    find_all_headers,
-    find_all_libraries,
     has_shebang,
     join_path,
     path_contains_subdirectory,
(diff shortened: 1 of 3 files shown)


haampie:
Just noticed @rgommers concern on the original PR, which I agree with: https://github.com/spack/spack/pull/28527#issuecomment-1076796674. The support for the original PR seems mostly anecdotal where people did not remember why something did not work.

rgommers:
> otherwise there are just a few static libraries (`site-packages/numpy/random/lib/libnpyrandom.a`, `site-packages/numpy/core/lib/libnpymath.a`)

FYI we are intending to remove those when we can. `statsmodels` just removed its usage completely, and `scipy` removed `libnpyrandom` already (`libnpymath` is harder). I'm not aware of other major users, so hopefully we'll get this done next year.

alalazo (approved):
@adamjstewart Do you have any comment on this PR?

adamjstewart (approved):
If pipelines are passing, then I'm happy. There may be some old versions that break because of this. I would prefer to deprecate these when we encounter them.

PR #1563, opened 2025-09-18
flux-python: depends on flux-core, flux-security, json-glib

becker33:
`flux-python` build errors without these dependencies.


alecbcs (approved):
Looks good to me.

PR #1568, opened 2025-09-18
mapl: Add v2.61.0, conflict ESMF v9 for older versions

mathomp4:
This PR adds a `conflicts` to `mapl` for `esmf@9`. An issue was found in testing ESMF v9.0.0b03 (needed for MAPL 3 development) where a specific use of the (deprecated) `ESMF_AttributeGet` call triggered a failure (https://github.com/esmf-org/esmf/issues/493)

We also add a MAPL 2.61 which has the fixes needed to work with ESMF v9

NOTE: If needed, we could of course backport these fixes to older tags. When that happens, we can relax this.

### DIFF
--- a/repos/spack_repo/builtin/packages/mapl/package.py
+++ b/repos/spack_repo/builtin/packages/mapl/package.py
@@ -39,6 +39,7 @@ class Mapl(CMakePackage):
     version("develop", branch="develop")
     version("main", branch="main")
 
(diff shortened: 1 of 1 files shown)


climbfuji (approved):
Thanks!

alecbcs (approved):
Looks good to me. Thanks @mathomp4!

PR #1570, opened 2025-09-18
qt-tools: adds depends for c, cxx

emwjacobson:
This PR adds depends_on statements for c and cxx to the qt-tools package.

Without these, the installation attempts to use the system gcc/g++ compiler.

Tagging @wdconinc as the listed maintainer.

### DIFF
--- a/repos/spack_repo/builtin/packages/qt_tools/package.py
+++ b/repos/spack_repo/builtin/packages/qt_tools/package.py
@@ -43,6 +43,9 @@ class QtTools(QtPackage):
     # use of relative path in https://github.com/qt/qttools/blob/6.8.2/.gitmodules
     conflicts("+assistant", when="@6.8.2", msg="Incorrect git submodule prevents +assistant")
 
+    depends_on("c")
+    depends_on("cxx")
+
     depends_on("llvm +clang")
 
     depends_on("qt-base +network")


wdconinc (approved):
Good catch. Looks good to me. Thanks for the fix!

repos/spack_repo/builtin/packages/qt_tools/package.py
```diff
@@ -43,6 +43,9 @@ class QtTools(QtPackage):
     # use of relative path in https://github.com/qt/qttools/blob/6.8.2/.gitmodules
     conflicts("+assistant", when="@6.8.2", msg="Incorrect git submodule prevents +assistant")
 
+    depends_on("c")
+    depends_on("cxx")
```
alalazo, line 47: @wdconinc Is it correct to have `build,link` here, or should it be just `build`?

wdconinc: Yeah. I overlooked that.

emwjacobson: ~~Is this a change you'd like me to push through (as I was the one that added it), or something that either of you would like to change? Sorry about not catching this initially.~~

Nevermind, seems #3627 resolves this

PR #1573, opened 2025-09-18
libtree: add cxx compiler

Chrismarsh:
the cmake build looks for cxx and fails without this dep


Chrismarsh:
@haampie

haampie:
Only for version @:2, the rewrite of v3 is in pure C. Do you need v2 specifically?

Chrismarsh:
I didn't intentionally install `@2` and thus didn't realize it wasn't the newest. I just did a `spack install libtree` and got `@2`. I made the fix

haampie (approved):

PR #1579, opened 2025-09-19
memtailor: new package

d-torrance:
[Memtailor](https://github.com/Macaulay2/memtailor) is a C++ library of special-purpose memory allocators.  It is a dependency of Mathic, Mathicgb, and Macaulay2.


tldahlgren (approved):
Confirmed homepage and commit.

PR #1583, opened 2025-09-19
snakemake: add new versions

aphecetche:
@marcusboden @w8jcik I've just added new versions. Did not want to venture into messing with the dependencies just yet. I can give it a try though, unless of course you're already on it.

### DIFF
--- a/repos/spack_repo/builtin/packages/snakemake/package.py
+++ b/repos/spack_repo/builtin/packages/snakemake/package.py
@@ -16,6 +16,14 @@ class Snakemake(PythonPackage):
 
     license("MIT")
 
+    version("9.11.4", sha256="37b13337c87124d3cbd345f7a6ec976d4c485ac5b60d3d3379dcfe41b426dbb2")
+    version("9.11.3", sha256="db4e3405fbb104e561e4e466520df071ef77bf3e4faa205a1964dcb9c8519d25")
+    version("9.11.2", sha256="d0905992f6db8f887bb40234336a7979d9088886a7cabfb4334557c400d66ec9")
+    version("9.11.1", sha256="b04bd4804821b0f0910fc967154b72ec113cbfc171837322bf26ad9d3798f75c")
+    version("9.11.0", sha256="f4b64d1b52525487e74f05a4ee6912fca9a42af2736c522e6c3abd6d361ba1c5")
+    version("9.10.1", sha256="538308d13c0848c3d1a8b166f3a608d7a4800c3bac2313bd02b29fdf8c5a2c9f")
+    version("9.10.0", sha256="4a934412dbb1a205f7b698be9c59bbc29aeaae6e9c271bbf49defae81e11d351")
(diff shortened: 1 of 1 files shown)


w8jcik:
@aphecetche You need to update the dependencies, like here https://github.com/spack/spack/pull/43437/files. It is not necessary to update all the Snakemake plugins, but if the main package requires newer version of a dependency, one needs to look into their recipes as well.

Just adding the hashes to the main package is not very useful. Installed Snakemake might work at first glance, but it is not how the author of Snakemake intended and tested it. It will most likely start crashing at some point. Changes to the dependencies need to be checked all down to 8.25.2 version.

About hashes themselves, did you test all these versions? There are at least two approaches, either keep adding all the versions which bloats the package definition with untested versions or just add what you need and tested.

Why add all of these

```
    version("9.11.4", sha256="37b13337c87124d3cbd345f7a6ec976d4c485ac5b60d3d3379dcfe41b426dbb2")
    version("9.11.3", sha256="db4e3405fbb104e561e4e466520df071ef77bf3e4faa205a1964dcb9c8519d25")
    version("9.11.2", sha256="d0905992f6db8f887bb40234336a7979d9088886a7cabfb4334557c400d66ec9")
    version("9.11.1", sha256="b04bd4804821b0f0910fc967154b72ec113cbfc171837322bf26ad9d3798f75c")
    version("9.11.0", sha256="f4b64d1b52525487e74f05a4ee6912fca9a42af2736c522e6c3abd6d361ba1c5")
```

`9.11.0` is not worth adding if there is `9.11.4` already. Snakemake authors sometimes release a new version after a single commit.

As a side note, in case you just want to force Spack to install a version you can say `spack install snakemake@=9.11.4` without adding these hashes. It will be wrong, but Spack will obey.

aphecetche:
ok, I'll go over the dependencies then.

Concerning the hashes, no, I did not test all of them (currently, the one I needed is 9.10.1, btw). 
I could imagine that someone, for comparison purposes, would require all the tags, but I'd also agree that's probably only an edge case. I'll remove all but the latest of 9.10 and 9.11 in the next iteration of this PR.

thanks,

w8jcik:
> I could imagine that someone, for comparison purposes, would require all the tags

One can still install any version if the dependencies are correctly defined.

Adding all minor versions to the list is fine, but adding all patch versions is not necessary.

tldahlgren:
Added missing maintainers to Reviewers list.

aphecetche:
I do have updates on the following packages (which was indeed a little bit more involved than my initial PR :wink: ), but I need to review those once more before putting them in the PR. Will try to do that soon(ish).

py_snakemake_executor_plugin_slurm
py_snakemake_executor_plugin_slurm_jobstep
py_snakemake_interface_common
py_snakemake_interface_logger_plugins
py_snakemake_interface_report_plugins
py_snakemake_interface_scheduler_plugins
py_snakemake_interface_storage_plugins
py_snakemake_logger_plugin_snkmt
py_snakemake_storage_plugin_fs
snakemake

aphecetche:
superseded by #2604

PR #1584, opened 2025-09-19
py-codecarbon: add new package

thomas-bouvier:


repos/spack_repo/builtin/packages/py_codecarbon/package.py
```diff
@@ -0,0 +1,36 @@
...
+
+    homepage = "https://mlco2.github.io/codecarbon/"
+    pypi = "codecarbon/codecarbon-3.2.1.tar.gz"
+
+    version("3.2.1", sha256="04571e4d4758936587b7145de557f4aa0f53c83c771318f9a12d67ff039fc95c")
+
+    depends_on("python@3.7:", type=("build", "run"))
+    depends_on("py-setuptools@61:", type="build")
+
+    depends_on("arrow", type=("build", "run"))
```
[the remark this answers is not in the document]
bernhardkaindl: @thomas-bouvier, there are two `arrow` packages in spack, `arrow` and `py-arrow`.

Please check in the package which itself is this a Pyhon dependency:

https://github.com/mlco2/codecarbon/blob/master/pyproject.toml#L31

As this lists "arrow" in pyproject.toml#L31, this is a Python dependency, hence the review by Copilot appars to be spot-on, and you should add "py-" before "arrow" here.

repos/spack_repo/builtin/packages/py_codecarbon/package.py
```diff
@@ -0,0 +1,36 @@
...
+    depends_on("py-click", type=("build", "run"))
+    depends_on("py-fief-client +cli", type=("build", "run"))
+    depends_on("py-pandas@2.3.3:", when="^python@3.14:", type=("build", "run"))
+    depends_on("py-pandas", type=("build", "run"))
```
bernhardkaindl: As it looks like you should change the deps to depend on `py-arrow` instead of `arrow`, I think it also looks better to me to re-order the pandas dependency to list the easy, always-applied dependency first::

```suggestion
    depends_on("py-pandas", type=("build", "run"))
    depends_on("py-pandas@2.3.3:", when="^python@3.14:", type=("build", "run"))
```

repos/spack_repo/builtin/packages/py_codecarbon/package.py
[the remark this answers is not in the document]
thomas-bouvier: Good catch, thanks

repos/spack_repo/builtin/packages/py_fief_client/package.py
```diff
@@ -0,0 +1,27 @@
...
+
+from spack_repo.builtin.build_systems.python import PythonPackage
+
+from spack.package import *
+
+
+class PyFiefClient(PythonPackage):
+    """Fief Client for Python."""
+
+    homepage = "https://docs.fief.dev/integrate/python/"
```
tldahlgren: This URL doesn't appear to be valid as I get `Failed to open page`.

tldahlgren: Looking at `pypi.org` links,  perhaps the source link would be better: hhttps://github.com/fief-dev/fief-python.

Note the repository has been read-only since April 2025.

tldahlgren (commented):
Confirmed the homepage of one package -- the other is not available -- and the version checksums of both.

repos/spack_repo/builtin/packages/py_codecarbon/package.py
```diff
@@ -0,0 +1,37 @@
...
+
+
+class PyCodecarbon(PythonPackage):
+    """Track emissions from Compute and recommend ways to reduce their
+    impact on the environment."""
+
+    homepage = "https://mlco2.github.io/codecarbon/"
+    pypi = "codecarbon/codecarbon-3.2.1.tar.gz"
+
+    version("3.2.1", sha256="04571e4d4758936587b7145de557f4aa0f53c83c771318f9a12d67ff039fc95c")
```
tldahlgren, line 18: FWIW. `3.2.2` was released on February 1st.

tldahlgren (changes requested):
Confirmed the dependencies as well.

Only request is to provide a readable URL for the homepage.

repos/spack_repo/builtin/packages/py_fief_client/package.py
thomas-bouvier: Updated the homepage

repos/spack_repo/builtin/packages/py_codecarbon/package.py
thomas-bouvier: Added version 3.2.2

tldahlgren (approved):
Confirmed the newest version's sha256 and dependencies.

Thanks for the updates.

PR #1588, opened 2025-09-19
dealii: add v9.7

dabele:
- adds recently released dealii package versions 9.7.0 and 9.7.1
- upgrades required versions of dependencies 
  - taskflow: dealii 9.7 requires taskflow >3.10 
  - arborx: dealii 9.7 requires arborx  >1.3. I thought it would be easier to skip to version 1.5, doesn't use trilinos
- limit "compile examples" variant to version <9.7 since the corresponding cmake flag DEAL_II_COMPILE_EXAMPLES has been removed. There is a new process to compile examples, this can probably be added as a script to the package. not sure it's worth it.
- there is a new cmake flag `DEAL_II_USE_LTO`. I didn't add a variant for this flag since all cmake packages have the `ipo` variant already.

Tested locally with the default variants on ubuntu 22

pinging maintainers @jppelteret  @luca-heltai

### DIFF
--- a/repos/spack_repo/builtin/packages/dealii/package.py
+++ b/repos/spack_repo/builtin/packages/dealii/package.py
@@ -32,6 +32,8 @@ class Dealii(CMakePackage, CudaPackage):
     generator("make")
 
     version("master", branch="master")
+    version("9.7.1", sha256="0f2096ef83db54fdcebe9f3d148fa713f63f1c3f567941b53bcb4a1a8ea7de43")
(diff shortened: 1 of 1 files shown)


luca-heltai:
@dabele, thanks for working on this! Everything looks good from my side, but you'll have to make the pre-checks happy. :)

luca-heltai (approved):

alecbcs (approved):
Looks good to me. Thanks @dabele!

PR #1594, opened 2025-09-19
quantum-espresso patch for build_type cmake +libxc ^libxc+shared 

brunobruzzo:
**Package maintainers**: @bellenlau @tgorni @ye-luo 

**Issue that fixes:**

This build fails:
```console
$ spack install quantum-espresso +libxc
```

<details><summary>Error message</summary>

<pre>
  CMake variable LIBXC_ROOT is set to:

    /home/bbruzzo/spack/opt/spack/linux-icelake/libxc-7.0.0-jf2mj5ljx4tmpygbkuec7bzfqa2zmlfz/lib64

  For compatibility, find_package is ignoring the variable, but code in a
  .cmake module might still use it.
This warning is for project developers.  Use -Wno-dev to suppress it.

-- Found PkgConfig: /home/bbruzzo/spack/opt/spack/linux-icelake/pkgconf-2.5.1-qnrfdy6hwgjgxnyllge6jy5w5sgmsyvl/bin/pkg-config (found version "2.5.1")
-- Checking for one of the modules 'libxc>=5.1.2'
-- Could NOT find Libxc (missing: LIBXC_LIBRARIES) (Required is at least version "5.1.2")
-- Libxc searching failed in CMake Module mode, trying Config mode
CMake Warning at CMakeLists.txt:574 (find_package):
  Could not find a configuration file for package "Libxc" that is compatible
  with requested version "5.1.2".

  The following configuration files were considered but not accepted:

    /home/bbruzzo/spack/opt/spack/linux-icelake/libxc-7.0.0-jf2mj5ljx4tmpygbkuec7bzfqa2zmlfz/lib64/cmake/Libxc/LibxcConfig.cmake, version: 7.0.0
    /home/bbruzzo/spack/var/spack/environments/qe-patch/.spack-env/view/lib64/cmake/Libxc/LibxcConfig.cmake, version: 7.0.0

CMake Error at CMakeLists.txt:588 (message):
  Failed to find Libxc package (>=5.1.2) with Fortran enabled.

</pre></details>

[spack-build-out.txt](https://github.com/user-attachments/files/22429188/spack-build-out.txt)
[spack-build-env.txt](https://github.com/user-attachments/files/22429187/spack-build-env.txt)

Building qe with cmake, find_package() fails to accept valid libxc builds (>= 5.1.2) because of libxc claiming  SameMajorVersion compatibility.
https://github.com/ElectronicStructureLibrary/libxc/blob/4bd0e1e36347c6d0a4e378a2c8d891ae43f8c951/CMakeLists.txt#L589

**Proposed Patch:**
Already commited into develop branch of qe. https://gitlab.com/QEF/q-e/-/commit/cf52afc88f7fa61ba7a17aa7dbec5ca5f0b60ff3

This merge requests adds the patch on the QE repo to `package.py`.

```diff
diff --git a/repos/spack_repo/builtin/packages/quantum_espresso/package.py b/repos/spack_repo/builtin/packages/quantum_espresso/package.py
index 15f332dce1..0aebb0c2b2 100644
--- a/repos/spack_repo/builtin/packages/quantum_espresso/package.py
+++ b/repos/spack_repo/builtin/packages/quantum_espresso/package.py
@@ -443,6 +443,13 @@ class QuantumEspresso(CMakePackage, Package):
     # gipaw.x will only be installed with cmake if the qe-gipaw version is >= 5c4a4ce.
     patch("gipaw-eccee44.patch", when="@7.2+gipaw build_system=cmake")
 
+    # QE 7.4.1 +libxc build_system cmake can't rely on find_package() to handle 
+    # version check. With patch applied cmake checks after ${Libxc_FOUND} 
+    patch(
+        "https://gitlab.com/QEF/q-e/-/commit/cf52afc88f7fa61ba7a17aa7dbec5ca5f0b60ff3.diff",
+        sha256="f15616cac7a8de6a693c406a988db93347498522b219c67a6086613dcc3b67ce",
+        when="@7.4.1+libxc",
+    )
 
 class CMakeBuilder(cmake.CMakeBuilder):
     def cmake_args(self):
```

Let me know if I can further assist the maintainers.


ye-luo:
Your way only fixes 7.4.1 but the issue exists in all released QE versions. Could you just change instead
```
depends_on("libxc@5.1.2:", when="+libxc")
```
to
```
depends_on("libxc@5.1.2:~shared", when="+libxc")
```

ye-luo:
I should have mentioned earlier. Could you remove the patch line which is not needed since we are forcing static libxc?

brunobruzzo:
Not a problem. Patch removed. [a59e98c](https://github.com/spack/spack-packages/pull/1594/commits/a59e98c617340b6abec5583a10d1cd11566f23c3)

ye-luo (approved):

becker33 (approved):

becker33 (commented):
@ye-luo why are we preferring the solution of building with static libxc?

The issue is that an environment with `quantum-espresso` is now forcing `libxc~shared` on all of the rest of the packages, which is a substantial negative. The QE project upstream adopted the patch, that seems to be the way forward.

I see your argument that the patch only applies to the most recent version, but I think we can work around that.

Proposal:
1. bring back the patch for `quantum-espresso@7.4:7.4.1`
2. change the dependency to be
```
depends_on("libxc@5.1.2:", when="+libxc")
depends_on("libsc@5.1.2:~shared", when="@:7.3+libxc")  # cmake bug with +shared in older versions

**Review by becker33 — CHANGES_REQUESTED** [review]
Alternatively, it's possible the patch could apply cleanly to older versions, in which case that is a preferable solution.

**Comment by becker33** [comment]
Sorry for the review chaos -- I became more convinced over time that there was a better solution to this problem.

**Comment by ye-luo** [comment]
If you built libxc manually without using spack, the default is always static. For this reason, statlic libxc usage by QE is more taken care of historically. My preference is chosen to minimize the burden of spack recipe on maintainers and do minimal work to get things working for most users. For this reason, I'm generally against patches which usually only work for specific versions. As a result, it is not necessary to always yield the "best" spack recipe.

> Alternatively, it's possible the patch could apply cleanly to older versions, in which case that is a preferable solution.

This requires manually checking out old versions and testing them. It takes time and effort. I'm in favor of fast turnaround and unblock users.

FYI: I'm going to introduce QE 7.5 release in a separate PR and adjust to
```
depends_on("libxc@5.1.2:", when="+libxc")
depends_on("libsc@5.1.2:~shared", when="@:7.5+libxc")  # cmake bug with +shared in older versions
```
In this way, newer releases of QE won't be restricted as upstream QE has been fixed.

ye-luo:
ping @becker33

becker33 (approved):

PR #1597, opened 2025-09-19
Casacore +tests depend integrates casacore-measures

preshanth:
Add support for running casacore tests with automatic measures data integration.
When +tests variant is enabled, casacore automatically detects and uses
casacore-measures data, enabling the full test suite.

Changes:
- Added +tests variant (default=False for backward compatibility)
- Auto-detection of casacore-measures dependency and DATA_DIR configuration
- Added boost dependency with test framework when +tests enabled
- BUILD_TESTING enabled when +tests variant specified
- No changes to default build behavior
- 486/492 tests pass the failures are not measures related and differ due to exceeding thresholds


preshanth:
@mpokorny In the event it does not automatically assign it to you. This is what I put together. We can discuss if it does not fully cover the case of existing measures data. I am also having trouble updating measures from the casacore update-measures method.

preshanth:
```==> Error: ProcessError: Command exited with status 2:
    '/home/pjaganna/Software/spack/opt/spack/linux-zen/gmake-4.4.1-zid2l7c7ykkftqtbb2b3nowneo7qsv57/bin/make' '-j8' 'test'

2 errors found in build log:
     12428    492/492 Test  #45: tMultiFileLarge ..................   Passed   71.13 sec
     12429    
     12430    99% tests passed, 6 tests failed out of 492
     12431    
     12432    Total Test time (real) =  71.33 sec
     12433    
  >> 12434    The following tests FAILED:
     12435    	156 - tExternalStMan (Failed)
     12436    	169 - tStMan (Failed)
     12437    	225 - tLSQaips (Failed)
     12438    	226 - tLSQFit (Failed)
     12439    	256 - tFFTServer (Failed)
     12440    	355 - tMSSelection (Failed)
     12441    Errors while running CTest
     12442    Output from these tests are in: /tmp/pjaganna/spack-stage/spack-stage-casacore-3.7
              .1-pechndrl2r5qj2fxlwuqcvhg45ohhd4n/spack-build-pechndr/Testing/Temporary/LastTest
              .log
     12443    Use "--rerun-failed --output-on-failure" to re-run the failed cases verbosely.
  >> 12444    make: *** [Makefile:74: test] Error 8

See build log for details:
  /tmp/pjaganna/spack-stage/spack-stage-casacore-3.7.1-pechndrl2r5qj2fxlwuqcvhg45ohhd4n/spack-build-out.txt

See test log for details:
  /tmp/pjaganna/spack-stage/spack-stage-casacore-3.7.1-pechndrl2r5qj2fxlwuqcvhg45ohhd4n/install-time-test-log.txt
```
For the sake of completeness these are the failing tests.

repos/spack_repo/builtin/packages/casacore/package.py
```diff
@@ -51,6 +51,7 @@ class Casacore(CMakePackage):
     variant("shared", default=True, description="Build shared libraries")
     variant("tablelocking", default=True, description="Enable table locking")
     variant("threads", default=True, description="Use mutex thread synchronization")
+    variant("tests", default=False, description="Build and run tests")
```
mpokorny, line 54: The standard Spack way to build and run unit tests is with `spack install --test=root`. Seeing that setting `+tests` doesn't actually run or install the tests, I think that it's not a legitimate variant, and we should stick with keying off the `--test=root` argument. I think that's `self.run_tests`.

repos/spack_repo/builtin/packages/casacore/package.py
```diff
@@ -100,6 +103,11 @@ def cmake_args(self):
         args.append(self.define_from_variant("USE_MPI", "mpi"))
         args.append("-DPORTABLE=ON")  # let Spack determine arch build flags
 
+        # Use casacore-measures data directory if available
+        if spec.satisfies("^casacore-measures"):
+            measures_spec = spec["casacore-measures"]
+            args.append(self.define("DATA_DIR", measures_spec.prefix.share.data))
```
mpokorny, line 109: Setting `DATA_DIR` during the build, whether it's for testing or not, kind of requires `casacore-measures` to be available during runtime, or applications might fail unexpectedly. Yes, users can override the measures location, but I think that by default, they should not have to if we've built casacore with a valid `DATA_DIR` value.

repos/spack_repo/builtin/packages/casacore/package.py
```diff
@@ -121,12 +129,8 @@ def cmake_args(self):
...
             args.append(self.define("USE_PCH", True))
 
-        # tests won't pass unless measures data are installed, which
-        # we don't do in this package, and for which we don't yet
-        # provide any way of specifying at build time
-        #
-        # args.append(self.define('BUILD_TESTING', self.run_tests))
-        args.append(self.define("BUILD_TESTING", False))
+        # Enable testing when tests variant is enabled and measures data is available
+        args.append(self.define("BUILD_TESTING", spec.satisfies("+tests")))
```
mpokorny, line 133: The value of `BUILD_TESTING` can be simply `self.run_tests`

repos/spack_repo/builtin/packages/casacore/package.py
```diff
@@ -81,6 +82,8 @@ class Casacore(CMakePackage):
     depends_on("py-numpy@:1", when="@:3.6.0 +python")
     depends_on("py-numpy", when="@3.6.1: +python")
     depends_on("gsl", when="+dysco")
+    depends_on("casacore-measures", when="+tests")
+    depends_on("boost +system +filesystem +test", when="+tests")
```
mpokorny, line 86: I ran some tests building this package under various configurations, and I think that the dependencies needed for testing are not correct, beyond anything that you've added here. What I found works is
```python
    depends_on("python@2.6:", when="@:3.5.0", type=("build","run"))
    depends_on("python@3:", when="@3.6.0:", type=("build","run"))
    depends_on("boost +python", when="+python")
    depends_on("py-numpy@:1", when="@:3.6.0 +python")
    depends_on("py-numpy", when="@3.6.1: +python")
    depends_on("gsl", when="+dysco")
    # casacore-measures dependency without +measures is logically of
    # type "test", but that fails to set the environment for casacore
    # tests properly
    depends_on("casacore-measures", type="build")
    # When we build casacore with DATA_DIR value derived from
    # casacore-measures, we set a dependency on the casacore-measures
    # installation to prevent breaking the casacore installation by
    # removal of casacore-measures.
    depends_on("casacore-measures", when="+measures", type="run")
    depends_on("boost +system +filesystem +test")
```
Note that I also added a`measures` variant
```python
    variant(
        "measures", default=False,
        description="Set default measures directory from casacore-measures"
    )
```
that is specifically used to set the `DATA_DIR` variable during the build. But I also made changes to `casacore-measures` that optionally allows a user to refer to the measures data installed by that package through the use of an environment variable.

repos/spack_repo/builtin/packages/casacore_measures/package.py
```diff
@@ -28,3 +28,7 @@ def install(self, spec, prefix):
         mkdirp(self.prefix.share.data)
         update = Executable(self.prefix.bin.update_measures)
         update()
+
+    def setup_dependent_package(self, module, dependent_spec):
+        """Provide measures data directory to dependent packages."""
+        module.casacore_measures_data_dir = self.spec.prefix.share.data
```
mpokorny, line 34: The following provides access to the measures directory through an environment variable plus an aipsrc file.
```python

        # Install script to update measures
        mkdirp(self.prefix.bin)
        install("bin/update_measures", self.prefix.bin)
        # Install measures
        mkdirp(self.prefix.share.data)
        update = Executable(self.prefix.bin.update_measures)
        update()
        # Write an aipsrc file into share sub-directory that sets
        # "measures.directory" to our measures directory
        with tempfile.NamedTemporaryFile("w") as aipsrc:
            aipsrc.write(f"measures.directory: {self.prefix.share.data}\n")
            aipsrc.flush()
            install(aipsrc.name, join_path(self.prefix.share, "aipsrc"))

        """Add to CASARCFILES an aipsrc file that sets path to our
        measures data directory"""
        env.prepend_path("CASARCFILES", join_path(self.prefix.share, "aipsrc"))

    def setup_dependent_package(self, module, dependent_spec):
        """Provide measures data directory to dependent packages."""
        module.casacore_measures_data_dir = self.spec.prefix.share.data
```

mpokorny (changes requested):
I tried building this package in a variety of ways, focusing on the unit tests and use of `casacore-measures`. I've put together some changes that I think have the following advantages:
- Hard dependency on `casacore-measures` only when running tests. If there were some other method of ensuring that measures are available during the build (in a Spack world), we could reconsider this. casacore itself has enough hooks to avoid a runtime dependency on the `casacore-measures` package. 
- Option to provide a value of `DATA_DIR` at configure time through a variant that adds a dependency on `casacore-measures`
- A modification of the runtime environment of `casacore-measures` that adds the path to the measures directory in a way that casacore understands. This allows a user to create an environment (or `spack load casacore-measures`) under which casacore will naturally access the desired measures data.

There is still one potential issue that might occur
1. build `casacore` with `+measures`, which sets the default measures directory, but also adds a Spack dependency on `casacore-measures`
2. build an application with a dependency on `casacore`, but with a different `casacore-measures` package

Then there would be a  conflict when `concretizer:unify:true` because of the two different `casacore-measures` packages. This situation seems unlikely, and could be worked around outside of Spack, but I don't see an easy solution to it inside Spack

I added some code in my comments with suggested changes, but the association with the affected lines may not be entirely clear because of the limitations of this form.

repos/spack_repo/builtin/packages/casacore/package.py
mpokorny: The following suggested dependency
```python
    depends_on("casacore-measures", when="+measures", type="run")
```
is what could cause a problem with downstream packages that are built against a different `casacore-measures` package. The only reason to have this dependency is to ensure that the default measures directory exists while the `casacore` package exists. Removing the default measures directory isn't actually fatal to this package, however, as a user can always override the value at runtime...which is exactly what a runtime dependency on `casacore-measures` does. I'm not sure whether to keep the suggested dependency. Your comments?

PR #1598, opened 2025-09-19
frobby: new package

d-torrance:
Frobby is a software system for monomial ideal computations.  It is used by Macaulay2, Sage, and CoCoA.

### DIFF
--- a/repos/spack_repo/builtin/packages/frobby/package.py
+++ b/repos/spack_repo/builtin/packages/frobby/package.py
@@ -1,0 +1,43 @@
+# Copyright Spack Project Developers. See COPYRIGHT file for details.
+#
+# SPDX-License-Identifier: (Apache-2.0 OR MIT)
+
+from spack_repo.builtin.build_systems.makefile import MakefilePackage
+
+from spack.package import *
+
+
+class Frobby(MakefilePackage):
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/frobby/package.py
```diff
@@ -0,0 +1,43 @@
...
+    Frobenius problems with very large numbers."""
+
+    homepage = "https://github.com/Macaulay2/frobby"
+    git = "https://github.com/Macaulay2/frobby"
+
+    maintainers("d-torrance")
+
+    license("GPL-2.0-or-later", checked_by="d-torrance")
+
+    version("0.9.5", tag="v0.9.5")
```
alecbcs:
```suggestion
    version("0.9.5", tag="v0.9.5", commit="cbda56e8bb0d706f8cd7e6594a8a034797f53eb5")
```

To protect against the tag possibly moving in the future we should add an additional pin here on the commit.

alecbcs (dismissed):
One suggestion regarding version pinning to a specific commit. 

Otherwise this looks mostly good to me. I think you could extend the package with a shared variant to enable the shared library, but that isn't necessary for this to merge.

alecbcs (approved):
Looks good to me. Thanks!

PR #1599, opened 2025-09-20
cddlib: add gmp variant

d-torrance:
GMP support is expected by some dependencies such as gfan.

### DIFF
--- a/repos/spack_repo/builtin/packages/cddlib/package.py
+++ b/repos/spack_repo/builtin/packages/cddlib/package.py
@@ -23,8 +23,10 @@ class Cddlib(AutotoolsPackage):
     version("0.94m", sha256="70dffdb3369b8704dc75428a1b3c42ab9047b81ce039f12f427e2eb2b1b0dee2")
     version("0.94h", sha256="7382782c3834214b022c8b2898ed775a7bf915f2cb2acb73fa045d6fd9a3de33")
 
-    depends_on("c", type="build")  # generated
+    depends_on("c", type="build")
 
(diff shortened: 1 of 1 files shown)


d-torrance:
@spack/spack-releasers

Still waiting for a review. Should not be marked stale.

becker33 (commented):
How does the cddlib determine where to find gmp? Is it just from `LD_LIBRARY_PATH`?

It's slightly concerning to see the variant added with no code to distinguish build instructions based on it.

d-torrance:
> How does the cddlib determine where to find gmp? Is it just from `LD_LIBRARY_PATH`?

Basically, yes.  It looks to see if it can compile with `gmp.h` and link against `-lgmp` using the standard autoconf macros:

```autoconf
AC_CHECK_HEADER(gmp.h, [
  AC_CHECK_LIB([gmp], [__gmpz_init], [true])
])
```

> It's slightly concerning to see the variant added with no code to distinguish build instructions based on it.

There's no `--with-gmp` option or anything.  The configure script just looks to see if it gmp exists, and if it does, supports it:

```autoconf
dnl We do not build the gmp enabled library if we don't have gmp.
AM_CONDITIONAL([GMP], [test "x$ac_cv_lib_gmp___gmpz_init" = "xyes"])
```

becker33:
Ok, this will cause failures when `gmp` is available in the same directory as an external for another dependency and you want to install `~gmp`, but since there are no other link dependencies I don't think we'll hit that in this case.

becker33 (approved):

PR #1602, opened 2025-09-20
py-setuptools-cpp: remove optional developer deps

adamjstewart:
Only required dependencies are required. Optional developer dependencies should either be removed or controlled by a variant.


alecbcs (approved):
Agreed. Looks good to me.

PR #1603, opened 2025-09-20
py-torchgeo: fix SMP dep

adamjstewart:
Extracted out of #350 so we can fix this bug while waiting for the concretizer fix needed to merge #350.


alecbcs (approved):
Looks good to me. Thanks!

PR #1604, opened 2025-09-20
Python: add new versions

adamjstewart:
Also re-adding myself as a maintainer 🙂

### DIFF
--- a/repos/spack_repo/builtin/packages/python/package.py
+++ b/repos/spack_repo/builtin/packages/python/package.py
@@ -46,7 +46,7 @@ class Python(Package):
     list_depth = 1
     tags = ["windows", "build-tools"]
 
-    maintainers("skosukhin", "scheibelp")
+    maintainers("adamjstewart", "skosukhin", "scheibelp")
 
     phases = ["configure", "build", "install"]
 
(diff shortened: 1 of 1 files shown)


adamjstewart:
@spackbot style fix

adamjstewart:
@spackbot fix style

scheibelp (commented):
(I want to ask a couple questions before approving)

This seems to ratchet up the "deprecation aggressiveness" 

* previously: allowed some patch versions for 3.13 and 3.12 (3.11, 3.10 and 3.9 only had one patch version, so those are consistent with this PR)
* Now: appears to be moving to "one patch version per x.y version"

In the past we've spoken about keeping versions around for some length of time (e.g. at least 6 months, and e.g. 3.13.3-5 were added 3 months ago).

The effect in this case I think would be that users who have recently installed python and choose to "freeze" their x.y.z version may get concretizer errors without `--deprecated` (that is always a possibility with deprecation, but as mentioned above this is for some versions that were added recently).

So my questions are:

- Was this already discussed somewhere (that I could/should be reading)?
- If not, then is this deprecation based on a general principle (something that may be applied to other packages), or is it specific to the Python package?

adamjstewart:
History of the Python deprecation craze:

* I started deprecating all EOL unsupported versions: https://devguide.python.org/versions/
* @haampie (I think) started deprecating everything except the most recent patch version of every minor release

The only reason that old Python 3.13 patch releases aren't deprecated is because no one bothered to deprecate them. I wouldn't consider this PR as changing any policies.

There are good reasons to deprecate these (possible CVE fixes, faster concretizer, etc.) but I'm also fine with keeping them if people still want them.

scheibelp (approved):
I'm interested in discussing an "official" policy (perhaps just for Python, perhaps for packages in general), but that does not need to hold this up.

PR #1605, opened 2025-09-20
lrslib: add v7.3

d-torrance:
Also:
* Update to MakefilePackage
* Stop depending on c++ compiler
* Use upstream Makefile after v6.2
* Add patch fixing "multiple definition" error in v6.2
* Remove "generated" comments


d-torrance:
@spack/spack-releasers

Still waiting for a review. Should not be marked stale.

becker33 (approved):

PR #1615, opened 2025-09-20
py-nipype: add v1.10.0

manuelakuhn:
https://github.com/nipy/nipype/tree/1.10.0

but requirements for dependency versions can be found here:
https://github.com/nipy/nipype/blob/1.10.0/nipype/info.py


alecbcs (approved):
Looks good to me. Thanks!

PR #1624, opened 2025-09-21
bdw-gc: add "cxx" variant for C++ support

d-torrance:
Also remove some "generated" comments.


tldahlgren (approved):
Confirmed the option is available with the ~oldest listed and newest versions.

PR #1627, opened 2025-09-21
libunwind: fix build and add v1.8.2

manuelakuhn:
Added a patch to solve
```
==> Error: ProcessError: Command exited with status 2:
    '/spack/opt/spack/linux-skylake/gmake-4.4.1-2mokir4qvk25s327xhzajy5frrkytic6/bin/make' '-j16' 'V=1'

3 errors found in build log:
     879    mv -f $depbase.Tpo $depbase.Po
     880    depbase=`echo Gtest-resume-sig.o | sed 's|[^/]*$|.deps/&|;s|\.o$||'`;\
     881    /spack/opt/spack/linux-skylake/compiler-wrapper-1.0-7ovawteudbzdbaxhap54v3kxdsq2dngj/li
            bexec/spack/gcc/gcc -DHAVE_CONFIG_H -I. -I../include  -I../include -I../include/tdep-x8
            6_64 -I../src -D_GNU_SOURCE -DNDEBUG -fno-optimize-sibling-calls -g -O2 -fexceptions -W
            all -Wsign-compare -D__EXTENSIONS__ -MT Gtest-resume-sig.o -MD -MP -MF $depbase.Tpo -c 
            -o Gtest-resume-sig.o Gtest-resume-sig.c &&\
     882    mv -f $depbase.Tpo $depbase.Po
     883    In file included from Ltest-nomalloc.c:4:
     884    Gtest-nomalloc.c: In function 'malloc':
  >> 885    Gtest-nomalloc.c:49:12: error: too many arguments to function 'func'; expected 0, have 
            1
     886       49 |     return func(s);
     887          |            ^~~~ ~
  >> 888    make[1]: *** [Makefile:1740: Ltest-nomalloc.o] Error 1
     889    make[1]: *** Waiting for unfinished jobs....
     890    make[1]: Leaving directory '/tmp/root/spack-stage/spack-stage-libunwind-1.8.2-uqzdhohdw
            a7ktxem24isbri7lpzap2ui/spack-src/tests'
  >> 891    make: *** [Makefile:620: all-recursive] Error 1
```
 
and without adding a `libtool` build dependency I get errors like
```
>> 935     /spack/opt/spack/linux-skylake/binutils-2.45-2y4uwvl5bqxogicvdtfm7b4hz6fyeg3h/bin/ld: 
		 ../src/.libs/libunwind-x86_64.so: undefined reference to `dwarf_reg_state_pool'
>> 936     /spack/opt/spack/linux-skylake/binutils-2.45-2y4uwvl5bqxogicvdtfm7b4hz6fyeg3h/bin/ld: 
		 ../src/.libs/libunwind-x86_64.so: undefined reference to `dwarf_cie_info_pool'
>> 937     /spack/opt/spack/linux-skylake/binutils-2.45-2y4uwvl5bqxogicvdtfm7b4hz6fyeg3h/bin/ld: 
		 ../src/.libs/libunwind-x86_64.so: undefined reference to `_Ux86_64_dwarf_init'  
```

While at it I also added version 1.8.2

### DIFF
--- a/repos/spack_repo/builtin/packages/libunwind/package.py
+++ b/repos/spack_repo/builtin/packages/libunwind/package.py
@@ -12,8 +12,9 @@ class Libunwind(AutotoolsPackage):
     the call-chain of a program."""
 
     homepage = "https://www.nongnu.org/libunwind/"
-    url = "https://github.com/libunwind/libunwind/releases/download/v0.0.0/libunwind-0.0.0.tar.gz"
+    url = "https://github.com/libunwind/libunwind/releases/download/v1.8.2/libunwind-1.8.2.tar.gz"
     git = "https://github.com/libunwind/libunwind"
+
     maintainers("mwkrentel")
 
     tags = ["e4s"]
@@ -22,6 +23,7 @@ class Libunwind(AutotoolsPackage):
 
     version("master", branch="master")
     version("1.8-stable", branch="v1.8-stable")
+    version("1.8.2", sha256="7f262f1a1224f437ede0f96a6932b582c8f5421ff207c04e3d9504dfa04c8b82")
     version("1.8.1", sha256="ddf0e32dd5fafe5283198d37e4bf9decf7ba1770b6e7e006c33e6df79e6a6157")
     version("1.7-stable", branch="v1.7-stable")
     version("1.7.2", sha256="a18a6a24307443a8ace7a8acc2ce79fbbe6826cd0edf98d6326d0225d6a5d6e6")
@@ -80,12 +82,13 @@ class Libunwind(AutotoolsPackage):
     depends_on("c", type="build")  # generated
     depends_on("cxx", type="build")  # generated
 
+    depends_on("libtool", type="build")
+
     # The libunwind releases contain the autotools generated files,
     # but the git repo snapshots do not.
     reconf_versions = "@master,1.6-stable,1.7-stable,1.8-stable"
     depends_on("autoconf", type="build", when=reconf_versions)
     depends_on("automake", type="build", when=reconf_versions)
-    depends_on("libtool", type="build", when=reconf_versions)
     depends_on("m4", type="build", when=reconf_versions)
 
     depends_on("xz", type="link", when="+xz")
@@ -101,6 +104,13 @@ class Libunwind(AutotoolsPackage):
     conflicts("target=aarch64:", when="@1.8:")
 
     provides("unwind")
+
+    # Fix bad prototype for malloc() in test
+    patch(
+        "https://src.fedoraproject.org/rpms/libunwind/raw/49b1c9d51f8194546ba559f3f20e10889c8a073a/f/457612f470f8c0e718cdf7f14ef1ecb583f3b3a6.patch",
+        sha256="4562c231f1051bd327cf27b6940445e5c0d83e5d8427a6ca36c9f0853b3e4a6d",
+        when="@1.8:",
+    )
 
     def url_for_version(self, version):
         if version == Version("1.5.0"):


alecbcs (approved):
Looks good to me. Thanks!

mwkrentel:
@alecbcs First, I protest the quick merging without giving me as maintainer
adequate time to review.  In the future, please don't submit something
on a Sunday and expect it to be done by Monday morning.

You could have added version 1.8.3 which has been out for about 3 weeks.
```
version("1.8.3", sha256="be30d910e67f58d82e753231f1357f326a1a088acf126b21ff77e60aab19b90b")
```

Why add `libtool` unconditionally?
```
depends_on("libtool", type="build")
```
The releases are not supposed to need autotools, and they build
cleanly for me.

I'm willing to believe it fails for you and I guess adding the
dependency fixes it, but I'd like to better understand why,
what's different?

About the patch, applying it `when="@1.8:"` breaks the build for
`master` and will break the build for `1.8-stable` if/when the fix is
backported there.

Now, `master` and `1.8-stable` are already broken for different
reasons that I don't really understand, but you shouldn't make it worse.

Since `1.8-stable` is a moving target, there is no perfect solution.
But you should avoid including `master` in `when`.  Maybe use
`when="@1.8"` for just the 1.8.x versions.

manuelakuhn:
Sorry if this was merged too fast. @alecbcs is working on reviewing a lot of PRs these days, which I am extremely thankful for. I guess not waiting for the maintainer was simply an overlook.

There is no restriction on `libtool` because none of the versions will install without `libtool` (I tested 1.8, 1.7, 1.6, 1.5, 1.4, and 1.3 and got the error message above for all of them) including `master`. The only version where I get a different error is 1.2 (there I run into https://github.com/libunwind/libunwind/issues/58 independently if `libtool` is added or not). `1.8-stable` does also not build independently of the patch or `libtool`.

The error for `master` is:

```
==> libunwind: Executing phase: 'autoreconf'
==> Error: RuntimeError: Cannot generate configure: missing dependencies: libtool.

Please add the following lines to the package:

    depends_on("libtool", type="build", when="@master build_system=autotools")

Update the version (when="@master") as needed.

/spack-packages/repos/spack_repo/builtin/build_systems/autotools.py:600, in autoreconf:
        597        ensure_build_dependencies_or_raise(
        598            spec=spec,
        599            dependencies=["autoconf", "automake", "libtool"],
  >>    600            error_msg="Cannot generate configure",
        601        )
        602
        603        tty.msg("Configure script not found: trying to generate it")
```

For the newer version and the patch:
I am happy to open a follow up pull request that adds version 1.8.3 and restricts the patch to version `1.8` (or even further depending if it is still needed for 1.8.3). Thanks for the hint.
I did only add 1.8.2 because I encountered this problem for the first time already some weeks ago, now wanted to finally open a PR for it and simply forgot to check if there was a newer version released in the meantime.

As additional info:
The system I am testing this with, is a rocky 8 minimal container and I am using `gcc@15` (built with spack). This is what I will need for the production system.

mwkrentel:
All of these things (except the stable branches) build cleanly for me.

You may as well leave the recipe alone for now.
There are a few things that need fixing for libunwind.
But I can't really address them until I understand why they fail for
you and what are we doing differently.

I'll try the container, that could be it.

alecbcs:
@mwkrentel while we may defer to a maintainer to support in reviewing a PR, it's not a hard block for simple PRs the merge to develop team is comfortable testing themselves. @manuelakuhn is a long time trusted member of the team and I verified that their modifications fixed the build in minimal environments where `libtool` is not already available.

Autotools components from a host system often "bleed" into builds since we do not strip everything out of `/usr/bin` in Spack's isolated build environments. As such it's often the case that builds will work on a machine even if you haven't specified every required build dependency. To validate builds in a minimal environment you can check builds with https://github.com/spack/spack-distroless or with a minimal container like rocky 8.

> You may as well leave the recipe alone for now.

Please do not tell contributors not to modify a recipe. As a maintainer you'll be pinged on proposed PRs, but you do not have authority to unilaterally block contributions that follow best practices and fix bugs that the community can verify. For follow up PRs in the future we're happy to give you a week to respond to a complex PRs before the merge-to-develop team begins work on them. It's fine if you're annoyed with me. We've gotten behind on accepting PRs in a timely manner and I wanted to work my way through a backlog of PRs from specific maintainers and this PR did get reviewed and accepted rather quickly, but please do not take out that frustration on other members of the community who are not part of merge-to-develop.

mwkrentel:
@alecbcs As I said, this PR breaks the build for master because it tries to
apply a patch that is already there.  Were you comfortable testing that?

> You may as well leave the recipe alone for now.

I think you're misreading what I was trying to say.
All I meant was that although the current recipe (after this PR) has
some remaining things that should be fixed, they're not urgent.
Don't feel pressured to drop everything and fix them right now.
Of course, I didn't mean, "keep your hands off."

Basically, you're telling me that my opinion is not desired or useful,
and I consider that somewhat inappropriate.

PR #1638, opened 2025-09-22
Added conflicts for `cuda_arch` values in range 50-60 and `py-torch@2.8: ^cuda@12.8:`

abhishek1297:
See this: https://github.com/pytorch/pytorch/releases/tag/v2.8.0#backwards-incompatible-changes

### DIFF
--- a/repos/spack_repo/builtin/packages/py_torch/package.py
+++ b/repos/spack_repo/builtin/packages/py_torch/package.py
@@ -149,7 +149,7 @@ class PyTorch(PythonPackage, CudaPackage, ROCmPackage):
             conflicts(
                 f"cuda_arch={val}",
                 msg="PyTorch 2.8.0+ does not support sm_50–sm_60 architectures "
-                "with CUDA 12.8 or newer. Use CUDA 12.6 instead."
+                "with CUDA 12.8 or newer. Use CUDA 12.6 instead.",
             )
 
     # Required dependencies
--- a/repos/spack_repo/builtin/packages/py_torch/package.py
+++ b/repos/spack_repo/builtin/packages/py_torch/package.py
@@ -143,6 +143,18 @@ class PyTorch(PythonPackage, CudaPackage, ROCmPackage):
         "https://developer.nvidia.com/cuda-gpus",
     )
 
+    # https://github.com/pytorch/pytorch/releases/tag/v2.8.0#backwards-incompatible-changes
+    with when("@2.8: +cuda ^cuda@12.8:"):
+        for val in CudaPackage.cuda_arch_values:
+            if val.isdigit() and 50 <= int(val) <= 60:
+                conflicts(
+                    f"cuda_arch={val}",
+                    msg="PyTorch 2.8.0+ does not support "
+                    "sm_50–sm_60 architectures "
+                    "with CUDA 12.8 or newer. "
+                    "Use CUDA 12.6 instead.",
+                )
+
     # Required dependencies
     depends_on("c", type="build")
     depends_on("cxx", type="build")
--- a/repos/spack_repo/builtin/packages/py_torch/package.py
+++ b/repos/spack_repo/builtin/packages/py_torch/package.py
@@ -145,15 +145,12 @@ class PyTorch(PythonPackage, CudaPackage, ROCmPackage):
 
     # https://github.com/pytorch/pytorch/releases/tag/v2.8.0#backwards-incompatible-changes
     with when("@2.8: +cuda ^cuda@12.8:"):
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/py_torch/package.py
```diff
@@ -143,6 +143,15 @@ class PyTorch(PythonPackage, CudaPackage, ROCmPackage):
...
     )
 
+    # https://github.com/pytorch/pytorch/releases/tag/v2.8.0#backwards-incompatible-changes
+    with when("@2.8: +cuda ^cuda@12.8:"):
+        for val in [50, 52, 53, 60]:
+            conflicts(
+                f"cuda_arch={val}",
+                msg="PyTorch 2.8.0+ does not support sm_50–sm_60 architectures "
+                "with CUDA 12.8 or newer. Use CUDA 12.6 instead.",
+            )
```
adamjstewart, line 153: Based on https://github.com/pytorch/pytorch/blob/v2.8.0/torch/cuda/__init__.py#L259, it's a bit more nuanced. Let's add a comment to check this file for each new release. @alalazo @becker33 is it possible to add ranges for variant values, e.g.,
```python
# See CUDA_ARCHES_SUPPORTED in torch/cuda/__init__.py
requires("cuda_arch=50:90", when="@2.8:+cuda^cuda@12.6")
requires("cuda_arch=70:120", when="@2.8:+cuda^cuda@12.8:12.9")
```
It's also unclear to me if this is a compatibility issue or a support issue. Like, do older GPUs not work because the binary wheels PyTorch ships aren't compiled for them, or because the source code is no longer compatible?

adamjstewart (changes requested):

repos/spack_repo/builtin/packages/py_torch/package.py
abhishek1297: It's deprecation for older GPUs by CUDA toolkit. It seems 12.8 still supports Maxwell and Pascal architectures but PyTorch has dropped them.

https://docs.nvidia.com/cuda/cuda-toolkit-release-notes/index.html#deprecated-architectures

https://dev-discuss.pytorch.org/t/cuda-toolkit-version-and-architecture-support-update-maxwell-and-pascal-architecture-support-removed-in-cuda-12-8-and-12-9-builds/3128

adamjstewart: Dropped them in their sdist (source code build) or bdist (pre-compiled wheels)?

adamjstewart: Also, is 12.6 the oldest supported CUDA? We currently allow CUDA 11+.

abhishek1297: > Dropped them in their sdist (source code build) or bdist (pre-compiled wheels)?

Not sure about the source build but the wheels are now built only for sm 70+

https://github.com/pytorch/pytorch/blob/60b4791d084bf1c0e7e737297f905c18660a1d8e/.ci/manywheel/build_cuda.sh#L57

adamjstewart: Can you try building with sm60 and see if the build fails or if there is only a harmless warning?

abhishek1297: > Also, is 12.6 the oldest supported CUDA? We currently allow CUDA 11+.

> Based on https://github.com/pytorch/pytorch/blob/v2.8.0/torch/cuda/__init__.py#L259, it's a bit more nuanced. 


It seems like that the link you shared only has 3 CUDA versions (12.6, 12.8, and 12.9) supported for pytorch 2.8. This should be added:
```python
depends_on("cuda@12.6:", when="@2.8:+cuda")
```

adamjstewart: Still not sure if those are working versions or just documented versions. It would be nice to know if cuda 12.5 causes any build failures before adding such a limit.

abhishek1297: It works fine with CUDA 12.5.

However, a harmless warning may appear if the compute capability is below 50 or above 120. To address this warning for PyTorch `2.8:` versions, it may be necessary to adjust the lower CUDA version bound from `cuda@11:` to `cuda@12:` based on [CUDA Wikipedia page](https://en.wikipedia.org/wiki/CUDA#:~:text=12.0%20%E2%80%93%2012.6,5.0) which shows that CUDA 12.x only supports CC >= 50

PR #1639, opened 2025-09-22
vecgeom: new pre-release 2.0.0-rc.7

sethrj:
This adds a new development version for VecGeom. The only cmake change is that it doesn't pull in Doxygen from the system environment. (It wouldn't use it before, but doesn't even look for it now.)


tldahlgren (approved):
Version sha256 has been confirmed.

@sethrj @drbenmorgan Any objections to this PR being merged as is?

sethrj:
It's pretty minor. Thanks for approving!

PR #1640, opened 2025-09-22
gtksourceview: add version 4.8.4

KineticTheory:
+ Newer versions require the meson build system, so adjustments were made to allow Autotools for older versions and Meson for newer versions.
+ Dependencies corrected by removing max versions of `gtksourceview`. All newer versions of `gtksourceview` require `glib`, `gtkplus`, and `libxml`.
+ New dependency on `vala` added for versions `gtksourceview@4.8:`.

### DIFF
--- a/repos/spack_repo/builtin/packages/gtksourceview/package.py
+++ b/repos/spack_repo/builtin/packages/gtksourceview/package.py
@@ -3,7 +3,6 @@
 # SPDX-License-Identifier: (Apache-2.0 OR MIT)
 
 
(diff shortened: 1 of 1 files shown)


KineticTheory:
@spackbot fix style

KineticTheory:
@michaelkuhn Can you review this change? Thx.

michaelkuhn (approved):
Built successfully on Fedora 42 with GCC 15 (after applying #1723 and #1724).

KineticTheory:
@rbberger Can you review this one? TIA

PR #1642, opened 2025-09-22
build_systems/python: fix check for whether a file is executable

becker33:
PythonExtension overrides the default `add_files_to_view` method to ensure that shebang lines in python `bin/` files are rewritten to point to the python in the view.

The logic in this override has a bit-mask for determining whether the file is executable. The bit-mask is not correctly identifying files that are executable. This PR changes the hardcoded bitmask and replaces it with one constructed from constraints from python's `stat` module.


kwryankrattiger (approved):

alecbcs (approved):
Also looks good to me. Thanks!

PR #1644, opened 2025-09-22
py-frozendict: add v2.4.6

manuelakuhn:
https://github.com/Marco-Sulla/python-frozendict/tree/v2.4.6


alecbcs (approved):
Looks good to me. Thanks!

PR #1645, opened 2025-09-22
build_systems/oneapi: Fix for openmp lib detection

teabagk7:
Removing the context manager that caused the failure.

### DIFF
--- a/repos/spack_repo/builtin/build_systems/oneapi.py
+++ b/repos/spack_repo/builtin/build_systems/oneapi.py
@@ -226,10 +226,9 @@ class IntelOneApiLibraryPackage(IntelOneApiPackage):
             )
 
         # query the compiler for the library path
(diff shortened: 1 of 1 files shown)


hfp (approved):
Looks good to me!

I think the context was perhaps necessary to make this subsequent properties referring valid data (see also https://github.com/spack/spack/pull/45189).

Apparently, determining the information as currently implemented is brittle. This line was subject of previous maintenance.

Thank you, Konstantin for fixing this. We now need a writer to approve this PR for merge.

teabagk7:
Hey @alalazo, would you mind taking a look when you have a moment?

alalazo (approved):
`openmp_libs` is still very brittle as a method, but I agree the context manager is not needed since the compilers became dependencies. As such they affect the environment as every other node, and shouldn't need a special way of doing things.

PR #1650, opened 2025-09-22
Resolves a python@12 ^py-numpy@1.26.2:1.26:3 conflict; adds depends_on('cxx') to lib tool (replaces PR 1572)

markyoder:
Resolves compile issues originally encountered building py-torch. Our application requires [python@12.x](mailto:python@12.x), which I believe is what triggers py-torch to prefer [py-numpy@1.26.x](mailto:py-numpy@1.26.x). The That combination of python@3.12 ^py-torch@1.26.2:1.26.3 breaks with a "cannot generate metadata" error.

It may be possible to fix this by specifying a different version of py-setuptools, but I fixed it with
py_numpy/package.py:
conflicts('@:1.26.3', when='^python@3.12', msg="Known bug with @1.26.2, 1.26.3 when ^python@3.12")

In the course of building that, I also added:
libtool/package.py


markyoder:
Ok. I think it will also not be possible to push up this one. I believe the problem is that my GitHub repo is an organization, not a personal account. I'll move the repo and see how that goes...

markyoder:
`spack style --fix py-numpy`

markyoder:
Apologies for the mess. Closing this PR; reopening the original PR. "Allow edits" was not available, because I had forked Spack-packages to an Organization, not personal repo. I transferred the repo, now the option is available.

PR #1652, opened 2025-09-22
SST-Core: Do not set compilers manually

plavin:
I do not believe there is a need to set the compilers in this manner. Autotools should detect the correct compilers. 

Setting `CXX=mpicxx` causes sst-elements to also choose `mpicxx` instead of `gcc`, which causes issues with a library in Ariel, as it ends up searching for the wrong `libmpi.so`. If we do need to keep setting compilers this way, we should be using `MPICC` and `MPICXX` instead of `CC` and `CXX`.

### DIFF
--- a/repos/spack_repo/builtin/packages/sst_core/package.py
+++ b/repos/spack_repo/builtin/packages/sst_core/package.py
@@ -126,10 +126,6 @@ class SstCore(AutotoolsPackage):
 
         if self.spec.satisfies("+pdes_mpi"):
             args.append("--enable-mpi")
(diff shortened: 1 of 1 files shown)


berquist (approved):
I installed via
```bash
spack install --keep-stage --fresh sst-elements +ariel_mpi ^openmpi@=4.1.8 %gcc@=8.5.0
```
(and without `+ariel_mpi`) and confirmed that `CC`, `SPACK_CC`, and `MPICC` are correct.

alalazo (approved):

PR #1653, opened 2025-09-22
libproxy: Added new versions, build system

emwjacobson:
Bumping libproxy to newer versions.

The build system changed from cmake to meson with release 0.5.

Another subtle change is the change of the "python" variant from @0.4.**18** to @0.4.**14**. Python 3 bindings were introduced and split apart from the Python2 bindings in 0.4.14.

### DIFF
--- a/repos/spack_repo/builtin/packages/libproxy/package.py
+++ b/repos/spack_repo/builtin/packages/libproxy/package.py
@@ -2,12 +2,14 @@
 #
 # SPDX-License-Identifier: (Apache-2.0 OR MIT)
 
+from spack_repo.builtin.build_systems import cmake, meson
 from spack_repo.builtin.build_systems.cmake import CMakePackage
+from spack_repo.builtin.build_systems.meson import MesonPackage
 
 from spack.package import *
 
 
-class Libproxy(CMakePackage):
+class Libproxy(CMakePackage, MesonPackage):
     """libproxy is a library that provides automatic proxy configuration
     management."""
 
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/libproxy/package.py
```diff
@@ -16,23 +18,53 @@ class Libproxy(CMakePackage):
...
+    version("0.4.18", sha256="0b4a9218d88f6cf9fa25996a3f38329a11f688a9d026141d9d0e966d8fa63837")
     version("0.4.17", sha256="88c624711412665515e2800a7e564aabb5b3ee781b9820eca9168035b0de60a9")
     version("0.4.16", sha256="9e7959d6ae1d6c817f0ac1e253105ce8d99f55d7821c1b6eaef32bf6879c6f0a")
     version("0.4.15", sha256="18f58b0a0043b6881774187427ead158d310127fc46a1c668ad6d207fb28b4e0")
     version("0.4.14", sha256="6220a6cab837a8996116a0568324cadfd09a07ec16b930d2a330e16d5c2e1eb6")
     version("0.4.13", sha256="d610bc0ef81a18ba418d759c5f4f87bf7102229a9153fb397d7d490987330ffd")
 
     variant("perl", default=False, description="Enable Perl bindings")
-    variant("python", default=False, description="Enable Python bindings", when="@0.4.18:")
+    variant("python", default=False, description="Enable Python bindings", when="@0.4.14:0.4.18")
```
becker33:
```suggestion
    variant("python", default=False, description="Enable Python bindings", when="@0.4.14:0.4")
```
This is "any 0.4 version after 0.4.14".

Is the python support non-conditional or removed in version 5+? Your comment about it being separate python2/3 support doesn't clarify why it's removed from the future versions. If it still has python bindings in version 5, we need to modify the `extends` directive for them.

becker33 (changes requested):

emwjacobson:
Copying my response as a comment, as it might be hard to find in the now-outdated review -

`@0.4.14:0.4` is the right call here, thanks.

As for python support, it appears that starting with 0.5 onwards, Python support is available through the use of [PyGObject](https://pygobject.gnome.org/index.html):
https://libproxy.github.io/libproxy/python.html

becker33 (approved):

PR #1654, opened 2025-09-22
emacs: add +sqlite variant

lpoirel:
The most accurate description of emacs behavior I found is [this reddit thread](https://www.reddit.com/r/emacs/comments/123klvf/emacs29_build_from_git_with_sqlite3/).
- by default (no options given), emacs detects if sqlite3 is available
  - if present, emacs is built with sqlite3 support
  - if not present, emacs is built without sqlite3 support
- there is an option to bypass the detection and build without sqlite3 support, even if sqlite3 is present
- there is no option to force sqlite3 support

### DIFF
--- a/repos/spack_repo/builtin/packages/emacs/package.py
+++ b/repos/spack_repo/builtin/packages/emacs/package.py
@@ -63,6 +63,7 @@ class Emacs(AutotoolsPackage, GNUMirrorPackage):
     )
     variant("json", default=False, when="@27:", description="Build with json support")
     variant("native", default=False, when="@28:", description="Enable native compilation of elisp")
+    variant("sqlite", default=False, when="@29.1:", description="Build with sqlite3 support")
     variant("tls", default=True, description="Build with gnutls support")
     variant("treesitter", default=False, when="@29:", description="Build with tree-sitter support")
 
@@ -90,6 +91,7 @@ class Emacs(AutotoolsPackage, GNUMirrorPackage):
     depends_on("tree-sitter", when="+treesitter")
     depends_on("gcc@11: +strip languages=jit", when="+native")
     depends_on("jansson@2.7:", when="+json")
+    depends_on("sqlite@3", when="+sqlite")
 
     # GUI dependencies
     with when("gui=x11"):


repos/spack_repo/builtin/packages/emacs/package.py
```diff
@@ -127,6 +129,9 @@ def configure_args(self):
         args.extend(self.with_or_without("tree-sitter", variant="treesitter"))
         args.extend(self.with_or_without("json"))
 
+        if self.spec.satisfies("~sqlite"):
+            args.append("--with-sqlite3=no")
+
```
alecbcs:
```suggestion
        args.extend(self.with_or_without("sqlite3")
```

We should be able to continue using `with_or_without()` here.

repos/spack_repo/builtin/packages/emacs/package.py
```diff
@@ -63,6 +63,7 @@ class Emacs(AutotoolsPackage, GNUMirrorPackage):
     )
     variant("json", default=False, when="@27:", description="Build with json support")
     variant("native", default=False, when="@28:", description="Enable native compilation of elisp")
+    variant("sqlite", default=False, when="@29:", description="Build with sqlite3 support")
```
alecbcs:
```suggestion
    variant("sqlite", default=False, when="@29.1:", description="Build with sqlite3 support")
```

It seems like `sqlite-mode` is only available in v29.1 and newer.

alecbcs (changes requested):
Two quick suggestions. Otherwise this looks good to me.

repos/spack_repo/builtin/packages/emacs/package.py
lpoirel: I do not think so:
image
reddit user in [this discussion](https://www.reddit.com/r/emacs/comments/123klvf/emacs29_build_from_git_with_sqlite3/) says that `--with-sqlite3=no` exists, but `with-sqlite3=yes` does not exist.

The build options are:
- nothing -> build with sqlite support if sqlite3 available
- `--with-sqlite3=no` -> build without sqlite support

alecbcs: In this case I think we could just drop the `--with-sqlite3=no` and have the presence / absence of the dependency controlled by the variant do the logic for us.

lpoirel: done!
I removed the commit that added `--with-sqlite3=no`, squashed the correction `@29:` -> `@29.1:`, and rebased on develop

alecbcs (approved):
Looks good to me. Thanks @lpoirel!

PR #1656, opened 2025-09-23
plog: add missing C build dependency

plexoos:
Fixes build with Spack 1.0.1


alecbcs (approved):
Confirmed this fixes a build error when the C compiler is not defined. 

Looks good to me. Thanks @plexoos!

PR #1667, opened 2025-09-23
Fixing gdbm for gcc@15

davide-q:
# FIXME: Merge is blocked due to requested changes:

@davide-q, see this commit how it should be done instead:
https://github.com/spack/spack-packages/pull/1470/changes

### DIFF
--- a/repos/spack_repo/builtin/packages/gdbm/package.py
+++ b/repos/spack_repo/builtin/packages/gdbm/package.py
@@ -68,4 +68,8 @@ class Gdbm(AutotoolsPackage, GNUMirrorPackage):
         #   https://patchwork.ozlabs.org/patch/771300/
         #   https://stackoverflow.com/questions/5582211
         #   https://www.gnu.org/software/automake/manual/html_node/Flag-Variables-Ordering.html
-        return ["--enable-libgdbm-compat", "CPPFLAGS=-D_GNU_SOURCE"]
+        opts = ["--enable-libgdbm-compat", "CPPFLAGS=-D_GNU_SOURCE"]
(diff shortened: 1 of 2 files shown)


repos/spack_repo/builtin/packages/gdbm/package.py
```diff
@@ -68,4 +68,8 @@ def configure_args(self):
         #   https://patchwork.ozlabs.org/patch/771300/
         #   https://stackoverflow.com/questions/5582211
         #   https://www.gnu.org/software/automake/manual/html_node/Flag-Variables-Ordering.html
-        return ["--enable-libgdbm-compat", "CPPFLAGS=-D_GNU_SOURCE"]
+        opts = ["--enable-libgdbm-compat", "CPPFLAGS=-D_GNU_SOURCE"]
+        if self.spec.satisfies("%gcc@15:"):
+            opts.append("CFLAGS=-std=gnu17")
+
+        return opts
```
michaelkuhn, line 75: I think we should rather be doing this with a `flag_handler`. This also doesn't seem to be necessary for all versions since recent versions build fine for me with GCC 15.

michaelkuhn (changes requested):

repos/spack_repo/builtin/packages/gdbm/package.py
bernhardkaindl: @davide-q See this diff how it should be done:
https://github.com/spack/spack-packages/pull/1470/changes

bernhardkaindl:
@davide-q, do you plan to work on this PR?

The comment says the changes / the PR should be updated to use `flag_handler()`.

PR #1668, opened 2025-09-23
opendatadetector: find libs and use it for LD_LIBRARY_PATH

jmcarcell:
instead of adding both lib and lib64. Tested successfully on Alma 9.


tldahlgren (approved):
LGTM.

@vvolkl Any objections to this PR being merged as is?

alecbcs:
Merged without a response from @vvolkl after > 1 week.

PR #1671, opened 2025-09-23
various py-versioneer fixes

Chrismarsh:
Various py-versioneer fixes:
- `py-dask-mpi`  vendors py-versioneer and needs a patch for `python@3.12:`
- `py-distributed` has incorrect overlapping upper bounds resulting in concretization error
- `py-donfig` has `py-versioneer` incorrectly as a build, run dep


tldahlgren (dismissed):
@adamjstewart Do you have any objections to this PR being merged since you made significant changes recently to `py-distributed`?

repos/spack_repo/builtin/packages/py_donfig/package.py
```diff
@@ -22,5 +22,7 @@ class PyDonfig(PythonPackage):
     )
 
     depends_on("py-setuptools@62.6:", type="build")
-    depends_on("py-versioneer@0.28: +toml")
+
+    # strictly should be 0.28 but 0.29 is a safe version
+    depends_on("py-versioneer@0.28:0.29 +toml", type="build")
```
adamjstewart: Not sure if an upper bound is really needed here. A lot of projects like numpy/scipy/matplotlib will pin build deps because they have no control over them via pip, when in fact all newer versions work fine. Until a newer version of py-versioneer is released that breaks things, I think we can leave off the upper bound.

adamjstewart (dismissed):
Seems fine. We could probably even relax these more if we want to.

repos/spack_repo/builtin/packages/py_donfig/package.py
Chrismarsh: Sure

Chrismarsh:
I've relaxed both as you suggested

Chrismarsh:
I think along with https://github.com/spack/spack/pull/51351 this should bite much less frequently now

repos/spack_repo/builtin/packages/py_dask_mpi/package.py
```diff
@@ -32,3 +32,9 @@ class PyDaskMpi(PythonPackage):
     # jupyter-server-proxy is not a needed dependency; https://github.com/dask/dask-mpi/pull/102
     # this significantly reduces the dependency tree of py-dask-mpi
     patch("remove-dependency-jupyter-proxy.patch", when="@:2022.4.0")
+
+    # fix for python 3.12+, merged into mainline
+    # https://github.com/dask/dask-mpi/pull/123
+    # this PR-as-a-patch doesn't apply onto top of @2022, so this patch is rebased ontop of
+    # 2022.4.0 tag release
+    patch("py312-versioneer.patch", when="@2022 ^python@3.12:")
```
adamjstewart: Don't really want to add a 3K line patch. Can we instead add a newer version (no releases, only main?) and add:
```python
conflicts("^python@3.12:", when="@:2022")
```

Chrismarsh: `@2022` is the newest release. Given they vendor `py-versioneer` it's the only way to make it compatible with python 3.12. Are you proposing adding a new release that is just the git mainline?

adamjstewart: Either a pinned dated commit or just a main branch, yes.

adamjstewart: Alternatively, deprecate the whole package if it's abandoned.

Chrismarsh: It's a slow moving project, but it's definitely not abandoned. I've asked if they would consider a new release
https://github.com/dask/dask-mpi/issues/136

Chrismarsh: Ok let me know if you like what I've done to impliment what you've suggested

Chrismarsh:
@spackbot fix style

repos/spack_repo/builtin/packages/py_dask_mpi/package.py
```diff
@@ -17,6 +17,16 @@ class PyDaskMpi(PythonPackage):
...
+    # This is not an official release
+    # There have been no releases post-2022, c.f.,
+    # https://github.com/dask/dask-mpi/issues/136
+    # but various fixes have been implimented so this is
+    # the most recent main commit as of 2025-09-26
+    version(
+        "2025.0.0",
+        sha256="39ee0564af77be6eb34e52581dafbb1335ba82d0fccd2d7fa6182e530251a19d",
+        url="https://www.github.com/dask/dask-mpi/tarball/3d4742e7fca906e265234fb48360990eafe51950",
+    )
```
adamjstewart: I would instead add:
```python
git = "https://github.com/dask/dask-mpi.git"
```
to the top and change this to:
```suggestion
    version("2025.4.1", commit="3d4742e7fca906e265234fb48360990eafe51950")
```
This is the date and hash of the most recent commit.

I would also be fine with adding:
```python
```
as well if that makes things easier.

repos/spack_repo/builtin/packages/py_dask_mpi/package.py
```diff
@@ -32,3 +42,17 @@ class PyDaskMpi(PythonPackage):
     # jupyter-server-proxy is not a needed dependency; https://github.com/dask/dask-mpi/pull/102
     # this significantly reduces the dependency tree of py-dask-mpi
     patch("remove-dependency-jupyter-proxy.patch", when="@:2022.4.0")
+
+    # fix for python 3.12+, merged into mainline
+    # https://github.com/dask/dask-mpi/pull/123
+    conflicts("^python@3.12:", when="@:2022")
+    requires("@2025:", when="^python@3.12:")
```
adamjstewart: This is implied by the previous line, don't need both

repos/spack_repo/builtin/packages/py_dask_mpi/package.py
```diff
@@ -32,3 +42,17 @@ class PyDaskMpi(PythonPackage):
     # jupyter-server-proxy is not a needed dependency; https://github.com/dask/dask-mpi/pull/102
     # this significantly reduces the dependency tree of py-dask-mpi
     patch("remove-dependency-jupyter-proxy.patch", when="@:2022.4.0")
+
+    # fix for python 3.12+, merged into mainline
+    # https://github.com/dask/dask-mpi/pull/123
+    conflicts("^python@3.12:", when="@:2022")
```
adamjstewart:
```suggestion
    conflicts("^python@3.12:", when="@:2024.4.28")
```
If you want to be more specific

repos/spack_repo/builtin/packages/py_dask_mpi/package.py
```diff
@@ -32,3 +42,17 @@ class PyDaskMpi(PythonPackage):
...
+    requires("@2025:", when="^python@3.12:")
+
+    # https://github.com/dask/dask-mpi/issues/137
+    @when("@2025")
+    def patch(self):
+        filter_file(
+            """    python_requires=">=3.6,<3.12",""",
+            """    python_requires=">=3.6",""",
```
adamjstewart:
```suggestion
            '    python_requires=">=3.6,<3.12",',
            '    python_requires=">=3.6",',
```

repos/spack_repo/builtin/packages/py_dask_mpi/package.py
Chrismarsh: I doubt a 2024 will be released at this point, so I'm happy with the bound as is as it's clearly matching what is in the package.py

repos/spack_repo/builtin/packages/py_dask_mpi/package.py
Chrismarsh: Oh ok, I like this more thanks

repos/spack_repo/builtin/packages/py_dask_mpi/package.py
```diff
@@ -32,3 +39,16 @@ class PyDaskMpi(PythonPackage):
     # jupyter-server-proxy is not a needed dependency; https://github.com/dask/dask-mpi/pull/102
     # this significantly reduces the dependency tree of py-dask-mpi
     patch("remove-dependency-jupyter-proxy.patch", when="@:2022.4.0")
+
+    # fix for python 3.12+, merged into mainline
+    # https://github.com/dask/dask-mpi/pull/123
+    conflicts("^python@3.12:", when="@:2022")
```
adamjstewart:
```suggestion
    depends_on("python@:3.11", when="@:2022")
```
This will make it easier for `spack audit packages` to notice when Python 3.11 is removed and the corresponding versions of py-dask-mpi need to be removed.

Chrismarsh:
@spackbot fix style

adamjstewart (approved):

PR #1672, opened 2025-09-24
log4cplus: add missing dependency on c and update url to github

csbnw:
The `log4cplus` package fails to build due to a missing dependency on `c`.

The source has moved from SourceForge to GitHub, the package has been updated accordingly. This also includes updating the versions to the latest patch versions of `1.2`, `2.0` and `2.1`.


alalazo (commented):
It seems the project moved to GitHub. Can you update the `url` etc. ?

csbnw:
I just pushed the following changes:

- The `homepage` and `url` are updated to github. 
- The versions are updated to contain the latest patch version for `1.2` and `2.0` that were already present, and `2.1` is added too.
 
There are no longer two patch versions of `2.0`, since I assume that no one really needs the old `2.0.1` anyway.

alalazo (approved):

PR #1674, opened 2025-09-24
madgraph5amc: Add version 3.5.9 + fix ninja path

paulgessinger:
This PR adds the 3.5.9 version of madgraph and fixes the discovery path for ninja: this is currently setting the base install dir, but it needs to be the `lib` dir for madgraph to correctly locate it.


alalazo (approved):

jmcarcell:
This version is not one of the LTS ones and what has been happening in the past is that they disappear at some point and then the recipe is broken and needs to be fixed. I would prefer if it was removed, 3.5.7 is in the LTS list for 3.5.x.

paulgessinger:
3.5.9 is in lcg and I was trying to reproduce that.

jmcarcell:
Sources for LCG are uploaded to EOS where they don't disappear. Some packages use these sources here on Spack, I guess that's also OK :shrug:

PR #1675, opened 2025-09-24
syscalc: Use CMake mirror

paulgessinger:
I don't know if a change like this controversial, but here we go:

- I've discovered that the syscalc package points at a source url that does not exist anymore (https://bazaar.launchpad.net/~mgtools/mg5amcnlo/SysCalc/tarball/17). https://cp3.irmp.ucl.ac.be/projects/madgraph/wiki/SysCalc still exists, but does not contain any relevant information. The only reason spack can still download the sources is through the mirror. **I did not manage to figure out if this was moved anywhere else**.
- The syscalc package is **tiny** and uses a custom brittle Makefile build, that uses a C++ and fortran compiler, and doesn't portably link to the fortran runtime library (it fails to build on macOS). 
- I've created a sort of "fork" of the package here (https://github.com/paulgessinger/SysCalc) where I removed the Makefiles and replace it with a straight forward CMake build
- This PR updates the package in spack to point at that repository instead and switches the package type to a CMake one
- I'm also removing the vendored version of tinyxml2 from syscalc and instead let it use a version installed from spack.

Concerning versions: since I've only had access to the source code through the mirror, I only have the 1.1.7 version of the code. However, the spack package was not referencing any other versions anyway. I don't know if or where this package is developed, so I don't know if there are updated versions.

As it stands, the package configuration as is is essentially broken, hence my suggetion.

### DIFF
--- a/repos/spack_repo/builtin/packages/syscalc/package.py
+++ b/repos/spack_repo/builtin/packages/syscalc/package.py
@@ -20,6 +20,7 @@ class Syscalc(CMakePackage):
 
     depends_on("cxx", type="build")  # generated
     depends_on("fortran", type="build")  # generated
+    depends_on("pkgconfig", type="build")
 
     tags = ["hep"]
 
--- a/repos/spack_repo/builtin/packages/syscalc/package.py
+++ b/repos/spack_repo/builtin/packages/syscalc/package.py
@@ -2,23 +2,21 @@
 #
 # SPDX-License-Identifier: (Apache-2.0 OR MIT)
 
-from spack_repo.builtin.build_systems.makefile import MakefilePackage
+from spack_repo.builtin.build_systems.cmake import CMakePackage
 
 from spack.package import *
 
 
-class Syscalc(MakefilePackage):
+class Syscalc(CMakePackage):
     """A tool to derive theoretical systematic uncertainties"""
 
     homepage = "https://cp3.irmp.ucl.ac.be/projects/madgraph/wiki/SysCalc"
     url = "https://bazaar.launchpad.net/~mgtools/mg5amcnlo/SysCalc/tarball/17"
 
-    version(
-        "1.1.7",
-        sha256="ac73df0f9f195eb62601fafc2eede3db17a562750f7971616870d6df4abd1b6c",
-        url="https://bazaar.launchpad.net/~mgtools/mg5amcnlo/SysCalc/tarball/17",
-        extension=".tgz",
-    )
+    # This is a mirror + added CMakeLists.txt file
+    git = "https://github.com/paulgessinger/SysCalc.git"
+
+    version("1.1.7", commit="d62edf92b04a6cf89cde6837b0a999bc79601e8f")
 
     depends_on("cxx", type="build")  # generated
     depends_on("fortran", type="build")  # generated
@@ -26,6 +24,7 @@ class Syscalc(MakefilePackage):
     tags = ["hep"]
 
     depends_on("lhapdf@6:")
+    depends_on("tinyxml2")
 
     def url_for_version(self, version):
         url = self.url.rsplit("/", 1)[0]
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/syscalc/package.py
```diff
@@ -2,44 +2,34 @@
...
     url = "https://bazaar.launchpad.net/~mgtools/mg5amcnlo/SysCalc/tarball/17"
 
-    version(
-        "1.1.7",
-        sha256="ac73df0f9f195eb62601fafc2eede3db17a562750f7971616870d6df4abd1b6c",
-        url="https://bazaar.launchpad.net/~mgtools/mg5amcnlo/SysCalc/tarball/17",
-        extension=".tgz",
-    )
+    # This is a mirror + added CMakeLists.txt file
+    git = "https://github.com/paulgessinger/SysCalc.git"
```
wdconinc, line 17: In these cases of upstream issues which are resolved in a fork, we try to keep the option open that upstream gets their act together by adding the fork without removing the upstream version. So, you'd keep the 1.7 in place, deprecate it, add a 1.7.paulgessinger.0 which points to the fork, likely by adding a git attribute to the version directive instead of package wide. That would them pick your forked version preferentially but not invalidate systems that have the upstream version installed (somehow).

The changing build system makes that a bit trickier, since you'd likely have to keep adding the forked versions to the conditional that says they should be built with CMake. An alternative (not sure it's worth it in this case) is to use a variant to pick the fork, and then indicate to use CMake when that fork variant is used.

paulgessinger: In that case, I wouldn't be able to use `CMakePackage` though, would I?

I think they haven't made a new version in about a decade.

This seems like a lot of trouble, so I would lean towards keeping a private fork of the package and keeping `builtin.syscalc` broken as is.

wdconinc: Can we ask https://github.com/mg5amcnlo to host the fork? Then I think there are lot less reasons to be concerned since it wouldn't be a fork as much as an archival repository owned by the original authors.

paulgessinger: I don't know who to contact about this, do you @wdconinc?

wdconinc: Tagging @oliviermattelaer.

Olivier, it seems in the past SysCalc was distributed via https://bazaar.launchpad.net/~mgtools/mg5amcnlo/SysCalc, but not anymore. @paulgessinger has created a 'fork' at https://github.com/paulgessinger/SysCalc (uses CMake, not Makefile). Would that be something that you would be willing to host under https://github.com/mg5amcnlo for the benefit of the community since it is apparently used in the context of mg5amc?

oliviermattelaer:
I'm actually surprised that anyone is still using that fork of SysCalc.
That code has been fully re-written (in python) 10 years ago and that new code is now fully integrated within mg5amc and shipped with all version of mg5amc (with full backward compatibility with SysCalc).

That C++ code was never fully functioning, neither really supported (because his main authors left physics before the completion of that project). So my suggestion would be to simply remove any reference to SysCalc within spack. (and I can help @paulgessinger and any other user to move to our internal module).

Now, I do not see any objection to host inside our github part (but I will likely archive it directly).
But if @paulgessinger wants to revive SysCalc and/or keep it alive, it does make more sense to use that repo as the source of truth.

wdconinc:
@paulgessinger what say you?

wdconinc:
@paulgessinger what say you?

paulgessinger:
@wdconinc @oliviermattelaer I don't have a specific need for this if there's an alternative. Madgraph's configuration for sure are still generated with a key that points at the syscalc executable, but I'm not sure if it's actually used anywhere. 

I'll try removing it from the madgraph package build and see if that works.

paulgessinger:
Seems to work, I'm closing this and will update the `magraph5amc` package to drop it

PR #1676, opened 2025-09-24
mochi-margo: new versions, variants, and build systems

mdorier:
This PR does the following:
- Converts the mochi-margo package into a dual Autotools/CMake package, as mochi-margo@main and versions > 0.21.0 now use cmake;
- Add the `+hwlock` variant;
- Add versions 0.20.0 and 0.21.0

### DIFF
--- a/repos/spack_repo/builtin/packages/mochi_margo/package.py
+++ b/repos/spack_repo/builtin/packages/mochi_margo/package.py
@@ -18,6 +18,12 @@ class MochiMargo(cmake.CMakePackage, autotools.AutotoolsPackage):
 
     maintainers("carns", "mdorier", "fbudin69500")
 
+    build_system(
+        conditional("cmake", when="@0.22.0:"),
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/mochi_margo/package.py
```diff
@@ -18,6 +19,8 @@ class MochiMargo(AutotoolsPackage):
     maintainers("carns", "mdorier", "fbudin69500")
 
     version("main", branch="main")
```
wdconinc, line 27:
```suggestion
    build_system("cmake", "autotools", default="CMake")

    version("main", branch="main")
```
with `conditional` as appropriate.

mdorier: Thanks, I added. Now all tests are passing.

wdconinc (approved):
Looks good to me.

PR #1680, opened 2025-09-24
duckdb: convert to cmake package

haampie:
The `Makefile` is just a wrapper to invoke `cmake`.

With `CMakePackage` we benefit from Spack's recommended/default CMake arguments.

* Define `httpfs` conditionally, since it's also out-of-tree since v1.2.0
* Ensure that Python extension is installed in duckdb's prefix (previously it installed in `python-venv`'s prefix).
* Remove `cli`, `excel`, `inet` variants as these extensions no longer exist or are out-of-tree in all of the versions supported by Spack
* Remove redundant `ninjabuild` variant, that's part of `CMakePackage`.
* Remove `static_openssl` variant, which only applies with `httpfs`, and is imo a non-feature because of CVEs in openssl.
* Remove redundant patch

### DIFF
--- a/repos/spack_repo/builtin/packages/duckdb/package.py
+++ b/repos/spack_repo/builtin/packages/duckdb/package.py
@@ -3,12 +3,12 @@
 # SPDX-License-Identifier: (Apache-2.0 OR MIT)
 
 
-from spack_repo.builtin.build_systems.makefile import MakefilePackage
+from spack_repo.builtin.build_systems.cmake import CMakePackage
 
 from spack.package import *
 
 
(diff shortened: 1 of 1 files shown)


alecbcs (approved):
Looks good to me. Thanks @haampie 

Will give @glentner @teaguesterling a couple days for a final review before merging.

teaguesterling (approved):
LGTM. Happy to ser the simplification! We should add v1.4.0 next.

glentner:
Apologies, was going to get to testing this this afternoon but hadn't gotten to it yet.

PR #1681, opened 2025-09-24
gdal: make `+arrow` default to improve build robustness

Chrismarsh:
GDAL's cmake searches for packages even when disabled: https://github.com/spack/spack/issues/42809#issuecomment-1960352436; https://github.com/OSGeo/gdal/issues/9287. This results in bizarre errors that are difficult to track down that only show up when building in certain combinations.

This protects against one such case:
https://github.com/OSGeo/gdal/issues/12327

@adamjstewart, this was sufficient for me with gdal 3.10, does it resolve the 3.11 issue for you?


Chrismarsh:
@spackbot fix style

adamjstewart:
This does fix the issue for me, but I don't really love the solution...

Chrismarsh:
Neither do I. But I like gdal building.

Unless https://github.com/OSGeo/gdal/issues/9287 is fixed, is there an alternative?

adamjstewart:
I guess the alternative would be to build +arrow by default. That's what I'm using locally.

Chrismarsh:
but `~arrow` will still be broken. Almost feels like `+arrow` is required and non-optional? Given it has to build to exist for build depend which is then never used.

adamjstewart:
My experience was that `+arrow` is only required if CMake finds a partial install on the system. So if I have `arrow~filesystem` present on my system, then the build will fail, but if I don't, then the build will succeed.

Chrismarsh:
Mmmh ya. Complicated indeed.

```
spack install gdal@3.10+arrow ^arrow~filesystem
==> Error: failed to concretize `gdal@3.10+arrow ^arrow~filesystem` for the following reasons:
     1. 'arrow' requires conflicting variant values '~filesystem' and '+filesystem'
     2. 'arrow' requires conflicting variant values '~filesystem' and '+filesystem'
        required because gdal depends on arrow+filesystem when +arrow
          required because gdal@3.10+arrow ^arrow~filesystem requested explicitly
        required because gdal@3.10+arrow ^arrow~filesystem requested explicitly
```

so while not amazing, I think this PR then correctly works around those edge cases to ensure a robust build until GDAL changes how it's cmake works.

adamjstewart:
You are correct that this PR results in a more robust build. I just wonder how hard it would be to fix GDAL. I unfortunately don't know anything about CMake. Not sure what to do here. I'm still personally in favor of making these variants default to True. Most people won't bother to disable things, only enable them.

Chrismarsh:
Given Evan hasn't done it, I assume it's hard enough that it isn't going to happen any time soon. 

If `+arrow` is default True, then `~arrow` is still uncertain if it fails. Could we make `+arrow` true, and remove the variant from `3.10:`?

adamjstewart:
> Given Evan hasn't done it, I assume it's hard enough that it isn't going to happen any time soon.

Or he's insanely busy.

> If `+arrow` is default True, then `~arrow` is still uncertain if it fails.

Yes, but how many people are going to disable `~arrow` and then be confused when it fails? It works sometimes, fails other times. I'm hesitant to remove the ability to disable it, but if you really want we could add a conflict for `~arrow`. That's what we did for `~png/~jpeg` in older versions.

Chrismarsh:
Ok I've made `+arrow` default and added a comment that links to your issue with a small desc. I've removed the build only complexity this PR originally added.

I am also adding the error message copy-pasted from the gdal ticket for future searching 

```
-- Performing Test HAS_NO_UNDEFINED - Failed
CMake Error at port/CMakeLists.txt:3 (include):
  include could not find requested file:

    TargetPublicHeader


CMake Error at port/CMakeLists.txt:4 (include):
  include could not find requested file:

    GdalDriverHelper


CMake Error at port/CMakeLists.txt:6 (target_public_header):
  Unknown CMake command "target_public_header".


```

adamjstewart (approved):

PR #1687, opened 2025-09-24
fixup to build additivefoam utilities

colemanjs:
Fix-up on the AdditiveFOAM build script to build utilities in applications/utilities folder.


gknapp1 (approved):
This looks good to me. 

I doublechecked that this installs the utilities within the expected spack installation path. It does, so the utilities associated with the spack additivefoam install are loaded/unloaded as expected.

tldahlgren (approved):

PR #1689, opened 2025-09-24
fftw: Fix NEON usage on armv8+

green-br:
It looks like NEON was not being enabled on armv8+ due to "neon" CPU feature moving to "asimd".  If "asimd" is available we can enable NEON for float and double precision.

Tested as part of a wider test of the SVE patch in https://github.com/FFTW/fftw3/pull/315 where there is still a wait on 3.3.11 version of FFTW which should include this.  Without NEON enabled the performance improvement with the SVE patch was not as expected which led to discovering this issue.

Improved timer with `--enable-armv8-cntvct-el0` could also be used on armv8.

Example output from tests in fftw

Current version
```
$ ./bench -v2 -r 20 -s ocf2048
planner time: 0.001217 s
(dft-ct-dit/32
  (dftw-direct-32/62 "t1_32")
  (dft-direct-64-x32 "n1_64"))
flops: 47616 add, 11904 mul, 9344 fma
estimated cost: 78208.000000, pcost = 78208.000000
Problem: ocf2048, setup: 1.25 ms, time: 7.56 us, ``mflops'': 14897.431
Took 20 measurements for at least 10.00 ms each.
Time: min 7.56 us, max 7.62 us, avg 7.57 us, median 7.57 us
```

With fix (using Nvidia Grace CPU Superchip - neoverse-v2)
```
$ ./bench -v2 -r 20 -s ocf2048
planner time: 0.002544 s
(dft-ct-dit/32
  (dftw-direct-32/124 "t2fv_32_neon")
  (dft-direct-64-x32 "n2fv_64_neon"))
flops: 25792 add, 7936 mul, 2688 fma
estimated cost: 39104.000000, pcost = 39104.000000
Problem: ocf2048, setup: 2.57 ms, time: 5.66 us, ``mflops'': 19890.216
Took 20 measurements for at least 10.00 ms each.
Time: min 5.66 us, max 5.67 us, avg 5.67 us, median 5.67 us
```

### DIFF
--- a/repos/spack_repo/builtin/packages/fftw/package.py
+++ b/repos/spack_repo/builtin/packages/fftw/package.py
@@ -80,13 +80,13 @@ class FftwBase(AutotoolsPackage):
         # to check NEON.
         if self.spec.satisfies("target=m1:"):
             filter_file(
-                r" aarch64)",
-                r" aarch64 | arm)",
+                r" aarch64\)",
+                r" aarch64 | arm\)",
                 "configure",
             )
             filter_file(
-                r" aarch64)",
-                r" aarch64 | arm)",
+                r" aarch64\)",
+                r" aarch64 | arm\)",
                 "configure.ac",
             )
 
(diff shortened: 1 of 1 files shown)


green-br:
@spackbot fix style

psakievich (approved):
The relationship between `asimd` and `neon` is confusing as implemented, but I don't clearly see a way to improve it. It seems we're at the mercy of the build-system's confusing implementation.

PR #1691, opened 2025-09-25
pkgconfig: virtual has no version

haampie:
Remove version bounds on the `pkgconfig` virtual, because no versions
are defined for the "api" by `pkg-config` and `pkgconf`.

If these were meant to be `pkg-config` lowerbounds, they're redundant,
cause all we have is 0.28 and higher.


alalazo (approved):

alalazo:
Failure is unrelated:
```
$ git fetch --depth 1 origin ${SPACK_CHECKOUT_VERSION}
fatal: unable to access 'https://github.com/spack/spack.git/': Failed to connect to github.com port 443 after 132646 ms: Couldn't connect to server
```

adamjstewart:
These were definitely meant to be pkg-config versions. pkgconf has a different versioning scheme. Is there any known mapping between versions?

PR #1692, opened 2025-09-25
scotch: add determinism variant

klevzoff:
From Scotch CMake build guide:

```
  - SCOTCH_DETERMINISTIC:STRING (default value FIXED_SEED): set this
    flag to NONE to use faster, non-deterministic algorithms in
    Scotch, and a variable random seed; FIXED_SEED to use these
    algorithms with a fixed random seed; FULL, to use fully
    deterministic algorithms with a fixed random string, enabling
    fully deterministic behavior with any number of threads and the
    same number of MPI processes. These three levels are obtained by a
    combination of the Make compilation flags
    "-DCOMMON_RANDOM_FIXED_SEED" and "-DSCOTCH_DETERMINISTIC"; see
    sections 3.9) and 3.10) of this document.

```

Scotch CMake defaults to `FIXED_SEED`, but we often need to build with full determinism. Hence, makes sense to expose this option as a Spack variant.

### DIFF
--- a/repos/spack_repo/builtin/packages/scotch/package.py
+++ b/repos/spack_repo/builtin/packages/scotch/package.py
@@ -63,6 +63,13 @@ class Scotch(CMakePackage, MakefilePackage):
         when="@7.0.1",
         description="Link error handling library to libscotch/libptscotch",
     )
+    variant(
+        "determinism",
+        default="FULL",
+        values=("NONE", "FIXED_SEED", "FULL"),
+        multi=False,
+        description="Determinism configuration",
+    )
 
     depends_on("c", type="build")
     depends_on("cxx", type="build")
@@ -151,6 +158,9 @@ class CMakeBuilder(cmake.CMakeBuilder):
         if self.pkg.version > Version("7.0.4"):
             args.append(self.define("ENABLE_TESTS", self.pkg.run_tests))
 
+        if self.pkg.version >= Version("7.0.7"):
+            args.append(self.define_from_variant("SCOTCH_DETERMINISTIC", "determinism"))
+
         if "+int64" in self.spec:
             args.append("-DINTSIZE=64")
         elif self.is_64bit():
@@ -174,7 +184,14 @@ class MakefileBuilder(makefile.MakefileBuilder):
 
     def edit(self, pkg, spec, prefix):
         makefile_inc = []
-        cflags = ["-O3", "-DCOMMON_RANDOM_FIXED_SEED", "-DSCOTCH_DETERMINISTIC", "-DSCOTCH_RENAME"]
+        cflags = ["-O3", "-DSCOTCH_RENAME"]
+
+        if "determinism=FIXED_SEED" in self.spec:
+            cflags.append("-DCOMMON_RANDOM_FIXED_SEED")
+
+        if "determinism=FULL" in self.spec:
+            cflags.append("-DCOMMON_RANDOM_FIXED_SEED")
+            cflags.append("-DSCOTCH_DETERMINISTIC")
 
         # SCOTCH_Num typedef: size of integers in arguments
         # SCOTCH_Idx typedef: indices for addressing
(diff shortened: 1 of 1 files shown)


klevzoff:
@climbfuji @AlexanderRichert-NOAA apologies for tagging you directly, but the PR workflow hasn't assigned any reviewers and you are listed as package maintainers for Scotch. Could you please help review the change and drive the approval (e.g. assign reviewers with write access) if it makes sense?

AlexanderRichert-NOAA (approved):
Looks reasonable to me. @klevzoff are you by any chance on the Spack Slack server? If so I'd suggest posting this in the PRs channel.

klevzoff:
Thank you for review and the suggestion @AlexanderRichert-NOAA, I was not aware of the slack server - joined now!

repos/spack_repo/builtin/packages/scotch/package.py
```diff
@@ -63,6 +63,13 @@ class Scotch(CMakePackage, MakefilePackage):
         when="@7.0.1",
         description="Link error handling library to libscotch/libptscotch",
     )
+    variant(
```
climbfuji, line 66: Is this variant only valid for `@7.0.7:` as suggested below? If so, then shouldn't the `variant` should have a `when=` clause?

I also think that version comparisons can be done in a much simpler way, namely `if self.spec.satisfies("@7.0.4:")` etc.

klevzoff: It's complicated: the CMake option `SCOTCH_DETERMINISTIC` was only added in 7.0.7, but in Makefile build `COMMON_RANDOM_FIXED_SEED` and `SCOTCH_DETERMINISTIC` were supported for all versions listed in `package.py`. Therefore I didn't add `when` clause to the variant.

But it's also reasonable to say that anyone building earlier versions were ok with default flags and don't need the variant (whether they use Makefile or CMake build), and it could be restricted only to `@7.0.7:`. I will go ahead and refactor to follow this logic because it's more straightforward for users.

AlexanderRichert-NOAA: Would it overcomplicate things to do something like `when='@7.0.7: build_system=cmake`/`when='build_system=makefile'`? Just tossing that out there.

klevzoff: I implemented the suggestions in https://github.com/spack/spack-packages/pull/1692/commits/c0b42da32edbfca8b87901783e461ce9c6b4256b. Could not find a way to avoid defining the variant twice for the two `when` conditions. On the other hand, this allows them to have different defaults (`FULL` for Makefile, `FIXED_SEED` for CMake) to match current behavior of the package, avoiding surprises for users.

repos/spack_repo/builtin/packages/scotch/package.py
```diff
@@ -63,6 +63,22 @@ class Scotch(CMakePackage, MakefilePackage):
...
+        description="Determinism configuration",
+        when="build_system=makefile",
+    )
+    variant(
+        "determinism",
+        default="FIXED_SEED",
+        values=("NONE", "FIXED_SEED", "FULL"),
+        multi=False,
+        description="Determinism configuration",
+        when="@7.0.7: build_system=cmake",
```
climbfuji, line 80: According to line 44, the cmake build was added for version 7.0.0.

The way the variant `determinism` is defined now means that the variant doesn't exist for versions 7.0.0 through 7.0.6 when using cmake. Is that the intention?

klevzoff: Yes, that is the intention for the CMake based build. CMake option `SCOTCH_DETERMINISM` did not exist until 7.0.7 ([added here](https://gitlab.inria.fr/scotch/scotch/-/commit/6e5eb68327b30dd8cc6469031b7d38cc9b4f506e)) so while we could allow the Spack variant to exist for prior versions, it would have no effect on the build which would be confusing.

climbfuji (approved):
Thanks for the clarification. This now looks all good to me.

becker33 (approved):

PR #1697, opened 2025-09-25
acts: deprecate versions :38

stephenswat:
This commit deprecates older versions of the ACTS package so they can be removed in the future.


repos/spack_repo/builtin/packages/acts/package.py
```diff
@@ -56,155 +56,580 @@ class Acts(CMakePackage, CudaPackage):
...
+        "34.0.0",
+        commit="daafd83adf0ce50f9667f3c9d4791a459e39fd1b",
+        submodules=submodules,
+        deprecated=True,
+    )
+    version(
+        "33.1.0",
+        commit="00591a593a648430820e980b031301d25c18f1c7",
+        submodules=submodules,
+        deprecated=True,
```
alalazo: You can use:
```python
    ...
```
to shorten the diff

stephenswat: Thank you, updated.

alalazo (approved):

PR #1705, opened 2025-09-25
factory: new package

d-torrance:
Factory (which is maintained as part of [Singular](https://www.singular.uni-kl.de/)) is a library for computations with multivariate polynomials.  In addition to Singular, it is also used by Macaulay2.

### DIFF
--- a/repos/spack_repo/builtin/packages/singular_factory/package.py
+++ b/repos/spack_repo/builtin/packages/singular_factory/package.py
@@ -1,0 +1,35 @@
+# Copyright Spack Project Developers. See COPYRIGHT file for details.
+#
+# SPDX-License-Identifier: (Apache-2.0 OR MIT)
+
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/factory/package.py
```diff
@@ -0,0 +1,35 @@
+# Copyright Spack Project Developers. See COPYRIGHT file for details.
+#
+# SPDX-License-Identifier: (Apache-2.0 OR MIT)
+
+from spack_repo.builtin.build_systems.autotools import AutotoolsPackage
+
+from spack.package import *
+
+
+class Factory(AutotoolsPackage):
```
alalazo: Since "factory" is a very common name in computer science, what do you think about renaming this to `singular-factory`?

d-torrance: Seems reasonable -- done!

alalazo (approved):

PR #1708, opened 2025-09-25
json-glib: Add v1.10.8 and v1.9.2

KineticTheory:
I am unable to build existing version 1.6.6, so I added versions 1.9.2 and 1.10.8.  Both of the new versions build cleanly w/o new dependencies.  However, I needed to add new mason build options to manually disable documentation options that would have required new dependencies.

### DIFF
--- a/repos/spack_repo/builtin/packages/json_glib/package.py
+++ b/repos/spack_repo/builtin/packages/json_glib/package.py
@@ -68,6 +68,7 @@ class JsonGlib(MesonPackage):
     def url_for_version(self, version):
         return f"https://download.gnome.org/sources/json-glib/{version.up_to(2)}/json-glib-{version}.tar.xz"
 
+
 class MesonBuilder(meson.MesonBuilder):
     @when("@1.9.2:")
     def meson_args(self):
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/json_glib/package.py
```diff
@@ -61,3 +64,13 @@ def build(self, spec, prefix):
...
         make("install")
+
+    def url_for_version(self, version):
+        return f"https://download.gnome.org/sources/json-glib/{version.up_to(2)}/json-glib-{version}.tar.xz"
+
+
+class MesonBuilder(meson.MesonBuilder):
+    @when("@1.9.2:")
+    def meson_args(self):
+        args = ["-Ddocumentation=disabled", "-Dman=false", "-Dtests=false", "-Dconformance=false"]
```
michaelkuhn: I only needed to disable `documentation` to get the newer versions to build. Any particular reason for disabling man pages and the tests as well?

KineticTheory: > I only needed to disable `documentation` to get the newer versions to build. Any particular reason for disabling man pages and the tests as well?

I can remove those extra configure arguments.  I was just trying to minimize all extra dependencies.

michaelkuhn (approved):
LGTM, thanks!

PR #1710, opened 2025-09-25
tsne-cuda: new package

Markus92:
tsnecuda is a CUDA implementation of the TSNE algorithm. It's a dependency of Relion; it will build/run without this package but eventually give up when you try to use functionality depending on it.

A few patches to CMakeLists.txt are included (all in one .patch file): the first is as because it hardcodes the CUDA compute architectures a little too hard for my taste, not really fitting with Spack philosophy and breaking if you use Cuda 12/13; second is to use Spack's version of cxxopts instead of a submodule from 2018; third is to add an install target.


Markus92:
@spackbot fix-style

Markus92:
@spackbot fix style

Markus92:
Merged as port of #2054

PR #1711, opened 2025-09-25
py-scifem: add v0.14.0 and fix build/link dependencies

jorgensd:
Skipping intermediate releases.

### DIFF
--- a/repos/spack_repo/builtin/packages/py_scifem/package.py
+++ b/repos/spack_repo/builtin/packages/py_scifem/package.py
@@ -22,7 +22,7 @@ class PyScifem(PythonPackage):
     license("MIT", checked_by="jorgensd")
 
     version("main", branch="main")
-    version("0.12.0", sha256="d866d6d87277e8f541c336730d49fdcce9e3576252f855d051f9739b1ff0abb9")
+    version("0.13.1", sha256="8c5570c0a0184a1c44cfca1bf3427bebb0d72181251989ef62ecaeb2fd0ef957")
     version("0.7.0", sha256="016556573a05d2bebb983017c41427d906df870633d23e5bd9d68af8d4b81de2")
     version("0.6.0", sha256="548c9af8997537bc6830c898a9ffe7007dda16b5e40f3240c97e646cfd0a30b3")
     version("0.5.0", sha256="1e5978ab97889c2d6bad80e375c9db1b050bfb68c197eada17928e6908f15372")
--- a/repos/spack_repo/builtin/packages/py_scifem/package.py
+++ b/repos/spack_repo/builtin/packages/py_scifem/package.py
@@ -22,7 +22,7 @@ class PyScifem(PythonPackage):
     license("MIT", checked_by="jorgensd")
 
     version("main", branch="main")
-    version("0.11.1", sha256="369110e55fe51e32d38ff5708d79bd582ffb4034518830baebb2baeaa062784e")
+    version("0.12.0", sha256="d866d6d87277e8f541c336730d49fdcce9e3576252f855d051f9739b1ff0abb9")
     version("0.7.0", sha256="016556573a05d2bebb983017c41427d906df870633d23e5bd9d68af8d4b81de2")
     version("0.6.0", sha256="548c9af8997537bc6830c898a9ffe7007dda16b5e40f3240c97e646cfd0a30b3")
     version("0.5.0", sha256="1e5978ab97889c2d6bad80e375c9db1b050bfb68c197eada17928e6908f15372")
@@ -40,11 +40,15 @@ class PyScifem(PythonPackage):
     depends_on("py-scikit-build-core+pyproject", type="build")
     depends_on("py-setuptools@42:", type="build")
     depends_on("cmake@3.21:", type="build")
+    depends_on("fenics-dolfinx@0.9:", when="@0.4:", type=("build", "link"))
+    depends_on("fenics-dolfinx@main", when="@main", type=("build", "link"))
+    depends_on("fenics-basix@0.9:", when="@0.4:", type=("build", "link"))
+    depends_on("fenics-basix@main", when="@main", type=("build", "link"))
 
     depends_on("py-fenics-dolfinx@0.9:", when="@0.4:", type="run")
     depends_on("py-fenics-dolfinx@main", when="@main", type="run")
-    depends_on("py-numpy", type=("run"))
-    depends_on("py-packaging", type=("run"))
+    depends_on("py-numpy", type="run")
+    depends_on("py-packaging", type="run")
 
     with when("+adios2"):
         depends_on("adios2+python", type="run")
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/py_scifem/package.py
alecbcs, line 40:
```suggestion
    depends_on("py-scikit-build-core@0.4.3:+pyproject", type="build")
```

alecbcs, line 39:
```suggestion
    depends_on("py-nanobind@1.3.2:", type="build")
```

alecbcs, line 42:
```suggestion
    depends_on("cmake@3.19:", type="build")
```

alecbcs:
```suggestion
    depends_on("py-numpy", type=("run"))
```

alecbcs (commented):
It appears our dependencies in Spack are inconsistent with the `pyproject.toml` do you all know if this was intentional or if they can be realigned?

repos/spack_repo/builtin/packages/py_scifem/package.py
jorgensd: This one is intentional.
Scifem supports older versions of  `py-fenics-dolfinx` in general. 
However, on spack, I decided to only support 0.9 and onwards, where the requirement is:
https://github.com/spack/spack-packages/blob/develop/repos/spack_repo/builtin/packages/py_fenics_dolfinx/package.py#L87
Therefore, we cannot change this lower bound.

jorgensd: This one points out an inconsistency in what is required in pyproject.toml, and what is required in the cmake file (which requires 3.21). I'll make a post release of scifem to rectify this, and leave the current bound at 3.21

jorgensd: Same comment as with nanobind. Newer versions of DOLFINx requires a specific version of scikit build core, that is more modern than the lower bound of scifem in general:
https://github.com/spack/spack-packages/blob/develop/repos/spack_repo/builtin/packages/py_fenics_dolfinx/package.py#L89-L90
Should be kept as is and be determined by `py-fenics-dolfinx`.

repos/spack_repo/builtin/packages/py_scifem/package.py
```diff
@@ -22,6 +22,7 @@ class PyScifem(PythonPackage):
     license("MIT", checked_by="jorgensd")
 
     version("main", branch="main")
+    version("0.11.0", sha256="93e99a10439f1b2619ffb8113649c817958132a1d04741359b55912da7d7cb9e")
```
jorgensd:
```suggestion
    version("0.11.1", sha256="369110e55fe51e32d38ff5708d79bd582ffb4034518830baebb2baeaa062784e")
```

jorgensd: Bump to patch release for cmake consistency

jorgensd:
> It appears our dependencies in Spack are inconsistent with the `pyproject.toml` do you all know if this was intentional or if they can be realigned?

@alecbcs I've tried to address all your comments. The one regarding cmake and numpy have been taken into account, while the others remains with existing pins, as `scifem` on Github supports older versions of DOLFINx than what I want to support through spack.

finsberg:
Hi @alecbcs. Among our [dependencies](https://github.com/scientificcomputing/scifem/blob/efc5e51fdcb541c3098c6aa1ab00f087f887b2b6/pyproject.toml#L22) the only dependency that might be inconsistent is `fenics-dolfinx` which is the same as `py-fenics-dolfinx` on spack. However `fenics-dolfinx` is not a package on `pypi` but the name of the packaged that is installed when fenics-dolfinx is installed see https://github.com/FEniCS/dolfinx/blob/9d517ddfbabfd5a5f5fd97712675663272ba30ed/python/pyproject.toml#L12. In other words, `fenics-dolfinx` as to be installed in a different way (e.g from source, conda, docker etc) before `scifem` can be installed. Hope this clarifies the issue.

finsberg:
Hi @alecbcs. I have updated the build dependencies now. This PR should be OK to merge now.

repos/spack_repo/builtin/packages/py_scifem/package.py
```diff
@@ -22,6 +22,7 @@ class PyScifem(PythonPackage):
     license("MIT", checked_by="jorgensd")
 
     version("main", branch="main")
+    version("0.12.0", sha256="d866d6d87277e8f541c336730d49fdcce9e3576252f855d051f9739b1ff0abb9")
```
jorgensd:
```suggestion
    version("0.13.1", sha256="8c5570c0a0184a1c44cfca1bf3427bebb0d72181251989ef62ecaeb2fd0ef957")
```

finsberg:
@alecbcs @alalazo Any update on this? Would be good to get this one merged.

repos/spack_repo/builtin/packages/py_scifem/package.py
```diff
@@ -39,11 +40,15 @@ class PyScifem(PythonPackage):
     depends_on("py-scikit-build-core+pyproject", type="build")
     depends_on("py-setuptools@42:", type="build")
     depends_on("cmake@3.21:", type="build")
+    depends_on("fenics-dolfinx@0.9:", when="@0.4:", type=("build", "link"))
+    depends_on("fenics-dolfinx@main", when="@main", type=("build", "link"))
+    depends_on("fenics-basix@0.9:", when="@0.4:", type=("build", "link"))
+    depends_on("fenics-basix@main", when="@main", type=("build", "link"))
```
alecbcs:
```suggestion
    depends_on("fenics-dolfinx@main", when="@main", type=("build", "link"))
    depends_on("fenics-dolfinx@0.9:", when="@0.4:", type=("build", "link"))
    depends_on("fenics-basix@main", when="@main", type=("build", "link"))
    depends_on("fenics-basix@0.9:", when="@0.4:", type=("build", "link"))
```

Let's put `main` above v0.9 since it is classified as an infinite version.

alecbcs (changes requested):
One quick nitpick, otherwise looks good to me! Thanks @jorgensd!

repos/spack_repo/builtin/packages/py_scifem/package.py
```diff
@@ -22,6 +22,7 @@ class PyScifem(PythonPackage):
     license("MIT", checked_by="jorgensd")
 
     version("main", branch="main")
+    version("0.13.1", sha256="8c5570c0a0184a1c44cfca1bf3427bebb0d72181251989ef62ecaeb2fd0ef957")
```
jorgensd:
```suggestion
    version("0.14.0", sha256="a875311cdd21f1c6395be67c0220df2f03e52a3571041eb75deef0972d511cbb")
```

alecbcs (approved):
Looks good to me. Thanks @jorgensd!

PR #1712, opened 2025-09-25
Netcdf-c && Proj Fixes for CMake based Curl (fixes broken develop CI)

johnwparent:
This PR modifies both packages in one shot as they are both broken in develop and neither PR can pass CI alone while the other has not been merged.

## Netcdf-c: Prevent globbing of arbitrary files for installation on Windows w/ CMake

Netcdf-c globs the CMAKE_PREFIX_PATH (incorrectly pre 4.9.3) to install "runtime" dependencies on Windows
This is dangerous for a multitude of reasons and results in extremely error prone installs on Windows, often bringing in unescaped Windows paths (which CMake cannot handle).

Additionally, Spack already does this quite well thanks and the Netcdf-c CMake operation is redundant and less robust.

Remove the CML code in netcdf-c that affects this installation procedure.

The patches change Windows only CML code (gated behind `if(msvc)`), so while they're applicable on all platforms, they're effectively no-ops everywhere else, so we don't bother applying them.

## Proj: 8:9.1 use documented curl include interface

Proj `@8:9.1` references the curl include dirs via `CURL_INCLUDE_DIR`, which is not a part of the `FindCURL` interface, but is a side effect of the CMake vendored curl find module. 

Both CMake's find curl module and Curl's CMake config support `CURL_INCLUDE_DIRS` (note the plural) so use that instead.

This issue is resolved in upstream proj 9.2 and later.


johnwparent:
@zackgalbreath I didn't realize this hadn't been merged, this should fix the hanging on failure VTK/paraview pipelines (just the root cause, not the hangs)

johnwparent:
@zackgalbreath this is the amalgamated branch that _should_ get us passing CI

zackgalbreath (approved):

adamjstewart (approved):

PR #1713, opened 2025-09-25
MSVC: Retool vcvars environmental setup approach with ARM considerations

johnwparent:
Arm support is becoming more widespread on Windows, as such our vcvars handling needs an update.

Previously we assumed invoking `vcvars64.bat` would properly construct a 64bit environment, but it does not handle ARM, so we must instead invoke the generic `vcvarsall.bat` environmental setup script with properly constructed arguments to properly handle both ARM and amd/intel architectures. 

We now construct the arch argument we give to vcvarsall a bit more carefully as well.
Typically `x86_64` means "the 64bit extension of the x86 ISA", however with VCVARS, that string means "compile for 64bit with a 32bit host running the 32bit compiler", so we must translate the vcvars invocations in Spack to use the correct specification for 64bit host and 64bit target, with handling for arm. 
To do this, we determine the host bitness and intel/arm, and then compose the two values together with an underscore as vcvarsall expects.

### DIFF
--- a/repos/spack_repo/builtin/packages/msvc/package.py
+++ b/repos/spack_repo/builtin/packages/msvc/package.py
@@ -2,7 +2,9 @@
 #
 # SPDX-License-Identifier: (Apache-2.0 OR MIT)
 import os.path
+import platform
 import re
+import struct
 import subprocess
 
 from spack_repo.builtin.build_systems import compiler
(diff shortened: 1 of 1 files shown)


johnwparent:
@spackbot fix style

repos/spack_repo/builtin/packages/msvc/package.py
```diff
@@ -143,12 +145,40 @@ def init_msvc(self):
...
+        # run on a 64bit host, but is not accurate, and doesn't scale to arm.
+        # compose a proper vcvars arch string
+
+        # get spec target
+        spec_target_arch = self.spec.target.family.name
+        # get host bitness
+        host_machine_type = platform.machine().lower()
+
+        # generic check for arm arch
+        check_arm = lambda x: "arm" in x or "aarch" in x
```
scheibelp, line 163: IIRC this was because `platform.machine()` was potentially inconsistent  across systems. I was wondering if `archspec.HOST_TARGET_FAMILY` would help with that. Were you able to check into that?

bernhardkaindl:
@johnwparent, see the review question by scheibelp at https://github.com/spack/spack-packages/pull/1713#discussion_r2496657438

PR #1714, opened 2025-09-25
simsipm: Add conflict for aarch64

tmadlener:
Package uses x86intrin.h header which makes it incompatible with aarch64

### DIFF
--- a/repos/spack_repo/builtin/packages/simsipm/package.py
+++ b/repos/spack_repo/builtin/packages/simsipm/package.py
@@ -43,7 +43,12 @@ class Simsipm(CMakePackage):
     depends_on("python@3.6:", when="+python", type=("build", "run"))
     depends_on("py-pybind11", when="+python", type=("build", "link"))
 
-    conflicts("arch=aarch64", msg="package uses x86 internals and does not support aarch64")
+    conflicts("arch=aarch64", msg="SimSiPM < 2.1.0 uses x86 internals and does not support aarch64", when="@:2.0")
+    patch(
+        "https://github.com/EdoPro98/SimSiPM/commit/b08cc27151d70298b43ca7e216add9e46c056b04.patch?full_index=1",
+        sha256="354208cd110976a948e4770d7ae086cff367cef74534d07c22524f107ffbb7fd",
+        when="@2.1.0",
+    )
 
     def cmake_args(self):
         args = [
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/simsipm/package.py
```diff
@@ -43,6 +43,8 @@ class Simsipm(CMakePackage):
     depends_on("python@3.6:", when="+python", type=("build", "run"))
     depends_on("py-pybind11", when="+python", type=("build", "link"))
 
+    conflicts("arch=aarch64", msg="package uses x86 internals and does not support aarch64")
+
```
wdconinc: Just ran into this as well. Would you be interested in a patch instead? This appears fixed in main, https://github.com/EdoPro98/SimSiPM/commit/b08cc27151d70298b43ca7e216add9e46c056b04

```suggestion
    conflicts("arch=aarch64", msg="SimSiPM < 2.1.0 uses x86 internals and does not support aarch64", when="@:2.0")
    patch(
        "https://github.com/EdoPro98/SimSiPM/commit/b08cc27151d70298b43ca7e216add9e46c056b04.patch?full_index=1",
        sha256="354208cd110976a948e4770d7ae086cff367cef74534d07c22524f107ffbb7fd",
        when="@2.1.0",
    )

```

tmadlener: Nice, I guess, I opened this just the week too early :D 

Did you check whether this builds with the patch? Otherwise, I could try do that.

wdconinc: > Nice, I guess, I opened this just the week too early :D

Or the developer is keeping an eye on spack or key4hep issue trackers :-)

> Did you check whether this builds with the patch? Otherwise, I could try do that.

I haven't tested the build. I don't have an arm system easily accessible locally, so I'm having to get this into github workflows to test: https://github.com/eic/containers/actions/runs/19217416962/job/54943884917?pr=21 (without the patch).

wdconinc: Rather than figuring out how to cherry-pick from your fork as a new remote, I just changed the version to the commit of the patch: https://github.com/eic/containers/pull/21/commits/4c7acf4d18e72c76f6985c5cb388e29add5151b0. If that builds, it'll be a good sign that this patch indeed works. Just keep an eye on the pipeline at https://github.com/eic/containers/actions/runs/19236407689/job/54988113769?pr=21. If that passes, I'll mark as approved here too.

wdconinc (approved):
Builds on my arm runners with the commit pointed to by the patch. I am assuming the patched release will therefore also work.

repos/spack_repo/builtin/packages/simsipm/package.py
tmadlener: Thanks for checking.

tmadlener:
I can now also confirm that this builds on arm with the patch applied via spack.

PR #1721, opened 2025-09-25
py-starfile: new package

Markus92:
py-starfile is one of the (many) dependencies for RELION. Quoting: "[it] can be used interactively to inspect/explore files or in scripts and larger software packages to provide basic STAR file I/O functions."

It's a pretty simple pip wheel, no crazy tricks necessary to get it to build or function. Tried the examples out, just works as it should out of the box.


Markus92:
@spackbot fix style

repos/spack_repo/builtin/py_starfile/package.py
```diff
@@ -0,0 +1,46 @@
...
+    version("0.5.11", sha256="cc3161ab31a2e602b5feec3d2973ac6cc4209ca685affb8f05756bebb115af1f")
+    version("0.5.10", sha256="b9065776ffe088f0d4c11d7edeaa2bf0388253f9c20e47efcad30457cf6924f0")
+    version("0.5.8", sha256="877698060dbf729e90cae4094d0137dbe1a138ca0fab98ad851d468e59be3d29")
+    version("0.5.7", sha256="75b81959fc5d2a955907a01684c592108439f75d9ba2d5e85f102be4a5f67af5")
+    version("0.5.6", sha256="e6efe40a43ceae8ec6d85ec2fae62ae570cbadb981fdc76fe4a66a1b90c36631")
+    version("0.5.5", sha256="e472c24554a179e74ab2c0a6c43c21a185f081b9d1c3477b009ef9ba4461fd97")
+    version("0.5.4", sha256="4e7cea4f871f5aeeff6829b2cd95efc0c2212ca7a3a583f5fc8ff53329bc4417")
+    version("0.5.3", sha256="923a05e93612546f3071d8afa2b1673a143836ed519b73ba5dad3c323b9c761b")
+    version("0.5.2", sha256="f65ad12bad6b64b40bd101bf0aeb4c6ed2342e8f6e10473ac68a3dfe7a236ed4")
+    version("0.5.1", sha256="b8902d81256ac44faa3a09fd8e102a1493c673892f6a1fabb19655f5d505a946")
```
alecbcs: These three versions use `setuptools` as the build backend instead of hatchling. I'd suggest dropping them unless you really need the older versions. In which we should include `py-setuptools` as a build dependency for older versions.

repos/spack_repo/builtin/py_starfile/package.py
```diff
@@ -0,0 +1,46 @@
...
+
+    depends_on("python@3.8:", type=("build", "run"))
+
+    depends_on("py-hatchling", type="build")
+
+    depends_on("py-numpy", type=("build", "run"))
+    depends_on("py-pandas@2.1.1:", type=("build", "run"))
+    depends_on("py-arrow", type=("build", "run"))
```
alecbcs:
```suggestion
    depends_on("py-arrow", type=("build", "run"), when="@0.5.5:")
```

`py-arrow` is only a dependency from v0.5.5 and newer.

alecbcs (changes requested):
Just a couple suggestions. Otherwise this looks mostly good to me. Thanks!

repos/spack_repo/builtin/py_starfile/package.py
Markus92: Must've missed that one, removed the old versions. Thanks!

Markus92:
Merged in #2054

PR #1724, opened 2025-09-25
vala: add 0.56.18

michaelkuhn:


KineticTheory:
* Hash for v0.56.18 confirmed.  (I've been using an equivalent update in my local testing.)
* New `depends_on("gobject-introspection")` is also confirmed as required for the new version.
  * Should you add a `when="@0.56:"`?
* New `url_for_version` is also needed.

michaelkuhn:
> New `depends_on("gobject-introspection")` is also confirmed as required for the new version.
> * Should you add a `when="@0.56:"`?

Good point, fixed.

alecbcs (approved):
Looks good to me. Thanks!

PR #1726, opened 2025-09-25
mpich: Updated conditional for -fallow-argument-mismatch.

AcerP-py:
Starting with version 20 LLVM's fortran compiler `flang` is now what used to be `new-flang`. The new compiler does not accept the `-fallow-argument-mismatch` argument. Now (with spack v1 :partying_face:) we can actually check the name/version of the fortran compiler being used!

I am not quite sure how to handle the `apple-clang` case. My own experimentation showed that `clang` should be replaced with `llvm`. I am not entirely sure that `apple-clang` will be caught under the same conditional.


repos/spack_repo/builtin/packages/mpich/package.py
```diff
@@ -487,9 +487,7 @@ def flag_handler(self, name, flags):
         if name == "fflags":
             # https://bugzilla.redhat.com/show_bug.cgi?id=1795817
             # https://github.com/spack/spack/issues/17934
-            # TODO: we should add the flag depending on the real Fortran compiler spec and not the
-            #  toolchain spec, which might be mixed.
-            if any(self.spec.satisfies(s) for s in ["%gcc@10:", "%apple-clang@11:", "%clang@11:"]):
+            if any(self.spec["fortran"].satisfies(s) for s in ["gcc@10:", "llvm@11:19.1.7"]):
```
alalazo, line 497: You cannot assume "fortran" is always there, since it's conditionally added on `mpich+fortran`.

alalazo (dismissed):

repos/spack_repo/builtin/packages/mpich/package.py
AcerP-py: This should be resolved now. "fflags" will only be set when the spec depends on "fortran".

alalazo:
@raffenet  @yfguo For some reason you were not pinged on this PR. Waiting on your input on whether we should drop `apple-clang`

raffenet (approved):
LGTM

raffenet (approved):

alalazo:
@spackbot run pipeline

raffenet (approved):

AcerP-py:
@spackbot run pipeline

alalazo (approved):

PR #1727, opened 2025-09-25
kokkos ecosystem: v4.7.01

ndellingwood:
Updating the versions for the 4.7.01 patch release.

### DIFF
--- a/repos/spack_repo/builtin/packages/kokkos/package.py
+++ b/repos/spack_repo/builtin/packages/kokkos/package.py
@@ -28,6 +28,7 @@ class Kokkos(CMakePackage, CudaPackage, ROCmPackage):
 
     version("develop", branch="develop")
 
+    version("4.7.01", sha256="404cf33e76159e83b8b4ad5d86f6899d442b5da4624820ab457412116cdcd201")
     version("4.7.00", sha256="126b774a24dde8c1085c4aede7564c0b7492d6a07d85380f2b387a712cea1ff5")
     version("4.6.02", sha256="baf1ebbe67abe2bbb8bb6aed81b4247d53ae98ab8475e516d9c87e87fa2422ce")
     version("4.6.01", sha256="b9d70e4653b87a06dbb48d63291bf248058c7c7db4bd91979676ad5609bb1a3a")
@@ -445,7 +446,7 @@ class Kokkos(CMakePackage, CudaPackage, ROCmPackage):
             self.define_from_variant("Kokkos_ENABLE_IMPL_CUDA_MALLOC_ASYNC", "alloc_async")
         )
 
-        if self.version == Version("4.7.00"):
+        if self.version == Version("4.7.00") or self.version == Version("4.7.01"):
             options.append(self.define("Kokkos_ENABLE_IMPL_VIEW_LEGACY", True))
 
         # Remove duplicate options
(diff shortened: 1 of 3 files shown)


ndellingwood:
I'm unable to add/request reviewers, pinging a few people active with the kokkos ecosystem package support:
@cedricchevalier19 @lucbv @crtrott @masterleinad @rbberger

repos/spack_repo/builtin/packages/kokkos/package.py
```diff
@@ -445,7 +446,7 @@ def cmake_args(self):
             self.define_from_variant("Kokkos_ENABLE_IMPL_CUDA_MALLOC_ASYNC", "alloc_async")
         )
 
-        if self.version == Version("4.7.00"):
+        if self.version == Version("4.7.00") or self.version == Version("4.7.01"):
```
cedricchevalier19: This conditional was introduced as a temporary fix until 4.7.01, is it necessary to extend it to this new patch release?

tpadioleau: It does not seem necessary as the release 4.7.01 now defaults `Kokkos_ENABLE_IMPL_VIEW_LEGACY` to `True`, see https://github.com/kokkos/kokkos/blob/4.7.01/CHANGELOG.md#4701

tpadioleau (changes requested):
Thanks for the update, as the Kokkos release 4.7.01 also fixes build issues with CUDA 13, I would suggest to add a conflict with CUDA >=13 for Kokkos <4.7.1, in terms of Spack it should be `conflicts("^cuda@13:", when="@:4.7.0")`.

repos/spack_repo/builtin/packages/kokkos/package.py
ndellingwood: thanks @cedricchevalier19 @tpadioleau , I dropped the 4.7.01 version check in [f9f0c71](https://github.com/spack/spack-packages/pull/1727/commits/f9f0c71ce3d34adf47b210cb0ce89b33fcd871a5)

ndellingwood:
> I would suggest to add a conflict with CUDA >=13 for Kokkos <4.7.1, in terms of Spack it should be `conflicts("^cuda@13:", when="@:4.7.0")`.

@tpadioleau thanks for the suggestion and spack snippet, I added this in [f9f0c71](https://github.com/spack/spack-packages/pull/1727/commits/f9f0c71ce3d34adf47b210cb0ce89b33fcd871a5) , let me know if there is a better location in the package file where this should be placed

lucbv (approved):
Thanks @ndellingwood will keep an eye on the last CI check and ping someone to merge once it passes

tpadioleau (approved):
Looks ok to me, thank you.

By the way, this makes me realize that the conflict on cmake 3.28 is likely not working as expected. I can open a PR later.

cedricchevalier19 (approved):
Thank you!

lucbv:
@rbberger could you have a look and maybe merge this? Thx : )

tpadioleau:
> Looks ok to me, thank you.
> 
> By the way, this makes me realize that the conflict on cmake 3.28 is likely not working as expected. I can open a PR later.

see https://github.com/spack/spack-packages/pull/1766

rbberger (approved):

PR #1731, opened 2025-09-26
Update package.py for hypre version 3.0.0

rfalgout:

### DIFF
--- a/repos/spack_repo/builtin/packages/hypre/package.py
+++ b/repos/spack_repo/builtin/packages/hypre/package.py
@@ -28,6 +28,7 @@ class Hypre(AutotoolsPackage, CudaPackage, ROCmPackage):
     license("MIT")
 
     version("develop", branch="master")
+    version("3.0.0", sha256="d9dbfa34ebd07af1641f04b06338c7808b1f378e2d7d5d547514db9f11dffc26")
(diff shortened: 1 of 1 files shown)


alalazo (changes requested):
This is failing in pipelines

victorapm:
Hi @alalazo,

MFEM <= 4.8 is incompatible with >= hypre 3.0.0
The PR on MFEM has been merged this week and I expect it to be available on MFEM 4.9
Therefore I guess we need something like:
 
`conflicts('hypre@3.0.0:', when='@:4.8.0')`

in MFEM's spack package

victorapm:
On a related note, should spack packages in general not assume compatibility to dependencies when a new major version of the dependency comes in?

alalazo:
> On a related note, should spack packages in general not assume compatibility to dependencies when a new major version of the dependency comes in?

That depends on the specific package. We cannot assume any global constraint of this type since not all packages follow semver, and many packages are compatible with different major versions of a dependency.

victorapm:
Thanks @alalazo, that makes sense

Just checking: would you be able to implement the suggested change on MFEM's package? It seems I can't push to this branch otherwise I would do it. Otherwise, please let me know if you would like a separate PR

Thanks for helping!

victorapm:
@alalazo I opened PR #1855 mainly for CMake support in hypre and added the contents of this PR in there as well

victorapm:
Closing since hypre v3.0.0 support was added in #1855

PR #1736, opened 2025-09-26
mpibind: add support for NVIDIA's Grace Hopper architecture

eleon:
mpibind v0.23+ and hwloc v2.12+ are necessary to support NVIDIA's Grace Hopper.

### DIFF
--- a/repos/spack_repo/builtin/packages/mpibind/package.py
+++ b/repos/spack_repo/builtin/packages/mpibind/package.py
@@ -3,8 +3,11 @@
 # SPDX-License-Identifier: (Apache-2.0 OR MIT)
 
 import sys
+
 from spack_repo.builtin.build_systems.autotools import AutotoolsPackage
+
 from spack.package import *
+
 
 class Mpibind(AutotoolsPackage):
     """A portable runtime library that automatically maps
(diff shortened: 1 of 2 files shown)


repos/spack_repo/builtin/packages/mpibind/package.py
```diff
@@ -27,14 +25,33 @@ class Mpibind(AutotoolsPackage):
...
-    version("0.8.0", commit="ff38b9dcd150ca1e8a8796835d8e1e1847b3ba68", no_cache=True)
-    version("0.7.0", commit="3c437a97cd841b9c13abfbe1062a0285e1a29d3e", no_cache=True)
-    version("0.5.0", commit="8698f07412232e4dd4de4802b508374dc0de48c9", no_cache=True)
-
-    variant("cuda", default=False, description="Build w/support for NVIDIA GPUs.")
-    variant("rocm", default=False, description="Build w/support for AMD GPUs.")
-    variant("flux", default=False, description="Build the Flux plugin.")
-    variant("python", default=False, description="Build the Python bindings.")
+    version("0.23.0", commit="7d23407726004c7092a20ebdf6e9661f020b4ae7",
+            no_cache=True)
```
tgamblin: why are these all set to `no_cache=True`?  The only place we currently put this in the mainline package repository is if someone fetches `master` or `main` -- i.e. moving targets.  And since https://github.com/spack/spack/pull/50604, we store commit data on branch versions as well, so I don't think we should really have `no_cache` anywhere in `spack-packages`.

repos/spack_repo/builtin/packages/mpibind/package.py
```diff
@@ -27,14 +25,33 @@ class Mpibind(AutotoolsPackage):
...
+            no_cache=True, deprecated=True)
+    version("0.7.0", commit="3c437a97cd841b9c13abfbe1062a0285e1a29d3e",
+            no_cache=True, deprecated=True)
+    version("0.5.0", commit="8698f07412232e4dd4de4802b508374dc0de48c9",
+            no_cache=True, deprecated=True)
+
+    variant("cuda", default=False,
+            description="Build with support for NVIDIA GPUs")
+    variant("rocm", default=False,
+            description="Build with support for AMD GPUs")
```
tgamblin: Does `mpibind` itself depend on `cuda` or is this just to control `hwloc`?  Wondering whether this should use `CudaPackage` or `RocmPackage`:
* https://spack.readthedocs.io/en/latest/build_systems/cudapackage.html
* https://spack.readthedocs.io/en/latest/build_systems/rocmpackage.html

repos/spack_repo/builtin/packages/mpibind/package.py
```diff
@@ -46,24 +63,50 @@ class Mpibind(AutotoolsPackage):
...
+               when="+cuda")
+    depends_on("hwloc@2.4: +rocm +opencl", type="link",
+               when="@:0.19 +rocm")
+    depends_on("hwloc@2.4: +rocm", type="link",
+               when="@0.20: +rocm")
+
+    # Need mpibind v0.23+ and hwloc v2.12+ for NV Grace Hopper
+    depends_on("hwloc@2.12: +cuda +nvml", type="link",
+               when="@0.23: +cuda target=neoverse_v2")
+    conflicts("@:0.22 +cuda target=neoverse_v2",
```
tgamblin: Would also need mpibind and hwloc at at least these versions for newer ARM processors? Or is it really *just* `neoverse_v2` that requires this?

```suggestion
               when="@0.23: +cuda target=neoverse_v2:")
    conflicts("@:0.22 +cuda target=neoverse_v2:",
```

tgamblin (changes requested):

repos/spack_repo/builtin/packages/mpibind/package.py
eleon: Thanks @tgamblin ! 
The `no_cache=True` is (was?) needed because I pass the mpibind version number to `AC_INIT` (`configure.ac`) via the command `git describe --tags`. When I first created the spack recipe, using `no_cache=True` was the only way I could get this to work.

repos/spack_repo/builtin/packages/mpibind/package.py
eleon: This is just to control `hwloc`

tgamblin: can you add a note so that nobody gripes at you to make this a CudaPackage or RocmPackage?

eleon: Sure, done.

repos/spack_repo/builtin/packages/mpibind/package.py
eleon: Other thoughts, @tgamblin ?

eleon:
Anything else I need to do to get this approved?

tgamblin (dismissed):

tgamblin:
@eleon looks like style is falling -- so you will need tor resolve the style issues.

eleon:
@spackbot fix style

tgamblin (approved):

eleon:
Thanks very much, @tgamblin

PR #1739, opened 2025-09-26
libcap-ng: add 0.8.5

michaelkuhn:
This also replaces all versions to use the actual upstream tarballs (based on https://src.fedoraproject.org/rpms/libcap-ng) since those contain pre-generated `configure` scripts and newer versions do not work with our generic autoreconf setup.


tldahlgren (approved):
Confirmed the new homepage and new/updated version sha256s.

PR #1744, opened 2025-09-26
gsl: fix external cblas

Chrismarsh:
The cblas find doesn't seem to reliably work when an openblas virtual is used because openblas puts `cblas.h` into `prefix/include/openblas` which the patch doesn't handle. 

```
     156    checking for cblas.h... no
  >> 157    configure: error:
     158    	   	*** Header file cblas.h not found.
     159    	   	*** If you installed cblas header in a non standard place,
     160    	   	*** specify its install prefix using the following option
     161    	   	***  --with-cblas-external-cflags="-I<include_dir>"
```

This PR checks the virtual being used and appends the correct openblas suffix

@cessenat @mathomp4

### DIFF
--- a/repos/spack_repo/builtin/packages/gsl/package.py
+++ b/repos/spack_repo/builtin/packages/gsl/package.py
@@ -3,6 +3,7 @@
 # SPDX-License-Identifier: (Apache-2.0 OR MIT)
 
 import re
+import os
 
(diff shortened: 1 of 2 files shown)


Chrismarsh:
@spackbot fix style

cessenat (approved):
Thanks a lot, good thing.

mathomp4 (approved):

tldahlgren (approved):

PR #1747, opened 2025-09-27
terminalimageviewer: new package

ExplorerRay:
Terminal Image Viewer (tiv) is a program that can be used to check image in terminal.

### DIFF
--- a/repos/spack_repo/builtin/packages/terminalimageviewer/package.py
+++ b/repos/spack_repo/builtin/packages/terminalimageviewer/package.py
@@ -1,0 +1,32 @@
+# Copyright Spack Project Developers. See COPYRIGHT file for details.
+#
+# SPDX-License-Identifier: (Apache-2.0 OR MIT)
+
+from spack_repo.builtin.build_systems.makefile import MakefilePackage
+
+from spack.package import *
+
+
+class Terminalimageviewer(MakefilePackage):
+    """
+    Small C++ program to display images in a (modern) terminal
+    using RGB ANSI codes and unicode block graphic characters.
+    """
+
+    homepage = "https://github.com/stefanhaustein/TerminalImageViewer"
+    url = "https://github.com/stefanhaustein/TerminalImageViewer/archive/refs/tags/v1.2.1.tar.gz"
+
+    license("Apache-2.0	OR GPL-3.0-or-later", checked_by="ExplorerRay")
+
+    version("1.2.1", sha256="08d0c30e3ffa47b69d1bce07bea56f04b7deb4a8a79307ce435a4f0852fbcd5f")
+    version("1.2", sha256="f9b3c96554a24eccc3b4cdce2ce37fd3bade24a16a2e13038233f3b26a662542")
+
+    depends_on("gcc", type="build")
+
+    depends_on("imagemagick")
+
+    build_directory = "src"
+
+    def edit(self, spec, prefix):
+        makefile = FileFilter("src/Makefile")
+        makefile.filter(r"prefix *\?=\s*\/usr\/local", f"prefix ?= {prefix}")


tldahlgren (dismissed):
Confirmed the homepage and two version sha256s.

@ExplorerRay Please paste the output from `spack debug report` on the system you are able to successfully build each version using this package.

repos/spack_repo/builtin/packages/terminalimageviewer/package.py
```diff
@@ -0,0 +1,39 @@
...
+    version("1.2", sha256="f9b3c96554a24eccc3b4cdce2ce37fd3bade24a16a2e13038233f3b26a662542")
+
+    depends_on("cxx", type="build")
+
+    def edit(self, spec, prefix):
+        # Require g++ to build
+        which("g++", required=True)
+
+        makefile = FileFilter("src/Makefile")
+        makefile.filter(r"prefix *\?=\s*\/usr\/local", f"prefix ?= {prefix}")
```
alecbcs:
```suggestion
    depends_on("gcc", type="build")
    
    build_directory = "src"

    def edit(self, spec, prefix):
        makefile = FileFilter("src/Makefile")
        makefile.filter(r"prefix *\?=\s*\/usr\/local", f"prefix ?= {prefix}")
```

Instead of calling `which()` it's better practice to directly depend on the specific c++ compiler you expect. (Here depending on `gcc` instead of `cxx`.)

**Edit:** added `build_directory` as described in the other comment to help if you accept both suggestions.

repos/spack_repo/builtin/packages/terminalimageviewer/package.py
```diff
@@ -0,0 +1,39 @@
...
+
+    def build(self, spec, prefix):
+        with working_dir("src"):
+            make()
+
+    def install(self, spec, prefix):
+        with working_dir("src"):
+            make("install")
```
alecbcs:
```suggestion
```

You should be able to drop both of these methods as they'll be inherited from the `MakefilePackage` class and your reimplementations are essentially identical to the parent class. Instead you can define the following under `depends_on()` to modify the build directory of the package.

```
build_directory = "src"
```

alecbcs (changes requested):
@ExplorerRay two quick suggestions. Otherwise the PR looks mostly good to me. Thanks!

ExplorerRay:
```bash
$ spack debug report

* **Spack:** 1.1.0.dev0 (https://github.com/spack/spack/commit/2dc1993bc751440667dd6dc2f6d98741839d4eb0)
* **Builtin repo:** https://github.com/spack/spack-packages/commit/0214537f3472beca8e19c3d9a493a4dc3d27dad2
* **Python:** 3.13.5
* **Platform:** linux-debian13-icelake
```

By the way, I found that installing `terminalimageviewer@1.2` which also installs 1.2.1 because it also fits 1.2 but it makes me unavailable to check whether 1.2 is good or not. Do you have any advice on this?

repos/spack_repo/builtin/packages/terminalimageviewer/package.py
ExplorerRay: I just tested this code `makefile = FileFilter("Makefile")` and found it broken.

The [`edit()` in makefile.py](https://github.com/spack/spack-packages/blob/develop/repos/spack_repo/builtin/build_systems/makefile.py#L101) doesn't specify working directory and this makes it unable to find `Makefile` when installing.

alecbcs: Ahh interesting. I've updated the suggestion so that edit should now work.

alecbcs:
Yes to test installing `v1.2` you can type:
```console
$ spack install terminalimageviewer@=1.2
```

(You've happened to find the interesting edge case where `@` in Spack picks the most up-to-date version fitting the constraint, but `@=` will look for the exact version string requested.)

ExplorerRay:
I just revised those changes and add `imagemagick` dependency because I found that `tiv` can be successfully built but `tiv <image file>` will cause "unrecognized file format" error which needs imagemagick to fix.

alecbcs (approved):
Looks good to me. Thanks!

PR #1625, opened 2025-09-21
glib: add recent versions and fix introspection support

michaelkuhn:
This requires adding a glib-bootstrap package with introspection disabled, which can be used to bootstrap gobject-introspection. glib itself requires gobject-introspection in recent versions.

PR #1748, opened 2025-09-28
meson: deprecate superseded point releases

eli-schwartz:
Per usual, assume that a feature release line with multiple bugfix releases does not need every bugfix release.

As a special example, meson 1.8.* was a story in "we released a daring refactor and then had repeated regressions fixed over the course of several bugfix releases". You *really* do not want to be using 1.8.2 at this point.

1.8.4 is "probably safe", as that was the specific version to include a particularly large series of bugfixes, but... why chance it if users can be encouraged to prefer 1.8.5 instead?

### DIFF
--- a/repos/spack_repo/builtin/packages/meson/package.py
+++ b/repos/spack_repo/builtin/packages/meson/package.py
@@ -22,10 +22,7 @@ class Meson(PythonPackage):
     license("Apache-2.0")
 
     version("1.8.5", sha256="1cd0b5b013b4208ab450f5aca93b592b707f3fb2afe96b101dc710e6e5a8245c")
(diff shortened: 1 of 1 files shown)


michaelkuhn (approved):

michaelkuhn:
@spackbot run pipeline

KineticTheory:
@michaelkuhn Will #1741 fix the failing pipeline?

michaelkuhn:
> @michaelkuhn Will #1741 fix the failing pipeline?

I don't know, but I hope so. I created that PR while trying to fix CI failures for #1625. I don't really understand why and when it's failing, though.

PR #395, opened 2025-06-28
add googletest 1.17.0

plindsa:

PR #1749, opened 2025-09-28
googletest: add cxxstd restriction

sethrj:
In my review of #395 I missed a new restriction that Googletest requires C++17. https://github.com/google/googletest/releases/tag/v1.17.0


alecbcs (approved):
Looks good to me. Thanks @sethrj!

(P.S. it's best not to request review from a single person. The team will get to PRs as fast as we're able.)

PR #1755, opened 2025-09-29
py-pytest-fail-slow: add v0.6.0

manuelakuhn:
https://github.com/jwodder/pytest-fail-slow/tree/v0.6.0


tldahlgren (approved):
Confirmed new version sha256 and dependencies and the older version's sha256.

PR #1756, opened 2025-09-29
detray: deprecate versions :75

stephenswat:
This commit deprecates older versions of the detray package which are no longer useful.


tldahlgren (approved):
LGTM

PR #1760, opened 2025-09-29
py-pytest-runner: add v6.0.1

manuelakuhn:
https://github.com/pytest-dev/pytest-runner/tree/v6.0.1


alecbcs (approved):
Looks good to me. Thanks!

PR #1762, opened 2025-09-29
py-python-json-logger: add v3.3.0

manuelakuhn:
https://github.com/nhairs/python-json-logger/tree/v3.3.0

### DIFF
--- a/repos/spack_repo/builtin/packages/py_python_json_logger/package.py
+++ b/repos/spack_repo/builtin/packages/py_python_json_logger/package.py
@@ -27,7 +27,7 @@ class PyPythonJsonLogger(PythonPackage):
     depends_on("py-typing-extensions", type="build", when="@3.1: ^python@:3.9")
 
     def url_for_version(self, version):
-        if version >= Version("3.1"):
+        if self.spec.satisfies("@3.1:"):
             name = "python_json_logger"
         else:
             name = "python-json-logger"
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/py_python_json_logger/package.py
```diff
@@ -11,12 +11,24 @@ class PyPythonJsonLogger(PythonPackage):
...
+    depends_on("python@3.7:", type="build", when="@3:")
+
     depends_on("py-setuptools", type="build")
+    depends_on("py-typing-extensions", type="build", when="@3.1: ^python@:3.9")
+
+    def url_for_version(self, version):
+        if version >= Version("3.1"):
+            name = "python_json_logger"
+        else:
+            name = "python-json-logger"
```
alecbcs:
```suggestion
        if self.spec.satisfies("@3.1:"):
            name = "python_json_logger"
        else:
            name = "python-json-logger"
```

We can drop the addition version object creation here by instead using satisfies.

alecbcs (changes requested):
One quick suggestion. Otherwise looks good to me.

repos/spack_repo/builtin/packages/py_python_json_logger/package.py
manuelakuhn: Thanks for the hint.

alecbcs (approved):

PR #1768, opened 2025-09-29
py-breathe: add v4.36.0, add self as maintainer

sethrj:
This was released in February as a bugfix release. 

Version 4.36 now contains a pyproject.toml file. I have overhauled the package recipe to use the latest URL, remove incorrect dependencies, and use `default_args` for cleaner presentation.

### DIFF
--- a/repos/spack_repo/builtin/packages/py_breathe/package.py
+++ b/repos/spack_repo/builtin/packages/py_breathe/package.py
@@ -14,6 +14,8 @@ class PyBreathe(PythonPackage):
     homepage = "https://github.com/michaeljones/breathe"
     url = "https://github.com/michaeljones/breathe/archive/v4.11.1.tar.gz"
 
+    maintainers("sethrj")
+
     license("BSD-3-Clause")
 
     version("4.36.0", sha256="94fc2277e639929e898cb95559d9d2c51cf764b787eaf6fa04f848a6550c2ff5")
@@ -31,9 +33,6 @@ class PyBreathe(PythonPackage):
     version("4.7.2", sha256="982d47909d22fcd71b48bad5aef3644294340b24f612b2887cde2e3be464d960")
     version("4.7.1", sha256="afb1ab0084b25d3670fa8f5cf2eeaee6fe61bfc77876e3816b140eacd4949875")
     version("4.7.0", sha256="5629c67f5adb41f39375d36c5f0d60d34b1230be268125e535205d77f69211e4")
-
-    depends_on("c", type="build")  # generated
-    depends_on("cxx", type="build")  # generated
 
     depends_on("py-setuptools", type="build")
 
--- a/repos/spack_repo/builtin/packages/py_breathe/package.py
+++ b/repos/spack_repo/builtin/packages/py_breathe/package.py
@@ -16,6 +16,7 @@ class PyBreathe(PythonPackage):
 
     license("BSD-3-Clause")
 
+    version("4.36.0", sha256="94fc2277e639929e898cb95559d9d2c51cf764b787eaf6fa04f848a6550c2ff5")
     version("4.35.0", sha256="55b54723752fc04b892a0f868782b1df65e69db6ca94fb32cf04be495bfd7841")
     version("4.34.0", sha256="b22e70eb4000492508d687d71f258c8f9678398e277bcee0daf34cd438a46e25")
     version("4.33.1", sha256="acda75b216b227ec09ffd74378a6c7ed8f023df4539693c6e20574d1df52e75b")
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/py_breathe/package.py
tldahlgren: The python dependency for the new release -- per https://github.com/breathe-doc/breathe/blob/v4.36.0/pyproject.toml-- is `python@3.9:`.

tldahlgren: There is also a requirement on `py-flit-core@3.7:`.

tldahlgren: `py-docutils` does not show up as a requirement for the new version.

repos/spack_repo/builtin/packages/py_breathe/package.py
```diff
@@ -14,8 +14,11 @@ class PyBreathe(PythonPackage):
     homepage = "https://github.com/michaeljones/breathe"
```
tldahlgren: FYI. This URL redirects to https://github.com/breathe-doc/breathe.

repos/spack_repo/builtin/packages/py_breathe/package.py
tldahlgren: The new version has a dependency on `py-sphinx@7.2:`.

tldahlgren: This no longer appears to be a dependency in the new version.

repos/spack_repo/builtin/packages/py_breathe/package.py
```diff
@@ -31,9 +34,6 @@ class PyBreathe(PythonPackage):
     version("4.7.1", sha256="afb1ab0084b25d3670fa8f5cf2eeaee6fe61bfc77876e3816b140eacd4949875")
     version("4.7.0", sha256="5629c67f5adb41f39375d36c5f0d60d34b1230be268125e535205d77f69211e4")
 
-    depends_on("c", type="build")  # generated
-    depends_on("cxx", type="build")  # generated
-
     depends_on("py-setuptools", type="build")
```
tldahlgren: AFAICT. This is not a dependency of the latest version being added here in that it isn't listed in `pyproject.toml`.

repos/spack_repo/builtin/packages/py_breathe/package.py
```diff
@@ -14,8 +14,11 @@ class PyBreathe(PythonPackage):
     homepage = "https://github.com/michaeljones/breathe"
     url = "https://github.com/michaeljones/breathe/archive/v4.11.1.tar.gz"
```
tldahlgren: The urls also get re-directed, e.g., https://github.com/breathe-doc/breathe/archive/refs/tags/v4.36.0.tar.gz for versions since `4.29.0`.

tldahlgren: FWIW. I can confirm all of the version sha256s with the addition (at the bottom) of `url_for_version`:

```
        if version < Version("4.29.0"):
            acct = "michaeljones"
            path = ""
        else:
            acct = "breathe-doc"
            path = "refs/tags/"
        return f"https://github.com/{acct}/breathe/archive/{path}v{version}.tar.gz"
```

repos/spack_repo/builtin/packages/py_breathe/package.py
tldahlgren, line 58: This is not an explicit requirement of the new version per `pyproject.toml`.

tldahlgren (changes requested):
Confirmed the version sha256 but there are inconsistencies with the dependencies in the new version's `pyproject.toml`. I also some suggestions for avoiding the URL redirect.

When you're done, please paste the results of `spack debug report` from a successful build with the final changes to this package.

repos/spack_repo/builtin/packages/py_breathe/package.py
sethrj: Apologies, I didn't notice the new toml file (4.35 didn't have one).

tldahlgren: No worries.

repos/spack_repo/builtin/packages/py_breathe/package.py
sethrj: Even the old versions correctly use the new URL so I have just updated it.

repos/spack_repo/builtin/packages/py_breathe/package.py
sethrj: It is an implicit dependency since py-breathe consumes the output of doxygen, so I've left it in there. If you disagree I can remove it, but I think most users would be surprised if it's not available as a dependency.

sethrj: These and py-pygments were/are implicitly pulled through py-sphinx. I checked that the version ranges for the  transient dependencies match.

sethrj:
@spackbot fix style

sethrj:
@tldahlgren Thanks for catching the massive error on my part. I've overhauled and double checked the dependencies.

sethrj:
@tldahlgren This is ready for  a second look. Thank you!

tldahlgren (approved):

PR #1773, opened 2025-09-30
py-pandas: add v2.3.3

adamjstewart:
https://github.com/pandas-dev/pandas/releases/tag/v2.3.3

### DIFF
--- a/repos/spack_repo/builtin/packages/py_pandas/package.py
+++ b/repos/spack_repo/builtin/packages/py_pandas/package.py
@@ -23,6 +23,7 @@ class PyPandas(PythonPackage):
 
     tags = ["e4s"]
 
+    version("2.3.3", sha256="e05e1af93b977f7eafa636d043f9f94c7ee3ac81af99c13508215942e64c993b")
     version("2.3.2", sha256="ab7b58f8f82706890924ccdfb5f48002b83d2b5a3845976a9fb705d36c34dcdb")
     version("2.3.1", sha256="0a95b9ac964fe83ce317827f80304d37388ea77616b1425f0ae41c9d2d0d7bb2")
     version("2.3.0", sha256="34600ab34ebf1131a7613a260a61dbe8b62c188ec0ea4c296da7c9a06b004133")
@@ -91,11 +92,8 @@ class PyPandas(PythonPackage):
 
     with default_args(type=("build", "run")):
         # Based on PyPI wheel availability
-        depends_on("python@3.10:", when="@2.2:")
-        depends_on("python@3.9:", when="@2.1:")
-        depends_on("python@3.8:", when="@1.4:")
-
-        depends_on("python@:3.13")
+        depends_on("python@:3.14")
+        depends_on("python@:3.13", when="@:2.3.2")
         depends_on("python@:3.12", when="@:2.2.2")
         depends_on("python@:3.11", when="@:2.1.0")
         depends_on("python@:3.10", when="@:1.4")


repos/spack_repo/builtin/packages/py_pandas/package.py
```diff
@@ -91,11 +92,8 @@ class PyPandas(PythonPackage):
 
     with default_args(type=("build", "run")):
         # Based on PyPI wheel availability
-        depends_on("python@3.10:", when="@2.2:")
```
adamjstewart, line 94: This line is incorrect, I don't know where this came from. Other min versions are deprecated anyway.

repos/spack_repo/builtin/packages/py_pandas/package.py
tldahlgren, line 77: Should this be capped `meson@1.2.1:1` per https://github.com/pandas-dev/pandas/blob/v2.3.3/pyproject.toml?

tldahlgren, line 102: There should be a dependency on `py-numpy@2:` per https://github.com/pandas-dev/pandas/blob/v2.3.3/pyproject.toml, line 11.

tldahlgren: Interesting that the listed `numpy` `dependencies` (lines 30-32 of `pyproject.toml`) indicate ranges dependent on the `python` version but are not represented here -- probably because of the comment on  lines 9-10.

tldahlgren, line 113: There now appears to be a lower bound for this package: `python-dateutil@2.8.2:`.

repos/spack_repo/builtin/packages/py_pandas/package.py
```diff
@@ -91,11 +92,8 @@ class PyPandas(PythonPackage):
...
-        depends_on("python@3.9:", when="@2.1:")
-        depends_on("python@3.8:", when="@1.4:")
-
-        depends_on("python@:3.13")
+        depends_on("python@:3.14")
```
tldahlgren, line 95: FYI. The `pyproject.toml` file for the new version says `python@3.9:`.

tldahlgren (commented):
Confirmed the  new version sha256. There are discrepancies between the dependencies and the `pyproject.toml` file in the GitHub repository.

tldahlgren:
> Confirmed the new version sha256. There are discrepancies between the dependencies and the `pyproject.toml` file in the GitHub repository.

And the version of the file in the archive file.

tldahlgren (changes requested):
See previous comments.

repos/spack_repo/builtin/packages/py_pandas/package.py
adamjstewart: Yes, but Python 3.9 is the minimum supported Spack version and will be deprecated as well in a few weeks. So that's why there is no lower bound.

Pandas has close integration with CPython and generally doesn't work with Python versions before they are released. So that's why there is an upper bound.

repos/spack_repo/builtin/packages/py_pandas/package.py
adamjstewart: We don't usually add upper bounds on build dependencies, they tend to be overconstrained because PyPI doesn't allow fixes of prior releases. @rgommers may have opinions on this.

adamjstewart: The build dependency on `numpy>=2` is required to generate wheels that are compatible with both numpy 1 and 2. However, Spack ensures a single version of numpy is present in the DAG, so the build-time numpy version always matches the run-time version.

The Python version-specific numpy versions are unnecessary because the numpy package has the same Python versions as dependencies. I have no idea why pandas duplicated this logic.

adamjstewart: This is on line 110 in the `package.py`

tldahlgren: 🤦

rgommers: Agreed, no need to use `<2` here. Meson 2.0 isn't a thing, and in case it does materialize and it happens to break `pandas` (both unlikely), making a change at that point to add the necessary constraint would be the normal way of doing this.

rgommers (approved):
LGTM, thanks

tldahlgren (approved):

PR #1776, opened 2025-09-30
py-pytest-cov: add v7.0.0

manuelakuhn:
https://github.com/pytest-dev/pytest-cov/tree/v7.0.0


tldahlgren (approved):
Confirmed all of the version sha256s and the dependencies for the new version.

PR #1789, opened 2025-09-30
RepeatModeler: Fix directory used during configuration

emwjacobson:
While attempting to install RepeatModeler, I found that it's configuration script looks for the `LTR_retriever` program within `spec["ltr-retriever"].prefix.bin`. The problem is that there is no `LTR_retriever` program within `bin/`, but rather it's within the project's root. ([LTR_retriever repo](https://github.com/oushujun/LTR_retriever)).

Obligatory mention of the maintainer: @snehring


snehring (approved):
good catch, not sure how this has gone unnoticed so long.

alalazo (approved):

PR #1790, opened 2025-09-30
CI: write_dotenv.py (needed by #342)

kwryankrattiger:
Add script for writing the dotenv file.

### DIFF
--- a/.ci/gitlab/scripts/common/write_dotenv.py
+++ b/.ci/gitlab/scripts/common/write_dotenv.py
@@ -6,13 +6,13 @@ CI_AUX_DIR = os.path.join(".ci", "tmp", "spack")
 
 
 def configure_packages_repo(dotenv):
-    dotenv["SPACK_PACKAGES_ROOT"] = CI_PROJECT_DIR
-    dotenv["SPACK_ROOT"] = CI_AUX_DIR
+    dotenv["SPACK_CI_PACKAGES_ROOT"] = CI_PROJECT_DIR
+    dotenv["SPACK_CI_SPACK_ROOT"] = CI_AUX_DIR
 
 
 def configure_core_repo(dotenv):
-    dotenv["SPACK_PACKAGES_ROOT"] = CI_AUX_DIR
-    dotenv["SPACK_ROOT"] = CI_PROJECT_DIR
+    dotenv["SPACK_CI_PACKAGES_ROOT"] = CI_AUX_DIR
+    dotenv["SPACK_CI_SPACK_ROOT"] = CI_PROJECT_DIR
 
 
 def read_dotenv(file):
--- a/.ci/gitlab/scripts/common/write_dotenv.py
+++ b/.ci/gitlab/scripts/common/write_dotenv.py
@@ -1,0 +1,47 @@
+import os
+import sys
+
+CI_PROJECT_DIR = "."
+CI_AUX_DIR = os.path.join(".ci", "tmp", "spack")
+
+
+def configure_packages_repo(dotenv):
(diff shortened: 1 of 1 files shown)


.ci/gitlab/scripts/common/write_dotenv.py
```diff
@@ -0,0 +1,47 @@
...
+def configure_core_repo(dotenv):
+    dotenv["SPACK_PACKAGES_ROOT"] = CI_AUX_DIR
+    dotenv["SPACK_ROOT"] = CI_PROJECT_DIR
+
+
+def read_dotenv(file):
+    dotenv = {}
+    for line in file:
+        if "=" in line:
+            key, value = line.rsplit("=", 1)
```
johnwparent: Would you want an lsplit here? I.e. the first equals from the left will always be the delineator between the key and value, but the value itself could have an equals (i.e. if its a hash or similar). Probably an edge case.

.ci/gitlab/scripts/common/write_dotenv.py
```diff
@@ -0,0 +1,47 @@
...
+
+def write_dotenv(env, file):
+    for var, value in env.items():
+        file.write(f"{var}={value}\n")
+
+
+if __name__ == "__main__":
+    dotenv = {}
+    base_dotenv = sys.argv[1] if len(sys.argv) > 1 else None
+    if os.path.exists(base_dotenv):
```
johnwparent: (uber nit) Warning if this doesn't exist? Just trying to get ahead of future debugging efforts.

johnwparent (dismissed):

.ci/gitlab/scripts/common/write_dotenv.py
kwryankrattiger: Yeah...I am trying to remember why I did it this way. maybe I was standing on my head 🙃

johnwparent: Must have just been in the southern hemisphere

.ci/gitlab/scripts/common/write_dotenv.py
kwryankrattiger: Only if `base_dotenv is not None`, but yeah it should die on that.

johnwparent (dismissed):

johnwparent:
@becker33 Do you want to give this another pass before I merge?

kwryankrattiger:
So I found an issue with using `SPACK_ROOT` here. This gets overwritten by `setup-env.sh` which uses the absolute path instead of the relative path.

I am thinking about renaming the variables to `SPACK_CI_SPACK_ROOT` and `SPACK_CI_PACKAGES_ROOT` to avoid the conflicts. Any objection?

johnwparent:
> So I found an issue with using `SPACK_ROOT` here. This gets overwritten by `setup-env.sh` which uses the absolute path instead of the relative path.
> 
> I am thinking about renaming the variables to `SPACK_CI_SPACK_ROOT` and `SPACK_CI_PACKAGES_ROOT` to avoid the conflicts. Any objection?

Sounds good to me

johnwparent (approved):

kwryankrattiger:
Passing here: https://gitlab.spack.io/spack/spack-packages/-/pipelines/1255924

alalazo:
Why the script is added separately, and not in #342 (since it seems unused in this PR)?

kwryankrattiger:
> Why the script is added separately, and not in #342 (since it seems unused in this PR)?

There was concern about merging the other PR referencing a PR instead of a hash. This was merged first to satisfy that a static commit on develop could be used.

PR #1791, opened 2025-09-30
mvapich2: add autoconf as build dependency 

peterivatt:
…include it as a dependency.

If m4 is used used as a spec in the same environment as mvapich2 it causes autoconf to not link correctly resulting in autom4te not being found. It would be safer to just include autoconf as a dependency.


bernhardkaindl (approved):
Hi @harisubramoni, @natshineman and @ndcontini,

spackbot invited you to review this change two months ago. Could you take a look?

While you didn't review change in this time frame, because this change should definitely not hurt, I assume we can merge this safely.

PR #1793, opened 2025-09-30
restic: add v0.18.1

alecbcs:
https://github.com/restic/restic/releases/tag/v0.18.1


manuelakuhn (approved):
LGTM and builds fine on rocky 8 with gcc@15

PR #1796, opened 2025-10-01
py-torch: Populate libs property properly

tmadlener:
This populates the `libs` properties of `py-torch` properly. Without this `libs` returns an empty list, as the fact that they are installed in the python platform directory puts them outside of the default root for searching.

### DIFF
--- a/repos/spack_repo/builtin/packages/py_torch/package.py
+++ b/repos/spack_repo/builtin/packages/py_torch/package.py
@@ -800,6 +800,10 @@ class PyTorch(PythonPackage, CudaPackage, ROCmPackage):
             python("run_test.py")
 
     @property
+    def libs(self):
+        return find_libraries("libtorch*", root=python_platlib, recursive=True, shared=True)
+
+    @property
     def cmake_prefix_paths(self):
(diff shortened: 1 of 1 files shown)


haampie (approved):
LGTM. What package requires this?

tmadlener:
None really, so far (as far as I am aware), but I just stumbled over https://github.com/spack/spack/pull/49267 once again, and while debugging saw that `libs` is also not populated. Since I was already there, this was a quick fix.

---------------
For context: What I tried to do was something along the lines of
```python
env.prepend_path("CMAKE_PREFIX_PATH", self.spec["py-torch"].cmake_prefix_paths[0])
```
but that doesn't work because `cmake_prefix_paths` is not an attribute of `Spec`. Hence, I intermediately went for a solution based on
```python
self.spec["py-torch"].libs
```
before realizing that `cmake_prefix_paths` is accessible as
```python
self.spec["py-torch"].package.cmake_prefix_paths
```

haampie:
Ah okay. If it doesn't fix any build or runtime issues, I think it's better to leave it out. Otherwise it's just a redundant filesystem scan as part of setting up Spack's build environment (which is expensive if py-torch is marked external).

tmadlener:
fair enough. Closing this again.

Issue #1779, opened 2025-09-30
Installation issue: py-fastjsonschema (won't concretize with ^python@3.13)


payerle:
### Steps to reproduce the issue

```console
$ spack spec -l py-fastjsonschema@2.21.1 ^python@3.13.5 |& tee spec.out
==> Error: failed to concretize `py-fastjsonschema@2.21.1 ^python@3.13.5` for the following reasons:
     1. Cannot satisfy 'python@3.13.5'
     2. Cannot satisfy 'python@3.13.5'
        required because py-fastjsonschema@2.21.1 ^python@3.13.5 requested explicitly 
     3. Cannot satisfy 'python@3.9:' and 'python@3.13.5'
        required because py-fastjsonschema@2.21.1 ^python@3.13.5 requested explicitly 
        required because py-setuptools depends on python@3.9: when @75.4: 
...
```
Full output of spec command is attached in "Error message"

### Error message

Spec fails to concretize

<details><summary>Error message</summary>
<pre>
.==> Error: failed to concretize `py-fastjsonschema@2.21.1 ^python@3.13.5` for the following reasons:
     1. Cannot satisfy 'python@3.13.5'
     2. Cannot satisfy 'python@3.13.5'
        required because py-fastjsonschema@2.21.1 ^python@3.13.5 requested explicitly 
     3. Cannot satisfy 'python@3.9:' and 'python@3.13.5'
        required because py-fastjsonschema@2.21.1 ^python@3.13.5 requested explicitly 
...
</pre>
</details>
Full output of spec command is attached
[spec.txt](https://github.com/user-attachments/files/22620768/spec.txt)

### Information on your system

spack debug report
* **Spack:** 1.1.0.dev0
* **Builtin repo:** https://github.com/spack/spack-packages/commit/4b8bb3514838ce9e8bbc2ca7f98e62dbd7100c63
* **Python:** 3.6.8
* **Platform:** linux-rhel8-zen2

### Additional information

According to information on PyPI, e.g. https://pypi.org/project/fastjsonschema/2.21.1/, recent versions of py-fastjsonspec should support python@3.13

The issue appears to the the following lines in the recipe
<pre>
    depends_on("python@3.3:3.13", when="@2.21:", type=("build", "run"))
    depends_on("python@3.3:3.12", when="@2.20:", type=("build", "run"))
    depends_on("python@3.3:3.11", when="@2.16.3:", type=("build", "run"))
[...]

PR #1805, opened 2025-10-01
py-fastjsonschema: Fix issue with python@3.12:

payerle:
This should #1779

The original depends_on(python) lines restricting python versions were using open ended version ranges for py-fastjsonschema in the when clause, effectively making the lines for python@3.3.12: ignored.

This fixes that by adding upper limits on the version ranges in the when clauses.


payerle:
Sorry, previous comment should be
This should fix issue #1779

bernhardkaindl (approved):
Correct fix, but in develop #2330 was merged, as three is no reason to believe that even the older versions won't work with newer Python versions.

While they weren't tested with those versions, they should not break it.
Thus, the version dependencies have been removed.

In this regard, the removed version dependencies do not do that.

They left 2.15 open for `3.3:`, but its setup.py lists only up to 3.8.
In any case, they were broken from the start and should not have been merged in the first place, which was relatively recent: https://github.com/spack/spack/pull/47926

We could instead deprecate some of the older versions, but there is no need to.

Closing this PR as obsolete.

PR #1807, opened 2025-10-01
eigen: add versions 5.0.0 and 3.4.1

### LINKED ISSUE #1786: Add new Eigen versions 5.0.0 and 3.4.1
The long-awaited new Eigen releases are there !

See the release notes :
- https://gitlab.com/libeigen/eigen/-/releases/5.0.0
- https://gitlab.com/libeigen/eigen/-/releases/3.4.1

@HaoZeke: Since you are the package maintainer, I let you the initiative to start a PR, unless you are ok with xomeone elseto do it.

jcortial-safran:
Closes https://github.com/spack/spack-packages/issues/1786

On top of adding 5.0.0 and 3.4.1, I had to make 2 other changes:
1) Add a build dependency on Fortran. By default, if a Fortran compiler is detected, Eigen uses it to build its internal BLAS and LAPACK implementations. This behavior makes building recent (3.4.0, 3.4.1 and 5.0.0) versions of Eigen fail, at least on my machine, unless an explicit dependency on Fortran is declared. I never experienced such an issue before, I wonder if it is related to recent changes in spack develop.
2) Declare version 3.4.0-44-ge7248b26a deprecated otherwise the specification eigen@3.4.0 builds it instead of the actual 3.4.0 release. In most cases 3.4.1 should be preferred anyway.

Note that there is no official 4.x.x release and in theory there will never be due to numbering changes, see release notes https://gitlab.com/libeigen/eigen/-/releases#versioning for details.


jcortial-safran:
@spackbot fix style

repos/spack_repo/builtin/packages/eigen/package.py
```diff
@@ -23,12 +23,16 @@ class Eigen(CMakePackage, ROCmPackage):
     license("MPL-2.0")
 
     version("master", branch="master")
-    version("3.4.0-44-ge7248b26a", commit="e7248b26a1ed53fa030c5c459f7ea095dfd276ac")
     version(
-        "3.4.0",
-        sha256="8586084f71f9bde545ee7fa6d00288b264a2b7ac3607b974e54d13e7162c1c72",
+        "5.0.0",
+        sha256="315c881e19e17542a7d428c5aa37d113c89b9500d350c433797b730cd449c056",
         preferred=True,
```
wdconinc: `preferred` is not necessary unless you want to change the ordering. In this case, that's not the case (anymore).
```suggestion
```

repos/spack_repo/builtin/packages/eigen/package.py
```diff
@@ -73,6 +77,7 @@ class Eigen(CMakePackage, ROCmPackage):
 
     depends_on("c", type="build")
     depends_on("cxx", type="build")
+    depends_on("fortran", type="build")
```
wdconinc, line 77: There is a change in the minimum cmake version required. It would be a good time to add this dependency with a version.

jcortial-safran:
It seems the update broke a few packages, including lammps, vtk and paraview.

I would argue that the recipes were not specific enough and should request more specific version ranges :)

vicentebolea (commented):
ParaView and VTK changes look good.

jcortial-safran:
> ParaView and VTK changes look good.

Unfortunately there seems to be a failure on Windows. We will have to investigate.

johnwparent:
> > ParaView and VTK changes look good.
> 
> Unfortunately there seems to be a failure on Windows. We will have to investigate.

If it fails again and you can't get access to a Windows machine to debug if need be, you can ping me.

rbberger (commented):
lammps changes look good enough for now.

jcortial-safran:
There seems to be an issue with newer Eigen versions on Windows : https://gitlab.com/libeigen/eigen/-/issues/2980

I added a conflict directive to use 3.4.0 rather than 3.4.1 on Windows.

There is still an issue with proj but I think it is unrelated to Eigen. I have seen that there is a PR to address it : https://github.com/spack/spack-packages/pull/1862

jcortial-safran:
I added upper bounds for eigen in the acts and geomodel packages

jcortial-safran:
Waiting to see if https://github.com/spack/spack-packages/pull/1910 will fix the CI

johnwparent:
Fixes for CI are in, lets see how this does now

melven:
Remark on 1:

> Add a build dependency on Fortran. By default, if a Fortran compiler is detected, Eigen uses it to build its internal BLAS and LAPACK implementations. This behavior makes building recent (3.4.0, 3.4.1 and 5.0.0) versions of Eigen fail, at least on my machine, unless an explicit dependency on Fortran is declared. I never experienced such an issue before, I wonder if it is related to recent changes in spack develop.

I observed the same problem with an existing older spack installation where `eigen@3.4.0` builds fine (not using these changes here). So I assume the problem originates from changes in Eigen.
We could use `depends_on("fortran", type="build", when="@3.4.1:")` to not break the old behavior.

jcortial-safran:
> Remark on 1:
> 
> > Add a build dependency on Fortran. By default, if a Fortran compiler is detected, Eigen uses it to build its internal BLAS and LAPACK implementations. This behavior makes building recent (3.4.0, 3.4.1 and 5.0.0) versions of Eigen fail, at least on my machine, unless an explicit dependency on Fortran is declared. I never experienced such an issue before, I wonder if it is related to recent changes in spack develop.
> 
> I observed the same problem with an existing older spack installation where `eigen@3.4.0` builds fine (not using these changes here). So I assume the problem originates from changes in Eigen. We could use `depends_on("fortran", type="build", when="@3.4.1:")` to not break the old behavior.

Fair enough. I just checked again that eigen@3.4.0 builds without explicit dependency on fortran, at least on my machine. I made the change you proposed.

To follow up, we should add an option to control whether to build the internal BLAS/LAPACK implementations since these are the components that require fortran, if I understand correctly.

jcortial-safran:
@wdconinc : Now that the CI passes, the PR is ready

jcortial-safran:
> @wdconinc : Now that the CI passes, the PR is ready

@wdconinc: Would it be possible to consider merging the PR, or to get some feedback to improve it ? Thanks :)

repos/spack_repo/builtin/packages/eigen/package.py
```diff
@@ -73,11 +73,18 @@ class Eigen(CMakePackage, ROCmPackage):
...
+
+    depends_on("cmake@3.10:", when="@3.4.1:", type="build")
+    depends_on("cmake@3.5:", when="@3.4.0", type="build")
+    depends_on("cmake@:3", when="@:3.3", type="build")
 
     depends_on("boost@1.53:", when="@master", type="test")
     # TODO: latex and doxygen needed to produce docs with make doc
     # TODO: Other dependencies might be needed to test this package
 
+    conflicts("@3.4.1", when="platform=windows")
```
johnwparent, line 86: Note to self(non blocking): Figure out if this can be resolved.

johnwparent (commented):
Looks like the fix for the Windows build issues is persistent in 5.0 (based on the merge date of https://gitlab.com/libeigen/eigen/-/merge_requests/2019) can we add a conflict for that on Windows as well? I'll make another PR backporting the patch from that MR so we don't delay this PR any longer.

johnwparent (approved):
Looks great! LGTM

jcortial-safran:
> Looks like the fix for the Windows build issues is persistent in 5.0 (based on the merge date of https://gitlab.com/libeigen/eigen/-/merge_requests/2019) can we add a conflict for that on Windows as well? I'll make another PR backporting the patch from that MR so we don't delay this PR any longer.

https://gitlab.com/libeigen/eigen/-/merge_requests/2017 may also be related.

johnwparent:
> > Looks like the fix for the Windows build issues is persistent in 5.0 (based on the merge date of https://gitlab.com/libeigen/eigen/-/merge_requests/2019) can we add a conflict for that on Windows as well? I'll make another PR backporting the patch from that MR so we don't delay this PR any longer.
> 
> https://gitlab.com/libeigen/eigen/-/merge_requests/2017 may also be related.

Very much yes, good catch.
The Windows symbol visibility model is always a fun surprise for predominately *nix based projects.

johnwparent:
If no one has any qualms with this I'll merge tonight/tomorrow morning. 

cc @wdconinc

PR #1818, opened 2025-10-02
py-referencing: add v0.36.2

manuelakuhn:
https://github.com/python-jsonschema/referencing/tree/v0.36.2


alalazo (approved):
I'm merging this, since it's small and has been waiting for a long time. @wdconinc let me know if you have any request, I'll follow up with those in case.

PR #1829, opened 2025-10-02
py-rpds-py: add v0.27.1

manuelakuhn:
https://github.com/crate-py/rpds/tree/v0.27.1

I don't know where the rust version restriction came from, but I left it in anyway.


tldahlgren:
Added missing maintainer to the Reviewers list.

wdconinc (dismissed):
Looks good to me. If I recall correctly, cargo-msrv on the previous versions returned the list rust version, but it is not imposed by the pyproject.toml. Listing explicitly avoids later build errors when an older rust version is used.

wdconinc:
@manuelakuhn Let me know if you want to look into the msrv.

manuelakuhn:
Thanks, I added the rust dependency versions now. I used `cargo msrv list` for that (I hope that is correct, I am not that familiar with cargo and rust and `cargo msrv find` did not work)

wdconinc (approved):
Thanks! I'm myself not a user of rust much either...

PR #1830, opened 2025-10-02
py-rst2pdf: add v0.103.1

manuelakuhn:
https://github.com/rst2pdf/rst2pdf/tree/0.103.1

The dependency upper bound restriction in [pyproject.toml](https://github.com/rst2pdf/rst2pdf/blob/0.103.1/pyproject.toml#L26) seemed too strict. And the package seems to work fine with newer versions, so they were loosened.

Needs #1351 to build


repos/spack_repo/builtin/packages/py_rst2pdf/package.py
```diff
@@ -20,18 +20,32 @@ class PyRst2pdf(PythonPackage):
...
+    depends_on("python@3.8:", type=("build", "run"), when="@0.100:0.102")
+
+    depends_on("py-setuptools@64:", type="build", when="@0.103:")
+    depends_on("py-setuptools", type="build", when="@:0.102")
+    depends_on("py-setuptools-scm@8:", type="build", when="@0.103:")
+    depends_on("py-setuptools-scm", type="build", when="@:0.102")
+
+    # ignore upper bound restrictions, because they seem to be too strict
+    depends_on("py-docutils@0.21.2:", type=("build", "run"), when="@0.103:")
+    depends_on("py-docutils", type=("build", "run"), when="@:0.102")
```
alalazo: In cases like this one, it's fine to leave the unconstrained declaration like it was before, and just add a lower bound for newer versions. You can probably leave the previous block untouched, and add[^1]:
```python
with default_args(type=("build", "run")), when("@0.103:"):
    depends_on("py-importlib-metadata@8:")
    ...
```

[^1]: Beware of typos, I didn't try the code 🙂

repos/spack_repo/builtin/packages/py_rst2pdf/package.py
```diff
@@ -20,18 +20,32 @@ class PyRst2pdf(PythonPackage):
...
+    depends_on("py-importlib-metadata", type=("build", "run"), when="@:0.102")
+    depends_on("py-jinja2@3:", type=("build", "run"), when="@0.103:")
+    depends_on("py-jinja2", type=("build", "run"), when="@:0.102")
+    depends_on("py-packaging@24:", type=("build", "run"), when="@0.103:")
+    depends_on("py-packaging", type=("build", "run"), when="@:0.102")
+    depends_on("py-pygments@2:", type=("build", "run"), when="@0.103:")
+    depends_on("py-pygments", type=("build", "run"), when="@:0.102")
+    depends_on("py-pyyaml@6:", type=("build", "run"), when="@0.103:")
+    depends_on("py-pyyaml", type=("build", "run"), when="@:0.102")
+    depends_on("py-reportlab@4:", type=("build", "run"), when="@0.103")
```
alalazo: Missing `:` ?

```suggestion
    depends_on("py-reportlab@4:", type=("build", "run"), when="@0.103:")
```

alalazo (changes requested):

repos/spack_repo/builtin/packages/py_rst2pdf/package.py
manuelakuhn: Thanks. I have removed the constraint now.

alalazo (approved):

PR #1834, opened 2025-10-02
libgit2-glib: new spack recipe for gnome package.

KineticTheory:
**New package:**  `libgit2-glib` is a `glib` wrapper library around the `libgit2` git access library. `libgit2` only implements the core plumbing functions, not really the higher-level porcelain stuff.

### DIFF
--- a/repos/spack_repo/builtin/packages/libgit2_glib/package.py
+++ b/repos/spack_repo/builtin/packages/libgit2_glib/package.py
@@ -1,0 +1,53 @@
+# Copyright Spack Project Developers. See COPYRIGHT file for details.
+#
+# SPDX-License-Identifier: (Apache-2.0 OR MIT)
+
+from spack_repo.builtin.build_systems import meson
+from spack_repo.builtin.build_systems.meson import MesonPackage
+
+from spack.package import *
+
+
+class Libgit2Glib(MesonPackage):
+    """libgit2-glib is a glib wrapper library around the libgit2 git access library.
+    libgit2 only implements the core plumbing functions, not really the higher
+    level porcelain stuff."""
+
+    homepage = "https://gitlab.gnome.org/GNOME/libgit2-glib"
+    url = "https://download.gnome.org/sources/libgit2-glib/1.2/libgit2-glib-1.2.1.tar.xz"
+
+    maintainers("KineticTheory")
+
+    license("LGPL-2.1-or-later")
+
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/libgit2_glib/package.py
```diff
@@ -0,0 +1,45 @@
...
+    """libgit2-glib is a glib wrapper library around the libgit2 git access library.
+    libgit2 only implements the core plumbing functions, not really the higher
+    level porcelain stuff."""
+
+    homepage = "https://gitlab.gnome.org/GNOME/libgit2-glib"
+    url = "https://download.gnome.org/sources/libgit2-glib/1.2/libgit2-glib-1.2.1.tar.xz"
+
+    maintainers("KineticTheory")
+
+    license("GNU LGPL-2.1")
```
rbberger: This should be:
```suggestion
    license("LGPL-2.1-or-later")
```

KineticTheory: Fixed! TY.

KineticTheory:
@rbberger I implemented your suggested change last week.  I think this is ready for a 2nd look and possible merge.  TIA.

rbberger (dismissed):

repos/spack_repo/builtin/packages/libgit2_glib/package.py
```diff
@@ -0,0 +1,45 @@
...
+    depends_on("py-pygobject", type="build")
+    depends_on("vala", type="build")
+
+    def url_for_version(self, version):
+        return f"https://download.gnome.org/sources/libgit2-glib/{version.up_to(2)}/libgit2-glib-{version}.tar.xz"
+
+
+class MesonBuilder(meson.MesonBuilder):
+    def meson_args(self):
+        args = ["-Dssh=false"]
```
rbberger: It might be worth expanding this boolean from `self.spec.satisfies("^libgit2+ssh)`. Similarly, other dependencies are only needed for certain [options](https://gitlab.gnome.org/GNOME/libgit2-glib/-/blob/master/meson_options.txt?ref_type=heads), which could be mapped as variants.

KineticTheory: @rbberger I've added variants to support some of the other meson build options including logic to enable ssh support.  I did not add a variant `+docs` because this build mode requires additional packages that do not yet exist in the spack ecosystem.

rbberger (approved):

PR #1835, opened 2025-10-02
CI: Stop using about-to-be-removed image "macos-13"

hartwork:
> The macOS 13 runner image will be retired by December 4th, 2025.

### DIFF
--- a/.github/workflows/bootstrap.yml
+++ b/.github/workflows/bootstrap.yml
@@ -84,7 +84,7 @@ jobs:
     runs-on: ${{ matrix.runner }}
     strategy:
       matrix:
-        runner: ['macos-13', 'macos-14', "ubuntu-latest"]
+        runner: ['macos-14', "ubuntu-latest"]
     steps:
       - name: Setup macOS
         if: ${{ matrix.runner != 'ubuntu-latest' }}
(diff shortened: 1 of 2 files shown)


.github/workflows/bootstrap.yml
```diff
@@ -84,7 +84,7 @@ jobs:
     runs-on: ${{ matrix.runner }}
     strategy:
       matrix:
-        runner: ['macos-13', 'macos-14', "ubuntu-latest"]
+        runner: ['macos-14', "ubuntu-latest"]
```
alalazo: Thanks for the PR. Can you:
- substitute `macos-13` with `macos-15-intel`
- substitute `macos-14` with `macos-latest`

instead?

alalazo (changes requested):

.github/workflows/bootstrap.yml
hartwork: @alalazo I can try it, but if CI is not all green I will have to revert.

hartwork: @alalazo PS: `-latest` is not a good idea because CI should have as few moving pieces as possible, ideally zero. Moving pieces break pull requests in just the wrong moment.

hartwork: @alalazo pushed.

alalazo (approved):

PR #1838, opened 2025-10-02
kokkos-kernels: Make +lapack depend on +blas

brian-kelley:
In KokkosKernels, we had the ``+blas`` and ``+lapack`` variants being independent of each other. But in the library code, certain macros were defined only when the BLAS TPL is enabled, and the LAPACK TPL wrappers use these same macros. Effectively this means the LAPACK TPL depends on the BLAS TPL.

For spack, this meant specs like ``kokkos-kernels +lapack`` would fail during the build step:
```
  >> 153    .../spack-stage/spack-stage-kokkos-kernels-4.7.01-gpgzhtduestvywivx5iz4zu2m75mo5cp/spack-src/lapack/tpls/KokkosLapack_Host_tpl.cpp:32:27: error: expected ')' before ',' token
     154       32 | void F77_BLAS_MANGLE(sgesv, SGESV)(int*, int*, float*, int*, int*, float*, int*, int*);
```

This PR ensures that if ``+lapack`` is enabled, then ``+blas`` is also enabled. Now the spec ``+lapack`` is equivalent to ``+blas +lapack``, so the installation works as expected.

### DIFF
--- a/repos/spack_repo/builtin/packages/kokkos_kernels/package.py
+++ b/repos/spack_repo/builtin/packages/kokkos_kernels/package.py
@@ -187,7 +187,7 @@ class KokkosKernels(CMakePackage, CudaPackage):
         depends_on(spackname, when=f"+{tpl}")
 
     # lapack TPL depends on blas TPL
-    conflicts('+lapack', when='~blas')
+    conflicts("+lapack", when="~blas")
 
     patch("pr_2296_430.patch", when="@4.3.00:4.4.00")
     patch("pr_2296_400.patch", when="@4.0.00:4.2.01")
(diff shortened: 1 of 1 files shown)


tpadioleau (approved):
LGTM

tldahlgren (approved):

tldahlgren:
@brian-kelley @lucbv @srajama1 Any objections to merging this PR without further reviews?

brian-kelley:
@tldahlgren I think it's ready to go, but I would like Luc to see it before before merging.

lucbv (approved):
Yes, no problem with me.

PR #1839, opened 2025-10-02
libunwind: make m4 directory before autoreconf step

mwkrentel:
Aclocal installs files into m4, but the master and stable branches for libunwind don't include an m4 directory which causes aclocal to fail. This fixes the build for the master and stable branches.

The fgsl, httperf and libgdsii packages also do this.

### DIFF
--- a/repos/spack_repo/builtin/packages/libunwind/package.py
+++ b/repos/spack_repo/builtin/packages/libunwind/package.py
@@ -136,6 +136,11 @@ class Libunwind(AutotoolsPackage):
 
         return (wrapper_flags, None, flags)
 
+    # The master/stable branches don't have an m4 directory.
+    @run_before("autoreconf")
+    def make_m4_dir(self):
+        mkdirp("m4")
+
     def configure_args(self):
         spec = self.spec
         args = []


mwkrentel:
There must be something wonky inside autotools.

The master and stable branches of `libunwind` don't include the
autotools generated files and don't come with an `m4` directory, so
spack needs to run the `autoreconf` step for these versions.

The following creates an `m4` directory, copies files into it and succeeds.
There's a warning, but it creates the directory on demand.
```
$ cd libunwind
$ autoreconf -f -i
aclocal: warning: couldn't open directory 'm4': No such file or directory
libtoolize: putting auxiliary files in AC_CONFIG_AUX_DIR, 'config'.
libtoolize: copying file 'config/ltmain.sh'
libtoolize: putting macros in AC_CONFIG_MACRO_DIRS, 'm4'.
libtoolize: copying file 'm4/libtool.m4'
libtoolize: copying file 'm4/ltoptions.m4'
...
```

But spack runs the following which fails because somewhere inside
autoreconf, it's not creating the `m4` directory as needed.
```
$ autoreconf -f -i -I /home/krentel/unwind/install/linux-broadwell/libtool-2.4.7-nfe52udc3tzhzbrj4m2yllmxel55mptm/share/aclocal 
aclocal: error: couldn't open directory 'm4': No such file or directory
autoreconf: error: aclocal failed with exit status: 1
```

WTF!?  Adding the `-I dir` option causes autoreconf not to create the
`m4` directory on demand??  That makes no sense.

Anyway, the workaround is simple, just create the `m4` directory
before running `autoreconf`.

Three other spack packages: fgsl, httperf and libgdsii also do this.

I thought about suggesting that the spack autotools build system
should always do this, create the `m4` directory as needed before the
`autoreconf` step.

That would be a reasonable thing to do, and if someone wants to do
that, I'd support them.  But it's only 3 packages, now 4, so whatever.
I'm kinda surprised this isn't more common.

@alecbcs @manuelakuhn  Would you like to review?

mwkrentel:
@tldahlgren Would you like to review?   It's short and simple.
Thanks

scheibelp (approved):

PR #1840, opened 2025-10-03
Add a new SPR - Babeltrace2

minghangli-uni:
This PR introduces a new SPD receipe for [Babeltrace2](https://github.com/efficios/babeltrace). While there is an existing babeltrace package, Babeltrace2 is a major rewrite and differs significantly from verion 1 in API, build options, Python bindings. plugins, internal structures etc, ref: [Babeltrace2 doc](https://babeltrace.org/#:~:text=Babeltrace%C2%A02%20is%20a%20complete%20rewrite%20of%20the%20library%2C%20Python%20bindings%2C%20and%20CLI.%20It%20is%20plugin%20based%20and%20offers%20much%20more%20features%20and%20potential%20than%20Babeltrace%C2%A01.).

Because Babeltrace2 is not a drop-in replacement, this SPD does not override or replace the existing babeltrace package. They are intended to coexist side by side.

### DIFF
--- a/repos/spack_repo/builtin/packages/babeltrace2/package.py
+++ b/repos/spack_repo/builtin/packages/babeltrace2/package.py
@@ -1,0 +1,73 @@
+# Copyright Spack Project Developers. See COPYRIGHT file for details.
+#
+# SPDX-License-Identifier: (Apache-2.0 OR MIT)
+
+from os.path import join as pjoin
+
+from spack_repo.builtin.build_systems.autotools import AutotoolsPackage
+
+from spack.package import *
+
+
+class Babeltrace2(AutotoolsPackage):
+    """
+    Babeltrace 2: CTF trace processing library, CLI, and Python3 bindings (bt2).
+    """
+
+    homepage = "https://babeltrace.org/"
+    git = "https://github.com/efficios/babeltrace.git"
+    url = "https://github.com/efficios/babeltrace/archive/refs/tags/v2.1.2.tar.gz"
+
+    license("MIT")
+
+    version("2.1.2", sha256="92e261e1811f4a7f747ee4bc5ac21c0054cd7906f88ad799fab81f16e08c2122")
+    version("2.1.1", sha256="1ad8cea3fc31e8d18e869d0c01581318e534f3976a9f3636fdab4423f1113d47")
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/babeltrace2/package.py
```diff
@@ -0,0 +1,73 @@
...
+    version("2.1.0", sha256="808426a78efdeb1db71c71dba9ddb39902f7cc2604cee698849bc291d0af377e")
+    version("2.0.7", sha256="a8c1965590b0cf04b1985881b2a5727a0e49d628a99bdaade4f3e29f0c36e5b1")
+    version("2.0.6", sha256="a3a548b3e5b2a1d18c865299f083aa9716fcb6fed4a5713f29a87d3e9e39041c")
+    version("2.0.5", sha256="c058f9a2dc0b1287dcc20ef59fc0e13f54b9c43c14ffd58d5dc1c10786884a81")
+    version("2.0.4", sha256="37d3e81bc4abc42e4fa99488685eab3458212c6531ccadcf53cf1534d762bf22")
+    version("2.0.3", sha256="6cdeaa3bfc12d47936e7c664c5a2610c376ad3d2dfc8cf947137c4b3a2051dd3")
+    version("2.0.2", sha256="1c09428fec2d0000bf6f5332da2624b39fbf110477b82d2cb0856dcb74c58afc")
+
+    def url_for_version(self, v):
+        return f"https://github.com/efficios/babeltrace/archive/refs/tags/v{v}.tar.gz"
```
tldahlgren: Spack already handles this simple translation so you can remove this method.

(If you were to leave the method in, it should be implemented  AFTER all of the directives, so inserted around line 48.)

tldahlgren (changes requested):
Confirmed the homepage takes the same approach to distinguish between the two "versions" of the package.

Also confirmed the homepage and version sha256s (without the `url_for_version` method).

Please remove the unnecessary method implementation.

repos/spack_repo/builtin/packages/babeltrace2/package.py
minghangli-uni: Thanks @tldahlgren ! I've addressed your suggestion and also made other changes in 
6e26fe9

tldahlgren (approved):

PR #1845, opened 2025-10-03
qt: Patch MathExtras to remove ref to wtf_ceil on macOS

Aiden2244:
@johnwparent for review, please.


repos/spack_repo/builtin/packages/qt/package.py
```diff
@@ -195,6 +195,9 @@ class Qt(Package):
         when="@5.9.2: %gcc@14:",
     )
 
+    # Do not define `wtf_ceil()` in MathExtras.h on macOS
+    patch("qt515-mathextras.patch", when="@5.15.4:5 platform=darwin")
```
johnwparent:
```suggestion
    patch("qt515-mathextras.patch", when="@5.15.4:5.15 platform=darwin")
```

johnwparent (commented):
Just one small change for the version range. While I think your range would technically do what we want, I think adding the `.15` is a bit clearer as to the intent.

repos/spack_repo/builtin/packages/qt/package.py
```diff
@@ -195,6 +195,9 @@ class Qt(Package):
         when="@5.9.2: %gcc@14:",
     )
 
+    # Do not define `wtf_ceil()` in MathExtras.h on macOS
```
johnwparent: sorry, but on second thought, can we also expand this docstring a bit? basically just describe
What is does (you did that)
Why it does that
Why we need that patched
i.e. Prevents references to removed api, required to prevent compilation errors for webkit on macos, etc

Aiden2244: Feel free to review the verbage I used here!

zackgalbreath (approved):

PR #1850, opened 2025-10-03
AutotoolsPackage: add missing diffutils dependency

michaelkuhn:
While debugging a build failure in yaksa, I noticed that many, if not all, configure scripts require diffutils, e.g.:
```
checking if $SPACK/opt/spack/linux-skylake/compiler-wrapper-1.0-famqftdrjs5p3x2icdckr35xxszwhzhn/libexec/spack/gcc/gcc supports -fno-rtti -fno-exceptions... $SPACK_STAGE/spack-stage-yaksa-0.3-rwuygjddrlbs76e64zgg2fszdlvidl5b/spack-src/configure: line 9713: diff: command not found
```
However, in this case, the error was hidden:
```
checking whether C compiler accepts option -std=c11... no
```
This is obviously not true and caused the build failure. `config.log` contains:
```
configure:14539: checking whether C compiler accepts option -std=c11
configure:14594: $SPACK/opt/spack/linux-skylake/compiler-wrapper-1.0-famqftdrjs5p3x2icdckr35xxszwhzhn/libexec/spack/gcc/gcc -o conftest  -std=c11   conftest.c  > pac_test1.log 2>&1
configure:14594: $? = 0
configure:14629: $SPACK/opt/spack/linux-skylake/compiler-wrapper-1.0-famqftdrjs5p3x2icdckr35xxszwhzhn/libexec/spack/gcc/gcc -o conftest -std=c11  -std=c11   conftest.c  > pac_test2.log 2>&1
configure:14629: $? = 0
configure:14638: diff -b pac_test1.log pac_test2.log > pac_test.log
$SPACK_STAGE/spack-stage-yaksa-0.3-rwuygjddrlbs76e64zgg2fszdlvidl5b/spack-src/configure: line 14639: diff: command not found
configure:14641: $? = 127
configure: program exited with status 127
*** diff -b pac_test1.log pac_test2.log :
configure:14753: result: no
```
Adding diffutils as a build dependency required changing diffutils and libiconv (diffutil's only dependency) from AutotoolsPackage to Package, which is obviously a bit ugly since AutotoolsPackage provides quite a bit of convenience functionality.

This error happens in, e.g., Rocky Linux and Fedora containers since both do not contain `diff`.

### DIFF
--- a/repos/spack_repo/builtin/build_systems/autotools.py
+++ b/repos/spack_repo/builtin/build_systems/autotools.py
@@ -60,6 +60,7 @@ class AutotoolsPackage(PackageBase):
         depends_on("gnuconfig", type="build", when="target=aarch64:")
         depends_on("gnuconfig", type="build", when="target=riscv64:")
         depends_on("gmake", type="build")
+        depends_on("diffutils", type="build")
         conflicts("platform=windows")
 
     def flags_to_build_system_args(self, flags):
--- a/repos/spack_repo/builtin/packages/diffutils/package.py
+++ b/repos/spack_repo/builtin/packages/diffutils/package.py
@@ -4,13 +4,13 @@
 
 import re
 
-from spack_repo.builtin.build_systems.autotools import AutotoolsPackage
+from spack_repo.builtin.build_systems.generic import Package
 from spack_repo.builtin.build_systems.gnu import GNUMirrorPackage
 
 from spack.package import *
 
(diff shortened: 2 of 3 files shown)


michaelkuhn:
Marking as draft, since it causes test failures. I can fix those if this is deemed a valid way forward. Otherwise we should find another way to fix this.

PR #1851, opened 2025-10-03
mvapich: fix issue building against Slurm process manager

nhanford:
Hello @natshineman and @MatthewLieber ,

When building MVAPICH 4.1 against Slurm, I hit a configure error of the type:

```
'--with-pm=slurm' '--with-slurm=/usr' 'CFLAGS=-I/usr/include/slurm' '--with-device=ch4:ucx' '--with-file-system=lustre+nfs'

1 error found in build log:
     978    config.status: creating adio/include/romioconf.h
     979    config.status: executing depfiles commands
     980    config.status: executing libtool commands
     981    configure: ===== done with src/mpi/romio configure =====
     982    configure: sourcing src/mpi/romio/localdefs
     983    configure: WARNING: PM slurm is unknown
  >> 984    configure: error: Aborting configure because an error was seen in the selection of process managers
```

which I tracked down to the following lines (35599-35631) in `configure` for both version 4.0 and 4.1:

```bash
for pm_name in $pm_names ; do
    if test -z "$first_pm_name" ; then
       first_pm_name=$pm_name
       export first_pm_name
    fi

    case "$pm_name" in
      gforker)
        build_gforker=yes
        if test "$with_pmi" = "default" -o "$with_pmi" = "pmi1"; then
            with_pmi=pmi1
        else
            as_fn_error $? "$pm_name requires PMI-v1, but $with_pmi was chosen." "$LINENO" 5
        fi
        ;;
      remshell)
        build_refshell=yes
        if test "$with_pmi" = "default" -o "$with_pmi" = "pmi1"; then
            with_pmi=pmi1
        else
            as_fn_error $? "$pm_name requires PMI-v1, but $with_pmi was chosen." "$LINENO" 5
        fi
        ;;
      hydra)
        subsystems="$subsystems src/pm/hydra"
        ;;
      *)
        { $as_echo "$as_me:${as_lineno-$LINENO}: WARNING: PM $pm_name is unknown" >&5
$as_echo "$as_me: WARNING: PM $pm_name is unknown" >&2;}
	hasError=yes
        ;;
    esac
done
```

This error case is not present in version 3.0, which contains logic to handle the case where `--with-pm=slurm` is set. Therefore, I removed the erroneous setting for all but version 3 and confirmed the regression goes away for version 4.1. Please let me know what you think.

Thanks,
Nate

### DIFF
--- a/repos/spack_repo/builtin/packages/wgrib2/package.py
+++ b/repos/spack_repo/builtin/packages/wgrib2/package.py
@@ -179,7 +179,9 @@ class Wgrib2(MakefilePackage, CMakePackage):
     depends_on("ip@5.1:", when="@3.5: +ipolates")
     depends_on("lapack", when="@3.5: +ipolates")
     depends_on("libaec@1.0.6:", when="@3.2: +aec")
-    depends_on("netcdf-c", when="@3.2: +netcdf4")
+    # Options to use netcdf3 or netcdf4 merged into a single option with v3.4.0
+    depends_on("netcdf-c", when="@3.4: +netcdf")
+    depends_on("netcdf-c", when="@3.2:3.3 +netcdf4")
     depends_on("jasper@:2", when="@3.2:3.4 +jasper")
     depends_on("g2c", when="@3.5: +jasper")
     depends_on("zlib-api", when="@3.2: +png")


natshineman:
@nhanford thanks for catching this. 

Yes, we no longer support the `--with-pm=slurm` option since it was causing confusion with users. To use srun as the process manager, we use `--with-pm=none` now. That said, slurm typically stores it's PMI headers in a non-standard location, so we may need to support that. Let me get a version of this up that correctly handles the 4.0 use case.

PR #1856, opened 2025-10-06
ESMF: workaround for legacy linking to MPI C++ interface

climbfuji:
## Description

`repos/spack_repo/builtin/packages/esmf/package.py`: workaround for legacy linking to MPI C++ interface.

A fix for this will be made upstream for newer versions of ESMF (release/beta snapshot unknown). The `filter_file` function doesn't cause any harm if the search string can't be found, therefore not applying any version bounds to this patch.

## Issues

See https://github.com/esmf-org/esmf/issues/497.

## Testing

Tested on Navy DSRC Nautilus system that has the MPI C++ bindings installed.


theurich:
A fix for this has been implemented upstream with https://github.com/esmf-org/esmf/releases/tag/v9.0.0b04 and later. 

As @climbfuji points out, this spack-packages PR is safe independent of ESMF version (with or without the upstream fix), therefore good to have!

theurich (approved):
Looks good. Thank you!

tldahlgren (approved):

PR #1857, opened 2025-10-06
Fix Cabana configure error in CI

zackgalbreath:
After #1382 was merged, Cabana began failing to find Kokkos during the configure step.

Fix this by making Cabana unconditionally depend on Kokkos.

### DIFF
--- a/repos/spack_repo/builtin/packages/cabana/package.py
+++ b/repos/spack_repo/builtin/packages/cabana/package.py
@@ -55,6 +55,8 @@ class Cabana(CMakePackage, CudaPackage, ROCmPackage):
     depends_on("cmake@3.16:", type="build", when="@0.5.0:")
 
     depends_on("googletest", type="build", when="+testing")
+
+    depends_on("kokkos")
     _versions = {"0.3:": "@3.1:", "0.4:": "@3.2:", "0.6:": "@3.7:"}
     for _version in _versions:
         _kk_version = _versions[_version]


eugeneswalker (dismissed):
Thanks!

stacks/e4s-cray-rhel/spack.yaml
```diff
@@ -98,7 +98,7 @@ spack:
   - arborx
   - argobots
   - boost +python +filesystem +iostreams +system
-  - cabana
+  - cabana +openmp
```
kwryankrattiger: Constrain this at the packages config level for kokkos with for all of these stacks. Cabana will detect this correctly then.

```
kokkos:
  prefer: ["+serial +openmp"]
```

kwryankrattiger (commented):
This also needs a fix in the cabana recipe to always depend on kokkos. Kokkos enforces the requirement of at least one backend being specified, so the constraint in the stacks should be on kokkos instead of cabana.

```
depends_on("kokkos")
```

repos/spack_repo/builtin/packages/cabana/package.py
```diff
@@ -51,6 +51,8 @@ class Cabana(CMakePackage, CudaPackage, ROCmPackage):
     depends_on("c", type="build", when="+mpi")
     depends_on("cxx", type="build")
 
+    depends_on("kokkos")
```
kwryankrattiger: small nit, but can you put this right above the loops for the kokkos variant propagation to keep them together

eugeneswalker (approved):

PR #1860, opened 2025-10-06
py-tokenizers: add v0.22.1

qwertos:


repos/spack_repo/builtin/packages/py_tokenizers/package.py
```diff
@@ -24,7 +25,8 @@ class PyTokenizers(PythonPackage):
 
     # TODO: This package currently requires internet access to install.
```
tldahlgren, line 28: FYI. The archive from `PyPi.org` indicates the minimum python for the new version is `python@3.9:`.

repos/spack_repo/builtin/packages/py_tokenizers/package.py
```diff
@@ -24,7 +25,8 @@ class PyTokenizers(PythonPackage):
...
     depends_on("py-maturin@1", when="@0.14:", type="build")
```
tldahlgren: FWIW. Based on prior comments from Adam, I'm ignoring the upper bound for `py-maturin` in the `pyproject.toml` file.

tldahlgren (changes requested):
Confirmed the version sha256. 

Note the `pyproject.toml` downloaded from PyPi has a lower bound for `python` still listed as a safe version. Since someone may already have a version of `python` installed, I would think it would be prudent to include the lower bound.

tldahlgren (approved):

PR #1863, opened 2025-10-06
py-peft: new package

qwertos:


wdconinc (approved):
LGTM. I'll rebase to make sure this passes current CI checks (no reason it won't).

PR #1864, opened 2025-10-06
mvapich: remove cuda variant

nhanford:
Hello @natshineman and @MatthewLieber,

Initially this PR was designed to shore up GPU support in the MVAPICH non-Plus package. However, in the conversation below, it was determined that GPU support should not be built in this package and users should instead be directed to MVAPICH-Plus. Therefore, this PR now removes the erroneous CUDA variant.

Thanks,
Nate

### DIFF
--- a/repos/spack_repo/builtin/packages/beatnik/package.py
+++ b/repos/spack_repo/builtin/packages/beatnik/package.py
@@ -35,8 +35,8 @@ class Beatnik(CMakePackage, CudaPackage, ROCmPackage):
     depends_on("mpi")
     with when("+cuda"):
         depends_on("mpich +cuda", when="^[virtuals=mpi] mpich")
-        depends_on("mvapich +cuda", when="^[virtuals=mpi] mvapich")
-        depends_on("mvapich2 +cuda", when="^[virtuals=mpi] mvapich2")
+        depends_on("mvapich-plus +cuda", when="^[virtuals=mpi] mvapich")
+        depends_on("mvapich2-gdr +cuda", when="^[virtuals=mpi] mvapich2-gdr")
         depends_on("openmpi +cuda", when="^[virtuals=mpi] openmpi")
 
     with when("+rocm"):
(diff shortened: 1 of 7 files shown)


nhanford:
@spackbot fix style

natshineman:
Hi Nate, I think we need to check with Dr Panda, on this one. Officially MVAPICH does not support GPUs in the open source version. Technically we inherit whatever GPU support MPICH provides, but the official position is that GPU users should use MVAPICH-Plus and not MVAPICH. Let me see how we want to handle this so we don't confuse users.

nhanford:
> Hi Nate, I think we need to check with Dr Panda, on this one. Officially MVAPICH does not support GPUs in the open source version. Technically we inherit whatever GPU support MPICH provides, but the official position is that GPU users should use MVAPICH-Plus and not MVAPICH. Let me see how we want to handle this so we don't confuse users.

Okay in that case then I can modify this PR to ditch the `+cuda` variant that was already present in the package as well.

nhanford:
@natshineman It looks like you'll have to talk to the maintainers of [beatnik](https://github.com/spack/spack-packages/blob/develop/repos/spack_repo/builtin/packages/beatnik/package.py) and [sphexa](https://github.com/spack/spack-packages/blob/develop/repos/spack_repo/builtin/packages/sphexa/package.py) about a license for MVAPICH-Plus as these packages are causing the audit to fail.

```bash
spack pkg grep "mvapich +cuda"
```

Thomas-Ulrich (approved):
ok for tandem and seissol.

repos/spack_repo/builtin/packages/beatnik/package.py
```diff
@@ -35,12 +35,13 @@ class Beatnik(CMakePackage, CudaPackage, ROCmPackage):
     depends_on("mpi")
     with when("+cuda"):
         depends_on("mpich +cuda", when="^[virtuals=mpi] mpich")
-        depends_on("mvapich +cuda", when="^[virtuals=mpi] mvapich")
-        depends_on("mvapich2 +cuda", when="^[virtuals=mpi] mvapich2")
+        depends_on("mvapich-plus +cuda", when="^[virtuals=mpi] mvapich")
```
MatthewLieber: should this be mvapich-plus at the end?

nhanford: Yes it should--good catch, thanks.

MatthewLieber (approved):

patrickb314:
Beatnik package change is good.

alalazo (approved):

PR #1868, opened 2025-10-07
libtirpc: Fix macOS build of 1.3.7

paulgessinger:
Add patch to enable building 1.3.7 on macOS.

### DIFF
--- a/repos/spack_repo/builtin/packages/libtirpc/package.py
+++ b/repos/spack_repo/builtin/packages/libtirpc/package.py
@@ -26,6 +26,20 @@ class Libtirpc(AutotoolsPackage):
 
     provides("rpc")
 
+    # Allow build on macOS:
+    # - Adds a preprocessor define to make IPV6 defined available under expected names
+    # - Fixes configure template to correctly detect absence of `gss_pname_to_uid`
+    # - Avoids linking with shared cache
+    with when("@1.3.7 platform=darwin"):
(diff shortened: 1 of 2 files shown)


repos/spack_repo/builtin/packages/libtirpc/package.py
```diff
@@ -26,6 +26,20 @@ class Libtirpc(AutotoolsPackage):
...
+    # - Avoids linking with shared cache
+    with when("@1.3.7 platform=darwin"):
+        patch("libtirpc-fix-ipv6-gss-macos-1.3.7.patch")
+        depends_on("autoconf", type="build")
+        depends_on("automake", type="build")
+        depends_on("libtool", type="build")
+
+    @property
+    def force_autoreconf(self):
+        return self.version == Version("1.3.7")
```
alalazo:
```suggestion
    @property
    def force_autoreconf(self):
        return self.spec.satisfies("@1.3.7 platform=darwin")
```

alalazo (changes requested):
See failures in pipelines

repos/spack_repo/builtin/packages/libtirpc/package.py
paulgessinger: I think I grabbed this from the documentation somewhere, but I'm happy to change it.

paulgessinger:
@alalazo pipeline is green now.

alalazo (approved):

PR #1869, opened 2025-10-07
lhapdfsets: Fix index, allow explicit list of pdfs

paulgessinger:
This PR:

- Sets the empirical arbitrary version to `6.5.5`
- Allows explicitly passing the set of pdfs to install as a multi value variant. The existing sentinel values `all` and `default` should still work
- Updates `pdfsets.index` which is a seemingly undocumented file I grabbed from their website.


repos/spack_repo/builtin/packages/lhapdfsets/package.py
```diff
@@ -32,8 +32,8 @@ class Lhapdfsets(BundlePackage):
 
     variant(
         "sets",
-        description="Individiual lhapdf sets to install",
-        values=("all", "default"),
+        description="Individiual lhapdf sets to install (all, default, or comma-separated list)",
+        multi=True,
```
wdconinc, line 36: I think we want to use the disjoint sets functionality to add all and none. That then disallows the case where someone specifies all,cteq5 and it allows all to satisfy requirements for cteq5.

repos/spack_repo/builtin/packages/lhapdfsets/package.py
```diff
@@ -19,7 +19,7 @@ class Lhapdfsets(BundlePackage):
 
     maintainers("vvolkl", "wdconinc")
 
-    version("6.3.0")
+    version("6.5.5")
```
wdconinc, line 22: If the version is primarily based on the date when the index was downloaded, should we suffix that date?

repos/spack_repo/builtin/packages/lhapdfsets/package.py
wdconinc: I think that's what fastjet does?

repos/spack_repo/builtin/packages/lhapdfsets/package.py
paulgessinger: I honestly don't understand what this version does. The PDFs themselves are not versioned, but if this number doesn't line up with the version number in `lhapdf` is doesn't seem to work.

repos/spack_repo/builtin/packages/lhapdfsets/package.py
paulgessinger: But that requires me to list all PDFs, right? If you look at the `pdfsets.index` thats over 1500, and they change them arbitrarily without notice.

paulgessinger:
@wdconinc any further thoughts?

paulgessinger:
@wdconinc can we proceed with this?

wdconinc (approved):
Fair point on the number of sets. Looks good to me.

PR #1891, opened 2025-10-08
sentieon-genomics(GENESIS): add 2.0.x, 2.1.x, tests, fix for A64FX 

chig:
The initial description was empty, updated by @bernhardkaindl:

# This PR
- updates the GENESIS package recipe to the 2.0.x and 2.1.x release series
- switches the package URL to the current upstream source, and
- adds regression-test support.
- also includes patching needed for CUDA/NVTX on Arm platforms and for Fujitsu compiler builds on A64FX systems.
----
[alalazo](https://github.com/alalazo) / @chig, these are comments from GitHub Copilot from review of the PR on 25.6.2006:

# Latest Updated Review Summary (25.6.2006)

GitHub Copilot checked the earlier changes requested by @alalazo, and they appear to be applied:

- old versions/resources were removed
- obsolete patch directives/files were removed
- the obsolete `@master` condition was removed
- the `--host=Fugaku` logic was narrowed to the `a64fx` target
- explanatory comments were added above the patch directives

It also did another pass over the current diff and found one remaining risk worth addressing before merge:

1. `force_autoreconf = True` is now unconditional, and `autoconf`, `automake`, `libtool`, and `m4` are unconditional build dependencies for all builds.

Previously, autoreconf and those autotools dependencies were only needed for the patched Fujitsu-specific cases. Making them unconditional broadens the build behavior for every version and compiler, not just `%fj`. That means this PR is no longer only adding versions / updating URLs / `%fj` patch handling; it also changes the default build path for non-Fujitsu builds.

Copilot thinks this should either:
- be limited back to the cases that actually patch autotools input files, or
- be justified explicitly if all supported versions now require autoreconf.

Other than that, the earlier requested changes from @alalazo look applied to me.
----
# Detailed review:

Q: Can you check if the changes requested by @alalazo by reviewing with changes-requested status have been applied? Also, can you review the PR for further issues or risks and add a comment to the PR with your results?

----
The specific changes requested by @alalazo in the earlier changes-requested reviews appear to have been applied, but GitHub Copilot also found one substantive remaining risk in the current PR diff.

Checked:

- @alalazo requested removal of old versions/resources: applied.
- @alalazo requested removal of obsolete patch directives/files: applied.
- @alalazo requested fixing the obsolete @master condition: applied.
- @alalazo requested keeping the --host=Fugaku logic only for a64fx: applied by changing the condition to spec.target == "a64fx".
- @alalazo later requested comments above patch directives explaining effect/need/source: partially applied. The PR now has explanatory comments for the patch directives, but I did not verify provenance/source text for each patch beyond what is visible in the current diff.

Further review result:

- I found a likely regression risk in package.py: the PR sets force_autoreconf = True unconditionally and makes autoconf, automake, libtool, and m4 unconditional build dependencies for all versions. Previously those were limited to the Fujitsu-patched builds / @master path. Since the main purpose of the PR is adding packaged releases and patching Fujitsu builds, broadening autoreconf to every build changes behavior for non-Fujitsu builds too. That may be intentional, but it is a materially wider change than the review thread established and could introduce avoidable build risk.

Why this matters:

- Running autoreconf on every build can change the generated configure machinery for versions that previously built from shipped release tarballs without regeneration.
- Making autotools dependencies unconditional increases the dependency surface for all users, not only the %fj cases that seem to require patching configure.ac.


chig:
@spack/spack-releasers 
This PR is still relevant and useful for our HPC application (molecular dynamics software, GENESIS).
I would appreciate a review when possible.

repos/spack_repo/builtin/packages/genesis/package.py
```diff
@@ -15,39 +15,28 @@ class Genesis(AutotoolsPackage, CudaPackage):
...
-        placement="doc",
-    )
-    resource(
-        when="@1.5.1",
-        name="user_guide",
-        url="https://www.r-ccs.riken.jp/labs/cbrt/wp-content/uploads/2019/10/GENESIS-1.4.0.pdf",
-        sha256="da2c3f8bfa1e93adb992d3cfce09fb45d8d447a94f9a4f884ac834ea7279b9c7",
-        expand=False,
-        placement="doc",
-    )
```
alalazo, line 50: Usually we deprecate old versions first, but I think in this case we can directly remove them.

repos/spack_repo/builtin/packages/genesis/package.py
alalazo: These two patch directives (and the corresponding files) can be removed.

alalazo:
```suggestion
```

alalazo:
```suggestion
```

alalazo:
```suggestion
```

alalazo, line 77:
```suggestion
            options.append("--with-cuda=%s" % self.spec["cuda"].prefix)
```

alalazo:
```suggestion
```

alalazo, line 69:
```suggestion
```

alalazo (changes requested):

repos/spack_repo/builtin/packages/genesis/package.py
chig: I agree with you.

repos/spack_repo/builtin/packages/genesis/package.py
chig: I agree with you.

chig: I agree with you.

chig: I agree with you.

chig: I agree with you.

chig: Thanks for the review. I addressed the requested changes.

I also kept `--host=Fugaku` for A64FX targets, since it is still needed there, but generalized the condition to `spec.target == "a64fx"`.

chig:
Thank you very much for your suggestion. I've updated the recipe.

repos/spack_repo/builtin/packages/genesis/package.py
```diff
@@ -120,3 +89,17 @@ def cache_test_sources(self):
...
+            test_name = join_path(
+                    self.install_test_root, "tests", "regression_test", "test.py"
+                    )
+            bin_name = join_path(self.prefix.bin, "spdyn")
+            opts = [
+            test_name,
+            self.spec["mpi"].prefix.bin.mpirun + " -np 8 " + bin_name,
+            ]
+            env["OMP_NUM_THREADS"] = "1"
+            self.run_test(exe_name, options=opts, expected="Passed  61 / 61")
```
alalazo: Why did we add this method?

repos/spack_repo/builtin/packages/genesis/package.py
```diff
@@ -15,83 +15,52 @@ class Genesis(AutotoolsPackage, CudaPackage):
...
-    depends_on("fortran", type="build")  # generated
-
-    depends_on("autoconf", type="build", when="@1.5.1 %fj")
-    depends_on("autoconf", type="build", when="@master")
-    depends_on("automake", type="build", when="@1.5.1 %fj")
-    depends_on("automake", type="build", when="@master")
-    depends_on("libtool", type="build", when="@1.5.1 %fj")
-    depends_on("libtool", type="build", when="@master")
-    depends_on("m4", type="build", when="@1.5.1 %fj")
-    depends_on("m4", type="build", when="@master")
```
alalazo, line 68: We probably removed too much now

repos/spack_repo/builtin/packages/genesis/package.py
```diff
@@ -120,3 +89,17 @@ def cache_test_sources(self):
         """Copy test files after the package is installed for test()."""
         if self.spec.satisfies("@master"):
             cache_extra_test_sources(self, ["tests"])
+
+    def test(self):
+        if self.spec.satisfies("@master"):
```
alalazo: There is no `@master` version declared

alalazo (changes requested):

repos/spack_repo/builtin/packages/genesis/package.py
chig: We use 8 MPI ranks because it allows us to run the largest number of test cases. However, this is not a strict requirement.

repos/spack_repo/builtin/packages/genesis/package.py
chig: Thanks for pointing that out. I will restore the necessary parts and refine the changes to avoid removing required functionality.

repos/spack_repo/builtin/packages/genesis/package.py
chig: I'll update.

repos/spack_repo/builtin/packages/genesis/package.py
chig: There seems to be some uncertainty about how tests should be added for this package in Spack, so I kept the test configuration simple. We use 8 MPI ranks because it allows us to cover the largest number of test cases, but this is not a strict requirement.

chig:
Thanks for the review. I believe I have addressed the requested changes:

* removed the old versions/resources
* removed the obsolete patch directives
* fixed the `@master` condition in the test logic
* kept `--host=Fugaku` only for the `a64fx` target

I’ve requested review again. Thank you for taking another look.

chig:
@alalazo I believe I have addressed the remaining requested change by removing the obsolete `@master` condition. Could you please take another look when you have time?

Also, the workflow is currently awaiting maintainer approval, so I would appreciate it if the CI could be approved to run.

repos/spack_repo/builtin/packages/genesis/package.py
```diff
@@ -15,83 +15,62 @@ class Genesis(AutotoolsPackage, CudaPackage):
...
 
     variant("openmp", default=True, description="Enable OpenMP.")
     variant("single", default=False, description="Enable single precision.")
+    variant("mixed",  default=False, description="Enable mixed precision.", when="@2.0.0:")
     variant("hmdisk", default=False, description="Enable huge molecule on hard disk.")
+    
+    patch("fix-nvtx-include.patch", when="+cuda")
+    patch("fix-nvtx-include.patch", when="+gpu")
+    patch("fj_compiler_2.0.0.patch", when="@2.0.0:2.1.3 %fj")
+    patch("fj_compiler_2.1.4.patch", when="@2.1.4: %fj")
```
alalazo, line 46: Can you add a brief comment above the `patch` directive explaining:
2. Why it is needed
3. Where it was taken from

?

alalazo (commented):
Just one small comment. I'll let CI run in the meanwhile

repos/spack_repo/builtin/packages/genesis/package.py
chig: Thank you for reviewing the pull request. I added the simple comments you pointed out, and also removed unnecessary spaces that were causing errors in the CI tool.

chig:
I've updated package.py using style fix.

bernhardkaindl (dismissed):
@alalazo, I guess your GitHub-requested changes were applied,

PR would be ready to merge?

Checked: sentieon-genomics@202503.03 installs on Linux

bernhardkaindl (commented):
[alalazo](https://github.com/alalazo) / @chig, these are comments from GitHub Copilot from review of the PR on 25.6.2006:

# Latest Updated Review Summary (25.6.2006)

GitHub Copilot checked the earlier changes requested by @alalazo, and they appear to be applied:

- old versions/resources were removed
- obsolete patch directives/files were removed
- the obsolete `@master` condition was removed
- the `--host=Fugaku` logic was narrowed to the `a64fx` target
- explanatory comments were added above the patch directives

It also did another pass over the current diff and found one remaining risk worth addressing before merge:

1. `force_autoreconf = True` is now unconditional, and `autoconf`, `automake`, `libtool`, and `m4` are unconditional build dependencies for all builds.

Previously, autoreconf and those autotools dependencies were only needed for the patched Fujitsu-specific cases. Making them unconditional broadens the build behavior for every version and compiler, not just `%fj`. That means this PR is no longer only adding versions / updating URLs / `%fj` patch handling; it also changes the default build path for non-Fujitsu builds.

Copilot thinks this should either:
- be limited back to the cases that actually patch autotools input files, or
- be justified explicitly if all supported versions now require autoreconf.

Other than that, the earlier requested changes from @alalazo look applied to me.
----
# Detailed review:

Q: Can you check if the changes requested by @alalazo by reviewing with changes-requested status have been applied? Also, can you review the PR for further issues or risks and add a comment to the PR with your results?

----
The specific changes requested by @alalazo in the earlier changes-requested reviews appear to have been applied, but GitHub Copilot also found one substantive remaining risk in the current PR diff.

What I checked:

    @alalazo requested removal of old versions/resources: applied.
    @alalazo requested removal of obsolete patch directives/files: applied.
    @alalazo requested fixing the obsolete @master condition: applied.
    @alalazo requested keeping the --host=Fugaku logic only for a64fx: applied by changing the condition to spec.target == "a64fx".
    @alalazo later requested comments above patch directives explaining effect/need/source: partially applied. The PR now has explanatory comments for the patch directives, but I did not verify provenance/source text for each patch beyond what is visible in the current diff.

Further review result:

    I found a likely regression risk in package.py: the PR sets force_autoreconf = True unconditionally and makes autoconf, automake, libtool, and m4 unconditional build dependencies for all versions. Previously those were limited to the Fujitsu-patched builds / @master path. Since the main purpose of the PR is adding packaged releases and patching Fujitsu builds, broadening autoreconf to every build changes behavior for non-Fujitsu builds too. That may be intentional, but it is a materially wider change than the review thread established and could introduce avoidable build risk.

Why this matters:

    Running autoreconf on every build can change the generated configure machinery for versions that previously built from shipped release tarballs without regeneration.
    Making autotools dependencies unconditional increases the dependency surface for all users, not only the %fj cases that seem to require patching configure.ac.

chig:
I rebased the branch on the current develop branch and added a comment explaining why autoreconf is required unconditionally for GENESIS GitHub source archives.

repos/spack_repo/builtin/packages/genesis/package.py
```diff
@@ -1,5 +1,4 @@
-# Copyright Spack Project Developers. See COPYRIGHT file for details.
-#
+# Copyright Spack Project Developers. See COPYRIGHT file for details.  #
```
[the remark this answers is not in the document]
chig: I update the header.

repos/spack_repo/builtin/packages/genesis/package.py
```diff
@@ -15,83 +14,70 @@ class Genesis(AutotoolsPackage, CudaPackage):
...
     variant("single", default=False, description="Enable single precision.")
+    variant("mixed", default=False, description="Enable mixed precision.", when="@2.0.0:")
     variant("hmdisk", default=False, description="Enable huge molecule on hard disk.")
 
-    conflicts("%apple-clang", when="+openmp")
+    # Fix NVTX include path for CUDA 12 on Arm sbsa-linux platforms
+    # (e.g., GH200). nvToolsExt.h is located under
+    # targets/sbsa-linux/include/nvtx3.
+    patch("fix-nvtx-include.patch", when="+cuda")
+    patch("fix-nvtx-include.patch", when="+gpu")
```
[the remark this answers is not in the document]
chig: I remove the pacth line.

repos/spack_repo/builtin/packages/genesis/package.py
```diff
@@ -15,83 +14,70 @@ class Genesis(AutotoolsPackage, CudaPackage):
...
+        options.extend(self.enable_or_disable("mixed"))
         options.extend(self.enable_or_disable("hmdisk"))
         if spec.satisfies("+cuda"):
             options.append("--enable-gpu")
             options.append("--with-cuda=%s" % spec["cuda"].prefix)
         else:
             options.append("--disable-gpu")
-        if spec.target == "a64fx" and self.spec.satisfies("@master %fj"):
+        if spec.target == "a64fx":
             options.append("--host=Fugaku")
```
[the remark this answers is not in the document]
chig: I fix it.

repos/spack_repo/builtin/packages/genesis/package.py
```diff
@@ -15,83 +14,70 @@ class Genesis(AutotoolsPackage, CudaPackage):
...
     depends_on("lapack")
+    depends_on("c", type="build")  # generated
+    depends_on("fortran", type="build")  # generated
 
-    patch("fj_compiler.patch", when="@master %fj")
-    patch("fj_compiler_1.5.1.patch", when="@1.5.1 %fj")
+    depends_on("autoconf", type="build")
+    depends_on("automake", type="build")
+    depends_on("libtool", type="build")
+    depends_on("m4", type="build")
```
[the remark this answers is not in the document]
chig: This is intentional.

GENESIS GitHub source archives are not generated with make dist, so the bundled autotools-generated files are not guaranteed to be consistent with the current configure.ac and Makefile.am.

For this reason, force_autoreconf = True is applied unconditionally. I also added a comment in the package recipe explaining this rationale.

repos/spack_repo/builtin/packages/genesis/package.py
bernhardkaindl: seems to have been resolved.

repos/spack_repo/builtin/packages/genesis/package.py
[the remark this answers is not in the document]
bernhardkaindl: ack, is applied.

repos/spack_repo/builtin/packages/genesis/package.py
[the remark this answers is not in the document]
bernhardkaindl: ack.

repos/spack_repo/builtin/packages/genesis/package.py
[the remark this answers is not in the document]
bernhardkaindl: ack, is applied.

bernhardkaindl:
@alalazo, after another review with other good fixes, I think this PR is ready, your requested changes are also applied, AFAICS.

alalazo (approved):

PR #1894, opened 2025-10-08
py-soupsieve: add v2.8

manuelakuhn:
https://github.com/facelessuser/soupsieve/tree/2.8

### DIFF
--- a/repos/spack_repo/builtin/packages/py_soupsieve/package.py
+++ b/repos/spack_repo/builtin/packages/py_soupsieve/package.py
@@ -18,6 +18,7 @@ class PySoupsieve(PythonPackage):
     # Circular dependency on beautifulsoup4
     skip_modules = ["soupsieve"]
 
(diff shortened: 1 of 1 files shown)


alalazo (commented):
Can we deprecate old versions, so we could eventually get rid of the historical dependencies?

manuelakuhn:
If you want you can do that. I am in general not a fan to remove recipes for versions. 
I think having also recipes for older versions, even if not supported, is a feature of spack (at least it was one of the reasons for me back then to start with spack).

manuelakuhn:
@alalazo can you take another look, please?

alalazo (approved):

alalazo:
> I am in general not a fan to remove recipes for versions.

That's fine here. As a general rule we want to deprecate old versions and eventually remove them, since there is a trade-off between how many versions of a package we support, and the maintenance cost of the recipe / speed of the concretizer.

PR #1899, opened 2025-10-08
precice: add v3.3.0

fsimonis:
This PR adds preCICE version 3.3.0, which is the target release for the next xSDK release.

FYI @balay


MakisH (approved):

MakisH (approved):
Checksum is correct, and the dependency ranges have already been updated.

alalazo (approved):

PR #1900, opened 2025-10-08
IMOD: new package

Markus92:
IMOD is a set of image processing, modeling and display programs used for tomographic reconstruction and for 3D reconstruction of EM serial sections and optical sections. It's a dependency of Relion (and associted py-yet-another-imod-wrapper, which I'll submit later).

IMOD is quite open about what you can all build it with, but it makes a lot of assumptions on the  environment it's running in with some custom tcsh scripts, it doesn't use anything like Autotools or CMake unfortunately. I think with the few changes I make I cut through most of the stuff (together with Spack managing compiler flags itself very carefully), but please ping me if there's any issues.


Markus92:
Merged in #2054

PR #1901, opened 2025-10-08
py-statsmodels: add v0.14.5

manuelakuhn:
https://github.com/statsmodels/statsmodels/tree/v0.14.5

### DIFF
--- a/repos/spack_repo/builtin/packages/py_statsmodels/package.py
+++ b/repos/spack_repo/builtin/packages/py_statsmodels/package.py
@@ -21,17 +21,19 @@ class PyStatsmodels(PythonPackage):
 
     license("BSD-3-Clause")
 
+    version("0.14.5", sha256="de260e58cccfd2ceddf835b55a357233d6ca853a1aa4f90f7553a52cc71c6ddf")
     version("0.14.1", sha256="2260efdc1ef89f39c670a0bd8151b1d0843567781bcafec6cda0534eb47a94f6")
 
     depends_on("c", type="build")  # generated
     depends_on("fortran", type="build")  # generated
 
-    depends_on("python", type=("build", "link", "run"))
+    depends_on("python@3.9:", when="@0.14.2:", type=("build", "link", "run"))
+    depends_on("python@3.8:", type=("build", "link", "run"))
 
-    depends_on("py-setuptools@69.0.2:", when="@0.14.1: ^python@3.12:", type="build")
-    depends_on("py-setuptools@63.4.3:", when="@0.14.1:", type="build")
-
-    # pyproject.toml
+    depends_on("py-setuptools@69.0.2:", when="@0.14.1:", type="build")
+    depends_on("py-setuptools@69.0.2:", when="@0.14.1 ^python@3.12:", type="build")
+    depends_on("py-setuptools@63.4.3:", when="@0.14.1", type="build")
+    depends_on("py-cython@3.0.10:3", when="@0.14.2:", type="build")
     depends_on("py-cython@0.29.33:3", when="@0.14.1", type="build")
     depends_on("py-setuptools-scm+toml@8", when="@0.14.1:", type="build")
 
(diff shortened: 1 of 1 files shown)


tldahlgren:
Made sure maintainer in the Reviewers list.

tldahlgren (approved):
Confirmed the version sha56.

Will give @climbfuji time to review.

repos/spack_repo/builtin/packages/py_statsmodels/package.py
```diff
@@ -40,17 +42,20 @@ class PyStatsmodels(PythonPackage):
     #    https://github.com/pydata/patsy/pull/131
 
     # requirements.txt
-    depends_on("py-numpy@1.22.3:1", when="@0.14.1:", type=("build", "link", "run"))
-    # https://github.com/statsmodels/statsmodels/issues/9194
-    depends_on("py-numpy@:1", when="@:0.14.1", type=("build", "link", "run"))
-    depends_on("py-scipy@1.4:", when="@0.13.5:", type=("build", "run"))
-    conflicts("^py-scipy@1.9.2", when="@:0.14.1")
+    depends_on("py-numpy@1.22.3:2", when="@0.14.3:", type=("build", "link", "run"))
+    depends_on("py-numpy@1.22.3:1", when="@0.14.1", type=("build", "link", "run"))
```
climbfuji, line 46: What about 0.14.2 ?

repos/spack_repo/builtin/packages/py_statsmodels/package.py
```diff
@@ -40,17 +42,20 @@ class PyStatsmodels(PythonPackage):
...
+    depends_on("py-scipy@1.8:", when="@0.14.2:", type=("build", "run"))
```
tldahlgren, line 47: Which take precedence? The lower bound appears higher `@1.13:` for the new version in `pyproject.toml` (see https://github.com/statsmodels/statsmodels/blob/1107ea567121b80f90562d6c72085eed7e882113/pyproject.toml#L8) but what you have here in at https://github.com/statsmodels/statsmodels/blob/1107ea567121b80f90562d6c72085eed7e882113/requirements.txt#L3).

tldahlgren (commented):
Confirmed the version sha256 but have a question about one of the dependencies (see comment).

repos/spack_repo/builtin/packages/py_statsmodels/package.py
manuelakuhn: I took the versions from the [`requirments.txt](https://github.com/statsmodels/statsmodels/blob/v0.14.5/requirements.txt) and the first to have `numpy>=1.22.3,<3` was `0.14.3` (0.14.2 has no upper bound). Because `0.14.2` is not in spack I did not bother with it. I can also extend it to `0.14.2` if you want, but there is still the open question from @tldahlgren (see below)

repos/spack_repo/builtin/packages/py_statsmodels/package.py
manuelakuhn: Yes, I only went with the versions from `requirements.txt` and did not bother with the `py-numpy` and `py-scipy` versions from `pyproject.toml` for two reasons:
- the `pyproject.toml` ones are only for build and not run
- that's how it was handled in the recipe for previous version.
I don't know which ones take precedence. What is the spack policy here? Should I change it to the versions from `pyproject.toml`

repos/spack_repo/builtin/packages/py_statsmodels/package.py
climbfuji: Thanks, that makes sense.

repos/spack_repo/builtin/packages/py_statsmodels/package.py
tldahlgren: Excellent point. Thanks for adding the comment.

PR #1904, opened 2025-10-08
py-tifffile: add v2025.10.4

manuelakuhn:
https://github.com/cgohlke/tifffile/tree/v2025.10.4


tldahlgren (approved):
Confirmed sha256 and dependencies.

PR #1905, opened 2025-10-08
pbzip2: new package

Markus92:
pbzip2 is a parallel implementation of bzip2. It splits the file up in chunks, then uses bzip2 to compress it. It achieves significant speed-ups.

It's not very maintained anymore by upstream, but still fully functional and included in Ubuntu, Debian, RHEL, Arch and a lot of other Linux distributions, so it's still relevant I'd say.

It's a dependency for Relion.


Markus92:
@spackbot fix style

becker33 (approved):

PR #1908, opened 2025-10-08
Charliecloud: Fix URL and update bdw-gc dependency

nhanford:
Hello @j-ogas @reidpr and @loshak,

I have updated the `charliecloud` package to build the latest versions I could find over on your new Gitlab project/repo. I would like to request feedback on the following:

## URL update and intended tarballs for distribution

Please take a look at the tar URL I used. I played around with the Gitlab API somewhat and couldn't find a decent way to get the correct versions and checksums programmatically so I left it up to Spack and used the URLs it auto-resolved when I pointed it at the Gitlab URL. However, these tarballs *may not* be the ones you intended folks to download, in which case we should talk to the Spack maintainers about `lib/spack/spack/llnl/url.py` as I believe that's where the "auto-resolution" is coming from.

## Charliecloud version using newer symbols from `bdw-gc`

I ran into errors of the type: 
```
spack-src/bin/mem.c:316: undefined reference to `GC_get_full_gc_total_time'
```
when building "naively" on RHEL 8. Some digging found that the missing symbols were added in bdw-gc (the Boehm Garbage Collector) version 8. RHEL 8 only has version 7.6.2. I'm not sure when you started adding references to these symbols (maybe version 0.41?) so I just made `bdw-gc@8:` a blanket build dependency. However, there was more fun to be had as autotools appeared to happily ignore the existence of the Spack-installed newer version of `bdw-gc` and the link line didn't include it explicitly, so I made it a `link` dependency in Spack as well. Now it builds for me.

Anyway please take a look and tell me what you think.

Thanks,
Nate

### DIFF
--- a/repos/spack_repo/builtin/packages/charliecloud/package.py
+++ b/repos/spack_repo/builtin/packages/charliecloud/package.py
@@ -34,6 +34,7 @@ class Charliecloud(AutotoolsPackage):
     variant("cdi", default=True, description="Build with CDI support", when="@0.40:")
 
     depends_on("c", type="build")  # generated
+    depends_on("bdw-gc@8:", type=("build", "link"))
 
     # Autoconf.
     depends_on("m4", type="build")
--- a/repos/spack_repo/builtin/packages/charliecloud/package.py
+++ b/repos/spack_repo/builtin/packages/charliecloud/package.py
@@ -34,7 +34,8 @@ class Charliecloud(AutotoolsPackage):
     variant("cdi", default=True, description="Build with CDI support", when="@0.40:")
 
     depends_on("c", type="build")  # generated
-    depends_on("bdw-gc@8:", type=("build", "link"))
+    depends_on("bdw-gc", type=("build", "link"))
+    depends_on("bdw-gc@8:", type=("build", "link"), when="@0.40:")
 
     # Autoconf.
     depends_on("m4", type="build")
--- a/repos/spack_repo/builtin/packages/charliecloud/package.py
+++ b/repos/spack_repo/builtin/packages/charliecloud/package.py
@@ -12,7 +12,7 @@ class Charliecloud(AutotoolsPackage):
 
     maintainers("j-ogas", "reidpr", "loshak")
     homepage = "https://charliecloud.io/"
-    url = "https://gitlab.com/charliecloud/main/-/archive/v0.42/main-v0.42.tar.gz"
+    url = "https://gitlab.com/charliecloud/charliecloud/-/archive/v0.42/main-v0.42.tar.gz"
     git = "https://gitlab.com/charliecloud/charliecloud.git"
 
     tags = ["e4s"]
--- a/repos/spack_repo/builtin/packages/charliecloud/package.py
+++ b/repos/spack_repo/builtin/packages/charliecloud/package.py
@@ -11,15 +11,17 @@ class Charliecloud(AutotoolsPackage):
     """Lightweight user-defined software stacks for HPC."""
 
     maintainers("j-ogas", "reidpr", "loshak")
-    homepage = "https://hpc.github.io/charliecloud"
-    url = "https://github.com/hpc/charliecloud/releases/download/v0.18/charliecloud-0.18.tar.gz"
-    git = "https://github.com/hpc/charliecloud.git"
+    homepage = "https://charliecloud.io/"
+    url = "https://gitlab.com/charliecloud/main/-/archive/v0.42/main-v0.42.tar.gz"
+    git = "https://gitlab.com/charliecloud/charliecloud.git"
 
     tags = ["e4s"]
 
     license("Apache-2.0")
 
-    version("master", branch="master")
+    version("main", branch="main")
+    version("0.42", sha256="be98c025f58336a7b6e6d79804ef89dd489c5dcc5ad8faccb551ea0065dcd13a")
+    version("0.41", sha256="065cc50f8b7893f8a0e28d9d06e2e3640d0d8139d10ad59fe941aea1e33dfdc6")
     version("0.40", sha256="dcad81136d1fed905be6e573a7bf191ea655ae7827f7980bbe6559942f2affdd")
     version("0.39", sha256="38503b507119a970ac288df7181aefe6cd1a125b9d509f5cb162dacea7143fd1")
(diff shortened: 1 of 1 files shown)


reidpr:
@nhanford Do you think there are also changes we should make upstream on this?

nhanford:
> @nhanford Do you think there are also changes we should make upstream on this?

Hello @reidpr,

I don't know if Autotools has a convenient way to check the version of the Boehm Garbage Collector, but that could be a way to catch this at configure before we get all the way link time.

It would be nice to have SHA checksums that I could go see to verify this stuff, but I can't seem to find any on Gitlab that match what I get locally via Spack or `sha256sum`... I like what you have for the [package registry](https://gitlab.com/charliecloud/charliecloud/-/packages/46419666), but I would need a way to tell Spack where to locate the tarball from just a version number. Anyway, that's probably more of an issue with how Spack gets stuff from Gitlab, so I would defer to the Spack maintainers on that issue.

Regards,
Nate

nhanford:
@reidpr I just confirmed that I did the hashes correctly, so as long as you're cool with the urls I set and requiring `bdw-gc@8:` across all versions then this should be good to go from my end. We tested it here at LLNL and confirmed we got all the functionality we expected.

repos/spack_repo/builtin/packages/charliecloud/package.py
```diff
@@ -11,15 +11,17 @@ class Charliecloud(AutotoolsPackage):
     """Lightweight user-defined software stacks for HPC."""
 
     maintainers("j-ogas", "reidpr", "loshak")
-    homepage = "https://hpc.github.io/charliecloud"
-    url = "https://github.com/hpc/charliecloud/releases/download/v0.18/charliecloud-0.18.tar.gz"
-    git = "https://github.com/hpc/charliecloud.git"
+    homepage = "https://charliecloud.io/"
+    url = "https://gitlab.com/charliecloud/main/-/archive/v0.42/main-v0.42.tar.gz"
+    git = "https://gitlab.com/charliecloud/charliecloud.git"
```
reidpr, line 16: Good catch on these, thanks.

repos/spack_repo/builtin/packages/charliecloud/package.py
```diff
@@ -11,15 +11,17 @@ class Charliecloud(AutotoolsPackage):
     """Lightweight user-defined software stacks for HPC."""
 
     maintainers("j-ogas", "reidpr", "loshak")
-    homepage = "https://hpc.github.io/charliecloud"
-    url = "https://github.com/hpc/charliecloud/releases/download/v0.18/charliecloud-0.18.tar.gz"
-    git = "https://github.com/hpc/charliecloud.git"
+    homepage = "https://charliecloud.io/"
+    url = "https://gitlab.com/charliecloud/main/-/archive/v0.42/main-v0.42.tar.gz"
```
reidpr:
```suggestion
    url = "https://gitlab.com/charliecloud/charliecloud/-/archive/v0.42/main-v0.42.tar.gz"
```

because of a recent rename ([#1999](https://gitlab.com/charliecloud/charliecloud/-/issues/1999)).

repos/spack_repo/builtin/packages/charliecloud/package.py
```diff
@@ -32,6 +34,7 @@ class Charliecloud(AutotoolsPackage):
     variant("cdi", default=True, description="Build with CDI support", when="@0.40:")
 
     depends_on("c", type="build")  # generated
+    depends_on("bdw-gc@8:", type=("build", "link"))
```
reidpr: This will be needed only on v0.40 and up.

reidpr (commented):
Thanks Nathan, appreciate your work here. I found a few nits to consider. Also, if I’m reading the diff correctly it may need to be rebased since PR #1190 landed?

repos/spack_repo/builtin/packages/charliecloud/package.py
nhanford: I have added and tested the relevant lines to restrict this `bdw-gc` version dependency to `charliecloud` version 0.40+.

nhanford:
> Thanks Nathan, appreciate your work here. I found a few nits to consider. Also, if I’m reading the diff correctly it may need to be rebased since PR #1190 landed?

Hello @reidpr, I rebased and it appears there were no conflicts and updated the other things you mentioned. Please let me know what you think.
Thanks,
Nate

repos/spack_repo/builtin/packages/charliecloud/package.py
```diff
@@ -32,6 +34,8 @@ class Charliecloud(AutotoolsPackage):
     variant("cdi", default=True, description="Build with CDI support", when="@0.40:")
 
     depends_on("c", type="build")  # generated
+    depends_on("bdw-gc", type=("build", "link"))
```
reidpr, line 37: Did fd91121 miss removing this line, since the following line is more specific?

nhanford: Hello @reidpr, These two lines together are saying "This package always depends on `bdw-gc`. It depends on `bdw-gc` version 8 or later when this package is version 0.40 or later." Removing this line would change this to "This package only depends on `bdw-gc` *at all* when this package is version 0.40 or later. In that case, it depends on `bdw-gc` version 8 or later."

reidpr: Aha, thanks for clarifying. In that case, this would be the correct dependency:

> This package only depends on bdw-gc at all when this package is version 0.40 or later. In that case, it depends on bdw-gc version 8 or later.

Prior to 0.40, Charliecloud didn't use libgc (bdw-gc) at all.

(I may have said this already, but it's an optional dependency still. If not present, we just leak all heap allocations. 😬 So in my view, if it's available it should be used, i.e. I don't think making another variant is worth the hassle.)

nhanford: I usually don't dig in on these things, but this is a very small/lightweight dependency that is more widely supported than other Charliecloud required dependencies. I built it on my system in seconds. It also avoids memory leaks as you mention.  Bearing that in mind, I strongly recommend indeed not creating a variant and just requiring this dependency for all Spack builds as it is here.

j-ogas (approved):
After reviewing the MR and discussion, I approve as is. Merge at will.

alecbcs (approved):

PR #1920, opened 2025-10-09
py-types-psutil: add v7.0.0.20251001

manuelakuhn:
https://pypi.org/project/types-psutil/#files
(no git tags)


alecbcs (approved):
Looks good to me. Thanks!

PR #1924, opened 2025-10-09
acts: add v44.0.0, v44.0.1, v44.1.0

stephenswat:
This commit adds ACTS version 44.0.0, which is the first version in a while that can be built while air-gapped; therefore, there is a bit of a gap in the version history.


repos/spack_repo/builtin/packages/acts/package.py
```diff
@@ -50,6 +50,11 @@ class Acts(CMakePackage, CudaPackage):
 
     # Supported Acts versions
     version("main", branch="main")
+    version("44.0.1", commit="404f40aaa6211231b6c6726a364b08134a2e3fa4", submodules=submodules)
+    version("44.0.0", commit="d5d65c794d3676034f37d89e555c131b5b7ad807", submodules=submodules)
+    # NOTE: Versions between 39.2.0 and 44.0.0 are not available via Spack,
+    # as they cannot be built without downloading third-party dependencies
+    # from remote, non-Spack sources.
```
wdconinc, line 58: I think it would actually be OK to add them now, potentially with the deprecated flag or so. We mainly want to avoid having people get updates (even through git pull on develop) that suddenly change behavior to require an internet connection. Default behavior now is retained on the v44 releases, so anyone who installs a previous release is doing it consciously. Thoughts?

stephenswat: To me, deprecation indicates the explicit desire to remove the versions in the future. Is there any other flag that does not imply that which, like the sticky option for variants, can be used to avoid accidental concertising of these versions?

repos/spack_repo/builtin/packages/acts/package.py
```diff
@@ -423,18 +435,24 @@ class Acts(CMakePackage, CudaPackage):
     depends_on("python", when="+python")
     depends_on("python@3.8:", when="+python @19.11:19")
     depends_on("python@3.8:", when="+python @21:")
```
wdconinc, line 441: https://github.com/acts-project/acts/blob/7b1edddb4dad8db5a5ef5beb39a9dc68fbc1d704/codegen/src/codegen/sympy_common.py#L281

The sympy codegen includes nested f-strings with quoting only supported in python 3.12 and later.

repos/spack_repo/builtin/packages/acts/package.py
wdconinc: There's no other flag. One could add a variant. But let's just go ahead and ship this without the skipped versions. If someone (like me, for my community) needs it, they can add it.

repos/spack_repo/builtin/packages/acts/package.py
jmcarcell, line 432:
```suggestion
    depends_on("podio @:0", when="@:35 +edm4hep")
    depends_on("podio @:1.4", when="@:39.2 +edm4hep")
```
This is because the `EDM4hepParticleWriter.cpp` uses `->` which has been removed in podio (https://github.com/AIDASoft/podio/pull/812) in: https://github.com/acts-project/acts/blob/94cf48783efd713f38106b18211d1c59f4e8cdec/Examples/Io/EDM4hep/src/EDM4hepParticleWriter.cpp#L46

wdconinc: It's still there in acts 44 too, so the limit needs to be higher.

repos/spack_repo/builtin/packages/acts/package.py
```diff
@@ -423,18 +435,24 @@ class Acts(CMakePackage, CudaPackage):
...
+    depends_on("py-numpy @2.2", when="@44:")
     depends_on("py-onnxruntime@:1.12", when="+onnx @:23.2")
     depends_on("py-onnxruntime@1.12:", when="+onnx @23.3:")
+    depends_on("py-particle @0.24", when="@44:")
     depends_on("py-pybind11 @2.6.2:", when="+python @18:")
     depends_on("py-pybind11 @2.13.1:", when="+python @36:")
     depends_on("py-pytest", when="+python +unit_tests")
+    depends_on("py-setuptools", when="@44:")
+    depends_on("py-sympy @1.13", when="@44:")
```
jmcarcell:
```suggestion
    depends_on("py-sympy @1.13", when="@44:")
    # Fix the version when the next after 44.1.0 is released
    depends_on("py-hatchling", when="@44.1.1")
```
See https://github.com/acts-project/acts/pull/4770#issuecomment-3489785399, this will fix `main` builds.

stephenswat: Thanks, updated.

repos/spack_repo/builtin/packages/acts/package.py
stephenswat: Updated.

repos/spack_repo/builtin/packages/acts/package.py
stephenswat: Fixed, thanks.

stephenswat:
Incorporated feedback (thanks @wdconinc and @jmcarcell). Also added v44.1.0.

repos/spack_repo/builtin/packages/acts/package.py
```diff
@@ -420,22 +433,33 @@ class Acts(CMakePackage, CudaPackage):
...
     depends_on("python@3.8:", when="+python @21:")
+    depends_on("python@3.12:", when="+python @44:")
+    depends_on("py-numpy @2.2", when="@44:")
     depends_on("py-onnxruntime@:1.12", when="+onnx @:23.2")
     depends_on("py-onnxruntime@1.12:", when="+onnx @23.3:")
+    depends_on("py-particle @0.24", when="@44:")
     depends_on("py-pybind11 @2.6.2:", when="+python @18:")
     depends_on("py-pybind11 @2.13.1:", when="+python @36:")
     depends_on("py-pytest", when="+python +unit_tests")
+    depends_on("py-setuptools", when="@44:44.1.0")
```
wdconinc: Type build?

wdconinc: Similar suggestion for other codegen dependencies actually.

wdconinc (commented):
Looks good. Just one suggestion on adding type=build where possible.

repos/spack_repo/builtin/packages/acts/package.py
stephenswat: Done, thank you! Well spotted.

wdconinc (approved):

PR #1929, opened 2025-10-09
py_simple_slurm: new package

adamwitmer:
@rbberger I have added myself as the maintainer of this package, and had incorrectly tagged the wrong developer, who is actually @amq92

### DIFF
--- a/repos/spack_repo/builtin/packages/py_simple_slurm/package.py
+++ b/repos/spack_repo/builtin/packages/py_simple_slurm/package.py
@@ -13,10 +13,10 @@ class PySimpleSlurm(PythonPackage):
     homepage = "https://github.com/amq92/simple_slurm"
     pypi = "simple-slurm/simple_slurm-0.3.6.tar.gz"
 
-    maintainers = "mamachra"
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/py_simple_slurm/package.py
```diff
@@ -0,0 +1,22 @@
...
+
+    homepage = "https://github.com/amq92/simple_slurm"
+    pypi = "simple-slurm/simple_slurm-0.3.6.tar.gz"
+
+    license("AGPL-3.0")
+    maintainers("adamwitmer")
+
+    version("0.3.6", sha256="0f88cac96b39d7d7e34c5a124919cebb41eb8a8ef2ff4fb65c363d940ff8551e")
+
+    depends_on("python@3.11:", type=("build", "run"))
```
rbberger: Explicit Python dependency isn't needed. The setup.py says it needs Python 3.6+, which is always true in Spack.
```suggestion
```

rbberger (changes requested):

rbberger (approved):

PR #1938, opened 2025-10-10
kokkos: add missing amdgpu_target override

rbberger:
Without this change, the concretizer has the freedom to choose a combination of `amdgpu_target` values. We actually have the same override for `cuda_arch`. But recently this oversight started to become an issue for some concretizations.

### DIFF
--- a/repos/spack_repo/builtin/packages/kokkos/package.py
+++ b/repos/spack_repo/builtin/packages/kokkos/package.py
@@ -189,6 +189,7 @@ class Kokkos(CMakePackage, CudaPackage, ROCmPackage):
         when="+cuda",
     )
 
+    # Since Kokkos supports only one amdgpu_target at a time, the multi-value property is disabled.
     variant(
         "amdgpu_target",
         description="AMD GPU architecture",
--- a/repos/spack_repo/builtin/packages/kokkos/package.py
+++ b/repos/spack_repo/builtin/packages/kokkos/package.py
@@ -388,13 +388,7 @@ class Kokkos(CMakePackage, CudaPackage, ROCmPackage):
 
         spack_microarches = []
         if spec.satisfies("+cuda"):
-            if isinstance(spec.variants["cuda_arch"].value, str):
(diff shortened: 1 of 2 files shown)


rbberger:
@alalazo I've bisected this issue to come from spack@b70a028e067824cf57f9660fe983bd9169f534f9. Due to this change, the `amdgpu_target` multi variant will have "random" other values, despite me only requiring `amdgpu_target=gfx90a` in my environments. Examples I've seen: `amdgpu_target:=gfx942,gfx90a`, `amdgpu_target:=gfx90a,gfx940`, `amdgpu_target:=gfx1030,gfx900,gfx906,gfx908,gfx90a`. This PR is a workaround, but this might mean there are other such cases where the concretizer will be "more random".

repos/spack_repo/builtin/packages/kokkos/package.py
```diff
@@ -189,6 +189,16 @@ class Kokkos(CMakePackage, CudaPackage, ROCmPackage):
         when="+cuda",
     )
 
+    variant(
```
tpadioleau, line 193:
```suggestion
    # Since Kokkos supports only one amdgpu_target at a time, the multi-value property is disabled.
    variant(
```

tpadioleau (approved):
lgtm, I would just add a comment in the package to explain the addition as in the CUDA case.

masterleinad (approved):

tpadioleau (approved):
Still ok for me, I think we should be able to further simplify `!= "none"`. In the CUDA case we explicitly conflict with `none` but not in the AMD nor Intel cases. It may need its own PR.

rbberger:
@spackbot fix style

rbberger:
@kwryankrattiger I think I'm running into timeout issues here. Or is just a runner failure? (see rocblas).

rbberger:
@afzpatel since I'm touching `hip/package.py` could you please review this too?

afzpatel (approved):

alalazo (approved):

PR #1939, opened 2025-10-10
python: 3.14.0, 3.13.8, 3.12.12, 3.11.14, 3.10.19, 3.9.24

haampie:
I would suggest to make `3.14.0` the preferred version and add forward compat bounds if we see issues with particular packages.


skosukhin:
@haampie, could you please remove me from the maintainers, or will you make me create a separate PR for that?

adamjstewart (dismissed):
I'll open a separate PR to deprecate 3.9 at the end of the month.

haampie:
I'll fix the Windows issue and then force merge to avoid another 7h of CI.

PR #1945, opened 2025-10-10
py_mpi4py: add c++ dependency

AlexanderRichert-NOAA:
This PR adds a c++ dependency for pi_mpi4py. Without it, build fails on account of mpi_spec not having mpicxx attribute.

### DIFF
--- a/repos/spack_repo/builtin/packages/py_mpi4py/package.py
+++ b/repos/spack_repo/builtin/packages/py_mpi4py/package.py
@@ -37,7 +37,8 @@ class PyMpi4py(PythonPackage):
     version("2.0.0", sha256="6543a05851a7aa1e6d165e673d422ba24e45c41e4221f0993fe1e5924a00cb81")
     version("1.3.1", sha256="e7bd2044aaac5a6ea87a87b2ecc73b310bb6efe5026031e33067ea3c2efc3507")
 
-    depends_on("c", type="build")  # generated
+    depends_on("c", type="build")
+    depends_on("cxx", type="build")
 
     depends_on("py-setuptools@40.9:", type="build")
     depends_on("py-cython@3:", when="@4:", type="build")


rbberger:
@AlexanderRichert-NOAA with what MPI implementation and version are you seeing this? I've just tested this with mpich and it builds fine both for 4.0.1 and master.

Chrismarsh:
I build mpi4py against OpenMPI/gcc and I don't see this problem either:
```
 spack spec py-mpi4py
[+]  py-mpi4py@4.0.1 build_system=python_pip platform=linux os=rhel8 target=icelake %c=gcc@14.3.0
```

AlexanderRichert-NOAA:
Hm. I'm using cray-mpich. I'm guessing it's because `spec.mpicxx` is set conditionally depending on whether the dependent package actually needs it.

Are those with or without cxx enabled?

rbberger:
is your external cray-mpich defined with `+wrappers`? That's what you need to have those variables defined.

rbberger:
I think the issue is that with `+wrappers` it inherits the MPICH logic, which unconditionally adds the wrappers, while otherwise it'll depend on the language dependencies. This goes all the way back to 1deb6325c595

AlexanderRichert-NOAA:
Hm. I'm using `~wrappers` so I can use cc/CC/ftn (specifically, the paths set in my compiler config) from craype.

I may be missing something, but, if the mpi4py recipe is using mpicxx and is baking in a compiler name or path, doesn't that imply that there's a cxx dependency?

rbberger:
The recipe yes, the actual build itself doesn't apparently.

rbberger:
In any case, if we assume that all MPIs (OpenMPI and MPICH) are wrong to just always define `mpicxx` for dependent packages and only `cray-mpich` was correctly set up by checking for language virtuals, then this MR is the right fix. I'm gravitating toward that conclusion, but we should make @alalazo, OpenMPI maintainers (@hppritcha, @naughtont3) and MPICH maintainers (@raffenet, @yfguo) aware. I think we need to come up with a consistent answer for MPIs.

rbberger (approved):

raffenet:
> In any case, if we assume that all MPIs (OpenMPI and MPICH) are wrong to just always define `mpicxx` for dependent packages and only `cray-mpich` was correctly set up by checking for language virtuals, then this MR is the right fix. I'm gravitating toward that conclusion, but we should make @alalazo, OpenMPI maintainers (@hppritcha, @naughtont3) and MPICH maintainers (@raffenet, @yfguo) aware. I think we need to come up with a consistent answer for MPIs.

Sounds OK to me. Does the same need to be done for Fortran `mpifort`?

rbberger:
Yes, I believe so.

alalazo (approved):

PR #1946, opened 2025-10-10
Bug fix in ectrans: use cmake to detect LAPACK libraries

climbfuji:
## Description

Bug fix in ectrans: use cmake to detect LAPACK libraries unless MKL is requested explicitly, don't just go and look for cray-libsci.

See https://github.com/ecmwf-ifs/ectrans/issues/316 for details.

**Notes**
- I will be on leave Oct 11-20. Any feedback from the ectrans developers regarding the minimum cmake version required to support detecting LAPACK via cmake/ecbuild will be incorporated after my return.
- I tested the PR as is (hash https://github.com/spack/spack-packages/commit/7846a5f0e71315cd7ac45b747fc98a0541be7d08) on a system with cmake 3.31.8, and it correctly linked against the LAPACK provider from spack (openblas)
- Update 2025/10/23: I tested this with cmake@3.25.2 after hearing back from the ectrans developers that their most-recent release 1.7.0 requires cmake@3.25.


repos/spack_repo/builtin/packages/ectrans/package.py
```diff
@@ -67,6 +67,9 @@ class Ectrans(CMakePackage):
         when="@1.3.1:1.5.1",
     )
 
+    # LINK TO ISSUE HERE
```
climbfuji:
```suggestion
    # https://github.com/ecmwf-ifs/ectrans/issues/316
```

repos/spack_repo/builtin/packages/ectrans/package.py
```diff
@@ -47,15 +47,23 @@ class Ectrans(CMakePackage):
...
 
-    depends_on("c", type="build")  # generated
-    depends_on("fortran", type="build")  # generated
+    depends_on("c", type="build")
+    depends_on("cxx", type="build")
+    depends_on("fortran", type="build")
+
+    # Add explicit dependency on newer cmake versions in order to apply patch
+    # "find_lapack.patch", see below and https://github.com/ecmwf-ifs/ectrans/issues/316
+    # Newer versions of ectrans (not yet released) also require cmake@3.25: by default).
```
climbfuji:
```suggestion
    # Newer versions of ectrans (1.7.0+) also require cmake@3.25: by default.
```

alalazo (approved):

PR #1950, opened 2025-10-10
arrow: don't build mimalloc with intel 2025

Chrismarsh:
Works around https://github.com/apache/arrow/issues/47790 which prevents building with the new clang/llvm intel-oneapi compilers


Chrismarsh:
@spackbot fix style

adamjstewart (approved):
Seems fine to me. It doesn't seem like an important enough feature to warrant a variant, so this is fine for now.

Chrismarsh:
My hope was that it would be fixed upstream, but 3 weeks without any feedback on either issue 🤷

PR #1951, opened 2025-10-10
py-types-setuptools: add v80.9.0.20250822

manuelakuhn:
https://pypi.org/project/types-setuptools/#files
(no tags in git)


alecbcs (approved):
Looks good to me. Thanks!

PR #1957, opened 2025-10-11
Update package.py for orca 6.1.0

alansill:
Add sha256sum values for regular and avx2 versions of orca 6.1.0, as well as OpenMPI version specification (4.1.8) for these.

@snehring

### DIFF
--- a/repos/spack_repo/builtin/packages/orca/package.py
+++ b/repos/spack_repo/builtin/packages/orca/package.py
@@ -25,6 +25,10 @@ class Orca(Package):
     license("LGPL-2.1-or-later")
 
     version(
+        "avx2-6.1.0", sha256="f09a08b5971b18d1f85364f7e57becdb5179fbddc1a1636802551521ef778774"
+    )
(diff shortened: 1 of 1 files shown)


tldahlgren:
Added maintainer ping to the description since cannot be added to the Reviewers list.

Note that the version sha256s can't be automatically confirmed since this is a `manual download` package.

snehring (changes requested):
checksums check out. ~However line 71 will need to be changed to be ```@avx2-6.0.1:```~, it does look like they're carrying the versioning for this forward.

snehring:
Wait, that's wrong. Let me look into this.

snehring:
This should fix it. Since they've decided to go with the 6.0.1 scheme going forward we'll treat that as the default and handle 6.0.0 as a weird edge case.

```
        features = ver_parts[:-1]
        feature_text = "_".join([""] + features)

        url = f"file://{os.getcwd()}/orca_{ver_underscored}_linux_x86-64_shared_openmpi{openmpi_version}{feature_text}.tar.xz"
        if self.spec.satisfies("@=avx2-6.0.0"):
            url = f"file://{os.getcwd()}/orca_{ver_underscored}_linux_x86-64_avx2_shared_openmpi{openmpi_version}.tar.xz"
```

PR #1968, opened 2025-10-13
superlu-dist: add patch for compilation with CUDA

msimberg:
This backports a bugfix available on superlu-dist master, but not yet in a release, which fixes compilation with `+cuda` (possibly also `+rocm`, but I have not tried that): https://github.com/xiaoyeli/superlu_dist/commit/5a1946f347e6d813a250af874dee0942f4fdfc44. The patch applies to 9.0 and 9.1, though with 9.0 there is another build failure that I have not tried to fix, since 9.1 builds now.


liuyangzhuan (approved):

albestro (approved):
Thanks for the review @liuyangzhuan 

LGTM

PR #1969, opened 2025-10-13
Repeatmodeler 2.0.7

snehring:
Version update. I'm also deprecating the 1.0.11 version with the intent to remove it in the near future to allow a bit of recipe rework. The ltr_retreiver version requirement for repeatmodeler became 2.9.0 after the 2.0.4 release, but I assume it's equally relevant for it.


tldahlgren (approved):
Confirmed the homepage and all new version sha256s.

PR #1972, opened 2025-10-14
flecsi: simplify package after version removals

rbberger:

### DIFF
--- a/repos/spack_repo/builtin/packages/flecsi/package.py
+++ b/repos/spack_repo/builtin/packages/flecsi/package.py
@@ -18,7 +18,7 @@ class Flecsi(CMakePackage, CudaPackage, ROCmPackage):
     mesh topology, mesh geometry, and mesh adjacency information.
     """
 
-    homepage = "http://flecsi.org/"
+    homepage = "https://www.flecsi.org/"
     git = "https://github.com/flecsi/flecsi.git"
     maintainers("rbberger", "opensdh")
 
@@ -43,7 +43,7 @@ class Flecsi(CMakePackage, CudaPackage, ROCmPackage):
     variant("shared", default=True, description="Build shared libraries")
     variant("flog", default=False, description="Enable logging support")
     variant("graphviz", default=False, description="Enable GraphViz Support")
-    variant("doc", default=False, description="Enable documentation", when="@2.2:")
+    variant("doc", default=False, description="Enable documentation")
     variant("hdf5", default=True, description="Enable HDF5 Support")
     variant(
         "caliper_detail",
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/flecsi/package.py
```diff
@@ -66,13 +66,11 @@ class Flecsi(CMakePackage, CudaPackage, ROCmPackage):
     depends_on("hdf5+hl+mpi", when="+hdf5")
     depends_on("metis@5.1.0:", when="@:2.3.1")
     depends_on("parmetis@4.0.3:", when="@:2.3.1")
-    depends_on("boost@1.70.0: cxxstd=17 +program_options +stacktrace")
+    depends_on("boost@1.79.0: cxxstd=17 +program_options +stacktrace")
 
-    depends_on("cmake@3.15:")
     depends_on("cmake@3.19:", when="@2.2:")
```
opensdh: This and one other `@2.2:` (and a "2.2+") can be removed.

repos/spack_repo/builtin/packages/flecsi/package.py
```diff
@@ -18,7 +18,7 @@ class Flecsi(CMakePackage, CudaPackage, ROCmPackage):
     mesh topology, mesh geometry, and mesh adjacency information.
     """
 
-    homepage = "http://flecsi.org/"
+    homepage = "https://flecsi.org/"
```
opensdh:
```suggestion
    homepage = "https://www.flecsi.org/"
```
for consistency with our own repository.

repos/spack_repo/builtin/packages/flecsi/package.py
```diff
@@ -82,7 +80,6 @@ class Flecsi(CMakePackage, CudaPackage, ROCmPackage):
     depends_on("kokkos +openmp", when="+kokkos +openmp")
     requires("+openmp", when="@:2.3 ^kokkos +openmp")
     depends_on("legion@cr-20230307", when="backend=legion @2.2.0:2.2.1")
-    depends_on("legion@24.03.0:", when="backend=legion @2.2.2:")
```
opensdh, line 85: Don't we need this, since otherwise 2.3.0 won't have any version requirement at all?

We could simplify these two lines to use `@:2.2` and `@2.3:` to avoid referring to nonexistent versions (even though it might really have been that the end of the `2.2` branch needed this).

opensdh: There being no 2.3.0 (anymore), and it being a bit misleading to use `@2.3:` on a version that wasn't actually used for all 2.3.* versions, I guess there's nothing to do here.

opensdh (approved):

tldahlgren (approved):
Confirmed new homepage.

PR #1977, opened 2025-10-14
Harbuzz: add cmake builder

johnwparent:
Harfbuzz has had a CMake build system upstream for some time. Add it to Spack.
Harfbuzz package has difficulty detecting CMake based dependencies from its Meson system, so this is important to help reduce the amount of maintenance is required to support harfbuzz in Spack.

* Add CMake builder for Harfbuzz
* Specify that when we build freetype with CMake, we must also build harfbuzz with freetype as harbuzz has decided not to support Freetype built with CMake with their meson system.
* Removes patch backporting support for CMake built freetype when building harfbuzz with meson that was removed upstream


repos/spack_repo/builtin/packages/harfbuzz/package.py
```diff
@@ -75,17 +109,20 @@ class Harfbuzz(MesonPackage, AutotoolsPackage):
         depends_on("meson@0.60:", when="@11.1:")
         depends_on("meson@0.55:", when="@3.2.1:")
         depends_on("meson@0.52:")
+        # harfbuzz's Meson only supports autotools based
+        # freetype
+        depends_on("freetype build_system=autotools")
```
johnwparent, line 104: Probably a better way to express this.

johnwparent:
@spackbot fix style

johnwparent:
I realllly need to get style checks working again for spack-packages on Windows

scheibelp (approved):

PR #1978, opened 2025-10-14
py-matplotlib: matplotlib < 3.5.0 is incompatible with Python >= 3.12

tukss:
The setup routine of the older matplotlib version uses `SafeConfigParser`, which was removed from Python 3.12.


adamjstewart (changes requested):
I would rather use:
```python
depends_on("python@:3.11", when="@:3.4", type=("build", "link", "run"))
```
This way, when we remove Python 3.11, we will know to remove older matplotlib as well.

tukss:
Thanks for the suggestion. I've changed it accordingly. I've left the comment in but can remove it if you prefer.

adamjstewart (approved):

PR #1984, opened 2025-10-15
Update dotnet-core-sdk for windows and darwin

grospelliergilles:

### DIFF
--- a/repos/spack_repo/builtin/packages/dotnet_core_sdk/package.py
+++ b/repos/spack_repo/builtin/packages/dotnet_core_sdk/package.py
@@ -15,6 +15,7 @@ class DotnetCoreSdk(Package):
 
     homepage = "https://www.microsoft.com/net/"
 
+    maintainers("grospelliergilles")
     license("MIT")
 
     if platform.system() == "Linux" and platform.machine() == "x86_64":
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/dotnet_core_sdk/package.py
```diff
@@ -41,6 +42,34 @@ class DotnetCoreSdk(Package):
             url="https://builds.dotnet.microsoft.com/dotnet/Sdk/6.0.428/dotnet-sdk-6.0.428-linux-arm64.tar.gz",
             sha256="b5956b0d9ab3063c2886ec74adc953394e81a1aa3f5075c6b41b3e4f1d7a53f9",
         )
+    elif platform.system() == "darwin" and platform.machine() == "x86_64":
```
tldahlgren: The first letter returned by `platform.system()` is in upper case.

You can see this by running:
```
$ spack python
Spack version 1.1.0.dev0
Python 3.13.5, Darwin arm64
>>> import platform
>>> platform.system()
'Darwin'
>>> exit()
```

tldahlgren: I'll have to trust you on the sha256 because I'm not getting the value with this fix or when running `sha256sum < <archive-file>` on/for the above machine.

tldahlgren (changes requested):
See comment.

grospelliergilles:
Thanks @tldahlgren for the quick review. I have fixed the uppercase for platform.
You were right, some hash were bad because my downloads were truncated. I have also fixed that.

tldahlgren (approved):
Thanks.

PR #1990, opened 2025-10-15
plumed: add libtorch and metatomic support and update GROMACS compatibility

RMeli:
Add `libtorch` and `metatomic` support to `plumed` package, and update GROMACS-Plumed compatibility matrix.


RMeli:
@spackbot fix style

repos/spack_repo/builtin/packages/plumed/package.py
```diff
@@ -160,12 +164,16 @@ class Plumed(AutotoolsPackage):
     depends_on("m4", type="build")
     depends_on("py-cython", type="build")
 
+    depends_on("py-torch", when="+pytorch")
+    conflicts("+metatomic ~pytorch", msg="metatomic support requires PyTorch")
```
Luthaf, line 168: is there a way to automatically enable pytorch when the user request metatomic, and only error if the user explicitly disable pytorch?

repos/spack_repo/builtin/packages/plumed/package.py
```diff
@@ -248,19 +258,55 @@ def configure_args(self):
...
+            extra_libs.append(spec["py-torch"].libs.link_flags)
+            # Add include paths manually
+            # Spack HeaderList.cpp_flags does not support include paths within include paths
+            extra_cppflags.extend(
+                [
+                    f"-I{pytorch_path / 'include'}",
+                    f"-I{pytorch_path / 'include' / 'torch' / 'csrc' / 'api' / 'include'}",
+                ]
+            )
+            print(extra_cppflags[-2:])
```
Luthaf:
```suggestion
```

repos/spack_repo/builtin/packages/plumed/package.py
RMeli: This is what this conflicts does. It allows `+pytorch ~metatomic` (which is valid, as far as I understand), but disallows `~pytorch +metatomic`. If you select only `+metatomic` then the `pytorch` variant automatically needs to be `+pytorch` if not specified in order to satisfy the constraint. However, if a user explicitly requires `+metatomic ~pytorch`, then an error is triggered.

This is essentially because the `conflicts` is stronger than the variant default value, see https://spack.readthedocs.io/en/latest/frequently_asked_questions.html#why-does-spack-pick-particular-versions-and-variants

repos/spack_repo/builtin/packages/plumed/package.py
RMeli: Good catch, thanks!

repos/spack_repo/builtin/packages/plumed/package.py
Luthaf: Thanks for the explanation! 🙏

mabraham:
Please mention in the description that the GROMACS-Plumed support matrix has been updated

mabraham (approved):

repos/spack_repo/builtin/packages/plumed/package.py
```diff
@@ -248,19 +258,54 @@ def configure_args(self):
             if spec.satisfies("^[virtuals=mpi] intel-oneapi-mpi"):
                 configure_opts.extend(["STATIC_LIBS=-mt_mpi"])
 
+        enable_libmetatomic = self.spec.satisfies("+metatomic")
+        enable_libtorch = self.spec.satisfies("+pytorch") or self.spec.satisfies("+metatomic")
```
marcodelapierre: Please consider having only `enable_libtorch = self.spec.satisfies("+pytorch")`, to have a more consistent logic.

In fact, having `+metatomic` as a variant will trigger `+pytorch` thanks to the `conflicts` above; and in turn having `+pytorch` will trigger `enable_libtorch`. Hence, `+metatomic` will always result in having `enable_libmetatomic` true, even without the additional `or` statement.

Does it make sense, or I have missed something on my side? Thanks!

marcodelapierre (approved):
Please consider my comment on line 262 of the Plumed recipe

repos/spack_repo/builtin/packages/plumed/package.py
RMeli:
```suggestion
        enable_libtorch = self.spec.satisfies("+pytorch")
```

I think I was trying to be more explicit, but agree it's better to have the `conflicts` do the work.

marcodelapierre (approved):

adamjstewart:
@haampie just removed the `libs` property from the PyTorch package. Can you explain why this is necessary?

RMeli:
@adamjstewart it is necessary because `plumed` doesn't use CMake and therefore one has to use `search_flags` and `link_flags` to properly set things up. I can do it manually as I had to do for the link flags (because the way `HeaderList` works doesn't seem to allow for `include` subdirectories), but thought it was cleaner to add the property to the PyTorch package itself. Is there any reason why it had to be removed @haampie?

adamjstewart:
We removed `libs`/`headers` from all Python packages because 1) it isn't used 99% of the time, and 2) it's very slow if using external packages. However, it sounds like this falls under the 1% and shouldn't be too slow. I'm personally fine with re-adding it.

junghans (dismissed):

RMeli:
I was able to reproduce the [plumed+metatomic](https://atomistic-cookbook.org/examples/metatomic-plumed/metatomic-plumed.html) with the `plumed` package updated in this PR, therefore I think this is good to go.

marcodelapierre (commented):
There is one CI check that is failing. Any clue?  Thanks

RMeli:
@spackbot fix style

kwryankrattiger:
@spackbot run pipeline

haampie (approved):

PR #1992, opened 2025-10-15
napari: new package

Markus92:
Napari is a fast, interactive viewer for multi-dimensional images in Python (according to their developers at least). It's a pretty nifty tool for image visualization and manipulation, see also https://napari.org

It has a whole hoist of dependencies that were not in Spack yet. To prevent from flooding you with a whole bunch of separate pull requests, I added them all in this one but as separate commits.

There is not just new packages but also some changes. Unfortunately, only scikit-image has an activate maintainer listed: @adamjstewart and the others look pretty much unmaintained.

Let me know if this is acceptable, I can split it off in separate pull requests if necessary,

Last but not least: napari is, of course, a depency for Relion.


Markus92:
@spackbot fix style
(should just do this on my own machine...)

adamjstewart (approved):
scikit-image changes look good, will let someone else review the other packages.

PR #2002, opened 2025-10-16
Nat/enhance gpu support

natshineman:
Removes some outdated options from the package and provides some more options and dependencies for GPU enabled builds.

Also adds the "cray" PM for builds on Cray systems with slurm as the launcher. These builds require different PMI libraries from standard slurm builds.


tldahlgren:
Added missing maintainers to Reviewer list.

tldahlgren:
@spackbot fix style

tldahlgren (approved):
Can't confirm `mvapich-plus` sha256 (since manual download).

Assuming spackbot fixes the style, is there any reason to delay merging this PR (e.g., feedback from other maintainers wanted)?

PR #2015, opened 2025-10-19
LinuxPerf: fix python-config not found for external Python

ExplorerRay:
Original setting `python-config` is not available for Python package which is found externally (no symlink python-config -> python3-config). Therefore, I add `which` to check `python3-config` which is commonly used for external python package and avoid `python-config` command not found issue.


ExplorerRay:
@spackbot fix style

tldahlgren:
Added missing maintainer.

Jordan474 (approved):

tldahlgren (approved):

PR #2023, opened 2025-10-21
flecsi: add v2.4.1

rbberger:


opensdh (approved):
Do we already want to deprecate 2.4.0?  (Unlike 2.3.0 and 2.3.1, it doesn't have a build failure, just the usual sort of bugs.)

tldahlgren (approved):
Confirmed commit+tag exist and the variant associated with the dependency changes is defined.

PR #2026, opened 2025-10-21
nghttp2: add v1.66.0, v1.67.0, v1.67.1

m-fila:


tldahlgren (approved):
Confirmed the version sha256s.

PR #2029, opened 2025-10-21
openssl: add recent versions, deprecate older versions

m-fila:
The v3.1 release reached its End-of-Life and is also deprecated


michaelkuhn (approved):

PR #2032, opened 2025-10-21
detray: add v0.104.0

stephenswat:
This commit adds version 0.104.0 of the detray package.


tldahlgren (approved):
Confirmed the version sha256.

PR #2035, opened 2025-10-21
Handle checkout vars passed by environment

kwryankrattiger:
Get environment variables from environment. Previous testing didn't require the reference dotenv file exist.

### DIFF
--- a/.ci/gitlab/scripts/common/write_dotenv.py
+++ b/.ci/gitlab/scripts/common/write_dotenv.py
@@ -33,13 +33,17 @@ if __name__ == "__main__":
     dotenv = {}
     base_dotenv = sys.argv[1] if len(sys.argv) > 1 else None
     if base_dotenv:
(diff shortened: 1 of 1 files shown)


kwryankrattiger:
Tested here: https://gitlab.spack.io/spack/spack/-/pipelines/1277301

.ci/gitlab/scripts/common/write_dotenv.py
```diff
@@ -35,11 +35,15 @@ def write_dotenv(env, file):
     if base_dotenv:
         if not os.path.exists(base_dotenv):
```
zackgalbreath: can we rewrite this to avoid the double negative?

```
if os.path.exists(base_dotenv):
...
else:
...
```

zackgalbreath (changes requested):

zackgalbreath (approved):

PR #2041, opened 2025-10-21
Package for ERF (Energy Research and Forecasting) software

larenspear:
Git repo for software can be found at https://github.com/erf-model/ERF.

https://erf.readthedocs.io/en/latest/index.html


repos/spack_repo/builtin/packages/erf/package.py
```diff
@@ -0,0 +1,174 @@
...
+    version("24.08", tag="25.08", submodules=True)
+    version("24.06", tag="25.08", submodules=True)
+    version("24.05", tag="25.08", submodules=True)
+    version("24.04", tag="25.08", submodules=True)
+    version("24.03", tag="25.08", submodules=True)
+    version("24.02", tag="25.08", submodules=True)
+    version("24.01", tag="25.08", submodules=True)
+    version("23.12", tag="25.08", submodules=True)
+    version("23.11", tag="25.08", submodules=True)
+    version("23.10", tag="25.08", submodules=True)
```
becker33: It's preferable if you can include the commit reference in addition to the tag for tag-based versions. Also, it looks like you updated the versions but not the tags when you added the older versions.

repos/spack_repo/builtin/packages/erf/package.py
```diff
@@ -0,0 +1,174 @@
...
+
+    variant("mpi", default=False, description="Enable MPI support")
+    variant("openmp", default=False, description="Enable OpenMP support")
+    variant("hip", default=False, description="Enable HIP/rocm support")
```
becker33: The `RocmPackage` has an implicit `rocm` variant that you should use instead of adding this one.

repos/spack_repo/builtin/packages/erf/package.py
```diff
@@ -0,0 +1,174 @@
...
+    depends_on("cmake@3.20:", type="build")
+    depends_on("git", type="build")
+    depends_on("amrex", type=("build", "link", "run"))
+    for v in ("mpi", "openmp", "cuda", "sycl", "particles"):
+        depends_on(f"amrex+{v}", when=f"+{v}")
+        depends_on(f"amrex~{v}", when=f"~{v}")
+    for sm in CudaPackage.cuda_arch_values:
+        depends_on(f"amrex+cuda cuda_arch={sm}", when=f"+cuda cuda_arch={sm}")
+    depends_on("amrex+rocm", when="+hip")
+    depends_on("amrex~rocm", when="~hip")
```
becker33:
```suggestion
    depends_on("amrex", type=("build", "link", "run"))
    for v in ("mpi", "openmp", "cuda", "sycl", "particles"):
        depends_on(f"amrex+{v}", when=f"+{v}")
        depends_on(f"amrex~{v}", when=f"~{v}")
    for sm in CudaPackage.cuda_arch_values:
        depends_on(f"amrex+cuda cuda_arch={sm}", when=f"+cuda cuda_arch={sm}")
    depends_on("amrex+rocm", when="+hip")
    depends_on("amrex~rocm", when="~hip")
```
You need to either repeat the type arguments for all of the dependencies they apply to, or use a `with default_args(type=("build", "link", "run")):` context manager around them.

repos/spack_repo/builtin/packages/erf/package.py
```diff
@@ -0,0 +1,174 @@
...
+            self.define_from_variant("ERF_ENABLE_PARTICLES", "particles"),
+            self.define_from_variant("ERF_ENABLE_MULTIBLOCK", "multiblock"),
+            self.define_from_variant("ERF_ENABLE_RADIATION", "radiation"),
+            self.define_from_variant("ERF_BUILD_TESTS", "tests"),
+            self.define_from_variant("ERF_BUILD_FCOMPARE", "fcompare"),
+            self.define_from_variant("ERF_ENABLE_FFT", "fft"),
+            self.define("ERF_DIM", "3"),
+        ]
+        args.append(self.define("ERF_USE_EXTERNAL_AMREX", False))
+        args.append(self.define("ERF_CLONE_AMREX", True))
```
becker33: Please make Spack responsible for downloading amrex, rather than relying on the cmake system. This will make the package play much more nicely with spack features like source mirrors.

repos/spack_repo/builtin/packages/erf/package.py
```diff
@@ -0,0 +1,174 @@
...
+        args.append(self.define("GIT_SUBMODULE_PROTOCOL", "https"))
```
becker33: Does the ERF build system run submodules even after the whole source (including submodules) is downloaded? I don't think this line should be necessary.

repos/spack_repo/builtin/packages/erf/package.py
```diff
@@ -0,0 +1,174 @@
...
+        args.append(self.define("MPIEXEC_PREFLAGS", "--oversubscribe"))
+        args.append(self.define("CMAKE_BUILD_TYPE", "Release"))
```
becker33: Rely on the `build_type` variant instead for this -- if debug builds are not possible, redefine the variant so that only the `Release` value is a valid value.

repos/spack_repo/builtin/packages/erf/package.py
```diff
@@ -0,0 +1,174 @@
...
+            args.extend(
+                [
+                    self.define("NetCDF_C_PATH", self.spec["netcdf-c"].prefix),
+                    self.define("NetCDF_FORTRAN_PATH", self.spec["netcdf-fortran"].prefix),
+                ]
+            )
+
+        if "+radiation" in self.spec:
+            if "+cuda" in self.spec:
+                args.append(self.define("Kokkos_ARCH_AMPERE80", True))
```
becker33: This should not be hardcoded to `ampere80`, it should probably take its value from the `cuda_arch` variant.

If `ampere` is the only valid cuda target for the package, please override the `cuda_arch` variant to ensure it's the only valid value. We shouldn't let people try to build for a different arch and "secretly" give them an ampere build without errors.

repos/spack_repo/builtin/packages/erf/package.py
```diff
@@ -0,0 +1,174 @@
...
+            args.append(self.define("CMAKE_CUDA_ARCHITECTURES", ";".join(archs)))
+            args.extend(
+                [
+                    self.define("ERF_ENABLE_CUDA", "ON"),
+                    self.define("CUDAToolkit_ROOT", self.spec["cuda"].prefix),
+                ]
+            )
+
+        if "+hip" in self.spec:
+            args.append(self.define("CMAKE_HIP_ARCHITECTURES", "gfx906;gfx908;gfx90a"))
```
becker33: Same for these being hardcoded, use the `amdgpu_target` variant

repos/spack_repo/builtin/packages/erf/package.py
```diff
@@ -0,0 +1,174 @@
...
+
+        print(args)
```
becker33: The arguments are already logged without needing to additionally print them, please remove the debug print statement when you're done with it.

repos/spack_repo/builtin/packages/erf/package.py
```diff
@@ -0,0 +1,174 @@
...
+        return args
+
+    def setup_build_environment(self, env):
+        super(Erf, self).setup_build_environment(env)
```
becker33:
```suggestion
        super().setup_build_environment(env)
```
Python 3 lets us be a little less verbose with this.

repos/spack_repo/builtin/packages/erf/package.py
```diff
@@ -0,0 +1,174 @@
...
+        if "+openmp" in self.spec:
+            if "%clang" in self.spec or "%gcc" in self.spec:
+                env.append_flags("CFLAGS", self.compiler.openmp_flag)
+                env.append_flags("CXXFLAGS", self.compiler.openmp_flag)
+                env.append_flags("FFLAGS", self.compiler.openmp_flag)
+            elif "%intel" in self.spec:
+                env.append_flags("CFLAGS", "-qopenmp")
+                env.append_flags("CXXFLAGS", "-qopenmp")
+                env.append_flags("FFLAGS", "-qopenmp")
+                print(self.compiler.openmp_flag)
```
becker33:
```suggestion
        if "+openmp" in self.spec:
            env.append_flags("CFLAGS", self["c"].openmp_flag)
            env.append_flags("CXXFLAGS", self["cxx"].openmp_flag)
            env.append_flags("FFLAGS", self["fortran"].openmp_flag)
```
If this can't be handled properly by the `intel-oneapi-compilers-classic` package (that provides the intel compiler) then we should modify that package to properly handle the `openmp_flag` option.

If that's necessary and you don't feel comfortable adding it to this PR yourself, let me know and I can work that in.

repos/spack_repo/builtin/packages/erf/package.py
```diff
@@ -0,0 +1,174 @@
...
+
+    def configure_args(self):
+        args = super(Erf, self).configure_args()
+        return args
```
becker33: This package doesn't need a `configure_args` option, since it's not using a build system that uses `configure`.

becker33 (changes requested):

repos/spack_repo/builtin/packages/erf/package.py
```diff
@@ -0,0 +1,161 @@
...
+                [
+                    self.define("NetCDF_C_PATH", self.spec["netcdf-c"].prefix),
+                    self.define("NetCDF_FORTRAN_PATH", self.spec["netcdf-fortran"].prefix),
+                ]
+            )
+
+        if "+radiation" in self.spec:
+            args.extend(
+                [
+                    self.define("ERF_ENABLE_KOKKOS", True),
```
becker33: This seems to imply that erf needs a `depends_on("kokkos", when="+radiation")` added above

becker33 (changes requested):

becker33 (dismissed):
I'm still a little concerned that the git config implies a TPL that's being hidden by the build system, but we can always improve it later if there's an issue

larenspear:
> I'm still a little concerned that the git config implies a TPL that's being hidden by the build system, but we can always improve it later if there's an issue

As it turns out you're correct - cloning submodules and specifying the Spack AMReX instead of the submodule AMReX works, but choosing not to clone submodules doesn't work. May need to make some changes upstream in ERF. Regardless, will be taking this back to the drawing board to some extent. Will solicit help from the Slack.

becker33:
If all it needs is submodules, you can put `submodules=True` on the version directive to have it clone submodules before Spack turns control over to the build system.  But if those submodules are the full source for TPLs then that's not as effective because it negates Spack's ability to protect you from loader races that come from incompatible dependencies.

larenspear:
> If all it needs is submodules, you can put `submodules=True` on the version directive to have it clone submodules before Spack turns control over to the build system. But if those submodules are the full source for TPLs then that's not as effective because it negates Spack's ability to protect you from loader races that come from incompatible dependencies.

So a few things:

1: ERF directly depends on AMReX - that one I have covered. I ran `ldd` on the binary and confirms it links against Spack's AMReX.

2: For some configurations, ERF also depends on NoahMP (which does not have a Spack package) as well as E3SM's EKAT (which does not have a spack package, and also depends on a custom fork of kokkos). Does that mean I need to write Spack packages for NoahMP and EKAT before I can tackle ERF +radiation?

becker33:
For (2), yes that's probably the best way forward -- I think it makes sense to merge this without the `radiation` option (and without the git config for submodules) and then tackle that option as a separate PR.

I'm less worried about the ones that don't have spack packages, but if you link the custom fork of kokkos into the same dependency DAG as a normal kokkos you're going to end up with loader races.

becker33 (approved):

PR #2054, opened 2025-10-22
Relion: update for Relion 5.0.1

Markus92:
Hi all,

Took some time and effort, here is a draft PR for updating Relion to 5.0.1

It's almost finished bar a few things: some CUDA stuff (cuda_arch propagation mainly) and setting the TORCH_HOME directory for downloading the weights for some of the dependencies. It's useful and appears to be working otherwise.

Mainly putting this out here to prevent duplicate efforts while finishing the final things up.

This PR touches a lot of packages. However, none of them have an active maintainer listed (except scikit-image but that's already in PR #1992 where I still need to fix a small thing or two up).

Thanks for all the help on Slack!


Markus92:
@spackbot fix style

Markus92:
@spackbot fix style

repos/spack_repo/builtin/packages/aretomo/package.py
```diff
@@ -0,0 +1,50 @@
...
+
+    # AreTomo can be downloaded from the UCSF website
+    # https://msg.ucsf.edu/software
+    # AreTomo is not open-source. It is free for academic use only.
+    homepage = "https://doi.org/10.1016/j.yjsbx.2022.100068"
+    manual_download = True
+
+    maintainers("Markus92")
+
+    version_date = {"1.3.4": "Feb222023"}
```
becker33: This variable doesn't appear to be used anywhere.

repos/spack_repo/builtin/packages/py_app_model/package.py
```diff
@@ -0,0 +1,41 @@
...
+    depends_on("py-pydantic@2.8:", when="@0.5:", type=("build", "run"))
+    depends_on("py-pydantic@1.10.18:", when="@0.4", type=("build", "run"))
+    depends_on("py-pydantic@1.8.2:1", when="@0.1.4:0.1", type=("build", "run"))
+    depends_on("py-pydantic-compat@0.1.1:", when="@0.4", type=("build", "run"))
+    depends_on("py-in-n-out@0.1.5:", type=("build", "run"))
+    depends_on("py-typing-extensions@4.12:", type=("build", "run"))
+
+    with default_args(when="+qt", type=("build", "run")):
+        depends_on("py-qtpy@2.4.0:")
+        depends_on("py-superqt@0.7.2: +iconify")
```
becker33: [minor] this might be easier to read

```suggestion
    with default_args(type=("build", "run")):
        depends_on("py-psygnal@0.10:", when="@0.4:")
        depends_on("py-psygnal@0.3.4:")
        depends_on("py-pydantic@2.8:", when="@0.5:")
        depends_on("py-pydantic@1.10.18:", when="@0.4")
        depends_on("py-pydantic@1.8.2:1", when="@0.1.4:0.1")
        depends_on("py-pydantic-compat@0.1.1:", when="@0.4")
        depends_on("py-in-n-out@0.1.5:")
        depends_on("py-typing-extensions@4.12:")
        
        with when("+qt"):
            depends_on("py-qtpy@2.4.0:")
            depends_on("py-superqt@0.7.2: +iconify")
```

repos/spack_repo/builtin/packages/py_cachey/package.py
```diff
@@ -0,0 +1,25 @@
...
+    homepage = "http://github.com/dask/cachey/"
+    pypi = "cachey/cachey-0.2.1.tar.gz"
+
+    maintainers("Markus92")
+
+    license("BSD-3-Clause", checked_by="Markus92")
+
+    version("0.2.1", sha256="0310ba8afe52729fa7626325c8d8356a8421c434bf887ac851e58dcf7cf056a6")
+
+    depends_on("python@3.6:", type=("build", "run"))
```
becker33, line 21: The version constraint here isn't necessary since Spack doesn't support older pythons anyway. It's not necessary to remove it, but I think it will have a very minor effect on concretizer performance and those add up.

repos/spack_repo/builtin/packages/py_hsluv/package.py
```diff
@@ -0,0 +1,24 @@
...
+    homepage = "https://www.hsluv.org"
+    pypi = "hsluv/hsluv-5.0.4.tar.gz"
+
+    maintainers("Markus92")
+
+    license("MIT", checked_by="Markus92")
+
+    version("5.0.4", sha256="2281f946427a882010042844a38c7bbe9e0d0aaf9d46babe46366ed6f169b72e")
+
+    depends_on("python@3.7:", type=("build", "run"))
```
becker33, line 22: Same here for the version

repos/spack_repo/builtin/packages/py_napari/package.py
```diff
@@ -0,0 +1,102 @@
...
+    depends_on("py-setuptools-scm@3.4: +toml", type="build")
+
+    depends_on("py-appdirs@1.4.4:", type=("build", "run"))
+    depends_on("py-app-model@0.4.0:0.4", when="@0.6:", type=("build", "run"))
+    depends_on("py-app-model@0.1:0.2", when="@0.4:0.5", type=("build", "run"))
+    depends_on("py-cachey@0.2.1:", type=("build", "run"))
+    depends_on("py-certifi@2018.1.18:", type=("build", "run"))
+    depends_on("py-dask@2021.10.0: +array", type=("build", "run"))
+    depends_on("py-imageio@2.20:", type=("build", "run"))
+    conflicts("^py-imageio@2.28.0")
```
becker33: [minor] this can be simplified
```suggestion
    depends_on("py-imageio@2.20:2.27,2.28.1:", type=("build", "run"))
```

repos/spack_repo/builtin/packages/py_napari/package.py
```diff
@@ -0,0 +1,102 @@
...
+    depends_on("py-magicgui@0.7.0:", when="@0.6:", type=("build", "run"))
+    depends_on("py-magicgui@0.3.6:", type=("build", "run"))
+    depends_on("py-numpy@1.24.2:", when="@0.6:", type=("build", "run"))
+    depends_on("py-numpy@1.21:", type=("build", "run"))
+    depends_on("py-numpydoc@1.0.0:", when="@0.6:", type=("build", "run"))
+    depends_on("py-numpydoc@0.9.2:", type=("build", "run"))
+    depends_on("py-pandas@1.3.3:", type=("build", "run"))
+    depends_on("py-pillow@9.0:", when="@0.6:", type=("build", "run"))
+    conflicts("^py-pillow@7.1.0")
+    conflicts("^py-pillow@7.1.1")
```
becker33: These seem redundant

repos/spack_repo/builtin/packages/py_napari/package.py
```diff
@@ -0,0 +1,102 @@
...
+    variant("all", default=False, description="Install all optional dependencies")
+
+    depends_on("python@3.10:", type=("build", "run"))
+
+    depends_on("py-setuptools@69:", when="@0.6:", type="build")
+    depends_on("py-setuptools@56:", type="build")
+    depends_on("py-setuptools-scm@8: +toml", when="@0.6:", type="build")
+    depends_on("py-setuptools-scm@3.4: +toml", type="build")
+
+    depends_on("py-appdirs@1.4.4:", type=("build", "run"))
```
becker33: A `with default_args(type=("build", "run")):` block here might make things easier to maintain.

repos/spack_repo/builtin/packages/py_relion/package.py
```diff
@@ -0,0 +1,74 @@
...
+    maintainers("Markus92")
+
+    license("GPL-2", checked_by="Markus92")
+
+    version("5.0.1", sha256="acbf898e96513b092514a56ff2a255c69a795e7a6f04131eacc8f55e2a900c23")
+    version("5.0.0", sha256="5d02d529bfdb396204310b35963f35e5ec40ed9fd10bc88c901119ae7d7739fc")
+
+    variant("cuda", default=True, description="Build with CUDA (recommended)")
+
+    depends_on("python@3.10")
```
becker33, line 31: I think a bunch of the deptypes in this list are probably wrong -- the default deptype is `build/link`, which is unlikely for python components.

I'm also confused by the when clauses -- they apply to all of the versions in the package currently, are there other versions you intend to add that make them relevant?

repos/spack_repo/builtin/packages/py_superqt/package.py
```diff
@@ -15,15 +15,24 @@ class PySuperqt(PythonPackage):
...
     depends_on("python@3.8:", type=("build", "run"))
     depends_on("py-hatchling", type="build")
     depends_on("py-hatch-vcs", type="build")
 
     depends_on("py-packaging", type=("build", "run"))
     depends_on("py-pygments@2.4:", type=("build", "run"))
+    depends_on("py-qtpy@2.4:", when="@0.7.6:", type=("build", "run"))
     depends_on("py-qtpy@1.1:", type=("build", "run"))
-    depends_on("py-typing-extensions@3.7.4.3:", type=("build", "run"))
+    depends_on("py-typing-extensions@4.12:", when="@0.7.6: ^python@3.13:", type=("build", "run"))
```
becker33: Is this actually a `py-superqt` constraint, or is this a misplaced constraint that should be on `py-typing-extensions`? If it's the latter, we can remove it here because adding it there will ensure the compatibility we need.

repos/spack_repo/builtin/packages/relion/package.py
```diff
@@ -17,11 +17,12 @@ class Relion(CMakePackage, CudaPackage):
     homepage = "https://www2.mrc-lmb.cam.ac.uk/relion"
     git = "https://github.com/3dem/relion.git"
     url = "https://github.com/3dem/relion/archive/4.0.0.zip"
-    maintainers("dacolombo")
+    maintainers("Markus92")
```
becker33: It's probably better to leave the other maintainer listed unless they say they want to be removed

```suggestion
    maintainers("dacolombo", "Markus92")
```

repos/spack_repo/builtin/packages/relion/package.py
```diff
@@ -63,13 +64,14 @@ class Relion(CMakePackage, CudaPackage):
...
     # do not seem to cause problems with < 3
-    variant("mklfft", default=True, description="Use MKL rather than FFTW for FFT")
+    variant("mklfft", default=False, description="Use MKL rather than FFTW for FFT")
     variant(
         "allow_ctf_in_sagd",
         default=True,
         description=(
             "Allow CTF-modulation in SAGD, as specified in Claim 1 of patent US10,282,513B2"
         ),
+        when="@3",
```
becker33, line 80: Are these only for version 3, or should this be @3:?

becker33 (changes requested):

repos/spack_repo/builtin/packages/aretomo/package.py
Markus92: Removed

repos/spack_repo/builtin/packages/py_cachey/package.py
Markus92: The reason I left it in is we have re-instated some of the older Python versions to our current system at explicit user request (python 2 even...); I can remove it if needed

repos/spack_repo/builtin/packages/py_napari/package.py
Markus92: They came from from here: https://github.com/napari/napari/blob/82a6f31b5c29602a02bb19b7e35ec59e9ed03a64/setup.cfg#L63C1-L63C88
Not sure if Spack would realize that it applies to downstream packages though?

repos/spack_repo/builtin/packages/relion/package.py
Markus92: This flag only exists in Relion 3 from what I could find.

repos/spack_repo/builtin/packages/py_superqt/package.py
Markus92: So I ported it over from https://github.com/pyapp-kit/superqt/blob/v0.7.6/pyproject.toml#L51, might be better to handle it in the other package. I removed it for now

repos/spack_repo/builtin/packages/py_napari/package.py
Markus92: Agreed and done!

repos/spack_repo/builtin/packages/py_relion/package.py
Markus92: You're right, I did miss that.

No intention for other versions as of now, Relion 4 and below did not have an associated Python environment. I removed the version constraints to make life a little bit easier for the concretizer.

Markus92:
@spackbot fix style

Markus92:
@becker33 Thank you so much for taking the time to go through this way too big pull request!

I left the Python requirements in for selfish reasons (see above) but can remove them if you find them not clean. In practice, the `py-relion` package pins the Python minor version so it won't matter too much for us.

repos/spack_repo/builtin/packages/py_cachey/package.py
becker33: I will defer to the `spack audit packages` command on this -- if the audit doesn't flag this as a problem, I can live with it.

repos/spack_repo/builtin/packages/py_superqt/package.py
```diff
@@ -15,15 +15,23 @@ class PySuperqt(PythonPackage):
...
     depends_on("py-hatchling", type="build")
     depends_on("py-hatch-vcs", type="build")
 
     depends_on("py-packaging", type=("build", "run"))
     depends_on("py-pygments@2.4:", type=("build", "run"))
+    depends_on("py-qtpy@2.4:", when="@0.7.6:", type=("build", "run"))
     depends_on("py-qtpy@1.1:", type=("build", "run"))
-    depends_on("py-typing-extensions@3.7.4.3:", type=("build", "run"))
+    depends_on("py-typing-extensions@4.5:", when="@0.7.6:", type=("build", "run"))
+    depends_on("py-typing-extensions@3.7.4.3:", when="@0.6.1", type=("build", "run"))
```
becker33: Why not move these other `build/run` deps into the context manager below instead of having to type it out every time?

repos/spack_repo/builtin/packages/relion/package.py
```diff
@@ -17,11 +17,12 @@ class Relion(CMakePackage, CudaPackage):
...
     git = "https://github.com/3dem/relion.git"
     url = "https://github.com/3dem/relion/archive/4.0.0.zip"
-    maintainers("dacolombo")
+    maintainers("dacolombo", "Markus92")
 
     license("GPL-2.0-only")
 
-    version("5.0.0", sha256="800ad0c0aa778cbf584fcf8986976645f2b25d677a80f168e5397975b9db6e47")
+    version("5.0.1", sha256="3253230cd4b3d9633a5cac906937039b9971eb9430c3e2d838473777fb811f4c")
+    # version("5.0.0", sha256="800ad0c0aa778cbf584fcf8986976645f2b25d677a80f168e5397975b9db6e47")
```
becker33: Is this version commentd out for a reason? If you're removing it, it should be marked as deprecated for a release before being removed.

becker33 (changes requested):

becker33:
@Markus92 mostly looks good now, feel free to ping me on slack if either of these needs more discussion.

repos/spack_repo/builtin/packages/relion/package.py
Markus92: I'll deprecate it then. There's a few bugs in 5.0.0 that are fixed in 5.0.1; The current 5.0.0 doesn't fully function anyways (tomography packages are broken), which was the main reason for this entire undertaking :)

repos/spack_repo/builtin/packages/py_superqt/package.py
Markus92: I'll go over it and do all of them. Tried a bit too hard to try out all VIM shortcuts I guess.

Markus92:
@becker33 thanks again so much. I've done some more cleanup work, hopefully all good now.

The package audit didn't complain about the Python versions so I hope we can just leave it as is for now.

becker33 (approved):

samfux84:
@Markus92 : Is there a particular reason why you set FETCH_WEIGHTS=OFF for version 5 and newer? Aren't those weights needed for the Python packages?

Markus92:
@samfux84 by default they get stored in the home directory, not package directory and even if we were to patch that, I'm not sure if Pytorch expects that folder to be writable by the end user. Some of our users told me it fetches them at runtime and it gets stored in their own home directory which is fine.

samfux84:
Thank you. I figured out that by setting -DTORCH_HOME_PATH, you can specify a directory where the weights are stored.

I don't like the idea that every user has to store 12 GB of data in their home for this (waste of storage if many users are using the software). I stored it now in a central place and asked a user to test. Let's see. If torch needs to write to this directory, then I agree that the user home is a better place for this, but then the developers should better split it, such that the weights can be stored in a central place and torch should use another directory in the user home to store whatever it needs to store.

Markus92:
@samfux84 completely agree! Let me know how it goes, if it doesn't try to write in it during runtime we can definitely update the package to store these weights centrally.

PR #2057, opened 2025-10-22
Updates to GL related packages #192

kwryankrattiger:
Extracted updates to GL detection from #192

### DIFF
--- a/repos/spack_repo/builtin/packages/apple_glu/package.py
+++ b/repos/spack_repo/builtin/packages/apple_glu/package.py
@@ -14,3 +14,8 @@ class AppleGlu(AppleGlBase):
     provides("glu@1.3")
 
     requires("platform=darwin", msg="Apple-GL is only available on Darwin")
+
+    def setup_dependent_build_environment(
+        self, env: EnvironmentModifications, dependent_spec: Spec
+    ):
+        env.prepend_path("OpenGL_ROOT", self.prefix)
--- a/repos/spack_repo/builtin/packages/mesa/package.py
+++ b/repos/spack_repo/builtin/packages/mesa/package.py
@@ -176,6 +176,9 @@ class Mesa(MesonPackage):
                 flags.append("-std=c99")
         return super().flag_handler(name, flags)
 
+    def dependent_cmake_args(self, dependent_spec: Spec) -> List[str]:
+        return ["-DOpenGL_GL_PREFERENCE:STRING=LEGACY"]
+
     @property
     def libglx_headers(self):
         return find_headers("GL/glx", root=self.spec.prefix.include, recursive=False)
--- a/repos/spack_repo/builtin/packages/mesa_glu/package.py
+++ b/repos/spack_repo/builtin/packages/mesa_glu/package.py
@@ -26,6 +26,11 @@ class MesaGlu(AutotoolsPackage):
     # patch switches all instances of register long to long to fix this.
     patch("register-long.patch")
 
+    def setup_dependent_build_environment(
+        self, env: EnvironmentModifications, dependent_spec: Spec
+    ):
+        env.prepend_path("OpenGL_ROOT", self.prefix)
+
     def configure_args(self):
         args = ["--disable-libglvnd"]
 
--- a/repos/spack_repo/builtin/packages/openglu/package.py
+++ b/repos/spack_repo/builtin/packages/openglu/package.py
@@ -64,6 +64,11 @@ class Openglu(Package):
     def fetcher(self):
         _ = self.fetcher
 
+    def setup_dependent_build_environment(
+        self, env: EnvironmentModifications, dependent_spec: Spec
+    ):
+        env.prepend_path("OpenGL_ROOT", self.prefix)
+
     @property
     def libs(self):
         return find_libraries("libGLU", self.prefix, shared=True, recursive=True)


repos/spack_repo/builtin/build_systems/cmake.py
```diff
@@ -60,6 +60,13 @@ def _maybe_set_python_hints(pkg: PackageBase, args: List[str]) -> None:
     )
 
 
+def _dependencies_cmake_extra_args(pkg: PackageBase, args: List[str]) -> None:
+    for dep in pkg.spec.traverse(deptype=("build", "link")):
+        extra_args = getattr(dep.package, "cmake_extra_args", None)
+        if extra_args:
+            args.extend(extra_args)
```
alalazo: Hmm, do we really need something this general only for mesa? Can't we use one of the other callbacks?

kwryankrattiger: I didn't see other callbacks for injecting cmake configure flags. And this is can be used for any package that doesn't provide a cmake config file, which is many, to control options in built-in cmake find modules. Mesa just happens to be the first one.

I can't use link flags, because that will still result in the current issue where `libOpenGL` and `libGL` both get linked, and then cause runtime errors.

I am open to other suggestions, but this is fixing one of the longest standing GL issues in Spack so I would rather not hold it up too long. And when I say long standing, I mean since CMake `FindOpenGL` started supporting GLVND (3.10, circa 2017).

Since it is packages repo API, there is no guarantee on stability, so if we get a better idea later there isn't really much harm in deprecating/dropping it.

kwryankrattiger: @alalazo do you have a better solution here? If not, I would really like to move this forward for the next release of packages as it fixes some bugs that have been around for a couple of years.

alalazo: I'm looking at this again, and it seems very controversial to me:
- Why `build, link` dependencies only? Don't we ever need run dependencies?
- Why do we go beyond direct dependencies?
- There is no way to turn off this behavior from a dependent, e.g. if we don't want extra arguments to be injected
- The name `cmake_extra_args` is confusing, since it's not clear these arguments will be injected to transitive dependencies

repos/spack_repo/builtin/build_systems/cmake.py
```diff
@@ -406,6 +413,7 @@ def std_args(pkg: PackageBase, generator: Optional[str] = None) -> List[str]:
 
         _conditional_cmake_defaults(pkg, args)
         _maybe_set_python_hints(pkg, args)
+        _dependencies_cmake_extra_args(pkg, args)
```
alalazo: > do you have a better solution here?

Can't we just deal with this in the same way we do for Python above, and code a `_maybe_set_mesa_args`? If we get more cases we can think about generalizing the mechanism. In any case it's probably sensible to stick to just direct deps.

repos/spack_repo/builtin/build_systems/cmake.py
kwryankrattiger: I agree this can probably be direct deps only. The reason for only build/link is I copied the requirement for python. I didn't convert python to do this because it was outside the immediate scope of the change, but I can to reduce the number of different ways this is handled.

If GL is provided by Spack built mesa, the _only_ option is LEGACY and not having this set for a `FindOpenGL` will result in borked builds. So there is no option for this, it is what it has to be.

I would say we have now increased the use case to two for this feature, both Mesa and Python. It would acually probably be nice to also provide this for MPI to set `MPI_C_COMPILER` and `MPI_CXX_COMPILIER` hints since that is something that is done in a number of places for fix `FindMPI`. So the usefulness of this type of feature is beyond just GL.

For the API and function, it would probably be named better. `set_dependent_cmake_args(self, pkg, args)`. This would allow something like this, so it is up to the provider whether or not it is optional.

```python
def set_dependent_cmake_args(self, pkg: PackageBase, cmake_args: List[str]):
    if getattr(pkg, "find_python_hints", True):
        return

    cmake_args.extend([...])
```

repos/spack_repo/builtin/build_systems/cmake.py
kwryankrattiger: I also want to do this for MPI, there are 27 packages where we do the same thing we do for python for MPI. That is at least three packages that I deal with prefequently enough that are trying to inject spack related hints into the CMake configure.

I think it makes some sense to put this on the packages themselves. a) it makes the `CMakePackage` more adaptable and b) changing the configure flags in the providing package is more likely to trigger a rebuild of the things that depended on it than changing the `CMakePackage`

repos/spack_repo/builtin/packages/python/package.py
```diff
@@ -1289,6 +1280,28 @@ def setup_dependent_package(self, module, dependent_spec):
...
+            return
+
+        python_executable = pkg.spec["python"].command.path
+        args.extend(
+            [
+                define("PYTHON_EXECUTABLE", python_executable),
+                define("Python_EXECUTABLE", python_executable),
+                define("Python3_EXECUTABLE", python_executable),
+            ]
+        )
```
alalazo: Hmm, not sure if it's wise that a dependent can potentially modify a list that is passed. It should just be able to return its own arguments.

repos/spack_repo/builtin/build_systems/cmake.py
```diff
@@ -42,22 +42,12 @@ def _extract_primary_generator(generator):
...
-    python_executable = pkg.spec["python"].command.path
-    args.extend(
-        [
-            define("PYTHON_EXECUTABLE", python_executable),
-            define("Python_EXECUTABLE", python_executable),
-            define("Python3_EXECUTABLE", python_executable),
-        ]
-    )
+def _dependencies_cmake_extra_args(pkg: PackageBase, args: List[str]) -> None:
+    for dep in pkg.spec.dependencies(deptype=("build", "link")):
```
alalazo: We should not use arbitrarily `build,link` to traverse dependencies. Maybe it's better to decide when to perform a callback using the `UseMode` defined in `spack.build_environment`.

repos/spack_repo/builtin/build_systems/cmake.py
```diff
@@ -42,29 +42,12 @@ def _extract_primary_generator(generator):
...
-
-
 def _dependencies_cmake_extra_args(pkg: PackageBase, args: List[str]) -> None:
-    for dep in pkg.spec.traverse(deptype=("build", "link")):
-        extra_args = getattr(dep.package, "cmake_extra_args", None)
-        if extra_args:
-            args.extend(extra_args)
+    for dep in pkg.spec.dependencies(deptype=("build", "link")):
+        try:
+            dep.package.set_dependent_cmake_args(pkg, args)
```
alalazo: I think we should just call this `dependent_cmake_args` and it should return arguments, without having the power of potentially reordering or modifying them.

alalazo (dismissed):

repos/spack_repo/builtin/packages/python/package.py
kwryankrattiger: sure, I don't think there is a strong reason for packages to need that level of functionality. I was mostly just copying the interface of the old method here.

repos/spack_repo/builtin/build_systems/cmake.py
kwryankrattiger: Is `UseMode` in the package API? I am not seeing it imported in `package.py`.

I am also not sure how that would be used here, this particular function is run as a part of the setup for the cmake configure. My understanding is core shouldn't know anything about since it is a repo defined extension of the package API.

The choice to use `build, link` here is not arbitrary, we are looking at the dependencies that are needing during the build/link of the package. Packages that marked `run` shouldn't influence the build imo. If there is an exception to that later that can be addressed, but I am not thinking of any case for that now.

kwryankrattiger:
@spackbot run pipeline

repos/spack_repo/builtin/build_systems/cmake.py
alalazo: What about `test` dependencies? Also, is there no case where we should record a PATH or anything of a `run` dependency?

alalazo: > I am also not sure how that would be used here, this particular function is run as a part of the setup for the cmake configure. My understanding is core shouldn't know anything about since it is a repo defined extension of the package API.

In general core can prescribe new workflows, e.g. "build system methods listed in ... will be called at this point with this signature", so if we want to go for `UseMode` I guess we can.

repos/spack_repo/builtin/packages/python/package.py
```diff
@@ -1289,6 +1280,26 @@ def setup_dependent_package(self, module, dependent_spec):
         module.python_platlib = join_path(dependent_spec.prefix, self.platlib)
         module.python_purelib = join_path(dependent_spec.prefix, self.purelib)
 
+    def dependent_cmake_args(self, pkg: PackageBase) -> List[str]:
```
adamjstewart: I may be misunderstanding, but it seems like this injects flags to 99% of CMake packages that depend on Python unless they explicitly disable this? I don't love injecting non-standard flags, as they end up as red herring warning messages for downstream packages. I would much rather opt-in, and opt-in one flag at a time. `spec['python']` already has an easy and convenient way to grab the Python command. I think packages should manually use this only when needed.

kwryankrattiger: That is already true, I just moved the logic for this from `build_systems/cmake.py` to `python/package.py`. This change specifically _should_ amount to a noop.

Without these flags, cmake cannot find spack installed python. It is a similar issue for MPI, which is why you see so many cmake projects passing MPI hints in a few different iterations of the same general form (see the `mpi/package.py` changes for a flavor of it)

repos/spack_repo/builtin/build_systems/cmake.py
kwryankrattiger: > What about test dependencies?

I would classify those the same as `run` for injecting cmake args.

Is an example of a test-only dep that would need to inject cmake arguments? It seems very unlikely to be the case where a test dep that is not also a build or link dep would influence build system flags.

> if we want to go for UseMode I guess we can.

I still don't see how this would be better, or even implemented. We are injecting args into build system details that are more or less opaque to core. I am also uncertain how many other cases outside of CMake are going to have this kind of standard configure customization that can be applied like this.

repos/spack_repo/builtin/packages/python/package.py
adamjstewart: Thanks. I'm okay with no-ops, but I also think we should remove this and only add it to packages that need it. Doesn't have to be in this PR though.

kwryankrattiger: I am not sure what you mean by this.

These flags are only added to projects that directly depend on python (or whatever project has the cmake args function) and are CMake projects. Of those projects, very few _don't_ need to have these hints passed, and even fewer will have issues if the are passed. And for the projects where that is the case, we already of that opt-out option to set the package property `find_python_hints = False`.

adamjstewart: Very few need all 3 flags, and often only need them when `+python`.

kwryankrattiger: Sure. This is only applied when the spec has a python build or link dep. For packages with a python variant that will usually only be true for the `+python`. I have a feeling there are vanishingly few cases where the conditions for these flags are not met under those conditions.

We could be smarter about which flags we pass, they could be conditionally applied based on the CMake version. But since it doesn't hurt anything to pass all three all the time I don't know if there is value in making it more complicated.

kwryankrattiger:
@spackbot run pipeline

repos/spack_repo/builtin/build_systems/cmake.py
alalazo: This doesn't seem to be answered:
> Also, is there no case where we should record a PATH or anything of a run dependency?

I still feel like `build,link` is rather arbitrary as a choice. What if a tool needs `python` or `perl` only as a `run` dependency, and those packages want to inject some flag to CMake to be found properly?

repos/spack_repo/builtin/build_systems/cmake.py
```diff
@@ -42,22 +42,12 @@ def _extract_primary_generator(generator):
...
-            define("PYTHON_EXECUTABLE", python_executable),
-            define("Python_EXECUTABLE", python_executable),
-            define("Python3_EXECUTABLE", python_executable),
-        ]
-    )
+def _dependencies_cmake_extra_args(pkg: PackageBase, args: List[str]) -> None:
+    for dep in pkg.spec.dependencies(deptype=("build", "link")):
+        try:
+            args.extend(dep.package.dependent_cmake_args(pkg, args))
+        except Exception:
```
alalazo: This one is definitely too broad. Can we use a narrower list of exceptions?

alalazo: Maybe even better would be to not use exceptions at all:
```python
if not hasattr(dep.package, "dependent_cmake_args"):
    continue
```

repos/spack_repo/builtin/build_systems/cmake.py
alalazo: Thinking about it the best choice would probably be to iterate over all dependencies, and let each package decide when to inject arguments or not.

repos/spack_repo/builtin/build_systems/cmake.py
alalazo: Given the implementation, I think we can inline `_dependencies_cmake_extra_args`. 

Also, given we are generalizing the way dependents inject arguments, we should extract the `boost` related bits of `_conditional_cmake_defaults` and move them to the `boost` package.

repos/spack_repo/builtin/packages/apple_glu/package.py
```diff
@@ -14,3 +14,8 @@ class AppleGlu(AppleGlBase):
     provides("glu@1.3")
 
     requires("platform=darwin", msg="Apple-GL is only available on Darwin")
+
+    def setup_dependent_build_environment(
+        self, env: EnvironmentModifications, dependent_spec: Spec
+    ):
+        env.prepend_path("OpenGL_ROOT", self.prefix)
```
alalazo, line 21: Can this be extracted in a separate PR to be merged straight away?

repos/spack_repo/builtin/packages/cairo/package.py
```diff
@@ -143,7 +143,7 @@ class Cairo(AutotoolsPackage, MesonPackage):
     # patch from https://gitlab.freedesktop.org/cairo/cairo/issues/346
     patch("fontconfig.patch", when="@1.16.0:1.17.2")
     # Don't regenerate docs to avoid a dependency on gtk-doc
-    patch("disable-gtk-docs.patch", when="build_system=autotools ^autoconf@2.70:")
+    patch("disable-gtk-docs.patch", when="build_system=autotools")
```
alalazo: Can we extract also this seemingly unrelated change? If we are adding a new feature to the CMake build system, I'd like the PR to contain only changes relevant for that feature.

repos/spack_repo/builtin/packages/python/package.py
alalazo: > This is only applied when the spec has a python build or link dep. 

I think the Python package should have code to decide when to inject these flags, _if_ that is relevant. 

If there is any need to check e.g. the cmake version being used or the dependency types of `python`, before injecting Python flags, we should do it here. The code in the build system should probably be as agnostic as possible, and just perform a callback on packages implementing the method.

repos/spack_repo/builtin/packages/mpi/package.py
```diff
@@ -15,6 +15,23 @@ class Mpi(Package):
     homepage = "https://www.mpi-forum.org/"
     virtual = True
 
+    def dependent_cmake_args(self, pkg: PackageBase) -> List[str]:
```
alalazo: Is this actually being called? If I read code correctly we just iterate on `.dependencies`:
```python
Spack version 1.2.0.dev0
Python 3.13.2, Linux x86_64
>>> from spack.concretize import concretize_one
>>> s = concretize_one("hdf5")
>>> s.dependencies()
[cmake@3.31.6/ju2z2nr3nclj2ip2rhexrthq6mt7flt6, zlib-ng@2.2.4/3mewdwgmlykrd2ie4cftrpifzovbs2er, pkgconf@2.5.1/iguplooome2oktywbgjvoq5nwxmgdw6y, openmpi@5.0.9/tk7gfkssl7mkbglkbitxu6l4i3gpgysa, gcc@15.1/5mmswryndxgqgzxf4vhd27drooj2zywh, glibc@2.31/nm2n32h34mhcte6fm3s4fjjnnrdyqcvz, gcc-runtime@15.1/4obv4eoaimwzbfcgq3s7lf3wvrpd5emq, compiler-wrapper@1.0/tidruedm3z3lhfee2vyu3baxfpilprh7, gmake@4.4.1/x6dhwcecsmwwctcho5mmmvp4cmbryyqk]
```
and that doesn't include virtual packages, right?

alalazo (commented):
Left a few more comments. Summary:
1. I think we should use this PR only for changes related to the `cmake` build system, anything else should be extracted in a separate PR and merged
2. The build system should not make assumption on when to perform the callback depending on the dependency type. It should be on the called package to decide whether it wants to return something or not
3. No broad `Exception` catch in build systems, which could mask other errors

I also asked @haampie to give a second review to this PR. Since it touches a part of the repo that is used in many other places.

repos/spack_repo/builtin/packages/mpi/package.py
alalazo: I _think_ we need to iterate on edges and call the virtual package explicitly if the virtual is provided, if we want this to take effect (unless I'm forgetting some magic we do).

repos/spack_repo/builtin/build_systems/cmake.py
kwryankrattiger: oh wow, I think I was tired when I wrote that or something, 100% agree..

repos/spack_repo/builtin/build_systems/cmake.py
kwryankrattiger: The selection of `build` and `link` here are not arbitrary at all. As I pointed out previously I was quite intentional picking those.

My feeling is if a package is a `run` dep only then CMake shouldn't be finding them in the first place, or aware of them at all. If CMake is being used to generate runtime launcher scripts that run python or perl or something else then those should be labeled as `build` deps since we need the information about the python at build time to generate the files.

I think if there is ever a real case where we need to collect cmake flags from a dependency that is only labeled `run` then we can update the API accordingly, but I think the use case you are describing will likely only arise in ill-constrained dependencies and we should fix those dependency relationships rather than work around them with more complexity.

repos/spack_repo/builtin/packages/apple_glu/package.py
kwryankrattiger: Yeah I can do that.

repos/spack_repo/builtin/packages/cairo/package.py
kwryankrattiger: yeah makes sense, this was all migrated out of another PR, so what is a couple more 😄

repos/spack_repo/builtin/build_systems/cmake.py
alalazo: > Thinking about it the best choice would probably be to iterate over all dependencies, and let each package decide when to inject arguments or not.

:point_up: This is such that each package is responsible for deciding when to return arguments (the base class asks all dependencies and each one decides what to return). 

If a package knows it doesn't need to inject flags when it is a `run` dependency (why?) it can inspect `pkg` and return nothing. Likewise, if the flags we need are different depending on the version of CMake being used (PYTHON_EXECUTABLE?), we can inspect `pkg` and avoid injecting redundant flags.

Also, we don't account for `test` dependencies which may be used for build time tests, right? What if `python` is needed to run some test script etc ? Why are we creating a self imposed constraint instead of being agnostic of the dependency type and let the package decide?

repos/spack_repo/builtin/packages/python/package.py
```diff
@@ -1305,6 +1305,26 @@ def setup_dependent_package(self, module, dependent_spec):
...
+        """Set the {Python3,Python,PYTHON}_EXECUTABLE CMake variables explicitly if the package
+        has the variable ``find_python_hints`` set to True. This ensures that CMake locates the
+        right Python in its builtin FindPython3, FindPython, and FindPythonInterp modules.
+        Spack does CMake's job here because CMake's modules only search for Python versions known
+        at the time of release by default.
+        """
+
+        # Allow packages to disable these hints be setting a the member variable
+        # ``find_python_hints`` to False. By default this is always applied.
+        if not getattr(pkg, "find_python_hints", True):
```
alalazo: Another thought I had is that the `CMakePackage` class should define an attribute, for instance:
```python
class CMakePackage:
    disable_hints_from: List[str] = []
```
and we should check in the base class whether to make the callback or not. This would save use from inventing a new attribute name for each package defining this method, and having this check at the top of each implementation.

haampie: For me personally, I think the class attribute is fine, because:

1. Easier to document
2. Editor completion in derived classes from CMakePackage
3. I don't think the list will be long

repos/spack_repo/builtin/build_systems/cmake.py
haampie: I agree that `pkg.spec.dependencies()` with all deptypes is more straight-forward and least surprising.

> My feeling is if a package is a run dep only then CMake shouldn't be finding them in the first place

In practice this is not the case. There are enough instances of a run type dep where CMake still wants to know things about its executables. Sometimes people like to create absolute shebangs, etc. Specifying those are "build" deps would be wrong conceptually, as those are for the host, whereas the shebangs are for the target arch (would we ever do cross compiling)

Test deps should definitely be included.

`UseMode` should not be relevant here, given that the package being considered is always the one that's being built. `UseMode` is only relevant if you care about runtime vs build time and packages a few deps deep from the root.

haampie (dismissed):
At the very least, this PR should remove the MPI related changes. I think it will annoy people to see this compiler wrapper pop up. MPI is just a library like any other dependency.

In general, I would say: split the functionality you're introducing from the use cases you're contributing. Make the PR small, so it's merged faster.

repos/spack_repo/builtin/build_systems/cmake.py
kwryankrattiger: > If a package knows it doesn't need to inject flags when it is a run dependency (why?) it can inspect pkg and return nothing. Likewise, if the flags we need are different depending on the version of CMake being used (PYTHON_EXECUTABLE?), we can inspect pkg and avoid injecting redundant flags.

I just don't think this is a reasonable thing for a package to "know" as a dependency.  The only thing the build system or the dependency can "know" is if the package is needed for `build` or `link`, and really only `build`, that it should pass the  hint flags to cmake.

If a dependency is needed for `test`, then it will be properly configured in the environment when the _test_ runs. If the build is configuring the package to run tests then it is needed as part of `build` since it is being pulled in at `build` time to build the tests. If it is not, then the build shouldn't need to know anything about it.

Imagine the case below. Here we have a base test dependency for the test drivers to use some python. This python has nothing to do with the build. However, when the variant `+python` we do want to configure with python. Now this package has a policy of "if I can detect it, then enable the feature" policy. If we started injecting python into the build it would enable python even though python was not configured that way in the DAG. Since it is not marked as a `run` dependency now `python` is not going to be loaded correctly and our runtime will likely be broken/wrong.

```
depends_on("python", type="test")
with when("+python"):
    depends_on("python", type=("build", "test", "run"))
    depends_on("py-numpy", type="test")
```

It is up to a package to say, via the dependency type, _when_ a dependency should be affecting it. If it is not supposed to affect the `build` then we shouldn't be injecting it haphazardly into the build. By adding the ability for non-build dependencies to start injecting themselves into the build is more likely going to be problematic than convenient.

The part of about checking cmake-version or tertiary dependencies all make sense to happen in the `dependent_cmake_args` function. And there is information already passed to do so, so I don't really see the issue there.

repos/spack_repo/builtin/packages/python/package.py
kwryankrattiger: See my other comment. I am very much opposed to this.  The longer I think about it the more ways I find that is can end up creating broken packages. If a package needs to be known by the build system it will be present as a `build` or `link` dependency. If not, then we should not be injecting it into the build system at all.

kwryankrattiger:
@haampie 

> At the very least, this PR should remove the MPI related changes.

yeah, those were added in error, I have another PR to deal with that. 

> I think it will annoy people to see this compiler wrapper pop up. MPI is just a library like any other dependency.

Yeah, this is a feature of the FindMPI module. The wrappers are used to help with the find, they shouldn't be forced to be used by any of the packages. CC/CXX should remain what they were set to by the package.

> split the functionality you're introducing from the use cases you're contributing. Make the PR small, so it's merged faster.

Yeah, it is on my list to do so. There were enough comments on the functionality here already that I have not moved it. I was hoping to have some consensus on how this should work looking at the full picture before I broke it out. The GL/Python/MPI changes that use the feature here are meaningless without this change, and then provide context for the discussion.

alalazo:
> yeah, those were added in error, I have another PR to deal with that.

Were they? It seems to me they were listed in https://github.com/spack/spack-packages/pull/2057#discussion_r2515015866 as a motivation for going more general, instead of just coding a `_maybe_set_mesa_args` (which I think would be merged by now?).

kwryankrattiger:
> _maybe_set_mesa_args

The result of doing everything this way in an ad hoc manner will result in multiple `_maybe_set_*_args` for the cmake build system. This is hacky and will fail to scale as we have to patch more of these cases in the future. I have at least 4 cases where this greatly improves consistency in Spack and avoids common pitfalls (MPI, Python, Mesa(GL), and HDF5). There are likely many more of these cases, so it makes sense to just do it better.

So far the main hold up on the extensible solution is to add a feature that doesn't make sense to me and seems more likely to result in subtle bugs than provide value. I am yet to see a case where `run` or `test` deptypes would need to be added based on information knowable by the dependency. Relying on the dependent to specify that the build system requires hints for a package via the `build` dependency makes significantly more sense and is more expressive of actual use.

I keep asking for a motivating case for passing deptypes here where the added complexity is actually going to inform something useful. I fail to see a single instance where this would be needed and have provided at least one situation where this would cause a bug. I am happy to be proven wrong here, but I don't believe we should do something simply because information _can_ be extracted at a higher level. Rather, hiding some details to simplify higher level implementations and avoid pitfalls in usage seems like a more prudent approach.

repos/spack_repo/builtin/packages/python/package.py
kwryankrattiger: Yeah this makes sense to me. Adding in split out part of this that is just new feature and python conversion.

alalazo:
> I keep asking for a motivating case for passing deptypes here where the added complexity is actually going to inform something useful.

From https://github.com/spack/spack-packages/pull/2057#discussion_r2717599453

> Also, we don't account for test dependencies which may be used for build time tests, right? What if python is needed to run some test script etc ? 

From https://github.com/spack/spack-packages/pull/2057#discussion_r2721522631

> In practice this is not the case. There are enough instances of a run type dep where CMake still wants to know things about its executables. Sometimes people like to create absolute shebangs, etc. Specifying those are "build" deps would be wrong conceptually, as those are for the host, whereas the shebangs are for the target arch (would we ever do cross compiling)
>
> Test deps should definitely be included.

Aren't these motivating cases?

kwryankrattiger:
> Aren't these motivating cases?

I don't think so, both of those are covered by `link` imo. It doesn't make any sense to me to defer to a package that knows nothing about how a dependent will  use it to make any determination about whether or not it should provide flags. The only way to know would either be from the `disable_cmake_args_from` flag or the `deptype`. I cannot think of a single package that could make the determination differently than simply `build` and `link` with any degree of confidence.

I am honestly tired of arguing about it at this point though. For the cases this causes an actual bug we can just put it in the disable list. This doesn't fix the case where they should be passed when a variant is on and not when it is off (see my example above). Those cases can be addressed later.

johnwparent (approved):
Looks good, LGTM

PR #2058, opened 2025-10-22
Updates to Qt6 related packages #192

kwryankrattiger:
Extracted Qt6 updates from #192

### DIFF
--- a/repos/spack_repo/builtin/packages/qt_base/package.py
+++ b/repos/spack_repo/builtin/packages/qt_base/package.py
@@ -55,17 +55,15 @@ class QtPackage(CMakePackage):
                     if dep in vendor_deps_to_remove:
                         shutil.rmtree(dep)
 
+    @staticmethod
+    def _qt_feature_flag(feature):
+        return f"FEATURE_{feature}"
+
+    def define_qt_feature_from_variant(self, feature, variant=None):
+        return self.define_from_variant(QtPackage._qt_feature_flag(feature), variant or feature)
+
     def define_qt_feature(self, feature, value=None):
-        assert type(value) in (type(None), str, bool)
-
-        if value is None:
-            value = feature
-
-        flag = f"FEATURE_{feature}"
-        if isinstance(value, str):
-            return self.define_from_variant(flag, value)
-        else:
-            return self.define(flag, value)
+        return self.define(QtPackage._qt_feature_flag(feature), value)
 
     def cmake_args(self):
         # Start with upstream cmake_args
@@ -302,16 +300,16 @@ class QtBase(QtPackage):
             [
                 self.define_from_variant("BUILD_SHARED_LIBS", "shared"),
                 self.define_qt_feature("optimize_size", spec.satisfies("build_type=MinSizeRel")),
-                self.define_qt_feature("accessibility"),
+                self.define_qt_feature_from_variant("accessibility"),
                 # concurrent: default to on
-                self.define_qt_feature("dbus"),
-                self.define_qt_feature("framework"),
-                self.define_qt_feature("gui"),
-                self.define_qt_feature("network"),  # note: private feature
+                self.define_qt_feature_from_variant("dbus"),
+                self.define_qt_feature_from_variant("framework"),
+                self.define_qt_feature_from_variant("gui"),
+                self.define_qt_feature_from_variant("network"),  # note: private feature
                 # testlib: default to on
                 # thread: default to on
-                self.define_qt_feature("widgets"),  # note: private feature
-                self.define_qt_feature("sql"),  # note: private feature
+                self.define_qt_feature_from_variant("widgets"),  # note: private feature
+                self.define_qt_feature_from_variant("sql"),  # note: private feature
                 # xml: default to on
             ]
         )
--- a/repos/spack_repo/builtin/packages/qt_base/package.py
+++ b/repos/spack_repo/builtin/packages/qt_base/package.py
@@ -55,6 +55,18 @@ class QtPackage(CMakePackage):
                     if dep in vendor_deps_to_remove:
                         shutil.rmtree(dep)
 
+    def define_qt_feature(self, feature, value=None):
+        assert type(value) in (type(None), str, bool)
+
+        if value is None:
+            value = feature
+
+        flag = f"FEATURE_{feature}"
+        if isinstance(value, str):
+            return self.define_from_variant(flag, value)
+        else:
+            return self.define(flag, value)
+
     def cmake_args(self):
         # Start with upstream cmake_args
         args = super().cmake_args()
@@ -218,7 +230,8 @@ class QtBase(QtPackage):
         depends_on("libdrm")
         depends_on("at-spi2-core", when="+accessibility")
     depends_on("dbus", when="+dbus")
-    depends_on("gl", when="+opengl")
+    depends_on("gl", when="+opengl", type=("build", "link"))
+    depends_on("glu", when="+opengl", type=("build", "link"))
     depends_on("sqlite", when="+sql")
 
     with when("+gui"):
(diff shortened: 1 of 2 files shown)


repos/spack_repo/builtin/packages/qt_base/package.py
```diff
@@ -55,6 +55,18 @@ def remove_vendor_deps(self, vendor_dir, vendor_deps_to_remove):
...
+        assert type(value) in (type(None), str, bool)
+
+        if value is None:
+            value = feature
+
+        flag = f"FEATURE_{feature}"
+        if isinstance(value, str):
+            return self.define_from_variant(flag, value)
+        else:
+            return self.define(flag, value)
```
alalazo: Can we stick to Spack provided API, without inventing a new wrapper method for Qt? That should improve readability (since readers don't have to learn the Qt specific way of doing things).

repos/spack_repo/builtin/packages/qt_base/package.py
```diff
@@ -214,7 +226,12 @@ class QtBase(QtPackage):
         depends_on("libdrm")
         depends_on("at-spi2-core", when="+accessibility")
     depends_on("dbus", when="+dbus")
-    depends_on("gl", when="+opengl")
+    # OpenGL headers are required at build time
+    # GL, provided by the Qt application is required at runtime.
+    # Do not include GL in the "link" rpaths as GL may be loaded at runtime
+    # and differ from the GL that provided the headers.
+    depends_on("gl", when="+opengl", type=("build", "link"))
+    depends_on("glu", when="+opengl", type=("build", "link"))
```
alalazo: I don't understand this.

> Do not include GL in the "link" rpaths as GL may be loaded at runtime

but:
```python
depends_on("gl", when="+opengl", type=("build", "link"))
```
which is exactly the same as the previous:
```python
depends_on("gl", when="+opengl")
```
with the dependencies made explicit. It's also of `link` type, which seems to contradict the comment. 🤔

repos/spack_repo/builtin/packages/qt_base/package.py
```diff
@@ -283,32 +300,24 @@ def cmake_args(self):
         def define(cmake_var, value):
             args.append(self.define(cmake_var, value))
```
alalazo: Can we remove this too, in favor of just `self.define`?

repos/spack_repo/builtin/packages/qt_tools/package.py
```diff
@@ -39,14 +39,24 @@ class QtTools(QtPackage):
...
+    )
 
     # use of relative path in https://github.com/qt/qttools/blob/6.8.2/.gitmodules
     conflicts("+assistant", when="@6.8.2", msg="Incorrect git submodule prevents +assistant")
 
     depends_on("c")
     depends_on("cxx")
 
-    depends_on("llvm +clang")
+    depends_on("llvm +clang", when="+qdoc")
```
alalazo, line 63: Can you explain in which way we depend on `llvm+clang`, since we don't require it to be the `c` or `cxx` compiler?

repos/spack_repo/builtin/packages/qt_base/package.py
kwryankrattiger: It looks like some of the components do in fact link GL...These changes have been going on since before May, so it is possible the comment went stale.

repos/spack_repo/builtin/packages/qt_base/package.py
kwryankrattiger: Yeah...this can also go.

repos/spack_repo/builtin/packages/qt_base/package.py
kwryankrattiger: The point of keeping the feature one around was because it is ubiquitous pattern to prepend "FEATURE_" everywhere, this just makes it a little easier to implement everywhere.

I suppose I could remove it though. @wdconinc any objections?

repos/spack_repo/builtin/packages/qt_tools/package.py
kwryankrattiger: Qdoc uses clang to parse C++ files https://doc.qt.io/qt-6/qdoc-guide-clang.html

alalazo: I guess my question is: can we ever be in a situation where `qt-tools` uses `gcc` for `c` and `cxx`, while needing `clang` for `qdoc`?

repos/spack_repo/builtin/packages/qt_base/package.py
wdconinc: I think this is a bit of local interface infrastructure that is useful for Qt packages. They already inherit the base package to make things easier, so I think that makes clear to users where to look for the definition. Maybe we can rename it to `define_qt_feature` to make it abundantly clear that this is a local feature.

repos/spack_repo/builtin/packages/qt_tools/package.py
kwryankrattiger: yes, that is how we build in CI for #192. GCC for the build, Clang for qdoc only

repos/spack_repo/builtin/packages/qt_base/package.py
```diff
@@ -214,7 +226,12 @@ class QtBase(QtPackage):
         depends_on("libdrm")
         depends_on("at-spi2-core", when="+accessibility")
     depends_on("dbus", when="+dbus")
-    depends_on("gl", when="+opengl")
+    # OpenGL headers are required at build time
+    # GL, provided by the Qt application is required at runtime.
+    # Do not include GL in the "link" rpaths as GL may be loaded at runtime
+    # and differ from the GL that provided the headers.
+    depends_on("gl", when="+opengl", type=("build"))
+    depends_on("glu", when="+opengl", type=("build"))
```
alalazo:
```suggestion
    depends_on("gl", when="+opengl", type="build")
    depends_on("glu", when="+opengl", type="build")
```

alalazo (dismissed):
There are legitimate failures in pipelines:
- https://gitlab.spack.io/spack/spack-packages/-/jobs/18566210

kwryankrattiger:
> There are legitimate failures in pipelines:

This PR needs #2057 which fixes issues with FindOpenGL

wdconinc (dismissed):
Looks good to me with the changes suggested by @alalazo .

repos/spack_repo/builtin/packages/qt_base/package.py
```diff
@@ -214,7 +226,12 @@ class QtBase(QtPackage):
         depends_on("libdrm")
         depends_on("at-spi2-core", when="+accessibility")
     depends_on("dbus", when="+dbus")
-    depends_on("gl", when="+opengl")
+    # OpenGL headers are required at build time
+    # GL, provided by the Qt application is required at runtime.
+    # Do not include GL in the "link" rpaths as GL may be loaded at runtime
+    # and differ from the GL that provided the headers.
+    depends_on("gl", when="+opengl", type="build")
+    depends_on("glu", when="+opengl", type="build")
```
alalazo: Can you explain this? I thought the fix was to also `link`? Are we trying to enable `LD_LIBRARY_PATH` with Spack?

alalazo (commented):
Still one question

repos/spack_repo/builtin/packages/qt_base/package.py
kwryankrattiger: Hm, I think you are right, I don't remember why the link got dropped anymore.

repos/spack_repo/builtin/packages/qt_base/package.py
```diff
@@ -55,6 +55,18 @@ def remove_vendor_deps(self, vendor_dir, vendor_deps_to_remove):
                     if dep in vendor_deps_to_remove:
                         shutil.rmtree(dep)
 
+    def define_qt_feature(self, feature, value=None):
```
johnwparent: I think the concept behind this is a good one, but it does leave a layer of obfuscation as to what the actual method is doing, and makes reading the recipe slightly harder the first time. I definitely had to remind myself what the implicit behavior behind `value` was a few times. 

Since the actual business of this logic doesn't really have anything to do with define vs define from variant, why not keep that interface consistent and just do the value and flag name processing in this method?

i.e.

`self.define(self.qt_feature_name("gui"), "gui")`

johnwparent (dismissed):
Functionality wise (and somewhat relying on Sankesh's expertise here) looks great, just a nit that doesn't have to block (so I'm approving).

repos/spack_repo/builtin/packages/qt_base/package.py
kwryankrattiger: The point was to avoid a bunch of boilerplate and make it _more_ readable and avoid typos as where possible. You see in almost all cases we don't use the value argument at all.

I agree it is kind of strange how we treat value. I am handling three cases. The feature matches the variant, the feature does not match the variant, and the boolean value is explicit for a feature (ie. not a variant).

We could do the longer form as you suggest, but then we are just adding more boiler plate rather than reducing it.

I could also just not reuse `value` and do

```python
def define_qt_feature(self, feature, value: Optional[Any] = None, variant: Optional[str] = None):
    flag = f"FEATURE_{feature}"
    if value:
        return self.define(flag, value)

    if not variant:
        variant = feature
    return self.define_from_variant(flag, variant)
```

johnwparent: I think maybe split out the vanilla define and variant define behavior? That way we reduce the boilerplate but have a clearer interface with an obvious mapping to the existing behavior.

johnwparent (approved):

PR #2060, opened 2025-10-22
sysbench: add mysql variant

tukss:
This adds a variant to make the mysql dependency optional. It was always on before.


tukss:
@rbberger Can you have look at this?

rbberger (approved):

PR #2070, opened 2025-10-23
dyninst: add dependency for +openmp  %llvm-amdgpu variant 

vladgl:
This pr aims to fix this issue: https://github.com/dyninst/dyninst/issues/2035. Without `rocm-openmp-extras`, cmake cannot find the OpenMP package and produces a  `Could NOT find OpenMP_C (missing: OpenMP_C_FLAGS OpenMP_C_LIB_NAMES)` error. LAMMPS fixes this problem in a similar way: https://github.com/spack/spack-packages/blob/1562c2ce3066ec4249a7b8dae844cccb0f7094a9/repos/spack_repo/builtin/packages/lammps/package.py#L402


hainest:
I was just getting ready to submit this as a patch. You beat me to it by minutes!

vladgl:
@hainest It seems that merging develop branch was a bad idea because it led to a CI suspension. I don't know what to do with that.

hainest:
> @hainest It seems that merging develop branch was a bad idea because it led to a CI suspension. I don't know what to do with that.

I see 16 successful checks. Only ci/gitlab-ci is waiting. It's only run a few times a day, so it might be a while before it finishes.

hainest (approved):
So sorry, I thought I had approved this! Eventually, I think we need the llvm-amdgpu package to depend on rocm-openmp-extras, but this will work for now.

alalazo (approved):

PR #2072, opened 2025-10-23
py-torchgeo: add pandoc dependency

adamjstewart:
https://github.com/torchgeo/torchgeo/pull/3044

@blaz-r


blaz-r:
I'm not sure about this package since I had to install it from the website. I did also try the package from pip but it didn't work. I'll retry this, if it even matters.

adamjstewart:
Spack is relatively niche. If you want to install TorchGeo on a supercomputer, this is probably the best way. Unlike pip, it can list non-Python dependencies.

blaz-r:
Ah okay, so since this does not have a "py" prefix it just means that this is the pandoc app. I think that makes sense then 👍

alalazo (approved):

PR #2080, opened 2025-10-24
r: add gmake dep to build system if package needs compiling

manuelakuhn:
Packages that need compiling and hence have a dependency on `c` or `cxx` also need a dependency on `gmake`. Instead of adding that to all those packages manually, this PR adds it to the build system.


repos/spack_repo/builtin/build_systems/r.py
```diff
@@ -100,6 +100,10 @@ class RPackage(Package):
 
     extends("r")
 
+    # needed for packages that need compiling
+    depends_on("gmake", type="build", when="%c")
+    depends_on("gmake", type="build", when="%cxx")
```
michaelkuhn, line 105: This should probably also include:
```
depends_on("gmake", type="build", when="%fortran")
```

manuelakuhn: Done

michaelkuhn (approved):

PR #2083, opened 2025-10-24
py-numba: add rocm-openmp-extras dependency when compiler is amdclang

vladgl:
This pr fixes the same issue as in https://github.com/spack/spack-packages/pull/2070. py-numba includes `omp.h` and amdclang fails to find it without  `rocm-openmp-extras`.

### DIFF
--- a/repos/spack_repo/builtin/packages/py_numba/package.py
+++ b/repos/spack_repo/builtin/packages/py_numba/package.py
@@ -36,6 +36,8 @@ class PyNumba(PythonPackage):
     depends_on("c", type="build")  # generated
     depends_on("cxx", type="build")  # generated
 
+    depends_on("rocm-openmp-extras", when="%llvm-amdgpu", type="build")
+
     # Be careful that the bounds given in setup.py are exclusive on the upper bound
     # i.e., [min, max)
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/py_numba/package.py
```diff
@@ -36,6 +36,8 @@ class PyNumba(PythonPackage):
     depends_on("c", type="build")  # generated
     depends_on("cxx", type="build")  # generated
 
+    depends_on("rocm-openmp-extras", when="%llvm-amdgpu", type="build")
```
alalazo, line 39: If this needs to be added everywhere `llvm-amdgpu` is needed, we should fix it in the compiler package. Is there some issue already open about `llvm-amdgpu` + OpenMP, with `llvm-amdgpu` maintainers pinged?

vladgl: @alalazo 
> we should fix it in the compiler package

I think that would be the best solution. Perhaps it's also worth adding the `+openmp` variant, as it is done with `llvm`.
> Is there some issue already open

No, such an issue was not opened.

PR #2069, opened 2025-10-23
pkg-config: disable strict integer conversion check for amdclang

vladgl:
Add missing `-Wno-error=int-conversion` flag for pkg-config when compiled with amdclang 7, because without it compiler will fail to compile internal glib with the following error:
```
>> 670    gatomic.c:392:10: error: incompatible integer to pointer conversion passing 'gssize' (aka 'long') to parameter of type 'gpointer' (aka 'void *') 
            [-Wint-conversion]
     671      392 |   return g_atomic_pointer_add ((volatile gpointer *) atomic, val);
     672          |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
     673    ./gatomic.h:170:46: note: expanded from macro 'g_atomic_pointer_add'
     674      170 |     (gssize) __sync_fetch_and_add ((atomic), (val));                         \
     675          |                                              ^~~~~
  >> 676    gatomic.c:416:10: error: incompatible integer to pointer conversion passing 'gsize' (aka 'unsigned long') to parameter of type 'gpointer' (aka 'v
            oid *') [-Wint-conversion]
     677      416 |   return g_atomic_pointer_and ((volatile gpointer *) atomic, val);
     678          |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
     679    ./gatomic.h:177:45: note: expanded from macro 'g_atomic_pointer_and'
     680      177 |     (gsize) __sync_fetch_and_and ((atomic), (val));                          \
     681          |                                             ^~~~~
  >> 682    gatomic.c:440:10: error: incompatible integer to pointer conversion passing 'gsize' (aka 'unsigned long') to parameter of type 'gpointer' (aka 'v
            oid *') [-Wint-conversion]
     683      440 |   return g_atomic_pointer_or ((volatile gpointer *) atomic, val);
     684          |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
     685    ./gatomic.h:184:44: note: expanded from macro 'g_atomic_pointer_or'
     686      184 |     (gsize) __sync_fetch_and_or ((atomic), (val));                           \
     687          |                                            ^~~~~

[...]

PR #2019, opened 2025-10-20
openmpi: add missing header for +rocm build

afzpatel:
The following error is seen when building `+rocm` with `rocmcc`:
```
  >> 5300    accelerator_rocm_module.c:283:9: error: call to undeclared library
              function 'memcpy' with type 'void *(void *, const void *, unsigne
             d long)'; ISO C99 and later do not support implicit function decla
             rations [-Wimplicit-function-declaration]
     5301      283 |         memcpy(dest, src, size);
     5302          |         ^
```
adding this patch to resolve the issue: https://github.com/open-mpi/ompi/commit/15d97a054a286c7bf8e4296f6fed2ca6cd43f063

PR #2084, opened 2025-10-24
hpctoolkit: add rocm-openmp-extras dependency when compiler is amdclang

vladgl:
This is the same as in https://github.com/spack/spack-packages/pull/2070 and https://github.com/spack/spack-packages/pull/2083: compilation fails without `rocm-openmp-extras`.

### DIFF
--- a/repos/spack_repo/builtin/packages/hpctoolkit/package.py
+++ b/repos/spack_repo/builtin/packages/hpctoolkit/package.py
@@ -134,6 +134,8 @@ class Hpctoolkit(AutotoolsPackage, MesonPackage):
     depends_on("c", type="build")  # generated
     depends_on("cxx", type="build")  # generated
 
+    depends_on("rocm-openmp-extras", when="%llvm-amdgpu", type="build")
+
     with when("@2024.01: build_system=autotools"):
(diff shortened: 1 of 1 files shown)


mwkrentel:
@blue42u Does hpctoolkit build with `llvm-amdgpu` ?

vladgl:
@mwkrente  I successfully compiled it (`hpctoolkit~viewer~papi+rocm+mpi ^openmpi+rocm`) for our system using these prs: https://github.com/spack/spack-packages/pull/2070, https://github.com/spack/spack-packages/pull/2083, https://github.com/spack/spack-packages/pull/2069 and https://github.com/spack/spack-packages/pull/2019. Also, you need to compile rust with gcc for one of the py-* dependencies, because it gives an error when compiled with amdclang, maybe I'll live an issue for that later.

blue42u:
@mwkrentel I've never tried building `hpctoolkit` using `llvm-amdgpu`, but we do regularly build against LLVM/Clang and I would expect `amdclang` to be close enough for some practical purposes. So, no complaints here from me.

mwkrentel (approved):

tldahlgren (approved):

PR #2085, opened 2025-10-24
Update salmon-tddft package

freifrauvonbleifrei:
Updates to the existing SALMON2 package, to include all recent releases. Also, maintainers and actual license.

Starting this as a draft pull request, we're still validating for more setups.


repos/spack_repo/builtin/packages/salmon_tddft/package.py
```diff
@@ -17,10 +17,41 @@ class SalmonTddft(CMakePackage):
...
-    url = "https://salmon-tddft.jp/download/SALMON-v.2.0.0.tar.gz"
+    git = "https://github.com/SALMON-TDDFT/SALMON2.git"
 
-    license("BSD-3-Clause-Open-MPI")
+    maintainers("syamada0", "freifrauvonbleifrei")
+    license("Apache-2.0")
 
+    version(
+        "2.2.2",
+        url="http://salmon-tddft.jp/download/SALMON-v.2.2.2.tar.gz",
```
alalazo: Why did you remove the `url` to then specify it on each version?

alalazo (changes requested):

repos/spack_repo/builtin/packages/salmon_tddft/package.py
freifrauvonbleifrei: good question @alalazo ! My impression was that because the development happens on git and only releases are on this page, it might be better to make that explicit. But maybe it would be preferred to have both the git and url at the package level?

freifrauvonbleifrei: but how do I tell spack how to unpack the version into the url path?

alalazo: Spack interpolates urls automatically.

freifrauvonbleifrei: thanks! I changed it back :)

freifrauvonbleifrei:
@alalazo sorry for the late action. could you re-review please?

repos/spack_repo/builtin/packages/salmon_tddft/package.py
```diff
@@ -43,6 +54,7 @@ class SalmonTddft(CMakePackage):
     depends_on("fortran", type="build")  # generated
 
     depends_on("cmake@3.14:", type="build")
+    depends_on("cuda", when="+cuda")
```
alalazo: Why aren't we just subclassing `CudaPackage`, like many other packages do?

freifrauvonbleifrei: excellent idea, I adapted it.

kshea21:
@alalazo could you re-review this with the new changes?

repos/spack_repo/builtin/packages/salmon_tddft/package.py
```diff
@@ -17,15 +18,25 @@ class SalmonTddft(CMakePackage):
     with norm-conserving pseudopotentials."""
 
     homepage = "https://salmon-tddft.jp"
-    url = "https://salmon-tddft.jp/download/SALMON-v.2.0.0.tar.gz"
+    git = "https://github.com/SALMON-TDDFT/SALMON2.git"
+    url = "https://salmon-tddft.jp/download/SALMON-v.2.2.2.tar.gz"
 
-    license("BSD-3-Clause-Open-MPI")
+    maintainers("syamada0", "freifrauvonbleifrei")
```
alalazo, line 24: @syamada0 Do you agree being a maintainer of the package?

repos/spack_repo/builtin/packages/salmon_tddft/package.py
```diff
@@ -92,17 +106,28 @@ def cmake_args(self):
...
+        if "%nvhpc" in spec:
+            if "+openacc" in spec:
+                if "+cuda" in spec:
+                    args.append(define("CMAKE_TOOLCHAIN_FILE", "nvhpc-openacc-cuda"))
+                else:
+                    args.append(define("CMAKE_TOOLCHAIN_FILE", "nvhpc-openacc"))
+            else:
+                args.append(define("CMAKE_TOOLCHAIN_FILE", "nvhpc-openmp"))
+        elif "%fj" in spec:  # TODO add toolchain files if necessary
+            args.append(self.define("CMAKE_Fortran_MODDIR_FLAG", "-M"))
```
alalazo, line 118: Can we make this more linear? E.g.
```python
if spec.satisfies("~openacc %nvhpc"):
    args.append(define("CMAKE_TOOLCHAIN_FILE", "nvhpc-openmp"))
elif spec.satisfies("+openacc~cuda %nvhpc"):
    ...
```

alalazo (commented):
Just waiting for a confirmation that the second maintainers agrees being pinged

repos/spack_repo/builtin/packages/salmon_tddft/package.py
```diff
@@ -55,8 +66,11 @@ class SalmonTddft(CMakePackage):
     conflicts("+eigenexa", when="~scalapack")
     conflicts("+manycore", when="@2.0.0:")
     conflicts("+current_processing", when="@2.0.0:")
+    conflicts("+cuda", when="~openacc")
 
-    patch("fjmpi.patch", when="@2.0.0: %fj")
+    requires("%nvhpc", when="+cuda")
+
+    patch("fjmpi.patch", when="@2.0.0 %fj")
```
alalazo, line 73: Also, I assume this ☝️ was done on purpose, but just want to confirm.

repos/spack_repo/builtin/packages/salmon_tddft/package.py
```diff
@@ -17,15 +18,25 @@ class SalmonTddft(CMakePackage):
...
+    version("2.2.0", sha256="b0b6eabad48f4547ae2a97f53771047ccb759b8c126ab0f1653b0e0f0d02a28b")
+    version("2.1.0", sha256="18267818cdfa82ea762441e2d751abeff7b553c8ce92cabf5fb010248d2cfcbe")
+    version("2.0.2", sha256="742007d3684a478199ba959ce135ad0020b70676a49f52a5e1dc25438123d50e")
+    version("2.0.1", sha256="6fcd72ddd484a1d2b4700bbe5c1717fabb1e449288da3253d74b4a8ab24e7255")
     version("2.0.0", sha256="c3bb80bc5d338cba21cd8f345acbf2f2d81ef75af069a0a0ddbdc0acf358456c")
     version("1.2.1", sha256="a5045149e49abe9dd9edefe00cd1508a1323081bc3d034632176b728effdbaeb")
 
     variant("mpi", default=False, description="Enable MPI")
     variant("libxc", default=False, description="Enable libxc")
+    variant("openacc", default=False, description="Enable OpenACC")
```
alalazo, line 39: The variant is unconditional. This means it applies to all versions, including v1.2.1 and v2.0.0 Is that correct?

repos/spack_repo/builtin/packages/salmon_tddft/package.py
```diff
@@ -55,8 +66,11 @@ class SalmonTddft(CMakePackage):
     conflicts("+eigenexa", when="~scalapack")
     conflicts("+manycore", when="@2.0.0:")
     conflicts("+current_processing", when="@2.0.0:")
+    conflicts("+cuda", when="~openacc")
 
-    patch("fjmpi.patch", when="@2.0.0: %fj")
+    requires("%nvhpc", when="+cuda")
```
alalazo, line 71: Is this correct or should we force the compiler when we use `+openacc` ?

alalazo (commented):
A few more comments

repos/spack_repo/builtin/packages/salmon_tddft/package.py
freifrauvonbleifrei: yes, `nvhpc` is expected for the cuda variant.

repos/spack_repo/builtin/packages/salmon_tddft/package.py
syamada0: Thank you very much! I agree!

repos/spack_repo/builtin/packages/salmon_tddft/package.py
freifrauvonbleifrei: Yes, I think it should apply to old versions too. @syamada0 please correct if I'm wrong!

repos/spack_repo/builtin/packages/salmon_tddft/package.py
freifrauvonbleifrei: just openacc with non-cuda backend should work, so this looks good to me. again please correct if wrong @syamada0 !

repos/spack_repo/builtin/packages/salmon_tddft/package.py
freifrauvonbleifrei: good suggestion -> addressed in commit 4391fe8b55e1e6ac6ff9b99a3df62c2a3df5e24f 

thanks!

PR #2088, opened 2025-10-24
Dorado: Updated versions

emwjacobson:
Adds the latest version of each minor release. Had to adjust how some patching worked, as different versions of dorado made slight changes that prevented the existing patch files from "just working".

@snehring the maintainer, so obligatory mention (Hello again! :smile:)


repos/spack_repo/builtin/packages/dorado/package.py
```diff
@@ -39,13 +48,42 @@ class Dorado(CMakePackage, CudaPackage):
...
+                "add_dependencies(dorado_lib htslib_project)", "", "CMakeLists.txt", string=True
+            )
+            filter_file(
+                "add_dependencies(dorado_utils htslib_project)",
+                "",
+                "dorado/utils/CMakeLists.txt",
+                string=True,
+            )
+
+            with when("@0.9.5:0.9.6"):
```
snehring: I'm not sure this syntax works here, at the very least this filter_file function is being called with 0.6.3 resulting in an error.
```suggestion
           if self.spec.satisfies("@0.9.5:0.9.6"):
```
probably replace all the 'with when's in this function.

snehring:
It will likely be Monday before I can run through this fully, I need to resolve some issues in my environment around cuda.

emwjacobson:
Good catch, sorry about that! Latest push should address that. I was able to install all 9 versions, with and without the CUDA flag. Let me know if you have any troubles getting them installed!

snehring:
I was getting a linking error during building for 1.2.0 with cuda@12.6 (it looked like it was looking for cuda 11.8 libs)  however I've been unable to reproduce the issue since trying a build with cuda@11.8 (successful), uninstalling all dorado versions, and building again. Not sure what was going on there, but I think we're good.

snehring (approved):

snehring:
@alalazo This is good to go

alalazo (approved):
Approved by maintainer

PR #2093, opened 2025-10-24
mmseqs2: Updated versions

emwjacobson:
Versions 16, 17, and 18 were added.

GPU support was also added in version 16, so steps allowing it's compilation was added as well.

### DIFF
--- a/repos/spack_repo/builtin/packages/mmseqs2/package.py
+++ b/repos/spack_repo/builtin/packages/mmseqs2/package.py
@@ -35,6 +35,7 @@ class Mmseqs2(CMakePackage, CudaPackage):
     depends_on("mpi", when="+mpi")
     depends_on("cuda", when="@16: +cuda")
 
+    conflicts("cuda", when="@:15 +cuda")
     conflicts("cuda_arch=none", when="+cuda", msg="CUDA architecture is required")
     conflicts(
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/mmseqs2/package.py
```diff
@@ -29,6 +33,12 @@ class Mmseqs2(CMakePackage):
 
     depends_on("zstd")
     depends_on("mpi", when="+mpi")
+    depends_on("cuda", when="@16: +cuda")
```
becker33, line 36: Should this be coupled with a conflict for `@:15 +cuda`?

emwjacobson: Makes sense. Just pushed a commit to address this.

repos/spack_repo/builtin/packages/mmseqs2/package.py
```diff
@@ -29,6 +33,13 @@ class Mmseqs2(CMakePackage):
...
+
+    conflicts("cuda", when="@:15 +cuda")
```
becker33:
```suggestion
    conflicts("@:15 +cuda")
```

becker33 (changes requested):

becker33 (approved):

PR #2094, opened 2025-10-24
pygtrie: add version 2.5.0

kwryankrattiger:


manuelakuhn (approved):

PR #2095, opened 2025-10-24
py-parse: add versions 1.19.1 and 1.20.2

kwryankrattiger:


manuelakuhn (commented):
needs `py-setuptools@61.2:` starting from `1.19.1`
(see [`pyproject.toml`](https://github.com/r1chardj0n3s/parse/blob/1.20.2/pyproject.toml#L2))

manuelakuhn (approved):

PR #2097, opened 2025-10-24
cray-mpich: add `+gtl` variant and change it to default when `+rocm` or `+cuda`

nhanford:
The common case when building against a GPU-Aware `cray-mpich` is to set up the GPU Transport Library for accelerated communication between GPUs. However, the current method of doing so by manually updating all dependent packages to include the GTL is clunky. I have changed the cray-mpich package to setup dependents to build and run with the GTL by default. This has been tested and works on a system very similar to El Capitan at LLNL with the `osu-micro-benchmarks` as the tested package.

@scheibelp This is my implementation of what we had discussed earlier. Please let me know what you think.

TODO:

- [X] Need to update `when=` clause to allow `+cuda` as well.

Thanks,
Nate


nhanford:
@spackbot fix style

repos/spack_repo/builtin/packages/cce/package.py
```diff
@@ -58,7 +58,27 @@ class Cce(Package, CompilerPackage):
...
+    version("15.0.1")
     version("16.0.0")
+    version("16.0.1")
+    version("17.0.0")
+    version("17.0.1")
+    version("18.0.0")
+    version("18.0.1")
+    version("19.0.0")
+    version("19.0.1")
+    version("20.0.0")
```
scheibelp, line 61: Generally speaking Spack understands version constraints even without these versions being defined. My inclination is to suggest keeping this version list "sparse" (likewise for `cray-mpich`).

repos/spack_repo/builtin/packages/cray_mpich/package.py
```diff
@@ -114,6 +124,16 @@ def setup_dependent_package(self, module, dependent_spec):
                 spec.mpifc = dependent_spec["fortran"].package.fortran
                 spec.mpif77 = dependent_spec["fortran"].package.fortran
 
+    def setup_dependent_build_environment(self, env, dependent_spec):
+        if self.spec.satisfies("+gtl"):
+            for flag in self.gtl_lib["ldflags"]:
+                env.prepend_path("LDFLAGS", flag)
+            env.prepend_path("LIBS", self.gtl_lib["ldlibs"][0])
```
scheibelp: From looking into this I get the impression this will generally inform Autotools-based builds on how to use GTL libs, but I think it will not address GTL functionality in all cases. When `%cce` is used, it appears that `CRAY_ACCEL_TARGET` will tell the wrappers directly to insert `-lmpi_gtl_hsa` (so may work across build systems).

The changes here may be preferable in the sense that it may get things working when using `amdclang` etc. rather than `%cce`, although I'd have to think on that.

scheibelp (commented):
(I have to think on whether this is the way we want this generally handled - more thoughts on that below)

nhanford:
Hello @becker33 @tgamblin,

@scheibelp brings up a lot of good points. Greg and I were just talking about either adding a `cray-mpich-wrapper` dependency that does this out of concern that:

1. *Always* linking `-lmpi_gtl_hsa`, even for non-MPI code is not a good idea.
2. As Peter mentions above, not all build methods may be responsive to these flags, but will be responsive to setting it in a wrapper.

Or we could, in theory, hack existing Spack compiler wrappers similar to `cce` such that they could detect `+gtl` and `-lmpi` and automatically add `-lmpi_gtl_hsa` in such cases.

Thanks,
Nate

etiennemlb:
I agree with @scheibelp, while this may be working with autotools, it won't work generally.

What I dislike the most with the current GTL setup is that it puts the burden of linking the GTL on the package depending on cray-mpich (see https://github.com/spack/spack-packages/pull/1164 and https://github.com/spack/spack-packages/blob/develop/repos/spack_repo/builtin/packages/lammps/package.py#L534).

The issue I had was that I do not know how to inject an ldlibs in the cray-mpich dependent packages. 
I'd like to see the cray-mpich-wrapper POC.

cheers

repos/spack_repo/builtin/packages/cray_mpich/package.py
rbberger: Just to add another perspective. `%cce` might not automatically mean that `cc`, `CC`, `ftn` is used, but might be defined to use the Cray compilers directly. At least in my Spack deployments and CMake-based projects, I've been convinced (initially by @becker33) that avoiding the Cray compiler wrappers with Spack saves us more headaches.

repos/spack_repo/builtin/packages/cce/package.py
nhanford: For someone "naively" doing `spack external find cray-mpich`, wouldn't they need the versions to match or the command would fail?

rbberger: Typically you have the latest versions on the top.

tpadioleau:
Hi, here are my 2 cents. Unfortunately the PR did not work for me. I needed to compile a stack with `gcc@13` as a host compiler. To make the gtl work I needed a bit more modifications, see https://github.com/gyselax/spack/blob/508ca8087defc5546964b138b968dc703d5a47a5/spack_repo/patches_v2025_11/packages/cray_mpich/package.py#L119-L121. In the current state, the modifications are too specific to my use case so that I can propose them. Also I had trouble with `append("LDFLAGS", flag)` called multiple times. The compiler was not happy with the resulting flags.

nhanford:
@tpadioleau and others,
Thanks for trying things out and providing feedback. @scheibelp and I are currently working on a solution that would involve a new `cray-mpich-gtl` package that depends on `cray-mpich` and adds `-lmpi_gtl_hsa` or `-lmpi_gtl_cuda` to the `mpicc` invocation and sets up the dependent run environment accordingly and will soon either modify this PR or close it and open a new PR that hopefully addresses this issue.

etiennemlb:
Don't we have other libraries in spack that need to inject dependencies in a parent ?

The fundamental specificity of craympich & gtl, is that it's two libraries.

You could trivially make a copy of the craympich lib, patch elf the dt needed to add the appropriate gtl, store the resulting modified binary as part of spack installed products and call it done. Every dependent software will link against this modified mpich binary.

Writting such a modified cray mpich spack package is easy enough, though I'm not sure it's fine by cray's TOS.

nhanford:
@spackbot fix style

PR #2098, opened 2025-10-25
py-pydantic-core: add v2.41.4

adamjstewart:
closes #1528 (adds py-maturin v1.9.6)

Add Python 3.14 support, fix missing run-time dep on typing-extensions


adamjstewart:
@alalazo this is the same bug reported in other PRs where multiple versions of setuptools are being added to PYTHONPATH and the wrong version is first.

alalazo (approved):

PR #2099, opened 2025-10-25
py-parso: add v0.8.5

adamjstewart:
Add Python 3.14 support.

Not entirely sure if the upper bound is necessary, this is the version of Python to check for, not the run-time version of Python supported.


manuelakuhn (approved):

PR #2102, opened 2025-10-25
py-charset-normalizer: add v3.4.4

adamjstewart:
Python 3.14 support

### DIFF
--- a/repos/spack_repo/builtin/packages/py_charset_normalizer/package.py
+++ b/repos/spack_repo/builtin/packages/py_charset_normalizer/package.py
@@ -33,6 +33,7 @@ class PyCharsetNormalizer(PythonPackage):
         depends_on("py-mypy@1.4.1:1.15.0", when="@3.4.2")
         depends_on("py-mypy@1.4.1:1.14.0", when="@3.4.1")
 
+    # Based on PyPI wheel availability
     with default_args(type=("build", "run")):
         depends_on("python@3.8:3.14", when="@3.4.3:")
         depends_on("python@3.7:3.13", when="@3.4.0:3.4.2")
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/py_charset_normalizer/package.py
```diff
@@ -24,13 +25,17 @@ class PyCharsetNormalizer(PythonPackage):
...
-        # depends_on("py-mypy@1.4.1:1.15.0", when="@3.4.2:")
+
+        # Historical dependencies
+        depends_on("py-setuptools-scm", when="@3.4.1:3.4.3")
+        depends_on("py-mypy@1.4.1:1.15.0", when="@3.4.2")
+        depends_on("py-mypy@1.4.1:1.14.0", when="@3.4.1")
 
     with default_args(type=("build", "run")):
-        depends_on("python@3.7:3.13", when="@3.4:")
+        depends_on("python@3.8:3.14", when="@3.4.3:")
```
manuelakuhn, line 38:
```suggestion
        depends_on("python@3.7:3.14", when="@3.4.3:")
```
according to [`pyproject.toml#L38`](https://github.com/jawah/charset_normalizer/blob/3.4.4/pyproject.toml#L38)

adamjstewart: 3.8 according to https://pypi.org/project/charset-normalizer/#files. At the very least, 3.7 is no longer being tested or supported. Neither version is supported in Spack. I could also remove the lower bounds if you want to make things simpler.

manuelakuhn: Thanks for the clarification. Can you then add a comment that the version restriction comes from the wheels listed in PyPI to prevent confusion in the future?

manuelakuhn (approved):
Thanks for the change. LGTM and builds successfully on rocky 8 with gcc@15

PR #2107, opened 2025-10-26
madgraph5amc: Inject JetMatching header into existing Pythia8 install

paulgessinger:
In order to do FxFx matching, madgraph wants to see it's own `JetMatching.h` byte-for-byte in Pythia8's include directory. The only way to achieve this, as far as I can tell, is to overwrite the header from the madgraph build in-place into pythia8's install directory. **This feels super icky to me**, but I think this is a consequence of the choices these HEP packages have made.

The only alternative I see is to detect whether madgraph is concretized from inside of the pythia8 package, and then modify its own source code accordingly, but that's not really less icky.

Any thoughts?


repos/spack_repo/builtin/packages/madgraph5amc/package.py
```diff
@@ -129,6 +130,17 @@ def installfile(filename):
 
         # TODO: Fix for reproducibility, see https://github.com/spack/spack/pull/41128#issuecomment-2305777485
         if "+pythia8" in spec:
+            # Overwrite Pythia8's version of `JetMatching.h` with the one from MadGraph5_aMC@NLO
+            src = join_path(self.prefix, "Template", "NLO", "MCatNLO", "Scripts", "JetMatching.h")
+            dst = join_path(
+                self.spec["pythia8"].prefix, "include", "Pythia8Plugins", "JetMatching.h"
```
wdconinc, line 133: I guess madgraph doesn't really honor any CXXFLAGS=-I<path> injections, does it? We do avoid writing in unrelated prefixes, yeah.

paulgessinger: I don't think so, they do this at **runtime** not at compile time. They compute all the flags themselves.

What they do is:

- Lookup Pythia8's install directory that was configured at build time
- They locate `$PYTHIA8_INSTALL/include/Pythia8Plugins/JetMatching.h`
- They read that file and compare it to the one they ship in-source byte-for-byte:

```python
# check that if FxFx is activated the correct shower plugin is present
if shower == 'PYTHIA8' and self.run_card['ickkw'] == 3:
    misc.sprint(self.options['pythia8_path'])
    f1 = open(pjoin(self.me_dir, "MCatNLO","Scripts", "JetMatching.h"), "r").read()
    py8_plugin = pjoin(self.options['pythia8_path'], "include","Pythia8Plugins","JetMatching.h")
    if not os.path.exists(py8_plugin):
        raise Exception("FxFx requires a dedicated plugin to be installed in Pythia8. See http://cern.ch/amcatnlo/FxFx_merging.htm")
    f2 = open(py8_plugin).read()
    if f1 != f2:
        raise Exception("FxFx requires a dedicated plugin to be installed in Pythia8. Incorrect plugin detected. See http://cern.ch/amcatnlo/FxFx_merging.htm")
```

I agree this is **terrible** but they've chosen this. The alternative is to let MadGraph build it's own Pythia8, but that seems even less in line with spack's philosophy.

paulgessinger: A possible alternative would be to patch this check out of MadGraph, but afaiu then you get wrong physics, which is arguably worse.

paulgessinger:
@wdconinc any further thoughts?

paulgessinger:
@wdconinc can we proceed witht this?

paulgessinger:
@wdconinc any further thoughts?

repos/spack_repo/builtin/packages/madgraph5amc/package.py
wdconinc: It is just a quite hard no to replace a user-facing header like this when it is supposed to reflect the compiled libraries, in particular in a way that applies forward to future pythia8 releases. There are a sufficiently large number of dependents of pythia8 that I don't think we should overwrite the header in the pythia8 prefix without imposing the use of this header for the compilation of pythia8 as well. It opens up all kinds of cans of worms, including things like a dependency on whether another dependent of pythia8 is installed before or after madgraph5amc  So, I think we should apply this as a patch to pythia8 before its compilation.

This would then call for the use of a patch imposed by the dependent, i.e. `madgraph5amc` would include a dependency like follows:
```py
    depends_on(
        "pythia8@8.301:",
        patches=[
            patch("madgraph5amc-3.3-pythia8-8.301-JetMatching.patch", when="@8.301:8.307"),
            # patches verified from 8.301 to 8.307; 8.308 includes the functionality natively
        ],
        when="+pythia8 @3.3:"
    )
```
Any spec that includes both pythia8 and madgraph5amc would then apply the patch to pythia8. Other dependents would get a consistent header/library for consistent interface/implementation.

Based on my reading, https://pythia.org/latest-manual/UpdateHistory.html#section8 and https://pythia.org/latest-manual/JetMatching.html#section3, the changes that are copied by the patch were merged into pythia8 as of 8.308. Copilot seems to confirm that when I ask it to do a three-way merge::

>    The changes from the original patch (JetMatching.patch for pythia8307) have
>    already been i ntegrated into pythia8308 by the Pythia8 developers in a cleaner,
>    refactored form. The pythia8308 version already includes all six steps of
>    modifications:
> 
>      - Step-FxFx-1: Removal of FxFx preclustering
>      - Step-FxFx-2: Unified eventProcess for both MLM and FxFx  
>      - Step-FxFx-3: Modified nRequested calculation with typeIdx[2] handling
>      - Step-FxFx-4: Veto logic for real emissions with only typeIdx=2 partons
>      - Step-FxFx-5 & 6: npNLO() condition changes in matching checks
>      - Updated scale comparison for FxFx using trunc(1000.*scale)
> 
>    These changes are properly documented in the pythia8308 code with comments
>    referencing "arXiv:2108.07826". No additional patching is required for
>    pythia8308 as it already contains the desired functionality.

This also seems to indicate that the copying over of the JetMatching header is only needed when one uses a pythia8 version before 8.308.

So, let me suggest the following strategy instead of copying into the pythia8 prefix:
```py
    # MadGraph5_aMC@NLO installs its own version of include/Pythia8Plugins/JetMatching.h
    # in the pythia8 installation, see Template/NLO/MCatNLO/Scripts/JetMatching.h.
    # In spack, we impose this as a patch to pythia8 when used as a dependency.
    # The version in MadGraph5_aMC@NLO has not meaningfully changed since v3.3.x;
    # this patch approach intends to bridge changes to JetMatching.h until they were
    # integrated in pythia8.308.
    depends_on(
        "pythia8@8.301:",
        patches=[
            patch("madgraph5amc-3.3-pythia8-8.301-JetMatching.patch", when="@8.301:8.307"),
            # patches verified from 8.301 to 8.307; 8.308 includes the functionality natively
        ],
        when="+pythia8 @3.3:"
    )
```
with the attached patch uploaded to the `madgraph5amc` directory in spack:
[madgraph5amc-3.3-pythia8-8.301-JetMatching.patch](https://github.com/user-attachments/files/23692426/madgraph5amc-3.3-pythia8-8.301-JetMatching.patch)

Finally, sorry it took me a while to get back to this, but I hope this seems like a more sensible and long-term approach.

wdconinc: > A possible alternative would be to patch this check out of MadGraph, but afaiu then you get wrong physics, which is arguably worse.

And once we properly patch in the right physics this way, then I think we can safely patch out the runtime check from madgraph5amc.

wdconinc: Some of this in https://github.com/spack/spack-packages/tree/madgraph5amc-pythia8-patch for cherry-picking (I was working on this in codespaces, so might as well provide it). Doesn't include the patching out of the runtime check.

paulgessinger: Hi @wdconinc, thanks for taking a look. I've verified that the `JetMatching.h` file shipped by Pythia8 is **not identical** to the one madgraph expects. The diff is here:

```diff
--- mg5_JetMatching.h	2025-11-23 09:26:27.948930229 +0100
+++ py8_JetMatching.h	2025-11-23 09:26:47.904903379 +0100
@@ -1,5 +1,5 @@
 // JetMatching.h is a part of the PYTHIA event generator.
-// Copyright (C) 2021 Torbjorn Sjostrand.
+// Copyright (C) 2025 Torbjorn Sjostrand.
 // PYTHIA is licenced under the GNU GPL v2 or later, see COPYING for details.
 // Please respect the MCnet Guidelines, see GUIDELINES for details.
 
@@ -87,8 +87,8 @@
 public:
 
   // Constructor and destructor
- JetMatching() : cellJet(NULL), slowJet(NULL), slowJetHard(NULL),
-    hjSlowJet(NULL) {}
+ JetMatching() : cellJet(nullptr), slowJet(nullptr), slowJetHard(nullptr),
+    hjSlowJet(nullptr) {}
  ~JetMatching() {
     if (cellJet) delete cellJet;
     if (slowJet) delete slowJet;
@@ -276,7 +276,7 @@
 public:
 
   // Constructor and destructor
-  JetMatchingMadgraph() : slowJetDJR(NULL) { }
+  JetMatchingMadgraph() : slowJetDJR(nullptr) { }
   ~JetMatchingMadgraph() { if (slowJetDJR) delete slowJetDJR; }
 
   // Initialisation
@@ -1139,21 +1139,21 @@
   jetAlgorithm = 2;
   slowJetPower = 1;
   slowJet = new SlowJet(slowJetPower, coneRadius, eTjetMin,
-    etaJetMaxAlgo, 2, 2, NULL, false);
+    etaJetMaxAlgo, 2, 2, nullptr, false);
 
   // For FxFx, also initialise jet algorithm to define matrix element jets.
   // Currently, this only supports the kT-algorithm in SlowJet.
   // Use the QCD distance measure by default.
   slowJetHard = new SlowJet(slowJetPower, coneRadius, qCutME,
-    etaJetMaxAlgo, 2, 2, NULL, false);
+    etaJetMaxAlgo, 2, 2, nullptr, false);
 
   // To access the DJR's
   slowJetDJR = new SlowJet(slowJetPower, coneRadius, qCutME,
-    etaJetMaxAlgo, 2, 2, NULL, false);
+    etaJetMaxAlgo, 2, 2, nullptr, false);
 
   // A special version of SlowJet to handle heavy and other partons
   hjSlowJet = new HJSlowJet(slowJetPower, coneRadius, 0.0,
-    100.0, 1, 2, NULL, false, true);
+    100.0, 1, 2, nullptr, false, true);
 
   // Setup local event records
   eventProcessOrig.init("(eventProcessOrig)", particleDataPtr);
@@ -1372,83 +1372,15 @@
   clearDJR();
   clear_nMEpartons();
 
-  // Step-FxFx-1: remove preclustering from FxFx
+  // Note: Compared to older versions (cf. arXiv:2108.07826):
+  // No more preclustering for FxFx.
 
- /* 
+  // Note: Compared to older versions (cf. arXiv:2108.07826):
+  // Set the same eventProcess for both MLM and FxFx. This was done separately
+  // for FxFx before. Add the type 2 selection also for FxFx
 
-  // For FxFx, pre-cluster partons in the event into jets.
-  if (doFxFx) {
-
-    // Get final state partons
-    eventProcess.clear();
-    workEventJet.clear();
-    for( int i=0; i < workEvent.size(); ++i) {
-      // Original AG+Py6 algorithm explicitly excludes tops,
-      // leptons and photons.
-      int id = workEvent[i].idAbs();
-      if ((id >= ID_LEPMIN && id <= ID_LEPMAX) || id == ID_TOP
-        || id == ID_PHOTON || id == 23 || id == 24 || id == 25) {
-        eventProcess.append(workEvent[i]);
-      } else {
-        workEventJet.append(workEvent[i]);
-      }
-    }
-
-    // Initialize SlowJetHard jet algorithm with current working event
-    if (!slowJetHard->setup(workEventJet) ) {
-      errorMsg("Warning in JetMatchingMadgraph:sortIncomingProcess"
-        ": the SlowJet algorithm failed on setup");
-      return;
-    }
-
-    // Get matrix element cut scale.
-    double localQcutSq = qCutMESq;
-    // Cluster in steps to find all hadronic jets at the scale qCutME
-    while ( slowJetHard->sizeAll() - slowJetHard->sizeJet() > 0 ) {
-      // Done if next step is above qCut
-      if( slowJetHard->dNext() > localQcutSq ) break;
-      // Done if we're at or below the number of partons in the Born state.
-      if( slowJetHard->sizeAll()-slowJetHard->sizeJet() <= npNLO()) break;
-      slowJetHard->doStep();
-    }
-
-    // Construct a master copy of the event containing only the
-    // hardest nPartonsNow hadronic clusters. While constructing the event,
-    // the parton type (ID_GLUON) and status (98,99) are arbitrary.
-    int nJets = slowJetHard->sizeJet();
-    int nClus = slowJetHard->sizeAll();
-    int nNow = 0;
-    for (int i = nJets; i < nClus; ++i) {
-      vector<int> parts;
-      if (i < nClus-nJets) parts = slowJetHard->clusConstituents(i);
-      else parts = slowJetHard->constituents(nClus-nJets-i);
-      int flavour = ID_GLUON;
-      for(int j=0; j < int(parts.size()); ++j)
-        if (workEventJet[parts[j]].id() == ID_BOT)
-          flavour = ID_BOT;
-      eventProcess.append( flavour, 98,
-        workEventJet[parts.back()].mother1(),
-        workEventJet[parts.back()].mother2(),
-        workEventJet[parts.back()].daughter1(),
-        workEventJet[parts.back()].daughter2(),
-        0, 0, slowJetHard->p(i).px(), slowJetHard->p(i).py(),
-        slowJetHard->p(i).pz(), slowJetHard->p(i).e() );
-      nNow++;
-    }
-
-    // Done. Clean-up
-    workEventJet.clear();
-    
-   */
-
-  // Step-FxFx-2: : set the same eventProcess for both MLM and FxFx. This was only for MLM
-  // Add the type 2 selection also for FxFx
-
-  // For MLM matching, simply take hard process state from workEvent,
-  // without any preclustering.
-  //} else {
-    eventProcess = workEvent;
-  //}
+  // For jet matching, simply take hard process state from workEvent.
+  eventProcess = workEvent;
 
   // Sort original process final state into light/heavy jets and 'other'.
   // Criteria:
@@ -1475,16 +1407,17 @@
       || (eventProcess[i].idAbs() <= nQmatch) ) {
       orig_idx = 0;
       if (doFxFx) {
-          // Crucial point FxFx: MG5 puts the scale of a not-to-be-matched quark 1 MeV lower than scalup. For
-          // such particles, we should keep the default "2"
-          idx = ( trunc(1000. * eventProcess[i].scale()) == trunc(1000. * infoPtr->scalup()) ) ? 0 : 2;
+        // Crucial point FxFx: MG5 puts the scale of a not-to-be-matched
+        // quark 1 MeV lower than scalup. For such particles, we should
+        // keep the default "2"
+        idx = ( trunc(1000.*eventProcess[i].scale())
+             == trunc(1000.*infoPtr->scalup()) ) ? 0 : 2;
 
-      }
-      else {
-      // Crucial point: MG puts the scale of a non-QCD particle to eCM. For
-      // such particles, we should keep the default "2"
-      idx = ( eventProcess[i].scale() < 1.999 * sqrt(infoPtr->eA()
-        * infoPtr->eB()) ) ? 0 : 2;
+      } else {
+        // Crucial point: MG puts the scale of a non-QCD particle to eCM. For
+        // such particles, we should keep the default "2"
+        idx = ( eventProcess[i].scale() < 1.999 * sqrt(infoPtr->eA()
+          * infoPtr->eB()) ) ? 0 : 2;
       }
     }
 
@@ -1688,36 +1621,25 @@
   // Count of the number of hadronic jets in SlowJet accounting
   int nCLjets = nClus - nJets;
   // Get number of partons. Different for MLM and FxFx schemes.
-  //int nRequested = (doFxFx) ? npNLO() : nParton;
-  //Step-FxFx-3: Change nRequested subtracting the typeIdx[2] partons
-  //Exclude the highest multiplicity sample in the case of real emissions and all typeIdx[2]
-  //npNLO=multiplicity,nJetMax=njmax(shower_card),typeIdx[2]="Weak" jets
-  int nRequested = (doFxFx && !(npNLO()==nJetMax && npNLO()==(typeIdx[2].size()-1) )) ? npNLO()-typeIdx[2].size() : nParton;
-
-  //Step-FxFx-4:For FxFx veto the real emissions that have only typeIdx=2 partons
-  //From Step-FxFx-3 they already have negative nRequested, so this step may not be necessary
-  //Exclude the highest multiplicity sample
-  if (doFxFx && npNLO()<nJetMax && typeIdx[2].size()>0 && npNLO()==(typeIdx[2].size()-1)) {
-        return MORE_JETS;
-  }
-//----------------
-  //Exclude all Weak Jets for matching
-  //if (doFxFx && typeIdx[2].size()>0) {
-  //      return MORE_JETS;
-  //} // 2A
-  //keep only Weak Jets for matching
-  //if (doFxFx && typeIdx[2].size()==0) {
-  //      return MORE_JETS;
-  //} // 2B
-  //keep only lowest multiplicity sample @0
-  //if (doFxFx && npNLO()==1) {
-  //      return MORE_JETS;
-  //} // 2C
-  //keep only highest multiplicity sample @1
-  //if (doFxFx && npNLO()==0) {
-  //      return MORE_JETS;
-  //} // 2D
-//--------------
+  // Note: Difference compared to old versions of FxFx (cf. arXiv:2108.07826):
+  // For FxFx, change nRequested subtracting the typeIdx[2] partons
+  // Exclude the highest multiplicity sample in the case of real emissions,
+  // exclude all typeIdx[2] particles (npNLO=multiplicity,
+  // nJetMax=njmax(shower_card), typeIdx[2]="Weak" jets)
+  int nRequested = ( doFxFx
+                   && !(npNLO()==nJetMax
+                   && npNLO()==((int)typeIdx[2].size()-1) ))
+                 ? npNLO()-typeIdx[2].size()
+                 : nParton;
+
+  // Note: Difference compared to old versions of FxFx (cf. arXiv:2108.07826):
+  // For FxFx veto the real emissions that have only typeIdx=2 partons
+  // Exclude the highest multiplicity sample
+  if ( doFxFx && npNLO()<nJetMax && typeIdx[2].size()>0
+    && npNLO()==((int)typeIdx[2].size()-1)) {
+    return MORE_JETS;
+  }
+
   // Veto event if too few hadronic jets
   if ( nCLjets < nRequested ) return LESS_JETS;
 
@@ -1729,11 +1651,10 @@
     // For FxFx, in the non-highest multipicity, all jets need to matched to
     // partons. For nCLjets > nRequested, this is not possible. Hence, we can
     // veto here already.
-    // if ( doFxFx && nRequested < nJetMax && nCLjets > nRequested )
-    //Step-FxFx-5:Change the nRequested to npNLO() in the first condition
-    //Before in Step-FxFx-3 it was nRequested=npNLO() for FxFx
-    //This way we correctly select the non-highest multipicity regardless the nRequested
-     if ( doFxFx && npNLO() < nJetMax && nCLjets > nRequested )
+    // Note: Difference wrt. old versions of FxFx (cf. arXiv:2108.07826):
+    // Change the nRequested to npNLO() in the first condition. This way we
+    // correctly select the non-highest multipicity regardless the nRequested
+    if ( doFxFx && npNLO() < nJetMax && nCLjets > nRequested )
       return MORE_JETS;
 
     // Now continue in inclusive mode.
@@ -1863,13 +1784,10 @@
 
   // Jet matching veto for FxFx
   if (doFxFx) {
-//    if ( nRequested <  nJetMax && nMatched != nRequested )
-//      return UNMATCHED_PARTON;
-//    if ( nRequested == nJetMax && nMatched <  nRequested )
-//      return UNMATCHED_PARTON;
-      //Step-FxFx-6:Change the nRequested to npNLO() in the first condition (like in Step-FxFx-5)
-      //Before in Step-FxFx-3 it was nRequested=npNLO() for FxFx
-      //This way we correctly select the non-highest multipicity regardless the nRequested
+      // Note: Difference wrt. old versions of FxFx (cf. arXiv:2108.07826):
+      // Change the nRequested to npNLO() in the first condition (cf. above).
+      // This way we correctly select the non-highest multipicity regardless
+      // the nRequested
       if ( npNLO() <  nJetMax && nMatched != nRequested )
         return UNMATCHED_PARTON;
       if ( npNLO() == nJetMax && nMatched <  nRequested )
```

Even if the logic is the same (no idea), that's of course not what madgraph checks, but only if the file is identical, so we'd have to patch madgraph to accept the separate file.

> It is just a quite hard no to replace a user-facing header like this when it is supposed to reflect the compiled libraries,

I think the `JetMatching.h` header is generally only used at "runtime-compile-time" so it wouldn't matter for dependents of pythia8, as pythia8 itself never compiles this header.

Your point about future compatibility concerns is completely valid, but that's just a reflection of the fact that the HEP generator software ecosystem is comprehensively broken.

Your solution with the patch might work, but given how difficult it is to develop patches in spack, I'm not willing to take that on. I'll make a private fork of the madgraph package and continue with the changes from this PR.

paulgessinger:
As discussed in the comment, closing this in favor of a private fork.

PR #2109, opened 2025-10-26
Fix pkgconfig dependency for several packages

michaelkuhn:
pkgconfig is the virtual dependency that should be used in most cases.


fangohr (approved):
Looks good to me. 

Additional thoughts @lang-m ?

lang-m:
Should be fine for octopus.

srekolam (approved):
Look Okay for rocm-smi-lib

alalazo (approved):
Gitlab needed some manual restart, but that's ok 🫠

PR #2114, opened 2025-10-27
actsvg: patch version numbers into source

stephenswat:
This commit adds a patch method to the actsvg package, as most versions of this package have an incorrect version in the source code. The patch replaces this (usually outdated) version number with the version number known to Spack.


stephenswat:
@spackbot re-run pipeline

stephenswat:
@wdconinc this fix will be needed to get #1924 to work.

wdconinc (approved):

PR #2115, opened 2025-10-27
Update freetype versions

wraith1995:
Add two versions to freetype.


michaelkuhn (commented):
~~This is unrelated to this PR, but could you also remove the preference for version 2.13.2?~~

Apart from that, both versions work for me (Fedora 42, GCC 15).

michaelkuhn (approved):

PR #2123, opened 2025-10-27
darshan: add v3.5.0

wkliao:
Darshan version 3.5.0 was released on October 20, 2025.
Release tar ball and release note are available on
https://github.com/darshan-hpc/darshan/releases

cc: @carns

### DIFF
--- a/repos/spack_repo/builtin/packages/darshan_runtime/package.py
+++ b/repos/spack_repo/builtin/packages/darshan_runtime/package.py
@@ -120,7 +120,7 @@ class DarshanRuntime(AutotoolsPackage):
     variant("group_readable_logs", default=False, description="Write group-readable logs")
 
     def url_for_version(self, version):
-        if version < Version("3.5.0"):
+        if self.spec.satisfies("@:3.4"):
             return f"https://web.cels.anl.gov/projects/darshan/releases/darshan-{version}.tar.gz"
         else:
             return f"https://github.com/darshan-hpc/darshan/archive/refs/tags/{version}.tar.gz"
--- a/repos/spack_repo/builtin/packages/darshan_runtime/package.py
+++ b/repos/spack_repo/builtin/packages/darshan_runtime/package.py
@@ -17,7 +17,7 @@ class DarshanRuntime(AutotoolsPackage):
     systems where you intend to instrument MPI applications."""
 
     homepage = "https://www.mcs.anl.gov/research/projects/darshan/"
-    url = "https://web.cels.anl.gov/projects/darshan/releases/darshan-3.4.0.tar.gz"
+    url = "https://github.com/darshan-hpc/darshan/archive/refs/tags/3.5.0.tar.gz"
     git = "https://github.com/darshan-hpc/darshan.git"
 
     maintainers("carns", "wkliao")
@@ -26,6 +26,7 @@ class DarshanRuntime(AutotoolsPackage):
     test_requires_compiler = True
 
     version("main", branch="main", submodules=True)
+    version("3.5.0", sha256="76eac5a3ff556476ef8df53aa61851756bf6dcab59f5b92bb307abfdcbcd7f36")
     version("3.4.7", sha256="115a6d840b3bdb30751c271c0dec098b25bed2f8c77175125c133564d76afe5b")
     version("3.4.6", sha256="092b35e7af859af903dce0c51bcb5d3901dd0d9ad79d1b2f3282692407f032ee")
     version("3.4.5", sha256="1c017ac635fab5ee0e87a6b52c5c7273962813569495cb1dd3b7cfa6e19f6ed0")
@@ -104,8 +105,25 @@ class DarshanRuntime(AutotoolsPackage):
         default="none",
         description="Path to centralized, formatted Darshan log directory",
     )
-    variant("mmap_logs", default=False, description="Use mmap to store Darshan log data")
+
+    variant(
+        "mmap_logs",
+        default=False,
+        description="Use mmap to store Darshan log data",
+        when="@:3.4.7",
+    )
+
+    variant(
+        "mmap_logs", default=True, description="Use mmap to store Darshan log data", when="@3.5.0:"
+    )
+
     variant("group_readable_logs", default=False, description="Write group-readable logs")
+
+    def url_for_version(self, version):
+        if version < Version("3.5.0"):
+            return f"https://web.cels.anl.gov/projects/darshan/releases/darshan-{version}.tar.gz"
+        else:
+            return f"https://github.com/darshan-hpc/darshan/archive/refs/tags/{version}.tar.gz"
 
     @property
     def configure_directory(self):
--- a/repos/spack_repo/builtin/packages/darshan_util/package.py
+++ b/repos/spack_repo/builtin/packages/darshan_util/package.py
@@ -14,14 +14,15 @@ class DarshanUtil(AutotoolsPackage):
     log files produced by Darshan (runtime)."""
 
     homepage = "https://www.mcs.anl.gov/research/projects/darshan/"
-    url = "https://web.cels.anl.gov/projects/darshan/releases/darshan-3.4.0.tar.gz"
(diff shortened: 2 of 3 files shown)


carns (approved):

repos/spack_repo/builtin/packages/darshan_runtime/package.py
```diff
@@ -17,7 +17,6 @@ class DarshanRuntime(AutotoolsPackage):
     systems where you intend to instrument MPI applications."""
 
     homepage = "https://www.mcs.anl.gov/research/projects/darshan/"
-    url = "https://web.cels.anl.gov/projects/darshan/releases/darshan-3.4.0.tar.gz"
```
alecbcs, line 20: Instead of deleting this line, we should update the url property here to the new url format.

repos/spack_repo/builtin/packages/darshan_util/package.py
```diff
@@ -14,14 +14,14 @@ class DarshanUtil(AutotoolsPackage):
     log files produced by Darshan (runtime)."""
 
     homepage = "https://www.mcs.anl.gov/research/projects/darshan/"
-    url = "https://web.cels.anl.gov/projects/darshan/releases/darshan-3.4.0.tar.gz"
```
alecbcs, line 17: Same here. This property is used to find new versions with `spack versions` and `spack checksum` commands.

alecbcs (changes requested):
A few quick notes. Otherwise this looks good to me. Thanks @wkliao!

repos/spack_repo/builtin/packages/darshan_runtime/package.py
wkliao: Starting from this 3.5.0 version, we switch the software release repo to Github.
Will the URL pointing to the repo's release be the right one to use? i.e.
`https://github.com/darshan-hpc/darshan/releases` or it needs to contain
the tar ball name of a specific version?

alecbcs: It would be a specific version url like:
```python3
url = "https://web.cels.anl.gov/projects/darshan/releases/darshan-3.5.0.tar.gz"
```

Spack is smart enough find and replace the version in a url dynamically.

wkliao: Thanks. I have revised accordingly and force pushed the changes to this PR.

repos/spack_repo/builtin/packages/darshan_runtime/package.py
```diff
@@ -104,9 +105,26 @@ class DarshanRuntime(AutotoolsPackage):
...
+    )
+
     variant("group_readable_logs", default=False, description="Write group-readable logs")
 
+    def url_for_version(self, version):
+        if version < Version("3.5.0"):
+            return f"https://web.cels.anl.gov/projects/darshan/releases/darshan-{version}.tar.gz"
+        else:
+            return f"https://github.com/darshan-hpc/darshan/archive/refs/tags/{version}.tar.gz"
+
```
alecbcs, line 127:
```suggestion
    def url_for_version(self, version):
        if self.spec.satisfies("@:3.4"):
            return f"https://web.cels.anl.gov/projects/darshan/releases/darshan-{version}.tar.gz"
        else:
            return f"https://github.com/darshan-hpc/darshan/archive/refs/tags/{version}.tar.gz"

```

We can drop the additional version object creation here by instead using the `satisfies()` directive.

repos/spack_repo/builtin/packages/darshan_util/package.py
```diff
@@ -77,6 +78,12 @@ class DarshanUtil(AutotoolsPackage):
 
     patch("retvoid.patch", when="@3.2.0:3.2.1")
 
+    def url_for_version(self, version):
+        if version < Version("3.5.0"):
+            return f"https://web.cels.anl.gov/projects/darshan/releases/darshan-{version}.tar.gz"
+        else:
+            return f"https://github.com/darshan-hpc/darshan/archive/refs/tags/{version}.tar.gz"
```
alecbcs, line 85:
```suggestion
    def url_for_version(self, version):
        if self.spec.satisfies("@:3.4"):
            return f"https://web.cels.anl.gov/projects/darshan/releases/darshan-{version}.tar.gz"
        else:
            return f"https://github.com/darshan-hpc/darshan/archive/refs/tags/{version}.tar.gz"
```

Same suggestion here.

alecbcs (changes requested):
Two quick suggestions. Otherwise it looks good to me.

repos/spack_repo/builtin/packages/py_darshan/package.py
```diff
@@ -56,6 +57,8 @@ class PyDarshan(PythonPackage):
     for v in ["3.4.0", "3.4.1", "3.4.2", "3.4.3", "3.4.4", "3.4.5", "3.4.6", "3.4.7"]:
         depends_on(f"darshan-util@{v}", when=f"@{v}", type=("build", "run"))
 
+    depends_on("darshan-util@3.5.0", when="@3.5.0", type=("build", "run"))
```
alecbcs:
```suggestion
for v in ["3.4.0", "3.4.1", "3.4.2", "3.4.3", "3.4.4", "3.4.5", "3.4.6", "3.4.7", "3.5.0"]:
depends_on(f"darshan-util@{v}", when=f"@{v}", type=("build", "run"))
```

We can just add the following instead of creating a full new line.

repos/spack_repo/builtin/packages/darshan_runtime/package.py
```diff
@@ -104,9 +105,26 @@ class DarshanRuntime(AutotoolsPackage):
...
+    variant(
+        "mmap_logs",
+        default=False,
+        description="Use mmap to store Darshan log data",
+        when="@:3.4.7",
+    )
+
+    variant(
+        "mmap_logs", default=True, description="Use mmap to store Darshan log data", when="@3.5.0:"
+    )
```
alecbcs, line 118: Can you explain why we're now enabling this variant by default? We explicitly try to keep almost all variants as `default=False` in order to minimize the install size of packages and reduce the complexity of dependency trees.

alecbcs (changes requested):

repos/spack_repo/builtin/packages/py_darshan/package.py
wkliao: I just thought line 57 will grow quickly.
In the future, such dependency will be added for each new release.
Is there a better way in python for such task?

repos/spack_repo/builtin/packages/darshan_runtime/package.py
wkliao: This is because the default configure option `mmap_logs` has been changed to `True` in this new release, and we intend to keep it for all future releases.

wkliao:
Hi, @alecbcs 
I believe I have made modifications based on your comments. Please let me know if there is anything else that can improve this PR.

repos/spack_repo/builtin/packages/py_darshan/package.py
alecbcs: It's something on our roadmap, but not something we've implemented yet. For now I'd recommend either adding to the existing array, or rewriting all of the `depends_on` directives to be on their own lines.

alecbcs:
Hi @wkliao, apologies this PR fell off my radar. I'd still suggest a modification to the `depends_on` directives, either add to the existing array or refactor the array out into a set of separate lines so that we're doing one or the other but not both. There looks to be a simple merge conflict, if you can rebase everything else looks good to me.

wkliao:
Thanks. I have updated the PR with the suggested changes and rebased it.

wkliao:
Hi, @alecbcs 
Just a friendly reminder.
Please let me know if there are other suggestions.

alecbcs:
@spackbot run pipeline

alecbcs (approved):
Looks good to me. Thanks @wkliao!

PR #2124, opened 2025-10-27
py-flake8: new versions

KineticTheory:

### DIFF
--- a/repos/spack_repo/builtin/packages/py_flake8/package.py
+++ b/repos/spack_repo/builtin/packages/py_flake8/package.py
@@ -16,6 +16,8 @@ class PyFlake8(PythonPackage):
 
     license("MIT")
 
+    version("7.3.0", sha256="fe044858146b9fc69b551a4b490d69cf960fcb78ad1edcb84e7fbb1b4a8e3872")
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/py_flake8/package.py
```diff
@@ -16,6 +16,8 @@ class PyFlake8(PythonPackage):
 
     license("MIT")
 
+    version("7.3.0", sha256="fe044858146b9fc69b551a4b490d69cf960fcb78ad1edcb84e7fbb1b4a8e3872")
+    version("7.1.2", sha256="c586ffd0b41540951ae41af572e6790dbd49fc12b3aa2541685d253d9bd504bd")
```
alalazo, line 20: Did you check if the major version bump comes with new constraints on the dependencies?

alalazo (changes requested):

repos/spack_repo/builtin/packages/py_flake8/package.py
KineticTheory: @alalazo Thanks for the reminder.  I've updated the new dependency requirements.

alalazo (approved):

PR #2133, opened 2025-10-28
Trilinos: Adding cusparse variants

Ivan-Hrd:
Creates the variant cusparse to enable the cuSPARSE TPL.

### DIFF
--- a/repos/spack_repo/builtin/packages/trilinos/package.py
+++ b/repos/spack_repo/builtin/packages/trilinos/package.py
@@ -424,6 +424,7 @@ class Trilinos(CMakePackage, CudaPackage, ROCmPackage):
         depends_on("kokkos+hip_relocatable_device_code~shared", when="+rocm_rdc")
         depends_on("kokkos-kernels~shared", when="+cuda_rdc")
         depends_on("kokkos-kernels~shared", when="+rocm_rdc")
+        depends_on("kokkos-kernels+cusparse", when="+cusparse")
         depends_on("kokkos~complex_align")
         depends_on("kokkos@=4.7.01", when="@master:")
         depends_on("kokkos@=4.5.01", when="@16.1")
--- a/repos/spack_repo/builtin/packages/trilinos/package.py
+++ b/repos/spack_repo/builtin/packages/trilinos/package.py
@@ -131,6 +131,7 @@ class Trilinos(CMakePackage, CudaPackage, ROCmPackage):
     # TPLs (alphabet order)
     variant("adios2", default=False, description="Enable ADIOS2")
     variant("boost", default=False, description="Compile with Boost")
+    variant("cusparse", default=False, description="Enable cuSPARSE support")
     variant("hdf5", default=False, description="Compile with HDF5")
     variant("hypre", default=False, description="Compile with Hypre preconditioner")
     variant("mpi", default=True, description="Compile with MPI parallelism")
@@ -376,6 +377,7 @@ class Trilinos(CMakePackage, CudaPackage, ROCmPackage):
         msg="trilinos~wrapper+cuda can only be built with the Clang compiler",
     )
     conflicts("+cuda_rdc", when="~cuda")
+    conflicts("+cusparse", when="~cuda")
     conflicts("+rocm_rdc", when="~rocm")
     conflicts("+wrapper", when="~cuda")
     conflicts("+wrapper", when="%clang")
@@ -905,6 +907,7 @@ class Trilinos(CMakePackage, CudaPackage, ROCmPackage):
             ("ADIOS2", "adios2", "adios2"),
             ("Boost", "boost", "boost"),
             ("CUDA", "cuda", "cuda"),
+            ("CUSPARSE", "cusparse", "cuda"),
             ("HDF5", "hdf5", "hdf5"),
             ("HYPRE", "hypre", "hypre"),
             ("MUMPS", "mumps", "mumps"),


repos/spack_repo/builtin/packages/trilinos/package.py
```diff
@@ -905,6 +910,7 @@ def define_tpl(trilinos_name, spack_name, have_dep):
             ("ADIOS2", "adios2", "adios2"),
             ("Boost", "boost", "boost"),
             ("CUDA", "cuda", "cuda"),
+            ("CUSPARSE", "cusparse", "cuda"),
```
cgcgcg, line 911: Should this be
```
("CUSPARSE", "cusparse", "cusparse"),
```

cgcgcg (commented):
I am not sure about these changes. I believe spack encourages short tests, but not full test suites. I am also not sure if adding a "build tests" variant for a single Trilinos package is a good idea. We are in the process of refactoring the entire spack package anyway, so chances are that these changes will get blown away in the future.

For the cusparse change: does Trilinos currently not enable cusparse by default when cuda is enabled? That seems like a problem.

repos/spack_repo/builtin/packages/trilinos/package.py
rbourgeois33: Hello @cgcgcg !

Thanks for for the feedback. I work with @Ivan-Hrd. Let me give you more background.

### Why a new variant ?

The reason why we want a test/example variant is that we are building an algebraic multigrid library comparison tool (`solver-marker`). We install the several libraries (including yours) via spack. 

If the library includes a driver (sample code that reads a `.mtx`file and a set of parameters and produces a system resolution) such as ([this one](https://github.com/trilinos/Trilinos/blob/master/packages/muelu/example/basic/Stratimikos.cpp), [that one](https://github.com/trilinos/Trilinos/blob/master/packages/muelu/test/tutorial/TutorialDriver.cpp), and [also that one](https://github.com/trilinos/Trilinos/blob/7c8afbd28691f64cff071e579701bfe4ec65b837/packages/muelu/test/scaling/Driver.cpp)) we want to use it and not maintain one on our side.

In the case of Trilinos,  these drivers are from test/tutorial, not from the core sources. That is why I think we need a spack variant to build them. 

Maybe we could also simply install Trilinos without test/example variant, clone a driver and compile it as external code ?

side questions: which input deck is the most maintained for MueLu ?

### Cusparse
I just tested locally on the main branch of trilinos (commit `7c8afbd28691`)

By using the following `cmake` command:

```bash
        cmake -B build -S . \
        -D CMAKE_INSTALL_PREFIX=install \
        -D CMAKE_BUILD_TYPE=RELEASE \
        -D BUILD_SHARED_LIBS:BOOL=ON \
        -D Trilinos_ENABLE_EXPLICIT_INSTANTIATION:BOOL=ON \
        -D Trilinos_ENABLE_TESTS:BOOL=OFF \
        -D Trilinos_ENABLE_EXAMPLES:BOOL=OFF \
        -D Trilinos_ENABLE_MueLu:BOOL=ON \
        -D Kokkos_ENABLE_DEBUG=OFF \
        -D Kokkos_ENABLE_DEBUG_BOUNDS_CHECK=OFF \
        -D TPL_ENABLE_BLAS:BOOL=ON \
        -D TPL_ENABLE_MPI:BOOL=ON \
        -D TPL_ENABLE_CUSOLVER=ON \
        -D TPL_ENABLE_CUBLAS=ON \
        -D TPL_ENABLE_CUDA=ON \
        -D MueLu_ENABLE_TESTS:BOOL=ON \
        -D MueLu_ENABLE_EXAMPLES:BOOL=ON \
```

The following CUSPARSE / CUSOLVER cmake options are not set:

```bash
 Amesos2_ENABLE_CUSOLVER                                                                                                                                                                     
 Amesos2_ENABLE_CUSPARSE
 KokkosKernels_CUSOLVER_ROOT                                                                                                                                                                 
 KokkosKernels_CUSPARSE_ROOT  
 MueLu_ENABLE_CUSPARSE   
 TPL_ENABLE_CSparse      
  TpetraTSQR_ENABLE_CUBLAS                                                                                                                                                                    
 TpetraTSQR_ENABLE_CUSOLVER                                                                                                                                                                                                                                                                                                                                      
```

Therefore, activating CUDA through `TPL_ENABLE_CUDA` does not seem to imply the use of cusparse everywhere ? At least not explicitly.  

Cheers !

kuberry: It looks like cuBLAS and cuSOLVER are enabled by default when CUDA is enabled, but this isn't the case for cuSPARSE. Also, cuSOLVER is disabled by default (when +cuda) in Kokkos-Kernels because cuSPARSE is disabled by default.

Here is a comment and the kokkos-kernels logic to deal with cuSPARSE:
https://github.com/trilinos/Trilinos/blob/master/packages/kokkos-kernels/cmake/kokkoskernels_features.cmake#L43-L64

Echoing @cgcgcg, I would recommend going with the second option you all mentioned about removing `define_from_variant("MueLu_ENABLE_TESTS", "muelu_tests"),`,
 `conflicts("+muelu_tests", when="~muelu")`, and `variant("muelu_tests", default=False, description="Enable build of MueLu test")` and then cloning the tool that you want to use from muelu tests and building it as part of your dependent package.

`("CUSPARSE", "cusparse", "cuda")` is appropriate because the third argument is the spack spec containing the library, and cuSPARSE ships with CUDA runtime according to https://docs.nvidia.com/cuda-libraries/index.html.

rbourgeois33: Thanks for the insight. Let me know if I get this straight:

It is expected that using CUDA does not imply using Cusparse. Therefore, a Cusparse variant in the trilinos spack recipe should be added via the present PR.

However, we noticed that setting `TPL_ENABLE_CUSPARSE=ON` does not imply `MueLu_ENABLE_CUSPARSE ` being set to `ON`. Is that expected ? If not, should it be fixed at the CMakeLists.txt level or at the spack recipe level ? If yes, should there be a muelu_cusparse variant ?

We will go with the 2nd option, thanks !

Thanks,

kuberry: Can you provide a snippet of some output relating to `MueLu_ENABLE_CUSPARSE` not being enabled? 

I'm seeing
```
./packages/muelu//cmake/Dependencies.cmake:9:SET(LIB_OPTIONAL_DEP_TPLS Boost MATLAB AmgX ViennaCL MKL Avatar CUSPARSE MAGMASparse mlpack)
```
which populates 
```
./packages/muelu//cmake/MueLu_config.hpp.in:101:#cmakedefine HAVE_MUELU_CUSPARSE
```
and this seems like it should be set if the TPL for CUSPARSE is set (which it should be with your `+cusparse` variant being enabled).

rbourgeois33: Sure, buy running the following command, activating MueLU and TPL CUDA and Cusparse:

```bash
cmake .. -D Trilinos_ENABLE_MueLu=ON -D TPL_ENABLE_CUSPARSE=ON -D TPL_ENABLE_CUDA=ON 2>&1 | tee cmake_output.log
```

I get the this [cmake_output.log](https://github.com/user-attachments/files/23261163/cmake_output.log)

In which we can indeed see `Setting MueLu_ENABLE_CUSPARSE=ON since Trilinos_ENABLE_MueLu=ON AND TPL_ENABLE_CUSPARSE=ON`

But when inspecting the 
[CMakeCache.txt](https://github.com/user-attachments/files/23261300/CMakeCache.txt), we can see that `MueLu_ENABLE_CUSPARSE` is not set:

```bash
//Enable optional library support in the package MueLu for the
// package CUSPARSE.  Set to 'ON', 'OFF', or leave empty to allow
// for other logic to decide.
MueLu_ENABLE_CUSPARSE:STRING=
```

I do not know if that means `MueLu_ENABLE_CUSPARSE` is somehow set to `ON` internally ? The comment suggests that it might be the case but I am not sure how to double check that. 

Thanks !

cgcgcg: Please check packages/muelu/src/MueLu_config.hpp in your build directory. There should be a line 
```
#define HAVE_MUELU_CUSPARSE
```
if it is enabled and
```
/* #undef HAVE_MUELU_CUSPARSE */
```
if it is not enabled.

Just for context: this variable enables some additional testing where cusparse is used directly in MueLu. It does not change what backend Kokkos Kernels (and in turn Tpetra and MueLu) is using.

rbourgeois33: Thanks ! 

Yes, `#define HAVE_MUELU_CUSPARSE` is indeed un-commented in `build/packages/muelu/src/MueLu_config.hpp`

Therefore, it seems like we only need a `+cusparse` variant of the spack recipe. 

@Ivan-Hrd will update the present PR.

Cheers

Ivan-Hrd: Hello everyone,
I have updated the pull request, it now handle only the cusparse variant, activating ```TPL_ENABLE_CUSPARSE``` through the tpl_variant_map adding the ``` ("CUSPARSE", "cusparse", "cuda") ``` element.

Thanks

kuberry (approved):

repos/spack_repo/builtin/packages/trilinos/package.py
Ivan-Hrd: Hello again,
After a [discussion](https://kokkosteam.slack.com/archives/C067P67GNBC/p1762247351832019) with @cgcgcg and @rbourgeois33, I also activated the cusparse variant in the kokkos-kernels dependency, that way, whatever path is used is used for performing a sparse operation (through CUSPARSE directly or through Kokkos-Kernels) both will be available and cusparse we be used in the end.

rbourgeois33:
Hello !

What is the status of the integration of this branch ? Should we make any supplementary modifications ?

Thanks.

psakievich (approved):
LGTM now that the dust seems to have settled for the trilinos core devs.

kuberry (approved):

PR #2138, opened 2025-10-28
Build MCT with -fPIC

climbfuji:
## Description

All in the title.

`MCT` is no longer developed and only supports static builds. In order to link the library to shared libraries, it needs `-fPIC`.

Tested on my dev machine.


repos/spack_repo/builtin/packages/mct/package.py
```diff
@@ -28,3 +28,15 @@ class Mct(AutotoolsPackage):
     depends_on("fortran", type="build")  # generated
 
     depends_on("mpi")
+
+    # MCT build is static only, need -fPIC to link in shared environment;
+    # unfortunately, also need to set FCFLAGS and not just FFLAGS
+    def flag_handler(self, name, flags):
+        if name == "cflags":
+            flags.append(self.compiler.cc_pic_flag)
```
alalazo:
```suggestion
            flags.append(self["c"].pic_flag)
```

repos/spack_repo/builtin/packages/mct/package.py
```diff
@@ -28,3 +28,15 @@ class Mct(AutotoolsPackage):
...
+        elif name == "fflags":
+            flags.append(self.compiler.fc_pic_flag)
```
alalazo:
```suggestion
            flags.append(self["fortran"].pic_flag)
```

alalazo (changes requested):
Since compilers are now packages, I suggest using the syntax below. The other interface is there to support legacy code.

repos/spack_repo/builtin/packages/mct/package.py
```diff
@@ -28,3 +28,15 @@ class Mct(AutotoolsPackage):
...
+    # unfortunately, also need to set FCFLAGS and not just FFLAGS
+    def flag_handler(self, name, flags):
+        if name == "cflags":
+            flags.append(self["c"].pic_flag)
+        elif name == "fflags":
+            flags.append(self["fortran"].pic_flag)
+        return (None, flags, None)
+
+    def setup_build_environment(self, env: EnvironmentModifications) -> None:
+        env.set("FCFLAGS", self.compiler.fc_pic_flag)
```
climbfuji:
```suggestion
        env.set("FCFLAGS", self["fortran"].pic_flag)
```
I guess this should work here as well?

alalazo: Yes, it should work. Let me know if you see any issue.

climbfuji: Tested, works as expected

alalazo (approved):

PR #2141, opened 2025-10-28
pkg-config: add -Wno-error=int-conversion for older versions of rocmcc

afzpatel:
The changes from https://github.com/spack/spack-packages/pull/2069 should be used for all versions of `llvm-amdgpu`


afzpatel:
@tldahlgren please review this pr.

becker33 (approved):

PR #2137, opened 2025-10-28
py-fenics-ffcx: add version 0.10.0 and 0.10.1

schnellerhase:
Depends on https://github.com/spack/spack-packages/pull/2087 and https://github.com/spack/spack-packages/pull/2082

PR #2142, opened 2025-10-28
fenics-dolfinx: add version 0.10.0.post1

schnellerhase:
Depends on https://github.com/spack/spack-packages/pull/2136 and https://github.com/spack/spack-packages/pull/2137


jhale (approved):

repos/spack_repo/builtin/packages/fenics_dolfinx/package.py
```diff
@@ -70,7 +75,7 @@ class FenicsDolfinx(CMakePackage):
 
     depends_on("adios2@2.8.1:+mpi", when="@0.9: +adios2")
     depends_on("adios2+mpi", when="+adios2")
-    for ver in ("main", "0.9", "0.8", "0.7", "0.6"):
+    for ver in ("main", "0.10", "0.9", "0.8", "0.7", "0.6"):
```
jhale: Does this work with post0? In any case, we will not continue to tightly couple version numbers across components, so this will need to be removed.

schnellerhase: Instead write it out explicitly as
```python
depends_on("fenics-ufcx@0.10", when="@0.10")
depends_on("fenics-basix@0.10", when="@0.10")
...
```
to allow for future decoupling of releases?

jhale:
Not that we've yet implemented testing in Spack, but this should probably be switched to `v0.10.post2` which contains a fix for `pytest>=9`.

repos/spack_repo/builtin/packages/fenics_dolfinx/package.py
jhale: Yes.

jhale (changes requested):
Misses propagation of cmake build arg variant into cmake_args:

```
diff --git a/repos/spack_repo/builtin/packages/fenics_dolfinx/package.py b/repos/spack_repo/builtin/packages/fenics_dolfinx/package.py
index 14da6a085e..f60ad492eb 100644
--- a/repos/spack_repo/builtin/packages/fenics_dolfinx/package.py
+++ b/repos/spack_repo/builtin/packages/fenics_dolfinx/package.py
@@ -80,6 +80,7 @@ class FenicsDolfinx(CMakePackage):

     def cmake_args(self):
         return [
+            self.define_from_variant("CMAKE_BUILD_TYPE", "build_type"),
             self.define("DOLFINX_SKIP_BUILD_TESTS", True),
             self.define_from_variant("DOLFINX_ENABLE_PETSC", "petsc"),
             self.define_from_variant("DOLFINX_ENABLE_SLEPC", "slepc"),
```

repos/spack_repo/builtin/packages/fenics_dolfinx/package.py
```diff
@@ -18,6 +18,9 @@ class FenicsDolfinx(CMakePackage):
     license("LGPL-3.0-or-later")
 
     version("main", branch="main", no_cache=True)
+    version(
+        "0.10.0.post1", sha256="43ebfaf89fcb7809b11b94ba3e8ea8e56a99d724a45c51e99e39a35c577c441e"
```
jhale: Do we know if reflecting post release numbering is correct in Spack? After all, there is no functional difference with 0.10, just packaging tweaks.

repos/spack_repo/builtin/packages/fenics_dolfinx/package.py
schnellerhase: > Does this work with post0?

Think it does https://spack.readthedocs.io/en/latest/packaging_guide_creation.html#version-ordering.

repos/spack_repo/builtin/packages/fenics_dolfinx/package.py
schnellerhase: The hash is checked against the tar ball of the corresponding git tag, so we can't rename another git tag to `0.10.0` here.

schnellerhase:
Superseded by https://github.com/spack/spack-packages/pull/2673.

PR #2154, opened 2025-10-29
pdsh: use chaos/pdsh (offical) + add new versions

skwde:


michaelkuhn (commented):
Looks good to me, but does not compile with GCC 15. Could you add something like this (with the proper version constraint)?
```
patch(
    "https://github.com/chaos/pdsh/commit/01b1a2150e5d2c1a065aeb05a504f1f92b2cc147.patch?full_index=1",
    sha256="3effcc73c7b5efbe73d12579eb09982c160a116edf86f0bdb315daeebce0802e",
    when="@...",
)
```

skwde:
> Looks good to me, but does not compile with GCC 15. Could you add something like this (with the proper version constraint)?
> 
> ```
> patch(
>     "https://github.com/chaos/pdsh/commit/01b1a2150e5d2c1a065aeb05a504f1f92b2cc147.patch?full_index=1",
>     sha256="3effcc73c7b5efbe73d12579eb09982c160a116edf86f0bdb315daeebce0802e",
>     when="@...",
> )
> ```

@michaelkuhn I identified the commit after which the patch should still work. However spack refuses to read 

```py
    patch(
        "https://github.com/chaos/pdsh/commit/01b1a2150e5d2c1a065aeb05a504f1f92b2cc147.patch?full_index=1",
        sha256="3effcc73c7b5efbe73d12579eb09982c160a116edf86f0bdb315daeebce0802e",
        when="@8c173e645dfc5cb6619c4d0647f3fe495ad4ce7f:",
    ) # first commit after 2924dd8ff07b31a81c4a81254ae1e1f81b39595e (shortly after pdsh-2.32)

```

I get

```out
$ spack info pdsh
==> Error: cannot load package 'pdsh' from the 'builtin' repository: unexpected characters in the spec string
@8c173e645dfc5cb6619c4d0647f3fe495ad4ce7f:
```

and I cannot find info on how I give commit hashs to when. Any hint?

EDIT:

adding something like

```py
 # ensure also that gcc@15 is not used when the patch cannot be applied
 conflicts("%gcc@15:", when="@2924dd8ff07b31a81c4a81254ae1e1f81b39595e:")
```

would also be good I think.

EDIT2:

Of course the simple solution would be to do something like

```py
    patch(
        "https://github.com/chaos/pdsh/commit/01b1a2150e5d2c1a065aeb05a504f1f92b2cc147.patch?full_index=1",
        sha256="3effcc73c7b5efbe73d12579eb09982c160a116edf86f0bdb315daeebce0802e",
        when="@2.33:",
    ) # strictly the patch can be applied after commit 2924dd8ff07b31a81c4a81254ae1e1f81b39595e (shortly after pdsh-2.32)

    # ## conflict when patch cannot be applied before and at commit 2924dd8ff07b31a81c4a81254ae1e1f81b39595e
    conflicts("%gcc@15:", when="@:2.32")
```

skwde:
Note, force push easy, working and simple solution to get it merged...

michaelkuhn (approved):
Sorry for the late reply, LGTM.

PR #2156, opened 2025-10-29
py_keras: enable multi values in variant backend

LydDeb:
Allows the installation of multiple backends. This avoids potential compatibility issues due to version constraints caused by a decoupled installation. Specifically, the change would be from this:
```
spack install py-keras backend=torch
spack install py-tensorflow
```

to this:
```
spack install py-keras backend=torch,tensorflow
```

The KERAS_BACKEND variable takes the first value of the backend variant.


repos/spack_repo/builtin/packages/py_keras/package.py
```diff
@@ -78,10 +78,10 @@ class PyKeras(PythonPackage):
     # TODO: add openvino backend (keras 3.8+)
     variant(
         "backend",
-        default="tensorflow",
-        description="backend library",
+        default="torch",
```
adamjstewart: Not sure how I feel about defaulting to PyTorch. It's what I use, but it's probably not what most Keras users use.

LydDeb: Indeed, at first glance it would seem that the most popular backend in the Keras community is still TensorFlow.

adamjstewart (approved):

LydDeb:
I suspect the error:
```
File "/builds/spack/spack-packages/.ci/tmp/spack/lib/spack/spack/url_buildcache.py", line 1387, in from_string
return cls(parts[0], int(parts[1]))
^^^^^^^^^^^^^
ValueError: invalid literal for int() with base 10: 'ariant_backend/ml-linux-x86_64-rocm
```
is due to the fact that my branch name is `py_keras__variant_backend` and the `from_string` class method of the MirrorURLAndVersion class uses '__v' as the separator.

PR #2165, opened 2025-10-29
Update compiler_wrapper/cc.sh: choose compiler in mode 'ld'

climbfuji:
## Description 

Fixes an issue when checking which compiler is configured when the compiler wrapper is invoked in `ld` mode. Without this bug fix, spack assumes that a package that doesn't depend on C must depend on CXX. But there are packages that depend on Fortran only.

## Testing

Tested for a package that depends on Fortran only and that invokes `ld` directly during the build.

Because this is such an important script in spack, I suggest that more folks test this thoroughly.


climbfuji:
@spackbot run pipeline

climbfuji:
@spackbot run pipeline

repos/spack_repo/builtin/packages/compiler_wrapper/cc.sh
```diff
@@ -339,7 +339,15 @@ case "$command" in
     ld|ld.gold|ld.lld)
         mode=ld
         if [ -z "$SPACK_CC_RPATH_ARG" ]; then
-            comp="CXX"
+	        if [ ! -z "$SPACK_CXX_RPATH_ARG" ]; then
+                comp="CXX"
+	        elif [ ! -z "$SPACK_FC_RPATH_ARG" ]; then
+                comp="FC"
+            elif [ ! -z "$SPACK_F77_RPATH_ARG" ]; then
+                comp="F77"
```
climbfuji: Since `depends` only offers `Fortran` (and presumably then sets both `SPACK_FC_RPATH_ARG` and `SPACK_F77_RPATH_ARG`?), I wonder if these lines 346-347 can be removed.

becker33: Yes, that can be removed.

becker33:
```suggestion
	        if [ -n "$SPACK_CXX_RPATH_ARG" ]; then
                comp="CXX"
	        elif [ -n "$SPACK_FC_RPATH_ARG" ]; then
                comp="FC"
```

becker33 (changes requested):
We should probably use the `-n` operator instead of negating the `-z` operator.

becker33 (dismissed):

alalazo (dismissed):
I think we should do a full rebuild here, since the compiler wrapper is used everywhere. Also, I'm surprised this works given we didn't update the shasum. 🤔

becker33:
@alalazo are there any packages in CI that use the linker but don't depend on C? I imagine we would've hit this issue already if there were, and if there aren't then this isn't going to change anything in CI. We can do the full rebuild to confirm.

alalazo:
I think there might be packages that depend on C++ only and use the linker in CI, but that was covered.

climbfuji:
@alalazo @becker33 Is there anything I can do to advance this pull request?

becker33:
@climbfuji can you update the sha in the version directive for this package to reflect the new changes, and bump the version number to 1.1?

climbfuji:
> @climbfuji can you update the sha in the version directive for this package to reflect the new changes, and bump the version number to 1.1?

I am happy to do that, but slightly embarrassed that I don't know how to get the updated sha.

becker33:
`shasum -a 256 $(spack location -p compiler-wrapper)/cc.sh` should give you what you want.

climbfuji:
> `shasum -a 256 $(spack location -p compiler-wrapper)/cc.sh` should give you what you want.

Thanks very much. I hope I got it all right.

climbfuji:
@becker33 Anything else I can do to move this PR forward?

climbfuji:
@alalazo @becker33 Can you please advise what to do with this PR? There's now a merge conflict, because apparently v1.0 got updated/had the hash changed. I still think we need the changes here, unless you've made updates that solve the problem in other ways.

climbfuji:
@becker33 @alalazo This has gone stale. I think it's still relevant. If so, what do I need to do? Thanks.

becker33 (dismissed):

climbfuji:
I merged spack develop into my branch, fixed the merge conflict, and updated the shasum.

becker33 (approved):

becker33:
In the long run we need to make this file properly versioned in its own repo, but I don't think that's a change to enforce on this PR.

alalazo:
Now changes have to be made here https://github.com/spack/compiler-wrapper and be tested in a PR before being merged.

climbfuji:
> Fixes an issue when checking which compiler is configured when the compiler wrapper is invoked in `ld` mode. Without this bug fix, spack assumes that a package that doesn't depend on C must depend on CXX. But there are packages that depend on Fortran only.

I created https://github.com/spack/compiler-wrapper/pull/16. I would appreciate help with testing the changes. Thanks very much in advance.

climbfuji:
@becker33 Thanks for merging the new PR in the compiler-wrapper repo. I assume this can now be closed, correct?

climbfuji:
Please let me know if any follow-up changes are required now that https://github.com/spack/compiler-wrapper/pull/16 is merged. Thanks!

PR #2171, opened 2025-10-29
cups: added all missing versions from 2.4.5 to 2.4.14

fbarbari:


repos/spack_repo/builtin/packages/cups/package.py
```diff
@@ -21,8 +21,16 @@ class Cups(AutotoolsPackage):
 
     license("Apache-2.0", checked_by="wdconinc")
 
+    version("2.4.14", sha256="660288020dd6f79caf799811c4c1a3207a48689899ac2093959d70a3bdcb7699")
```
alalazo, line 24: Can't we add just the latest? Why all the old patch versions?

bernhardkaindl (changes requested):
See https://github.com/spack/spack-packages/pull/2171#discussion_r2491707981

PR #2174, opened 2025-10-29
cairo: swap default values of a few variants

alalazo:
Having a featureful cairo by default helps with concretizing dependent packages, since they usually ask to activate variants that would otherwise be penalized, see https://github.com/spack/spack/issues/51428

@michaelkuhn


michaelkuhn (approved):
This works around the problem, thanks!

PR #1855, opened 2025-10-06
Hypre CMake support + v3.0.0

victorapm:
CMake support in hypre has matured and achieved feature parity with its autotools build system. In this PR:
- Introduce `CMakePackage` implementation for hypre and make it the default for `hypre@3:`
- Add `v3.0.0` version for hypre
- Add `mixed` value for `precision` variant in hypre
- Add `pic` variant (passes `-fPIC` to CFLAGS/CXXFLAGS)
- Add `cxxstd` variant
- `umpire` becomes a dependency for `+cuda` and `+rocm` variants in hypre
- Fix dependency issues with MFEM + hypre v3.0.0
- Fix issue with `+complex %gcc@14:` Closes #1975 
- Add full hypre team to maintainers list: @rfalgout @oseikuffuor1 @liruipeng @waynemitchell

Future PRs:
- Packages using `hypre_cmake` should use `hypre`
- Remove old `hypre_cmake` package implementation

@acfisher @hughcars @markcmiller86

PR #2178, opened 2025-10-29
Fix hypre@2 with GPU when using autotools

Sbozzolo:
#1855 turned cxxstandard into a variant, so the syntax for accessing the value should have changed.

This was not caught because hypre@3 uses cmake and not autotools.


victorapm (approved):
Thanks @Sbozzolo !

jrood-nrel:
Would love to see this one merged so we don't have to put this in our custom local repo to have it fixed.

victorapm:
Hi Jon, hopefully this will be merged soon. Note this fix isn't required to build our newest release. Please, let me know if you have any problems using hypre v3

alalazo (approved):

PR #2181, opened 2025-10-30
PyConfigArgParse: Drop yanked version and add latest

kwryankrattiger:

### DIFF
--- a/repos/spack_repo/builtin/packages/py_configargparse/package.py
+++ b/repos/spack_repo/builtin/packages/py_configargparse/package.py
@@ -19,11 +19,11 @@ class PyConfigargparse(PythonPackage):
     features."""
 
     homepage = "https://github.com/bw2/ConfigArgParse"
-    url = "https://github.com/bw2/ConfigArgParse/archive/1.7.tar.gz"
+    url = "https://github.com/bw2/ConfigArgParse/archive/1.7.1.tar.gz"
 
     license("MIT")
 
-    version("1.7", sha256="4549d105790386d01f71beebc3aa457d4177315680b75415f05bc22e1e28183a")
+    version("1.7.1", sha256="b31bfcce183ef7226f00cd07a37cfd69fe654229fe462f9a1de6091f396fc0f9")
     version("1.5.7", sha256="2156f15ef4ccce4377427046789ce93e0b09ac425c7297f1c9572655bf11bdfe")
     version("1.5.5", sha256="5b8316f11985aa169e51126086d3d6d24d7ba976585266311491015ddffbd717")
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/py_configargparse/package.py
```diff
@@ -19,11 +19,11 @@ class PyConfigargparse(PythonPackage):
     features."""
 
     homepage = "https://github.com/bw2/ConfigArgParse"
-    url = "https://github.com/bw2/ConfigArgParse/archive/1.7.tar.gz"
+    url = "https://github.com/bw2/ConfigArgParse/archive/1.7.1.tar.gz"
 
     license("MIT")
 
-    version("1.7", sha256="4549d105790386d01f71beebc3aa457d4177315680b75415f05bc22e1e28183a")
+    version("1.7.1", sha256="b31bfcce183ef7226f00cd07a37cfd69fe654229fe462f9a1de6091f396fc0f9")
```
w8jcik, line 26: With package being already installed, I suspect that the Spack solver won't be pleased if the version entry disappears. I believe that removing versions without deprecating them first is against Spack policy.

Consider bringing back `1.7` marked as `deprecated=True`

```
version(
    "1.7",
    sha256="4549d105790386d01f71beebc3aa457d4177315680b75415f05bc22e1e28183a",
    deprecated=True,
)
```

w8jcik (changes requested):

repos/spack_repo/builtin/packages/py_configargparse/package.py
kwryankrattiger: The solver looks at installed/external versions as well as recipe versions. This is just saying that it is no longer buildable. Existing installs shouldn't have any issues and in environments requesting reuse, it will prefer to not build this package if it exists as another version.

Adding deprecated will weight the installed version with deprecated penalty and make it more likely to rebuild.

w8jcik: Ok, then you know Spack internals better than I do. I just thought it is a strict rule to not remove hashes without deprecation, whatever the reason.

It can be merged then.

w8jcik (approved):

w8jcik:
@alalazo I’m not sure there’s much point in assigning this PR to me — I’ve already approved the change and don’t have write permissions to merge it.

alalazo (approved):

alalazo:
>  I’ve already approved the change and don’t have write permissions to merge it.

:+1: I noticed the open discussion and was wondering if there was anything to be solved. Since it seems it'a all :heavy_check_mark: I'll merge. Thanks!

PR #2183, opened 2025-10-30
Add new versions for py-xxhash

kwryankrattiger:


RikkiButler20 (approved):
Verified versions & sha256.
Looks good to me

PR #2184, opened 2025-10-30
PyTorch Lightning: Add newer versions

kwryankrattiger:


adamjstewart (commented):
Also need to update dependencies. Note that py-lightning is basically identical to py-pytorch-lightning. I switched to the former and haven't been updating the latter, although neither has been officially deprecated.

kwryankrattiger:
@adamjstewart I fixed the deps. But version >=2.5 have changed the URL for the source so that is going to require additional effort and I currently don't need that version so I am punting in this PR.

I did leave the version constraint updates, so whoever does eventually need that version or later can handle the URL and the rest should be done.

repos/spack_repo/builtin/packages/py_pytorch_lightning/package.py
```diff
@@ -57,8 +64,13 @@ class PyPytorchLightning(PythonPackage):
     depends_on("py-setuptools", type=("build", "run"))
```
adamjstewart, line 64: build/run for lightning 2.3 and older, build-only for 2.4 and newer

repos/spack_repo/builtin/packages/py_pytorch_lightning/package.py
```diff
@@ -57,8 +64,13 @@ class PyPytorchLightning(PythonPackage):
...
 
     # requirements/pytorch/base.txt
-    depends_on("py-numpy@1.17.2:", when="@1.3:", type=("build", "run"))
+    depends_on("py-numpy@1.17.2:", when="@1.3:2.3", type=("build", "run"))
     depends_on("py-numpy@1.16.6:", when="@:1.2", type=("build", "run"))
+    depends_on("py-torch@2.1:", when="@2.4:", type=("build", "run"))
+    depends_on("py-torch@2.1:", when="@2.4:", type=("build", "run"))
```
adamjstewart, line 70: Duplicated

repos/spack_repo/builtin/packages/py_pytorch_lightning/package.py
```diff
@@ -67,23 +79,30 @@ class PyPytorchLightning(PythonPackage):
     depends_on("py-torch@1.4:", when="@:1.3", type=("build", "run"))
     depends_on("py-tqdm@4.57.0:", when="@1.6.3:", type=("build", "run"))
     depends_on("py-tqdm@4.41.0:", when="@:1.6.2", type=("build", "run"))
-    depends_on("py-pyyaml@5.4:", when="@1.6:", type=("build", "run"))
+    depends_on("py-pyyaml@5.4.1:", when="@1.6:", type=("build", "run"))
```
adamjstewart, line 82:
```suggestion
    depends_on("py-pyyaml@5.4.1:", when="@2.5.3:", type=("build", "run"))
    depends_on("py-pyyaml@5.4:", when="@1.6:", type=("build", "run"))
```

adamjstewart (commented):
Only checked 2.4 since that's the default. This package apparently doesn't have upper bounds on deps like lightning has. Otherwise, we could share the dep info like we do with py-pillow and py-pillow-simd.

PR #2191, opened 2025-10-30
sherpa: add v3.0.2

m-fila:


wdconinc (dismissed):
Looks good to me. Do we need to add a conflict with gcc-15 for older versions?

m-fila:
> Looks good to me. Do we need to add a conflict with gcc-15 for older versions?

Indeed, I missed it in the changelog. Added conflict

repos/spack_repo/builtin/packages/sherpa/package.py
```diff
@@ -37,6 +38,8 @@ class Sherpa(CMakePackage, AutotoolsPackage):
         conditional("cmake", when="@3:"), conditional("autotools", when="@:2"), default="cmake"
     )
 
+    conflicts("%gcc@15:", when="@:3.0.2")
```
wdconinc: Ranges inclusive in spack.
```suggestion
    conflicts("%gcc@15:", when="@:3.0.1")
```

wdconinc (approved):

PR #2200, opened 2025-10-30
py-numexpr: add versions

jrood-nrel:

### DIFF
--- a/repos/spack_repo/builtin/packages/py_numexpr/package.py
+++ b/repos/spack_repo/builtin/packages/py_numexpr/package.py
@@ -15,6 +15,12 @@ class PyNumexpr(PythonPackage):
 
     license("MIT", checked_by="lgarrison")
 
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/py_numexpr/package.py
```diff
@@ -34,6 +40,7 @@ class PyNumexpr(PythonPackage):
 
     depends_on("python@3.7:", when="@2.8.3:", type=("build", "run"))
     depends_on("python@3.9:", when="@2.8.7:", type=("build", "run"))
+    depends_on("python@3.10:", when="@2.11.0:", type=("build", "run"))
     depends_on("py-setuptools", type="build")
```
tldahlgren, line 44: There was a lower bound for `py-setuptools` introduced in `2.12.0` per https://github.com/pydata/numexpr/blob/v2.12.0/pyproject.toml.

tldahlgren (changes requested):
Confirmed the version sha256s and dependencies, though `py-setuptools` has a lower bound as of `2.12.0`.

PR #2210, opened 2025-10-30
MVAPICH: Update Checksums

nhanford:
@MatthewLieber @natshineman The checksums on the `mvapich` package appear outdated across several systems. This PR would update the checksums.

### DIFF
--- a/repos/spack_repo/builtin/packages/mvapich/package.py
+++ b/repos/spack_repo/builtin/packages/mvapich/package.py
@@ -29,7 +29,7 @@ class Mvapich(MpichEnvironmentModifications, AutotoolsPackage):
 
     # Prefer the latest stable release
     version("4.1", sha256="25a53d3725b669e2c648158fb7c9fc5b1388953f3a2f949748586c447d0e43ee")
-    version("4.0", sha256="ee076c4e672d18d6bf8dd2250e4a91fa96aac1db2c788e4572b5513d86936efb")
+    version("4.0", sha256="c532f7bdd5cca71f78c12e0885c492f6e276e283711806c84d0b0f80bb3e3b74")
     with default_args(deprecated=True):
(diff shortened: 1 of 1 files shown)


MatthewLieber:
Thanks for catching this it looks like 4.1 does need to be updated but on my end it looks like the sha for 4.0 is fine.  Where did you get the new 4.0 sha from?

nhanford:
> Thanks for catching this it looks like 4.1 does need to be updated but on my end it looks like the sha for 4.0 is fine. Where did you get the new 4.0 sha from?

We calculated it locally on a couple different machines downloading in a couple of places. Since you have the current one for 4.0 as correct, I'll revert that one.

nhanford:
@spackbot re-run pipeline

michaelkuhn:
Do we know *why* the checksums have changed? It might be harmless (maintainers updating the tarball) but could also indicate something problematic (the tarball suddenly containing malware). Someone should check with upstream.

j34ni:
Echoing the upstream concern: MVAPICH 4.1 tarball has mutated at least 3 times since GA (SHAs: 6c707f... → a36c45... → 25a53d...)

Conda-forge mvapich-feedstock hitting the same walls

+1 on pinging OSU for a repo/changelog

MatthewLieber:
Hi all this was an expected change.  There was a change that was requested to one of the ROMIO test that was causing a failure when compiled with GCC 13+.  This was just a test change not things with actual MVAPICH code.

natshineman (approved):
Apologies for the confusion, we had a missing CHANGELOG line (no code changes), and and then we had a selftest error in make check that we pulled from upstream MPICH to satisfy the OpenSUSE devs. Neither involved changes to the MVAPICH library. 

We are good to update the sha.

nhanford:
@MatthewLieber Please approve and merge when you get a chance.

MatthewLieber (approved):

michaelkuhn (approved):
LGTM after the explanations regarding the reasons for the checksum change

PR #2215, opened 2025-10-31
Glib: build fix

yizeyi18:
Added new package `glib-bootstrap` and `gir-scanner` to solve circular dependency between `glib@2.79:` and `gobject-introspection`.

The new packages are added according to suggestions in #381 and fixes the issue. To keep the environment clean, a static `g-ir-scanner` was boostrapped first, then used to build `glib` as a pure `build` dependent.

@spackbot fix style

### DIFF
--- a/repos/spack_repo/builtin/packages/gir_scanner/package.py
+++ b/repos/spack_repo/builtin/packages/gir_scanner/package.py
@@ -1,0 +1,92 @@
+# Copyright Spack Project Developers. See COPYRIGHT file for details.
+#
+# SPDX-License-Identifier: (Apache-2.0 OR MIT)
+from spack_repo.builtin.build_systems import meson
+from spack_repo.builtin.build_systems.meson import MesonPackage
+
+from spack.package import *
+
+
+class GirScanner(MesonPackage):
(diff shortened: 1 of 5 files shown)


yizeyi18:
@spackbot fix style

michaelkuhn (changes requested):
I've already done this in #1625, what are the differences here?

yizeyi18:
@michaelkuhn I'm sorry that I only searched the issues. This PR would be closed.

michaelkuhn:
No worries. Could you give #1625 a try to see if it works for you?

yizeyi18:
It seems good, I made this PR only because I cannot install `rocm-gdb` that depends on `glib`.

What I am afraid of is that, if there exists any package that depends on `glib` implicitly by depending on `g-i`, and `glib-bootstrap` lacks any necessary function, so that the package will not work. I'm not familiar with both `glib` itself and packages relying `glib`, so I choosed a 3-stage bootstrap here(`glib~introspection`->`g-ir-scanner`->`glib`); 1-stage is also good if no such implicit dependency exists.

PR #2220, opened 2025-10-31
abseil-cpp: add new versions, update default cxxstd

wdconinc:
This PR updates `abseil-cpp` to the versions up to 20250814.1 (with latest patch release for any version).
- Updates the `cxxstd=14` conditional since in newer versions 17 is required (which becomes the default for the most recent versions).
- Fixes the older `cxxstd=11` conditional, since the mere year is not a component of the version number so the conditional was essentially unreachable.

Specific intermediate versions are added since packages like `py-onnxruntime` are particular about which one they need, owing to their vendoring.

Test build (for the one I needed as a dependency):
```
-- linux-ubuntu25.10-skylake / %cxx=gcc@15.1.0 ------------------
abseil-cpp@20240116.3  abseil-cpp@20240722.1  abseil-cpp@20250127.1  abseil-cpp@20250512.1  abseil-cpp@20250814.1
==> 5 installed packages
```

### DIFF
--- a/repos/spack_repo/builtin/packages/abseil_cpp/package.py
+++ b/repos/spack_repo/builtin/packages/abseil_cpp/package.py
@@ -98,7 +98,12 @@ class AbseilCpp(CMakePackage):
 
     variant(
         "cxxstd",
-        values=(conditional("11", when="@:20220623"), conditional("14", when="@:20250127"), "17", "20"),
+        values=(
+            conditional("11", when="@:20220623"),
+            conditional("14", when="@:20250127"),
+            "17",
+            "20",
+        ),
         default="17",
         description="C++ standard used during compilation",
     )
--- a/repos/spack_repo/builtin/packages/abseil_cpp/package.py
+++ b/repos/spack_repo/builtin/packages/abseil_cpp/package.py
@@ -19,7 +19,22 @@ class AbseilCpp(CMakePackage):
     license("Apache-2.0", checked_by="wdconinc")
 
     version(
+        "20250814.1", sha256="1692f77d1739bacf3f94337188b78583cf09bab7e420d2dc6c5605a4f86785a1"
+    )
+    version(
+        "20250512.1", sha256="9b7a064305e9fd94d124ffa6cc358592eb42b5da588fb4e07d09254aa40086db"
+    )
+    version(
+        "20250127.1", sha256="b396401fd29e2e679cace77867481d388c807671dc2acc602a0259eeb79b7811"
+    )
+    version(
+        "20240722.1", sha256="40cee67604060a7c8794d931538cb55f4d444073e556980c88b6c49bb9b19bb7"
+    )
+    version(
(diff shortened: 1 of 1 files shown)


wdconinc:
RFR @jcftang

albestro:
**TL;DR**

- `protobuf` versions before v31 (excluded) should be constrained to use `abseil-cpp@20230802:20250127` (pre-C++17, which provide if_constexpr utilities), with something like `depends_on("abseil-cpp@20230802:20250127", when=":3.30")`
- in CI it fails `protobuf@3.28.2`, but IMHO it should be `protobuf@28`. I will open a separate PR/issue about that, and try to keep changes requested by this PR separated from that problem.

### The problem

I checked a couple of CI failures, e.g. https://gitlab.spack.io/spack/spack-packages/-/jobs/18650002#L262

And it seems that `protobuf@3.28.2` is looking for `absl::if_constexpr` target which is missing in the `abseil-cpp` version it is using.

### abseil-cpp

From a quick investigation in `abseil-cpp` repo, the target gets defined in `absl/utility/CMakeLists.txt`, but just in a range of versions.

In particular, these are the commits involved
- https://github.com/abseil/abseil-cpp/commit/56c5fc620161e12844b9b92d06f8954b29f40bf4 introduces it
- https://github.com/abseil/abseil-cpp/commit/14cfd97abe88d78406478e75f3d33479e57f78bb drops it

and respectively they apply to these version tags

|introduced|dropped|
|-|-|
|20230802<br/>20240116<br/>20240722<br/>20250127<br/>20250512<br/>20250814|<br/><br/><br/><br/>20250512<br/>20250814|

So the range is `@20230802:20250127`. From the [commit message](https://github.com/abseil/abseil-cpp/commit/56c5fc620161e12844b9b92d06f8954b29f40bf4), they were introduced for pre-C++17 releases. And it matches with the `cxxstd` conditions, actually.
So a constraint has to be added to `protobuf`.

### protobuf

With https://github.com/protocolbuffers/protobuf/pull/20413 the `absl::if_constexpr` dependency has been definitely dropped, and this has been merged for versions starting from 31.

It is also worth highlighting `protobuf` versioning https://protobuf.dev/support/version-support/

> Starting with the v21.x protoc release, we changed our versioning scheme to enable nimbler updates to language-specific parts of Protocol Buffers. In the new scheme, each language has its own major version that can be incremented independently of other languages. The minor and patch versions, however, remain coupled.

The version reported in spack as `3.28.2`, I'd say it is a `v28` actually. I'm going to open a PR to propose this change, but I'd try to keep it separate from this PR.

What has to be done in any case is updating the constraints for `protobuf`, with something like `depends_on("abseil-cpp@20230802:20250127", when=":3.30")`.

albestro:
There is an additional point that might be worth considering in this PR (but it can also be deferred for a next one), which is about enforcing the same C++ standard to dependents, as pointed out here https://github.com/abseil/abseil-cpp/commit/6ee80c1d4ec5eb89e0c4879160294079267bdb3d.

More details in this comment https://github.com/spack/spack-packages/pull/2307#discussion_r2500097189

PR #2222, opened 2025-10-31
Visit: Make blueprint features required together

kwryankrattiger:
Visit enables blueprints if both MFEM and Conduit are enabled, but then only checks for conduit in source when including blueprint headers.

Instead of patching to apply the checks consistently in the headers just force the variants both on until this is fixed upstream.

### DIFF
--- a/repos/spack_repo/builtin/packages/visit/package.py
+++ b/repos/spack_repo/builtin/packages/visit/package.py
@@ -187,6 +187,10 @@ class Visit(CMakePackage):
     depends_on("mfem+shared+exceptions+fms+conduit", when="+mfem")
     depends_on("libfms@0.2:", when="+mfem")
 
+    # The checks in the build system are not consistent
+    requires("+mfem", when="+conduit")
+    requires("+conduit", when="+mfem")
+
     with when("+adios2"):
         depends_on("adios2")
         # adios 2.8 removed adios2_taustubs (https://github.com/visit-dav/visit/issues/19209)


cyrush (approved):
thanks for this, we recommend both on

pauleonix:
Given that this issue on the VisIt side seems to be years old, do we know why it only appeared now? Both variants are on by default, so sth seems to have changed recently to cause `visit~mfem` to appear in CI.

zackgalbreath:
> Given that this issue on the VisIt side seems to be years old, do we know why it only appeared now? Both variants are on by default, so sth seems to have changed recently to cause `visit~mfem` to appear in CI.

I am not sure why, but here's a bit more data.

* For c1631ede, you can see `mfem` in the [concretization results](https://gitlab.spack.io/spack/spack-packages/-/jobs/18594438#L246) for the data-vis-sdk stack.
* But in 2c679cf4a, `mfem` [no longer appears](https://gitlab.spack.io/spack/spack-packages/-/jobs/18597795).

Unfortunately there's [a lot of changes](https://github.com/spack/spack-packages/compare/c1631ede71336b0c8f7ce6f861ed1095297e5bf7..2c679cf4a213ea16fbff8bde27088375a4f5111b) between these two commits. It could be caused by the new `depends_on()` in the `mfem` package, or by the spack version bump.

kwryankrattiger:
> thanks for this, we recommend both on

@cyrush Here is a PR to make it _possible_ to turn off mfem in blueprints. https://github.com/visit-dav/visit/pull/20693

cyrush:
The concretizer can be unpredictable, frustrating aspect of all of this

kwryankrattiger:
I am going to force this through since it fixes the issue with VisIt. Unrelated issues with windows runners not starting is blocking.

PR #2223, opened 2025-10-31
Vtk: make patches that require ipv4 fetches local

johnwparent:
Some SC networks are now ipv6 only, and these patches reference links that are ipv4 only, make the patches local so fetch source does not introduce errors.

cc @kwryankrattiger


johnwparent:
Visit failing, probably legitimately, looking into why the behavior is different.

kwryankrattiger:
I have a fix for visit in another branch

kwryankrattiger (approved):
LGTM, thanks!

PR #2236, opened 2025-11-02
py-memray: add v1.19.1

adamjstewart:
Fixes support for newer macOS versions

### DIFF
--- a/repos/spack_repo/builtin/packages/py_memray/package.py
+++ b/repos/spack_repo/builtin/packages/py_memray/package.py
@@ -15,6 +15,7 @@ class PyMemray(PythonPackage):
 
     license("Apache-2.0")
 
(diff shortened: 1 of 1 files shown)


manuelakuhn (dismissed):
In a minimal rocky8 container this fails because it cannot find make. For some strange reason adding `gmake` as a build dependency also does not help.
This is not a problem of the new version but also happens for the old ones.
On rocky 9 (were the minimal container comes with make) it builds fine. 
Since this PR does not break anything (just keeps it in the same state), I would be fine with it being merged.
I approved the PR so that you can decide if you want to debug it or just merge it.

adamjstewart:
I confirmed that make is indeed a dep. Can't easily debug if or why not Spack-installed make is used. Will leave that up to someone else who needs it.

manuelakuhn (approved):

PR #2237, opened 2025-11-02
py-pytest-memray: add new package

adamjstewart:
Tested and works perfectly on macOS


manuelakuhn (approved):
LGTM and buils successfully on rocky 9 with gcc@15

PR #2244, opened 2025-11-02
compiler dependency: octave_control new version 4.1.3 + new url

cessenat:
This new version compiles with recent compiler (gcc 13.3.0 from Ubuntu 24) and octave 9.4.0 at least.


alalazo (dismissed):

repos/spack_repo/builtin/packages/octave_control/package.py
```diff
@@ -13,10 +13,13 @@ class OctaveControl(OctavePackage, SourceforgePackage):
...
 
     homepage = "https://octave.sourceforge.io/control/"
-    sourceforge_mirror_path = "octave/control-3.2.0.tar.gz"
+    # sourceforge_mirror_path = "octave/control-3.2.0.tar.gz"
+    git = "https://github.com/gnu-octave/pkg-control/"
+    url = "https://github.com/gnu-octave/pkg-control/releases/download/control-4.1.3/control-4.1.3.tar.gz"
 
     license("GPL-3.0-or-later")
 
+    version("4.1.3", sha256="07ce19c121778333e409b4fd67d8946ff6fa9d0eb93b00312b331b86ea6954b4")
```
alalazo, line 19: Did you check the shasum of v3.2.0 with the new url?

cessenat: Oh, no ; indeed it fails : 3.2.0 cannot be found on the new url.
So I tried to add something like
```
        if version <= Version("3.2.0"):
            url = "https://sourceforge.net/projects/octave/files/Octave Forge Packages/Individual Package Releases/control-{0}.tar.gz"
            return url.format(version)
        else:
            url = "https://github.com/gnu-octave/pkg-control/releases/download/control-{0}/control-{1}.tar.gz"
            return url.format(version, version)
```
but it fails:
```
root@3ab6b6d3802c:/# spack checksum octave-control@3.2.0
==> Selected 0 of 10 versions. Filtered by @3.2.0

==> Enter number of versions to take, or use a command:
    [c]hecksum  [e]dit  [f]ilter  [a]sk each  [n]ew only  [r]estart  [q]uit
action> c
root@3ab6b6d3802c:/# 

```

cessenat: Well, despite spack checksum fails, as it does with so many other packages that have a `url_for_version`, the last commit is fine for spack install. So, I think you can take a final review @alalazo , thanks.

cessenat:
Hello @alalazo , your change request philosophy is applied to octave control as well, checksum is fine.

alalazo (approved):

PR #2251, opened 2025-11-03
add `gfx942` to `rocm-openmp-extras`

simonpintarelli:
The target for MI300 was missing in `rocm-openmp-extras`. 
Question for the maintainers @srekolam @renjithravindrankannath is there a reason why the `gfx_list` is hardcoded and not read from `amdgpu_target`?

I've tested compilation for `6.3.3`, not sure what the minimum supported version is for `gfx942`.


srekolam (approved):
Okay with the changes,  I have looked at older releases and tried to check from which release gfx942 can be added. Inside llvm-amdgpu,clang driver had the support for gfx942 in 6.1 release. I have not checked prior releases yet.

renjithravindrankannath (approved):

alalazo (approved):
Maintainers approved

PR #2257, opened 2025-11-03
libfabric: patch hardcoded fi version in cxi provider for 2.3.X

boeschf:
Fixes `CXIP_FI_VERSION` in cxi provider.


repos/spack_repo/builtin/packages/libfabric/package.py
```diff
@@ -114,6 +114,11 @@ class Libfabric(AutotoolsPackage, CudaPackage, ROCmPackage):
     variant("level_zero", default=False, description="Enable Level Zero support")
     variant("gdrcopy", default=False, when="@1.12: +cuda", description="Enable gdrcopy support")
 
+    # The CXI provider hardcodes CXIP_FI_VERSION to FI_VERSION(2, 2).
+    # Make it match the libfabric we're building so that 2.3.x
+    # builds pick up the correct FI version.
+    patch("cxi_fi_version_2.3.patch", when="@2.3.0:2.3.1")
```
msimberg: Perhaps you can refer to the upstream patch in https://github.com/ofiwg/libfabric/pull/11421 instead and update the commit to say you're backporting it from `main`? Something like:

```suggestion
    patch(
        "https://github.com/ofiwg/libfabric/commit/f565852cedc7b6fd3848ed2f11b1dd90ed37be05.patch?full_index=1",
        sha256="da2514252074c350fb5cbdb04f267cf227d0a575902fe6cad355afe1dc7c0102",
        when="@2.3",
    )
```
Note also `when="@2.3"`. This is so that the patch also applies to a potential 2.3.2 and newer patch releases.

This _could_ also have `when="@2.3 fabrics=cxi"`, but I think it's unnecessary? The patch doesn't hurt even if cxi isn't enabled. What do you think?

msimberg (changes requested):

msimberg (approved):
Thanks @boeschf, looks good!

PR #2033, opened 2025-10-21
julia: add v1.12.5

m-fila:
Requires:
- https://github.com/spack/spack-packages/pull/2029
- https://github.com/spack/spack-packages/pull/2028
- https://github.com/spack/spack-packages/pull/2027
- https://github.com/spack/spack-packages/pull/2026
-  https://github.com/spack/spack-packages/pull/2025
- https://github.com/spack/patches/pull/3
- https://github.com/spack/spack-packages/pull/2264

PR #2264, opened 2025-11-03
libuv-julia: add v1.48.1rc1

m-fila:
The libuv-julia repo doesn't tag all the releases. As suggested in https://github.com/spack/spack-packages/pull/2033#discussion_r2486293077 adding a v1.48.1rc1 with commit used by Julia v1.12

@haampie @giordano


giordano (approved):
Looks ok to me

alalazo (approved):

PR #2266, opened 2025-11-03
acts: add upper limit on podio

wdconinc:
This pull request adds a version constraint so that when building `acts` versions up to `44` with both `+edm4hep` and `+examples`, the package now depends on `podio` version `1.4` or earlier.

Ref: https://github.com/acts-project/acts/issues/4764


vvolkl (approved):

wdconinc:
@stephenswat Looks good to you too?

jmcarcell (approved):
I can confirm this is still an issue in v44

stephenswat (approved):
LGTM.

PR #2270, opened 2025-11-03
nalu-wind+kynema: define cmake variable from variant

mbkuhn:


mbkuhn:
@jrood-nrel

jrood-nrel (approved):

repos/spack_repo/builtin/packages/nalu_wind/package.py
```diff
@@ -74,6 +74,7 @@ class NaluWind(CMakePackage, CudaPackage, ROCmPackage):
     depends_on("mpi")
     depends_on("yaml-cpp@0.6.0:0.7.0")
     depends_on("kynema", when="+kynema")
+    depends_on("trilinos@master", when="+kynema")
```
psakievich, line 77:
```suggestion
    depends_on("trilinos@master:", when="+kynema")
```
Probably want to make this a range so `@develop` can be used too

psakievich: Actually I'd probably do it from the next possible trilinos release so we don't have to come back and change it

psakievich (commented):
LGTM. Small suggestion

repos/spack_repo/builtin/packages/nalu_wind/package.py
mbkuhn: How does that work? If the next trilinos release doesn't exist, will spack automatically pick master instead?

psakievich: Yes exactly that. It's a lower bound and doesn't have to exist.

jrood-nrel:
Trilinos 16.2 is out, so feels like we might as well try to switch to that tag instead of master now.

mbkuhn:
2 parts of this commit; one is out-of-date and the other will be addressed in a different PR

PR #2271, opened 2025-11-03
[TAU package] Enables the user to force either the new or legacy L0 profiler

jordialcaraz:
@spackbot fix style


jordialcaraz:
@spackbot fix style

wspear (commented):
Is it possible to detect when new vs legacy is required based on the L0 dependency being used, making the variants unnecessary? If not, what about making 'new' the default and only providing an additional variant to specify legacy L0?  It'd be nice if the variant description could include a very succinct explanation of why you would want to use the legacy version.

jordialcaraz:
By default, it automatically detects what it should use.
The flags are in the case you want to use the old profiler, or when building the package using a machine which doesn't have a GPU. If there is no GPU, the check fails and the old profiler is used, so this is a workaround.

jordialcaraz:
@spackbot fix style

jordialcaraz:
@spackbot fix style

jordialcaraz:
The "when" is missing the ":", after 2.35. Cannot change it at this moment.

eugeneswalker (dismissed):

eugeneswalker:
@spackbot run pipeline due to runner system failure unrelated to this PR in prior run

jordialcaraz:
Just as a comment, due to the removed variant, now by default it forces the use of the new profiler, this can be overwritten with `+force-legacy-l0`

eugeneswalker:
The TAU builds are failing everywhere with this PR:

Here are some example jobs failed:
https://gitlab.spack.io/spack/spack-packages/-/pipelines/1295377

jordialcaraz:
@spackbot fix style

eugeneswalker (dismissed):
Looks great to me! @wspear Any final thoughts? Should we merge now?

wspear:
@jordialcaraz correct me if I'm wrong, but it looks like:
```
        if spec.satisfies("@2.35:"):
            if "+force-legacy-l0" in spec:
                options.append("-force_legacy_l0")
            else:
                options.append("-force_new_l0")
```

Runs even if level zero isn't specified. Should the first if clause include  "+level_zero" in spec?  Otherwise we'll get -force_new_l0 in every configuration. Even if that flag does nothing when level_zero isn't specified that would be an unnecessary and easily avoided complication.

jordialcaraz:
Yes, you are right, let me move it.

jordialcaraz:
@spackbot fix style

wspear (approved):

eugeneswalker (approved):

eugeneswalker (approved):

PR #2274, opened 2025-11-03
of-precice: Added new version, conflicts

emwjacobson:
Added a new version, 1.3.1, which is made for precice 3. I also added a `depends_on` for `c`, as without it errors would arise during compilation complaining about the `SPACK_CC_*` environment variable.

Mentions of the maintainers: @MakisH @kjrstory

### DIFF
--- a/repos/spack_repo/builtin/packages/of_precice/package.py
+++ b/repos/spack_repo/builtin/packages/of_precice/package.py
@@ -38,8 +38,8 @@ class OfPrecice(Package):
     depends_on("yaml-cpp")
     depends_on("pkgconfig", type="build")
 
-    conflicts("precice@3:", when="@:1.2.3")
-    conflicts("precice@:2", when="@1.3.0:")
+    depends_on("precice@:2", when="@:1.2.3")
+    depends_on("precice@3:", when="@1.3.0:")
 
     # General patches
     common = ["change-userdir.sh", "spack-derived-Allwmake"]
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/of_precice/package.py
```diff
@@ -22,20 +22,25 @@ class OfPrecice(Package):
...
+    depends_on("c", type="build")  # generated
     depends_on("cxx", type="build")  # generated
 
     depends_on("openfoam+source")
     depends_on("precice")
     depends_on("yaml-cpp")
     depends_on("pkgconfig", type="build")
 
+    conflicts("precice@3:", when="@:1.2.3")
+    conflicts("precice@:2", when="@1.3.0:")
```
alalazo:
```suggestion
    depends_on("precice@:2", when="@:1.2.3")
    depends_on("precice@3:", when="@1.3.0:")
```

alalazo (commented):
Maybe use positive logic, and `depends_on` constraints?

repos/spack_repo/builtin/packages/of_precice/package.py
emwjacobson: Makes sense. Wasn't sure what the "standard" was for this. Thanks!

alalazo (approved):

MakisH:
Thanks for the clean update! That was fast, everything looks correct from the upstream side.

We could even drop the `yaml-cpp` dependency, which was only relevant before v1.0.

PR #2275, opened 2025-11-03
Add ittnotify variant for Intel ITTNotify support

wspear:
@jordialcaraz @eugeneswalker @nchaimov


eugeneswalker:
Do we need to restrict this variant to `@2.35:`?

What happens if the new configure arg is used in a version that doesn't support it?

nchaimov:
Yes, it needs to be version-restricted. If an unknown argument is passed to configure, it fails, for example:
```
ERROR: Command line switch `-nonexistent-option' not recognized
```

wspear:
@spackbot fix style

eugeneswalker (approved):

PR #2277, opened 2025-11-03
tutorial stack: use when_possible and bootstrap gcc12 (new container)

becker33:
Use new container that doesn't have gcc-12 installed

Use `when_possible` to bootstrap the compiler from the container in a multi-round solve as @haampie pointed out is possible.

### DIFF
--- a/.ci/gitlab/.gitlab-ci.yml
+++ b/.ci/gitlab/.gitlab-ci.yml
@@ -580,7 +580,7 @@ tools-sdk-build:
 
 tutorial-generate:
   extends: [ ".tutorial", ".generate-x86_64"]
-  image: ghcr.io/spack/tutorial-ubuntu-22.04:v2023-10-30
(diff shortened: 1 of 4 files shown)


kwryankrattiger:
Looks good to me. Will approve when image tags are changed

becker33:
I had to get more specific than ideal with some of the specs to keep them distinct while using the `when_possible` unification scheme... we'll see whether this works, but the runtime changes seem to be working.

becker33:
@kwryankrattiger @alecbcs this should be ready to go now, and I've tested the entire tutorial with the PR buildcache that it generated -- may require some light tweaking to the stacks tutorial but overall I'm happy with it and getting back to having automation is good.

zackgalbreath (approved):

psakievich (approved):

alecbcs (approved):
Looks good to me. Thanks for pushing on this @becker33!

PR #2288, opened 2025-11-04
libcxi: add new versions, remove version build requirement

jscook2345:
This adds several new versions from git tags and removed the `@main` version requirement from building.

### DIFF
--- a/repos/spack_repo/builtin/packages/libcxi/package.py
+++ b/repos/spack_repo/builtin/packages/libcxi/package.py
@@ -16,8 +16,10 @@ class Libcxi(AutotoolsPackage):
 
     license("LGPL-2.1-or-later or BSD-3-Clause")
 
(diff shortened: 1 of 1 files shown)


rbberger (changes requested):
You also need to remove this:

https://github.com/jscook2345/spack-packages/blob/1c533d6652bd220c1c7c3a6fce3559c4eac1dfec/repos/spack_repo/builtin/packages/libcxi/package.py#L62

jscook2345:
> You also need to remove this:
> 
> https://github.com/jscook2345/spack-packages/blob/1c533d6652bd220c1c7c3a6fce3559c4eac1dfec/repos/spack_repo/builtin/packages/libcxi/package.py#L62

Thanks for spotting this! Fixed, rebased, and pushed!

rbberger (approved):

PR #2291, opened 2025-11-05
acts: Add +gnn variant and add necessary dependencies

tmadlener:
This PR is currently meant as a basis for discussion on how to best achieve what we want in the end: being able to build the Gnn plugin of acts via spack.

Currently we add the `torch-scatter` package as a `CMakePackage` as the [python version of this package](https://github.com/spack/spack-packages/blob/develop/repos/spack_repo/builtin/packages/py_torch_scatter/package.py) does not build via cmake and hence **does not install the cmake config for downstream consumption** that we need in order to enable the Gnn plugin for acts.

I don't know what the policy in this case is as I don't see a realistic way of fixing the upstream package on a timescale that would be relevant for us.

This also currently holds all versions of `torch_scatter` but realistically not even the latest tagged version is enough to build against recent versions of `py-torch`, so we might want to pick a dedicated commit rather than just `@master`.

I have not yet tested the `+cuda` variant in combination with `+gnn`, but I can build against the Gnn plugin locally and also run it on CPU.

- [x] This builds on top of #1924 so that needs to be merged before this can be considered.


wdconinc:
We've been trying to avoid the duplicate versions of packages where one installs CMake and another installs python since they tend to go out of sync and are a pain to have at the same time. I'm going to tag @adamjstewart in case he sees an easy way to include the CMake install in `py-torch-scatter`.

adamjstewart:
I'm unfortunately quite inexperienced when it comes to CMake. Do what you need to do.

tmadlener:
I was hoping that the cmake build also installs the python sources if `WITH_PYTHON` is set to `ON`, but it looks like that doesn't help either. So I don't see an easy way (i.e. without touching the upstream package significantly) of getting this with one package at the moment.

repos/spack_repo/builtin/packages/torch_scatter/package.py
```diff
@@ -0,0 +1,46 @@
...
+    version("2.1.0", sha256="c587ddf7ffaf8e4ddc31d8016d7f445d5896ce491209574800166a9f85c29377")
+    version("2.0.9", sha256="f7f87395e3a87c1db5a48d675138a15f7e4f5d7fd1c5e5067ade9a01890d7f8e")
+    version("2.0.8", sha256="3f4fd8787006e55bfc38d716b484b27ba7d37291d80b69d94930676f6f990b51")
+    version("2.0.7", sha256="0d5568f907a3f948b3359cf741838bd34cd840cf31efd686b0c44f6671fc325e")
+    version("2.0.6", sha256="d95e552f69d1304b060fb72770d2ea543c3e6c2d41e69baba792117831379f8c")
+    version("2.0.5", sha256="e29b364beaa9c84a99e0e236be89ed19d4452d89010ff736184ddcce488b47f6")
+    version("2.0.4", sha256="4fdadd6587f16ef3ff63c52f313f0c9dd97d13ae6496867fe566c309a4ea4036")
+    version("2.0.3", sha256="ff2ca1468cb4e49b65bea8f889971f196f209231fbee0cc8bd1615ecb367400b")
+
+    variant("cuda", default=False, description="Build with CUDA support")
```
wdconinc: Inherit CudaPackage is preferred.

repos/spack_repo/builtin/packages/torch_scatter/package.py
```diff
@@ -0,0 +1,46 @@
...
+    depends_on("py-torch")
+    depends_on("py-torch +cuda", when="+cuda")
+
+    conflicts("py-torch@2.1:", when="@:2.1.2")
+
+    def cmake_args(self):
+        args = [
+            self.define_from_variant("WITH_CUDA", "cuda"),
+            self.define("CMAKE_CXX_STANDARD", "20"),
+            self.define("WITH_PYTHON", True),
```
wdconinc: Does your functionality need the python components? To keep the roles for torch-scatter and py-torch-scatter factorized, would it be acceptable to depend on py-torch-scatter when +python or something like that? I assume there would need to be some version propagation.

tmadlener: Good question. I don't think so, but I haven't checked actually.

Just for clarification on your suggestion. Are you suggesting to have one of these as a "main" package (e.g. `torch-scatter` and then add a variant (e.g. `python`) and if that is set the main package also depends on the other (`i.e. `py-torch-scatter` in this case)? Or would you go for a "dual build system" package that chooses how to install depending on variant values? At least in this case we would eventually need both versions in any case (for validation).

wdconinc: I don't think a dual build system between cmake and python is the right way here. They are too different.

tmadlener: (I was hoping you are going to say that, as I think that would have been quite cumbersome to implement ;) ).

I suppose we could do 
- add a `cpp` variant to `py-torch-scatter` and depend on `torch-scatter` if it's enabled
- vice versa for `torch-scatter`

And then somehow break the dependency cycle with a `conflicts`?

adamjstewart:
> So I don't see an easy way (i.e. without touching the upstream package significantly) of getting this with one package at the moment.

I would be okay with this if you need to. That package hasn't had a new release in a couple years. I don't know how many people other than you are actually using it.

tmadlener:
So, we played a bit with copilot yesterday and it looks like that was able to come up with a configuration that uses cmake for building via setuptools. We are still testing this to see if this truly works as expected in our case.

I can't judge how well received such a wide-reaching change would be in the original package (we will open a PR regardless once we are done testing). I suppose we could add it as a `patch` if it's not accepted upstream?

wdconinc:
@tmadlener #1924 was merged and resulted in minor conflicts here. Can you rebase/update?

tmadlener:
I have now done the following:
- Rebased after #1924 was merged
- Made `torch-scatter` inherit from `CudaPackage` to get the `cuda` variant
- Made `torch-scatter` have a `python` variant and depend on `py-torch-scatter` if it's set to `True`
- Removed all but the latest version from `torch-scatter` since we are in any case only interested in that (and `py-torch-scatter` doesn't define others either)

At least in our experiments the agent mode of copilot is not able to come up with a solution that uses cmake as a setuptools extension without intervention from our side. Hence, unless someone insists that we follow through on that with manual intervention to have a single package, we would prefer for this PR to be merged as is (plus whatever fixes are still necessary).

repos/spack_repo/builtin/packages/torch_scatter/package.py
```diff
@@ -0,0 +1,41 @@
+from spack_repo.builtin.build_systems.cmake import CMakePackage
```
adamjstewart, line 5: @alalazo I'm surprised this PR isn't failing the pre-checks, this file doesn't have a license.

repos/spack_repo/builtin/packages/acts/package.py
```diff
@@ -300,6 +300,12 @@ class Acts(CMakePackage, CudaPackage):
     variant("legacy", default=False, description="Build the Legacy package")
     variant("mlpack", default=False, description="Build MLpack plugin", when="@25:31")
     variant("onnx", default=False, description="Build ONNX plugin")
+    variant(
+        "torch",
+        default=True,
```
wdconinc: Do we want this to be default=True? It seems different from the way other plugins are treated (default false), and it pulls in a much larger dependency too.

```suggestion
        default=False,
```

repos/spack_repo/builtin/packages/acts/package.py
```diff
@@ -427,6 +433,8 @@ class Acts(CMakePackage, CudaPackage):
     depends_on("mlpack@3.1.1:", when="+mlpack")
     depends_on("nlohmann-json @3.9.1:", when="@0.14: +json")
     depends_on("nlohmann-json @3.10.5:", when="@37: +json")
+    depends_on("torch-scatter", when="+gnn")
+    depends_on("torch-scatter +cuda", when="+cuda")
```
wdconinc, line 437: We don't want to depend on torch-scatter when just cuda is set but gnn is not.
```suggestion
    depends_on("torch-scatter +cuda +gnn", when="+cuda +gnn")
```

wdconinc: Doesn't this need when +torch too then? Or should there be another relationship between the gnn and torch variants?

tmadlener: This is independent in principle. You can build `acts +gnn~torch` and you will have a `gnn` plugin that can e.g. run onnx inference, but not torch based inference. Regardless of the choice of inference engine, you will still need `torch-scatter` for parts of the graph building (that happens outside any model inference).

In practice `torch-scatter` will pull in `py-torch` in any case and make sure that all the dependencies are there, but `+torch` is still necessary for acts to enable the necessary cmake option for building the torch based model inference (and ensure `py-torch is also in the build environment).

wdconinc: It's more the other way round. Right now `~gnn+torch` is not really any different from `~gnn~torch`, so that seems to indicate that `torch` should only be defined when `+gnn`. Unless there is shortly going to be more `torch` functionality in ACTS that is independent of the GNN plugin.

repos/spack_repo/builtin/packages/torch_scatter/package.py
wdconinc:
```suggestion
# Copyright Spack Project Developers. See COPYRIGHT file for details.
#
# SPDX-License-Identifier: (Apache-2.0 OR MIT)

from spack_repo.builtin.build_systems.cmake import CMakePackage
```

repos/spack_repo/builtin/packages/acts/package.py
tmadlener: Ah, OK. Sorry, didn't get the question then the first time. Essentially, currently it should be like this, IIUC?

```python
variant("torch", when="+gnn")
```

wdconinc: Yes, that's what I meant. I wasn't clear in my first comment and had it inverted myself :-)

wdconinc (approved):
Looks good to me. CI failures look unrelated. I'd try a pipeline resubmit.

tmadlener:
@spackbot re-run pipeline

PR #2294, opened 2025-11-05
protobuf: fix versioning (i.e. 3.32.1 -> 32.1)

albestro:
From the official documentation https://protobuf.dev/support/version-support/

> Starting with the v21.x protoc release, we changed our versioning scheme [...]. In the new scheme, each language has its own major version that can be incremented independently of other languages. The minor and patch versions, however, remain coupled.

And it also reports some examples

> For example, a single release might include protoc version 24.0, Java runtime version 4.24.0 and C# runtime version 3.24.0.

> Prior to this versioning scheme, all languages had major version 3. The first instance of this new versioning scheme was the 4.21.0 version of the Python API, which followed the preceding version, 3.20.1. Other language APIs released at the same time were released as 3.21.0.

With this PR we are mainly aiming at fixing versioning of the spack package to be aligned with the actual one.


hyoklee (approved):

msimberg (dismissed):
Looks good to me. Giving maintainer some time to approve as well.

albestro:
I have an additional question for you @hyoklee about protobuf and its versioning.

From the official documentation https://protobuf.dev/support/version-support/

> Starting with the v21.x protoc release, we changed our versioning scheme [...]. In the new scheme, each language has its own major version that can be incremented independently of other languages. The minor and patch versions, however, remain coupled.

And it also reports some examples

> For example, a single release might include protoc version 24.0, Java runtime version 4.24.0 and C# runtime version 3.24.0.

> Prior to this versioning scheme, all languages had major version 3. The first instance of this new versioning scheme was the 4.21.0 version of the Python API, which followed the preceding version, 3.20.1. Other language APIs released at the same time were released as 3.21.0.

I'm wondering what we should use as version for this package, since AFAIK it includes both `protoc` and the `c`/`c++` extension. However, I'd still have some questions in this regards, e.g. given a project depending on protobuf if it requires a specific language version or a specific `protoc` version, and in the latter case how difficult it might become specifying constraints on minor versions to target protoc and not the language.

In any case, at the moment I'd say that things are mixed up. To me, looking at the package history, these
https://github.com/spack/spack-packages/blob/ab2071e8b228bdbd0e496c8e59253d6a24736ef7/repos/spack_repo/builtin/packages/protobuf/package.py#L20-L35

and these

https://github.com/spack/spack-packages/blob/00cfa33d83478c805d438e7d1d99526af5fd69e3/repos/spack_repo/builtin/packages/protobuf/package.py#L19-L20

it seems they belong to different versioning schemes.

**I'd say that we should at least change `32.1` and `33.0` to become `3.32.1` and `3.33.0` respectively.**

But about what's good or wrong, should be discussed by someone that knows better this ecosystem.

hyoklee:
> I'd say that we should at least change 32.1 and 33.0 to become 3.32.1 and 3.33.0 respectively.

@albestro , would you please change versions?

albestro:
@haampie is it ok to change version, i.e. 32.1 to 3.32.1 like we did? or we should duplicate it and mark it as deprecated?

In the while, I'll fix all dependencies...

albestro:
in the meanwhile I fix the `url_for_version`...

adamjstewart:
Only comment is on the version scheme.

This has always been confusing to me. For the longest time, protobuf would make 3 release tags for every version, e.g.:

* v25.8
* v4.25.8
* v3.25.8

or even more confusingly:

* v31.0
* v6.31.0
* v4.31.0

Google has finally stopped doing this and now only uses a single versioning scheme. We have no way to know when the major version should be incremented (3.25.8 vs. 4.25.8 vs. 6.31.0) so we would only be carrying over the previous major version.

I would personally prefer to use the version name that Google is using. We could also ask Google which version they prefer for C/C++ or see what other package managers are doing. Continuing the old scheme for consistency is probably not a good motivation.

albestro:
@adamjstewart 

To my little knowledge `protobuf` provides `protoc` and all languages runtime. My first thought is to switch versioning to `protoc` version (i.e. from version 21 included we should drop the "major", e.g. change 3.21.1 to 21.1).

To my eyes the current status of the spack package is wrong
- we are mixing versioning scheme (e.g. 3.21.0 and 32.1)
- we are neither following any language versioning scheme explicitly

It seems this package is somehow driven by github repo tags provided by the official repo. Unfortunately, those seems redundant but incomplete (IMHO), in the sense that they provide most (but not all) tags for all language runtimes.

Looking at the GitHub repo, the tags for `protoc` seems maintained correctly.

```
tag v21.0
Tagger: Deanna Garcia <deannagarcia@google.com>
Date:   Wed May 25 21:50:53 2022 +0000

Tag release v21.0

commit 7062d0a2d0075d5e7d5c294fd3984df67a976da3 (tag: v3.21.0, tag: v21.0)
```

```
tag v29.0
Tagger: Protobuf Team Bot <protobuf-team-bot@google.com>
Date:   Wed Nov 27 08:50:41 2024 -0800

Tag release v29.0

commit 2d4414f384dc499af113b5991ce3eaa9df6dd931 (tag: v5.29.0, tag: v3.29.0, tag: v29.0)
```

About the remaining tags, it seems they try to provide also the versioning for all language runtimes, but from a quick check it seems they are not complete (e.g. I'd have expected 4.29.0 too, see tables below) 

From https://protobuf.dev/support/version-support/#cpp
image

From https://protobuf.dev/support/version-support/#objc
image

From https://protobuf.dev/support/version-support/#java
image

IMHO the best versioning (from version 21 included) is to use the `protoc` versioning.

I don't think we should go down the way of maintaining all versions as duplicates for all the languages. Someone that requires a specific version of a language could target a specific "protoc version" for `protobuf`.

This looks like a breaking change, and we might have to discuss the best way forward (e.g. mark current "wrong" version entries as deprecated and duplicate them with the new "correct" versioning) before proceeding.

adamjstewart:
I trust your official documentation more than GitHub tags. Can you add comments to each package pointing to the official version scheme? Feel free to rename packages as needed as well.

hyoklee (approved):
Thank you, @albestro !

haampie:
Personally I don't think breaking changes to spack-packages aren't so bad if they fix a problem. In other package managers you also have version yanking sometimes.

adamjstewart:
We have an easy way to avoid breaking changes, which is to first deprecate them and remove them later. Might as well use that method.

albestro:
I added two separate commits, so that we are free to decide later
- bf11393 where all "wrongly mapped" versions are fixed
- cabecc6 where I add them back as deprecated

next step is going through all the dependents and see if any change is required.

albestro:
These are all the dependents

```
[daint][ialberto@daint-ln003 packages]$ spack dependents protobuf
arrow  chapel   fairroot       gnina             hepmc3     mgard      mosh    opencv    paraview    py-keras        py-or-tools  py-torch              rdc                        treelite
brpc   cntk     gem5           google-cloud-cpp  lbann      migraphx   nanopb  or-tools  protobuf-c  py-onnx         py-protobuf  py-vl-convert-python  root
caffe  express  geopm-service  grpc              libpulsar  mivisionx  onnx    orc       ps-lite     py-onnxruntime  py-tf-keras  qgis                  tensorflow-serving-client
```

but from a quick check, only these should require attention

```
gnina/package.py:    depends_on("protobuf@:3.21.12")  # TODO
lbann/package.py:    depends_on("protobuf@3.10.0:3.21.12")  # TODO
lbann/package.py:    depends_on("zlib-api", when="^protobuf@3.11.0:")  # TODO
mgard/package.py:    depends_on("protobuf@3.4:", when="@compat-2022-11-18:")  # TODO
mgard/package.py:    depends_on("protobuf@:3.28", when="@:1.5.2")  # TODO
mivisionx/package.py:    depends_on("protobuf@:3", type="build")  # TODO
onnx/package.py:        if self.spec.satisfies("@1.13:1.14 ^protobuf@3.22:"):  # TODO
onnx/package.py:        if self.spec.satisfies("@1.15: ^protobuf@3.22:"):  # TODO
paraview/package.py:    depends_on("protobuf@3.4:3.21", when="@5.8:%gcc")  # TODO
paraview/package.py:    depends_on("protobuf@3.4:3.21", when="@5.8:%clang")  # TODO
paraview/package.py:    depends_on("protobuf@3.4:3.21", when="@5.11:")  # TODO
paraview/package.py:    depends_on("protobuf@3.4:3.21", when="@master")  # TODO
protobuf_c/package.py:    depends_on("protobuf@:3.21.12")  # TODO
py_or_tools/package.py:    depends_on("protobuf@3.12.2")  # TODO
qgis/package.py:    depends_on("protobuf@:3.21", when="@:3.28")  # TODO
root/package.py:    depends_on("protobuf@3:", when="+tmva-sofie")  # TODO
```

adamjstewart:
I almost wonder if we should retroactively enforce this new version scheme even for 3.20 and earlier...

albestro:
> I almost wonder if we should retroactively enforce this new version scheme even for 3.20 and earlier...

Like if we always targeted `protoc` version, despite what google was doing at the time? why might you want to prefer it? a small cons is that we would not be aligned with google versioning and it might be confusing for users (I know...there seems to be already some confusion 😅).

adamjstewart:
The pro would be that the package has one consistent versioning scheme for all versions. 

Another way to get that would be to start deprecating older versions that aren't needed anymore. But a lot of packages rely on older versions I think.

albestro:
Looking at the other PR, it made me click about patches! I should look at versions used in patches as well and update them as needed. Going to do that soon.

albestro:
In 26dfdcc2ad03a0453a0aef9fc0ab784ec0e6dab6 I updated all self-references, grepping for `@3\.` and `@:3\.`, which lead me to fix `depends_on`, `pathces` and some conditional check on version.

Then, I reverted the change where wrong versions were added back as deprecated, and the rationale is that they would end up having problems/behave differently, since the aforementioned "self-references" would not apply correctly to them.

IMHO this closes the road for the choice of keeping the old versioning scheme as deprecated or not, and I think we should go for simply replacing old wrong versions with new ones (from 21:).

vicentebolea (approved):
ParaView changes.

afzpatel (approved):
mivisionx changes LGTM

albestro:
@spackbot run pipeline

adamjstewart (dismissed):
I don't see any issues with this, although perhaps we should deprecate the old versions before removing them. I would go even further and use this new version scheme even for older versions of protobuf. Or deprecate older versions of protobuf.

hyoklee (approved):

albestro:
As pointed out by @haampie https://github.com/spack/spack-packages/pull/2294#issuecomment-3516944645, I'd consider this breaking change a "good" one. It might have some negative impact somewhere, but I'd expect them to be more clear and maintenable than the alternatives.

> perhaps we should deprecate the old versions before removing them.

in case you might have missed this https://github.com/spack/spack-packages/pull/2294#issuecomment-3558868964 this is why IMHO the deprecation of wrong versions is not pursuable.

I agree that the deprecation intermediate stage should be in general respected, but in this case it would introduce problems, because we've fixed all references to comply with the new versioning scheme, and for deprecated wrong versions they don't apply anymore.

> I would go even further and use this new version scheme even for older versions of protobuf. Or deprecate older versions of protobuf.

I don't like very much this because it's misleading since we don't adhere to the real versioning. It would be "spack uses a different versioning for `protobuf`", and considering that `protobuf` versioning is already quite complex, it does not sound like a good idea.

In case, pursuing this would require additional work for all references to versions up to 21 excluded. An example is `py-protobuf`

https://github.com/spack/spack-packages/blob/6948d84e6792c97fc709b3e2a66b2393ab50d648/repos/spack_repo/builtin/packages/py_protobuf/package.py#L83-L84

kmorel (approved):
Looks OK w.r.t. the mgard package.

albestro:
@spackbot run pipeline

haampie (approved):
lgtm

adamjstewart (approved):

PR #2297, opened 2025-11-05
py-immutables: add v0.21

wdconinc:
This PR adds `py-immutables`, [v0.21](https://github.com/MagicStack/immutables/releases/tag/v0.21). Bugfix only. Needed for python@3.13.

### DIFF
--- a/repos/spack_repo/builtin/packages/py_immutables/package.py
+++ b/repos/spack_repo/builtin/packages/py_immutables/package.py
@@ -27,6 +27,7 @@ class PyImmutables(PythonPackage):
 
     depends_on("python@3.5:", type=("build", "run"))
     depends_on("python@3.6:", type=("build", "run"), when="@0.16:")
+    depends_on("python@:3.12", type=("build", "run"), when="@:0.20")
     depends_on("py-setuptools", type="build")
     # setuptools 68 is more strict about the format of pyproject.toml and fails to install older
     # versions of this package
--- a/repos/spack_repo/builtin/packages/py_immutables/package.py
+++ b/repos/spack_repo/builtin/packages/py_immutables/package.py
@@ -25,9 +25,15 @@ class PyImmutables(PythonPackage):
 
     depends_on("c", type="build")  # generated
 
-    depends_on("python@3.5:", type=("build", "run"))
-    depends_on("python@3.6:", type=("build", "run"), when="@0.16:")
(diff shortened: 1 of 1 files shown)


adamjstewart:
How tightly coupled is this to CPython? Usually when there are version-specific wheels on PyPI, I lock it to only support versions for which there are wheels (see numpy, etc.). I'm hoping to get Python 3.14 support for most things soon.

wdconinc:
> How tightly coupled is this to CPython? Usually when there are version-specific wheels on PyPI, I lock it to only support versions for which there are wheels (see numpy, etc.). I'm hoping to get Python 3.14 support for most things soon.

Would that change anything here? There are wheels for all supported python versions, right? 3.8 to 3.13?

I can add a section like in py-numpy, if you think it is a benefit.

repos/spack_repo/builtin/packages/py_immutables/package.py
```diff
@@ -26,6 +27,7 @@ class PyImmutables(PythonPackage):
 
     depends_on("python@3.5:", type=("build", "run"))
     depends_on("python@3.6:", type=("build", "run"), when="@0.16:")
+    depends_on("python@:3.12", type=("build", "run"), when="@:0.20")
```
wdconinc: I think I see your concern, ~~which is that there is no 3.13 wheel~~. So we'd get something like
```suggestion
    # Based on PyPI wheel availability
    with default_args(type=("build", "link", "run")):
        depends_on("python@3.8:3.13", when="@0.21")
        depends_on("python@3.8:3.12", when="@0.20")
        depends_on("python@3.6:3.11", when="@0.19")
        depends_on("python@3.6:3.10", when="@0.17:0.18")
        depends_on("python@3.6:3.9", when="@0.16")
        depends_on("python@3.5:3.9", when="@0.15")
        depends_on("python@3.5:3.8", when="@0.14")
```
~~But 0.21 was released specifically for 3.13 support, see https://github.com/MagicStack/immutables/pull/118.~~

adamjstewart: 0.21 has 3.13 wheels, but not 3.14 wheels. But yes, that is exactly the kind of structure I'm proposing.

adamjstewart (approved):
Perfect, thanks! This will result in the oldest couple versions becoming deprecated and removed eventually which will be good.

PR #2298, opened 2025-11-05
sst: update core, elements, macro to 15.1.0

jmlapre:
Update sst-core, sst-elements, and sst-macro to 15.1.0.


berquist (approved):
Each of core, elements, and macro have the same SHA256 as the respective release archives and they build with default options under Rocky 8 with GCC 8.5.0.

RikkiButler20 (approved):
Looks good to me

PR #2305, opened 2025-11-06
py-moo: add v0.6.1.5

alalazo:

### DIFF
--- a/repos/spack_repo/builtin/packages/py_pymoo/package.py
+++ b/repos/spack_repo/builtin/packages/py_pymoo/package.py
@@ -1,18 +1,13 @@
 # Copyright Spack Project Developers. See COPYRIGHT file for details.
 #
 # SPDX-License-Identifier: (Apache-2.0 OR MIT)
-
 from spack_repo.builtin.build_systems.python import PythonPackage
 
 from spack.package import *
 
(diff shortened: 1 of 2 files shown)


repos/spack_repo/builtin/packages/py_pymoo/package.py
```diff
@@ -1,39 +1,45 @@
...
+    with default_args(deprecated=True):
+        version("0.5.0", sha256="2fbca1716f6b45e430197ce4ce2210070fd3b6b9ec6b17bb25d98486115272c2")
+        version("0.4.2", sha256="6ec382a7d29c8775088eec7f245a30fd384b42c40f230018dea0e3bcd9aabdf1")
 
-    depends_on("cxx", type="build")  # generated
+    depends_on("c", type="build")
+    depends_on("cxx", type="build")
+
+    with when("@0.6.1.5"):
+        depends_on("python@3.9:")
```
becker33: It doesn't look like this is a link dependency from the repo, and below it's listed as a build/run dep. It looks like this constraint is redudant given line 44 as well.

becker33 (changes requested):

becker33 (approved):

PR #2308, opened 2025-11-06
bump sphexa version to 0.95

sekelle:
Bump dependencies: CUDA->12, GCC->12
Add C,CXX dependency
Add H5Hut dependency

### DIFF
--- a/repos/spack_repo/builtin/packages/sphexa/package.py
+++ b/repos/spack_repo/builtin/packages/sphexa/package.py
@@ -21,11 +21,7 @@ class Sphexa(CMakePackage, CudaPackage, ROCmPackage):
     license("MIT")
 
     version("0.95", sha256="1007ffa97eb2085d50173676ec5e6387d1da7a8b78f204308223fbdbbecc60a1")
-    version(
-            "0.93.1",
-            sha256="95a93d0063ac8857b9be12c1aca24f5b2eef9dd4ffe8cf3f6b552a4dd54b940f",
-            deprecated=True
-            )
+    version("0.93.1", sha256="95a93d0063ac8857b9be12c1aca24f5b2eef9dd4ffe8cf3f6b552a4dd54b940f")
     version("develop", branch="develop")
 
     variant("hdf5", default=True, description="Enable support for HDF5 I/O")
@@ -44,12 +40,13 @@ class Sphexa(CMakePackage, CudaPackage, ROCmPackage):
     depends_on("cxx", type="build")
 
     depends_on("mpi")
-    depends_on("cuda@12:", when="+cuda")
+    depends_on("cuda@12:", when="@0.95: +cuda")
+    depends_on("cuda@11.2:", when="@0.93: +cuda")
     depends_on("hip", when="+rocm")
     depends_on("rocthrust", when="+rocm")
     depends_on("hipcub", when="+rocm")
     depends_on("hdf5 +mpi", when="+hdf5")
-    depends_on("h5hut@master", when="+hdf5")
+    depends_on("h5hut@master", when="@0.95: +hdf5")
 
     # Build MPI with GPU support when GPU aware MPI is requested.
     # For cray-mpich, the user is responsible to configure it for GPU aware MPI.
@@ -61,7 +58,8 @@ class Sphexa(CMakePackage, CudaPackage, ROCmPackage):
 
         depends_on("mpich +rocm", when="+rocm ^[virtuals=mpi] mpich")
 
-    conflicts("%gcc@:12")
+    conflicts("%gcc@:11", when="@0.95:")
+    conflicts("%gcc@:10", when="@:0.93.1")
     conflicts("cuda_arch=none", when="+cuda", msg="CUDA architecture is required")
     conflicts("amdgpu_target=none", when="+rocm", msg="HIP architecture is required")
     conflicts("+cuda", when="+rocm", msg="CUDA and HIP cannot both be enabled")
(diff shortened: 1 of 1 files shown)


albestro (changes requested):
[Versions should not be removed](https://spack.readthedocs.io/en/latest/packaging_guide_creation.html#deprecating-old-versions) (in case it can be marked as deprecated if there is a reason to do so). In any case, just add the new one.

I think that the change above might translate to adapt also other changes in this PR accordingly.

sekelle:
> I think that the change above might translate to adapt also other changes in this PR accordingly.

Do you mean adding the necessary logic to keep supporting the old version?

repos/spack_repo/builtin/packages/sphexa/package.py
```diff
@@ -45,7 +61,7 @@ class Sphexa(CMakePackage, CudaPackage, ROCmPackage):
 
         depends_on("mpich +rocm", when="+rocm ^[virtuals=mpi] mpich")
 
-    conflicts("%gcc@:10")
```
becker33, line 48: This could also be versioned `when="@:0.93"` instead of removing it.

repos/spack_repo/builtin/packages/sphexa/package.py
```diff
@@ -20,20 +20,36 @@ class Sphexa(CMakePackage, CudaPackage, ROCmPackage):
...
+    depends_on("cxx", type="build")
 
     depends_on("mpi")
-    depends_on("cuda@11.2:", when="+cuda")
+    depends_on("cuda@12:", when="+cuda")
     depends_on("hip", when="+rocm")
     depends_on("rocthrust", when="+rocm")
     depends_on("hipcub", when="+rocm")
     depends_on("hdf5 +mpi", when="+hdf5")
+    depends_on("h5hut@master", when="+hdf5")
```
becker33:
```suggestion
    depends_on("h5hut@master", when="@0.95: +hdf5")
```

repos/spack_repo/builtin/packages/sphexa/package.py
```diff
@@ -54,7 +70,7 @@ def cmake_args(self):
         spec = self.spec
 
         args = [
-            self.define_from_variant("SPH_EXA_WITH_H5PART", "hdf5"),
+            self.define_from_variant("SPH_EXA_WITH_H5HUT", "hdf5"),
```
becker33: This could use a version guard to make the old version still build properly.

becker33 (changes requested):
Is 0.93.1 deprecated in the upstream?

repos/spack_repo/builtin/packages/sphexa/package.py
sekelle: done

repos/spack_repo/builtin/packages/sphexa/package.py
sekelle: done

repos/spack_repo/builtin/packages/sphexa/package.py
sekelle: done

sekelle:
Added all changes that allow the older v0.93.1 to keep building.

repos/spack_repo/builtin/packages/sphexa/package.py
```diff
@@ -45,16 +58,21 @@ class Sphexa(CMakePackage, CudaPackage, ROCmPackage):
...
     conflicts("cuda_arch=none", when="+cuda", msg="CUDA architecture is required")
     conflicts("amdgpu_target=none", when="+rocm", msg="HIP architecture is required")
     conflicts("+cuda", when="+rocm", msg="CUDA and HIP cannot both be enabled")
 
     def cmake_args(self):
         spec = self.spec
 
+        hdf5lib = "H5HUT"
+        with when("@:0.95"):
+            hdf5lib = "H5PART"
```
albestro, line 64: please also double check the range: `@:0.95` means "up to 0.95 included" and 0.95 is a range version encompassing `0.95*`.

```suggestion
        hdf5lib = "H5HUT"
        if self.spec.satisfies("@:0.94"):
            hdf5lib = "H5PART"
```

repos/spack_repo/builtin/packages/sphexa/package.py
```diff
@@ -45,16 +58,21 @@ class Sphexa(CMakePackage, CudaPackage, ROCmPackage):
 
         depends_on("mpich +rocm", when="+rocm ^[virtuals=mpi] mpich")
 
-    conflicts("%gcc@:10")
+    conflicts("%gcc@:11", when="@0.95:")
+    conflicts("%gcc@:10", when="@:0.93.1")
```
albestro, line 54: double checking, this means that:
- new versions starting from 0.95 (included) do not build with `gcc@11` or older, i.e. require at least `gcc@12`
- old versions up to 0.93.1 do not build with `gcc@10` or older, i.e. require at least `gcc@11`

repos/spack_repo/builtin/packages/sphexa/package.py
```diff
@@ -20,20 +20,33 @@ class Sphexa(CMakePackage, CudaPackage, ROCmPackage):
...
+    )
+
     depends_on("cmake@3.22:", type="build")
+    depends_on("c", type="build")
+    depends_on("cxx", type="build")
 
     depends_on("mpi")
-    depends_on("cuda@11.2:", when="+cuda")
+    depends_on("cuda@12:", when="@0.95: +cuda")
+    depends_on("cuda@11.2:", when="@0.93: +cuda")
```
albestro, line 36: double check: does `0.93` work with `cuda@12`?

repos/spack_repo/builtin/packages/sphexa/package.py
```diff
@@ -20,20 +20,33 @@ class Sphexa(CMakePackage, CudaPackage, ROCmPackage):
...
     variant("hdf5", default=True, description="Enable support for HDF5 I/O")
     variant("gpu_aware_mpi", default=True, description="GPU aware MPI")
 
+    variant(
+        "cxxstd",
+        default="20",
+        multi=False,
+        description="Compile using the specified C++ standard",
+        values=("20", "23"),
+    )
```
albestro: the variant has been introduced, but if I'm not wrong you are not using it anywhere.

albestro (changes requested):

repos/spack_repo/builtin/packages/sphexa/package.py
sekelle: Yes, correct.

repos/spack_repo/builtin/packages/sphexa/package.py
sekelle: Yes, it does.

albestro (approved):

becker33 (approved):

PR #2309, opened 2025-11-06
verdict: add shared and pic variants to the package

albestro:
I was building VTK with externals, and `verdict` is one of them.

At some point the build stopped with the following message

```
/usr/bin/ld: /user-environment/linux-zen2/verdict-1.4.4-5kpp5mib5lw45takv2jtex2mzjuuhmnc/lib64/libverdict.a(V_HexMetric.cpp.o): relocation R_X86_64_32 against `.rodata' can not be used when making a shared object; recompile with -fPIC
/usr/bin/ld: failed to set dynamic section sizes: bad value
```

because apparently VTK was creating a shared library and including `verdict`...but `verdict` in spack by default is built as static library, which by default are not built (by CMake) with positional independent code (`-fPIC`).

I opted for adding `pic` and `shared` variants copying what has been already done in other packages.

### DIFF
--- a/repos/spack_repo/builtin/packages/verdict/package.py
+++ b/repos/spack_repo/builtin/packages/verdict/package.py
@@ -23,6 +23,9 @@ class Verdict(CMakePackage):
     version("1.4.2", sha256="225c8c5318f4b02e7215cefa61b5dc3f99e05147ad3fefe6ee5a3ee5b828964b")
     version("1.4.1", sha256="26fa583265cb2ced2e9b30ed26260f6c9f89c3296221d96ccd5e7bfeec219de7")
 
+    variant("pic", default=False, description="Build position independent code")
+    variant("shared", default=False, description="Build shared libraries")
+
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/verdict/package.py
```diff
@@ -23,6 +23,9 @@ class Verdict(CMakePackage):
     version("1.4.2", sha256="225c8c5318f4b02e7215cefa61b5dc3f99e05147ad3fefe6ee5a3ee5b828964b")
     version("1.4.1", sha256="26fa583265cb2ced2e9b30ed26260f6c9f89c3296221d96ccd5e7bfeec219de7")
 
+    variant("pic", default=False, description="Build position independent code")
+    variant("shared", default=False, description="Build shared libraries")
```
albestro, line 27: by default both set to false, so that behavior of the package is kept the same if they are not specified.

RMeli (approved):

msimberg (approved):

PR #2310, opened 2025-11-06
warp: new package

Markus92:
New package for [Warp](https://github.com/warpem/warp), another set of Cryo-EM tools.

It uses .NET; It builds a library `libSkiaSharp.so` and it is unclear how/where it gets it's dependencies (and which ones it actually are). I patch the `rpath` of it to use Spack stuff and that seemed to do the trick, but it could be a bit fragile on systems not RHEL 9.

### DIFF
--- a/repos/spack_repo/builtin/packages/warp/package.py
+++ b/repos/spack_repo/builtin/packages/warp/package.py
@@ -135,7 +135,7 @@ class Warp(CMakePackage, CudaPackage):
                 "/p:PublishSingleFile=true",
                 "/p:DebugType=None",
                 "-o",
-                f"{self.spec.prefix}/bin",
+                self.spec.prefix.bin,
                 tool,
             )
 
--- a/repos/spack_repo/builtin/packages/warp/package.py
+++ b/repos/spack_repo/builtin/packages/warp/package.py
@@ -1,0 +1,187 @@
+# Copyright Spack Project Developers. See COPYRIGHT file for details.
+#
+# SPDX-License-Identifier: (Apache-2.0 OR MIT)
+# from spack_repo.builtin.build_systems.package import Package
+import shlex
+import shutil
+
+from spack_repo.builtin.build_systems.cmake import CMakePackage
+from spack_repo.builtin.build_systems.cuda import CudaPackage
+
+from spack.package import *
+
+
+class Warp(CMakePackage, CudaPackage):
+    """Warp is a set of tools for cryo-EM and cryo-ET data processing
+    including, among other tools: Warp, M, WarpTools, MTools, MCore, and
+    Noise2Map."""
+
+    homepage = "https://warpem.github.io/"
+    url = "https://github.com/warpem/warp/archive/refs/tags/v2.0.0dev36.tar.gz"
+
+    maintainers("Markus92")
+
+    license("GPL-3", checked_by="Markus92")
+
+    version(
+        "2.0.0dev36", sha256="a50b019bbe2143d3709c4782db3e689455b94eb7162b5f8dd955996e9710a291"
+    )
+
+    # The source is riddled with calls to avx intrinsics, so make sure target supports it
+    conflicts("target=:k10")  # last AMD processor not to support avx
+    conflicts("target=:westmere")  # last Intel processor not to support avx
+    conflicts("target=:x86_64_v2")  # last generic architecture not to support avx
+
+    variant("cuda", default=True, description="Build with CUDA")
+    conflicts("~cuda", msg="Cuda is required.")
+
+    depends_on("c", type="build")
+    depends_on("cxx", type="build")
+    # Upstream builds with gcc@9.5.0; The d_ValueFill function (in NativeAcceleration)
+    # is templated but not explicitly instantiated. This leads to unknown symbols in gcc 10+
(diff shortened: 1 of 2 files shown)


Markus92:
@spackbot fix style

(could've sworn I ran `spack style --fix` beforehand myself!)

repos/spack_repo/builtin/packages/warp/package.py
```diff
@@ -0,0 +1,190 @@
...
+    license("GPL-3", checked_by="Markus92")
+
+    version(
+        "2.0.0dev36", sha256="a50b019bbe2143d3709c4782db3e689455b94eb7162b5f8dd955996e9710a291"
+    )
+
+    # The source is riddled with calls to avx intrinsics, so make sure target supports it
+    conflicts("target=:k10")  # last AMD processor not to support avx
+    conflicts("target=:westmere")  # last Intel processor not to support avx
+    conflicts("target=:x86_64_v2")  # last generic architecture not to support avx
```
becker33, line 33: I think you also need 

```
requires("target=x86_64:")
```
to avoid all of the non-x86 architectures.

repos/spack_repo/builtin/packages/warp/package.py
```diff
@@ -0,0 +1,190 @@
...
+    variant("cuda", default=True, description="Build with CUDA")
+    conflicts("~cuda", msg="Cuda is required.")
+
+    depends_on("c", type="build")
+    depends_on("cxx", type="build")
+    # Upstream builds with gcc@9.5.0; The d_ValueFill function (in NativeAcceleration)
+    # is templated but not explicitly instantiated. This leads to unknown symbols in gcc 10+
+    # Cause is the removal of '-frepo' in gcc 10+
+    # conflicts("%gcc@10:", msg="gcc 10+ no longer supports -frepo")
+    # conflicts("%llvm", msg="gcc < 10 is required")
```
becker33: Why are these commented out?

repos/spack_repo/builtin/packages/warp/package.py
```diff
@@ -0,0 +1,190 @@
...
+            + f'PROPERTIES CUDA_ARCHITECTURES "{cuda_targets}")',
+            "NativeAcceleration/CMakeLists.txt",
+        )
+
+    def cmake_args(self):
+        spec = self.spec
+        args = []
+        args.append(
+            "-DTorch_DIR={0}/lib/python{1}/site-packages/torch/share/cmake/Torch".format(
+                spec["py-torch"].prefix, spec["python"].version.up_to(2)
```
becker33:
```suggestion
            f"-DTorch_DIR={os.path.join(self['py-torch'].prefix, self['python'].purelib)}/torch/share/cmake/Torch"
```
I'm not sure whether you want `purelib` or `platlib`. That depends whether the directory is always `lib` or whether it's `lib64` on some platforms.

repos/spack_repo/builtin/packages/warp/package.py
```diff
@@ -0,0 +1,190 @@
...
+                "--configuration",
+                "Release",
+                "--framework",
+                "net8.0",
+                "--self-contained",
+                "true",
+                "/p:PublishSingleFile=true",
+                "/p:DebugType=None",
+                "-o",
+                f"{self.spec.prefix}/bin",
```
becker33:
```suggestion
                self.spec.prefix.bin,
```

`Spec.prefix` is a `Prefix` object, which has a special gettatr property for appending directories like this.

becker33 (changes requested):

repos/spack_repo/builtin/packages/warp/package.py
Markus92: Had lot of linker errors, thought it was due to compiler at first. Turns out it was some bad templating that gets fixed with the patch it pulls in.

I've removed these, it compiles/links fine with gcc 14 now, didn't test Clang but no reason to assume it won't work.

repos/spack_repo/builtin/packages/warp/package.py
Markus92: Added! Thanks for the suggestion

repos/spack_repo/builtin/packages/warp/package.py
Markus92: Just needs the CMake stuff. I took this from the `ams` package, but I see there's a nice property for it in `py-torch` so pulled that in (with some inspiration from `redis-ai`).

becker33:
Going to close an reopen to re-run the package audits, since I don't understand what the error is supposed to be here.

Markus92:
It seems to fail as `dotnet-core-sdk` isn't installable on macos/windows. Not necessarily a problem of this package, but one of its dependencies.

bernhardkaindl:
Rebased onto the current develop branch using using the Rebase button on the web.

bernhardkaindl:
@Markus92
> It seems to fail as dotnet-core-sdk isn't installable on macos/windows. Not necessarily a problem of this package, but one of its dependencies.

In this case you'd need to submit a PR that makes dotnet-core-sdk isn't installable on macos/windows or add conflicts for macos/windows (but that would be a workaround, I guess).

PR #2319, opened 2025-11-06
Add recent superlu versions

Sbozzolo:
Versions 9.2.0 and version 7.0.1


Sbozzolo:
Some builds are failing because CUDA 13 seems to require C++17 or newer

Sbozzolo:
Will be taken care in https://github.com/spack/spack-packages/pull/2519

PR #2321, opened 2025-11-06
py-onnxruntime: add v1.21.1, v1.22.2

wdconinc:
This PR adds `py-onnxruntime`, v1.21.1, v1.22.2. No major changes, just a newer CMake required.

Not adding v1.23 yet since requires newer abseil-cpp and onnx.


alecbcs (approved):

Merging without approval from @adamjstewart after > 2 weeks without a response.

PR #2322, opened 2025-11-07
netcdf-c: fix +mpi cmake builds

eap:
This change adds a parameter to the cmake builder to fix a bug where MPI builds were failing to build the fortran compatibility layer even when it was present. See https://github.com/Unidata/netcdf-c/issues/3199 for background on the issue.

While a durable fix is proposed in the issue linked above, all current version are affected by this bug and all netcdf-fortran binaries based on the current spack `build_system=cmake` binary will have difficult to trace segfaults if they are built without this patch adding an argument for `NC_EXTRA_DEPS`.


climbfuji (approved):

climbfuji (changes requested):
Dismissing my earlier approval because of failures in our CI builds. Will come back here after investigating those.

eap:
Unclear followups on this. IIRC we figured out another path.

PR #2336, opened 2025-11-07
py-rasterio: fix rpath

Chrismarsh:
When using the `rio` cli tools, the cython shared objects don't have a baked in rpath and thus cannot find gdal.

This PR:
- changes gdal to be a run/build/link depend,
- ensures the underlying cython a) uses the correct gdal and b) embeds the rpath.

**maintainer**: @adamjstewart

### DIFF
--- a/repos/spack_repo/builtin/packages/py_rasterio/package.py
+++ b/repos/spack_repo/builtin/packages/py_rasterio/package.py
@@ -60,6 +60,12 @@ class PyRasterio(PythonPackage):
         # https://github.com/rasterio/rasterio/issues/3024
         depends_on("py-numpy@:1", when="@:1.3.9")
 
+        # From README.rst and setup.py
+        depends_on("gdal@3.5:", when="@1.4:")
+        depends_on("gdal@3.1:", when="@1.3:")
+        depends_on("gdal@2.4:3.3", when="@1.2.7:1.2")
+        depends_on("gdal@2.3:3.2", when="@1.2.0:1.2.6")
+
     with default_args(type=("build", "run")):
         depends_on("py-affine")
         depends_on("py-attrs")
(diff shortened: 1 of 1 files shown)


Chrismarsh:
I don't love the `LDFLAGS` but I was unable to get this to trick down to cython otherwise. Is there a better way?

Chrismarsh:
@spackbot fix style

adamjstewart:
> changes gdal to be a run/build/link depend

I don't think the CLI is used at run-time, is it?

> ensures the underlying cython a) uses the correct gdal and b) embeds the rpath

This is clever, but I'm curious why our compiler wrappers aren't working. @haampie may want to take a look, he's hacked on Python + RPATH before.

Chrismarsh:
>I don't think the CLI is used at run-time, is it?

it isn't, just needs build, link. I added the new section. Thanks for the catch

>This is clever, but I'm curious why our compiler wrappers aren't working.

I wonder if it is similar to https://github.com/spack/spack/issues/43255? Nested build systems seems to require setting various `env.` in my limited experience. In this case, I've poked around a bit and I don't know enough how about how, exactly, the compiler wrappers work and how cython works. 

Are you ok with the current solution until the @haampie has time to look at the wrapper situation?

repos/spack_repo/builtin/packages/py_rasterio/package.py
```diff
@@ -60,6 +60,13 @@ class PyRasterio(PythonPackage):
         # https://github.com/rasterio/rasterio/issues/3024
         depends_on("py-numpy@:1", when="@:1.3.9")
 
+    with default_args(type=("build", "link")):
```
adamjstewart: This is the default, technically don't need to change anything

repos/spack_repo/builtin/packages/py_rasterio/package.py
```diff
@@ -75,13 +82,14 @@ class PyRasterio(PythonPackage):
...
     conflicts("^gdal@3.11:", when="@:1.4.3")
     # https://github.com/rasterio/rasterio/pull/3212
     conflicts("^gdal@3.10:", when="@:1.4.1")
+
+    # ensure cython absolutely gets the right gdal and embeds the correct rpaths
+    def setup_build_environment(self, env):
+        gdal = self.spec["gdal"]
+        # looks for this envar in setup.py
+        env.set("GDAL_CONFIG", join_path(gdal.prefix.bin, "gdal-config"))
+        env.prepend_path("LDFLAGS", f"-Wl,-rpath,{gdal.libs.directories[0]}")
```
adamjstewart, line 94: We used to use `self.compiler.cc_rpath_arg` for this, but NAG was the only compiler that was different and we removed support for NAG.

adamjstewart:
Yeah I'm fine with the RPATH workaround for now.

repos/spack_repo/builtin/packages/py_rasterio/package.py
Chrismarsh: I don't understand this comment. Is it a FYI for me, or something you want changed?

adamjstewart: More of a musing. Once upon a time, we should probably use `self.compiler.cc_rpath_arg` instead. Nowadays, I don't think it really matters. Maybe @alalazo has thoughts.

repos/spack_repo/builtin/packages/py_rasterio/package.py
```diff
@@ -60,6 +60,12 @@ class PyRasterio(PythonPackage):
         # https://github.com/rasterio/rasterio/issues/3024
         depends_on("py-numpy@:1", when="@:1.3.9")
 
+    # From README.rst and setup.py
```
adamjstewart: Can we undo moving this around? Just for `git blame` tracing.

Chrismarsh: yes. I think it's back to where it was

repos/spack_repo/builtin/packages/py_rasterio/package.py
Chrismarsh: Gotcha!

adamjstewart (approved):
Completely fine with this workaround for now. Hopefully someone can explain why the compiler wrappers aren't working and come up with a better solution.

PR #2340, opened 2025-11-08
ML CI: Python 3.13+

adamjstewart:
Next up: let's see if we can force all packages to build with Python 3.13+. This will prevent older versions of numpy and setuptools from being pulled in and may result in fewer duplicate package builds.


adamjstewart:
CI failure is unrelated, should be safe to merge.

alalazo (approved):

PR #2343, opened 2025-11-08
claude-code: new package

wdconinc:
This PR adds `claude-code`, an npm package for the Anthropic Claude CLI.


wdconinc:
@spackbot fix style

rbberger:
@wdconinc can you add v2.1.50 before I merge this?

repos/spack_repo/builtin/packages/claude_code/package.py
```diff
@@ -0,0 +1,32 @@
...
+    natural language commands.."""
+
+    homepage = "https://github.com/anthropics/claude-code"
+    url = "https://registry.npmjs.org/@anthropic-ai/claude-code/-/claude-code-2.0.36.tgz"
+    git = "https://github.com/anthropics/claude-code.git"
+
+    maintainers("wdconinc")
+    license("Antropic-Claude")
+
+    version("2.0.36", sha256="42095aacc8e39d8b7d5c0162fb44d873a1cc39430681269bac492c004cfd0e13")
```
wdconinc, line 24:
```suggestion
    version("2.1.50", sha256="4d7ab6b04c666ebf2c966a9d22e5b8f57706c0d5a5941523b9b02af9f84e45d5")
    version("2.0.36", sha256="42095aacc8e39d8b7d5c0162fb44d873a1cc39430681269bac492c004cfd0e13")
```

Not tested yet (including whether the `AUTHORIZED` hack below still works).

wdconinc:
> @wdconinc can you add v2.1.50 before I merge this?

Added. I wish I could get `list_url` to work so the updates could be a bit more automatic (https://registry.npmjs.org/@anthropic-ai/claude-code/ has the right file names but not as html links so the spider doesn't catch them).

rbberger (approved):

rbberger:
@wdconinc I merged this now as is. Note, Anthropic now says that installation via npm is deprecated. So maybe we should move to binary installations similar to CUDA.

PR #2344, opened 2025-11-08
rust: add v.1.88.0, v1.89.0, v1.90.0 and v.1.91.0, also add the requisite rust_boostrap

davekeeshan:
Add rust 1.88, 1.89, 1.90 and 1.91, also add rust_bootstrap which is required to get rust up and running


adamjstewart:
@kwryankrattiger

repos/spack_repo/builtin/packages/rust_bootstrap/package.py
```diff
@@ -35,6 +35,61 @@ class RustBootstrap(Package):
...
+            "darwin": {
+                "x86_64": "b329b458c8074023e5f6934bcd6c0bbef5075ac0090548c3d45a7de82e0c5b0c",
+                "aarch64": "ec42d93940933340ee55e67003699ebe264aa82d7cf0d5ae08100c06b1bfacfa",
+            },
+            "linux": {
+                "x86_64": "bad9a353330d9f409fe9db790da5701074112f804073506bb2808dd97b940b3c",
+                "aarch64": "29c5a608861cc9c06d3f86852a7d7b1a868de2d7ab90d4ff625aeebfb9383390",
+                "powerpc64le": "1a357b1e44f7ec7c2da62461a4180d8d8b599dd06045e7e2b83cd7f93972d6d7",
+            },
+        },
```
alecbcs: We should be able to drop `rust-bootstrap@1.91`, `rust-bootstrap@1.89`, and `rust-bootstrap@1.87` due to the overlapping dependency ranges on `rust` itself. (Keep the existing ranges in `rust` just drop the versions here to minimize the size of the Spack source cache mirror.)

alecbcs (changes requested):
One quick nitpick on rust-bootstrap, otherwise looks good to me!

repos/spack_repo/builtin/packages/rust_bootstrap/package.py
davekeeshan: @alecbcs i am not sure i understand what you are saying completely.  I think you are saying because we have rust-boostrap for 1.92.0 we should be able to use that boostrap for 1.91.0, 1.90.0 etc. 

I think this means that in the rust package you could change this line (as a minimum change):

    depends_on("rust-bootstrap@1.90:1.91", type="build", when="@1.91")

to this line:

    depends_on("rust-bootstrap@1.91:1.92", type="build", when="@1.91")

However when I try to compile `rust@1.91.0` it fails on build with this error:

    Unexpected rustc version: 1.92.0, we should use 1.90.x/1.91.0 to build source with 1.91.0

davekeeshan:
@bernhardkaindl why is this being reassigned to me.  There was an nitpick raised by the original assignee, i queried the nitpick for clarification and got no reply, 5 months ago.

bernhardkaindl:
Hi @davekeeshan, thanks for your reply!

> @bernhardkaindl why is this being reassigned to me. There was an nitpick raised by the original assignee, i queried the nitpick for clarification and got no reply, 5 months ago.

On a quick glance, it looked like there was a remaining real problem that prevents merge:

If I understand correctly, this happens only when you try to apply change:

```diff
-    depends_on("rust-bootstrap@1.90:1.91", type="build", when="@1.91")
+    depends_on("rust-bootstrap@1.91:1.92", type="build", when="@1.91")
```

> However when I try to compile rust@1.91.0 it fails on build with this error:
> `Unexpected rustc version: 1.92.0, we should use 1.90.x/1.91.0 to build source with 1.91.0`

I think you must have misunderstood @alecbcs or he wasn't aware of the strict version chain needed to bootstrap rust, but I think @alecbcs is aware of it very well, so there must have been a misunderstanding in this communication, I think.

The other misunderstand that I see is about the so-called "nitpick":

@alecbcs, wrote it like this would be optional for him:

> One quick nitpick on rust-bootstrap, otherwise looks good to me!

While the word "nitpick" suggests what he's raising might be optional, it apparently wasn't optional for him:

I think he wanted to convey that this should be a minor issue, but it wants to clarify and have done, the reason is:

- But, the ``quick nitpick`` followed by ``, otherwise looks good to me!`` means its is not actually a "optional" nitpick (I know...), is is this ``, otherwise looks good to me!`` that overrides it.

The real blocker which can't be worked around and really makes it clear that he did want to see this addressed before he resolves the `, otherwise` and approves is that at the same time, he marked his review as `changes request` (the red ball above this comment).

With this he blocked this PR totally for everyone to merge it until he approves it, that's the unmistakable sign that didn't see this discussion as optional, and you or we to get to him and explain what's going on to either fix this issue he think he saw of convince him otherwise to give an approval.

I'll re-request his review (the circle left of this avatar to signal that he has to take action on this PR again), otherwise I assume that a busy maintainer won't get alerted to the PR enough.

This means the burden is on whoever who wants the PR to proceed to provide sufficient explanation to continue the conversation, and I assigned it to you because your last message wasn't clear enough to me.

Now, with today's message you said that you're asking @alecbcs to respond, but this wasn't fully clear to me. Given that looked like a remaining problem you'll need to solve. But it's clear that the way you understood @alecbcs isn't a solution, so it must be different.

I think there may none (none, in no easy fix for @alecbcs's question or suggestion), but it has to be clearly said either way, I think.

Maybe, to get some more input, I might get into a review with Copilot to see if there it can help to understand and turn what @alecbcs might have meant into code. I apologize that this is a long-winded reply, but it looks like restarting to talk needs to be the 1st step to move this PR forward.

repos/spack_repo/builtin/packages/rust_bootstrap/package.py
bernhardkaindl: Here's the recap of the suggestion (part 1)

```suggestion
```

@alecbcs said:
> We should be able to drop rust-bootstrap@1.91, rust-bootstrap@1.89, and rust-bootstrap@1.87 due to the overlapping dependency ranges on rust itself. (Keep the existing ranges in rust just drop the versions here to minimize the size of the Spack source cache mirror.)

Alec’s suggestion is to keep the dependency ranges in `rust/package.py` as they are, but reduce the number of binary bootstrap versions added in `rust_bootstrap/package.py`.

The important part is that these ranges overlap:

```python
depends_on("rust-bootstrap@1.90:1.91", type="build", when="@1.91")
depends_on("rust-bootstrap@1.89:1.90", type="build", when="@1.90")
depends_on("rust-bootstrap@1.88:1.89", type="build", when="@1.89")
depends_on("rust-bootstrap@1.87:1.88", type="build", when="@1.88")
```

So we do not need every intermediate `rust-bootstrap` binary version. With the current ranges, these bootstrap versions are enough:

| Building | Can use |
|---|---|
| `rust@1.91.0` | `rust-bootstrap@1.90.0` |
| `rust@1.90.0` | `rust-bootstrap@1.90.0` |
| `rust@1.89.0` | `rust-bootstrap@1.88.0` |
| `rust@1.88.0` | `rust-bootstrap@1.88.0` |

I checked the `spack spec` bootstrap choices, and they are as expected.

That means the suggested change should be to remove only the redundant bootstrap entries for `1.91.0`, `1.89.0`, and `1.87.0`, while keeping `1.90.0` and `1.88.0`.

The full suggested diff is:

```diff
diff --git a/repos/spack_repo/builtin/packages/rust_bootstrap/package.py b/repos/spack_repo/builtin/packages/rust_bootstrap/package.py
--- a/repos/spack_repo/builtin/packages/rust_bootstrap/package.py
+++ b/repos/spack_repo/builtin/packages/rust_bootstrap/package.py
@@ -35,17 +35,6 @@ class RustBootstrap(Package):
                 "powerpc64le": "e2fe00a3c91f21c52947ebf96b4da016c9def5ccfedd1c335f30746db58bbf35",
             },
         },
-        "1.91.0": {
-            "darwin": {
-                "x86_64": "b329b458c8074023e5f6934bcd6c0bbef5075ac0090548c3d45a7de82e0c5b0c",
-                "aarch64": "ec42d93940933340ee55e67003699ebe264aa82d7cf0d5ae08100c06b1bfacfa",
-            },
-            "linux": {
-                "x86_64": "bad9a353330d9f409fe9db790da5701074112f804073506bb2808dd97b940b3c",
-                "aarch64": "29c5a608861cc9c06d3f86852a7d7b1a868de2d7ab90d4ff625aeebfb9383390",
-                "powerpc64le": "1a357b1e44f7ec7c2da62461a4180d8d8b599dd06045e7e2b83cd7f93972d6d7",
-            },
-        },
         "1.90.0": {
             "darwin": {
                 "x86_64": "3d1d24e1d4bedb421ca1a16060c21f4d803eaefba585c0b5b5d0b1e56692ef4b",
@@ -57,17 +46,6 @@ class RustBootstrap(Package):
                 "powerpc64le": "4061405099dc0aba379fe7b7a616d320272ef9325114dfa8f106c303f9b5695c",
             },
         },
-        "1.89.0": {
-            "darwin": {
-                "x86_64": "8590528cade978ecb5249184112887489c9d77ae846539e3ef4d04214a6d8663",
-                "aarch64": "87baeb57fb29339744ac5f99857f0077b12fa463217fc165dfd8f77412f38118",
-            },
-            "linux": {
-                "x86_64": "542f517d0624cbee516627221482b166bf0ffe5fd560ec32beb778c01f5c99b6",
-                "aarch64": "26d6de84ac59da702aa8c2f903e3c344e3259da02e02ce92ad1c735916b29a4a",
-                "powerpc64le": "80db8e203357a050780fb8a2cdc027b81d5ae1634fa999c3be69cf8a2e10bbf6",
-            },
-        },
         "1.88.0": {
             "darwin": {
                 "x86_64": "b36b0bfac17e0a1f6cc06b9fdc4e2131ad578b4122a67792236b58650ae4c5c8",
@@ -79,17 +57,6 @@ class RustBootstrap(Package):
                 "powerpc64le": "e1f16b2885237695f3cce7fc2f0128a938fc07462b076cb61bd2f06e5f8baf38",
             },
         },
-        "1.87.0": {
-            "darwin": {
-                "x86_64": "867a84a93c6ba0b468b78b52004a6307d6f5e2e5598e64f65726c6810b6f7c82",
-                "aarch64": "249496972e6f845f052036b9d7e73f816418412de2b266ec717b9050c1810dc3",
-            },
-            "linux": {
-                "x86_64": "1f6f18ce19387c42968a474cf175e67f99280614ded9c752d5d2e37af3204bcd",
-                "aarch64": "2c66e31d774a0dcd4422db74584ebc6362ff3ae90c452caff9d2fb912c821e8d",
-                "powerpc64le": "3f8f68d79460475396a52b5ae70eee73d6fb194fff1c4ff4286b64fbebee4429",
-            },
-        },
         "1.86.0": {
             "darwin": {
                 "x86_64": "bf8121850b2f6a46566f6c2bbe9fa889b915b1039febf36853ea9d9c4256c67d",
```

The reason changing `rust@1.91.0` to use `rust-bootstrap@1.91:1.92` caused a build failure is that it permits `rust-bootstrap@1.92.0`, but Rust 1.91 rejects that compiler version.

The dependency range for `rust@1.91.0` should stay as `rust-bootstrap@1.90:1.91`; we only need to avoid adding unnecessary binary bootstrap tarballs to the source cache mirror.

I started builds of

```py
spack install rust@1.91.0 (interrupted just to give CPU to the other three for now)
spack install rust@1.90.0
spack install rust@1.89.0
spack install rust@1.88.0
```

with the full diff given above, which is what @alecbcs suggested and all started and are looking good so far.

If it all works out well (which I expect), I think this can be pushed and @alecbcs can then approve to unblock the PR.

bernhardkaindl (changes requested):

bernhardkaindl:
@davekeeshan, Alex's suggested changes look good so far. Please apply them and push them to proceed.

It looks like your PR branch or company's git repo policy does not enable maintainers of PR repos to push to PRs, I can't do it myself.

I pushed the additional commit to my repo, so you can pull it with a single command and then push it:
```py
git pull https://github.com/bernhardkaindl/spack-packages rust-v1.91.0
git push <this company repo's PR branch>
```

bernhardkaindl (dismissed):
@davekeeshan, I'm approving to drop my changes-requested status, which I only requested to confirm that the changes requested by appear to look good:

It looks like your PR branch or company's git repo policy does not enable maintainers of PR repos to push to PRs, I can't do it myself.

I pushed the additional commit to my repo, so you can pull it with a single command and then push it:
```py
git pull https://github.com/bernhardkaindl/spack-packages rust-v1.91.0
git push <this company repo's PR branch>
```

Please apply them, then we can ping Alec for his approval.

davekeeshan:
> ```python
> git pull https://github.com/bernhardkaindl/spack-packages rust-v1.91.0
> ```

@bernhardkaindl OK done

bernhardkaindl (approved):
@alecbcs (https://isc-hpc.com/ 👍), your requested changes have been applied, LGTM.

@davekeeshan, Alec is at ISC26, he would be back tomorrow.

bernhardkaindl (approved):
@alecbcs, your requested changes have been applied and I tested the bootstrap of these versions on Linux, could you change your `review:changes_requested` state to an approval to unblock the PR?

alecbcs (approved):
Looks good to me. Thanks @davekeeshan and @bernhardkaindl

PR #2346, opened 2025-11-09
Refactor Silo package for dual Autotools/CMake and other minor updates

markcmiller86:
In preparation for Silo 4.12.0 release (by end of November, 2025 at the latest)

* Refactor Silo package to support dual Autotools/CMake build systems
* Add dual BSD/LLNL-Legacy licensing logic variants
* Add zfp variant
* Adjust some of the dependency logic to more faithfully reflect Silo's dependencies
* Remove mpi variant
  * Silo is a serial-only library meaning you get the same library whether or not you compile it with an "mpi" compiler or with `+mpi` variant.  There is (or should be) no MPI dependence. That said, Silo did have logic to, if it was compiled against a parallel HDF5 library, touch a couple of HDF5's parallel methods (one of which has since 1.10.4 been deprecated) so that it would behave well in that setting.

This should remain in DRAFT until Silo 4.12.0 release is finalized and its version and checksum can be added here.


markcmiller86:
I see removing Silo's `mpi` variant causes problems in a number of other packages that apparently attempt to use that variant.

I would prefer not to have to go touch all those other package files to update the Silo package. But, wholly removing the `mpi` variant here I believe is the right thing to do for _all_ versions of Silo not just the upcoming release. 

If older versions of Silo were ever compiled against a parallel HDF5, Silo did include logic (perhaps incorrectly) to touch a couple of HDF5's parallel features with the intention of helping Silo to behave "better" under certain conditions. Some of those HDF5 features have now even been deprecated in newer versions of HDF5. 

So, I think the right action is to remove the `mpi` variant entirely but keep the `remove-mpiposix` patch for those older versions of Silo that might need it.

markcmiller86:
I am getting this error in one of the CI checks...

```
PKG-DIRECTIVES: 1 issue found
    return _main(argv)
1. silo: dependency on qt@6.0: cannot be satisfied by known versions of qt
  File "/Users/runner/work/spack-packages/spack-packages/spack-core/lib/spack/spack/main.py", line 1076, in _main
    happening in /Users/runner/work/spack-packages/spack-packages/repos/spack_repo/builtin/packages/silo/package.py
    return finish_parse_and_run(parser, cmd_name, args, env_format_error)
    known versions of qt are 5.15.17, 5.15.16, 5.15.15, 5.15.14, 5.15.13, 5.15.12, 5.15.11, 5.15.10, 5.15.9, 5.15.8, 5.15.7, 5.15.6, 5.15.5, 5.15.4, 5.15.3, 5.15.2, 5.14.2, 5.12.10, 5.9.9, 5.6.3, 5.3.2, 5.2.1, 4.8.7, 4.8.6, 4.8.5, 3.3.8b
```

It appears Spack doesn't know about versions of Qt greater than 5.15.7. But, I don't want to have to go edit the Qt package here to deal with that. The fact is, the `+silex` variant of Silo version 4.12.0 requires Qt6. The fact that Spack doesn't know about Qt6 I don't think should prevent a package developer of asserting the requirement.

tgamblin:
> It appears Spack doesn't know about versions of Qt greater than 5.15.7

it's qt-base after version 5 -- qt refactored at version 6

markcmiller86:
> > It appears Spack doesn't know about versions of Qt greater than 5.15.7
> 
> it's qt-base after version 5 -- qt refactored at version 6

Ok, thanks. Wasn't understanding your remark here until I went looking in packages repo for examples and found things for qt6 related as `depends_on(qt-base +comp1 +comp2 +...)` where `compK` is a Qt 6 module.

markcmiller86:
Silo 4.12.0 is now released so this PR is ready for review.

markcmiller86:
I am a little concerned about the gitlab-ci tests here. They appear to be running `autoreconf` on version 4.11.1-bsd. That release artifact is not really designed to permit `autoreconf`'ing it. Two source directories are removed from the tarball but without changing any of the Automake files. This was the previous method of releasing BSD and non-BSD tarballs. The BSD tarballs are missing two src directories for hzip and fpzip compression libs.

I was not anticipating anyone (or any bot) re-autoconfing those release artifacts. And, I think that is a reasonable position to take.

AFAICT, *fixing* the gitlab-ci failures here requires me adjusting a prior release artifact so that it can pass the autoreconf test.

Please advise.

markcmiller86:
Ok, I think I actually found logic in the `package.py` file itself controlling this.

repos/spack_repo/builtin/packages/silo/package.py
```diff
@@ -242,14 +295,27 @@ def configure_args(self):
...
+
+
+class CMakeBuilder(cmake.CMakeBuilder):
+    def cmake_args(self):
+        args = [
+            self.define("SILO_ENABLE_JSON", False),
+            self.define("SILO_ENABLE_ASAN", False),
+            self.define("SILO_ENABLE_SILOCK", True),
+            self.define("SILO_ENABLE_BROWSER", True),
+            self.define("SILO_ENABLE_INSTALL_LITE_HEADERS", True),
```
markcmiller86, line 320: Ok,just realized, I am not handling `SILO_BUILD_FOR_BSD` here yet. That is an oversight. Also, `SILO_ENABLE_ASAN` is not a Silo CMake variable. I need to fix these

markcmiller86: @patrickb314 what is the preferred way of dealing with CMake variables here that already default to the desired settings in the actual package? Options are to not reference them at all here or to explicitly set them here as well even though they simply repeat the package's default settings for those variables.

patrickb314: I don't know best practice. My general approach is to include them if they're options that the spack package is going to set (e.g. by adding a +asan or +silock build option), but if they're not things visible to the spack user to not include them and just let them default.

patrickb314 (approved):
These all look good to me and seem to make sense from what I see in silo 4.12.0.

markcmiller86:
@patrickb314 sorry for false alarms...I discovered I was missing the compression lib options in the CMake builds.

patrickb314 (approved):
Only changes from previous approval are minor ci and style corrections.

patrickb314 (approved):

patrickb314 (approved):

markcmiller86:
Just wanted to give praise for all of the CI checks here :clap: :tada:. Although it was sometimes a bit difficult to paw through the CI logs to diagnose failures, I was able ultimately to find and fix many issues without much hand-holding from other busy developers.

It also turns out that ChatGPT has quite deep knowledge of Spack and how to structure a `package.py` file and that too was enormously helpful.

patrickb314 (approved):

markcmiller86:
@tldahlgren I noticed you had some recent activity on an MFEM PR here and I wanted to reach out and ask if your (or someone you might wanna recommend) could merge this PR for Silo?

markcmiller86:
@tldahlgren or @tgamblin I am reaching out in hopes that you or someone you might be able to nudge could please merge this PR?

@cyrush thinks our codes will benefit from having this in Spack sooner rather than later.

@patrickb314 has reviewed and approved it a few times at this point and I've updated and resolved conflicts since it became stable but neither @patrickb314 nor I have the authority to actually do the merge.

repos/spack_repo/builtin/packages/silo/package.py
```diff
@@ -67,34 +73,52 @@ class Silo(AutotoolsPackage):
...
+        description="Build Silex, a GUI alternative to text browser for viewing Silo files",
+    )
     variant("pic", default=True, description="Produce position-independent code (for shared libs)")
-    variant("mpi", default=True, description="Compile with MPI Compatibility")
+    variant("mpi", default=False, when="@:4.11", description="(deprecated)")
     variant("hdf5", default=True, description="Support HDF5 for database I/O")
-    variant("hzip", default=True, description="Enable hzip support")
-    variant("fpzip", default=True, description="Enable fpzip support")
+    variant("zfp", default=True, description="Enable zfp compression features")
+    variant("hzip", default=False, description="Enable hzip compression features (!BSD)")
```
tldahlgren, line 85: I have some reservations about changing the defaults of variants (e.g., `hzip` and `fpzip`) from `True` to `False` given that there are 15 built-in packages directly depending on `silo`:

- amp
- beatnik
- bookleaf_cpp
- cabana
- coevp
- conduit
- exampm
- lulesh
- macsio
- paraview
- parflow
- samrai
- simulationio
- visit
- visit_silo

And only two explicitly reference those variants (`beatnik` enables both in one of two entries while `parflow` disables both).

tldahlgren (commented):
See my previous comment regarding changing a couple of variant defaults.

repos/spack_repo/builtin/packages/silo/package.py
markcmiller86: @tldahlgren thanks for taking a closer look. That is a reasonable concern.

First, I am 99.9% certain nobody is _using_ those compression features in Silo. They may want the features enabled just in case but I would be surprised if anyone is actually using them. The use cases where fpzip and hzip can be applied to a Silo object are somewhat limited in scope and the resulting compression somewhat mediocre in terms of size or quality. It would have to be a significantly niche usage scenario where someone could use these features and be happy with the results. Not saying it isn't possible...just not very likely. Since you mention `beatnik` enables both explicitly, can we reach out to anyone knowledgeable there and get some details about their use case(s)? As an aside, the new zfp variant is more likely to be used successfully and is BSD licensed code, already built into Silo with proper name manglin to avoid any name conflicts (which was never true for fpzip or hzip here).

Next, neither the fpzip nor hzip source code, which come built into Silo here, are BSD licensed. Newer versions of those libs may be available as BSD licensed code but not the versions we baked into Silo. I think it is important that default behavior be to build what is truly open source licensed code only. I don't think these subtleties were obvious to developers who previously worked on Silo's Spack recipe. The new `license` variant added here is also currently set to a default value of `bsdonly` and that connects with what I see as preferred behavior.

Next, I could, if you think its best, condition the new default behavior on the version number of Silo. This would mean at least that anyone building older versions would get older default behavior but the newer version would have correct (at least as I see it) default behavior. OTOH, it means default behavior varies by Silo version number too. So, I would prefer not to do this.

To summarize, while I agree its worthwhile issue to bring up, I believe its the right thing to do and the likelihood of issues is low.

markcmiller86: So, I took a quick look at the `package.py` file for `beatnik` and from there to the github [project pages for it](https://github.com/CUP-ECS/beatnik/tree/main). 

In order for a Silo caller to *actually use* any of Silo's compression features, there would have to be a call to `DBSetCompression()` somewhere. I [don't find any](https://github.com/search?q=repo%3ACUP-ECS%2Fbeatnik+DBSetCompression&type=code) there. That said, they appear be a 2D, regular gridded sort of thing and that is indeed one of the niche areas one needs to be to take advantage of fpzip/hzip in Silo. So, I also looked for the word [`compression`](https://github.com/search?q=repo%3ACUP-ECS%2Fbeatnik%20compression&type=code) anywhere in their github repo and didn't get any hits. I think we are safe.

patrickb314:
As one of the Beatnik developers, I can affirm that Beatnik doesn’t actually depend on or use these features.

From: Tamara Dahlgren ***@***.***>
Date: Tuesday, January 20, 2026 at 5:31 PM
To: spack/spack-packages ***@***.***>
Cc: Patrick Bridges ***@***.***>, Mention ***@***.***>
Subject: Re: [spack/spack-packages] Refactor Silo package for dual Autotools/CMake and other minor updates (PR #2346)


@tldahlgren commented on this pull request.

See my previous comment regarding changing a couple of variant defaults.

In repos/spack_repo/builtin/packages/silo/package.py<https://github.com/spack/spack-packages/pull/2346#discussion_r2710516611>:

>      variant("hdf5", default=True, description="Support HDF5 for database I/O")
-    variant("hzip", default=True, description="Enable hzip support")
-    variant("fpzip", default=True, description="Enable fpzip support")
+    variant("zfp", default=True, description="Enable zfp compression features")
+    variant("hzip", default=False, description="Enable hzip compression features (!BSD)")


I have some reservations about changing the defaults of variants (e.g., hzip and fpzip) from True to False given that there are 15 built-in packages directly depending on silo:

  *   amp
  *   beatnik
  *   bookleaf_cpp
  *   cabana
  *   coevp
  *   conduit
  *   exampm
  *   lulesh
  *   macsio
  *   paraview
  *   parflow
  *   samrai
  *   simulationio
  *   visit
  *   visit_silo

And only two explicitly reference those variants (beatnik enables both in one of two entries while parflow disables both).

cyrush:
@tldahlgren Mark has a good read on usage and potential impact of changing the variants. I also vote getting focus access to newer silo and the cmake build system is the priority.

tldahlgren (approved):

markcmiller86:
@tldahlgren, @patrickb314 and @cyrush thanks so much!

PR #2351, opened 2025-11-10
detray: add v0.105.0, v0.105.1

stephenswat:
This commit adds versions 0.105.0 and 0.105.1 of the detray package.


wdconinc (approved):
Looks good to me.

PR #2358, opened 2025-11-10
fish: add v4.1.1+

adamjstewart:
https://github.com/fish-shell/fish-shell/releases/tag/4.2.0


alecbcs (approved):

Merging without confirmation from @funnell after > 2 weeks without a response.

PR #2360, opened 2025-11-10
Build ascent with Python v3.11 or older

zackgalbreath:
ascent currently attempts to use the `distutils` module to determine the location of `site-packages`. `distutils` was removed in Python v3.12.

### DIFF
--- a/repos/spack_repo/builtin/packages/ascent/package.py
+++ b/repos/spack_repo/builtin/packages/ascent/package.py
@@ -204,6 +204,10 @@ class Ascent(CMakePackage, CudaPackage, ROCmPackage):
     # causes duplicate state issues when running compiled python modules.
     with when("+python"):
         depends_on("python+shared", type=("build", "link", "run"))
+
+        # https://github.com/Alpine-DAV/ascent/issues/1628
+        depends_on("python@:3.11", type=("build", "link", "run"))
+
         extends("python")
         depends_on("py-numpy", type=("build", "link", "run"))
         depends_on("py-pip", type="build")


repos/spack_repo/builtin/packages/ascent/package.py
```diff
@@ -204,6 +204,10 @@ class Ascent(CMakePackage, CudaPackage, ROCmPackage):
     # causes duplicate state issues when running compiled python modules.
     with when("+python"):
         depends_on("python+shared", type=("build", "link", "run"))
+
+        # https://github.com/Alpine-DAV/ascent/issues/1628
+        depends_on("python@:3.11", when="@:0.9.5", type=("build", "link", "run"))
```
adamjstewart: I wouldn't add an upper bound, we don't know if/when this will be fixed.

adamjstewart (commented):
Note that Python 3.11 only has two years left before it reaches [EOL](https://devguide.python.org/versions/). Hopefully we'll see support for 3.14 soon.

adamjstewart (approved):

cyrush:
@zackgalbreath  I resolved distutils usage in conduit, but not yet ascent b/c of the embedded needs. 

As far as I know, there is no direct `get_python_lib` equivalent. 

`distutils` had special logic for some platforms (ubuntu), which is not 1:1 compared to using `sysconfig`

If anyone has an example of a robust way to find the python lib in python 3.12 that would be very helpful.

adamjstewart:
This is what we do in Spack: https://github.com/spack/spack-packages/blob/v2025.07.0/repos/spack_repo/builtin/packages/python/package.py#L1215

Not perfect, but pretty darn close. For more info, see:
```python
    # https://docs.python.org/3/library/sysconfig.html#installation-paths
    # https://discuss.python.org/t/understanding-site-packages-directories/12959
    # https://github.com/pypa/pip/blob/22.1/src/pip/_internal/locations/__init__.py
    # https://github.com/pypa/installer/pull/103

    # NOTE: XCode Python's sysconfing module was incorrectly patched, and hard-codes
    # everything to be installed in /Library/Python. Therefore, we need to use a
    # fallback in the following methods. For more information, see:
    # https://github.com/pypa/pip/blob/22.1/src/pip/_internal/locations/__init__.py#L486
```
I basically just copied this from pip/installer to trust them to know better than me.

cyrush:
@adamjstewart thanks for this example!

PR #2361, opened 2025-11-10
Fix OpenMP support in libCEED/MFEM

Sbozzolo:
As I was working on the Palace package.py, I noticed that our tests were failing when the OpenMP variant is
enabled (https://github.com/awslabs/palace/issues/557).

I tracked this down to the libCEED package not building with the `OPENMP=1` option with OpenMP is enabled.

I added a new `+openmp` variant to libCEED to handle this case.

I looked at the dependents and found that mfem should also be updated to propagate the variant.

In this, I also noticed that MFEM describes the `threadsafe` variant with
```
Enable thread safe features. Required for OpenMP. May cause minor performance issues.
```

But does not activate the variant when `+openmp`. I now turn this variant on when mfem is being compiled with OpenMP.

### DIFF
--- a/repos/spack_repo/builtin/packages/mfem/package.py
+++ b/repos/spack_repo/builtin/packages/mfem/package.py
@@ -251,6 +251,7 @@ class Mfem(Package, CudaPackage, ROCmPackage):
     conflicts("+shared", when="@:3.3.2")
     conflicts("~static~shared")
     conflicts("~threadsafe", when="@:3+openmp")
+    requires("+threadsafe", when="+openmp")
 
     conflicts("+cuda", when="@:3")
     conflicts("+rocm", when="@:4.1")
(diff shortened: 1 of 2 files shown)


jeremylt:
I cannot speak to the MFEM changes, but that's the right way to enable it for libCEED

v-dobrev (approved):
I did not actually try it but it looks good to me.

Thanks, @Sbozzolo!

jeremylt (approved):

Sbozzolo:
Is there anyone I should ping to have this merged?

repos/spack_repo/builtin/packages/mfem/package.py
```diff
@@ -251,6 +251,7 @@ class Mfem(Package, CudaPackage, ROCmPackage):
     conflicts("+shared", when="@:3.3.2")
     conflicts("~static~shared")
     conflicts("~threadsafe", when="@:3+openmp")
+    requires("+threadsafe", when="+openmp")
```
alalazo, line 254: You can remove in a following PR the conflict in the line above, which seems redundant with the `requires`.

alalazo (approved):

PR #2363, opened 2025-11-10
update mfem to handle hypre 2.29.0 +rocm

chapman39:
hypre version 2.29+rocm and beyond have additional dependencies that need to be linked.

https://github.com/spack/spack-packages/blob/develop/repos/spack_repo/builtin/packages/hypre/package.py#L215-L216


repos/spack_repo/builtin/packages/mfem/package.py
```diff
@@ -762,6 +762,11 @@ def find_optional_library(name, prefix):
                     hypre_rocm_libs += hypre["rocsparse"].libs
                 if "^rocrand" in hypre:
                     hypre_rocm_libs += hypre["rocrand"].libs
+                if hypre.version >= Version("2.29.0"):
```
alecbcs:
```suggestion
                if hypre.satisfies("@2.39.0:"):
```

We can call `spec.satisfies()` here to minimize creating additional version objects and make the condition match with common Spack spec syntax.

alecbcs (changes requested):
One quick nitpick. Otherwise this looks pretty good to me. Thanks @chapman39!

alecbcs (approved):
Looks good to me. Thanks @chapman39!

PR #2366, opened 2025-11-10
ecmwf_atlas: add new versions, and new dependency versions

fmahebert:
Add new versions of ECMWF's Atlas library, as well as new versions of Atlas's ECMWF dependencies:
- atlas 0.41.1 through 0.44.1
- ecbuild 3.8.5 through 3.12.0
- eckit 1.29.3 through 1.32.3
- fckit 0.14.1
- fiat 1.5.1 and 1.6.1
- ectrans 1.6.2 and 1.7.0

### DIFF
--- a/repos/spack_repo/builtin/packages/fiat/intel_warnings_v151.patch
+++ b/repos/spack_repo/builtin/packages/fiat/intel_warnings_v151.patch
@@ -1,0 +1,42 @@
+--- a/cmake/fiat_compiler_warnings.cmake
++++ b/cmake/fiat_compiler_warnings.cmake
+@@ -5,15 +5,17 @@
+   ecbuild_add_c_flags("-Wextra"                                NO_FAIL)
+   ecbuild_add_c_flags("-Wno-unused-parameter"                  NO_FAIL)
+   ecbuild_add_c_flags("-Wno-unused-variable"                   NO_FAIL)
+-  ecbuild_add_c_flags("-Wno-gnu-zero-variadic-macro-arguments" NO_FAIL)
++  if (CMAKE_C_COMPILER_ID MATCHES "Clang|GNU")
++    ecbuild_add_c_flags("-Wno-gnu-zero-variadic-macro-arguments" NO_FAIL)
++  endif()
+ endif()
+ 
+ # Always disable some warnings
+ ecbuild_add_c_flags("-Wno-deprecated-declarations" NO_FAIL)
+-if( CMAKE_C_COMPILER_ID MATCHES Intel )
+-  ecbuild_add_c_flags("-diag-disable=279")   # controlling expression is constant
+-  ecbuild_add_c_flags("-diag-disable=11076") # inline limits
+-endif()
++#if( CMAKE_C_COMPILER_ID MATCHES Intel )
++#  ecbuild_add_c_flags("-diag-disable=279")   # controlling expression is constant
++#  ecbuild_add_c_flags("-diag-disable=11076") # inline limits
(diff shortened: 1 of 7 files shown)


repos/spack_repo/builtin/packages/fiat/intel_warnings_v151.patch
```diff
@@ -0,0 +1,42 @@
...
+ # Always disable some warnings
+ ecbuild_add_c_flags("-Wno-deprecated-declarations" NO_FAIL)
+-if( CMAKE_C_COMPILER_ID MATCHES Intel )
+-  ecbuild_add_c_flags("-diag-disable=279")   # controlling expression is constant
+-  ecbuild_add_c_flags("-diag-disable=11076") # inline limits
+-endif()
++#if( CMAKE_C_COMPILER_ID MATCHES Intel )
++#  ecbuild_add_c_flags("-diag-disable=279")   # controlling expression is constant
++#  ecbuild_add_c_flags("-diag-disable=11076") # inline limits
++#endif()
```
srherbener, line 22: Is there a chance these might be restored (partially or fully) someday and you want to preserve the details? If you think these will never be restored, it would be good to remove them. Thanks!

repos/spack_repo/builtin/packages/fiat/intel_warnings_v151.patch
```diff
@@ -0,0 +1,42 @@
...
+ ### Compilation flags
+
+-if( CMAKE_Fortran_COMPILER_ID MATCHES "Intel" )
+-  ## To disable checking of argument correctness of dummy mpi symbols
+-  ecbuild_add_fortran_flags( -nowarn nointerfaces NO_FAIL )
+-endif()
++#if( CMAKE_Fortran_COMPILER_ID MATCHES "Intel" )
++#  ## To disable checking of argument correctness of dummy mpi symbols
++#  ecbuild_add_fortran_flags( -nowarn nointerfaces NO_FAIL )
++#endif()
```
srherbener, line 39: Is there a chance these might be restored (partially or fully) someday and you want to preserve the details? If you think these will never be restored, it would be good to remove them. Thanks!

srherbener (commented):
Thanks for the updates! Note that I did not test these changes.

repos/spack_repo/builtin/packages/fiat/intel_warnings_v151.patch
fmahebert: @srherbener I'm not following your question/suggestion. Are you suggesting that we remove the flags from the fiat build system? Or that we alter the patch to delete these lines of CMake instead of commenting them out?

climbfuji (approved):

repos/spack_repo/builtin/packages/fiat/intel_warnings_v151.patch
srherbener: Sorry about the slow response, and the confusion in my comment. I didn't notice before that this was a patch file, so I was thinking this was commented code in a configuration and was thinking it might be good to just delete it. There's no need to change this (since it was a misunderstanding on my part).

repos/spack_repo/builtin/packages/fiat/intel_warnings_v151.patch
srherbener: Same here - misunderstanding on my part, and no need to change anything.

srherbener (approved):

alalazo (approved):

PR #2367, opened 2025-11-10
Added hypre 3.0.0 dependency for Cabana 0.8.0 and newer

JStewart28:
Added hypre 3.0.0 dependency for Cabana 0.8.0 and newer


repos/spack_repo/builtin/packages/cabana/package.py
```diff
@@ -93,8 +93,9 @@ class Cabana(CMakePackage, CudaPackage, ROCmPackage):
     # Dependencies for subpackages
     depends_on("all-library", when="@0.5.0:+all")
     depends_on("arborx", when="+arborx")
-    depends_on("hypre-cmake@2.22.0:", when="@0.4.0:+hypre")
-    depends_on("hypre-cmake@2.22.1:", when="@0.5.0:+hypre")
+    depends_on("hypre-cmake@2.22.0:", when="@0.4.0 +hypre")
+    depends_on("hypre-cmake@2.22.1:", when="@0.5.0:0.7.0 +hypre")
+    depends_on("hypre@3.0.0:", when="@0.8.0:+hypre")
```
streeve, line 98: To clarify, this covers `master` and the planned update in Cabana? Seems to be the case if this is building

JStewart28: Yes. This assumes the next release of Cabana (0.8.0) will use hypre 3.0.0, and also `master` and `develop` will use hypre 3.0.0

streeve: Tested Cabana master with hypre 3.0 locally. It may be worth a comment here saying that the 3.0 requirement is only necessary to integrate with their more broad cmake support (not yet strictly required in Cabana)

streeve (approved):

repos/spack_repo/builtin/packages/cabana/package.py
JStewart28: @streeve the intention is to open a pull request into Cabana to require hypre 3.0 once this pull request is merged. We want to update it here first so once the change is made in Cabana, it won't break the spack package.

streeve:
I could be wrong, but I don't think you need to keep updating on develop (unless there are conflicts). It resets the approvals

streeve (approved):
May want to add a conflict for `@:0.7.0 +hypre +rocm` since they don't support that

JStewart28:
> May want to add a conflict for `@:0.7.0 +hypre +rocm` since they don't support that

Added

streeve (approved):

JStewart28:
> I could be wrong, but I don't think you need to keep updating on develop (unless there are conflicts). It resets the approvals

Spackbot just re-requested review after my push. But even with your approval I can't merge. It says someone with write access needs to.

streeve:
> Spackbot just re-requested review after my push. But even with your approval I can't merge. It says someone with write access needs to.

Right, but same after those merge commits above

JStewart28:
What else needs to be done to get this merged?

alalazo:
> What else needs to be done to get this merged?

You can post your PR in the #pull-request channel on our Slack

alalazo (approved):

PR #2368, opened 2025-11-11
chapel: add mason variant

jabraham17:
Adds a mason variant to the Chapel package to allow building the mason package manager for Chapel

In the future, it would be good to have `+mason` imply `comm=none`, but I think to do that its best to support multiple variants for `comm` so if a user specified `comm=ofi +mason` the concertized spec is `comm=ofi,none +mason`

### DIFF
--- a/repos/spack_repo/builtin/packages/chapel/package.py
+++ b/repos/spack_repo/builtin/packages/chapel/package.py
@@ -951,15 +951,6 @@ class Chapel(AutotoolsPackage, CudaPackage, ROCmPackage):
             # TODO: Need to update checkChplDoc to work in the spack testing environment
             pass
 
-    def test_mason(self):
-        """Run the mason test"""
-        if not self.spec.satisfies("+mason"):
-            print("Skipping mason test as mason variant is not set")
-            return
-        else:
-            # TODO: Need to create a mason test for the spack testing environment
-            pass
-
     @run_after("install")
     def copy_test_files(self):
         """Copy test files to the install directory"""
--- a/repos/spack_repo/builtin/packages/chapel/package.py
+++ b/repos/spack_repo/builtin/packages/chapel/package.py
@@ -50,7 +50,7 @@ class Chapel(AutotoolsPackage, CudaPackage, ROCmPackage):
     test_requires_compiler = True
 
     # TODO: Re-enable these once we add determine_version and determine_variants
-    # executables = ["^chpl$", "^chpldoc$"]
+    # executables = ["^chpl$", "^chpldoc$", "^mason$"]
 
     # A list of GitHub accounts to notify when the package is updated.
     # TODO: add chapel-project github account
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/chapel/package.py
```diff
@@ -587,7 +596,7 @@ class Chapel(AutotoolsPackage, CudaPackage, ROCmPackage):
     # but many of these are ALSO run-time dependencies of the executable
     # application built by that Chapel compiler from user-provided sources.
     with default_args(type=("build", "link", "run", "test")):
-        depends_on("cuda@11:12", when="+cuda")
+        depends_on("cuda@11:", when="+cuda")
```
bonachea: This patch hunk looks like an unintentional regression. 

IIUC current releases of Chapel do not support CUDA 13, so this hunk should be reverted (until a new Chapel release advertises CUDA 13 support, which will also require additional changes elsewhere).

repos/spack_repo/builtin/packages/chapel/package.py
```diff
@@ -938,6 +951,15 @@ def test_chpldoc(self):
...
             pass
 
+    def test_mason(self):
+        """Run the mason test"""
+        if not self.spec.satisfies("+mason"):
+            print("Skipping mason test as mason variant is not set")
+            return
+        else:
+            # TODO: Need to create a mason test for the spack testing environment
+            pass
```
bonachea: I realize this procedure is templated off the (currently useless) `test_chpldoc` above. That being said, I don't love the proliferation of dead/useless testing framework code. In particular, I believe as written both of these give the end user the misleading impression that something has been successfully tested for chpldoc/mason, when in fact nothing was actually tested. 

In chpldoc's case IIRC there was originally real test code there that was later disabled due to technical problems with the hopes of re-enabling it "soon", but since that has not happened we should probably consider removing that test. That's orthogonal to this PR, but I'd prefer we not clone that code until/unless there's some real meaningful mason test code available that can run in Spack.

bonachea (changes requested):
Overall looks reasonable, but I noted a few nitpicks I'd like addressed.

repos/spack_repo/builtin/packages/chapel/package.py
jabraham17: Yes this was unintentional, removed

repos/spack_repo/builtin/packages/chapel/package.py
jabraham17: Fine by me, I have removed this.

I also actually remove `mason` from `test_version`, because `mason --version` does not report the version expected by spack. My own testing of this package did not hit that issue, but I noticed the potential issue while addressing this feedback

bonachea (approved):
LGTM, based on source inspection.

Thanks for the fixes!

alecbcs (approved):
Looks good to me. Thanks @jabraham17!

PR #2370, opened 2025-11-11
flecsi: loosen cxxstd for boost, kokkos and hpx

rbberger:
only prevents older standards than C++17

### DIFF
--- a/repos/spack_repo/builtin/packages/flecsi/package.py
+++ b/repos/spack_repo/builtin/packages/flecsi/package.py
@@ -67,7 +67,7 @@ class Flecsi(CMakePackage, CudaPackage, ROCmPackage):
     depends_on("hdf5+hl+mpi", when="+hdf5")
     depends_on("metis@5.1.0:", when="@:2.3.1")
     depends_on("parmetis@4.0.3:", when="@:2.3.1")
-    depends_on("boost@1.79.0: cxxstd=17 +program_options +stacktrace")
+    depends_on("boost@1.79.0: +program_options +stacktrace")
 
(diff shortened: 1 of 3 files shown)


repos/spack_repo/builtin/packages/flecsi/package.py
```diff
@@ -125,6 +125,11 @@ class Flecsi(CMakePackage, CudaPackage, ROCmPackage):
     conflicts("+hdf5", when="backend=hpx", msg="HPX backend doesn't support HDF5")
     conflicts("^hpx networking=none", when="backend=hpx")
 
+    for cxxstd in ("11", "14"):
+        conflicts(f"^kokkos cxxstd={cxxstd}")
```
opensdh: Can `conflicts("^kokkos cxxstd=11")` be a problem if (some version of) Kokkos doesn't have that variant at all, or is that considered irrelevant because it's `conflicts`?

rbberger: It could be a problem, yes. I just realized that kokkos has a hard conflict for its own `cxxstd=11` and `cxxstd=14` for any version newer or equal to `4.0.0`. So that one is actually redundant. @lucbv FYI, you can probably get rid of the variant value cxxstd=11 and remove the conflict altogether in the kokkos recipe.

rbberger:
@eugeneswalker needed to make a change for the CI to not do something different. More details in the commit message.

eugeneswalker (approved):

opensdh (approved):

PR #2377, opened 2025-11-11
goma: add version 7.10.1 and 7.11.0, sparse: fixes for gcc@15

wortiz:
Let me know if I need to break these into separate requests.

### DIFF
--- a/repos/spack_repo/builtin/packages/sparse/package.py
+++ b/repos/spack_repo/builtin/packages/sparse/package.py
@@ -33,10 +33,10 @@ class Sparse(MakefilePackage):
     def edit(self, spec, prefix):
         with working_dir("./src"):
             makefile = FileFilter("Makefile")
-            sparse_cflags = "CFLAGS = -O2 {0}".format(self.compiler.c99_flag)
+            cflags_override = f"CFLAGS = -O2 {self.compiler.c99_flag}"
             if "+pic" in self.spec:
-                sparse_cflags += " {0}".format(self.compiler.cc_pic_flag)
-            makefile.filter("CFLAGS = .*", sparse_cflags)
+                cflags_override += f" {self.compiler.cc_pic_flag}"
+            makefile.filter("CFLAGS = .*", cflags_override)
             makefile.filter("CC = .*", "CC = {0}".format(spack_cc))
             makefile.filter("LIBRARY = .*", "LIBRARY = ../lib/libsparse.a")
 
--- a/repos/spack_repo/builtin/packages/sparse/package.py
+++ b/repos/spack_repo/builtin/packages/sparse/package.py
@@ -22,6 +22,7 @@ class Sparse(MakefilePackage):
 
     variant("pic", default=True, description="Build with position independent code")
 
+    # Remove implicit usage of time() for newer compilers
     patch(
         "spTest_time_patch.patch",
         sha256="e10ee8e790bfbc198ba065bf3f9b4b4526b61997dfc238c922c8e5c429043657",
--- a/repos/spack_repo/builtin/packages/sparse/package.py
+++ b/repos/spack_repo/builtin/packages/sparse/package.py
@@ -33,10 +33,11 @@ class Sparse(MakefilePackage):
     def edit(self, spec, prefix):
         with working_dir("./src"):
             makefile = FileFilter("Makefile")
-            cflags_override = f"CFLAGS = -O2 {self.compiler.c99_flag}"
-            if "+pic" in self.spec:
-                cflags_override += f" {self.compiler.cc_pic_flag}"
-            makefile.filter("CFLAGS = .*", cflags_override)
+            makefile.filter(
+                "CFLAGS = .*",
+                f"CFLAGS = -O2 {self.compiler.c99_flag} "
+                + (self.compiler.cc_pic_flag if spec.satisfies("+pic") else ""),                
+            )
             makefile.filter("CC = .*", "CC = {0}".format(spack_cc))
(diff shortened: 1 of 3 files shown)


repos/spack_repo/builtin/packages/sparse/package.py
```diff
@@ -22,17 +22,20 @@ class Sparse(MakefilePackage):
 
     variant("pic", default=True, description="Build with position independent code")
 
+    patch(
+        "spTest_time_patch.patch",
+        sha256="e10ee8e790bfbc198ba065bf3f9b4b4526b61997dfc238c922c8e5c429043657",
+    )
```
alalazo, line 29: Can you add a brief comment above the `patch` directive explaining:
2. Why it is needed
3. Where it was taken from

?

repos/spack_repo/builtin/packages/sparse/package.py
```diff
@@ -22,17 +22,20 @@ class Sparse(MakefilePackage):
...
         with working_dir("./src"):
             makefile = FileFilter("Makefile")
+            sparse_cflags = "CFLAGS = -O2 {0}".format(self.compiler.c99_flag)
             if "+pic" in self.spec:
-                makefile.filter(
-                    "CFLAGS = .*", "CFLAGS = -O2 {0}".format(self.compiler.cc_pic_flag)
-                )
-            else:
-                makefile.filter("CFLAGS = .*", "CFLAGS = -O2")
+                sparse_cflags += " {0}".format(self.compiler.cc_pic_flag)
```
bernhardkaindl: @wortiz As alalazo alluded to, the specific gcc@15 problems fixed by the patch should be mentioned at least in the PR description if not in the commit message (only update it now if you need to make changes to avoid unneeded expensive GitLab pipeline runs).

For example, it is curious why -O2 was added before cc_pic_flag previously and why it can be removed now (it's good to remove not needed flags, but it deserves a little note to give some context on what were potential reasons why it was there in the 1st place to judge if it can be removed in for all configurations and versions).

repos/spack_repo/builtin/packages/sparse/package.py
wortiz: Added. 

Regarding 3. I wrote it myself.

repos/spack_repo/builtin/packages/sparse/package.py
wortiz: -O2 is still there. I just simplified the usage of if pic check (mistakenly commented something incorrect before.)

bernhardkaindl: Ack, apparently, the code was and is still a bit confusing, otherwise I think also also alalazo may have approved it right away. I'll try to propose an amendment to make the updated patching of `f"CFLAGS = -O2 {extra_cflags}"` more obvious.

repos/spack_repo/builtin/packages/sparse/package.py
```diff
@@ -22,17 +22,21 @@ class Sparse(MakefilePackage):
...
+            makefile.filter("CFLAGS = .*", sparse_cflags)
```
bernhardkaindl:
```suggestion

            # Override CFLAGS to use -O2, C99 and add PIC flag if requested
            cflags_override = f"CFLAGS = -O2 {self.compiler.c99_flag}"
            if "+pic" in self.spec:
                cflags_override += f" {self.compiler.cc_pic_flag}"
            makefile.filter("CFLAGS = .*", cflags_override)

```

bernhardkaindl (commented):
Hi @wortiz JFYI, as commented before, in the discussion started by alalazo,

I think the simplification of the filter patch to override `CFLAGS =` can be made more legible:

- Spack now uses f-strings, which improves legibility
- Use something like `cflags_override` to make the intent clear.
- Including `sparse` in the flags is redundant and confusing as it is the only cflags variable patched in this recipe and this is the sparse recipe.
- Add an empty line before and after the functional block and add a comment.

@wortiz, unless you find something I overlooked something in this suggestion,
I plan to apply this suggestion soon before GitLab CI starts

repos/spack_repo/builtin/packages/sparse/package.py
wortiz: ok, added.

bernhardkaindl: Hi @wortiz, an incremental suggestion on top of the change to use f-strings:

This puts it even into one statement and makes it so clear that no comment is needed:
```suggestion
            makefile.filter(
                "CFLAGS = .*",
                f"CFLAGS = -O2 {self.compiler.c99_flag} "
                + (self.compiler.cc_pic_flag if spec.satisfies("+pic") else ""),
            )
```

Update: Changed `"+pic" in spec` to `spec.satisfies('+pic')` for the newer spack method.
Update2: Simplified to use a trailing space in the 1st part to simplfiy the 2nd part avoiding the need for a 2nd f-string using `(self.compiler.cc_pic_flag if spec.satisfies("+pic") else "")`

bernhardkaindl (commented):
Thanks, another incremental suggestion:

Inline the conditional PIC suffix into the `makefile.filter(...)` call, so the package no longer needs a temporary cflags_override variable and it is all internal to the filter call.

@wortiz, unless you find something I overlooked something in this suggestion,
I'd like have this applied or apply this suggestion before GitLab CI starts.

repos/spack_repo/builtin/packages/sparse/package.py
wortiz: should be updated.

bernhardkaindl (approved):

PR #2378, opened 2025-11-12
madgraph5amc: drop syscalc dependency

paulgessinger:
As discussed in https://github.com/spack/spack-packages/pull/1675, this drops the `syscalc` dependency in the `madgraph5amc` package, and also deprecates the only available version of `syscalc` in preparation of deleting that package down the line.

@oliviermattelaer @wdconinc


wdconinc (approved):
As discussed. Looks good to me.

PR #2382, opened 2025-11-12
kokkos-fft: fix minimal version requirement

tpadioleau:
- fix minimal version requirement of Kokkos
- add missing requirement for cmake

### DIFF
--- a/repos/spack_repo/builtin/packages/kokkos_fft/package.py
+++ b/repos/spack_repo/builtin/packages/kokkos_fft/package.py
@@ -37,8 +37,7 @@ class KokkosFft(CMakePackage):
     variant("tests", default=False, description="Enable tests")
 
     depends_on("cxx", type="build")
-
-    depends_on("cmake@3.22:3", when="@0.3:")
+    depends_on("cmake@3.22:3", type="build")
 
     depends_on("kokkos +complex_align")
     depends_on("kokkos@4.4:4", when="@0.3")
--- a/repos/spack_repo/builtin/packages/kokkos_fft/package.py
+++ b/repos/spack_repo/builtin/packages/kokkos_fft/package.py
@@ -38,7 +38,11 @@ class KokkosFft(CMakePackage):
 
     depends_on("cxx", type="build")
 
-    depends_on("kokkos@4.4:4 +complex_align")
+    depends_on("cmake@3.22:3", when="@0.3:")
+
+    depends_on("kokkos +complex_align")
+    depends_on("kokkos@4.4:4", when="@0.3")
+    depends_on("kokkos@4.5:4", when="@0.4:")
     # kokkos-fft currently only supports compilation with the Kokkos nvcc wrapper
     requires("^kokkos +serial", when="host_backend=fftw-serial")
     requires("^kokkos +openmp", when="host_backend=fftw-openmp")


tpadioleau:
@cedricchevalier19 @yasahi-hpc Can I get your review ?

repos/spack_repo/builtin/packages/kokkos_fft/package.py
```diff
@@ -38,7 +38,11 @@ class KokkosFft(CMakePackage):
 
     depends_on("cxx", type="build")
 
-    depends_on("kokkos@4.4:4 +complex_align")
+    depends_on("cmake@3.22:3", when="@0.3:")
+
+    depends_on("kokkos +complex_align")
+    depends_on("kokkos@4.4:4", when="@0.3")
+    depends_on("kokkos@4.5:4", when="@0.4:")
```
cedricchevalier19, line 44: What about Kokkos 5?

yasahi-hpc: At least version 0.4 should work with Kokkos 5.0, but there may be a lot of warnings

repos/spack_repo/builtin/packages/kokkos_fft/package.py
```diff
@@ -38,7 +38,11 @@ class KokkosFft(CMakePackage):
 
     depends_on("cxx", type="build")
 
-    depends_on("kokkos@4.4:4 +complex_align")
+    depends_on("cmake@3.22:3", when="@0.3:")
```
yasahi-hpc: We always need `cmake@3.22:3`. Do we still need `when` argument in this case?

repos/spack_repo/builtin/packages/kokkos_fft/package.py
tpadioleau: I was not sure if we support it. Has it been tested ? kokkos-fft 0.4 was released on July 30th and Kokkos 5 is about to be released.

repos/spack_repo/builtin/packages/kokkos_fft/package.py
tpadioleau: Not necessary in this PR indeed, I can remove it.

repos/spack_repo/builtin/packages/kokkos_fft/package.py
yasahi-hpc: I can give it a try
Maybe we have a new release after resolving issues like https://github.com/kokkos/kokkos-fft/issues/356

tpadioleau: > Maybe we have a new release after resolving issues like [kokkos/kokkos-fft#356](https://github.com/kokkos/kokkos-fft/issues/356)

I think it would be safer.

repos/spack_repo/builtin/packages/kokkos_fft/package.py
tpadioleau: Done in 831fcc8c51d89dfa2bf45ed6e6d5b42df594e48a I also put the dependency as `build` only.

yasahi-hpc (approved):
LGTM 
Thanks

cedricchevalier19 (approved):
LGTM

rbberger (approved):

PR #2390, opened 2025-11-12
osu-micro-benchmarks: Add XCCL benchmarks

nhanford:
Hello @MatthewLieber, @natshineman, and @harisubramoni,

NCCL and RCCL are now included by default in installations of CUDA and HIP, respectively. Therefore, rather than adding another variant, I have added logic to detect the appropriate versions of CUDA or HIP in the DAG and build the XCCL benchmarks automatically. They will also build if NCCL or RCCL are explicitly detected in the DAG. This has been tested on LLNL Tuolumne, Corona, and Matrix. However, I do not have old enough CUDA/ROCm installations to test the version cutoff explicitly.

Thanks,
Nate

### DIFF
--- a/repos/spack_repo/builtin/packages/osu_micro_benchmarks/package.py
+++ b/repos/spack_repo/builtin/packages/osu_micro_benchmarks/package.py
@@ -65,12 +65,14 @@ class OsuMicroBenchmarks(AutotoolsPackage, CudaPackage, ROCmPackage):
 
         if "+cuda" in spec:
             config_args.extend(["--enable-cuda", "--with-cuda=%s" % spec["cuda"].prefix])
+            config_args.extend(["--enable-ncclomb", "--with-nccl=%s", spec["cuda"].prefix])
             cuda_arch = spec.variants["cuda_arch"].value
             if "none" not in cuda_arch:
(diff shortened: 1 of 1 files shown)


MatthewLieber:
For the most part this looks good to me.  My only concern is I have been on systems that have external rocm/hip that does not contain rccl and I think this solution would not handle that well.  Is there something we can do about that or does that count as a "user error"

nhanford:
> For the most part this looks good to me. My only concern is I have been on systems that have external rocm/hip that does not contain rccl and I think this solution would not handle that well. Is there something we can do about that or does that count as a "user error"

Yeah so that's what would make adding a variant more sure-fire--We could depend on `rccl` or `nccl` explicitly when `+xccl`. This would make installations a bit more complex, however, because you would need to make sure you have `rccl` or `nccl` defined (most likely as an external) in your `packages.yaml` or your environment.

Since you mention that you've hit these cases in the wild, I'm inclined to do it by adding a variant. Please confirm and I'll do it.

Thanks,
Nate

alecbcs:
@MatthewLieber what do you think about the added `xccl` variant?

MatthewLieber (approved):

MatthewLieber:
> @MatthewLieber what do you think about the added `xccl` variant?

lgtm

alecbcs (approved):

PR #2393, opened 2025-11-12
qt: Update compiler mapping for Intel

AlexanderRichert-NOAA:
This PR changes intel->intel-oneapi-compilers-classic in the compiler mapping in the qt package.py. Intel-based builds fail without it. `get_mkspec()` doesn't seem to be used if %oneapi, but I updated it on principle.

### DIFF
--- a/repos/spack_repo/builtin/packages/qt/package.py
+++ b/repos/spack_repo/builtin/packages/qt/package.py
@@ -329,11 +329,11 @@ class Qt(Package):
 
     # Mapping for compilers/systems in the QT 'mkspecs'
     compiler_mapping = {
-        "intel": ("icc",),
+        "intel-oneapi-compilers-classic": ("icc",),
         # This only works because we apply patch "qt51514-oneapi.patch"
         # above that replaces calls to "icc" with calls to "icx" in
         # qtbase/mkspecs/*
-        "oneapi": ("icc",),
+        "intel-oneapi-compilers": ("icc",),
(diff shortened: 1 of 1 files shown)


AlexanderRichert-NOAA:
@spackbot rerun pipeline

repos/spack_repo/builtin/packages/qt/package.py
```diff
@@ -329,11 +329,11 @@ class Qt(Package):
 
     # Mapping for compilers/systems in the QT 'mkspecs'
     compiler_mapping = {
-        "intel": ("icc",),
+        "intel-oneapi-compilers-classic": ("icc",),
         # This only works because we apply patch "qt51514-oneapi.patch"
         # above that replaces calls to "icc" with calls to "icx" in
         # qtbase/mkspecs/*
-        "oneapi": ("icc",),
+        "intel-oneapi-compilers": ("icc",),
```
climbfuji, line 336: this must be icx, or?

AlexanderRichert-NOAA: If I recall correctly, qt doesn't have a separate configuration for oneAPI, so setting "icc" here basically tells it to use the icc-related config/flags.

climbfuji: Ok, thanks

climbfuji (approved):

becker33 (approved):

PR #2394, opened 2025-11-13
root: Allow disabling cxxmodules, turn off by default for macOS

paulgessinger:
https://github.com/spack/spack-packages/pull/109 turned the value of `cxxmodules` to ON by default, and also removed the forwarding of the variant to CMake. ROOT's default for `runtime_cxxmodules` is ON, but is ostensibly still not working on macOS at all (see
https://root-forum.cern.ch/t/build-error-on-macos-macports-with-unctrl-h-ncurses-h/40239/22)

@jmcarcell


jmcarcell (approved):
Note that ROOT's CI and LCG stacks use the default (`runtime_cxxmodules` set to ON) to build on MacOS, but it's possible there is a single ncurses coming from the system ([from HEPOSlibs](https://gitlab.cern.ch/linuxsupport/rpms/HEP_OSlibs/-/blob/el9/dependencies/HEP_OSlibs.x86_64.all-installed-nvra.txt#L332)).

vvolkl (approved):

PR #2397, opened 2025-11-13
Kokkos: fix host backend requirement

tpadioleau:
As we also support the experimental backend HPX in the package, I think we should allow a build with only HPX as a host backend.


cedricchevalier19 (approved):
LGTM

tpadioleau:
@rbberger is it ok for you to merge ?

rbberger (approved):
LGTM

PR #2402, opened 2025-11-13
VTK requires HL HDF5

jhale:
It seems that VTK (perhaps newer versions - I am not a developer so unsure of when precisely this became needed) requires the HDF5 HL interface. Not including `+hl` leads to a CMake error for VTK finding HDF5 with HL at configure time.


jhale:
I'm also unsure on style here - perhaps another `depends_on` line is preferred over including `+hl` in the existing lines.

hyoklee (approved):

johnwparent:
> I'm also unsure on style here - perhaps another `depends_on` line is preferred over including `+hl` in the existing lines.

There's no enforced conventions here AFAIK but in this case I think a separate clause is probably best.

johnwparent (approved):

PR #2403, opened 2025-11-13
protobuf dependents: fix versioning in protobuf requires action on dependents

albestro:
This is a PR that will be merged into the work going on in #2294 about fixing protobuf versioning.

For more information, have a look at the official documentation here https://protobuf.dev/support/version-support

Long story short: `protobuf` package in spack kept versioning the old way (3.*) and never switched to the new one (21.0). We're trying to fix this one way or another, but the main messages is that since 21.0 included (that comes after 3.20.*):
- `protobuf` version is the `protoc` version (21, 22, ... nowadays 33)
- you might still be interested in language runtime, which use a versioning where the major is the language version, and minor+patch compose the `protoc` version

This change requires some actions also on dependents, that's the mission of this PR.

Guidelines (I'm going to update this message in case other ideas/suggestions comes up)
- dependents should use the new versioning starting from 21.0
- if a dependent needs a specific language runtime compatibility, e.g. c++ 5.x, they have to check the compatibility table and require the right `protobuf` version, e.g. c++ 5 requires `protobuf@26:29`

What I did in this PR: I looked for all dependents and filtered for the ones with any kind of constraint on `protobuf`. It was mostly easy (see e8b4afbbbc26733549186a6d97d7f06b4c5820b8) because of the convenient number (e.g. `:3.21` is semantically equivalent to `:21` with the new versioning).

Specific cases that need attention:
- [x] `mivisionx`

### DIFF
--- a/repos/spack_repo/builtin/packages/mivisionx/package.py
+++ b/repos/spack_repo/builtin/packages/mivisionx/package.py
@@ -154,7 +154,7 @@ class Mivisionx(CMakePackage):
 
     depends_on("cmake@3.5:", type="build")
     depends_on("ffmpeg@4.4:", type="build")
-    depends_on("protobuf@:3", type="build")  # TODO
+    depends_on("protobuf@3.12.4:", type="build")
     depends_on(
         "opencv@4.5:"
         "+calib3d+features2d+highgui+imgcodecs+imgproc"
--- a/repos/spack_repo/builtin/packages/mivisionx/package.py
+++ b/repos/spack_repo/builtin/packages/mivisionx/package.py
@@ -154,7 +154,7 @@ class Mivisionx(CMakePackage):
 
     depends_on("cmake@3.5:", type="build")
     depends_on("ffmpeg@4.4:", type="build")
-    depends_on("protobuf@:3", type="build")
+    depends_on("protobuf@:3", type="build")  # TODO
     depends_on(
         "opencv@4.5:"
         "+calib3d+features2d+highgui+imgcodecs+imgproc"
--- a/repos/spack_repo/builtin/packages/py_protobuf/package.py
+++ b/repos/spack_repo/builtin/packages/py_protobuf/package.py
@@ -73,13 +73,13 @@ class PyProtobuf(PythonPackage):
     depends_on("py-setuptools", type=("build", "run"))
     depends_on("py-six@1.9:", when="@3.0:3.17", type=("build", "run"))
 
-    # Minor version must match protobuf
-    for ver in range(32, 33):
+    # https://protobuf.dev/support/version-support/#python
+    for ver in range(30, 33):
         depends_on(f"protobuf@{ver}", when=f"@6.{ver}")
     for ver in range(26, 29):
-        depends_on(f"protobuf@3.{ver}", when=f"@5.{ver}")
-    for ver in range(21, 26):
-        depends_on(f"protobuf@3.{ver}", when=f"@4.{ver}")
+        depends_on(f"protobuf@{ver}", when=f"@5.{ver}")
+    for ver in range(21, 25):
+        depends_on(f"protobuf@{ver}", when=f"@4.{ver}")
     for ver in range(0, 21):
         depends_on(f"protobuf@3.{ver}", when=f"@3.{ver}")
 
--- a/repos/spack_repo/builtin/packages/py_protobuf/package.py
+++ b/repos/spack_repo/builtin/packages/py_protobuf/package.py
@@ -76,9 +76,9 @@ class PyProtobuf(PythonPackage):
     # https://protobuf.dev/support/version-support/#python
     for ver in range(30, 33):
         depends_on(f"protobuf@{ver}", when=f"@6.{ver}")
-    for ver in range(26, 29):
+    for ver in range(26, 30):
         depends_on(f"protobuf@{ver}", when=f"@5.{ver}")
-    for ver in range(21, 25):
+    for ver in range(21, 26):
         depends_on(f"protobuf@{ver}", when=f"@4.{ver}")
     for ver in range(0, 21):
         depends_on(f"protobuf@3.{ver}", when=f"@3.{ver}")
--- a/repos/spack_repo/builtin/packages/py_tf_keras/package.py
+++ b/repos/spack_repo/builtin/packages/py_tf_keras/package.py
@@ -43,7 +43,7 @@ class PyTfKeras(PythonPackage):
 
     depends_on("protobuf@3.20.3", type="build", when="@2.18:")
     # TODO: uncomment for 2.19
-    # depends_on("protobuf@4.23.0", type="build", when="@2.19:")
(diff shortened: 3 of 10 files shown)


repos/spack_repo/builtin/packages/py_protobuf/package.py
albestro, line 1: it slightly changes the ranges, according to the linked doc with python runtime compatibilities
- there was an overlap for 26, which I took the chance to fix.
- there was a gap 30:31 (despite `protobuf` versions are not there in spack at the moment of writing)

repos/spack_repo/builtin/packages/py_tf_keras/package.py
```diff
@@ -43,7 +43,7 @@ class PyTfKeras(PythonPackage):
 
     depends_on("protobuf@3.20.3", type="build", when="@2.18:")
     # TODO: uncomment for 2.19
-    # depends_on("protobuf@4.23.0", type="build", when="@2.19:")
+    # depends_on("protobuf@23.0", type="build", when="@2.19:")
```
albestro, line 46: this change just to let the maintainers know about the change...not sure about the correctness

repos/spack_repo/builtin/packages/mivisionx/package.py
```diff
@@ -154,7 +154,7 @@ def patch(self):
 
     depends_on("cmake@3.5:", type="build")
     depends_on("ffmpeg@4.4:", type="build")
-    depends_on("protobuf@:3", type="build")
+    depends_on("protobuf@:3", type="build")  # TODO
```
albestro: not sure what the constraint meant...

if we leave it like this it would constrain the package to protobuf versions up to 3.20 included (i.e. pre versioning scheme change in protobuf)

albestro: @afzpatel @renjithravindrankannath @srekolam

repos/spack_repo/builtin/packages/py_tf_keras/package.py
albestro: FYI @jonas-eschle

repos/spack_repo/builtin/packages/mivisionx/package.py
afzpatel: This restriction was added 4 years ago with ROCm 4.2, I don't think it's needed anymore. The 7.1.0 doc says 3.12.4 or later:
https://github.com/ROCm/MIVisionX/blob/rocm-7.1.0/docs/install/MIVisionX-prerequisites.rst

albestro: Thanks for the reply!

Looking at the project CHANGELOG, I see

- `mivisionx@2.0:2.1` `protobuf@3.12.0:`
  - https://github.com/ROCm/MIVisionX/blob/develop/CHANGELOG.md#mivisionx-200
  - https://github.com/ROCm/MIVisionX/blob/develop/CHANGELOG.md#mivisionx-201
  - https://github.com/ROCm/MIVisionX/blob/develop/CHANGELOG.md#mivisionx-210
- `mivisionx@2.2:2.5` `protobuf@3.12.4:`, (probably `mivisionx@2.5:` too, given your comment and link)
  - https://github.com/ROCm/MIVisionX/blob/develop/CHANGELOG.md#mivisionx-220
  - https://github.com/ROCm/MIVisionX/blob/develop/CHANGELOG.md#mivisionx-250

...but to me eyes it does not match with the versioning in the spack package, which seems to reflect the `rocm` version more than the `mivisionx` one.

I don't know the project, so please let me know what you'd prefer.
Otherwise I'd go for a generic `protobuf@3.12.4:` and in case any problem will arise later, anyone will eventually fix it in a separate PR.

albestro: After a bit more digging, I think I realised how the (questionable) versioning works.

The spack package `mivisionx` is actually using the `rocm` versioning, and not its own (questionable, but I'm not here to question that 😅).

From the changelog I see that `2.5.0 -> 5.7.0` and `3.0.0 -> 6.2.0`. So, I'd say that we can easily go for `depends_on("protobuf@3.12.4:")` as you also suggested.

albestro: @afzpatel @renjithravindrankannath @srekolam 

please review and accept if you agree with the change

srekolam (approved):
okay with mivisionx change

afzpatel (approved):

renjithravindrankannath (approved):

albestro:
@adamjstewart

A small summary:
- ✅ `mivision` we got the approval from its maintainers
- ✅ `py_tf_keras` maintainer has been notified about the suggested change in the comment
- ⏳ `py_protobuf` @adamjstewart you as package maintainer might want to review the changes

After that, I'm going to merge this into its base #2294 and complete the work there.

repos/spack_repo/builtin/packages/py_protobuf/package.py
```diff
@@ -73,13 +73,13 @@ class PyProtobuf(PythonPackage):
...
-    for ver in range(32, 33):
+    # https://protobuf.dev/support/version-support/#python
+    for ver in range(30, 33):
         depends_on(f"protobuf@{ver}", when=f"@6.{ver}")
     for ver in range(26, 29):
-        depends_on(f"protobuf@3.{ver}", when=f"@5.{ver}")
-    for ver in range(21, 26):
-        depends_on(f"protobuf@3.{ver}", when=f"@4.{ver}")
+        depends_on(f"protobuf@{ver}", when=f"@5.{ver}")
+    for ver in range(21, 25):
```
adamjstewart: I don't really understand these changes. It now looks like certain versions like 4.25 don't depend on protobuf at all. Remember that python ranges are not inclusive.

albestro: good catch! you're absolutely right!

adamjstewart (approved):
ONNX and PyProtobuf seem fine now, although any missing versions in the depends_on will likely fail the package audit.

albestro:
> any missing versions in the depends_on will likely fail the package audit

are you looking at `py-protobuf` logic where we link `py-protobuf` versioning to the `protobuf` one?

if yes, I had a similar concern about maintainability, as in "everytime a new `protobuf` version is detected, we should remember to update also `py-protobuf` logic". And this potentially applies to other packages. 

I don't know the audit tool, and I'd be glad if a package audit can already detect that.

adamjstewart:
We can kind of detect that. Among the following steps:

1. Add new version to py-protobuf
2. Update `depends_on` in py-protobuf
3. Add new version to protobuf

If you forget 3, the audit will complain. If you forget 1 or 2, we can't check for that. 

At the moment, you added a new dependency on protobuf@30 (3.30). Because that version doesn't exist, the package audit will complain. We can fix this later if you want.

PR #2404, opened 2025-11-13
MVAPICH-Plus: change package type to base from autotools

MatthewLieber:
This update would change what type of package MVAPICH-Plus is from an Autotools package to a base one with just an install script.  This was done due to the fact that MVAPICH-Plus is closed source and we were having issues getting buildcaches to work across systems consistently.  We took inspiration for this from the cuda package.  If users fail to install their preferred spec a message should print out saying this is most likely due to them requesting a version that has not been generated yet and direct them to a form that will give us the info we need and get them in contact with us so that their version can be made.


MatthewLieber:
@spackbot fix style

repos/spack_repo/builtin/packages/mvapich_plus/package.py
```diff
@@ -120,87 +121,47 @@ def libs(self):
...
+        gpu="nogpu"
+        gpu_ver=""
+        apu=''
+        if spec.satisfies("^cuda"):
+            gpu = "cuda"
+            gpu_ver=str(spec['cuda'].version)[:4]
+        elif spec.satisfies('+rocm'):
+            gpu = "rocm"
+            gpu_ver=spec['hip'].version
+            if spec.satisfies('+apu'):
```
nhanford: Since you have a `ROCmPackage` here, I would try to use the `amdgpu_target` value to do this since the user already has to set that. Although now I see `gfx942` covers both the A and the X. For your RPM naming scheme, perhaps I would use both, so then you would wind up with `gfx942a`. This could help with future-proofing if/when you do similar things for the CUDA Vera-Rubins, etc.

repos/spack_repo/builtin/packages/mvapich_plus/package.py
```diff
@@ -120,87 +121,47 @@ def libs(self):
...
+        netmod = "ucx" if spec.satisfies("netmod=ucx") else "ofi"
+        comp = str(spec.compiler)
+        comp = f"{comp[:comp.find('@')]}{comp[comp.find('@')+1:comp.find('/')]}"
+        el="el8"
+        os="rhel8"
+        if str(spec.os) == "rocky9" or str(spec.os)=="rhel9":
+
+           el="el9"
+           os="rhel9"
+        ofed="24.10"
```
nhanford: Is this necessary for `netmod=ofi` as well?

repos/spack_repo/builtin/packages/mvapich_plus/package.py
```diff
@@ -120,87 +121,47 @@ def libs(self):
...
+        elif spec.satisfies('+rocm'):
+            gpu = "rocm"
+            gpu_ver=spec['hip'].version
+            if spec.satisfies('+apu'):
+                apu='.mi300a'
+
+        netmod = "ucx" if spec.satisfies("netmod=ucx") else "ofi"
+        comp = str(spec.compiler)
+        comp = f"{comp[:comp.find('@')]}{comp[comp.find('@')+1:comp.find('/')]}"
+        el="el8"
```
nhanford: On a similar note, I would probably harden this a bit in case you build for more OSes in the future...

nhanford (commented):
Hello @MatthewLieber,

Overall, this looks good. I just have a few notes regarding integration, future-proofing, etc.

Regards,
Nate

repos/spack_repo/builtin/packages/mvapich_plus/package.py
MatthewLieber: Right now since we are using rpms and the main thing that the os version changes on these is the glibc that gets used.  Since Rhel8 is oldest supported os we use that as a default since generally binaries that are compiled with a earlier glibc can get ran on systems with later glibc but not vice versa.  Currently we do not have the infrastructure to make rpms for other distros besides rhel 8 and 9.  In the future we may change the rpm naming scheme a bit and would take this into account along with your comment about apu specific rpms naming.  If we change the naming convention and which distros we build specifically for we definitely will update this part of the package.  Does this plan work for you?

repos/spack_repo/builtin/packages/mvapich_plus/package.py
MatthewLieber: I was thinking about keeping the +apu variant since it seems that we can not differentiate between if it is the apu model or not based on the amdgpu_target.  However instead of it just being in the rocm part we would make it so that it is consistent between amd and nvidia in the naming ie instead of mi300a being in the rpm name it would just specify apu so in the future there are no changes required.

repos/spack_repo/builtin/packages/mvapich_plus/package.py
MatthewLieber: yes all rpms we build have the ofed version in the name.

repos/spack_repo/builtin/packages/mvapich_plus/package.py
nhanford: Yup that could work as well. It's up to you how specific you want to get. Being more specific in the naming by including `cuda_arch` or `amdgpu_target` could reduce the potential for bugs in the future if implementations diverge, but since I don't have the code in front of me, I'm not sure how likely that is...

repos/spack_repo/builtin/packages/mvapich_plus/package.py
nhanford: Yup that works.

repos/spack_repo/builtin/packages/mvapich_plus/package.py
natshineman: That said, we are not building OFED specific versions anymore so we could consider dropping this from the naming convention in the RPMs.

natshineman (approved):
Good to proceed here.

repos/spack_repo/builtin/packages/mvapich_plus/package.py
```diff
@@ -120,87 +122,47 @@ def libs(self):
...
+
+        netmod = "ucx" if spec.satisfies("netmod=ucx") else "ofi"
+        comp = str(spec.compiler)
+        comp = f"{comp[:comp.find('@')]}{comp[comp.find('@')+1:comp.find('/')]}"
+        el = "el8"
+        os = "rhel8"
+        if str(spec.os) == "rocky9" or str(spec.os) == "rhel9":
+
+            el = "el9"
+            os = "rhel9"
```
becker33: Is this an exhaustive list of supported operating systems? If so, there should be a conflict for other operating systems.

repos/spack_repo/builtin/packages/mvapich_plus/package.py
```diff
@@ -120,87 +122,47 @@ def libs(self):
...
+        ofed = "24.10"
+        slurm = ""
         if spec.satisfies("process_managers=slurm"):
-            opts = ["--with-pm=none", "--with-pmi=slurm"]
-        if spec.satisfies("process_managers=cray"):
-            opts = ["--with-pm=none", "--with-pmi=pmi2", "--with-pmi2=/opt/cray/pe/pmi/default"]
-        if "none" in spec.variants["process_managers"].value:
-            opts = ["--with-pm=none"]
+            slurm = ".slurm"
+        rpm = f"mvapich-plus-{mvp_ver}-{gpu}{gpu_ver}.{os}.ofed{ofed}.{netmod}.{comp}\
```
becker33: It looks like this needs a new build dependency on rpm

repos/spack_repo/builtin/packages/mvapich_plus/package.py
```diff
@@ -120,87 +122,47 @@ def libs(self):
...
-        for x in ("hydra", "gforker", "remshell"):
-            if spec.satisfies(f"process_managers={x}"):
-                other_pms.append(x)
-
-        opts = []
-        if len(other_pms) > 0:
-            opts = ["--with-pm=%s" % ":".join(other_pms)]
+    def install(self, spec, prefix):
+        print(self.stage.source_path)
+        runfile = glob(join_path(self.stage.source_path, "*"))[0]
```
becker33: Why not just use `which("mvapich-plus-installer.sh")` and call that instead of calling bash to call it?

repos/spack_repo/builtin/packages/mvapich_plus/package.py
```diff
@@ -120,87 +122,47 @@ def libs(self):
...
+            gpu = "cuda"
+            gpu_ver = str(spec["cuda"].version)[:4]
+        elif spec.satisfies("+rocm"):
+            gpu = "rocm"
+            gpu_ver = spec["hip"].version
+            if spec.satisfies("+apu"):
+                apu = ".mi300a"
+
+        netmod = "ucx" if spec.satisfies("netmod=ucx") else "ofi"
+        comp = str(spec.compiler)
```
becker33: How will this handle if the spec has different compilers for c/cxx/fortran?

repos/spack_repo/builtin/packages/mvapich_plus/package.py
```diff
@@ -120,87 +122,47 @@ def libs(self):
...
+        gpu_ver = ""
+        apu = ""
+        if spec.satisfies("^cuda"):
+            gpu = "cuda"
+            gpu_ver = str(spec["cuda"].version)[:4]
+        elif spec.satisfies("+rocm"):
+            gpu = "rocm"
+            gpu_ver = spec["hip"].version
+            if spec.satisfies("+apu"):
+                apu = ".mi300a"
```
becker33, line 143: Can't we get this from the `amdgpu_target` instead of needing a separate variable for it?

becker33 (changes requested):

repos/spack_repo/builtin/packages/mvapich_plus/package.py
MatthewLieber: These are the only ones that we build for explicitly for now but rhel8 ones should work on later operating systems since the differentiator is glibc version in most cases.

repos/spack_repo/builtin/packages/mvapich_plus/package.py
MatthewLieber: we use rpm2cpio is that covered in the rpm package or is there a separate one I will need?

repos/spack_repo/builtin/packages/mvapich_plus/package.py
MatthewLieber: This logic was copied from the cuda package I can change it though.

repos/spack_repo/builtin/packages/mvapich_plus/package.py
MatthewLieber: Based on prior conversation we did not think you can distinguish if the gpu is the apu version.  Also we wanted to keep the new variant for apu since it is not necessarily dependent.  mi300a is just the only one we had so that is how the rpms are named in the future we may go to just adding "apu" in the name instead mi300a.

repos/spack_repo/builtin/packages/mvapich_plus/package.py
MatthewLieber: We only support there being one compiler ie gcc or intel not a mix of both.  Is there a change you would suggest that would force that or is this fine?

repos/spack_repo/builtin/packages/mvapich_plus/package.py
MatthewLieber: I tried to just use which on the shell script but ran into permissions issues using the bash command gets around this.  If you want I can make it a subprocess run command if that it preferred.

MatthewLieber:
@becker33 are the edits made sufficient or is there more that needs to be done here?

repos/spack_repo/builtin/packages/mvapich_plus/package.py
becker33: I don't like the unconstrained glob for finding an executable, that feels like a giant footgun waiting to happen.

Is there something I don't understand, or is the name always `mvapich-plus-installer.sh`?

repos/spack_repo/builtin/packages/mvapich_plus/package.py
becker33: Then we should add a requirement or conflict. E.g.

```
for name in spack.compilers.config.supported_compilers():
    requires("%fortran={name}", when="%c={name}")
```
That isn't perfect since it doesn't account for versions, but it's at least a help.

In addition, `str(spec.compiler)` probably doesn't do what you want anymore since Spack 1.0. I'm assuming you want something like `spec["c"].format("{name}{version}")`, but not all the variants and dependencies of the compiler? That should eliminate the need for the ugly string parsing below

repos/spack_repo/builtin/packages/mvapich_plus/package.py
becker33: Can you change this to be based on the libc then?

First, in the directives we would need `conflicts("%glibc@:2.27")` to avoid building on any earlier OS.

Then, here the check should be something like:

```
el = "el9" if spec["glibc"].satisfies("@2.34:") else "el8"
rhel = "rh" + el
```

repos/spack_repo/builtin/packages/mvapich_plus/package.py
```diff
@@ -120,87 +123,47 @@ def libs(self):
...
-            opts = ["--with-device=ch4:ofi"]
-        elif self.spec.satisfies("netmod=ucx"):
-            opts = ["--with-device=ch4:ucx"]
-        return opts
+            slurm = ".slurm"
+        rpm = f"mvapich-plus-{mvp_ver}-{gpu}{gpu_ver}.{os}.ofed{ofed}.{netmod}.{comp}\
+{slurm}{apu}-4.1-1.{el}.x86_64.rpm"
+
+        install_shell = which("bash")
+        io = f"{self.spec['rpm'].prefix}/bin/rpm2cpio"
```
becker33:
```suggestion
        io = which("rpm2cpio").path
```
This is less fragile, and Spack guarantees that the rpm from your dependencies is the one in PATH.

becker33 (changes requested):

repos/spack_repo/builtin/packages/mvapich_plus/package.py
MatthewLieber: I resolved all of these comments the compiler matching one went a little haywire but after talking to Herman in the help channel for a little bit I just put requires for the common compilers since I was having issues both using the supported_compilers function and add the requires in a loop.  If there are more compilers you would like to see added let me know.

becker33 (approved):

PR #2407, opened 2025-11-13
py-tensorflow: modify cuDNN dependency versions when `+cuda`

wdconinc:
This PR fixes concretization of `py-tensorflow@2.20 +cuda` since previously the following two requirements were mutually exclusive:
```py
        depends_on("cudnn@9.3:", when="@2.18:")
        depends_on("cudnn@8.9:8", when="@2.15:")
```


adamjstewart (approved):
@alalazo I'm surprised your script didn't catch this one.

PR #2420, opened 2025-11-14
hip: set --gcc-install-dir instead of --gcc-toolchain

rbberger:
The usage of `--gcc-toolchain` has been discouraged by LLVM for some time now (see https://github.com/llvm/llvm-project/pull/77537). While `--gcc-toolchain` still works, this should future proofs the hip dependent package logic.

> [!NOTE]
> The `setup_dependent_build_environment` logic in the `hip` packages depends on a compiler dependency. However, > externals don't automatically gain one during concretization. So this should be better documented.
>
> Externals of `hip` can specify a GCC compiler so that the correct GCC installation is selected for `hipcc`. This becomes > relevant for compilations with newer C++ standards, where the default system GCC might be too old.
>
> ```yaml
> packages:
>   hip:
>     externals:
>     - spec: "hip@6.4.3 %gcc@13.2.0"
>       prefix: /opt/rocm-6.4.3
>     buildable: false
> ```

### DIFF
--- a/repos/spack_repo/builtin/packages/hip/package.py
+++ b/repos/spack_repo/builtin/packages/hip/package.py
@@ -427,17 +427,19 @@ class Hip(CMakePackage):
         # being used to compile. This is only important for external ROCm
         # installations, which may otherwise pick up the wrong GCC toolchain.
         if self.spec.external and self.spec.satisfies("%gcc"):
-            # This is picked up by hipcc.
-            env.append_path(
-                "HIPCC_COMPILE_FLAGS_APPEND",
-                f"--gcc-toolchain={self.compiler.prefix}",
-                separator=" ",
+            gcc = Executable(self.compiler.cc)
+            libgcc_path = gcc("-print-file-name=libgcc.a", output=str, fail_on_error=False).strip()
+            libgcc_dir = os.path.abspath(os.path.dirname(libgcc_path))
+            gcc_install_dir_flag = (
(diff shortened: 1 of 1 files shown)


afzpatel (approved):

srekolam (approved):

renjithravindrankannath (approved):

repos/spack_repo/builtin/packages/hip/package.py
```diff
@@ -415,17 +415,19 @@ def set_variables(self, env):
...
-            )
-            # This is picked up by CMake when using HIP as a CMake language.
-            env.append_path("HIPFLAGS", f"--gcc-toolchain={self.compiler.prefix}", separator=" ")
+
+            if gcc_install_dir_flag:
+                # This is picked up by hipcc.
+                env.append_path("HIPCC_COMPILE_FLAGS_APPEND", gcc_install_dir_flag, separator=" ")
+                env.append_path("HIPCC_LINK_FLAGS_APPEND", gcc_install_dir_flag, separator=" ")
+                # This is picked up by CMake when using HIP as a CMake language.
+                env.append_path("HIPFLAGS", gcc_install_dir_flag, separator=" ")
```
becker33: I think you can use `env.append_flags` to avoid having to explicitly set the separator.

becker33 (dismissed):

repos/spack_repo/builtin/packages/hip/package.py
rbberger: @becker33 Fixed

becker33 (approved):

PR #2421, opened 2025-11-14
Add 1.15.2-levelist version for metkit

ninaburg:


ninaburg:
Closing this PR because changes are not needed after having updated Spack.

In Spack 21.2, the metkit package had no git source defined (see: https://github.com/spack/spack/blob/v0.21.2/var/spack/repos/builtin/packages/metkit/package.py
), so I had to add it manually.
With Spack 23.2, the builtin package now includes the git source (https://github.com/spack/spack/blob/v0.23.1/var/spack/repos/builtin/packages/metkit/package.py#L14
), so I can simply use the tag directly, e.g. metkit@git.1.15.2-levelist, without adding a new version.

PR #2429, opened 2025-11-14
uwtools: add 'main' and update deps

AlexanderRichert-NOAA:
This PR:
 - adds 'main' branch to uwtools package.py to facilitate testing etc.
 - ensures py-lxml is always a dependency including when using 'main'
 - allows versions 2.12 onward to use a slightly older, non-rust dependent version of py-jsonschema.


sethrj (approved):
Confirmed increased compatibility via https://github.com/ufs-community/uwtools/commit/6d7541baf6f3417e5ba8e44a3ec8d9b607239bdb

PR #2430, opened 2025-11-14
parallelio: add placement's for resources

AlexanderRichert-NOAA:
This PR adds placement arguments for the resources in parallelio. This seems to be the only way to consistently have the path of the resource be correctly set, especially when using cached source code.


jedwards4b (approved):
Thanks for the fix.

sethrj (approved):

PR #2431, opened 2025-11-14
mapl: add placement's for resources

AlexanderRichert-NOAA:
This PR adds placement arguments for the resources in mapl. This seems to be the only way to consistently have the path of the resource be correctly set, especially when using cached source code.


mathomp4 (approved):
Interesting. I wonder why it was never needed until now? I mean, seems benign.

AlexanderRichert-NOAA:
It relates to an issue with installing from source caching that I've never been totally consistently able to reproduce, so it's a bit of an edge case (most people most of the time install from internet-connected systems). In any case this seems to work solidly both with and without pre-cached resources.

sethrj (approved):

PR #2432, opened 2025-11-14
cprnc: use placement instead of destination for genf90 resource

AlexanderRichert-NOAA:
This PR use placement instead of destination for the genf90 resource in cprnc. This seems to be the only way to consistently have the path of the resource be correctly set, especially when using cached source code.


AlexanderRichert-NOAA:
@spackbot fix style

jedwards4b (approved):

sethrj (approved):

PR #2434, opened 2025-11-15
mpitrampoline: add v5.5.0, deprecate old versions

alalazo:


eschnett:
The version update looks good to me. I am not familiar with modern Spack and cannot comment on the changes to the compiler wrapper setup.

PR #2436, opened 2025-11-16
ggml: add new package

rbberger:
Dependency for `llama-cpp`. https://github.com/spack/spack-packages/pull/2437

@adamjstewart do you know what happened to the apple accelerate / veclib spack package for BLAS?

### DIFF
--- a/repos/spack_repo/builtin/packages/ggml/package.py
+++ b/repos/spack_repo/builtin/packages/ggml/package.py
@@ -1,0 +1,81 @@
+# Copyright Spack Project Developers. See COPYRIGHT file for details.
+#
+# SPDX-License-Identifier: (Apache-2.0 OR MIT)
+
+from spack_repo.builtin.build_systems.cmake import CMakePackage
+from spack_repo.builtin.build_systems.cuda import CudaPackage
+from spack_repo.builtin.build_systems.rocm import ROCmPackage
+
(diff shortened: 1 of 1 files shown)


adamjstewart:
Not sure if we've ever had one. I've used `veclibfort` before which I think wraps the system veclib and adds Fortran support.

rbberger:
This one is what I remember https://github.com/spack/spack/pull/49395. Seems to just be forgotten / abandoned. I would just need something for the case `GGML_BLAS_VENDOR` being set to `Apple`.

aweits:
Breaking out ggml makes sense, can we grab the RPC setting/patch from #2314?

aweits:
#2317 can also be changed to make use of this.

rbberger:
@aweits sorry I didn't catch that you submitted a PR too. Yeah, it seems like having them split is the more spacky way of doing things and avoids duplicating the variants. I'll look at adding your changes too.

rbberger:
Combining this with https://github.com/spack/spack-packages/pull/2437 since they are tightly bound together.

elfprince13:
@rbberger fwiw I'm using the package no problems on our private fork, although it almost certainly could use some improvement, but we're still pre-1.0 so I haven't pulled it out of builtin into the standalone package repo yet.

PR #2437, opened 2025-11-16
llama-cpp and ggml: new packages

rbberger:
This is a useful LLM package for quickly running LLMs without many dependencies. Installing it via Spack allows you to then run something like:

```
llama-server -hf ggml-org/gpt-oss-20b-GGUF --jinja -c 0 --host 127.0.0.1 --port 8033
```

A python interface was added in https://github.com/spack/spack-packages/pull/2441

- [x] ~Depends on: https://github.com/spack/spack-packages/pull/2436~ (merged with this one)

Given the high development velocity of llama-cpp, it makes more sense to build with the embedded GGML, which is why both packages share a common base class for adding GGML build options, variants and dependencies.

### DIFF
--- a/repos/spack_repo/builtin/packages/ggml/package.py
+++ b/repos/spack_repo/builtin/packages/ggml/package.py
@@ -1,0 +1,81 @@
+# Copyright Spack Project Developers. See COPYRIGHT file for details.
+#
+# SPDX-License-Identifier: (Apache-2.0 OR MIT)
+
+from spack_repo.builtin.build_systems.cmake import CMakePackage
+from spack_repo.builtin.build_systems.cuda import CudaPackage
+from spack_repo.builtin.build_systems.rocm import ROCmPackage
+
+from spack.package import *
+
+
+class Ggml(CMakePackage, CudaPackage, ROCmPackage):
+    """Tensor library for machine learning"""
+
+    homepage = "https://github.com/ggml-org/ggml"
+    git = "https://github.com/ggml-org/ggml.git"
+
+    maintainers("rbberger")
+
+    license("MIT")
+
+    version("master", branch="master")
+    version("0.9.4", tag="v0.9.4", commit="72632094336524a9c809e129e8b1c52154543a5a")
+
+    variant("shared", default=True, description="build shared libraries")
+    variant("cpu", default=True, description="build CPU backend")
+    variant("blas", default=True, description="build BLAS backend")
+    variant("openmp", default=True, description="build OpenMP backend")
+    variant("cuda", default=False, description="build CUDA backend")
+    variant("rocm", default=False, description="build HIP backend")
+    variant("metal", default=True, description="build Metal backend", when="platform=darwin")
+
+    depends_on("c", type="build")
+    depends_on("cxx", type="build")
+
+    depends_on("git", type="build")
+
+    depends_on("blas", when="+blas")
+    depends_on("llvm-openmp", when="platform=darwin %apple-clang")
+
+    depends_on("hipblas", when="+rocm")
+    depends_on("rocblas", when="+rocm")
+
+    def cmake_args(self):
+        args = [
+            self.define_from_variant("BUILD_SHARED_LIBS", "shared"),
+            self.define_from_variant("GGML_CPU", "cpu"),
+            self.define_from_variant("GGML_BLAS", "blas"),
+            self.define_from_variant("GGML_OPENMP", "openmp"),
+            self.define_from_variant("GGML_CUDA", "cuda"),
+            self.define_from_variant("GGML_ROCM", "rocm"),
+            self.define_from_variant("GGML_METAL", "metal"),
+        ]
+
+        if self.spec.satisfies("+blas"):
+            if self.spec.satisfies("%blas=openblas"):
+                blas_vendor = "OpenBLAS"
+            elif self.spec.satisfies("%blas=intel-oneapi-mkl"):
+                blas_vendor = "Intel"
+            else:
+                blas_vendor = "Generic"
+            args.append(self.define("GGML_BLAS_VENDOR", blas_vendor))
+
+        if self.spec.satisfies("+cuda"):
+            args.append(self.define("CMAKE_CUDA_COMPILER", f"{self.spec['cuda'].prefix}/bin/nvcc"))
+            if not self.spec.satisfies("cuda_arch=none"):
+                archs = self.spec.variants["cuda_arch"].value
+                arch_str = ";".join(archs)
+                args.append(self.define("CMAKE_CUDA_ARCHITECTURES", arch_str))
+
+        if self.spec.satisfies("+rocm"):
+            args.append(
+                self.define(
+                    "CMAKE_HIP_COMPILER", f"{self.spec['llvm-amdgpu'].prefix}/bin/amdclang++"
+                )
+            )
+            if not self.spec.satisfies("amdgpu_target=none"):
+                archs = self.spec.variants["amdgpu_target"].value
+                arch_str = ";".join(archs)
+                args.append(self.define("CMAKE_HIP_ARCHITECTURES", arch_str))
+        return args
--- a/repos/spack_repo/builtin/packages/ggml/package.py
+++ b/repos/spack_repo/builtin/packages/ggml/package.py
@@ -29,6 +29,7 @@ class Ggml(CMakePackage, CudaPackage, ROCmPackage):
     variant("cuda", default=False, description="build CUDA backend")
     variant("rocm", default=False, description="build HIP backend")
     variant("metal", default=True, description="build Metal backend", when="platform=darwin")
+    variant("rpc", default=False, description="build with RPC support")
 
     depends_on("c", type="build")
     depends_on("cxx", type="build")
@@ -50,6 +51,7 @@ class Ggml(CMakePackage, CudaPackage, ROCmPackage):
             self.define_from_variant("GGML_CUDA", "cuda"),
             self.define_from_variant("GGML_ROCM", "rocm"),
             self.define_from_variant("GGML_METAL", "metal"),
+            self.define_from_variant("GGML_RPC", "rpc"),
         ]
 
         if self.spec.satisfies("+blas"):
--- a/repos/spack_repo/builtin/packages/ggml/package.py
+++ b/repos/spack_repo/builtin/packages/ggml/package.py
@@ -84,6 +84,4 @@ class Ggml(GGMLPackageBase):
     license("MIT")
 
     version("master", branch="master")
-    version("0.9.4-20251120", commit="781baf2a14d9e0aaee542b2e1bb918bfc4132199")
-    version("0.9.4-20251117", commit="c23776f22d616d8cb635145381cad365bac675e7")
     version("0.9.4", tag="v0.9.4", commit="72632094336524a9c809e129e8b1c52154543a5a")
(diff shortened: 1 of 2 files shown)


aweits:
see #2314, can we collab on this?

rbberger:
@aweits can you try this one out and see if it does what you want? I've used the merge request patch to fix the RPC install.

aweits:
I'll look, but i think the patch will have to be carried in the ggml package...

rbberger:
@aweits the main issue with the split I see is having a matching ggml release for a llama-cpp release/tag. There's always git versions to make it work with master, but we probably want at least the named versions have a functional counterpart. Should we just add our own for certain sync points?

aweits:
> @aweits the main issue with the split I see is having a matching ggml release for a llama-cpp release/tag. There's always git versions to make it work with master, but we probably want at least the named versions have a functional counterpart. Should we just add our own for certain sync points?

I haven't looked at if gitversions have matured enough to use them, but we certainly can put the "versions" we require in there i guess? I'll poke around in a bit...

repos/spack_repo/builtin/packages/llama_cpp/package.py
```diff
@@ -0,0 +1,57 @@
...
+
+    homepage = "https://github.com/ggml-org/llama.cpp"
+    git = "https://github.com/ggml-org/llama.cpp.git"
+
+    maintainers("rbberger")
+
+    license("MIT")
+
+    version("master", branch="master")
+    version("7086" tag="b7086", commit="7aaeedc098a77e9323044187101db4f6b69988da")
```
aweits:
```suggestion
    version("7086", tag="b7086", commit="7aaeedc098a77e9323044187101db4f6b69988da")
```

aweits:
ok, this works for me, can we add the rpc variant to llama-cpp as well?

rbberger:
@aweits you're asking for convenience for doing `llama-cpp+rpc` instead of `llama-cpp ^ggml+rpc`? Since it does seem to work without explicitly setting any more CMake flags, it would only be used to require `ggml+rpc`.

aweits:
> @aweits you're asking for convenience for doing `llama-cpp+rpc` instead of `llama-cpp ^ggml+rpc`? Since it does seem to work without explicitly setting any more CMake flags, it would only be used to require `ggml+rpc`.

That makes sense to me - easier for folks to install without having to infer more details around how ggml is/was built.

rbberger:
@aweits I went ahead and created a common base class for both ggml and llama-cpp. You probably can do the same once this one is merged with whisper.cpp. Given the speed of llama-cpp, I'm defaulting this now to prefer doing the embedded GGML build. This should give us the best of both worlds and a quicker way of running current llama-cpp without having to wait for ggml to get synced. I've dropped the patch for RPC and simply removed the older versions that aren't patched yet.

aweits:
> @aweits I went ahead and created a common base class for both ggml and llama-cpp. You probably can do the same once this one is merged with whisper.cpp. Given the speed of llama-cpp, I'm defaulting this now to prefer doing the embedded GGML build. This should give us the best of both worlds and a quicker way of running current llama-cpp without having to wait for ggml to get synced. I've dropped the patch for RPC and simply removed the older versions that aren't patched yet.

I like this a lot, let me do a couple of test builds...

aweits (dismissed):
LGTM.
```
[aweits@coltrane spack-packages]$ spack find -lv llama-cpp
-- linux-rocky9-skylake / %c,cxx=gcc@13.2.0 ---------------------
52kpqnt llama-cpp@7158+blas+cpu+cuda+curl~ipo+openmp~rocm+rpc+shared~system_ggml build_system=cmake build_type=Release commit=b3b03a7baf387cfeaf56641bd14c06dbd3d2fcf0 cuda_arch:=80,86 generator=make
```

repos/spack_repo/builtin/packages/ggml/package.py
```diff
@@ -0,0 +1,89 @@
...
+    git = "https://github.com/ggml-org/ggml.git"
+
+    maintainers("rbberger")
+
+    license("MIT")
+
+    version("master", branch="master")
+    version("0.9.4-20251120", commit="781baf2a14d9e0aaee542b2e1bb918bfc4132199")
+    version("0.9.4-20251117", commit="c23776f22d616d8cb635145381cad365bac675e7")
+    version("0.9.4", tag="v0.9.4", commit="72632094336524a9c809e129e8b1c52154543a5a")
```
adamjstewart, line 87: Should we default to this version because it's a real release?

repos/spack_repo/builtin/packages/llama_cpp/package.py
```diff
@@ -0,0 +1,61 @@
...
+    maintainers("rbberger")
+
+    license("MIT")
+
+    version("master", branch="master")
+    version("7158", tag="b7158", commit="b3b03a7baf387cfeaf56641bd14c06dbd3d2fcf0")
+    version("7086", tag="b7086", commit="7aaeedc098a77e9323044187101db4f6b69988da")
+
+    variant("curl", default=True, description="use curl for model download")
+    variant("system_ggml", default=False, description="build embedded GGML library")
```
adamjstewart: We should always rely on Spack-installed packages if possible. I would personally drop this variant and always build GGML separately. If we keep this variant, I find the current variant name and description to be confusing.

repos/spack_repo/builtin/packages/llama_cpp/package.py
```diff
@@ -0,0 +1,61 @@
...
+
+    depends_on("curl", when="+curl")
+    depends_on("ggml", when="+system_ggml")
+
+    depends_on("ggml@0.9.4-20251117:", when="@7086 +system_ggml")
+
+    for v in ("cpu", "blas", "openmp", "cuda", "rocm", "metal", "rpc"):
+        depends_on(f"ggml+{v}", when=f"+system_ggml +{v}")
+        depends_on(f"ggml~{v}", when=f"+system_ggml ~{v}")
```
adamjstewart, line 36: Why are these variants needed in `llama-cpp`? We try to avoid variant forwarding whenever possible, otherwise every package would have variants for all of its dependencies combined. Users can always install `llama-cpp ^ggml+cuda`, right?

repos/spack_repo/builtin/packages/llama_cpp/package.py
```diff
@@ -0,0 +1,61 @@
...
+import os
+
+from spack_repo.builtin.build_systems.cuda import CudaPackage
+from spack_repo.builtin.build_systems.rocm import ROCmPackage
+from spack_repo.builtin.packages.ggml.package import GGMLPackageBase
+
+from spack.package import *
+
+
+class LlamaCpp(GGMLPackageBase):
```
adamjstewart, line 14: I'm not sure if this base class is necessary. This package looks rather simple, all of the complexity is in GGML. If we always depend on GGML and keep all the variants there, none of this complexity is needed in llama-cpp.

adamjstewart (changes requested):
I think this could be greatly simplified to behave more like other Spack packages.

repos/spack_repo/builtin/packages/llama_cpp/package.py
aweits: Unfortunately, ggml is not always in-sync with what's vendored in whisper-cpp and llama-cpp. Breaking it out ostensibly creates more problems than it solves - thus this solution for now.

repos/spack_repo/builtin/packages/llama_cpp/package.py
aweits: That really is confusing for users that might not even be aware of the existence of ggml and want to distribute inference jobs across gpus - I'd prefer to hide the existence of ggml and leave the variants on llama/whisper in that case.

adamjstewart: It just makes this package orders of magnitude more complicated than it needs to be. If y'all want to maintain this complexity, I can't stop you.

repos/spack_repo/builtin/packages/llama_cpp/package.py
adamjstewart: Not in-sync as in a specific commit, or not in-sync as in there are internal modifications not present in GGML? For the former, this is how PyTorch is packaged, we just pin it to commits. For the latter, I wouldn't even allow external dependencies in that case.

rbberger: @adamjstewart I had the same thoughts as you initially. However, llama.cpp and whisper.cpp development is a bit weird. ggml is developed as part of those clients and periodically syncing back to ggml. There are no good current releases in ggml itself. It's worse than a submodule. It's inverted in terms of what's current.

adamjstewart: In that case, we can't guarantee that ggml is ever up-to-date enough to use to build llama-cpp, and there's no way to know which versions are compatible. I would remove the option to build with Spack-installed ggml.

repos/spack_repo/builtin/packages/ggml/package.py
rbberger: No. It's very outdated.

repos/spack_repo/builtin/packages/llama_cpp/package.py
rbberger: Argh, yes forgot to update the description. The variant name matches the cmake option.

repos/spack_repo/builtin/packages/llama_cpp/package.py
rbberger: The hope would be that as it matures they'll have proper releases. What you see now is an approach to avoid duplication. I do imagine ggml being useful on its own, but I didn't want to maintain the cmake logic in two places.

repos/spack_repo/builtin/packages/ggml/package.py
adamjstewart: I know how fast the ML world works, but 2 months ago is outdated?

rbberger: If you consider their "releases", yes.

repos/spack_repo/builtin/packages/llama_cpp/package.py
rbberger: Better?

repos/spack_repo/builtin/packages/ggml/package.py
rbberger: @adamjstewart I was considering removing all versions except `master`. What you see remaining were my attempts of finding usable versions between a pure llama-cpp and a ggml package. The fake versions are sync points. Since I basically gave up with the sync points and made the internal build the default, probably I can drop the fake ggml versions altogether.

rbberger: I removed the fake versions and only kept "real" releases in ggml.

repos/spack_repo/builtin/packages/llama_cpp/package.py
rbberger: @adamjstewart I agree. So much that I argue the same as you in https://github.com/spack/spack-packages/issues/2513 (join the fun ;-) ). What you see here is a middle ground after trying what you are proposing.

adamjstewart (approved):
It isn't necessarily how I would do it, but I don't want to block this PR from being merged.

aweits (approved):

PR #2445, opened 2025-11-17
cdi: change origin to gitlab repo, add version 2.5.4 and add cmake fix for latest versions

albestro:
I discovered there is a gitlab repo available for `cdi`, so I preferred switching the origin to it.

Changelog:
- switched origin for all versions
- differently from release tarballs (used previously), now `autoreconf` is needed
- add `depends_on("c")`
- add 2.5.4
- add a fix for cmake config provided by the package, which looks for the wrong library file (I'm going to report that)

### DIFF
--- a/repos/spack_repo/builtin/packages/cdi/package.py
+++ b/repos/spack_repo/builtin/packages/cdi/package.py
@@ -17,6 +17,7 @@ class Cdi(AutotoolsPackage):
     url = "https://gitlab.dkrz.de/mpim-sw/libcdi/-/archive/cdi-2.4.0/download.tar.gz"
     git = "https://gitlab.dkrz.de/mpim-sw/libcdi"
 
+    version("2.5.4", sha256="605d4a9192e9657d87bdbe544e31c96b81890e9acdfc137ecc8581721ce5390f")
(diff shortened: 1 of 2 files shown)


repos/spack_repo/builtin/packages/cdi/package.py
```diff
@@ -28,28 +14,35 @@ class Cdi(AutotoolsPackage):
...
-        if resource_id is None:
-            raise InstallError(f"don't know which url to use to download cdi@{version}")
-        return "https://code.mpimet.mpg.de/attachments/download/{}/cdi-{}.tar.gz".format(
-            resource_id, version
-        )
+    # note:
+    # starting from 2.5.1 the provided cmake config file looks for `cdi` instead of `libcdi`,
+    # but autotools still builds `cdi`.
+    # cmake build system is going to build `libcdi` in future releases.
+    patch("cmake-config-libname.patch", when="@2.5.1: build_system=autotools")
```
msimberg, line 45: I guess you don't know an upper bound for this yet?

msimberg (approved):

repos/spack_repo/builtin/packages/cdi/package.py
albestro: exactly, I just opened an issue and I'm waiting for confirmation.

currently it is still like that with the latest relase

PR #2449, opened 2025-11-17
mvapich2: deprecate all versions

alalazo:
The current package is called `mvapich`, latest version is v4.1


alalazo:
@spackbot fix style

natshineman (approved):
Looks good to me, we want folks to migrate to the latest version wherever possible.

PR #2451, opened 2025-11-17
Add tag to the latest MOAB version v5.6.0

vijaysm:
Updating hash to point to the latest MOAB v5.6.0 tag.

Additional reviewer: @xylar

### DIFF
--- a/repos/spack_repo/builtin/packages/moab/package.py
+++ b/repos/spack_repo/builtin/packages/moab/package.py
@@ -19,7 +19,7 @@ class Moab(AutotoolsPackage):
 
     homepage = "https://sigma.mcs.anl.gov/moab-library"
     git = "https://bitbucket.org/fathomteam/moab.git"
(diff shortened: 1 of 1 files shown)


vijaysm:
Tested installs on arch-linux, OSX and Ubuntu. Worked fine for different variant combos.

@xylar For E3SM workflows, I'd recommend something like the following to get the entire stack installed in one shot. 
`moab@5.6.0+eigen+fortran+hdf5+mpi+netcdf+pnetcdf~shared+zoltan+tempest`

xylar (approved):
Looks good!

xylar:
> @xylar For E3SM workflows, I'd recommend something like the following to get the entire stack installed in one shot.
moab@5.6.0+eigen+fortran+hdf5+mpi+netcdf+pnetcdf~shared+zoltan+tempest

I'll make sure that's what we're doing.  We still aren't ready for spack 1.x.x yet so we have our own fork of older spack but I've already added this version of moab there as well.

kshea21 (approved):

PR #2454, opened 2025-11-17
Added Sphinx Contrib Tikz

tuxfan:
Adds support for tikz Sphinx extension for PGF/TikZ LaTeX package.

### DIFF
--- a/repos/spack_repo/builtin/packages/py_sphinxcontrib_tikz/package.py
+++ b/repos/spack_repo/builtin/packages/py_sphinxcontrib_tikz/package.py
@@ -1,0 +1,44 @@
+# Copyright Spack Project Developers. See COPYRIGHT file for details.
+#
+# SPDX-License-Identifier: (Apache-2.0 OR MIT)
+
+from spack_repo.builtin.build_systems.python import PythonPackage
+
+from spack.package import *
+
+
+class PySphinxcontribTikz(PythonPackage):
+    """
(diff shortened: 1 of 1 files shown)


adamjstewart (commented):
All Python deps are now correct. There are some non-Python deps as well: https://sphinxcontrib-tikz.readthedocs.io/en/latest/#prerequisites-and-configuration.

I would suggest:
```python
depends_on("texlive", type="run")
depends_on("pdf2svg", type="run")
depends_on("poppler", type="run")
depends_on("netpbm", type="run")
depends_on("imagemagick", type="run")
depends_on("ghostscript", type="run")
```
Other than texlive, these deps could all be optional. For example, see the `py-matplotlib` package's `backend` variant. We could do the same thing here for `suite` if you want to make it easier to install.

repos/spack_repo/builtin/packages/py_sphinxcontrib_tikz/package.py
```diff
@@ -0,0 +1,49 @@
...
+    ]
+
+    default_suite = "pdf2svg"
+
+    variant(
+        "suite",
+        default=default_suite,
+        description="Conversion suite used to convert pdfs to various image formats",
+        values=all_suites,
+        multi=False,
```
adamjstewart:
```suggestion
        multi=True,
```
You can install all 4 at the same time if you want to

adamjstewart (approved):

PR #2462, opened 2025-11-18
openblas: deprecate old versions

alalazo:
Deprecate all versions before 2022


alalazo:
@spackbot fix style

mathomp4 (approved):
Looks good to me.

(And reminds me I've had some deprecated packages in `mapl` for a loooong time. Might be time to remove them...)

PR #2466, opened 2025-11-18
viskores: add v1.1.0

vicentebolea:
- **adios2: add non-default 2.11.0-rc1**
- **viskores: add v1.1.0 and set it as preferred**

### DIFF
--- a/repos/spack_repo/builtin/packages/viskores/package.py
+++ b/repos/spack_repo/builtin/packages/viskores/package.py
@@ -32,13 +32,12 @@ class Viskores(CMakePackage, CudaPackage, ROCmPackage):
 
     version("master", branch="master")
     version("release", branch="release")
-    version("1.1.0-rc1", sha256="cdf04a193798e284bf71dc7e22868c531f11e158431c19deae016565f450550d")
     version(
-        "1.0.0",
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/adios2/package.py
```diff
@@ -30,6 +30,9 @@ class Adios2(CMakePackage, CudaPackage, ROCmPackage):
     license("Apache-2.0")
 
     version("master", branch="master")
+    version(
+        "2.11.0-rc1", sha256="5885ef24e2567d0240acd18053e7a003a15ed94b59c967bce6456257d4e17abe"
```
ax3l: @vicentebolea I merged #2439, so this line here needs a rebase please

vicentebolea: @ax3l thanks for catching this. Resolved in the last push

vicentebolea: @ax3l all resolved

kmorel (approved):
I can verify that my laptop installs version 1.1 by default.

ax3l (approved):

PR #2467, opened 2025-11-18
WarpX 25.11

ax3l:
And dependencies.

- [x] rebase against #2469

### DIFF
--- a/repos/spack_repo/builtin/packages/py_picmistandard/package.py
+++ b/repos/spack_repo/builtin/packages/py_picmistandard/package.py
@@ -12,11 +12,12 @@ class PyPicmistandard(PythonPackage):
 
     homepage = "https://picmi-standard.github.io"
     git = "https://github.com/picmi-standard/picmi.git"
-    pypi = "picmistandard/picmistandard-0.33.0.tar.gz"
+    pypi = "picmistandard/picmistandard-0.34.0.tar.gz"
 
     maintainers("ax3l", "dpgrote", "RemiLehe")
 
     version("master", branch="master")
+    version("0.34.0", sha256="28db66c652c7ce7bedf79ecf7e764a02fa5e51ac05a2fb1cf9433354ac056ec1")
     version("0.33.0", sha256="bdab1643385d85da1462ba8cebc4460dd87c735946f4766181714d7cb8cf2188")
     version("0.30.0", sha256="28b892b242e0cc044ad987d6bdc12811fe4a478d5096d6bc5989038ee9d9dab6")
     version("0.29.0", sha256="dc0bf3ddd3635df9935ac569b3085de387150c4f8e9851897078bb12d123dde8")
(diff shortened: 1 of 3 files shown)


EZoni:
Just cross-referencing https://github.com/BLAST-WarpX/warpx/issues/5847 here in case it's easy to fix for you. One reason why I had backed down from the Spack updates was that I got stuck on this issue and couldn't find a way to fix it.

repos/spack_repo/builtin/packages/warpx/package.py
EZoni, line 55: Should we take care of the new geometries here or would you do that in a separate PR?

ax3l: Yeah, we should expose them, unless they break something. I am not yet fully up to speed on them, but since they are not yet supported in the Python bindings, I was a bit hesitant to ship them.

Do we have small enough tests we can run on them to see if the Spack build was successful (like the LWFA example that we run for the existing four geometries)?

EZoni: I think these are the only tests we have in those geometries:

https://github.com/BLAST-WarpX/warpx/blob/cbe73680443ea633c24ba79f133e8d8b63310baf/Examples/Tests/langmuir/CMakeLists.txt#L425-L443

```cmake
add_warpx_test(
    test_rcylinder_langmuir_multi  # name
    RCYLINDER  # dims
    2  # nprocs
    inputs_test_rcylinder_langmuir_multi  # inputs
    "analysis_r1d.py diags/diag1000080" # analysis
    "analysis_default_regression.py --path diags/diag1000080"  # checksum
    OFF  # dependency
)


add_warpx_test(
    test_rsphere_langmuir_multi  # name
    RSPHERE  # dims
    2  # nprocs
    inputs_test_rsphere_langmuir_multi  # inputs
    "analysis_r1d.py diags/diag1000080"  # analysis
    "analysis_default_regression.py --path diags/diag1000080"  # checksum
    OFF  # dependency
)
```

ax3l: I see. We can then use those.
But with the low test coverage I am unsure how production-ready these geoms are :D

EZoni: I agree, I would say probably not very production-ready.

ax3l: Let's keep it a TODO then. No need to burn CI/CD time to build and ship this just yet.

RemiLehe (approved):

repos/spack_repo/builtin/packages/py_picmistandard/package.py
```diff
@@ -12,11 +12,12 @@ class PyPicmistandard(PythonPackage):
 
     homepage = "https://picmi-standard.github.io"
     git = "https://github.com/picmi-standard/picmi.git"
-    pypi = "picmistandard/picmistandard-0.33.0.tar.gz"
+    pypi = "picmistandard/picmistandard-0.34.0.tar.gz"
 
     maintainers("ax3l", "dpgrote", "RemiLehe")
```
ax3l:
```suggestion
    maintainers("ax3l", "dpgrote", "EZoni", "RemiLehe")
```

EZoni (approved):

alalazo (approved):

PR #2468, opened 2025-11-18
py-cgsmiles: new package

adamwitmer:
@rbberger this is a new package for `cgsmiles`, which is referenced at: https://github.com/gruenewald-lab/CGsmiles

### DIFF
--- a/repos/spack_repo/builtin/packages/py_cgsmiles/package.py
+++ b/repos/spack_repo/builtin/packages/py_cgsmiles/package.py
@@ -1,0 +1,27 @@
+# Copyright Spack Project Developers. See COPYRIGHT file for details.
+#
+# SPDX-License-Identifier: (Apache-2.0 OR MIT)
+
+from spack_repo.builtin.build_systems.python import PythonPackage
+
+from spack.package import *
+
+
+class PyCgsmiles(PythonPackage):
+    """Coarse-Grained SMILES (CGsmiles)"""
+
+    homepage = "https://github.com/gruenewald-lab/CGsmiles#"
+    pypi = "cgsmiles/cgsmiles-1.0.0.tar.gz"
+    git = "https://github.com/gruenewald-lab/CGsmiles.git"
+
+    license("Apache-2.0")
+    maintainers("adamwitmer")
+
+    version("1.0.0", sha256="83fdb6dadfc4efa065fb8ef66af5d461ec50a629680ba6faf117877f37174aa9")
+    version("cg_smiles_polyply", commit="f53efae4e7dee69e63c51a69cb220e4a5f30b68f")
+
+    depends_on("py-networkx@2.0:", type=("build", "run"))
+    depends_on("py-setuptools@46.4.0:", type="build")
+    depends_on("py-pbr", type="build")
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/py_cgsmiles/package.py
```diff
@@ -0,0 +1,27 @@
...
+
+    homepage = "https://github.com/gruenewald-lab/CGsmiles#"
+    pypi = "cgsmiles/cgsmiles-1.0.0.tar.gz"
+    git = "https://github.com/gruenewald-lab/CGsmiles.git"
+
+    license("Apache-2.0")
+    maintainers("adamwitmer")
+
+    version("1.0.0", sha256="83fdb6dadfc4efa065fb8ef66af5d461ec50a629680ba6faf117877f37174aa9")
+    version("cg_smiles_polyply", commit="f53efae4e7dee69e63c51a69cb220e4a5f30b68f")
```
wdconinc: Named versions are a bit unpredictable in ordering. Do you still need this version-by-commit which is also in the other version?

wdconinc (commented):
Looks good. One comment on the named version. Also, 1.0.1 is released now.

repos/spack_repo/builtin/packages/py_cgsmiles/package.py
bernhardkaindl: > Named versions are a bit unpredictable in ordering. Do you still need this version-by-commit which is also in the other version?

Hi @adamwitmer, could you answer this question for the review or remove this version?

bernhardkaindl: @adamwitmer, may I convert this PR into a draft PR, or can you remove or explain why to add the named version?

adamwitmer: I have removed the named version and added the two new versions, i.e. `1.0.1` and `1.0.2`

bernhardkaindl:
@adamwitmer, I assume the older versions @1.0.0 and @1.0.1 are obsolete and do not need to be added with the new package, so they could be removed before adding the package?

adamwitmer:
> I assume the older versions @1.0.0 and @1.0.1 are obsolete and do not need to be added with the new package, so they could be removed before adding the package?

@wdconinc may be better able to answer this question. I can modify the package versions as they see fit.

wdconinc:
> > I assume the older versions @1.0.0 and @1.0.1 are obsolete and do not need to be added with the new package, so they could be removed before adding the package?
> 
> @wdconinc may be better able to answer this question. I can modify the package versions as they see fit. 

We typically only add the latest patch release per minor version, but we keep the older patch versions in place (in line with the deprecation policy) when a newer patch version is added in a later PR. So, here, I'd suggest not adding the older patch versions (even though they were added gradually) since the PR has not been merged yet and these versions haven't been in a develop branch commit yet.

adamwitmer:
@bernhardkaindl I have removed old versions and updated the `pypi` file path accordingly.

bernhardkaindl (approved):

PR #2472, opened 2025-11-19
valgrind: add version 3.26.0

albestro:


repos/spack_repo/builtin/packages/valgrind/package.py
```diff
@@ -29,6 +29,7 @@ class Valgrind(AutotoolsPackage, SourcewarePackage):
     license("GPL-2.0-or-later")
 
     version("develop", branch="master")
+    version("3.26.0", sha256="8d54c717029106f1644aadaf802ab9692e53d93dd015cbd19e74190eba616bd7")
```
RMeli, line 33: The license changed to GPL 3 with this version. (The `license` directive support `when=`.)

RMeli (changes requested):

repos/spack_repo/builtin/packages/valgrind/package.py
albestro: good catch! thanks @RMeli!

RMeli (approved):

PR #2476, opened 2025-11-19
PDI: fix MPI direct dependency

tpadioleau:
This PR adds a missing direct dependency for the plugins decl-hdf5 and decl-netcdf, see CMakeLists.txt [here](https://github.com/pdidev/pdi/blob/a1345591a799218d6f7666d772ce78306eaef632/plugins/decl_hdf5/CMakeLists.txt#L65) and [here](https://github.com/pdidev/pdi/blob/a1345591a799218d6f7666d772ce78306eaef632/plugins/decl_netcdf/CMakeLists.txt#L43).

It seems to need to be at least of type `build` and `link` to make it work, `link` only did not work for me. I don't know if it needs `run` as in the plugin mpi, see [here](https://github.com/spack/spack-packages/blob/119680aeee8ea802c6111b7167583bddef97e82f/repos/spack_repo/builtin/packages/pdiplugin_mpi/package.py#L32).


tpadioleau:
cc @benoitmartin88

jbigot:
Looks good to me!

jbigot (approved):
Looks great, thanks for the fix!

alalazo (approved):

PR #2482, opened 2025-11-19
precice: Fix compilation of v2.0.0 with GCC 14+

fsimonis:
This PR patches a compilation error for preCICE 2.0.0 which appeared with GCC 14.

### DIFF
--- a/repos/spack_repo/builtin/packages/precice/package.py
+++ b/repos/spack_repo/builtin/packages/precice/package.py
@@ -125,6 +125,13 @@ class Precice(CMakePackage):
     conflicts("%clang@:3.7")
     conflicts("%intel@:16")
 
+    # Fixes missing #include<array> in src/mesh/Edge.hpp
+    patch(
+        "edge-incomplete-type.patch",
+        when="@2.0",
+        sha256="4017a89e4f77f623807a6cd057d9a095788879310f1bddd98837920d252b1ac7",
+    )
+
     def xsdk_tpl_args(self):
         return [
             "-DTPL_ENABLE_BOOST:BOOL=ON",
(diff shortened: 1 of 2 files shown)


repos/spack_repo/builtin/packages/precice/package.py
```diff
@@ -125,6 +125,8 @@ class Precice(CMakePackage):
     conflicts("%clang@:3.7")
     conflicts("%intel@:16")
 
+    patch("edge-incomplete-type.patch", when="@2.0.0")
```
MakisH: The patch was [first released in v2.1.0](https://github.com/precice/precice/commit/a0d05f4e7f030837ca98181f786067bc793deae9), but there have also been v2.0.1 and v2.0.2 in the meantime, which are also affected. How about:

```suggestion
    patch("edge-incomplete-type.patch", when="@2.0")
```

This [should cover all v2.0.x](https://spack.readthedocs.io/en/latest/packaging_guide_creation.html#specifying-version-constraints).

MakisH (commented):
The patch looks correct, we could have a slightly wider version coverage.

MakisH (approved):

MakisH (approved):

repos/spack_repo/builtin/packages/precice/package.py
```diff
@@ -125,6 +125,12 @@ class Precice(CMakePackage):
     conflicts("%clang@:3.7")
     conflicts("%intel@:16")
 
+    patch(
```
alalazo, line 129: Can you add a brief comment above the `patch` directive explaining why it is needed? It could be be as simple as:
```suggestion
    # Add missing '#include<array>'
    patch(
```

That would be useful for future reference when reading the package.

fsimonis: @alalazo Done

alalazo (approved):

MakisH (approved):

PR #2487, opened 2025-11-19
libfabric: Add variants for supported sanitizers

darrylabbate:

### DIFF
--- a/repos/spack_repo/builtin/packages/libfabric/package.py
+++ b/repos/spack_repo/builtin/packages/libfabric/package.py
@@ -114,6 +114,16 @@ class Libfabric(AutotoolsPackage, CudaPackage, ROCmPackage):
     variant("level_zero", default=False, description="Enable Level Zero support")
     variant("gdrcopy", default=False, when="@1.12: +cuda", description="Enable gdrcopy support")
 
+    variant("asan", default=False, when="@1.12:", description="Enable AddressSanitizer (ASan)")
+    variant("lsan", default=False, when="@1.20:", description="Enable LeakSanitizer (LSan)")
+    variant("tsan", default=False, when="@1.20:", description="Enable ThreadSanitizer (TSan)")
+    variant(
+        "ubsan",
+        default=False,
+        when="@1.20:",
+        description="Enable UndefinedBehaviorSanitizer (UBSan)",
+    )
+
     # Backporting from main for versions 2.3.x
     # The CXI provider hardcodes CXIP_FI_VERSION to FI_VERSION(2, 2).
     # Make it match the libfabric we're building
@@ -170,6 +180,8 @@ class Libfabric(AutotoolsPackage, CudaPackage, ROCmPackage):
         "set by OPX, resulting in undefined behavior.",
     )
 
+    conflicts("+asan +tsan")
+
     flag_handler = build_system_flags
 
     @classmethod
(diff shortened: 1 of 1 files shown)


msimberg (commented):
Nice! Do you know if they can be enabled at the same time? From memory they don't all work together typically, but things might have changed since I last used them.

I suppose these are not likely to be used other than in ad-hoc debugging environments so not super critical to get the conflicts right, but if there's a clear answer might be worth adding some conflicts.

darrylabbate:
> Do you know if they can be enabled at the same time? From memory they don't all work together typically, but things might have changed since I last used them.

My understanding is ASan, MSan and TSan are mutually exclusive. LSan and UBSan are usually interoperable with other sanitizers. I'm not able to find an official source on this, though. I didn't want to introduce unnecessary complexity by making assumptions about compatibility.

We don't currently guard these options in Libfabric behind any kind of compatibility matrix: https://github.com/ofiwg/libfabric/blob/285abde881eebe182d83c84e42babc3d3d6ad399/configure.ac#L170-L197

The (now archived) [google/santizers wiki](https://github.com/google/sanitizers/wiki/AddressSanitizer#faq) only explicitly mentions GCC and Clang's implementations of ASan are incompatible with each other. There's an incomplete page on incompatibility here as well: https://github.com/google/sanitizers/wiki/AddressSanitizerIncompatiblity

The compiler should produce an error when incompatible sanitizers are enabled, but I don't think that works with Spack's concretization.

I think I'd be ok with adding an explicit `conflict` between ASan and TSan. What do you think?

msimberg:
> > Do you know if they can be enabled at the same time? From memory they don't all work together typically, but things might have changed since I last used them.
> 
> My understanding is ASan, MSan and TSan are mutually exclusive. LSan and UBSan are usually interoperable with other sanitizers. I'm not able to find an official source on this, though. I didn't want to introduce unnecessary complexity by making assumptions about compatibility.
> 
> We don't currently guard these options in Libfabric behind any kind of compatibility matrix: https://github.com/ofiwg/libfabric/blob/285abde881eebe182d83c84e42babc3d3d6ad399/configure.ac#L170-L197
> 
> The (now archived) [google/santizers wiki](https://github.com/google/sanitizers/wiki/AddressSanitizer#faq) only explicitly mentions GCC and Clang's implementations of ASan are incompatible with each other. There's an incomplete page on incompatibility here as well: https://github.com/google/sanitizers/wiki/AddressSanitizerIncompatiblity
> 
> The compiler should produce an error when incompatible sanitizers are enabled, but I don't think that works with Spack's concretization.
> 
> I think I'd be ok with adding an explicit `conflict` between ASan and TSan. What do you think?

I think that sounds very fair! Add conflicts for what we know doesn't work, and the rest can be added if/when people encounter them.

repos/spack_repo/builtin/packages/libfabric/package.py
```diff
@@ -170,6 +180,9 @@ class Libfabric(AutotoolsPackage, CudaPackage, ROCmPackage):
         "set by OPX, resulting in undefined behavior.",
     )
 
+    conflicts("+asan", when="+tsan")
+    conflicts("+tsan", when="+asan")
```
msimberg: This can also be expressed more simply as
```suggestion
    conflicts("+asan +tsan")
```
if you like.

msimberg (dismissed):
Small suggestion for the conflict, but approved as is. I'll merge in a few days if you prefer to keep the current version of the conflicts.

repos/spack_repo/builtin/packages/libfabric/package.py
darrylabbate: Oh, much cleaner, thanks! I'll update the PR

msimberg (approved):

PR #2491, opened 2025-11-19
asio: add actual C++20 standard to variant

rbberger:
- [x] Question to maintainers: I wasn't sure if I should completely remove `2a`.

### DIFF
--- a/repos/spack_repo/builtin/packages/asio/package.py
+++ b/repos/spack_repo/builtin/packages/asio/package.py
@@ -59,7 +59,7 @@ class Asio(AutotoolsPackage):
         msg="asio v1.22.1 fixed missing includes necessary for gcc v12 and above",
     )
 
-    stds = ("11", "14", "17", "2a", "20")
+    stds = ("11", "14", "17", "20")
     variant(
         "cxxstd",
         default="11",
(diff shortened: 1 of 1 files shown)


msimberg:
I'd be ok with removing `2a` completely. The question is how much breakage it might cause for others, but my guess is little? This is a completely unscientific guess though.

@pauleonix any thoughts?

pauleonix:
I'm fine with it as well. It will break support for GCC 8 and 9 which did not take `-std=c++20` yet afaiu. I would think that anyone who wants C++20 will not settle for those, given the only partial support of C++20.

pauleonix (approved):
LGTM

msimberg:
This looks good to me but looks like spack's audit checks don't like this change being done separately from the HPX changes (https://github.com/spack/spack-packages/pull/2492). https://github.com/spack/spack-packages/actions/runs/19551819876/job/55984972437?pr=2491#step:7:26 shows:
```
1. hpx: wrong variant used for dependency in 'depends_on()'
    return _main(argv)
    invalid values for variant 'cxxstd' in package asio: '2a'
```
It looks like the same, but opposite, error is on the HPX PR.

From my perspective it's ok if you want to merge the two PRs into one (probably into the HPX one). I don't really see another way out of this...

rbberger:
Combined with https://github.com/spack/spack-packages/pull/2492

PR #2499, opened 2025-11-20
rocmcc pre 7.0 might have amdflang in its prefix

scheibelp:
External instances of rocmcc may have amdflang.

`spack compiler find` for rocmcc < `7.0` can now `provides(fortran)` if 

* it finds `amdflang` in the prefix
* the output includes `rocm-x.y.z`

The latter point is not enforced by rocmcc pre-7.0: the official version string is in terms of Clang. This depends on the user installing rocmcc into a prefix that includes that string. 

The user can add `+fortran` themselves to the compiler entry for `< 7.0` in other cases.

### DIFF
--- a/repos/spack_repo/builtin/packages/llvm_amdgpu/package.py
+++ b/repos/spack_repo/builtin/packages/llvm_amdgpu/package.py
@@ -72,7 +72,7 @@ class LlvmAmdgpu(CMakePackage, LlvmDetection, CompilerPackage):
         "fortran",
         default=True,
         when="@:6.99",
-        description="External instances of llvm-amdgpu before 7 may provide fortran"
+        description="External instances of llvm-amdgpu before 7 may provide fortran",
     )
 
     variant(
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/llvm_amdgpu/package.py
```diff
@@ -66,6 +66,14 @@ class LlvmAmdgpu(CMakePackage, LlvmDetection, CompilerPackage):
 
     provides("c", "cxx")
     provides("fortran", when="@7.0:")
+    provides("fortran", when="+fortran")
+
+    variant(
+        "fortran",
+        default=True,
```
afzpatel: Default should be false since spack-built llvm-amdgpu doesn't provide fortran.
```suggestion
        default=False,
```

afzpatel:
@spackbot rerun pipeline

afzpatel:
https://gitlab.spack.io/spack/spack-packages/-/jobs/19552053
e4s-rocm-external-generate is failing with: `1. Multiple externals share the same spec: llvm-amdgpu@6.4.3+fortran`

Maybe adding `fortran: /opt/rocm-6.4.3/llvm/bin/amdflang` might fix it?
https://github.com/spack/spack-packages/blob/develop/stacks/e4s-rocm-external/spack.yaml#L139

PR #2500, opened 2025-11-20
lielab: new package

msparapa:
Adds new recipe: lielab@0.5.1

I might need some help with this. Locally, this does not build for me because of an issue with eigen@5.0.0 not building. I'm still new to spack and not good at debugging yet. I'm hoping the CI will give more details.


msparapa:
@spackbot fix style

msparapa:
Related issue #2514

msparapa:
@spackbot fix style

msparapa:
With #2514 fixed, I was able to update the recipe and test locally. This should be good to go now.

Also I'm just now reading that I shouldn't have been rebasing as often as I was. My bad.

msparapa:
Commenting for visibility. This MR is ready for review but I don't know who to ping about it.

repos/spack_repo/builtin/packages/lielab/package.py
```diff
@@ -0,0 +1,51 @@
...
+from spack.package import *
+
+
+class Lielab(CMakePackage):
+    """Lielab is a C++ library for numerical Lie-theory: Lie groups, Lie algebras, homogeneous
+    manifolds, and various functions and algorithms on these spaces."""
+
+    homepage = "https://github.com/sandialabs/Lielab"
+    maintainers("msparapa")
+    license("MIT")
```
alecbcs, line 19:
```suggestion
    homepage = "https://github.com/sandialabs/Lielab"
    url = "https://github.com/sandialabs/Lielab/archive/refs/tags/v0.5.2.tar.gz"
    
    maintainers("msparapa")
    
    license("MIT")
```

We can set the url as a property on the class and Spack is smart enough to find and replace the version string for each version.

repos/spack_repo/builtin/packages/lielab/package.py
```diff
@@ -0,0 +1,51 @@
...
+
+    version("0.5.2", sha256="d98ddf93fa317165891b69944c5ffab48c3955fd7c1c9428b06a0452f8fca453")
+    version("0.5.1", sha256="5a7545a675f630418634d9827e8db5035949bf8ae165f17600c03bf5a6da35af")
+
+    def url_for_version(self, version):
+        return f"https://github.com/sandialabs/Lielab/archive/refs/tags/v{version}.tar.gz"
```
alecbcs:
```suggestion
```

Suggestion to use the url property instead of a method here since the url is pretty standard.

repos/spack_repo/builtin/packages/lielab/package.py
```diff
@@ -0,0 +1,51 @@
...
+    variant("pic", default=True, description="Position independent code (-fPIC)")
+    variant(
+        "cxxstd",
+        default="20",
+        values=("20", "23", "26"),
+        multi=False,
+        sticky=True,
+        description="C++ standard",
+    )
+    variant("with_assertions", default=True, description="Build with assertions included.")
```
alecbcs, line 33: Variants should be placed after `version()` and before `depends_on()` directives.

repos/spack_repo/builtin/packages/lielab/package.py
```diff
@@ -0,0 +1,51 @@
...
+
+    def cmake_args(self):
+        args = []
+        args.append("-DLIELAB_INSTALL_LIBRARY=TRUE")
+        args.append("-DLIELAB_BUILD_TESTS=FALSE")
+        args.append("-DLIELAB_BUILD_PYTHON=FALSE")
```
alecbcs: Instead of writing `-DLIELAB_*` here we can instead make use of Spack's built in `define()` method.

E.g.
```python3
args.append(self.define("LIELAB_INSTALL_LIBRARY", True)
```

alecbcs (changes requested):
A couple quick suggestions, otherwise looks good to me.

Nice job on your first PR @msparapa!

msparapa:
Added the changes. Thank you!

alecbcs (dismissed):
Looks great! Thanks @msparapa

msparapa:
@spackbot fix style

msparapa:
@alecbcs I dunno if the gitlab CI was stuck because of formatting but 4-5 days seems like a long time. So I fixed style and reset approvals. I probably just made it take even longer 🤷.

bernhardkaindl (approved):
Existing review as concluded successfully, and I built both versions on Ubuntu/x86

PR #2504, opened 2025-11-20
protobuf: new version and add patch for aarch64 fix

green-br:
Added new version 33.1 (which includes fix) and a patch for aarch64 build issue at version 32.

Maintainer is: @hyoklee

### DIFF
--- a/repos/spack_repo/builtin/packages/protobuf/package.py
+++ b/repos/spack_repo/builtin/packages/protobuf/package.py
@@ -16,6 +16,7 @@ class Protobuf(CMakePackage):
 
     license("BSD-3-Clause")
 
+    version("33.1", sha256="0c98bb704ceb4e68c92f93907951ca3c36130bc73f87264e8c0771a80362ac97")
     version("32.1", sha256="d2081ab9528292f7980ef2d88d2be472453eea4222141046ad4f660874d5f24e")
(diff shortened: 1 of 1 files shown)


green-br:
Just spotted #2294 which is looking at versions and versioning scheme - patch would be nice for aarch64.

albestro:
@green-br considering your changes, I think this can go in independently:
- 33.1 respects the correct versioning that we are trying to fix in #2294 
- the patch targets version 32, which already respects the right versioning on `develop`

repos/spack_repo/builtin/packages/protobuf/package.py
```diff
@@ -142,6 +143,14 @@ class Protobuf(CMakePackage):
         sha256="c061356db31cdce29c8cdd98a3a8219ef048ebc2318d0dec26c1f2c5e5dae29b",
     )
 
+    # Broken build on aarch64 due to __builtin_arm_crc32cd
+    # See https://github.com/protocolbuffers/protobuf/pull/23164
```
albestro, line 147: My journey in fixing protobuf started exactly from this problem 😅
I wanted to add 33 which solves the problem, but then I realized the inconsistency in the versioning...and you know the rest 😁

Thanks for adding the patch!

albestro (approved):
Let's give @hyoklee some more time to review it.

hyoklee (approved):

PR #2506, opened 2025-11-20
lmod: add v8.7.60 - 8.7.67

afuetterer:
Hi,
I am unsure if all these versions are needed, or just the 8.7.67. What do you think?

After suggesting a faulty PR about snakemake once, which broke the installation, I did not include the major version jump to v9.0 in this PR. I tested the package with version v8.7.67 locally and it worked fine.

Please let me know how to proceed. Should I remove 8.7.60-8.7.66?

There is also 9.0, 9.0.1-4.
Ref: https://github.com/TACC/Lmod/releases


alecbcs (dismissed):
Looks good to me. @afuetterer I think we're fine adding all of these versions if you'd like them / need to use any of them in addition to 8.7.60. If you have no use for the intermediate versions, I'd vote to just add the latest.

afuetterer:
I agree with just adding the latest version. I don't need a version from the ones in between.

I don't know how exactly the deprecation policy works, but adding only one version, means fewer deprecations in the future.

Let me remove those other versions then.

afuetterer:
Please feel free to squash the commits. Or should I do this and force-push?

alecbcs (approved):
Looks good to me. I'll squash merge the commits.

PR #2481, opened 2025-11-19
Boost: fix pre 1.70.0 container tree insertion

fsimonis:
This PR patches a compilation issue with Boost.container tree-based data structures including `flat_set` and `flat_map`.

It was reported for version 1.66.0 in https://github.com/boostorg/container/issues/100.
The fix https://github.com/boostorg/container/pull/101 was released in 1.70.0.

Issue #2055, opened 2025-10-22
Installation issue: boost@1.48.0


aeblyve:
### Steps to reproduce the issue

```console
# spack spec -I boost@1.48.0
 -   boost@1.48.0~atomic~chrono~clanglibcpp~container~context~contract~coroutine~date_time~debug~exception~fiber~filesystem~graph~graph_parallel~icu~iostreams~json~locale~log~math~mpi~mqtt5+multithreaded~nowide~numpy~pic~program_options~python~random~regex~serialization+shared~signals~signals2~singlethreaded~stacktrace~system~taggedlayout~test~thread~timer~type_erasure~url~versionedlayout~wave build_system=generic cxxstd=11 patches:=4f0f7c0 visibility=hidden arch=linux-almalinux9-skylake_avx512 %c,cxx=gcc@11.5.0
[+]      ^compiler-wrapper@1.0 build_system=generic arch=linux-almalinux9-skylake_avx512 
[e]      ^gcc@11.5.0~binutils+bootstrap~graphite~nvptx~piclibs~profiled~strip build_system=autotools build_type=RelWithDebInfo languages:='c,c++' arch=linux-almalinux9-skylake_avx512 
[+]      ^gcc-runtime@11.5.0 build_system=generic arch=linux-almalinux9-skylake_avx512 
[e]      ^glibc@2.34 build_system=autotools arch=linux-almalinux9-skylake_avx512 


```

Simply attempt `spack install boost@1.48.0` or `spack install boost@1.48.0~system`. I am new to Spack, but I believe this may be an issue of the older package not supporting the variant specifications of the newer package. 

### Error message

<details><summary>Error message</summary>

<pre>
# spack install boost@1.48.0~system
[+] /root/spack/opt/spack/linux-skylake_avx512/compiler-wrapper-1.0-a2teqwnwwxwlpcp4luebo2eydkrlom2j
[+] /usr (external gcc-11.5.0-5b5rum4ucarf2zbjomqbslkpdgfjeo6j)
[+] /usr (external glibc-2.34-lpyethlr76auzvqcu6y6hfr4u5tycyjq)
[+] /root/spack/opt/spack/linux-skylake_avx512/gcc-runtime-11.5.0-mqobnxqeimx5ql2dg42yhqoqknc63aqf
==> No binary for boost-1.48.0-j2saci5nbqduv4r2s4v4o3f5npapksah found: installing from source
[...]

PR #2508, opened 2025-11-20
boost: deprecate versions before 1.70.0

alalazo:
v1.70.0 was released April 9, 2019


alalazo:
@spackbot fix style

hainest:
Generally, I would be happy to drop support for these ancient versions. However, recent activity (e.g., #2055, #2481) indicates that folks are still using them, so I'd rather not break their workflows unnecessarily. Of course, _any_ deprecation is likely to break someone's code.

@aeblyve @fsimonis How much extra work would it be to add `--deprecated` to your workflows?

fsimonis:
@hainest Adding a flag is trivial.
The deprecation however means that the workflow will stop working at some point in the future.

@alalazo Is it really necessary to abandon these versions? Boost is arguably one of the most used C++ library collections out there.
The versatility of Spack to build old versions especially for generating develop and runtime environments for legacy software makes it a very valuable tool. 

Maybe it would be a good idea to do some kind of user poll to check how many people rely on these builds.

Of course, it is always possible to use an older version of Spack.

alalazo:
> Maybe it would be a good idea to do some kind of user poll to check how many people rely on these builds.

The PR is here for that. I'm not merging it if it is controversial, and we can also adjust the cutoff.

On the other hand, I bet very old versions of software are usually unmaintained (not talking only about boost here), so they are unlikely to build fine. Also, more versions mean a larger search space for the concretizer - so in a sense we always pay _some_ price for having a ton of versions of widely used software.

fsimonis:
> The PR is here for that.

For such a busy project as Spack, PRs are visible to internal developers at best. Some package maintainers may stumble across this. I strongly doubt any external user of Spack will see this until their workflows start failing.

alalazo:
> For such a busy project as Spack, PRs are visible to internal developers at best.

Here it was the maintainer of the package that raised a flag, not an internal developer :thinking: 

> Some package maintainers may stumble across this. I strongly doubt any external user of Spack will see this until their workflows start failing.

Not sure what to take from your comment, besides this PR. Are you suggesting that we should never deprecate a version, or that we should poll for each package version deprecation? That would not scale given the number of packages we have.

For reference, our current guidelines for deprecation are [here](https://spack.readthedocs.io/en/latest/packaging_guide_creation.html#deprecate), happy to take suggestions to improve the workflow.

fsimonis:
> Not sure what to take from your comment

The only point I am trying to make is that a PR is not visible to users of the package. (Reply to `The PR is here for that.` https://github.com/spack/spack-packages/pull/2508#issuecomment-3563047835)

If you really want user feedback, you could tag all maintainers of packages depending on boost, send a notification in a suitable Slack channel, or merge this PR as a test to develop and see who complains.

alalazo:
6 months later, is there any version of boost we can deprecate? :slightly_smiling_face:

climbfuji (approved):
Approving for boost, ecflow, yaml_cpp

hainest:
@alalazo I think 1.70.0 is unnecessarily high. In my [research](https://github.com/hainest/spack-packages/commit/dbb0cd1f082a9b6f34641e4796a26b865150e278), I've found that 1.50.0 has a strong motivation for being the minimum version. That also allows bumping the minimum c++ version to 11 (https://github.com/hainest/spack-packages/commit/4a29759548aa1552a5ecc77bcead0bc3c03378c1). The only thing currently holding up those changes is a bug in the concretizer. I should really chat with you about that so we can fix a lot of Boost issues with my package rewrite.

viniciusvgp (approved):
Approving for simgrid

alalazo:
@hainest Cutting at 1.50.0 doesn't seem to buy us much, while cutting at 1.70.0 would simplify the recipe a lot in the longer term (i.e. when we actually remove versions). I have an analysis that can be a base for discussion [here](https://gist.github.com/alalazo/5ec75b5023e6ffa54fab696553e0f226). Its summary is more or less:


| | Remove ≤ 1.50.0 | Remove ≤ 1.70.0 |
|---|---|---|
| Versions removed | 13 | ~31 |
| Packages broken | 0 | 18 (all unmaintained, stalled version caps) |
| Library variants removed | 0 | 2 (`signals`, `coroutine` — discontinued by Boost itself) |
|  `when` clauses removed | 3 | 11 |
| Patches removed | 0 | 12 |
| Build logic branches simplified | 4 | 10 |

The 1.50.0 cut leaves the recipe essentially unchanged — the same variants, the same version-guarded branches, the same patches for old bugs. The 1.70.0 cut removes a lot of that, and the 18 packages that have at least a version requiring a `boost` in the cutoff are either unmaintained, or very old versions that were not deprecated yet.

A final consideration: for users who genuinely need very old Boost versions we could start another repository, called `legacy`. There we can extend `builtin` recipes via Python inheritance and support very old versions of widely used packages, if need be - so what's cleaned in `builtin.Boost` in 1 year or so may go in `legacy.Boost`.

fsimonis (approved):
Approving for precice

@alalazo Is the plan to create a release with the deprecations before removing the old boost versions?

alalazo:
Yes. I'd say deprecation for `boost` are under discussion with @hainest and won't happen in v1.2 so deletion will take at least ~1 year (first deprecation in the Nov 2026 release, then deletion in June 2027 release).

MakisH (approved):
Following-up https://github.com/spack/spack-packages/pull/2508#pullrequestreview-4449427205, approving for `precice` (and, by extension, `py_pyprecice`).

PR #2509, opened 2025-11-20
amdlibflame: deprecate versions before v4.2

alalazo:


alalazo:
@amd-toolchain-support

amd-toolchain-support:
@alalazo, could you assist me in understanding the purpose behind this change?

We have several other packages that are part of AOCL (https://www.amd.com/en/developer/aocl.html), such as amdlibm, amdfftw, amdblis, aocl-da, and a meta package amd-aocl that uses the same versioning.

alalazo:
@amd-toolchain-support It was just to deprecate old versions of widely used virtual providers, so we could clean up the recipes in a year or so. Do I understand correctly that these kind of deprecations for AMD packages should be done for all these packages at once?

bernhardkaindl:
@amd-toolchain-support, alalazo had a question for you here: https://github.com/spack/spack-packages/pull/2509#issuecomment-3670830656

PR #2520, opened 2025-11-20
py_vermin: add v1.8.0

netromdk:
I have released version [1.8.0](https://github.com/netromdk/vermin/releases/tag/v1.8.0) of Vermin 🎉 

@alalazo

PR #2521, opened 2025-11-20
ci: update vermin to v1.8.0

netromdk:
After having released version [1.8.0](https://github.com/netromdk/vermin/releases/tag/v1.8.0) of Vermin, it would make sense to also use it for the Spack workflows.

Relates to #2520.
@alalazo


alecbcs (approved):
Looks good to me. Thanks @netromdk!

Issue #2518, opened 2025-11-20
Installation issue: lua-ffi build fails


ebagrenrut:
### Steps to reproduce the issue

@trws (Because you're maintainer) @haampie (Because you last touched the package) @alalazo (Because you last touched the lua build system) @kwryankrattiger (because you're in my timezone :D):

I suspect that there is a deficiency in the lua build system when a Lua-based package is being isntalled from a commit and has an `archive_file` of `None`. See [this line](https://github.com/spack/spack-packages/blob/c75e10845431600b163c597545bd099e427c62f5/repos/spack_repo/builtin/build_systems/lua.py#L66) in `build_systems/lua.py`.

To reproduce:

In a brand-new Spack instance with no cached packages:

```console
$ spack env create test
$ spack env activate test
$ spack add lua-ffi
$ spack install
```

Now, if I run `spack install` a _second_ time in the same env, it works because the first `spack install` resulted in an archive file being created.

This is particularly an issue for me when trying to build `lua-ffi` in a Spack CI-generated `spack ci rebuild` job as there is no cached archive.

I wonder if the line in `LuaBuilder.unpack` should be something like:
```
    if pkg.stage.archive_file is not None and os.path.splitext(pkg.stage.archive_file)[1] == ".rock":
```

### Error message

<details><summary>Error message</summary>

<pre>
==> No binary for lua-ffi-0.0.1.fakever-wkizx3sgawgpayqsl5jqdemu6ttkbkec found: installing from source
==> Installing lua-ffi-0.0.1.fakever-wkizx3sgawgpayqsl5jqdemu6ttkbkec [22/22]
==> No patches needed for lua-ffi
==> lua-ffi: Executing phase: 'unpack'
==> Error: TypeError: expected str, bytes or os.PathLike object, not NoneType

/root/.spack/package_repos/fncqgg4/repos/spack_repo/builtin/build_systems/lua.py:66, in unpack:
         65    def unpack(self, pkg: LuaPackage, spec: Spec, prefix: Prefix) -> None:
  >>     66        if os.path.splitext(pkg.stage.archive_file)[1] == ".rock":
         67            directory = pkg.luarocks("unpack", pkg.stage.archive_file, output=str)
         68            dirlines = directory.split("\n")
         69            # TODO: figure out how to scope this better

See build log for details:
[...]

PR #2523, opened 2025-11-20
Fix LuaBuilder.unpack for non-existant archive_file

ebagrenrut:
`LuaBuilder.unpack` attempts to create a Path object of `pkg.stage.archive_file` without checking whether or not the value is `None`. When a Lua package, such as `lua-ffi`, is installed from a git commit, there is no archive file to unpack and an exception is thrown. Reference error shown in https://github.com/spack/spack-packages/issues/2518


ebagrenrut:
@spackbot style

ebagrenrut:
@spackbot fix style

becker33 (approved):

PR #2526, opened 2025-11-20
Add dep on make when the compiler is gcc

kwryankrattiger:
This is not meant to be a long term solution, but when there is no system installed make these packages fail to build.

### DIFF
--- a/repos/spack_repo/builtin/packages/pcre2/package.py
+++ b/repos/spack_repo/builtin/packages/pcre2/package.py
@@ -46,6 +46,8 @@ class Pcre2(AutotoolsPackage, CMakePackage):
 
     depends_on("c", type="build")
 
+    depends_on("gmake", type="build", when="%gcc")
+
     with when("build_system=cmake"):
         depends_on("zlib")
         depends_on("bzip2")
--- a/repos/spack_repo/builtin/packages/pcre2/package.py
+++ b/repos/spack_repo/builtin/packages/pcre2/package.py
@@ -46,8 +46,6 @@ class Pcre2(AutotoolsPackage, CMakePackage):
 
     depends_on("c", type="build")
 
-    depends_on("gmake", type="build", when="%gcc")
-
     with when("build_system=cmake"):
         depends_on("zlib")
         depends_on("bzip2")
--- a/repos/spack_repo/builtin/packages/pkgconf/package.py
+++ b/repos/spack_repo/builtin/packages/pkgconf/package.py
@@ -47,8 +47,6 @@ class Pkgconf(AutotoolsPackage):
 
     executables = ["^pkgconf$", "^pkg-config$"]
 
-    depends_on("gmake", type="build", when="%gcc")
-
     @classmethod
     def determine_version(cls, exe):
         exe = Executable(exe)
--- a/repos/spack_repo/builtin/packages/pkgconf/package.py
+++ b/repos/spack_repo/builtin/packages/pkgconf/package.py
@@ -47,6 +47,8 @@ class Pkgconf(AutotoolsPackage):
 
     executables = ["^pkgconf$", "^pkg-config$"]
 
+    depends_on("gmake", type="build", when="%gcc")
+
     @classmethod
     def determine_version(cls, exe):
         exe = Executable(exe)
--- a/repos/spack_repo/builtin/packages/py_matplotlib/package.py
+++ b/repos/spack_repo/builtin/packages/py_matplotlib/package.py
@@ -144,6 +144,9 @@ class PyMatplotlib(PythonPackage):
     depends_on("c", type="build")
     depends_on("cxx", type="build")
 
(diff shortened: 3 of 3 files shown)


rbberger:
`make` -> `gmake`?

adamjstewart:
Why GCC? Is it not needed for all compilers?

alalazo (dismissed):
Why is this needed? `AutotoolsPackage` and `CMakePackage` already specify when `gmake` is a dependency 🤔

kwryankrattiger:
> Why GCC? Is it not needed for all compilers?

@adamjstewart 
GCC brings in gmake under the hood, which if it is not in the path it will use the system gmake. If you have an old system gmake and use a newer gmake in your spack environment Spack will pass `fifo` pipes around and then old gmake will choke on it.

> Why is this needed?

@alalazo 
Excellent question. The CMake projects I wonder if I was getting `ninja`? My build environment/logs are gone. I will try to reproduce these again. I found these while using the new installer, running in Ubuntu 22 containers, system installed make was available, but `gmake@4.4:` was required. It has been very reliable to reproduce.

For sure matplotlib needs this, it isn't getting it from anywhere.

adamjstewart:
Will have to see some kind of stacktrace. I don't see make used anywhere in matplotlib, they use meson which uses ninja.

kwryankrattiger:
[spack-build-out.txt](https://github.com/user-attachments/files/23837466/spack-build-out.txt)

Here is the error out of the matplotlib build. I have gmake@4.4 required in my environment, yet I get the fifo pipe error.

<details>

```
  FAILED: [code=1] src/_tkagg.cpython-312-x86_64-linux-gnu.so
  /opt/spack/opt/spack/linux-skylake/compiler-wrapper-1.0-yhvkgqt6xoi2uyegjjslmzkg3wyiefzj/libexec/spack/gcc/g++  -o src/_tkagg.cpython-312-x86_64-linux-gnu.so src/_tkagg.cpython-312-x86_64-linux-gnu.so.p/_tkagg.cpp.o -flto -Wl,--as-needed -Wl,--allow-shlib-undefined -Wl,-O1 -shared -fPIC
  make: *** internal error: invalid --jobserver-auth string 'fifo:/tmp/tmplu3nts_n/jobserver_fifo'.  Stop.
  lto-wrapper: fatal error: make returned 2 exit status
  compilation terminated
```

</details>

adamjstewart:
Very strange. The matplotlib source code doesn't mention make or jobserver. Are we sure this isn't a flag injected by Spack? I know Spack recently added jobserver support for parallel builds. It's also in the middle of a `ninja -j10` run. Does ninja call out to make on occasion?

rgommers:
Ninja 1.13 (about half a year old) added jobserver support: https://ninja-build.org/manual.html#_gnu_jobserver_support

I think the correct fix here is to add a dependency on `gmake>=4.4` in the `package.py` for Ninja itself.

kwryankrattiger:
I will try this with ninja having a gmake dep, I know ninja add jobserver support, but I thought it implemented it's own interaction with the job server and didn't use make at all.

rbberger:
See https://github.com/spack/spack/pull/51444 for context.

rbberger:
@adamjstewart it's gcc with LTO that is causing this, since it internally generates a Makefile. Ninja/Meson have nothing to do with this directly.

@kwryankrattiger you're using the new installer with this, right? Since it seems like the revert mentioned above talks about moving the fifo jobserver support to the experimental installer.

alalazo:
> it's gcc with LTO that is causing this, since it internally generates a Makefile. Ninja/Meson have nothing to do with this directly.

:+1:

haampie (changes requested):
* Can you restrict to matplotlib, given the others already have the dependency
* Add a comment, otherwise this dependency is mysterious. You're adding it because `gcc` runs its `lto-wrapper` executable during the build, which generates a Makefile for parallelism, and external GCC lacks a `run` dependency on `gmake`.
* You *don't* need a lowerbound on the `gmake` version. Spack just need to know the gmake version and passes the appropriate `MAKEFLAGS= --jobserver-auth=fifo:PATH` if new, `MAKEFLAGS= --jobserver-auth:R,W` if old, `MAKEFLAGS= --jobserver-fds=R,W` if ancient.

This doesn't have anything to do with ninja.

You could also add `depends_on("gmake", type="run")` to `gcc` while you're at it, but that doesn't help with external GCC.

kwryankrattiger:
> You could also add depends_on("gmake", type="run") to gcc while you're at it, but that doesn't help with external GCC.

I think that compiler requirement makes it so that will almost never fix the issue unfortunately.

kwryankrattiger:
@spackbot run pipeline

haampie:
Can you add a comment for future reference? Otherwise people will be puzzled later. And something to grep in case we fix deps for externals over at spack/spack.

haampie (approved):

PR #2536, opened 2025-11-21
py-blis: new version 1.3.3

qwertos:

### DIFF
--- a/repos/spack_repo/builtin/packages/py_blis/package.py
+++ b/repos/spack_repo/builtin/packages/py_blis/package.py
@@ -17,15 +17,19 @@ class PyBlis(PythonPackage):
 
     license("BSD-3-Clause")
 
+    version("1.3.3", sha256="034d4560ff3cc43e8aa37e188451b0440e3261d989bb8a42ceee865607715ecd")
     version("0.9.1", sha256="7ceac466801f9d97ecb34e10dded8c24cf5e0927ea7e834da1cc9d2ed3fc366f")
     version("0.7.9", sha256="29ef4c25007785a90ffc2f0ab3d3bd3b75cd2d7856a9a482b7d0dac8d511a09d")
     version("0.4.1", sha256="d69257d317e86f34a7f230a2fd1f021fd2a1b944137f40d8cdbb23bd334cd0c4")
 
     depends_on("c", type="build")  # generated
 
+    depends_on("python@3.9:3.14", when="@1.3.3:", type=("build", "run"))
     depends_on("py-setuptools", type="build")
-    depends_on("py-cython@0.25:", when="@0.7.9:", type="build")
-    depends_on("py-numpy@1.15:", type=("build", "run"))
+    depends_on("py-cython@0.25:", when="@0.7.9:0.9.1", type="build")
+    depends_on("py-cython@3", when="@1.3.3:", type="build")
+    depends_on("py-numpy@1.15:", when="@:0.9.1", type=("build", "run"))
+    depends_on("py-numpy@2:", when="@1.3.3:", type=("build", "run"))
 
     def setup_build_environment(self, env: EnvironmentModifications) -> None:
         env.set("BLIS_COMPILER", spack_cc)
--- a/repos/spack_repo/builtin/packages/py_blis/package.py
+++ b/repos/spack_repo/builtin/packages/py_blis/package.py
@@ -24,6 +24,7 @@ class PyBlis(PythonPackage):
 
     depends_on("c", type="build")  # generated
 
+    depends_on("python@:3.14", when="@:0.9.1", type=("build", "run"))
     depends_on("python@3.9:3.14", when="@1.3.3:", type=("build", "run"))
     depends_on("py-setuptools", type="build")
     depends_on("py-cython@0.25:", when="@0.7.9:0.9.1", type="build")


repos/spack_repo/builtin/packages/py_blis/package.py
```diff
@@ -17,15 +17,19 @@ class PyBlis(PythonPackage):
...
     license("BSD-3-Clause")
 
+    version("1.3.3", sha256="034d4560ff3cc43e8aa37e188451b0440e3261d989bb8a42ceee865607715ecd")
     version("0.9.1", sha256="7ceac466801f9d97ecb34e10dded8c24cf5e0927ea7e834da1cc9d2ed3fc366f")
     version("0.7.9", sha256="29ef4c25007785a90ffc2f0ab3d3bd3b75cd2d7856a9a482b7d0dac8d511a09d")
     version("0.4.1", sha256="d69257d317e86f34a7f230a2fd1f021fd2a1b944137f40d8cdbb23bd334cd0c4")
 
     depends_on("c", type="build")  # generated
 
+    depends_on("python@3.9:3.14", when="@1.3.3:", type=("build", "run"))
```
alecbcs, line 28: We should add another `depends_on()` directive below this line specifying the maximum allowed python version prior to v1.3.3. Otherwise earlier versions will be allowed with python 3.14 / 3.15.

repos/spack_repo/builtin/packages/py_blis/package.py
```diff
@@ -17,15 +17,19 @@ class PyBlis(PythonPackage):
...
     depends_on("py-setuptools", type="build")
-    depends_on("py-cython@0.25:", when="@0.7.9:", type="build")
-    depends_on("py-numpy@1.15:", type=("build", "run"))
+    depends_on("py-cython@0.25:", when="@0.7.9:0.9.1", type="build")
+    depends_on("py-cython@3", when="@1.3.3:", type="build")
+    depends_on("py-numpy@1.15:", when="@:0.9.1", type=("build", "run"))
+    depends_on("py-numpy@2:", when="@1.3.3:", type=("build", "run"))
```
alecbcs, line 33: We should flip the ordering of these so that constraints that apply to newer versions come before older ones. (Mirroring the ordering from version directives.)

alecbcs:
Two quick nitpicks. Otherwise looks good to me. Thanks @qwertos!

repos/spack_repo/builtin/packages/py_blis/package.py
qwertos: Please no. That makes it so much more difficult to read and understand what's going on when updating, especially when it comes to more complex packages. If consistency with version directives is important, let's swap the order of version directives.

alecbcs: This is already the accepted standard in a majority of other `py-*` packages, and is the plan moving forward. 

I sympathize that it's a bit less intuitive to read when inspecting dependencies, but we can't at this point flip the ordering of the version directives that's been in place and used by the community for years. @adamjstewart started reordering dependencies to follow the version directives and I think it's the right idea to unify the formats.

adamjstewart: > That makes it so much more difficult to read and understand what's going on when updating, especially when it comes to more complex packages.

Can you elaborate on this? Because this is exactly why I like to sort `depends_on` in reverse order. The first numpy I see is the latest version, and if that's outdated, then I immediately know to update it.

qwertos: When I'm adding a new version to a package, I like to process and interpret all of the `depends_on` directives and see how my additions interact with what's already there. (Especially when it comes to version ranges that are open ended.)

Its the difference between seeing this:

```python
depends_on("foo@a:b", when="@m:n")
depends_on("foo@c:d", when="@o:p")
depends_on("foo@e:f", when="@q:r")
depends_on("foo@g:h", when="@s:t")
```

and being able to interpret it as:

```
m < n < o < p < q < r < s < t
```

vs having to wrangle my brain around:

```
( m < n ) > ( o < p ) > ( q < r ) > ( s < t )
```

Hopefully that all makes sense, describing how thoughts work is hard.

As far as me knowing what dependencies I need to update, I add the requirements to the package I'm working on and run `spack spec`. If I get an error something needs to be updated.

adamjstewart: No that makes perfect sense to me. It's not a bad argument. I would personally be okay with either if there was some way to enforce it via linters. While I personally prefer my order because it matches how we handle versions, we could also change how we handle versions.

adamjstewart: Until we figure out how to lint things like this, I think we should leave it up to the PR author or package maintainer to decide. I try to leave things in the order they are already in when editing existing packages.

qwertos: For some packages I think it makes sense to have dependencies listed newest first. Namely the rare packages where each version has a dedicated `with default_args(when="@n"):` (where `n` is not a range). 

At the end of the day, dependency wrangling can be tricky business (as I know you are well aware), and formatting of the `package.py` file is a tool in the toolbox to help make sense of the chaos. So long as the `depends_on` directives are internally consistent within a package, IMHO author/maintainer discretion is the way to go.

kshea21 (approved):
Discussion seems to have ended with the conclusion that the `depends_on` ordering can remain as the author intended and we can proceed with merging the PR.

PR #2541, opened 2025-11-21
py-typer-slim: new package (and discussion)

qwertos:
What is the best way to handle this package? Per [the readme][1] (what would be) `py-typer-slim+standard` is the same as the package `py-typer`. 

Options: 

1. Accept this package and have 2 packages that need to be updated and maintained, and potentially creating weird conflicts down the line as both import with `import typer`.
2. Update `py-typer` to have `provides("py-typer-slim")`, potentially adding unnecessary dependencies when `depends_on("py-typer-slim")` is specified.
3. When packaging dependent packages, specify `depends_on("py-typer")` instead of `depends_on("py-typer-slim")`.
4. Get rid of `py-typer` entirely and replace with `py-typer-slim` with a variant `+standard` and possibly `provides("py-typer", when="+standard")`

Personally, I lean towards option 2 even though this PR implements option 1, but I wanted to have some sort of discussion before I made that change. I'm happy to implement whatever is decided.

[1]: https://github.com/fastapi/typer/blob/master/README.md#typer-slim

### DIFF
--- a/repos/spack_repo/builtin/packages/py_typer_slim/package.py
+++ b/repos/spack_repo/builtin/packages/py_typer_slim/package.py
@@ -1,0 +1,23 @@
+# Copyright Spack Project Developers. See COPYRIGHT file for details.
+#
+# SPDX-License-Identifier: (Apache-2.0 OR MIT)
+
+from spack_repo.builtin.build_systems.python import PythonPackage
+
+from spack.package import *
+
+
+class PyTyperSlim(PythonPackage):
+    """Typer, build great CLIs. Easy to code. Based on Python type hints."""
+
+    homepage = "https://github.com/fastapi/typer"
+    pypi = "typer_slim/typer_slim-0.20.0.tar.gz"
+
+    license("MIT")
+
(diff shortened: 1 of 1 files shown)


qwertos:
bump

bernhardkaindl (changes requested):
Hi @qwertos,

I didn't find the reason why you want a new "slim" provides for `py-typer` yet.

I could not understand why your slim provides can't be done using a variant can't be done using e.g. a "`slim`" variant of `py-typer+slim`, but I'd like to advocate against such variant as it doesn't clearly define what it enables or disables.

It would likely be better for https://github.com/spack/spack-packages/blob/develop/repos/spack_repo/builtin/packages/py_typer/package.py to have variants which are enabled by default and which you can disable when you need a build with e.g. fewer dependencies, like it would be normal for spack packages.

These enabled-by-default variants (one per dependency) would enable the dependency when the variant is enabled as usual.

As you didn't add any further explanation and the actual package already exists, I'm considering to close this PR as adding a duplicate package which the official spack does not need. You can always add your own additional spack packages repo if you want a special customisation like this that isn't yet in the main spack-packages repo.

PR #2546, opened 2025-11-22
celeritas: new version v0.6.3

sethrj:


wdconinc (approved):
Looks good to me. Do we build with CUDA in CI? I think we do, right?

sethrj:
The HEP CI still seems to have celeritas+cuda+vecgeom disabled, but it has celeritas+rocm. We internally have CUDA building with CI.

PR #2550, opened 2025-11-24
intel-oneapi-mkl: Model optional SYCL dependency

mabraham:
Also example uses in GROMACS and HeFFTe

### DIFF
--- a/repos/spack_repo/builtin/packages/intel_oneapi_mkl/package.py
+++ b/repos/spack_repo/builtin/packages/intel_oneapi_mkl/package.py
@@ -211,7 +211,7 @@ class IntelOneapiMkl(IntelOneApiLibraryPackage):
     # For a downstream package to depend on this one and link to shared
     # objects from MKL that depend on SYCL, the linker must be able to find
     # the libsycl.so that came from the compiler that originally built MKL.
-    # The mappings in _versions ensures the spack solver can only choose
+    # This mapping ensures the spack solver can only choose
     # compatible MKL and compiler (and thus runtime) versions. Up to at least
     # 2026.0, that mapping is one-to-one, i.e MKL requires the compiler that
     # was in the same oneAPI release, but this may change when the SYCL ABI
--- a/repos/spack_repo/builtin/packages/intel_oneapi_mkl/package.py
+++ b/repos/spack_repo/builtin/packages/intel_oneapi_mkl/package.py
@@ -12,186 +12,6 @@ from spack_repo.builtin.build_systems.oneapi import (
 
 from spack.package import *
 
-_versions = {
-    "2025.3.0": {
-        "version_kwargs": {
-            "url": "https://registrationcenter-download.intel.com/akdlm/IRC_NAS/2ad98b49-1fb2-4294-ab3d-6889b434ebd3/intel-onemkl-2025.3.0.462_offline.sh",
-            "sha256": "6d0152abbfb85c51abb47bbf8d21711cb43c409a0e673d67970dbe6cf20c1970",
-            "expand": False,
-        },
-        "compatible_sycl_compiler_versions": ["%oneapi@2025.3"],
-    },
-    "2025.2.0": {
-        "version_kwargs": {
-            "url": "https://registrationcenter-download.intel.com/akdlm/IRC_NAS/47c7d946-fca1-441a-b0df-b094e3f045ea/intel-onemkl-2025.2.0.629_offline.sh",
-            "sha256": "7bfdde1379a9cd3c2784af4352931a0527a5c483c9da71e851a3f06e055af7c7",
-            "expand": False,
-        },
-        "compatible_sycl_compiler_versions": ["%oneapi@2025.2"],
-    },
-    "2025.1.0": {
-        "version_kwargs": {
-            "url": "https://registrationcenter-download.intel.com/akdlm/IRC_NAS/dc93af13-2b3f-40c3-a41b-2bc05a707a80/intel-onemkl-2025.1.0.803_offline.sh",
-            "sha256": "80a4b1338b48b3fbee55a8dc784f92e5e88d618f1b99d80f5f207a00c86a6638",
-            "expand": False,
-        },
-        "compatible_sycl_compiler_versions": ["%oneapi@2025.1"],
-    },
-    "2025.0.1": {
-        "version_kwargs": {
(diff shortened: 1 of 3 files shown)


repos/spack_repo/builtin/packages/intel_oneapi_mkl/package.py
```diff
@@ -196,6 +196,40 @@ class IntelOneapiMkl(IntelOneApiLibraryPackage):
...
+        description="Use SYCL support - requires oneAPI SYCL runtime from matching compiler",
+    )
+    # Let spack work out that a dependency on the oneAPI runtime is required
+    depends_on("sycl", when="+sycl")
+    # Ensure the solver can't try hipSYCL
+    conflicts("hipsycl", when="+sycl")
+    # Ensure the solver enforces the one-to-one mapping between versions of MKL libraries
+    # that depend on SYCL to the SYCL runtime library from the compiler that was in the
+    # same oneAPI release. This satisfies the linking requirements of the MKL shared objects.
+    depends_on("intel-oneapi-runtime@2025.3", when="@2025.3 +sycl", type=("run"))
```
mabraham: I see that CI doesn't like these dependencies. Perhaps it (or me) is confused about the status of `intel-oneapi-runtime`.

I think we should have something like these dependencies, because exactly one `libsycl.so` can satisfy the link-time dependencies of e.g. `libmkl_sycl.so` and that's the one distributed with the `intel-oneapi-compiler` from the same oneAPI release (which is also the one used to make the `intel-oneapi-runtime`. Users of `intel-oneapi-mkl` shouldn't have to rely on getting this right by hard-coding the right MKL version or having only one version of oneAPI available in spack. It looks to me like maybe the checker doesn't know how to create `intel-oneapi-runtime`, or similar.

Any ideas how I can do better, please?

mabraham: I don't see these errors with spack when run locally, but I do see the mapping honored correctly.

repos/spack_repo/builtin/packages/intel_oneapi_mkl/package.py
```diff
@@ -196,6 +196,40 @@ class IntelOneapiMkl(IntelOneApiLibraryPackage):
...
+    depends_on("intel-oneapi-runtime@2023.1", when="@2023.1 +sycl", type=("run"))
+    depends_on("intel-oneapi-runtime@2023.0", when="@2023.0 +sycl", type=("run"))
+    depends_on("intel-oneapi-runtime@2022.2", when="@2022.2 +sycl", type=("run"))
+    depends_on("intel-oneapi-runtime@2022.1", when="@2022.1 +sycl", type=("run"))
+    depends_on("intel-oneapi-runtime@2022.0", when="@2022.0 +sycl", type=("run"))
+    depends_on("intel-oneapi-runtime@2021.4", when="@2021.4 +sycl", type=("run"))
+    depends_on("intel-oneapi-runtime@2021.3", when="@2021.3 +sycl", type=("run"))
+    depends_on("intel-oneapi-runtime@2021.2", when="@2021.2 +sycl", type=("run"))
+    depends_on("intel-oneapi-runtime@2021.1", when="@2021.1 +sycl", type=("run"))
+    depends_on("intel-oneapi-runtime@2021.0", when="@2021.0 +sycl", type=("run"))
```
alalazo: These dependencies are wrong (runtimes are meant to be injected by compilers). Can you explain your use case better?

alalazo (changes requested):

repos/spack_repo/builtin/packages/intel_oneapi_mkl/package.py
mabraham: To use the SYCL support in MKL the linker must be able to find the libsycl.so that came from the matching oneAPI compiler package that built MKL. I'll try modeling with an explicit compiler.

mabraham: For example `gromacs %oneapi +sycl ^intel-oneapi-mkl +sycl` should work and only by constructing a dag where the oneapi compiler, the runtime injected by it, and the MKL package have a common oneAPI version, so that when gromacs links with SYCL aspects of MKL, the linker can also find a matching `libsycl.so`

mabraham:
@spackbot fix style

repos/spack_repo/builtin/packages/intel_oneapi_mkl/package.py
mabraham: OK that seems to have worked for me locally. Without the requirements the "preferred" MKL version is chosen, and with them the compatible version is chosen.

repos/spack_repo/builtin/packages/intel_oneapi_mkl/package.py
```diff
@@ -196,6 +196,45 @@ class IntelOneapiMkl(IntelOneApiLibraryPackage):
...
+        description="Use SYCL support - requires oneAPI SYCL runtime from matching compiler",
+    )
+    # Let spack work out that a dependency on the oneAPI runtime is required
+    depends_on("sycl", when="+sycl")
+    # Ensure the solver can't try hipSYCL
+    conflicts("hipsycl", when="+sycl")
+    # For a downstream package to depend on this one and link to shared
+    # objects from MKL that depend on SYCL, the linker must be able to find
+    # the libsycl.so that came from the compiler that originally built MKL.
+    # The mappings in _versions ensures the spack solver can only choose
```
mabraham:
```suggestion
    # This mapping ensures the spack solver can only choose
```

repos/spack_repo/builtin/packages/intel_oneapi_mkl/package.py
mabraham: I think this PR is in good working order now!

mkstoyanov (approved):

repos/spack_repo/builtin/packages/intel_oneapi_mkl/package.py
mabraham: Any further feedback, please @alalazo ?

mabraham:
Any further thoughts please @alalazo ?

alalazo:
I'll try to get to this in the next couple of weeks

mabraham:
I'm still keen for this :-)

PR #2560, opened 2025-11-25
pmix: fix dependencies for v6.0.0

thomas-bouvier:
I don't understand why some dependencies were tied to `@master`.

### DIFF
--- a/repos/spack_repo/builtin/packages/pmix/package.py
+++ b/repos/spack_repo/builtin/packages/pmix/package.py
@@ -71,29 +71,33 @@ class Pmix(AutotoolsPackage):
 
     depends_on("c", type="build")
     depends_on("pkgconfig", type="build")
-    depends_on("m4", type="build", when="@master")
-    depends_on("autoconf@2.69:", type="build", when="@master")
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/pmix/package.py
```diff
@@ -71,30 +71,34 @@ class Pmix(AutotoolsPackage):
 
     depends_on("c", type="build")
     depends_on("pkgconfig", type="build")
-    depends_on("m4", type="build", when="@master")
-    depends_on("autoconf@2.69:", type="build", when="@master")
-    depends_on("automake@1.13.4:", type="build", when="@master")
-    depends_on("libtool@2.4.2:", type="build", when="@master")
-    depends_on("flex@2.5.39:", type="build", when="@master")
-    depends_on("perl", type="build", when="@master")
```
alecbcs, line 79: These dependencies were likely tied to master because they're needed to run `reconfigure` which is needed to generate the configure script when pulling from git, but is already rendered when pulling the release tarball.

thomas-bouvier: Thanks for the clarification.

PR #2561, opened 2025-11-25
libint: add v2.11.2

hfp:


alecbcs (approved):
Looks good to me. Thanks @hfp!

PR #2564, opened 2025-11-25
libgdsii: depends on c

becker33:
Discovered in a debugging session with @wadudmiah

Adding to backports even though it adds a dependency because the package is unusable without it.


repos/spack_repo/builtin/packages/libgdsii/package.py
```diff
@@ -20,7 +20,8 @@ class Libgdsii(AutotoolsPackage):
 
     version("0.21", sha256="1adc571c6b53df4c08d108f9ac4f4a7fd6fbefd4bc56f74e0b7b2801353671b8")
 
-    depends_on("cxx", type="build")  # generated
+    depends_on("c", type"build")
```
alecbcs:
```suggestion
    depends_on("c", type="build")
```

alecbcs (changes requested):
Typo. Otherwise looks good to me.

alecbcs (approved):
Looks good to me. Thanks!

PR #2566, opened 2025-11-25
fd: add gmake as build dependency

alecbcs:
Add `gmake` as a build dependency because it is used and leaked from system installs in the build.


ashermancinelli (approved):
Thanks!

alalazo (approved):

PR #2567, opened 2025-11-25
libctl: build system tests for F77

becker33:
It doesn't appear to include any fortran code, but the libctl build system fails without fortran present.

Found in a debugging session with @wadudmiah


alecbcs (approved):
Looks good to me. Thanks!

PR #2568, opened 2025-11-25
mpb: depends_on fortran and cflags update

becker33:
Fails in configure without fortran

Fails at build time with gcc14 without `-Wno-int-conversion`

### DIFF
--- a/repos/spack_repo/builtin/packages/mpb/package.py
+++ b/repos/spack_repo/builtin/packages/mpb/package.py
@@ -18,7 +18,8 @@ class Mpb(AutotoolsPackage):
 
     version("1.11.1", sha256="7311fc525214c1184cad3e0626b8540c0b53b3c31c28e61ce6ec2860088eca46")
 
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/mpb/package.py
```diff
@@ -30,6 +31,11 @@ class Mpb(AutotoolsPackage):
     depends_on("hdf5")
     depends_on("libctl")
 
+    def flag_handler(self, name, flags):
+        if self.spec.satisfies("%c=gcc@14:") and name == "cflags":
+            flags.append("-Wno-int-conversion")
```
alecbcs: It might be less computationally intensive to switch the ordering of these two conditions.

becker33: Amdahl's law probably says don't bother when the build takes minutes, but I fixed it anyway.

alecbcs (approved):

PR #2569, opened 2025-11-25
meep: depends on python@:3.11

becker33:
meep uses imp which is removed in python 3.12.

Found in a debugging session with @wadudmiah


alecbcs (approved):
Looks good to me. Thanks!

PR #2316, opened 2025-11-06
(py-)onnx: add v1.18.0, v1.19.1

wdconinc:
This PR adds newer versions of `onnx` and `py-onnx`, in particular v1.18.0, v1.19.0, v1.19.1. Added all point versions since some ML dependencies are very specific. The patch is specific to onnxruntime.

PR #2570, opened 2025-11-25
onnx: conflict with abseil-cpp cxxstd=11,14

wdconinc:
As of 1.19 ([commit](https://github.com/onnx/onnx/commit/51df8fd8e66dfc446b8216a7ec081e3f2b7dd2ef)), `onnx` requires C++17 or better. Since `onnx` gets its `CMAKE_CXX_STANDARD` standard from `abseil-cpp` (via `protobuf`), this PR adds a conflict.

### DIFF
--- a/repos/spack_repo/builtin/packages/onnx/package.py
+++ b/repos/spack_repo/builtin/packages/onnx/package.py
@@ -69,7 +69,7 @@ class Onnx(CMakePackage):
     depends_on("cmake@3.26:", type="build", when="@1.20:")
     depends_on("python", type="build")
     depends_on("protobuf")
-    
+
     # ONNX uses the cxxstd from abseil-cpp, which is a dependency of protobuf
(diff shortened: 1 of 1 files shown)


wdconinc:
@paulgessinger Can you check if this resolves the issue you were referring to in https://github.com/spack/spack-packages/pull/2316#issuecomment-3574817753 ? It may need a similar treatment in `protobuf`...

paulgessinger:
@wdconinc I'm not able to reproduce this anymore at this time, so I'm not sure I can test this PR anymore.

wdconinc:
> @wdconinc I'm not able to reproduce this anymore at this time, so I'm not sure I can test this PR anymore.

Ok, maybe we can close it. What do you think?

paulgessinger:
@wdconinc Yes, I'll raise this again if I run into the same issue.

PR #2573, opened 2025-11-26
TensorFlow: fix build on macOS Tahoe

adamjstewart:
With this PR, I'm finally able to build TF on macOS for the first time in years! JAX will need the same treatment in a separate PR. This will help immensely with debugging #38.

### DIFF
--- a/repos/spack_repo/builtin/packages/py_tensorflow/package.py
+++ b/repos/spack_repo/builtin/packages/py_tensorflow/package.py
@@ -401,6 +401,14 @@ class PyTensorflow(Package, CudaPackage, ROCmPackage, PythonExtension):
     # https://github.com/tensorflow/tensorflow/issues/62416
     conflicts("%clang@17:", when="@:2.14")
 
+    # https://github.com/tensorflow/tensorflow/issues/103590
+    patch(
+        "https://github.com/tensorflow/tensorflow/pull/104948.patch?full_index=1",
+        sha256="93314b87dc9960aa5320f672dbe559a4ad95ba4f35c2ddca6094d169947f7c78",
+        when="@2.20",
+    )
+    conflicts("os=tahoe", when="@:2.19")
+
     # Fix build error with CUDA
     patch(
         "https://github.com/tensorflow/tensorflow/pull/99046.patch?full_index=1",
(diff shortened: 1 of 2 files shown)


repos/spack_repo/builtin/packages/py_tensorflow/package.py
```diff
@@ -401,6 +401,14 @@ class PyTensorflow(Package, CudaPackage, ROCmPackage, PythonExtension):
     # https://github.com/tensorflow/tensorflow/issues/62416
     conflicts("%clang@17:", when="@:2.14")
 
+    # https://github.com/tensorflow/tensorflow/issues/103590
+    patch(
+        "https://github.com/tensorflow/tensorflow/pull/104948.patch?full_index=1",
+        sha256="93314b87dc9960aa5320f672dbe559a4ad95ba4f35c2ddca6094d169947f7c78",
```
alecbcs, line 407:
```suggestion
        "https://github.com/tensorflow/tensorflow/commit/53d6f2a.patch?full_index=1",
        sha256="93314b87dc9960aa5320f672dbe559a4ad95ba4f35c2ddca6094d169947f7c78",
```

Although the PR that is referenced is closed, I'd like us to get out of the habit of including patches from PRs since reviewers might accept a patch from a PR that is still open and which could later create an installation issue via a change in checksum. I'm instead referencing the merge commit here.

alecbcs (changes requested):
One quick nitpick surrounding the patch used. Otherwise looks good to me.

repos/spack_repo/builtin/packages/py_tensorflow/package.py
adamjstewart: I 100% agree with you about open PRs, but slightly disagree about closed PRs. The URL for PR patches is much simpler, and the PR may contain multiple commits, even after merge.

If you want to push forward this policy, I would rather 1) mass update all existing packages (only 16) and 2) add a check to the package audit (we already do this to ensure `?full_index=1`).

alecbcs: I looked at mass updating all existing packages and then decided against it before the v1.1 release.

The note about a PR containing multiple commits after merge is a good one. I still believe it'll be too hard to train reviewers to ensure PRs are closed before we accept a patch, but I'll investigate if we'd able to validate PRs are merged with a CI check.

adamjstewart: So is this PR safe to merge until we vote on a different policy for PR patches?

alecbcs: Yep. I'll push on making a CI check for these. In the meantime this makes sense to keep.

alecbcs (approved):

PR #2574, opened 2025-11-26
flint: add version 3.4.0

d-torrance:

### DIFF
--- a/repos/spack_repo/builtin/packages/flint/package.py
+++ b/repos/spack_repo/builtin/packages/flint/package.py
@@ -19,6 +19,7 @@ class Flint(AutotoolsPackage):
     license("LGPL-2.1-or-later")
 
     version("main", branch="main")
+    version("3.4.0", sha256="9497679804dead926e3affeb8d4c58739d1c7684d60c2c12827550d28e454a33")
     version("3.1.2", sha256="fdb3a431a37464834acff3bdc145f4fe8d0f951dd5327c4c6f93f4cbac5c2700")
     version("3.0.1", sha256="7b311a00503a863881eb8177dbeb84322f29399f3d7d72f3b1a4c9ba1d5794b4")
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/flint/package.py
alecbcs:
```suggestion
    url = "https://flintlib.org/download/flint-3.1.2.tar.gz"
```

Looks like flint moved the download location on their servers for the source tarballs.

repos/spack_repo/builtin/packages/flint/package.py
```diff
@@ -11,14 +11,15 @@ class Flint(AutotoolsPackage):
     """FLINT (Fast Library for Number Theory)."""
 
     homepage = "https://flintlib.org"
-    url = "https://flintlib.org/flint-3.1.2.tar.gz"
+    url = "https://flintlib.org/download/flint-3.1.2.tar.gz"
```
alalazo, line 14: Why was this changed? I was able to retrieve 3.4.0 with the old url.

alalazo: Ugh, didn't notice @alecbcs was already reviewing. I'll leave the review to him, sorry for the noise.

alecbcs: No worries! I just jumped in when I saw the CI check was failing and started investigating that.

Curious that you were able to retrieve v3.4.0 with the old url. Both GitHub Actions and I get 404 codes when attempting to download from https://flintlib.org/flint-3.4.0.tar.gz.

Since the authors seemed to migrate all of the old releases I figured updating the url was cleaner than writing a `url_for_version()` method.

alalazo: I tried:
```console
spack versions ...
```
which probably uses `list_url`

alecbcs (approved):
Looks good to me. Thanks @d-torrance for the quick turnaround!

PR #2580, opened 2025-11-26
Fix static builds in libceed

Sbozzolo:
`libceed` did not have the `shared` variant and it was not possible to compile `libceed` as part of a static build. The reasons for this are two:

1. `libceed` has a build flag `STATIC `that has to turned on to build `libceed` as a static library
2. the dependencies of `libceed` were not compiled as static libraries (unless forced to).

This resulted in errors like:
```cpp
relocation R_X86_64_PC32 against symbol `magma2lapack_constants' can not be used when making a shared object; recompile with -fPIC
```
when building `libceed` as part of a static build.

### DIFF
--- a/repos/spack_repo/builtin/packages/libceed/package.py
+++ b/repos/spack_repo/builtin/packages/libceed/package.py
@@ -41,6 +41,8 @@ class Libceed(MakefilePackage, CudaPackage, ROCmPackage):
 
     variant("openmp", default=False, description="Enable OpenMP support")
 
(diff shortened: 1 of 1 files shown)


alalazo (approved):

alalazo:
Waiting for maintainers to review this too. @v-dobrev feel free to merge if the PR seems good already

Sbozzolo:
cc @jeremylt

jeremylt (approved):
LGTM

alalazo (approved):

PR #2587, opened 2025-11-27
acts: rename "tgeo" variant to "root"

stephenswat:
The "tgeo" variant for the ACTS package no longer clearly describes its intention. For this reason, I want to slowly rename it to ROOT. This commit starts this process by establishing a "root" alias for "tgeo", the latter of which we can phase out slowly.

Pinging @paulgessinger who requested this change.

### DIFF
--- a/repos/spack_repo/builtin/packages/acts/package.py
+++ b/repos/spack_repo/builtin/packages/acts/package.py
@@ -344,7 +344,6 @@ class Acts(CMakePackage, CudaPackage):
         conflicts("~root", when="+tgeo")
         conflicts("+root", when="~tgeo")
 
-
     variant("traccc", default=False, description="Build the Traccc plugin", when="@35.1:")
 
     # Variants that only affect Acts examples for now
--- a/repos/spack_repo/builtin/packages/acts/package.py
+++ b/repos/spack_repo/builtin/packages/acts/package.py
@@ -328,7 +328,7 @@ class Acts(CMakePackage, CudaPackage):
     # names as synonyms, ensuring that both must be true at the same time. We
     # also enforce that nothing explicitly relies on the TGeo naming anymore,
     # and we use ROOT instead. We then "deprecate" the TGeo naming by
-    # eliminating it in ACTS release 45. Finally, we retain the TGeo naming 
+    # eliminating it in ACTS release 45. Finally, we retain the TGeo naming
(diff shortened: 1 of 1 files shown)


wdconinc (dismissed):
Looks good to me.

I appreciate trying to make this backwards compatible, but the drawback is that users with `acts +tgeo` in their specs will not encounter an error but will remain silently stuck on the versions 44. Maybe this can be rolled out as you have it in this PR, but with a tty.warning that `tgeo` has been renamed. Then at some later point (maybe after the next spack-packages release) the `tgeo` variant can be removed entirely.

repos/spack_repo/builtin/packages/acts/package.py
```diff
@@ -322,10 +322,29 @@ class Acts(CMakePackage, CudaPackage):
...
+        "root", default=False, description="Build the ROOT plugin", when="@:34 +identification"
+    )
+    variant("root", default=False, description="Build the ROOT plugin", when="@35:")
+    # Establish a mutual implication between the tgeo and root variants; if
+    # one is enabled, so must be the other.
+    with when("@:44"):
+        conflicts("~root", when="+tgeo")
+        conflicts("+root", when="~tgeo")
+
+
```
wdconinc:
```suggestion
```

wdconinc (dismissed):

repos/spack_repo/builtin/packages/acts/package.py
```diff
@@ -322,10 +322,28 @@ class Acts(CMakePackage, CudaPackage):
         when="@19.3:",
     )
     variant("sycl", default=False, description="Build the SyCL plugin", when="@1:34")
+
+    # The TGeo and ROOT variants are synonyms, and the goal is to slowly phase
+    # out the TGeo name. The plan for this is as follow. First, we use both
+    # names as synonyms, ensuring that both must be true at the same time. We
+    # also enforce that nothing explicitly relies on the TGeo naming anymore,
+    # and we use ROOT instead. We then "deprecate" the TGeo naming by
+    # eliminating it in ACTS release 45. Finally, we retain the TGeo naming
```
wdconinc:
```suggestion
    # eliminating it in ACTS release 45. Finally, we retain the TGeo naming
```

wdconinc (approved):

PR #2588, opened 2025-11-27
py-eigenpy: init at 3.12.0

nim65s:
Hi,

This add https://github.com/stack-of-tasks/eigenpy following the nice spack tutorial from @RemiLacroix-IDRIS, thanks !

### DIFF
--- a/repos/spack_repo/builtin/packages/py_eigenpy/package.py
+++ b/repos/spack_repo/builtin/packages/py_eigenpy/package.py
@@ -1,0 +1,53 @@
+# Copyright Spack Project Developers. See COPYRIGHT file for details.
+#
+# SPDX-License-Identifier: (Apache-2.0 OR MIT)
+
+from spack_repo.builtin.build_systems.cmake import CMakePackage
+
+from spack.package import *
+
+
+class PyEigenpy(CMakePackage):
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/py_eigenpy/package.py
```diff
@@ -0,0 +1,53 @@
...
+
+class PyEigenpy(CMakePackage):
+    """Efficient bindings between Numpy and Eigen using Boost.Python"""
+
+    homepage = "https://github.com/stack-of-tasks/eigenpy"
+    git = "https://github.com/stack-of-tasks/eigenpy.git"
+
+    def url_for_version(self, version):
+        return f"https://github.com/stack-of-tasks/eigenpy/archive/refs/tags/v{version}.tar.gz"
+
```
tldahlgren: Please move this to around line 49 so it is the first method in the package and appears *after* the properties and directives.

tldahlgren (changes requested):
Confirmed the homepage, branch and version sha256. Please move the method so it appears below the properties and directives.

tldahlgren (commented):
~I'm seeing `pyproject.toml` in the repository. Why are you extending python instead of inheriting from it?~

tldahlgren (approved):
Thanks!

PR #2592, opened 2025-11-28
xtensor: fix constraints

gouarin:
The constraints set in the package definition are not well defined. 

What we have before was:

```py
    depends_on("xtl@0.8.0:0.8", when="@0.26.0:")
    depends_on("xtl@0.7.5:0.7", when="@0.24.5:")
    depends_on("xtl@0.7.2:0.7", when="@0.23.2:")
```

It seems that if we don't finalize the constraint on the second line as `when="@0.24.5:0.25"`, the `0.8` is not chosen for a package I am developing. This PR resolves the issue.

### DIFF
--- a/repos/spack_repo/builtin/packages/xtensor/package.py
+++ b/repos/spack_repo/builtin/packages/xtensor/package.py
@@ -14,7 +14,7 @@ class Xtensor(CMakePackage):
     url = "https://github.com/QuantStack/xtensor/archive/0.13.1.tar.gz"
     git = "https://github.com/QuantStack/xtensor.git"
 
-    maintainers("ax3l")
+    maintainers("ax3l", "gouarin")
 
     license("BSD-3-Clause")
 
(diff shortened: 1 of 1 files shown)


gouarin:
@spackbot help

gouarin:
@spackbot request review

repos/spack_repo/builtin/packages/xtensor/package.py
```diff
@@ -40,19 +40,19 @@ class Xtensor(CMakePackage):
     depends_on("cxx", type="build")  # generated
 
     depends_on("xtl", when="@develop")
-    depends_on("xtl@0.8.0:0.8", when="@0.26.0:")
-    depends_on("xtl@0.7.5:0.7", when="@0.24.5:")
-    depends_on("xtl@0.7.2:0.7", when="@0.23.2:")
```
alalazo, line 45: Thanks, these were definitely wrong

alalazo (approved):

PR #2593, opened 2025-11-28
highfive: fix url and add new versions

gouarin:
The highfive project changes its github url.

This PR updates the url of the project and adds the version 3 of this package.

### DIFF
--- a/repos/spack_repo/builtin/packages/highfive/package.py
+++ b/repos/spack_repo/builtin/packages/highfive/package.py
@@ -13,7 +13,6 @@ class Highfive(CMakePackage):
     """HighFive - Header only C++ HDF5 interface"""
 
     homepage = "https://github.com/highfive-devs/highfive"
-    url = "https://github.com/highfive-devs/highfive/archive/refs/tags/v3.2.0.tar.gz"
     git = "https://github.com/highfive-devs/highfive.git"
 
     maintainers("1uc", "gouarin")
@@ -70,3 +69,12 @@ class Highfive(CMakePackage):
             self.define("HIGHFIVE_UNIT_TESTS", False),
             self.define("HIGHFIVE_EXAMPLES", False),
         ]
+
+    def url_for_version(self, version):
+        print(help(Version))
+        if version < Version("3.0.0"):
+            url = "https://github.com/BlueBrain/HighFive/archive/v{0}.tar.gz"
+        else:
+            url = "https://github.com/highfive-devs/highfive/archive/v{0}.tar.gz"
+        print("URL **********************************", url.format(version))
+        return url.format(version)
--- a/repos/spack_repo/builtin/packages/highfive/package.py
+++ b/repos/spack_repo/builtin/packages/highfive/package.py
@@ -15,7 +15,7 @@ class Highfive(CMakePackage):
     homepage = "https://github.com/highfive-devs/highfive"
     git = "https://github.com/highfive-devs/highfive.git"
 
-    maintainers("1uc", "gouarin")
+    maintainers("1uc")
 
     license("BSL-1.0")
 
--- a/repos/spack_repo/builtin/packages/highfive/package.py
+++ b/repos/spack_repo/builtin/packages/highfive/package.py
@@ -12,14 +12,19 @@ from spack.package import *
 class Highfive(CMakePackage):
     """HighFive - Header only C++ HDF5 interface"""
 
-    homepage = "https://github.com/BlueBrain/HighFive"
-    url = "https://github.com/BlueBrain/HighFive/archive/v1.2.tar.gz"
-    git = "https://github.com/BlueBrain/HighFive.git"
-    maintainers("1uc")
+    homepage = "https://github.com/highfive-devs/highfive"
+    url = "https://github.com/highfive-devs/highfive/archive/refs/tags/v3.2.0.tar.gz"
+    git = "https://github.com/highfive-devs/highfive.git"
+
+    maintainers("1uc", "gouarin")
 
     license("BSL-1.0")
 
     version("develop", branch="master")
+    version("3.2.0", sha256="01ea2eed7dbce1cf5dfff59476cfa113a7822b641aecbd99c674592fe7a4e630")
+    version("3.1.1", sha256="622034f34badda41255d7793e1c5a3046954dcf0875b0bca076e7c77088a8890")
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/highfive/package.py
```diff
@@ -12,14 +12,19 @@
...
 
-    homepage = "https://github.com/BlueBrain/HighFive"
-    url = "https://github.com/BlueBrain/HighFive/archive/v1.2.tar.gz"
-    git = "https://github.com/BlueBrain/HighFive.git"
-    maintainers("1uc")
+    homepage = "https://github.com/highfive-devs/highfive"
+    url = "https://github.com/highfive-devs/highfive/archive/refs/tags/v3.2.0.tar.gz"
+    git = "https://github.com/highfive-devs/highfive.git"
+
+    maintainers("1uc", "gouarin")
```
1uc: Thank you for looking after the Spack recipe. However, I can't remember working with you nor do I remember reviewing any contributions to HighFive from you. Hence, adding yourself as a maintainer feels maybe a little too bold?

repos/spack_repo/builtin/packages/highfive/package.py
```diff
@@ -12,14 +12,19 @@
...
 
     license("BSL-1.0")
 
     version("develop", branch="master")
+    version("3.2.0", sha256="01ea2eed7dbce1cf5dfff59476cfa113a7822b641aecbd99c674592fe7a4e630")
+    version("3.1.1", sha256="622034f34badda41255d7793e1c5a3046954dcf0875b0bca076e7c77088a8890")
+    version("3.1.0", sha256="8c5aec8621b95b26028e129bb4f38e8c388cb7d1781e9721b1e5f8827b812b3b")
```
1uc: `3.1.0` is an annoying version that has broken version numbers, IIRC it's otherwise identical to `3.1.1`. I think it would be better to not add `3.1.0`.

repos/spack_repo/builtin/packages/highfive/package.py
```diff
@@ -12,14 +12,19 @@
 class Highfive(CMakePackage):
     """HighFive - Header only C++ HDF5 interface"""
 
-    homepage = "https://github.com/BlueBrain/HighFive"
-    url = "https://github.com/BlueBrain/HighFive/archive/v1.2.tar.gz"
-    git = "https://github.com/BlueBrain/HighFive.git"
-    maintainers("1uc")
+    homepage = "https://github.com/highfive-devs/highfive"
+    url = "https://github.com/highfive-devs/highfive/archive/refs/tags/v3.2.0.tar.gz"
+    git = "https://github.com/highfive-devs/highfive.git"
```
1uc, line 16: Will this approach work for v2? Those versions haven't been released under `highfive-devs/highfive`. They only exist under `BlueBrain/highfive`.

I looked at it a while back and it seemed like one solution would be `url_for_version`:
https://spack.readthedocs.io/en/latest/packaging_guide_creation.html#versions-and-urls

This would allow us to point to `BlueBrain/highfive` for everything upto 3; and then to `highfive-devs/highfive` for everything as of 3.

1uc (commented):
Thank you for adding the new version numbers! I'm not convinced it'll work for v2, I'd like to double check that.

repos/spack_repo/builtin/packages/highfive/package.py
gouarin: I applied the rules for being a maintainer of a conda-forge package. I added my name to indicate that I contributed to the Spack package, not the Highfive package. If this is a problem for you or if it doesn't fit the rules for a Spack package maintainer, I can remove my name.

repos/spack_repo/builtin/packages/highfive/package.py
gouarin: I removed this version.

repos/spack_repo/builtin/packages/highfive/package.py
gouarin: Good catch. I have implemented the `url_for_version` method and it works locally.

1uc: Thank you! Are the two printf's important or a leftover from debugging?

I'm trying to test it locally, but I'm running into the problem that it keeps downloading from mirror.spack.io:
```
==> Fetching https://mirror.spack.io/_source-cache/archive/60/60d66ba1315730494470afaf402bb40300a39eb6ef3b9d67263335a236069cce.tar.gz
```
and even the first approach worked. Do you know the trick to convince `spack` to not use mirror.spack.io?

gouarin: It was for debugging purpose. I cleaned up the file ! Sorry about that ...

I don't know if this is the right method but I added a `repos.yaml` file to the `.spack` directory with the path to the `spack-packages` github repo that I cloned, with the following content

```yaml
repos:
  - /path/to/spack-packages/repos/spack_repo/builtin
```

gouarin: Does it answer your question ?

1uc: Not quite. Those are just the steps to use your changes. In order to turn off mirrors.spack.io one needs to edit `~/git/spack/etc/spack/defaults/base/mirrors.yaml`.

I can now see it's downloading from the correct location; and checksums pass.

repos/spack_repo/builtin/packages/highfive/package.py
1uc: Thank you for fixing the recipe. Reading through: https://spack.readthedocs.io/en/latest/roles_and_responsibilities.html#package-contributors

I think Package Contributor describes your situation quite well. Would that work for you?

gouarin: Yes, it works for me. If you want another maintainer for this spack package, you can contact me later.

repos/spack_repo/builtin/packages/highfive/package.py
gouarin: great !

1uc (approved):
Thank you for fixing the recipe.

I checked all new versions and one old version, while manually disabling mirros.spack.io. The all installed ~with~ without error.

gouarin:
What kind of error ?

1uc:
No, error. Just me not writing the correct words. It installs with**out** error.

RikkiButler20 (approved):
Looks great.

PR #2595, opened 2025-11-28
mathicgb: new package

d-torrance:
mathicgb is a program and library for computing Gröbner bases and signature Gröbner bases.  It is a dependency of Macaulay2.


alecbcs (approved):
Looks good to me. Thanks @d-torrance!

PR #2596, opened 2025-11-28
gcc: backport fixincludes fix from GCC 15

msimberg:
GCC generates "fixed" headers that override system headers when it deems the system headers aren't usable (https://github.com/gcc-mirror/gcc/blob/master/fixincludes/README).

However, GCC had a bug in how it detects when the fixes should be generated: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=118009. In our case the manifestation was that GCC generates a "fixed" `pthread.h`, which overrides the system one. This makes GCC dependent on the exact glibc that was used when building (or at least a range of glibc versions). We updated glibc, and GCC that was compiled against an older glibc stopped being able to compile files with the value of `PTHREAD_COND_INITIALIZER` being different in the "fixed" `pthread.h` and the one provided by the newer system glibc. The issue is essentially the same as https://sourceware.org/bugzilla/show_bug.cgi?id=32625.

The broken fixincludes rule was fixed on the GCC 15 release branch, but is not yet in a release (15.2.0 is the latest release at the time of this PR). The patch that fixes the rule is https://github.com/gcc-mirror/gcc/commit/ea2798892de373b14f9fc7ae8a0d820eaddca98c.

In this PR I'm applying the above patch to 15.1 and 15.2. I'm also backporting the patch to 12-14 (line numbers have changed between versions). I'm not applying the patch to 11 since apparently the "fixed" include should actually be used for that version (see GCC commit description). I haven't even attempted to patch this for older versions. If someone needs it backported there I invite them to open a separate PR.

# Alternative

The fact that GCC even attempts to fix includes seems to be a bit controversial (opinionated writing from 2013: https://ewontfix.com/12/; not my text, but gives a bit of useful context around the feature). GCC comes with a `--disable-fixincludes` flag (see e.g. https://www.linuxfromscratch.org/lfs/view/development/chapter08/gcc.html and https://gcc.gnu.org/pipermail/gcc-patches/2022-May/595495.html, there doesn't seem to be any official documentation about it). This is a bit more nuclear, but the spack build of GCC could set `--disable-fixincludes` or introduce a variant to set it.

However, I think `--disable-fixincludes` is something that can be added independently of the patches in this PR if there's interest in it. I'd love to hear others' opinion (and maybe experience) about `--disable-fixincludes` before considering that though. 

# Testing

I've tested building GCC 15.2.0, 14.3.0, 13.4.0, and 12.5.0 on a system where we noticed `pthread.h` was generated. On develop `pthread.h` is indeed generated with all four versions, and with this PR `pthread.h` is not generated.

I've additionally tested "installing" all GCC versions from 12.1.0 to 15.2.0 until the configure stage to check that the patches actually apply cleanly.

### DIFF
--- a/repos/spack_repo/builtin/packages/gcc/package.py
+++ b/repos/spack_repo/builtin/packages/gcc/package.py
@@ -431,8 +431,9 @@ class Gcc(AutotoolsPackage, GNUMirrorPackage, CompilerPackage):
     # the installed GCC not portable across different glibc versions. Original
     # GCC bug report: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=118009. For
     # GCC 15 we can directly use the upstream patch. For GCC 12-14 the patch
-    # has been backported. For GCC 11 the fix should actually be applied and
-    # older versions have not been tested.
+    # has been backported. The patch is not applied to GCC 11 since the "fixinclude"
+    # is in fact needed for that version (see GCC commit description). Older versions
+    # have not been checked or tested.
     patch(
         "https://github.com/gcc-mirror/gcc/commit/ea2798892de373b14f9fc7ae8a0d820eaddca98c.patch?full_index=1",
         sha256="0999dbf856725566373f25a6f192a3520ea036db8e1f31928aae9750e6e38be7",
--- a/repos/spack_repo/builtin/packages/gcc/package.py
+++ b/repos/spack_repo/builtin/packages/gcc/package.py
@@ -425,6 +425,23 @@ class Gcc(AutotoolsPackage, GNUMirrorPackage, CompilerPackage):
     # https://github.com/iains/gcc-12-branch/issues/6
     conflicts("@:12", when="%apple-clang@14:14.0")
 
+    # Applies
+    # https://github.com/gcc-mirror/gcc/commit/ea2798892de373b14f9fc7ae8a0d820eaddca98c,
+    # which fixes an incorrectly applied fixincludes rule for pthread.h, making
+    # the installed GCC not portable across different glibc versions. Original
+    # GCC bug report: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=118009. For
+    # GCC 15 we can directly use the upstream patch. For GCC 12-14 the patch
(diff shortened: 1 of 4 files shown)


repos/spack_repo/builtin/packages/gcc/package.py
```diff
@@ -425,6 +425,23 @@ class Gcc(AutotoolsPackage, GNUMirrorPackage, CompilerPackage):
...
     conflicts("@:12", when="%apple-clang@14:14.0")
 
+    # Applies
+    # https://github.com/gcc-mirror/gcc/commit/ea2798892de373b14f9fc7ae8a0d820eaddca98c,
+    # which fixes an incorrectly applied fixincludes rule for pthread.h, making
+    # the installed GCC not portable across different glibc versions. Original
+    # GCC bug report: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=118009. For
+    # GCC 15 we can directly use the upstream patch. For GCC 12-14 the patch
+    # has been backported. For GCC 11 the fix should actually be applied and
+    # older versions have not been tested.
```
albestro: @msimberg not clear to me what this means, in particular about GCC11.

I mean that 

> I'm not applying the patch to 11 since apparently the "fixed" include should actually be used for that version (see GCC commit description)

is a bit more clearer.

albestro: My confusion comes from the fact that we are talking about a fix of a fix 😅

msimberg: Yeah, thanks for the comment @albestro! I admit that's not very clear :stuck_out_tongue: 

What do you think about this slightly reworded comment from the PR description:
```suggestion
    # has been backported. The patch is not applied to GCC 11 since the "fixinclude"
    # is in fact needed for that version (see GCC commit description). Older versions
    # have not been checked or tested.
```
?

albestro: 🔝

RMeli (approved):
Changes and explanation look sensible to me. I'll let the package maintainers comment.

PR #2599, opened 2025-11-29
gaudi: Allow builds against newer fmt versions

tmadlener:
Gaudi seems to be buildable without problems against `fmt@11`, but not against `fmt@12`. I have tested `gaudi@40.1` and `gaudi@39.4`. It's possible that even older versions build against `fmt@11`, but I could not as easily test them currently.

### DIFF
--- a/repos/spack_repo/builtin/packages/gaudi/package.py
+++ b/repos/spack_repo/builtin/packages/gaudi/package.py
@@ -162,6 +162,7 @@ class Gaudi(CMakePackage, CudaPackage):
     depends_on("cppgsl")
     depends_on("fmt@:8", when="@:36.9")
     depends_on("fmt@:10")
+    depends_on("fmt@10:11", when="@39:")
     depends_on("intel-tbb@:2020.3", when="@:37.0")
     depends_on("tbb", when="@37.1:")
     depends_on("uuid")


repos/spack_repo/builtin/packages/gaudi/package.py
```diff
@@ -162,6 +162,7 @@ class Gaudi(CMakePackage, CudaPackage):
     depends_on("cppgsl")
     depends_on("fmt@:8", when="@:36.9")
     depends_on("fmt@:10")
```
wdconinc, line 164: Note that this line will still limit fmt to 10 for 39 and later.

jmcarcell:
If you enable tests then no recent version should build with fmt 11 or 12: https://gitlab.cern.ch/gaudi/Gaudi/-/issues/345. Also any code that is formatting `Gaudi::Property<std::vector<int>>` or similar will not compile. If you disable tests, I have been building Gaudi for a while with fmt 12, unless some option enables code that has the formatting that fails. Being fully correct here would mean incompatible with fmt 11 or 12 but it's true that that formatting is probably not used often.

tmadlener:
OK. Then this might need some more thought. `fmt@12` doesn't build even without tests. I thought I needed `fmt@11` to get the Key4hep stack to concretize because `mariadb` has started to have `fmt` constraints that haven't existed before. But now I have a stack again that has `fmt@10.2`.

PR #2602, opened 2025-11-29
podio: ensure Python.h is found in ROOT ACLiC

wdconinc:
This PR ensures that `Python.h` in `podio` headers is found when the header is used from within a ROOT session.

Intends to resolve errors like https://eicweb.phy.anl.gov/EIC/benchmarks/physics_benchmarks/-/jobs/6873367#L72

### DIFF
--- a/repos/spack_repo/builtin/packages/podio/package.py
+++ b/repos/spack_repo/builtin/packages/podio/package.py
@@ -129,7 +129,7 @@ class Podio(CMakePackage):
 
         # Pythonizations require Python.h accessible for ACLiC
         if self.spec.satisfies("@1.5:"):
-            env.prepend_path("ROOT_INCLUDE_PATH", self.spec["python"].headers.directories)
+            env.prepend_path("ROOT_INCLUDE_PATH", self.spec["python"].headers.directories[0])
 
     def setup_dependent_build_environment(
         self, env: EnvironmentModifications, dependent_spec: Spec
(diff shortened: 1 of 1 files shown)


tmadlener (approved):

wdconinc:
@tmadlener This actually doesn't seem to fix the issue just yet...

tmadlener:
Hmm... we haven't run into any issues so far with this in our nightly builds, so I can't say anything from our experience.

Checking the directory layout of a spack python installation, I see `<prefix>/include/python3.13/<header-files>`. Does the `include.directories` of the `python` spec already do the "right thing" here?

wdconinc:
> Checking the directory layout of a spack python installation, I see `<prefix>/include/python3.13/<header-files>`. Does the `include.directories` of the `python` spec already do the "right thing" here?

Yeah, I'm confused. It's supposed to hit this:
https://github.com/spack/spack-packages/blob/87040045320496ae42a6e741dc065940a5daf9d2/repos/spack_repo/builtin/packages/python/package.py#L1118

Maybe the issue is that `prepend_path` is not happy with a list...

wdconinc:
> Maybe the issue is that `prepend_path` is not happy with a list...

Indeed:
```
==> Warning: /home/wdconinc/git/spack-packages/repos/spack_repo/builtin/packages/podio/package.py:132: when setting environment variable ROOT_INCLUDE_PATH=['/opt/software/linux-skylake/python-3.13.8-xzu2r3aoblk6eepzlqmkirfbwgqpazfm/include/python3.13']: value is of type `list`, but `str` or `PurePath` was expected. This is deprecated and will be an error in Spack v1.0
```

wdconinc:
ba94c07 indeed fixed the issue.

wdconinc:
@spackbot run pipeline

sethrj (approved):

PR #2603, opened 2025-11-29
py-ty: add new package

adamjstewart:
Fast type checker from the makers of ruff. Not yet stable, so we shouldn't switch Spack any time soon, but something to keep your eyes on.

@tgamblin type checks all of Spack core in < 4 s


alecbcs (approved):

PR #2619, opened 2025-11-30
py-mplhep(-data): add new versions

wdconinc:
This PR adds new versions to `py-mplhep` and `py-mplhep-data`. Minor versions, no changes to dependencies.


RikkiButler20 (approved):
Looks good to me.

PR #2631, opened 2025-12-01
velvet: Add c dependency and use of zlib.h from zlib-api

ebagrenrut:
- `velvet` is a C application and requires a dependency on `c`. If a Spack instance has been populated with all of `velvet`'s build dependencies from a binary mirror (and thus never installing a compiler as a build dependency), the `velvet` build will fail as there's no compiler. We add `depends_on("c")`.
- `velvet` depends on `zlib-api`, but does not actually use the `zlib.h` header from the Spack-installed package and thus fails to compile on a host that does not have a zlib devel package installed. We add `-L{spec['zlib-api'].prefix.include}` to `CFLAGS` in the `velvet` `Makefile` to ensure `zlib.h` can be found.


ebagrenrut:
@snehring

alecbcs (approved):
Looks good to me. Thanks @ebagrenrut!

PR #2634, opened 2025-12-01
git: bump version, add 2.48.2 2.49.0 2.49.1 2.50.0 2.50.1 2.51.0 2.51.1 2.51.2 2.52.0

albestro:
This PR started as a simple version bump, but it turned out to require a bit more work to solve a change in the build system, related also to some compatibility requirements on `glibc` dependency.

It can be summarised as:
- up to `git@:2.49` there was no `CSPRNG_METHOD` set, so `/dev/urandom` was used
- starting from `git@2.50:` it sets `CSPRNG_METHOD=getrandom` on `platform=linux`, which has been introduced in `glibc@2.25:`
- from `glibc@2.36:` we can start using `CSPRNG_METHOD=arc4random`

I opted for
- not altering the behavior of `git@:2.49`
- starting from `git@2.50:` on
  - fallback to default `/dev/urandom` if `glibc` does not support `getrandom`
  - if `glibc` supports `getrandom`, just use the default (no action)
  - if `glibc` supports `arc4random`, use it

In addition to this, I also took the chance to update some of the dependencies constraints.


msimberg (dismissed):
Looks good to me, but giving maintainer some time to have a look.

albestro:
CI reports a problem about `sys/random.h`
https://gitlab.spack.io/spack/spack-packages/-/jobs/19075835#L600

This is what I figured out:

- [here](https://github.com/git/git/blob/f0ef5b6d9bcc258e4cbef93839d1b7465d5212b9/compat/posix.h#L158-L160) is where `sys/random.h` is used, and it is conditional on `HAVE_GETRANDOM`
- Where [`HAVE_GETRANDOM`](https://github.com/git/git/blob/f0ef5b6d9bcc258e4cbef93839d1b7465d5212b9/Makefile#L2266-L2268) is detected in Makefile, which depends on `CSPRNG_METHOD` string
- [`CSPRNG_METHOD`](https://github.com/git/git/blob/f0ef5b6d9bcc258e4cbef93839d1b7465d5212b9/Makefile#L156-L162) documentation
- In https://github.com/git/git/commit/cdda67de0316ec29dfc1e290bb7f2154b7b95ee8 there has been a change of logic on how defaults are set

The last point above is very informative and detailed. A quick recap:
- `git@:2.49` there was no `CSPRNG_METHOD` set, so `/dev/urandom` was used
- `git@2.50:`, with that commit, `CSPRNG_METHOD` starts setting `getrandom` on `platform=linux`, which has been introduced in `glibc@2.25:`
- from `glibc@2.36:` we can start using `arc4random`

I tried to map this recap into https://github.com/spack/spack-packages/pull/2634/commits/edb2dcbcb52a84a63013e90b3ac6179efdef73e8 and I opted for not altering the behavior of `git@:2.49`, i.e. before this change (but open to apply this change more broadly, if it works and if you think it's worth).

In addition to this, I also took the chance to update some of the dependencies constraints.

EDIT: I kind of reported this message in the main PR message, so it ends up in the commit message.

albestro:
@jennfshr do you think it is ok to merge this?

jennfshr (approved):
Checked shas and changes. Looks good to me.

RMeli (approved):

PR #2639, opened 2025-12-02
py-python-lsp-server: update url, new version

Chrismarsh:
- update url pattern
- add new version
- update deps

### DIFF
--- a/repos/spack_repo/builtin/packages/py_python_lsp_server/package.py
+++ b/repos/spack_repo/builtin/packages/py_python_lsp_server/package.py
@@ -32,9 +32,7 @@ class PyPythonLspServer(PythonPackage):
     depends_on("py-setuptools-scm@3.4.3:+toml", type="build")
 
     variant(
-        "formatter",
-        values=any_combination_of("black", "ruff"),
-        description="Formatter to use",
+        "formatter", values=any_combination_of("black", "ruff"), description="Formatter to use"
     )
 
     with default_args(type=("build", "run")):
--- a/repos/spack_repo/builtin/packages/py_python_lsp_server/package.py
+++ b/repos/spack_repo/builtin/packages/py_python_lsp_server/package.py
@@ -31,6 +31,12 @@ class PyPythonLspServer(PythonPackage):
     depends_on("py-setuptools@69:", type="build", when="@1.13:")
     depends_on("py-setuptools-scm@3.4.3:+toml", type="build")
 
+    variant(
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/py_python_lsp_server/package.py
```diff
@@ -38,4 +40,13 @@ class PyPythonLspServer(PythonPackage):
         depends_on("py-python-lsp-jsonrpc@1.0.0:1")
         depends_on("py-python-lsp-jsonrpc@1.1.0:1", when="@1.8.0:")
         depends_on("py-ujson@3.0.0:")
+        depends_on("py-black", when="@1.13:")
```
alecbcs: It seems like the lsp-server only depends on `black` to validate format signatures in docstrings and can also depend on ruff for the same functionality. I'd propose we add a `formatter` variant that can either be `None`, `black`, or `ruff` that sets the relevant dependency info depending on which is chosen. (Default to None.)

https://github.com/python-lsp/python-lsp-server/commit/c8e4e992a7d0aee00804433a2211a7b3356f5b6d#diff-749bbcf2f7f5bd5289d6223d2a4cb36ac92ff1960ca2461b01cbd732ebfbd308R75-R78

alecbcs (changes requested):
One quick nitpick around adding `black` as a required dependency.

Otherwise looks good to me. Thanks @Chrismarsh!

repos/spack_repo/builtin/packages/py_python_lsp_server/package.py
Chrismarsh: Mmh I was just following what I see in the pyproject.toml. I don't see any ruff dependency in there. Is there a reason adding black here is sub-optimal? 

I only updated this because I wanted to use it with `jupyter lab`. It isn't clear to me how or where `pylsp.signature.formatter` should be set accordinly. Do you have any idea?

Chrismarsh:
@spackbot fix style

repos/spack_repo/builtin/packages/py_python_lsp_server/package.py
Chrismarsh: I've made the modification

Chrismarsh:
@alecbcs are you happy with how these changes are?

Chrismarsh:
gentle reminder on this @alecbcs

alecbcs (approved):

PR #2640, opened 2025-12-02
DIY depends on "c" in addition to "cxx"

tpeterka:


alecbcs (approved):
Looks good to me. Thanks @tpeterka!

Pending final review from @vicentebolea as maintainer.

vicentebolea (approved):
DIY use the default languages in of project in CMAKE: C and CXX. Thanks!

PR #2651, opened 2025-12-03
Braker 3.0.8

Keeper-of-the-Keys:


tldahlgren (changes requested):
You have a couple of commits by someone else that need to be dropped from this PR.

tldahlgren:
> You have a couple of commits by someone else that need to be dropped from this PR.

Otherwise, the checksum is confirmed.

Keeper-of-the-Keys:
Hey @tldahlgren 

In the end I'm not sure that this package should be included.
It depends on `genemark-et` which is no longer installable using spack as far as I can tell due to changes in how their site/licensing works.

(It fails with a reference to manual download instructions which involves manual registration for a 1 year license, I am not so familiar with spack yet but that does not seem to be something that can be automated).

Would be happy to hear what you think.

tldahlgren:
> Hey @tldahlgren
> 
> In the end I'm not sure that this package should be included. It depends on `genemark-et` which is no longer installable using spack as far as I can tell due to changes in how their site/licensing works.
> 
> (It fails with a reference to manual download instructions which involves manual registration for a 1 year license, I am not so familiar with spack yet but that does not seem to be something that can be automated).
> 
> Would be happy to hear what you think.

There are a few options depending on how you want to proceed. The most well-known manually downloaded dependency in Spack is probably `matlab`. 

Can you manually download `genemark-et` and successfully build it? And use it to build this package (see the docstring for a suggestion or check out how to set up a mirror in the docs)?

The package has been around for about 8 years so there *may* be people using it or they could be pinged if they would mind the package being removed if you would rather remove it.

IF you do want to keep the changes, there are a couple of unrelated commits that need to be dropped, one go which is causing a conflict.

Let me know how you want to proceed.

tldahlgren:
There are conflicts with the unrelated `.ci/env` package. You might want to simply drop all but the one commit you made OR you could start a new PR from `develop`. Up to you.

Add a comment here tagging me when you decide how you want to proceed.

kwryankrattiger:
This needs to be rebased on develop without the release changes in it.

bernhardkaindl:
@Keeper-of-the-Keys, tldahlgren has retired and has an open request for changes which can't be changed.

This means if you want to merge this PR, you need to consolidate the changes on top of the current develop branch and open a new PR.

Also you weren't sure that this can even go forward.
https://github.com/spack/spack-packages/pull/2651#issuecomment-3697548950

Closing because the PR does not show activity or changes since initially opening it and nearly 7 months have passed since.

Keeper-of-the-Keys:
Hey @bernhardkaindl,

We were unsure if we are continuing with spack because due to the way our env works it kept breaking.

Right now I am doing a build for our new env with spack 1.3.0dev, if that all goes well I probably will submit some PRs.

(there are a non-zero amount of packages that were written locally by my predecessor and never pushed up to spack)

Thanks!

bernhardkaindl:
@Keeper-of-the-Keys, thanks for the information!

If spack tends to work in a way do do things which break your environment, I think you're also welcome to join the `spackpm` Slack workspace. On it you have many channels where you can reach other uses as well as the maintainers, technical steering committee members and general leadership!

PR #2655, opened 2025-12-03
bedops updates

ebagrenrut:
- Add BEDOPS 2.4.42
- Add a run-time dependency on `python`. BEDOPS includes multiple Python scripts that end up in prefix.bin, including `starch-diff`, `update-sort-bed-migrate-candidates`, `update-sort-bed-slurm`, `update-sort-bed-starch-slurm`.
- Correct license directive. Though the BEDOPS suite _includes_ bzip2, its license is GPL 2.0. The updated directive indicates licenses for all code that appears to be in the BEDOPS source, including the `bzip2-1.0.6` license identifier for bzip2, the `MIT` license identifier for jansson, and the `Zlib` license identifier for Zlib.

### DIFF
--- a/repos/spack_repo/builtin/packages/bedops/package.py
+++ b/repos/spack_repo/builtin/packages/bedops/package.py
@@ -18,9 +18,9 @@ class Bedops(MakefilePackage):
 
     maintainers("jacorvar")
 
-    # More regarding the peculiar license() directive: bedops is licensed under the GPL v2, but the
-    # bedops source bundles source for bzlip2, jansson, and zlib, and its LICENSE has the licenses
-    # for those source codes appended.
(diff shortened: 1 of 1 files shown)


ebagrenrut:
@jacorvar

ebagrenrut:
More regarding the peculiar `license()` directive: bedops is licensed under the GPL v2, but as shown [here](https://github.com/bedops/bedops/tree/master/third-party), the `bedops` source bundles source for bzlip2, jansson, and zlib, and its [LICENSE](https://github.com/bedops/bedops/blob/master/LICENSE) has the licenses for those source codes appended.

becker33 (approved):

PR #2656, opened 2025-12-03
amrex: add v25.12

WeiqunZhang:


repos/spack_repo/builtin/packages/amrex/package.py
EZoni, line 20: Is the `url` kept to 23.05 intentionally? I understand it does not _need_ to be updated, but why is it kept outdated?

EZoni (approved):

repos/spack_repo/builtin/packages/amrex/package.py
EZoni: Answered on Slack. See original discussion in https://github.com/spack/spack/pull/37362#discussion_r1181726018.

ax3l: Does not really matter :)
Spack will discover the used syntax to grep when it searches new versions.

Important is just that we stay consistent how we tag our version numbers.

ax3l (approved):

PR #2669, opened 2025-12-05
Add necessary language support for gadap and grads

vanderwb:
These two packages have likely been broken since the 1.0 release. The `gadap` package only has a `cxx` dependency but needs `c` as well. Meanwhile, the opposite is the case for `grads`. Grads also has some implicitly declared functions for which we now bypass checks in relevant situations.


alecbcs (approved):
Looks good to me. Thanks @vanderwb!

PR #2676, opened 2025-12-05
r-terra: add 1.8-86, tbb variant

Chrismarsh:
- Adds 1.8-86
- Adds an optional (default False) TBB variant for parallel computing. I don't see the 1.7.x tags on the github so I am unsure if this is a 1.8x only or not.

Tested on linux with `+tbb`

### DIFF
--- a/repos/spack_repo/builtin/packages/r_terra/package.py
+++ b/repos/spack_repo/builtin/packages/r_terra/package.py
@@ -24,6 +24,7 @@ class RTerra(RPackage):
 
     license("GPL-3.0-or-later")
 
+    version("1.8-86", sha256="a446375352d1a01d6bb9ec0a0ac0031ac11a1252567278bd34b4a136f6de7474")
     version("1.7-78", sha256="658956b79d8a1371aefdf7300316f1756b58d436ba549ade012307684b2d4b7e")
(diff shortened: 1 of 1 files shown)


Chrismarsh:
@wdconinc do you have capacity to review this?

repos/spack_repo/builtin/packages/r_terra/package.py
```diff
@@ -42,3 +45,4 @@ class RTerra(RPackage):
     depends_on("geos@3.4.0:")
     depends_on("proj@4.9.3:")
     depends_on("sqlite")
+    depends_on("intel-oneapi-tbb", when="+tbb")
```
wdconinc, line 48: Don't we have a virtual for tbb? https://github.com/spack/spack/blob/d182b35601d84cf16dceca39b5cfc06472d483fd/etc/spack/defaults/base/packages.yaml#L62

But I see now that this doesn't even include this intel-oneapi-tbb... https://github.com/spack/spack/pull/51702

wdconinc (approved):

repos/spack_repo/builtin/packages/r_terra/package.py
wdconinc: Feel free to resolve this for merging. Or adapt to use the virtual, as you see fit.

PR #2681, opened 2025-12-05
Don't add binutils on rhel 7-8 systems when using mold.

danielsjensen1:
The gcc package forces +binutils when a rhel 7 or 8 system is detected even when mold is selected as the linker.  This behavior is undesirable and causes linking errors because binutils gets added to the search path (see `if self.spec.satisfies("+binutils")`).

Behavior on develop
```sh
$ spack spec gcc+mold
 -   gcc@15.2.0+binutils+bootstrap~graphite+mold~nvptx~piclibs~profiled~strip build_system=autotools build_type=RelWithDebInfo languages:='c,c++,fortran' platform=linux os=rhel8 target=skylake_avx512 %c,cxx=gcc@8.5.0
```
Behavior on gcc-fix-mold-on-rhel7-8
```sh
$ spack spec gcc+mold
 -   gcc@15.2.0~binutils+bootstrap~graphite+mold~nvptx~piclibs~profiled~strip build_system=autotools build_type=RelWithDebInfo languages:='c,c++,fortran' platform=linux os=rhel8 target=skylake_avx512 %c,cxx=gcc@8.5.0
```

### DIFF
--- a/repos/spack_repo/builtin/packages/gcc/package.py
+++ b/repos/spack_repo/builtin/packages/gcc/package.py
@@ -413,10 +413,10 @@ class Gcc(AutotoolsPackage, GNUMirrorPackage, CompilerPackage):
     # Ditto for RHEL7/8: OpenBLAS uses flags which the RHEL system-binutils don't have:
     # https://github.com/xianyi/OpenBLAS/issues/3805#issuecomment-1319878852
     conflicts(
-        "~binutils", when="@10: os=rhel7", msg="gcc: Add +binutils - preinstalled as might be old"
+        "~binutils", when="@10: os=rhel7 ~mold", msg="gcc: Add +binutils - preinstalled as might be old"
     )
     conflicts(
-        "~binutils", when="@10: os=rhel8", msg="gcc: Add +binutils - preinstalled as might be old"
+        "~binutils", when="@10: os=rhel8 ~mold", msg="gcc: Add +binutils - preinstalled as might be old"
     )
 
     # GCC 11 requires GCC 4.8 or later (https://gcc.gnu.org/gcc-11/changes.html)
--- a/repos/spack_repo/builtin/packages/gcc/package.py
+++ b/repos/spack_repo/builtin/packages/gcc/package.py
@@ -413,10 +413,14 @@ class Gcc(AutotoolsPackage, GNUMirrorPackage, CompilerPackage):
     # Ditto for RHEL7/8: OpenBLAS uses flags which the RHEL system-binutils don't have:
     # https://github.com/xianyi/OpenBLAS/issues/3805#issuecomment-1319878852
     conflicts(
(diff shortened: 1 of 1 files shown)


danielsjensen1:
Additional participant: @dc-snl

bernhardkaindl (commented):
Obsolete since this merge into develop:
```py
commit e6d347de0cecc04e9837ae1e90a42f99354501b7
Author: Harmen Stoppels <me@harmenstoppels.nl>
Date:   Wed Feb 4 18:00:00 2026 +0100

    gcc: default to +binutils (#3106)
    
    GCC has a conflict on `~binutils` for RHEL, where the system
    assembler and linker are ancient. However
    
    1. Spack has no clue whether or not the system toolchain is used, so the
       conflict can be too restrictive
    2. In general when building a compiler, it makes sense to also build the
       assembler and linker; we wanna avoid system dependencies.
    
    So, make things less restrictive by dropping the conflict, and instead
    use better defaults.
    
    
    Signed-off-by: Harmen Stoppels <me@harmenstoppels.nl>

diff --git a/repos/spack_repo/builtin/packages/gcc/package.py b/repos/spack_repo/builtin/packages/gcc/package.py
index ae55e1c9a1..57f20b2559 100644
--- a/repos/spack_repo/builtin/packages/gcc/package.py
+++ b/repos/spack_repo/builtin/packages/gcc/package.py
@@ -180,7 +180,7 @@ class Gcc(AutotoolsPackage, GNUMirrorPackage, CompilerPackage):
         multi=True,
         description="Compilers and runtime libraries to build",
     )
-    variant("binutils", default=False, description="Build via binutils")
+    variant("binutils", default=True, description="Use binutils linker and assembler")
     variant("mold", default=False, description="Use mold as the linker by default", when="@12:")
     variant(
         "piclibs", default=False, description="Build PIC versions of libgfortran.a and libstdc++.a"
@@ -408,18 +408,6 @@ class Gcc(AutotoolsPackage, GNUMirrorPackage, CompilerPackage):
         msg="Only GCC 11.3+ support aarch64-darwin",
     )
 
-    # Newer binutils than RHEL's is required to run `as` on some instructions
-    # generated by new GCC (see https://github.com/spack/spack/issues/12235)
-    conflicts("~binutils", when="@7: os=rhel6", msg="New GCC cannot use system assembler on RHEL6")
-    # Ditto for RHEL7/8: OpenBLAS uses flags which the RHEL system-binutils don't have:
-    # https://github.com/xianyi/OpenBLAS/issues/3805#issuecomment-1319878852
-    conflicts(
-        "~binutils", when="@10: os=rhel7", msg="gcc: Add +binutils - preinstalled as might be old"
-    )
-    conflicts(
-        "~binutils", when="@10: os=rhel8", msg="gcc: Add +binutils - preinstalled as might be old"
-    )
-
     # GCC 11 requires GCC 4.8 or later (https://gcc.gnu.org/gcc-11/changes.html)
     conflicts("%gcc@:4.7", when="@11:")
```

PR #2684, opened 2025-12-06
Paraconf: Add version 1.0.1

tpadioleau:


jbigot:
Thanks for the version bump! This looks great, but we should take into account that 1.0.1 requires cmake 3.10.

jbigot:
@tpadioleau can you update the dependency on cmake for 1.0.1, please

tpadioleau:
> @tpadioleau can you update the dependency on cmake for 1.0.1, please

Oh yes true sorry for the mistake

tpadioleau:
Done in 2424dc9e3d9e46bd37e1c4b0af05f56b043afaef

jbigot:
thank you very much! This looks perfect

jbigot (approved):

tldahlgren (approved):
Confirmed the sha256.

PR #2700, opened 2025-12-09
py-numba: add v0.63.0, Python 3.14 support

adamjstewart:

### DIFF
--- a/repos/spack_repo/builtin/packages/py_llvmlite/package.py
+++ b/repos/spack_repo/builtin/packages/py_llvmlite/package.py
@@ -16,7 +16,13 @@ class PyLlvmlite(PythonPackage):
 
     license("BSD-2-Clause")
 
-    version("0.45.0rc2", sha256="c264f8ac7ffc139d66cd243f6369d71c164098cc4fdf4e0af90dde004ba39fc1")
+    version("0.46.0", sha256="227c9fd6d09dce2783c18b754b7cd9d9b3b3515210c46acc2d3c5badd9870ceb")
+    version("0.45.1", sha256="09430bb9d0bb58fc45a45a57c7eae912850bedc095cd0810a57de109c69e1c32")
+    version(
+        "0.45.0rc2",
+        sha256="c264f8ac7ffc139d66cd243f6369d71c164098cc4fdf4e0af90dde004ba39fc1",
+        deprecated=True,
+    )
     version("0.44.0", sha256="07667d66a5d150abed9157ab6c0b9393c9356f229784a4385c02f99e94fc94d4")
     version("0.43.0", sha256="ae2b5b5c3ef67354824fb75517c8db5fbe93bc02cd9671f3c62271626bc041d5")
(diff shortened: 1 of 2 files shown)


repos/spack_repo/builtin/packages/py_llvmlite/package.py
```diff
@@ -30,16 +36,21 @@ class PyLlvmlite(PythonPackage):
...
     version("0.37.0", sha256="6392b870cd018ec0c645d6bbb918d6aa0eeca8c62674baaee30862d6b6865b15")
 
-    depends_on("c", type="build")
-    depends_on("cxx", type="build")
-    depends_on("cmake", type="build")
+    with default_args(type="build"):
+        depends_on("c")
+        depends_on("cxx")
+        depends_on("cmake")
+        depends_on("binutils", when="platform=linux")
```
adamjstewart, line 43: Binutils generally doesn't work on macOS. Most build systems are hard-coded to use certain flags per OS, but the linker coming from Xcode and binutils are completely different. I was able to build this package with the Xcode linker just fine (and binutils doesn't even compile with Apple Clang 17).

repos/spack_repo/builtin/packages/py_llvmlite/package.py
```diff
@@ -60,11 +71,6 @@ class PyLlvmlite(PythonPackage):
     ]:
         depends_on("llvm@10.0", when=f"@0.34:0.36 target={t}")
 
-    depends_on("binutils", type="build")
-
-    # TODO: investigate
-    conflicts("%apple-clang@15:")
```
adamjstewart, line 66: Builds just fine for me on macOS 26.1 with Apple Clang 17.0.0 after removing the binutils dep.

repos/spack_repo/builtin/packages/py_numba/package.py
```diff
@@ -33,41 +43,50 @@ class PyNumba(PythonPackage):
...
+        depends_on("python@3.10:3.13", when="@0.61:0.62")
+        depends_on("python@3.9:3.12", when="@0.59:0.60")
+        depends_on("python@3.8:3.11", when="@0.57:0.58")
+        depends_on("python@3.7:3.10", when="@0.55:0.56")
+        depends_on("python@3.7:3.9", when="@0.54")
+
+    with default_args(type=("build", "run")):
+        # Use min_numpy_run_version, not min_numpy_build_version
+        # min_numpy_build_version may be higher to ensure backwards-compatibility of wheels,
+        # but this doesn't matter for Spack which always guarantees compatibility
```
adamjstewart, line 63: This is one change from the previous recipe. When building binaries for PyPI, you have to build with numpy 2+ to ensure compatibility with both numpy 1 and 2. However, we don't build generic ABI-agnostic binaries with Spack, so there's no reason to require numpy 2+.

adamjstewart:
@spackbot run pipeline

wdconinc (approved):
This looks good to me. Thanks!

PR #2702, opened 2025-12-09
trilinos: update cxxstd variant default

alalazo:
In this way the default version of trilinos gets the default value of the variant

Discovered using https://github.com/spack/spack/pull/51674

### DIFF
--- a/repos/spack_repo/builtin/packages/trilinos/package.py
+++ b/repos/spack_repo/builtin/packages/trilinos/package.py
@@ -95,7 +95,7 @@ class Trilinos(CMakePackage, CudaPackage, ROCmPackage):
     variant("cuda_rdc", default=False, description="Turn on RDC for CUDA build")
     variant("rocm_rdc", default=False, description="Turn on RDC for ROCm build")
     variant(
-        "cxxstd", default="14", description="C++ standard", values=["11", "14", "17"], multi=False
+        "cxxstd", default="17", description="C++ standard", values=["11", "14", "17"], multi=False
     )
     variant("debug", default=False, description="Enable runtime safety and debug checks")
     variant(


repos/spack_repo/builtin/packages/trilinos/package.py
```diff
@@ -95,7 +95,7 @@ class Trilinos(CMakePackage, CudaPackage, ROCmPackage):
     variant("cuda_rdc", default=False, description="Turn on RDC for CUDA build")
     variant("rocm_rdc", default=False, description="Turn on RDC for ROCm build")
     variant(
-        "cxxstd", default="14", description="C++ standard", values=["11", "14", "17"], multi=False
+        "cxxstd", default="17", description="C++ standard", values=["11", "14", "17"], multi=False
```
cgcgcg, line 98: This needs to be updated. We probably want
```suggestion
        "cxxstd", default="20", description="C++ standard", values=["20", "23"], multi=False
```

cgcgcg: But that's version dependent, so it might require some additional work.

alalazo: My only concern at the moment is to have a recipe where the default value of that variant doesn't conflict with the default version of the package (`trilinos@16` cannot use `cxxstd=14`).

alalazo:
@psakievich

repos/spack_repo/builtin/packages/trilinos/package.py
cgcgcg: I agree that we don't want 14 as the default, but that we need to keep it because older versions of Trilinos did support it. The latest 16.2 release uses 17 as the default. The upcoming 17.0 release uses 20 as the default (and no longer supports 17). Develop and master also require 20. This means that we will need some additional logic in the package.

psakievich: @cgcgcg I recommend we specify requirements for the older versions to have valid ranges and set the default to 20. This way the older versions can't get the wrong cxxstd

alalazo: I agree with what you say, it's only that the scope of this PR is to fix the default value for the variant, without adding new versions :slightly_smiling_face:

tldahlgren (approved):

repos/spack_repo/builtin/packages/trilinos/package.py
kuberry: If we are only looking to fix the default value in this PR, can we replace the line with:
```
"cxxstd", default="20", description="C++ standard", values=["11", "14", "17","20", "23"], multi=False
```

Then, in a separate PR, we will have all of the option values we need for a version-dependent `cxxstd`? (instead of removing 11, 14, and 17 here and then adding them back in that PR?)

kuberry (changes requested):
We have existing, registered conflicts with `cxxstd` in 
https://github.com/spack/spack-packages/blob/5a6699e2860e192af40fedfb9bcc33aab9d71939/repos/spack_repo/builtin/packages/trilinos/package.py#L363-L365
which use the values 11, 14, and 17.

May I suggest the following changes:
```
diff --git a/repos/spack_repo/builtin/packages/trilinos/package.py b/repos/spack_repo/builtin/packages/trilinos/package.py
index 74641dfc32..476d597a0a 100644
--- a/repos/spack_repo/builtin/packages/trilinos/package.py
+++ b/repos/spack_repo/builtin/packages/trilinos/package.py
@@ -95,7 +95,11 @@ class Trilinos(CMakePackage, CudaPackage, ROCmPackage):
     variant("cuda_rdc", default=False, description="Turn on RDC for CUDA build")
     variant("rocm_rdc", default=False, description="Turn on RDC for ROCm build")
     variant(
-        "cxxstd", default="14", description="C++ standard", values=["11", "14", "17"], multi=False
+        "cxxstd",
+        default="20",
+        description="C++ standard",
+        values=["11", "14", "17", "20", "23"],
+        multi=False,
     )
     variant("debug", default=False, description="Enable runtime safety and debug checks")
     variant(
@@ -362,7 +366,7 @@ class Trilinos(CMakePackage, CudaPackage, ROCmPackage):
     conflicts("+adios2", when="@:12.14.1")
     conflicts("cxxstd=11", when="@13.2:")
     conflicts("cxxstd=14", when="@14:")
-    conflicts("cxxstd=17", when="@:12")
+    conflicts("cxxstd=17", when="@:12,17:")
     conflicts("cxxstd=11", when="+wrapper ^cuda@6.5.14")
     conflicts("cxxstd=14", when="+wrapper ^cuda@6.5.14:8.0.61")
     conflicts("cxxstd=17", when="+wrapper ^cuda@6.5.14:10.2.89")
```

repos/spack_repo/builtin/packages/trilinos/package.py
alalazo: @kuberry The preferred version of `trilinos` is currently `v16.1.0`, so the purpose of this PR is to pick a default value for `cxxstd` that can be taken by `v16.1.0` (otherwise we would take a penalty on variants during concretization for picking the default version).  The changes you propose here:

- https://github.com/spack/spack-packages/pull/2702#pullrequestreview-3578874323

don't seem correct because they add two new values without adding the corresponding constraints. This means with that changes `cxxstd=20` or `cxxstd=23` are valid values for any version of `trilinos` known to Spack.

Also, is `cxxstd=20` a valid value for `v16.1.0` - if not I'd stick with `17` being the default for now?

sebrowne: I think this change makes sense as a self-contained remedy to the specific issue at hand.

Agree that if the default version is `16.1.0` then the default C++ standard should be `17`.  Fixing `develop`/`master` (which were broken on the Trilinos side with https://github.com/trilinos/Trilinos/pull/14390) can be handled as a separate change IMO.

Other changes are certainly necessary given Trilinos' recent advance of C++ standard requirement and the new releases, and I'm happy to handle those with another PR.

@alalazo the difficulty here is that _parts_ of Trilinos are C++20 compliant/known working with `16.1.0`, it really depends on what the user wants to build.  We could try to map the individual package requirements between C++ standard and package enables, but I think that would be best left to the future (tentative but in-progress) work of splitting the way that Trilinos is handled within Spack to be package-by-package (e.g. `trilinos+tpetra+rol` -> `trilinos-rol trilinos-tpetra`).  For now, I think it's acceptable to leave the versions that are declared as supported in TriBITS as what are available in Spack, with the defaults likewise mapped.  E.g. we KNOW that C++11 is not supported by `trilinos@16.1.0`, but C++23 _may_ work for `trilinos@16.2.0` depending on which packages are required (can be seen [here](https://github.com/trilinos/Trilinos/blob/master/cmake/SetTrilinosCxxStandard.cmake#L3) currently).

kuberry (approved):

alalazo:
@tldahlgren Can we merge? Otherwise I'd close the PR.

PR #2711, opened 2025-12-09
hugo: Add cxx dependency for extended edition

ebagrenrut:
Per the `hugo` [README](https://github.com/gohugoio/hugo?tab=readme-ov-file#build-from-source), a compiler is needed when building the extended edition (using the `+extended` variant), but the package is missing this dependency. I've found that when building `hugo+extended`, a build dependency on `cxx` is sufficient to ensure that the binary is successfully linked to libstdc++ and libgcc_s from the `gcc-runtime` package, and the binary is successfully relocated when installed from buildcache.

@alecbcs

### DIFF
--- a/repos/spack_repo/builtin/packages/hugo/package.py
+++ b/repos/spack_repo/builtin/packages/hugo/package.py
@@ -46,9 +46,7 @@ class Hugo(GoPackage):
     depends_on("go@1.20:", type="build", when="@0.123:")
     depends_on("go@1.18:", type="build", when="@0.106:")
     depends_on("go@1.11:", type="build", when="@0.48:")
-
-    with when("+extended"):
-        depends_on("cxx", type="build")
+    depends_on("cxx", type="build", when="+extended")
 
     variant("extended", default=False, description="Enable extended features")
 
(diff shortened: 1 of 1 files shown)


alalazo (dismissed):

repos/spack_repo/builtin/packages/hugo/package.py
```diff
@@ -46,6 +46,7 @@ class Hugo(GoPackage):
     depends_on("go@1.20:", type="build", when="@0.123:")
     depends_on("go@1.18:", type="build", when="@0.106:")
     depends_on("go@1.11:", type="build", when="@0.48:")
+    depends_on("cxx", type="build", when="+extended")
```
alalazo, line 50: Can you explain why it's `cxx` being needed and not just `c`?

ebagrenrut: The resulting binary is linked to `libstdc++`, so I know it needs `cxx`. However, I don't actually know that it _doesn't_ need `c`; I'm not actually sure how cgo works :) It may very well be that it uses the c compiler while linking, and in my setup `gcc` is available. Perhaps I should also add a `c` dependency?

ebagrenrut: I've added a `c` dependency.

alalazo (approved):

PR #2715, opened 2025-12-10
libcatalyst: fixes +python and drop conduit as internal package

albestro:
Changes:
- drop `conduit` variant, promoting use of spack to build dependencies
- confirm language dependencies and make `fortran` conditional
- add `extends("python")`
- make `conduit+python` a ("build", "link", "run") dependency
- force `CMAKE_INSTALL_LIBDIR=lib`, it avoids problem when used with `py-venv`

### DIFF
--- a/repos/spack_repo/builtin/packages/amrex/package.py
+++ b/repos/spack_repo/builtin/packages/amrex/package.py
@@ -194,7 +194,7 @@ class Amrex(CMakePackage, CudaPackage, ROCmPackage):
         depends_on("conduit")
         depends_on("conduit +mpi", when="+mpi")
     with when("+catalyst"):
-        depends_on("libcatalyst@2.0: +conduit")
+        depends_on("libcatalyst@2.0:")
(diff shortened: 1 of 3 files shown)


mathstuf (approved):

albestro:
@spackbot run pipeline

albestro:
About the drop of the variant without any deprecation policy, there is this proposal https://github.com/spack/spack/issues/38829 but at the moment we just go for dropping them (if there is sense and agreement on doing so).

alalazo (approved):

alalazo:
Giving some more time to maintainers. Ping me in a few days for a merge if you don't get further comments.

albestro:
ping @alalazo

cquammen:
Dropping the conduit variant has caused a regression and a lot of build headaches for a collaborator who uses libcatalyst. Would it be possible to restore this variant? I know internal/vendored libraries go against spack philosophy, but this change has caused some practical difficulties in an important libcatalyst consumer.

PR #2716, opened 2025-12-10
occa: patch for build problem and bump version

albestro:
Changelog:
- patch for a build problem started with 1.2.0 and for which a fix has been merged in 1.6.0
- add `url` that enables the usage of `spack checksum` which ease the task of bumping the package
- add some newer versions

I'm not sure why it is preferred to go for the pair `tag`/`commit`, but for the moment I left them as they were.


tldahlgren (approved):
Confirmed the new version sha256s (and all commit+tags).

@v-dobrev Any objections to this PR?

PR #2717, opened 2025-12-10
py-scikit-learn: add v1.8.0

adamjstewart:
https://scikit-learn.org/stable/whats_new/v1.8.html


tldahlgren (approved):
Confirmed the sha256.

PR #2721, opened 2025-12-10
kokkos: cxxstd cleanup

tpadioleau:
- remove the always in conflict variant cxxstd=11
- add variant cxxstd=23 starting from [Kokkos 4](https://github.com/kokkos/kokkos/blob/develop/CHANGELOG.md#build-system-changes-9)

### DIFF
--- a/repos/spack_repo/builtin/packages/kokkos/package.py
+++ b/repos/spack_repo/builtin/packages/kokkos/package.py
@@ -289,16 +289,27 @@ class Kokkos(CMakePackage, CudaPackage, ROCmPackage):
     conflicts("+wrapper", when="~cuda")
     conflicts("+wrapper", when="+cmake_lang")
 
-    cxxstds = ["14", "17", "20", "23"]
     variant(
         "cxxstd",
         default="17",
-        values=(
-            conditional("14", when="@:3"),
-            conditional("17", when="@:4"),
-            "20",
-            conditional("23", when="@4:"),
-        ),
+        values=("14", "17", "20"),
+        when="@3",
+        multi=False,
+        description="C++ standard",
+    )
+    variant(
+        "cxxstd",
+        default="17",
+        values=("17", "20", "23"),
+        when="@4",
+        multi=False,
+        description="C++ standard",
+    )
+    variant(
+        "cxxstd",
+        default="20",
+        values=("20", "23"),
+        when="@5",
         multi=False,
         description="C++ standard",
     )
@@ -312,19 +323,13 @@ class Kokkos(CMakePackage, CudaPackage, ROCmPackage):
     variant("alloc_async", default=False, description="Use CudaMallocAsync", when="@4.2: +cuda")
 
     # SYCL and OpenMPTarget require C++17 or higher
-    for cxxstdver in cxxstds[: cxxstds.index("17")]:
-        conflicts(
-            "+sycl", when="cxxstd={0}".format(cxxstdver), msg="SYCL requires C++17 or higher"
-        )
-        conflicts(
-            "+openmptarget",
-            when="cxxstd={0}".format(cxxstdver),
-            msg="OpenMPTarget requires C++17 or higher",
-        )
+    conflicts("+sycl", when="cxxstd=14", msg="SYCL requires C++17 or higher")
+    conflicts("+openmptarget", when="cxxstd=14", msg="OpenMPTarget requires C++17 or higher")
 
     # HPX should use the same C++ standard
-    for cxxstd in cxxstds[: cxxstds.index("23")]:
-        depends_on("hpx cxxstd={0}".format(cxxstd), when="+hpx cxxstd={0}".format(cxxstd))
+    for cxxstd in ["14", "17", "20"]:
+        depends_on(f"hpx cxxstd={cxxstd}", when=f"+hpx cxxstd={cxxstd}")
+    conflicts("+hpx", when="cxxstd=23", msg="hpx package does not yet support cxxstd=23")
 
     # HPX version constraints
     depends_on("hpx@1.7:", when="+hpx")
--- a/repos/spack_repo/builtin/packages/kokkos/package.py
+++ b/repos/spack_repo/builtin/packages/kokkos/package.py
@@ -307,9 +307,8 @@ class Kokkos(CMakePackage, CudaPackage, ROCmPackage):
     conflicts("+openmptarget", when="cxxstd=14", msg="OpenMPTarget requires C++17 or higher")
 
     # HPX should use the same C++ standard
-    for cxxstd in ["14", "17", "20"]:
+    for cxxstd in ["14", "17", "20", "23"]:
         depends_on(f"hpx cxxstd={cxxstd}", when=f"+hpx cxxstd={cxxstd}")
-    conflicts("+hpx", when="cxxstd=23", msg="hpx package does not yet support cxxstd=23")
 
     # HPX version constraints
     depends_on("hpx@1.7:", when="+hpx")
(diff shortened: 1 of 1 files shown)


tpadioleau:
- Should we continue adding conflicts with CUDA, like `conflicts("+cuda", when="cxxstd=23")` ? I think if one wants to compile with CUDA clang++, it should work ? I think this limitation only concerns `nvcc` ?
- Not a big fan of the fix for HPX in C++23...

rbberger (commented):
Instead of operating with conflicts, we could adopt what is described here: https://spack.readthedocs.io/en/latest/packaging_guide_creation.html#conditional-possible-values

repos/spack_repo/builtin/packages/kokkos/package.py
```diff
@@ -288,14 +288,21 @@ class Kokkos(CMakePackage, CudaPackage, ROCmPackage):
     conflicts("+wrapper", when="~cuda")
     conflicts("+wrapper", when="+cmake_lang")
 
-    cxxstds = ["11", "14", "17", "20"]
-    variant("cxxstd", default="17", values=cxxstds, multi=False, description="C++ standard")
+    cxxstds = ["14", "17", "20", "23"]
+    variant(
+        "cxxstd",
+        default="17",
```
JBludau: Discussion WG: we probably want 20 as default
```suggestion
        default="20",
```

cedricchevalier19: I feel a little bit sceptical about this. We were very conservative trying not to change default for older releases and here, you want to push legacy version to default on c++ 20.
Kokkos 5 will end up using 20 as default thanks to conflict but I am not sure that we want Kokkos < 5 to be c++ 20 by default.

tpadioleau: I need to test, I wonder what happens when the default value is not part of the possible values...

JBludau: > I feel a little bit sceptical about this. We were very conservative trying not to change default for older releases and here, you want to push legacy version to default on c++ 20. Kokkos 5 will end up using 20 as default thanks to conflict but I am not sure that we want Kokkos < 5 to be c++ 20 by default.

Isn't the std something that ultimately comes from the user or other packages. If we are able to build with >=17 for version 4, any other setting in someone's build environment that wants 17 would be fine right? For standalone builds with just Kokkos we would use 20, but given that we want people to transition to 5 eventually, nudging them in this direction looks like a good thing to me

tpadioleau:
```
spack spec --fresh kokkos@5
 -   kokkos@5.0.0~aggressive_vectorization~cmake_lang+complex_align~cuda~debug~debug_bounds_check~hip_relocatable_device_code~hpx~hwloc~ipo~openmp~openmptarget~pic~rocm+serial+shared~sycl~tests~threads~tuning~wrapper build_system=cmake build_type=Release cxxstd=23 generator=make intel_gpu_arch=none platform=linux os=ubuntu20.04 target=icelake %cxx=gcc@10.5.0
```

`cxxstd=23` wut ? @cedricchevalier19 could you try ?

rbberger: I agree with @cedricchevalier19 that it might be too aggressive to change the default for older versions. Kokkos 5 already defaults to 20, that's good enough for future development.

tpadioleau: > ```
> spack spec --fresh kokkos@5
>  -   kokkos@5.0.0~aggressive_vectorization~cmake_lang+complex_align~cuda~debug~debug_bounds_check~hip_relocatable_device_code~hpx~hwloc~ipo~openmp~openmptarget~pic~rocm+serial+shared~sycl~tests~threads~tuning~wrapper build_system=cmake build_type=Release cxxstd=23 generator=make intel_gpu_arch=none platform=linux os=ubuntu20.04 target=icelake %cxx=gcc@10.5.0
> ```
> 
> `cxxstd=23` wut ? @cedricchevalier19 could you try ?

With the default value to 20, this does not happen.

tpadioleau: otherwise:
```
variant("cxxstd", default="17", values=("14", "17", "20",), when="@3", multi=False, description="C++ standard",)
variant("cxxstd", default="17", values=("17", "20", "23"), when="@4", multi=False, description="C++ standard",)
variant("cxxstd", default="20", values=("20", "23"), when="@5", multi=False, description="C++ standard",)
```

rbberger: @tpadioleau while explicit, it might be the best solution.

tpadioleau: I have updated the package, let me know what you think

repos/spack_repo/builtin/packages/kokkos/package.py
```diff
@@ -305,19 +323,13 @@ class Kokkos(CMakePackage, CudaPackage, ROCmPackage):
...
-        )
+    conflicts("+sycl", when="cxxstd=14", msg="SYCL requires C++17 or higher")
+    conflicts("+openmptarget", when="cxxstd=14", msg="OpenMPTarget requires C++17 or higher")
 
     # HPX should use the same C++ standard
-    for cxxstd in cxxstds:
-        depends_on("hpx cxxstd={0}".format(cxxstd), when="+hpx cxxstd={0}".format(cxxstd))
+    for cxxstd in ["14", "17", "20"]:
+        depends_on(f"hpx cxxstd={cxxstd}", when=f"+hpx cxxstd={cxxstd}")
+    conflicts("+hpx", when="cxxstd=23", msg="hpx package does not yet support cxxstd=23")
```
tpadioleau: Should I open an issue to ask the HPX package maintainers to add C++23 ?

cedricchevalier19 (approved):

repos/spack_repo/builtin/packages/kokkos/package.py
rbberger: LGTM

rbberger: We got one more suggestion on Slack for extracting common parts:
```python
with default_args(multi=False, description="C++ standard"):
    variant("cxxstd", default="17", values=("14", "17", "20"), when="@3")
    variant("cxxstd", default="20", values=("20", "23"), when="@5")
```
@tpadioleau can you try that out?

tpadioleau:
In a previous gitlab run, there was an issue with vtk-m, likely not supporting C++20 yet (but also likely not a big bug). What should we do about it ?

cedricchevalier19:
> In a previous gitlab run, there was an issue with vtk-m, likely not supporting C++20 yet (but also likely not a big bug). What should we do about it ?

Was it with kokkos 4? If so, they should add a conflict in their recipe.
 I am not sure that it is maintained now as the upstream project is now viskores.

repos/spack_repo/builtin/packages/kokkos/package.py
tpadioleau: https://github.com/spack/spack-packages/issues/2944

tpadioleau:
> > In a previous gitlab run, there was an issue with vtk-m, likely not supporting C++20 yet (but also likely not a big bug). What should we do about it ?
> 
> Was it with kokkos 4? If so, they should add a conflict in their recipe. I am not sure that it is maintained now as the upstream project is now viskores.

It was Kokkos 5 that triggered the use of C++20.

lucbv:
We can reach out to the vtk-m/viskores maintainers if necessary, considering the project went through some changes they might want to clean things up and deprecate old things as well.
I like the current changes, it looks clean, thanks @tpadioleau !

lucbv (approved):

cedricchevalier19 (approved):

tpadioleau:
Given https://github.com/spack/spack-packages/pull/2945, I am in favor to wait.

rbberger (approved):

rbberger:
@tpadioleau more cleanup: https://github.com/spack/spack-packages/pull/2994

rbberger:
@tpadioleau fingers crossed for the next rebase ;-)

PR #2722, opened 2025-12-10
blast-plus updates

ebagrenrut:
- Add 2.17.0
- Per the [BLAST+ release notes](https://www.ncbi.nlm.nih.gov/books/NBK131777/):
    - BLAST+ 2.15.0 introduced a dependency on SQLite. I found that when building on a host without any sqlite3 devel package(s) installed, the build does not fail, but most of the BLAST+ binaries (`blastn`, `blastp`, etc) are not built, which makes BLAST+ pretty useless. Thus, we add a dependency on `sqlite`.
    - BLAST+ 2.17.0 introduced a dependency on Zstandard. Thus, we add a dependency on `zstd`.

@weijianwen

### DIFF
--- a/repos/spack_repo/builtin/packages/blast_plus/package.py
+++ b/repos/spack_repo/builtin/packages/blast_plus/package.py
@@ -108,17 +108,17 @@ class BlastPlus(AutotoolsPackage):
             config_args.extend(["--with-dll", "--without-static", "--without-static-exe"])
 
         if spec.satisfies("+jpeg"):
-            config_args.append("--with-jpeg={self.spec['jpeg'].prefix}")
+            config_args.append(f"--with-jpeg={self.spec['jpeg'].prefix}")
         else:
             config_args.append("--without-jpeg")
 
         if spec.satisfies("+png"):
-            config_args.append("--with-png={self.spec['libpng'].prefix}")
+            config_args.append(f"--with-png={self.spec['libpng'].prefix}")
         else:
             config_args.append("--without-png")
 
         if spec.satisfies("+freetype"):
-            config_args.append("--with-freetype={self.spec['freetype'].prefix}")
+            config_args.append(f"--with-freetype={self.spec['freetype'].prefix}")
         else:
             config_args.append("--without-freetype")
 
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/blast_plus/package.py
```diff
@@ -168,4 +171,10 @@ def configure_args(self):
         else:
             config_args.append("--without-python")
 
+        with when("@2.15:"):
+            config_args.append("--with-sqlite={0}".format(self.spec["sqlite"].prefix))
+
+        with when("@2.17:"):
+            config_args.append("--with-zstd={0}".format(self.spec["zstd"].prefix))
```
tldahlgren: Nit: You could use f-strings here, e.g., 


```suggestion
            config_args.append(f"--with-sqlite={self.spec["sqlite"].prefix}")

        with when("@2.17:"):
            config_args.append(f"--with-zstd={self.spec["zstd"].prefix}")
```

tldahlgren (commented):
Confirmed the new version sha256.

@weijianwen Do you have objections or additional feedback for this PR before it is merged?

tldahlgren (dismissed):

repos/spack_repo/builtin/packages/blast_plus/package.py
ebagrenrut: @tldahlgren I considered that, though all of the other strings in the package used `.format()`, so I followed that pattern. If you'd prefer, I could change all of the formatted strings to f-strings.

tldahlgren: Makes sense. You can leave it.

I'll give the maintainer some time to comment before merging.

ebagrenrut: Sorry, I already changed it :D

tldahlgren: 🤦 Thanks for catching the double-to-single quote correction in my suggestion.

repos/spack_repo/builtin/packages/blast_plus/package.py
```diff
@@ -105,67 +108,73 @@ def configure_args(self):
             config_args.extend(["--with-dll", "--without-static", "--without-static-exe"])
 
         if spec.satisfies("+jpeg"):
-            config_args.append("--with-jpeg={0}".format(self.spec["jpeg"].prefix))
+            config_args.append("--with-jpeg={self.spec['jpeg'].prefix}")
```
tldahlgren: You need to prepend `f` to each of these strings in order for the substitution to work (see https://docs.python.org/3/tutorial/inputoutput.html).

tldahlgren (changes requested):
The `f` is required for value substitution.

tldahlgren (changes requested):
The `f` is required for value substitution.

repos/spack_repo/builtin/packages/blast_plus/package.py
ebagrenrut: Gah! You're right, of course. Trying again...

tldahlgren (approved):
LGTM

ebagrenrut:
@tldahlgren I suspect the maintainer is not going to respond. Are you willing to merge this? Thanks!

PR #2727, opened 2025-12-10
Add procps dependency to lmod

ebagrenrut:
Way back in Lmod 6.0.12, with [this commit](https://github.com/TACC/Lmod/commit/0ddb7bd2fdc462915205bf4855b18b9229a57548), Lmod added a dependency on the `ps` command. I encountered a failure to build Lmod when building in a container that did not have `ps` available. Thus, we add a dependency on `procps`.

### DIFF
--- a/repos/spack_repo/builtin/packages/lmod/package.py
+++ b/repos/spack_repo/builtin/packages/lmod/package.py
@@ -69,6 +69,8 @@ class Lmod(AutotoolsPackage):
 
     depends_on("pkgconfig", type="build")
 
+    depends_on("procps", type="build")
+
     # GNU sed is required instead of bsd sed on macOS
     depends_on("sed", type="build", when="platform=darwin")
 


repos/spack_repo/builtin/packages/lmod/package.py
```diff
@@ -69,6 +69,8 @@ class Lmod(AutotoolsPackage):
 
     depends_on("pkgconfig", type="build")
 
+    depends_on("procps", type="build")
```
tldahlgren, line 72: Is this dependency for all of the listed versions or from a specific version on?

tldahlgren (commented):
Have a question.

repos/spack_repo/builtin/packages/lmod/package.py
ebagrenrut: @tldahlgren Yes, the dependency on `ps` was added to LMod 6.0.12, while the oldest version in the Spack package is 6.3.7.

tldahlgren (approved):

PR #2733, opened 2025-12-11
Fix preseq package and add 3.2.0

ebagrenrut:
Though I didn't initially intend it to be so, this is effectively an overhaul of the `preseq` package.

- Add `preseq` 3.2.0
- Add a missing `libefence` dependency for `preseq` 2.x and earlier. The `libefence` requirement was dropped in `preseq` 3.0.
- Remove the `samtools` dependency. `preseq` 3 does not use `samtools` while building. It turns out that `preseq` 2 does not have an external `samtools` dependency at all, and the Spack-built `samtools` is not sufficient in any case (as it merely builds and isntalls the samtools binaries and not the samtools source code). `preseq` 2 requires the raw `samtools` source, and it includes it in its release tarball.
- Constrain `gsl` dependency to `preseq` 3.0.2 and earlier. Per the `preseq` changelog, 3.0.2 does not use `gsl` at all.
- The `Preseq` class now also extends the `AutotoolsPackage` class. While `preseq` 2.x and earlier simply use a Makefile, `preseq` 3.0 adopted autotools. When building `preseq` 2.x, select functions of `AutotoolsPackage`/`AutotoolsBuilder` are overridden to ensure that they do not cause the build to fail for autotools-related reason (e.g. a missing `configure` script).
- Add the `hts` variant. `preseq` 3.0 and newer can optionally use `htslib` while converting BAM inputs to mapped read format. Despite using autotools, one cannot pass the location of `htslib`'s includes and libs via options to `configure` and must add `CPPFLAGS` and `LDFLAGS`, respectively.
- When `+hts`, add a conflict with `gcc@13:`. The `preseq` code that interfaces with HTSlib does not compile beginning with gcc 13.

### DIFF
--- a/repos/spack_repo/builtin/packages/preseq/package.py
+++ b/repos/spack_repo/builtin/packages/preseq/package.py
@@ -72,5 +72,6 @@ class Preseq(MakefilePackage, AutotoolsPackage):
     def configure(self, spec, prefix):
         return
 
-    with when("+hts"):
-        configure_args=["--enable-hts"]
+    @when("+hts")
+    def configure_args(self):
+        return ["--enable-hts"]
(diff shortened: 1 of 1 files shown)


kwryankrattiger (approved):
LGTM

PR #2734, opened 2025-12-11
memtailor: add v1.1

d-torrance:
The latest version adds C++17 support, so we may use googletest 1.17.0.  Also, a tarball is now released, so we use it instead of a git commit.


RikkiButler20 (approved):
Looks good to me

Issue #2658, opened 2025-12-04
Installation issue: libfabric - CUDA runtime not available.


skwde:
### Steps to reproduce the issue

```console
$ spack spec -l libfabric+cuda+gdrcopy+uring fabrics=sockets,tcp,udp,verbs,shm,ucx,mlx,xpmem cuda_arch=80,89 %gcc@12.5.0 ^cuda@12.2.2
 -   zc6yq7t  libfabric@2.3.1~asan+cuda~debug+gdrcopy~kdreg~level_zero~lsan~rocm~tsan~ubsan+uring build_system=autotools cuda_arch:=80,89 fabrics:=mlx,shm,sockets,tcp,ucx,udp,verbs,xpmem platform=linux os=rocky8 target=x86_64_v3 %c=gcc@12.5.0
[+]  eiy76kp      ^compiler-wrapper@1.0 build_system=generic platform=linux os=rocky8 target=x86_64_v3 
[+]  2wbluft      ^cuda@12.2.2~allow-unsupported-compilers~dev build_system=generic platform=linux os=rocky8 target=x86_64_v3 
[+]  bqvcwsc          ^libxml2@2.13.5~http+pic~python+shared build_system=autotools platform=linux os=rocky8 target=x86_64_v3 %c=gcc@8.5.0
[+]  vu6462b              ^libiconv@1.18 build_system=autotools libs:=shared,static platform=linux os=rocky8 target=x86_64_v3 %c=gcc@8.5.0
[+]  2lsfyjp              ^xz@5.6.3~pic build_system=autotools libs:=shared,static platform=linux os=rocky8 target=x86_64_v3 %c=gcc@8.5.0
[+]  a7o54p7      ^gcc@12.5.0+binutils+bootstrap~graphite~mold~nvptx~piclibs~profiled~strip build_system=autotools build_type=RelWithDebInfo languages:='c,c++,fortran' platform=linux os=rocky8 target=x86_64_v3 %c,cxx=gcc@8.5.0
[+]  xy6t6pc          ^binutils@2.45~debuginfod+gas~gold~gprofng~headers~interwork+ld~libiberty~lto~nls~pgo+plugins build_system=autotools compress_debug_sections=zlib libs:=shared,static platform=linux os=rocky8 target=x86_64_v3 %c,cxx=gcc@8.5.0
[...]

PR #2736, opened 2025-12-12
libfabric: add variants for gdrcopy-dlopen and cuda-dlopen

brandongc:
These variants expose additional libfabric configuration options.

They allow e.g. `libcudart.so` to be `dlopen`'d at runtime.

### DIFF
--- a/repos/spack_repo/builtin/packages/libfabric/package.py
+++ b/repos/spack_repo/builtin/packages/libfabric/package.py
@@ -113,6 +113,8 @@ class Libfabric(AutotoolsPackage, CudaPackage, ROCmPackage):
     variant("uring", default=False, when="@1.17.0:", description="Enable uring support")
     variant("level_zero", default=False, description="Enable Level Zero support")
     variant("gdrcopy", default=False, when="@1.12: +cuda", description="Enable gdrcopy support")
+    variant("cuda_dlopen", default=False, description="Enable dlopen of CUDA libraries")
+    variant("gdrcopy_dlopen", default=False, description="Enable dlopen of gdr libraries")
 
     variant("asan", default=False, when="@1.12:", description="Enable AddressSanitizer (ASan)")
     variant("lsan", default=False, when="@1.20:", description="Enable LeakSanitizer (LSan)")
@@ -231,6 +233,8 @@ class Libfabric(AutotoolsPackage, CudaPackage, ROCmPackage):
     def configure_args(self):
         args = [
             *self.enable_or_disable("debug"),
+            *self.enable_or_disable("cuda_dlopen"),
+            *self.enable_or_disable("gdrcopy_dlopen"),
             *self.enable_or_disable("asan"),
             *self.enable_or_disable("lsan"),
             *self.enable_or_disable("tsan"),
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/libfabric/package.py
```diff
@@ -113,6 +113,8 @@ class Libfabric(AutotoolsPackage, CudaPackage, ROCmPackage):
     variant("uring", default=False, when="@1.17.0:", description="Enable uring support")
     variant("level_zero", default=False, description="Enable Level Zero support")
     variant("gdrcopy", default=False, when="@1.12: +cuda", description="Enable gdrcopy support")
+    variant("cuda_dlopen", default=False, description="Enable dlopen of CUDA libraries")
+    variant("gdrcopy_dlopen", default=False, description="Enable dlopen of gdr libraries")
```
msimberg: Do you know if these are dependent on the regular (`cuda`/`gdrcopy`) variants being enabled as well, or are they in fact mutually exclusive? Either case would be good to encode here.

darrylabbate: Libfabric requires header files to be available for `enable-{cuda,gdrcopy}-dlopen` for the build, but not the libraries themselves.

brandongc: So the suggestion is to add the following?

```
depends_on("gdrcopy", when="+gdrcopy_dlopen")
depends_on("cuda", when="+cuda_dlopen")
```

msimberg: I did a bit of digging and as far as I can tell* at least `--with-cuda-dlopen` requires `--with-cuda` to be set as well (I assume it's the same for gdrcopy), so I would set this to:
```suggestion
    variant("cuda_dlopen", default=False, when="+cuda", description="Enable dlopen of CUDA libraries")
    variant("gdrcopy_dlopen", default=False, when="+gdrcopy", description="Enable dlopen of gdr libraries")
```
That would make sure the `depends_on` are correctly set up from `+cuda` and `+gdrcopy` already.

\*CUDA dlopen support requires `HAVE_CUDA` to be enabled (https://github.com/ofiwg/libfabric/blob/f6888542d3e8f6512481773d344544fa96d377a5/src/hmem_cuda.c#L42-L48) and from testing configuring libfabric with `--without-cuda` and `--with-cuda-dlopen` `HAVE_CUDA` does not get set in that situation (https://github.com/ofiwg/libfabric/blob/f6888542d3e8f6512481773d344544fa96d377a5/configure.ac#L748-L767).

As far as I can tell, with `--with-cuda` and `--with-cuda-dlopen`, the configure script does not require CUDA symbols to be available (https://github.com/ofiwg/libfabric/blob/f6888542d3e8f6512481773d344544fa96d377a5/configure.ac#L751-L757). Would you mind testing if this setup works for https://github.com/spack/spack-packages/issues/2658?

If yes, I think you could simply add the same dlopen option for `rocr` as well.

brandongc: This works without issue in my environment. @skwde might be the person confirm for #2658.

I don't have any HW to test rocr, but the change is easy enough that's ok and desired.

msimberg: Thanks @brandongc! Let's give @skwde some time to comment there before merging, but this looks good to me now.

msimberg (approved):
Let's go ahead with this and wait for separate confirmation before closing #2658.

PR #2738, opened 2025-12-12
llvm-openmp: avoid fortran dependency

rbberger:
`LIBOMP_FORTRAN_MODULES` is `OFF` by default. This adds a `fortran` variant which follows that default.


rbberger:
@spackbot fix style

becker33 (approved):

PR #2741, opened 2025-12-12
Update heat package

LeonKaem:
Update the py-heat package to include versions higher than heat 1.4 as well as to enable simple cuda and rocm support via `+cuda` and `+rocm`  in spec.


ClaudiaComito (approved):
Thanks a lot @LeonKaem!

JuanPedroGHM (approved):

repos/spack_repo/builtin/packages/py_heat/package.py
```diff
@@ -36,30 +39,85 @@ class PyHeat(PythonPackage):
...
     with when("@1.3"):
         depends_on("python@3.8:3.10", type=("build", "run"))
         depends_on("py-mpi4py@3:", type=("build", "run"))
         depends_on("py-numpy@1.20:1", type=("build", "run"))
-        depends_on("py-torch@1.8:2.0.1", type=("build", "run"))
         depends_on("py-scipy@0.14:", type=("build", "run"))
         depends_on("pil@6:", type=("build", "run"))
         depends_on("py-torchvision@0.8:", type=("build", "run"))
 
+        with when("~cuda~rocm"):
```
alalazo: I think you can avoid this `when` context manager (the package seems to always depend on `py-torch@1.8:2.0.1`, it just gains more constraints when `+cuda` or `+rocm`)

repos/spack_repo/builtin/packages/py_heat/package.py
```diff
@@ -36,30 +39,85 @@ class PyHeat(PythonPackage):
...
     with when("@1.4"):
         depends_on("python@3.8:3.11", type=("build", "run"))
         depends_on("py-mpi4py@3:", type=("build", "run"))
         depends_on("py-numpy@1.22:1", type=("build", "run"))
-        depends_on("py-torch@1.11:2.3.2", type=("build", "run"))
         depends_on("py-scipy@1.10:", type=("build", "run"))
         depends_on("pil@6:", type=("build", "run"))
         depends_on("py-torchvision@0.12:", type=("build", "run"))
 
+        with when("~cuda~rocm"):
```
alalazo: Same

repos/spack_repo/builtin/packages/py_heat/package.py
```diff
@@ -36,30 +39,85 @@ class PyHeat(PythonPackage):
...
+    with when("@1.5"):
+        depends_on("python@3.9:3.12", type=("build", "run"))
+        depends_on("py-mpi4py@3:", type=("build", "run"))
+        depends_on("py-numpy@1.22:1", type=("build", "run"))
+        depends_on("py-scipy@1.10:", type=("build", "run"))
+        depends_on("pil@6:", type=("build", "run"))
+        depends_on("py-typing-extensions", type=("build", "run"))
+        depends_on("py-torchvision@0.15.2:0.21.1", type=("build", "run"))
+
+        with when("~cuda~rocm"):
```
alalazo: Also here

repos/spack_repo/builtin/packages/py_heat/package.py
```diff
@@ -36,30 +39,85 @@ class PyHeat(PythonPackage):
...
+
+    with when("@1.6"):
+        depends_on("python@3.10:", type=("build", "run"))
+        depends_on("py-mpi4py@3:", type=("build", "run"))
+        depends_on("py-scipy@1.14:", type=("build", "run"))
+        depends_on("pil@6:", when=("+examples"), type=("build", "run"))
+        depends_on("py-typing-extensions", type=("build", "run"))
+        depends_on("py-torchvision@0.15:", type=("build", "run"))
+
+        with when("~cuda~rocm"):
```
alalazo: and here

alalazo (commented):
Basically LGTM. I think we can remove the `when` context managers in the `~cuda~rocm` cases.

repos/spack_repo/builtin/packages/py_heat/package.py
LeonKaem: if I understand this correctly, if I specify multiple specs of the same dependency the more specific is selected because it fulfills both dependencies?
So this
```python
depends_on("py-torch@2.0:2.8.0", type=("build", "run"))
depends_on("py-torch@2.0:2.8.0+cuda", type=("build", "run"))
```
would have the same result as just writing
```python
depends_on("py-torch@2.0:2.8.0+cuda", type=("build", "run"))
```
Is that correct?

alalazo: In the case above what you say is true. In general, all conditions that apply are imposed on the spec, so for instance:
```python
depends_on("py-torch@2.0:2.8.0")
depends_on("py-torch+cuda")
```
is equivalent to:
```python
depends_on("py-torch@2.0:2.8.0+cuda")
```
Dependencies with a `when=` are applied only if the `when=` condition is satisfied.

LeonKaem: Ah, I see, that is convenient

alalazo (approved):

PR #2743, opened 2025-12-12
cubelib: set --with-nocross-compiler-suite in configure step

sebastiankreutzer:
Cubelib currently does not respect the selected compiler and will fall back to GCC, even if LLVM is specified.
This is because `--with-nocross-compiler-suite` is not set in the `configure` command.
This patch fixes this, adopting the implementation already present in the `scorep` package.
I've simply copied `clean_compiler` from there, not sure if there is a cleaner option that avoids code duplication.
@wrwilliams

### DIFF
--- a/repos/spack_repo/builtin/packages/cubelib/package.py
+++ b/repos/spack_repo/builtin/packages/cubelib/package.py
@@ -33,7 +33,21 @@ class Cubelib(AutotoolsPackage):
 
         return url.format(version.up_to(2), version)
 
+    def clean_compiler(self, compiler):
+        renames = {
+            "cce": "cray",
+            "rocmcc": "amdclang",
+            "intel-oneapi-compilers": "oneapi",
+            "llvm": "clang",
+        }
+        if compiler in renames:
+            return renames[compiler]
+        return compiler
+
+
     def configure_args(self):
+        cname = self.clean_compiler(self.spec.compiler.name)
+        configure_args.append("--with-nocross-compiler-suite={0}".format(cname))
         configure_args = ["--enable-shared"]
         configure_args.append("--with-frontend-zlib=%s" % self.spec["zlib-api"].prefix.lib)
         return configure_args
--- a/repos/spack_repo/builtin/packages/cubelib/package.py
+++ b/repos/spack_repo/builtin/packages/cubelib/package.py
@@ -46,10 +46,10 @@ class Cubelib(AutotoolsPackage):
 
 
     def configure_args(self):
-        cname = self.clean_compiler(self.spec.compiler.name)
-        configure_args.append("--with-nocross-compiler-suite={0}".format(cname))
         configure_args = ["--enable-shared"]
         configure_args.append("--with-frontend-zlib=%s" % self.spec["zlib-api"].prefix.lib)
+        cname = self.clean_compiler(self.spec.compiler.name)
+        configure_args.append("--with-nocross-compiler-suite={0}".format(cname))
         return configure_args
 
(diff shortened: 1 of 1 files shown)


sebastiankreutzer:
@spackbot fix style

repos/spack_repo/builtin/packages/cubelib/package.py
```diff
@@ -33,7 +33,20 @@ def url_for_version(self, version):
...
+        renames = {
+            "cce": "cray",
+            "rocmcc": "amdclang",
+            "intel-oneapi-compilers": "oneapi",
+            "llvm": "clang",
+        }
+        if compiler in renames:
+            return renames[compiler]
+        return compiler
+
```
wrwilliams, line 46: I am also unsure what to do with this as a multi-project convenience function; when a Spack maintainer has a look I would happily take suggestions to avoid the duplication.

repos/spack_repo/builtin/packages/cubelib/package.py
```diff
@@ -33,7 +33,20 @@ def url_for_version(self, version):
...
     def configure_args(self):
+        cname = self.clean_compiler(self.spec.compiler.name)
+        configure_args.append("--with-nocross-compiler-suite={0}".format(cname))
```
wrwilliams: Does `cubew` need the same treatment while we're on this?

wrwilliams (commented):
Forward progress as-is, can probably be improved as noted, I have no strong opinion re: one or many PRs.

repos/spack_repo/builtin/packages/cubelib/package.py
sebastiankreutzer: Yeah, seems to be the same issue there.

Thyre: OPARI2 as well, only OTF2 gets the correct compilers by hard-setting the compilers. Not ideal, but should be fine.

repos/spack_repo/builtin/packages/cubelib/package.py
```diff
@@ -33,9 +33,22 @@ def url_for_version(self, version):
...
+        }
+        if compiler in renames:
+            return renames[compiler]
+        return compiler
+
     def configure_args(self):
         configure_args = ["--enable-shared"]
         configure_args.append("--with-frontend-zlib=%s" % self.spec["zlib-api"].prefix.lib)
+        cname = self.clean_compiler(self.spec.compiler.name)
+        configure_args.append("--with-nocross-compiler-suite={0}".format(cname))
```
bernhardkaindl, line 51: Not for now, but if you need to make some change, consider using an f-string instead.

bernhardkaindl (approved):

bernhardkaindl (approved):
LGTM

repos/spack_repo/builtin/packages/cubelib/package.py
bernhardkaindl: It seems different packages need different compiler maps. `compiler_map` would likely be a better name, and `clean_compiler()` sounds a bit misleading to me. For example, `map_compiler_name()` might be more informative in the future.

However, you can add such changes and changes to use f-strings when you update the package next time. I'm closing this conversation; it's a precondition for merging the PR.

repos/spack_repo/builtin/packages/cubelib/package.py
bernhardkaindl: @sebastiankreutzer, you can open new PRs for these issues.

PR #2744, opened 2025-12-12
Compiler wrapper: Fix for calling compiler wrapper with non-empty 'mode' variable

sibcse:
Where the compiler wrapper determines the `mode`, the `mode` variable was not emptied beforehand. This caused an issue: I was creating a Makefile package, where the Makefile takes an argument that was unfortunately also called `mode`, i.e., in `package.py`:
```
@property
def build_targets(self):
    return ["mode=my_arg"]
```
(this is exactly as explained in the docs: [https://spack.readthedocs.io/en/latest/build_systems/makefilepackage.html#command-line-arguments](https://spack.readthedocs.io/en/latest/build_systems/makefilepackage.html#command-line-arguments))

It turns out that the variable `mode` is then propagated to the compiler wrapper, such that the compiler wrapper script did not determine the correct mode (it was kept at the value that was passed to the Makefile). This finally prevented applying the correct `SPACK_TARGET_ARGS_FORTRAN`, which is how I noticed this in the first place. You can check this propagation easily: Create a tiny Makefile:
```make
all:
	./script.sh
```
and script.sh with `echo "var = ${var}"`; then just call `make var=some_value`. It will print `var = some_value`.

In this PR, `mode` is just set to an empty string before it is actually determined.

### DIFF
--- a/repos/spack_repo/builtin/packages/compiler_wrapper/cc.sh
+++ b/repos/spack_repo/builtin/packages/compiler_wrapper/cc.sh
@@ -296,6 +296,9 @@ fi
 # including version checks (SPACK_XFLAGS variants are not applied
 # for version checks).
 mode=""
+vdep=""
+lang_flags=""
+debug_flags=""
 command="${0##*/}"
 comp="CC"
 vcheck_flags=""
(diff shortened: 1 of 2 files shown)


sibcse:
@haampie May you perhaps review this PR, or add some other suitable reviewers? Thanks a lot!

haampie:
Thanks for this! That looks like a serious oversight.

Can you extend to

```
mode=""
vdep=""
lang_flags=""
debug_flags=""
```

Feel also free to add me as a maintainer in package.py

haampie (dismissed):
Thanks for catching this!

PR #2732, opened 2025-12-11
adds v2025.09.30 to arkouda and py_arkouda

ajpotts:
## PR Description

This PR adds the **v2025.09.30** release of both `arkouda` and `py-arkouda` to the Spack builtin repository and updates dependency constraints to align with the versions tested in CI and supported by the upstream project.

### 1. Add new versions
- Introduces `arkouda@2025.09.30`
- Introduces `py-arkouda@2025.09.30`

Both entries include verified `sha256` checksums.

## Dependency updates (Python, Chapel, Arrow)

This PR formalizes compatibility windows that were previously implicit, ensuring that Spack concretizes environments consistent with upstream-supported versions.

### Python support aligned with CI

- **≤ 2025.01.13:** `python@3.9:3.12.3`
- **2025.07.03–2025.08.20:** `python@3.9:3.13`
- **2025.09.30+:** `python@3.10:3.13`

### Updated Chapel ranges

- `2025.07.03–2025.08.20`: `chapel@2.0:2.4 +hdf5 +zmq`
- `2025.09.30+`: `chapel@2.0:2.5 +hdf5 +zmq`

### Arrow version alignment

For newer releases (`2025.07.03+`):

```
arrow@15:19 +brotli +bz2 +lz4 +parquet +snappy +zlib +zstd
```

PR #2749, opened 2025-12-12
chapel: add v2.7

e-kayrakli:
Add the Chapel 2.7 release.

Besides the new version, this PR:

- Adds a conflict with CUDA 12.9 which triggers a lot deprecation warnings that need to be fixed in the Chapel runtime
- Adds an `import` (and a `TODO`) to avoid a deprecation warning with newer Package APIs

### DIFF
--- a/repos/spack_repo/builtin/packages/chapel/package.py
+++ b/repos/spack_repo/builtin/packages/chapel/package.py
@@ -11,7 +11,6 @@ from spack_repo.builtin.build_systems.cuda import CudaPackage
 from spack_repo.builtin.build_systems.rocm import ROCmPackage
 
 from spack.package import *
-from spack.util.environment import is_system_path  # avoid deprecation warnings
 
 
 def slingshot_network():
--- a/repos/spack_repo/builtin/packages/chapel/package.py
+++ b/repos/spack_repo/builtin/packages/chapel/package.py
@@ -68,7 +68,7 @@ class Chapel(AutotoolsPackage, CudaPackage, ROCmPackage):
     version(
         "2.7.0",
         url="https://chapel-lang.org/tmp/chapel-2.7.0.tar.gz",
-        sha256="85e2fe7fcffb7f71e9173abb455678c14a1a8aa4997e10ed4d96d29d81672d0e",
+        sha256="5e3269babdae334c80fc3f25114698fdfe53e84ea06626af22d2b54eeb75bee6",
     )
     version("2.6.0", sha256="e469c35be601cf1f59af542ab885e8a14aa2b087b79af0d5372a4421976c74b6")
     version("2.5.0", sha256="020220ca9bf52b9f416e9a029bdc465bb1f635c1e274c6ca3c18d1f83e41fce1")
--- a/repos/spack_repo/builtin/packages/chapel/package.py
+++ b/repos/spack_repo/builtin/packages/chapel/package.py
@@ -483,6 +483,7 @@ class Chapel(AutotoolsPackage, CudaPackage, ROCmPackage):
 
     conflicts("+rocm", when="+cuda", msg="Chapel must be built with either CUDA or ROCm, not both")
 
+    # CUDA conflicts and dependencies
     conflicts(
         "^llvm@20",
         when="@:2.5 +cuda",
@@ -490,6 +491,9 @@ class Chapel(AutotoolsPackage, CudaPackage, ROCmPackage):
         "https://github.com/chapel-lang/chapel/issues/27273",
     )
 
+    conflicts("cuda@12.9:", when="+cuda") # deprecation warnings otherwise
+
+    # ROCm conflicts and dependencies
     conflicts("+rocm", when="@:2.1", msg="ROCm support in spack requires Chapel 2.2.0 or later")
     # Chapel restricts the allowable ROCm versions
     with when("@2.2: +rocm"):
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/chapel/package.py
```diff
@@ -10,7 +10,9 @@
 from spack_repo.builtin.build_systems.cuda import CudaPackage
 from spack_repo.builtin.build_systems.rocm import ROCmPackage
 
+# TODO be more selective in importing packages
 from spack.package import *
```
alecbcs: We're not importing other packages here. This is importing the common methods and directives for a package.py like `version()` and `variant()`. Although you could be more selective here, I wouldn't recommend it. The standard Spack practice/format is to import all utilities in all packages.

e-kayrakli: Thanks, I'll drop the comment, then. I should have looked at other packages out there.

If it wasn't clear from the next line in the diff, I put that comment there while adding a new import to grab the non-deprecated `is_system_path` to avoid deprecation warnings. As you have not commented on that next line, I am guessing that's fine. Although on a cursory look, I don't see other packages that use `is_system_path` to have added that line, probably accepting the deprecation warnings.

e-kayrakli:
@spackbot fix style

bonachea (commented):
I've successfully installed this new release on several systems, including NERSC Perlmutter and OLCF Frontier, using a variety of relevant configurations settings.

arifthpe:
I see [the gitlab pipeline](https://gitlab.spack.io/spack/spack-packages/-/pipelines/1354253) had some time outs in queue waiting for darwin runners, as well as a [true failure in `e4s-generate`](https://gitlab.spack.io/spack/spack-packages/-/jobs/19630865). The latter occurs after our deprecation of version 2.4 with:
```
RuntimeError: failed to concretize `arkouda` for the following reasons:
     1. Cannot satisfy 'chapel@2.0:2.4' 1(2.7.0)
     2. Cannot satisfy 'chapel@2.0:2.4' 4(2.7.0)
        required because arkouda depends on chapel@2.0:2.4+hdf5+zmq when @2025.07.03:2025.8.20 
          required because arkouda requested explicitly 
```

Until https://github.com/spack/spack-packages/pull/2732 (merged yesterday), the latest version of Arkouda had the `chapel@2.0:2.4` dependency; now the latest accepts `chapel@2.0:2.5`.

Is it the case that _every_ (non-deprecated) version of every package must be able to concretize without deprecated dependencies, or just the _latest_ version? If it's the latter then I think we'd just need to rebase off `develop` to incorporate that Arkouda PR. Or maybe I'm misunderstanding what `e4s-generate` is doing.

bonachea:
@arifthpe Thanks for looking into this
> I see [the gitlab pipeline](https://gitlab.spack.io/spack/spack-packages/-/pipelines/1354253) had some time outs in queue waiting for darwin runners, as well as a [true failure in `e4s-generate`](https://gitlab.spack.io/spack/spack-packages/-/jobs/19630865). 
...
> Is it the case that _every_ (non-deprecated) version of every package must be able to concretize without deprecated dependencies, or just the _latest_ version? If it's the latter then I think we'd just need to rebase off `develop` to incorporate that Arkouda PR. Or maybe I'm misunderstanding what `e4s-generate` is doing.

IIUC it's technically neither behavior. I believe what this pipeline is doing is concretizing [this spack.yaml](https://github.com/spack/spack-packages/blob/develop/stacks/e4s/spack.yaml) which includes `arkouda` (with no version number of variants). Thus it's attempting to concretize the default version of `arkouda` (so effectively your latter option). As you say, on this branch that currently appears to be `arkouda@2025.08.20`, which as you say depends on `"chapel@2.0:2.4` that is now deprecated. As you mention, on the develop branch the default version appears to be `arkouda@2025.09.30`, which supports `chapel@2.0:2.5`.

TL:DR; yes, we should probably rebase this branch

arifthpe:
With the update to use the 2.7.0 release tarball in its permanent github-release location, this is out of pre-release and ready for review and merge (pending CI checks ofc).

arifthpe (approved):

arifthpe (approved):

bonachea (approved):
LGTM and works as expected in manual testing.
Thanks for all the work on this!

arezaii (approved):

becker33 (commented):
I'm checking with other core devs about whether the correct solution with `is_system_path` is to undeprecate it in the API, rather than this import that bypasses the defined API layer.

e-kayrakli:
@becker33 -- Thanks. Is there any update on the subject? If it is a point of contention, we can definitely remove the `import` and accept the deprecation warnings. Let me know.

e-kayrakli:
@becker33 / others -- given that the import I had in the previous version was more controversial than I thought, I just removed it after discussing with other maintainers of the package.

bonachea (approved):
LGTM

arifthpe (approved):

arezaii (approved):

scheibelp (approved):
@becker33 it looks like your concern in https://github.com/spack/spack-packages/pull/2749#pullrequestreview-3645837734 has been resolved by removing the controversial code, so I'm approving.

bradcray:
Thank you, Peter!

Issue #2728, opened 2025-12-11
Installation issue: hdf-eos2


danrosen25:
### Steps to reproduce the issue

Download `hdf-eos2-3.0-src.tar.gz` from https://wiki.earthdata.nasa.gov/spaces/DAS/pages/202211788/Toolkit+Downloads#ToolkitDownloads-HDFEOSHDF-EOS

This can also be done via cloning the git repo containing this tarball
`git clone -b HDFEOS2_3.0 https://git.earthdata.nasa.gov/scm/das/hdfeos.git`
or using wget
`wget "https://git.earthdata.nasa.gov/projects/DAS/repos/hdfeos/raw/hdf-eos2-3.0-src.tar.gz?at=refs%2Fheads%2FHDFEOS2_3.0` and then renaming the file

```sh
spack install hdf-eos2
```

```console
 -   hdf-eos2@3.0+shared+static build_system=autotools patches:=5878c11 platform=linux os=ubuntu25.04 target=m1 %c=gcc@14.2.0
[+]      ^compiler-wrapper@1.0 build_system=generic platform=linux os=ubuntu25.04 target=m1 
[e]      ^gcc@14.2.0~binutils+bootstrap~graphite~mold~nvptx~piclibs~profiled~strip build_system=autotools build_type=RelWithDebInfo languages:='c,c++,fortran' platform=linux os=ubuntu25.04 target=aarch64 
[+]      ^gcc-runtime@14.2.0 build_system=generic platform=linux os=ubuntu25.04 target=m1 
[e]      ^glibc@2.41 build_system=autotools platform=linux os=ubuntu25.04 target=aarch64 
[+]      ^gmake@4.4.1~guile build_system=generic platform=linux os=ubuntu25.04 target=m1 %c=gcc@14.2.0
[+]      ^gnuconfig@2024-07-27 build_system=generic platform=linux os=ubuntu25.04 target=m1 
[+]      ^hdf@4.2.15+external-xdr~fortran~java~netcdf+pic~shared~szip build_system=autotools patches:=49733dd,5434f29,ce75518,d54592d,f7d67e8 platform=linux os=ubuntu25.04 target=m1 %c,cxx,fortran=gcc@14.2.0
[+]          ^bison@3.8.2~color build_system=autotools platform=linux os=ubuntu25.04 target=m1 %c,cxx=gcc@14.2.0
[...]

PR #2751, opened 2025-12-12
hdf-eos2: fix url and hdf4 linking, add fortran

danrosen25:
Fix hdf-eos2 package builds and add support for Fortran
https://github.com/spack/spack-packages/issues/2728

* download archive via web raw interface instead or rest api
* add danrosen25 to maintainers
* add variant for fortran support
* add conflict for gcc 14+
* add compiler flag to ignore implicit-function-declaration (newer compilers)
* add compiler flag to ignore implicit-int (newer compilers)
* use h4cc to set CC in order to fix hdf4 linking issues
* fix passing dependency library directories to configure args

@climbfuji - I think I've addressed everything that we discussed but please let me know if you find issues. Thank you!

### DIFF
--- a/repos/spack_repo/builtin/packages/hdf_eos2/package.py
+++ b/repos/spack_repo/builtin/packages/hdf_eos2/package.py
@@ -17,48 +17,55 @@ class HdfEos2(AutotoolsPackage):
     """
 
     homepage = "https://hdfeos.org"
-    # Starting with @3, download requires authentication.  So reverting
-    # to a manual download
-    url = "file://{0}/hdf-eos2-3.0-src.tar.gz".format(os.getcwd())
+    url = "https://git.earthdata.nasa.gov/projects/DAS/repos/hdfeos"
+    local_url = "file://{0}/hdf-eos2-{1}-src.tar.gz"
     manual_download = True
 
-    # The download URLs for @2 versions are messing, and include sha256 checksum.
-    # Templates for url_for_version. 0 is sha256 checksum, 1 is filename
-    # This is just a template.  See version_list and url_for_version below
-    v2url = "https://git.earthdata.nasa.gov/rest/git-lfs/storage/DAS/hdfeos/{0}?response-content-disposition=attachment%3B%20filename%3D%22{1}%22%3B%20filename*%3Dutf-8%27%27{1}"
+    # The old_url for archives is messy and is no longer used
+    old_url = "https://git.earthdata.nasa.gov/rest/git-lfs/storage/DAS/hdfeos/{0}?response-content-disposition=attachment%3B%20filename%3D%22{1}%22%3B%20filename*%3Dutf-8%27%27{1}"
 
-    maintainers("climbfuji")
+    maintainers("climbfuji", "danrosen25")
 
-    # Crazy URL scheme, differing with each version, and including the
-    # sha256 checksum in the URL.  Yuck
-    # The data in version_list is used to generate versions and urls
-    # In basename expansions, 0 is raw version,
-    # 1 is for version with dots => underscores
+    # Archives are committed to a branch related to the version
+    # but the branch name and archive name are subject to change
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/hdf_eos2/package.py
```diff
@@ -93,13 +103,14 @@ class HdfEos2(AutotoolsPackage):
     def url_for_version(self, version):
         vrec = [x for x in self.version_list if x["version"] == version.dotted.string]
         if vrec:
-            fname = vrec[0]["basename"].format(version.dotted, version.underscored)
-            sha256 = vrec[0]["sha256"]
+            branch = vrec[0]["branch"]
+            archive = vrec[0]["archive"]
             can_auto_download = vrec[0].get("can_auto_download", False)
```
climbfuji: If all the relevant versions above can be auto_downloaded, should we get rid of the logic that deals with manual downloads?

repos/spack_repo/builtin/packages/hdf_eos2/package.py
```diff
@@ -140,6 +156,19 @@ def setup_build_environment(self, env: EnvironmentModifications) -> None:
             extra_ldflags.append(tmp)
         # Set LDFLAGS
         env.set("LDFLAGS", " ".join(extra_ldflags))
+        # Use h4cc compiler wrapper and flags
+        if self.spec.satisfies("^hdf"):
+            tmp = self.spec["hdf"].prefix
+            env.set("CC", os.path.join(tmp.bin, "h4cc"))
```
climbfuji:
```suggestion
            env.set("CC", os.path.join(self.spec["hdf"].prefix.bin, "h4cc"))
```

repos/spack_repo/builtin/packages/hdf_eos2/package.py
```diff
@@ -17,66 +17,76 @@ class HdfEos2(AutotoolsPackage):
     """
 
     homepage = "https://hdfeos.org"
-    # Starting with @3, download requires authentication.  So reverting
-    # to a manual download
-    url = "file://{0}/hdf-eos2-3.0-src.tar.gz".format(os.getcwd())
+    url = "https://git.earthdata.nasa.gov/projects/DAS/repos/hdfeos"
+    local_url = "file://{0}/hdf-eos2-{1}-src.tar.gz"
```
climbfuji: See my comment below. Can we remove the local_url and old_url stuff, even if that means removing deprecated versions v2 altogether?

danrosen25: @climbfuji I can remove `can_auto_download`, `local_url`, and `old_url`. All versions `3.0`, `2.20v1.00`, `2.19v1.00`, `2.19b` have archive files that can be downloaded using the https raw file interface.  The only reason I deprecated `2.19b` is because it's binaries and I don't think this recipe is set up to install binaries. Plus `2.19b` fails when it tries to apply `hdf-eos2.configure.patch` because there's no `configure` file. I added `2.19v1.00` so that this version can still be installed, but from source.

repos/spack_repo/builtin/packages/hdf_eos2/package.py
danrosen25: Done

repos/spack_repo/builtin/packages/hdf_eos2/package.py
danrosen25: Removed can_auto_download

climbfuji (approved):
I think this looks great, I didn't test myself though.

We can probably do a little more cleanup (remove some unnecessary `tmp` steps) in the future, but for now this should be good enough.

Thanks for contributing these changes!

tldahlgren (approved):
I was able to manually download and verify the sha256s for each of the listed versions.

PR #2756, opened 2025-12-12
ci: bump Spack commit

haampie:
* Bump Spack commit to spack/spack@63a41706da7db43ccbabf8e1d0d5233456596907
   1. enables parallel concretization on macOS
   2. speeds up concretization by ~40%. E4S is now 16 minutes instead of 25+.
* Do not install unused `setuptools`, `pip`, `coverage` and `xdist` Python packages in GHA.
* Remove non-existent `lib/spack/spack/external` from `PYTHONPATH`


tldahlgren (dismissed):

haampie:
Seems to trigger hash changes, don't remember if that's intentional. Do you @alalazo?

No package hash change, just a hash change. Would have to bisect.

haampie:
Bisected to spack/spack@cba4a704a3b15a0c7f7f7e917aa4f662a106d573, but this was a bug fix so the hash change is expected.

haampie:
@spackbot rerun pipeline

haampie:
OK, so 3 issues had to be fixed, of which one still pending in spack/spack.

haampie:
This should be ready to go. Waiting for develop pipelines to finish (and then these).

alalazo (approved):

PR #2757, opened 2025-12-13
ripgrep: add pcre2 variant

ebagrenrut:
- Add `ripgrep` 15.1.0, which requires `rust` 2024 edition
- Add a `pcre2` variant. Per the [CHANGELOG](https://github.com/BurntSushi/ripgrep/blob/master/CHANGELOG.md#0100-2018-09-07), `ripgrep` has had optional support for PCRE2 since version 0.10.0. This is much older than any of the versions supported by this package, so the `pcre2` variant applies to any version supported by the package. PCRE2 support requires a provider of `pkgconfig` and the PCRE2 library. The user can now `+pcre2` to enable PCRE2 support during the `ripgrep` build.

@alecbcs

### DIFF
--- a/repos/spack_repo/builtin/packages/ripgrep/package.py
+++ b/repos/spack_repo/builtin/packages/ripgrep/package.py
@@ -33,15 +33,14 @@ class Ripgrep(CargoPackage):
     depends_on("rust@1.72:1.84", type="build", when="@14")
     depends_on("c", type="build")
 
-    build_args=["--release"]
-
     with when("+pcre2"):
-        build_args.append("--features 'pcre2'")
+        build_args=["--features", "pcre2"]
         depends_on("pcre2", type="build")
         depends_on("pkgconfig", type="build")
 
     @when("+pcre2")
     def setup_build_environment(self, env: EnvironmentModifications) -> None:
+        super().setup_build_environment
         env.set("PCRE2_SYS_STATIC", "1")
 
     @run_after("install")
--- a/repos/spack_repo/builtin/packages/ripgrep/package.py
+++ b/repos/spack_repo/builtin/packages/ripgrep/package.py
@@ -20,14 +20,29 @@ class Ripgrep(CargoPackage):
 
     license("MIT OR Unlicense")
 
+    version("15.1.0", sha256="046fa01a216793b8bd2750f9d68d4ad43986eb9c0d6122600f993906012972e8")
     version("14.1.1", sha256="4dad02a2f9c8c3c8d89434e47337aa654cb0e2aa50e806589132f186bf5c2b66")
     version("14.1.0", sha256="33c6169596a6bbfdc81415910008f26e0809422fda2d849562637996553b2ab6")
     version("14.0.3", sha256="f5794364ddfda1e0411ab6cad6dd63abe3a6b421d658d9fee017540ea4c31a0e")
     version("13.0.0", sha256="0fb17aaf285b3eee8ddab17b833af1e190d73de317ff9648751ab0660d763ed2")
     version("11.0.2", sha256="0983861279936ada8bc7a6d5d663d590ad34eb44a44c75c2d6ccd0ab33490055")
 
-    depends_on("rust@1.72:", type="build", when="@14:")
+    variant("pcre2", default=False, description="Add support for Perl-style regex via PCRE2")
+
+    depends_on("rust@1.85:", type="build", when="@15:")
+    depends_on("rust@1.72:1.84", type="build", when="@14")
     depends_on("c", type="build")
+
+    build_args=["--release"]
+
+    with when("+pcre2"):
+        build_args.append("--features 'pcre2'")
+        depends_on("pcre2", type="build")
+        depends_on("pkgconfig", type="build")
+
+    @when("+pcre2")
+    def setup_build_environment(self, env: EnvironmentModifications) -> None:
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/ripgrep/package.py
```diff
@@ -27,22 +27,37 @@ class Ripgrep(CargoPackage):
...
+            mkdirp(fish_completion_path(self.prefix))
+            with open(fish_completion_path(self.prefix) / "rg.fish", "w") as file:
+                rg("--generate", "complete-fish", output=file)
 
-        mkdirp(zsh_completion_path(self.prefix))
-        with open(zsh_completion_path(self.prefix) / "_rg", "w") as file:
-            rg("--generate", "complete-zsh", output=file)
+            mkdirp(zsh_completion_path(self.prefix))
+            with open(zsh_completion_path(self.prefix) / "_rg", "w") as file:
+                rg("--generate", "complete-zsh", output=file)
```
alecbcs: Instead of gating the entire method under

```python3
if self.spec.satisfies("@14:"):
```

We can instead put a when qualifier on the `@run_after()` decorator.
```patch
- @run_after("install")
+ @run_after("install", when="@14:")
  def install_completions(self):
  ...
```

repos/spack_repo/builtin/packages/ripgrep/package.py
```diff
@@ -27,22 +27,37 @@ class Ripgrep(CargoPackage):
     version("13.0.0", sha256="0fb17aaf285b3eee8ddab17b833af1e190d73de317ff9648751ab0660d763ed2")
     version("11.0.2", sha256="0983861279936ada8bc7a6d5d663d590ad34eb44a44c75c2d6ccd0ab33490055")
 
+    variant("pcre2", default=False, description="Add support for Perl-style regex via PCRE2")
+
     depends_on("rust@1.85:", type="build", when="@15:")
-    depends_on("rust@1.72:", type="build", when="@14:")
+    depends_on("rust@1.72:", type="build", when="@14")
+    depends_on("rust@1.31:", type="build", when="@:13")
```
alecbcs:
```suggestion
    depends_on("rust@1.72:", type="build", when="@14:")
    depends_on("rust@1.31:", type="build")
```

I would drop the `when="@:13"` from here as you're effectively creating a base case for all older versions. I'd then retain the `when="@14:"` (colon at the end) such that the newer major version's requirements layer on top of each other.

repos/spack_repo/builtin/packages/ripgrep/package.py
```diff
@@ -27,22 +27,37 @@ class Ripgrep(CargoPackage):
...
     depends_on("c", type="build")
 
+    with when("+pcre2"):
+        depends_on("pcre2")
+        depends_on("pkgconfig", type="build")
```
alecbcs:
```suggestion
    depends_on("pkgconfig", type="build", when="+pcre2")
    
    depends_on("pcre2", when="+pcre2")
```

Personal preference, but since there's only two dependencies related to `pcre2` support I'd just manually specify the when on both of them. (I'd also break up the sections so that the build dependencies are visually separate from the build/link dependencies.)

If additional features are added in the future that depend on `pkgconfig` it'll likely make sense to just make that build dependency required. However we'll keep it gated for now since there's only a single variant.

repos/spack_repo/builtin/packages/ripgrep/package.py
```diff
@@ -27,22 +27,37 @@ class Ripgrep(CargoPackage):
...
+
+    @property
+    def build_args(self):
+        if self.spec.satisfies("+pcre2"):
+            return ["--features", "pcre2"]
+
+        return []
```
alecbcs:
```suggestion
    @property
    def build_args(self):
        args = []
        if self.spec.satisfies("+pcre2"):
            args.extend(["--features", "pcre2"])

        return args
```

This suggestion is a bit less technically fun, but I'd argue to do this the boring way to create a more usable function in the future. If additional arguments are added later on it'll require less work by a future contributor to simply extend or append to the args.

alecbcs (changes requested):
@ebagrenrut thanks for the contribution! A couple quick suggestions, but then this is looking good to me. Thanks!

repos/spack_repo/builtin/packages/ripgrep/package.py
ebagrenrut: I didn't realize the `run_after` decorator supported that. Thanks!

ebagrenrut: Addressed with 9645980.

repos/spack_repo/builtin/packages/ripgrep/package.py
ebagrenrut: Addressed with 9645980.

repos/spack_repo/builtin/packages/ripgrep/package.py
ebagrenrut: Addressed with 9645980.

repos/spack_repo/builtin/packages/ripgrep/package.py
ebagrenrut: Addressed with 9645980.

alecbcs (approved):
Looks good to me. Thanks @ebagrenrut!

PR #2758, opened 2025-12-13
mathic: add v1.1

d-torrance:
The latest release uses a tarball, which we use instead of a git commit, that includes a configure script, so we don't need to depend on autoconf, etc.

### DIFF
--- a/repos/spack_repo/builtin/packages/mathic/package.py
+++ b/repos/spack_repo/builtin/packages/mathic/package.py
@@ -16,12 +16,14 @@ class Mathic(AutotoolsPackage):
     currently works best with dense term representations."""
 
     homepage = "https://github.com/Macaulay2/mathic"
+    url = "https://github.com/Macaulay2/mathic/releases/download/v1.1/mathic-1.1.tar.gz"
     git = "https://github.com/Macaulay2/mathic"
 
     maintainers("d-torrance")
 
     license("LGPL-2.0-or-later", checked_by="d-torrance")
 
+    version("1.1", sha256="2499fb3df3c2f8a201ae5627cad95538aaabee0eee235002b8737bdb842b694a")
     version("1.0.2025.05.13", commit="7abf77e4ce493b3830c7f8cc09722bbd6c03818e")
 
     depends_on("cxx", type="build")


repos/spack_repo/builtin/packages/mathic/package.py
```diff
@@ -17,18 +17,21 @@ class Mathic(AutotoolsPackage):
...
 
     depends_on("cxx", type="build")
 
-    depends_on("autoconf", type="build")
-    depends_on("automake", type="build")
-    depends_on("libtool", type="build")
+    # configure script distributed beginning w/ version 1.1
+    depends_on("autoconf", type="build", when="@:1.0.2025.0.13")
+    depends_on("automake", type="build", when="@:1.0.2025.0.13")
+    depends_on("libtool", type="build", when="@:1.0.2025.0.13")
```
alecbcs: We'll still need to leave in these dependencies if someone is to checkout a version of the project from the git url that's not specified in a version directive. (E.g. `spack install mathic@git.sha1_ref`)

It's a bit less optimized, but I'd recommend just leaving them in until we have a good way to specify they're only required for git versions and not those coming from release tarballs.

d-torrance: Makes sense -- fixed!

repos/spack_repo/builtin/packages/mathic/package.py
```diff
@@ -17,11 +17,13 @@ class Mathic(AutotoolsPackage):
 
     homepage = "https://github.com/Macaulay2/mathic"
     git = "https://github.com/Macaulay2/mathic"
+    url = "https://github.com/Macaulay2/mathic/releases/download/v1.1/mathic-1.1.tar.gz"
```
alecbcs, line 19:
```suggestion
homepage = "https://github.com/Macaulay2/mathic"
url = "https://github.com/Macaulay2/mathic/releases/download/v1.1/mathic-1.1.tar.gz"
git = "https://github.com/Macaulay2/mathic"
```

Let's swap `url` and `git` for consistency across packages.

alecbcs (commented):
One quick suggestion. Otherwise looks good to me. Thanks!

repos/spack_repo/builtin/packages/mathic/package.py
d-torrance: Done!

alecbcs (approved):
Looks good to me. Thanks @d-torrance!

PR #2763, opened 2025-12-14
graphviz: add package autodetect

sethrj:


repos/spack_repo/builtin/packages/graphviz/package.py
```diff
@@ -149,6 +152,22 @@ class Graphviz(AutotoolsPackage):
         msg="graphviz-2.40.1 needs gcc-6 or greater to compile with QT5 suppport",
     )
 
+    tags = ["build-tools"]
```
johnwparent, line 155: n.b. extremely pedantic nit:  is graphviz a "build-tool"?

repos/spack_repo/builtin/packages/graphviz/package.py
```diff
@@ -149,6 +152,22 @@ class Graphviz(AutotoolsPackage):
...
+
+    executables = ["^dot$"]
+
+    @classmethod
+    def determine_version(cls, exe):
+        print(exe)
```
johnwparent: Probably don't want to leave a print here

repos/spack_repo/builtin/packages/graphviz/package.py
```diff
@@ -149,6 +152,22 @@ class Graphviz(AutotoolsPackage):
...
+        output = Executable(exe)("--version", output=str, error=str)
+        match = re.search(r"graphviz\s+version\s+([\d\.]+)", output)
+        return match.group(1) if match else None
+
+    @classmethod
+    def determine_variants(cls, exes, version_str):
```
johnwparent: If this is intended to be a no-op then I think we can just do away with it

repos/spack_repo/builtin/packages/graphviz/package.py
sethrj: I would argue so; it's never linked into anything, and it's frequently used to build documentation. The main argument is that it can be found and used by default with `spack external find` without consequences (not true for libraries and such that may link to system libs).

repos/spack_repo/builtin/packages/graphviz/package.py
sethrj: Ah I didn't realize it was fully optional. Done!

repos/spack_repo/builtin/packages/graphviz/package.py
johnwparent: Fair enough

johnwparent (approved):

sethrj:
Thanks @johnwparent! 🙏

becker33 (approved):

PR #2767, opened 2025-12-15
ginkgo: add v1.11.0

MarcelKoch:
This PR adds Ginkgo v1.11.0.


MarcelKoch:
I would need some help regarding the CI failure. I've not been able to reproduce this issue locally. If I install cuda 13.0.2 locally via spack, the CMake `find_package(Thrust|CCCL)` works without issue.
Note that we need this `find_package`, since Ginkgo can't be compiled otherwise if cuda is available as a module.

tldahlgren (approved):
Confirmed the version commit.

PR #2768, opened 2025-12-15
root: avoid non-determinism of cxxstd

haampie:
The default is `cxxstd=11`, but it conflicts with all recent versions of root:

```python
    conflicts("cxxstd=11", when="@6.25.02:", msg="This version of root requires at least C++14")
    conflicts("cxxstd=14", when="@6.30.00:", msg="This version of root requires at least C++17")
```

That leads to non-determinism of the default value thanks to
https://github.com/spack/spack/issues/51112.

So, for now pick a default value of C++17 to get something that's
deterministic when picking recent versions of root.

### DIFF
--- a/repos/spack_repo/builtin/packages/root/package.py
+++ b/repos/spack_repo/builtin/packages/root/package.py
@@ -353,7 +353,7 @@ class Root(CMakePackage):
 
     variant(
         "cxxstd",
-        default="11",
+        default="17",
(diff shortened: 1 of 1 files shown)


alalazo (approved):
Purpose is similar to #2702 so I completely agree. If maintainers have follow up requests, they can ping any of us for a review on a subsequent PR.

PR #2774, opened 2025-12-15
ERF - add support for Noah-MP Land Surface Model

larenspear:
Adds support (via submodule) for NCAR's Noah-MP Land Surface Model (https://github.com/NCAR/noahmp).


repos/spack_repo/builtin/packages/erf/package.py
```diff
@@ -4,6 +4,15 @@
 from spack.package import *
 
 
+def submodules(package):
+    submodules = []
+
+    with when("+noahmp"):
+        submodules.append("Submodules/Noah-MP")
```
becker33: `when` blocks don't work this way for imperative code, only for the declarative directives (things that take a `when` argument).

```suggestion
    if package.spec.satisfies("+noahmp"):
        submodules.append("Submodules/Noah-MP")
```

becker33 (changes requested):
I'd prefer that this use a separate package for noahmp if possible, but since no one has cared about a standalone package for noahmp yet I'm not inclined to insist on it.

repos/spack_repo/builtin/packages/erf/package.py
larenspear: I'm working on a package for NoahMP as well; however, development on that is going to be delayed because of a lack of funding. So once I've finished that I intend to swap it out for the Spack package.

becker33 (approved):

PR #2775, opened 2025-12-15
bbmap: add 39.59; fixes/improvements

ebagrenrut:
- Add BBMap 39.59
- Change home page to https://bbmap.org/tools/bbmap, the current homepage for BBMap
- `Bbmap` class now extends `MakefilePackage`
- Add `usejni` variant that, when true, will cause the package to build libbbtoolsjni.so for acceleration.
- Constrain the compiler dependency to `gcc`. Per the [BBMap docs](https://github.com/bbushnell/BBTools/blob/master/jni/README.txt), and the contents of the makefile(s), gcc should be used.
- BBMap requires Java 8 or later, and it is also a run-time dependency. The `java` dependency has been updated accordingly.
- The BBMap source comes with makefiles for Linux and macOS (OS/X). The `edit` function is overridden to rename, based on the spec's `platform`, the appropriate file to `Makefile`.
- The `build` function has been overridden to first run `make clean`, as the BBMap distribution ships with a pre-compiled library that is only for x86_64.

### DIFF
--- a/repos/spack_repo/builtin/packages/bbmap/package.py
+++ b/repos/spack_repo/builtin/packages/bbmap/package.py
@@ -32,7 +32,8 @@ class Bbmap(MakefilePackage, SourceforgePackage):
     )
 
     # Building BBMap's jni libraries requires gcc, per the BBMap docs
-    depends_on("gcc", type="build", when="+usejni")
+    depends_on("c", type="build", when="+usejni")
+    requires("%c=gcc", when="+usejni")
     depends_on("java@8:", type=("build", "link", "run"))
 
     def edit(self, spec, prefix):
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/bbmap/package.py
```diff
@@ -2,28 +2,61 @@
...
+        default=False,
+        description=(
+            "Compile the libbbtoolsjni library for accelerated versions of BBMap, Dedupe, "
+            "BBMerge, and IceCreamFinder"
+        ),
+    )
 
-    depends_on("java")
+    # Building BBMap's jni libraries requires gcc, per the BBMap docs
+    depends_on("gcc", type="build", when="+usejni")
```
becker33: I think this should be
```suggestion
    depends_on("c", type="build", when="+usejni")
    requires("%c=gcc", when="+usejni")
```
I think you need to ensure the `c` virtual is represented in the DAG.

becker33 (changes requested):
I haven't looked at the makefile, but we often see makefiles that hardcode GCC out of a lack of imagination (never considered the idea of another compiler) instead of because it fails with other compilers.

If it's worth pursuing, it's often as simple as having the `edit` phase use `filter_file` to replace the old settings with whatever compiler is being used.

becker33 (approved):

PR #2780, opened 2025-12-16
libvdwxc: add  v0.5.0

mtaillefumier:
Version update

### DIFF
--- a/repos/spack_repo/builtin/packages/libvdwxc/package.py
+++ b/repos/spack_repo/builtin/packages/libvdwxc/package.py
@@ -33,9 +33,9 @@ class Libvdwxc(AutotoolsPackage):
     depends_on("mpi@2:", when="+mpi")
     depends_on("pfft", when="+pfft")
 
-    depends_on("autoconf", type="build")
-    depends_on("automake", type="build")
-    depends_on("libtool", type="build")
+    depends_on("autoconf", type="build", when="@master")
+    depends_on("automake", type="build", when="@master")
+    depends_on("libtool", type="build", when="@master")
 
     # pfft needs MPI
     conflicts("~mpi", "+pfft")
@@ -71,7 +71,3 @@ class Libvdwxc(AutotoolsPackage):
     # The relevant upstream fix for the m4 would be:
     # https://gitlab.com/libvdwxc/libvdwxc/-/commit/9340f857515c4a2e56d2aa7cf3a21c41ba8559c3.diff
     patch("fftw-detection.patch", when="@:0.4.0")
-
-    def url_for_version(self, version):
-        url = "https://launchpad.net/libvdwxc/stable/{0}/+download/libvdwxc-{0}.tar.gz"
-        return url.format(version)
--- a/repos/spack_repo/builtin/packages/libvdwxc/package.py
+++ b/repos/spack_repo/builtin/packages/libvdwxc/package.py
@@ -12,6 +12,7 @@ class Libvdwxc(AutotoolsPackage):
     """Portable C library of density functionals with van der Waals
     interactions for density functional theory"""
 
+    maintainers = "mtaillefumier"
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/libvdwxc/package.py
```diff
@@ -13,12 +13,13 @@ class Libvdwxc(AutotoolsPackage):
     interactions for density functional theory"""
 
     homepage = "https://libvdwxc.gitlab.io/libvdwxc/"
-    url = "https://launchpad.net/libvdwxc/stable/0.4.0/+download/libvdwxc-0.4.0.tar.gz"
+    url = "https://launchpad.net/libvdwxc/stable/0.5.0/+download/libvdwxc-0.5.0.tar.gz"
```
RMeli, line 17: Doesn't hurt, but FYI this does not need to be changed.

repos/spack_repo/builtin/packages/libvdwxc/package.py
```diff
@@ -65,3 +70,7 @@ def configure_args(self):
     # The relevant upstream fix for the m4 would be:
     # https://gitlab.com/libvdwxc/libvdwxc/-/commit/9340f857515c4a2e56d2aa7cf3a21c41ba8559c3.diff
     patch("fftw-detection.patch", when="@:0.4.0")
+
+    def url_for_version(self, version):
+        url = "https://launchpad.net/libvdwxc/stable/{0}/+download/libvdwxc-{0}.tar.gz"
+        return url.format(version)
```
RMeli: Why was this necessary? It should be done automatically by Spack. `url_for_version` is usually needed if different versions need different URLs etc, but it does not seem to be the case here.

repos/spack_repo/builtin/packages/libvdwxc/package.py
```diff
@@ -31,6 +32,10 @@ class Libvdwxc(AutotoolsPackage):
     depends_on("mpi@2:", when="+mpi")
     depends_on("pfft", when="+pfft")
 
+    depends_on("autoconf", type="build")
+    depends_on("automake", type="build")
+    depends_on("libtool", type="build")
```
RMeli: Are these needed only for/from version `0.5.0`? The `configure` script is not packaged? If yes, we should add the `when=` clause.

Maybe you had to add them when building from `master`? If this is the case, then it should probably be as follows?

```suggestion
    depends_on("autoconf", type="build", when="@master")
    depends_on("automake", type="build", when="@master")
    depends_on("libtool", type="build", when="@master")
```
(See https://spack.readthedocs.io/en/latest/build_systems/autotoolspackage.html#build-system-dependencies)

RMeli (changes requested):

repos/spack_repo/builtin/packages/libvdwxc/package.py
mtaillefumier: `spack checksum libvdwxc` does not work properly. It can not find new versions of libvdwxc

repos/spack_repo/builtin/packages/libvdwxc/package.py
mtaillefumier: I will add it for completeness.

repos/spack_repo/builtin/packages/libvdwxc/package.py
mtaillefumier: it does not work with the url_for_version either. I am puzzled by this

RMeli: Does `spack checksum libvdwxc 0.5.0` work?

mtaillefumier: Yes it does. I removed the 'url_for_version' method.

RMeli (approved):
LGTM, thanks @mtaillefumier!

PR #2788, opened 2025-12-16
Tau package patch for +dyninst +cuda +python

jordialcaraz:
When building TAU with

spack install tau@2.35 +dyninst +cuda +python

This configuration uses boost, cuda and python,  which enables  part of PYCUDA that are no longer used and prevent TAU from building.
This patch disables the PYCUDA part inside the configure, preventing the related Makefile lines to be enabled.


jordialcaraz:
May need to do a similar patch for previous versions. Is it possible to patch multiple versions with the same file or should a patch be done for each version?

eugeneswalker:
> May need to do a similar patch for previous versions. Is it possible to patch multiple versions with the same file or should a patch be done for each version?

As long as the patch applies correctly, you can update the `when` clause to apply the same patch to multiple versions, I believe.

eugeneswalker:
Do you want to try updating this to apply patch to multiple versions, or should we merge this as-is? I will approve and mege if the latter. Thanks!

jordialcaraz:
Let me check until which version this may apply. I will get back to you later.

This issue is fixed @master.

jordialcaraz:
Updated to apply the patch to apply since 2.33, where we have dyninst, which is the part that generates the conflict.
2.35 should be the last version where it is applied. 2.35.1 will not have this issue.

eugeneswalker (approved):

eugeneswalker:
Should we merge now? Do you want any additional changes before we merge? What do you think @wspear ?

jordialcaraz:
I tested the versions involved, there should be no issues.
Maybe we can also include the TAU version update? Sameer should have it ready soon.

PR #2789, opened 2025-12-16
awscli-v2: relax dependencies to support Python 3.14

adamjstewart:
Builds fine, but would appreciate a bit of testing to make sure that most functionality works in Python 3.14.


teaguesterling:
Hey @adamjstewart I was able to build this with Python 3.14 and confirm some basic functionality (s3, ec2, config, rds, etc.). All looks good.

climbfuji (approved):
I am not using Python 3.14 yet, but anything that relaxes constraints and still works is fine with me. Seems to be ok with an earlier version of Python (3.11).

alalazo (approved):

PR #2792, opened 2025-12-16
py-flit-core: mark Python 3.14 support

adamjstewart:
Discovered when the concretizer chose an older version and it didn't build.

### DIFF
--- a/repos/spack_repo/builtin/packages/py_flit_core/package.py
+++ b/repos/spack_repo/builtin/packages/py_flit_core/package.py
@@ -42,7 +42,7 @@ class PyFlitCore(PythonPackage):
         depends_on("python@3.4:", when="@3.0:")
         depends_on("python@2.7,3.4:", when="@2.3:")
         # https://github.com/pypa/flit/pull/684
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/py_flit_core/package.py
```diff
@@ -41,6 +41,8 @@ class PyFlitCore(PythonPackage):
         depends_on("python@3.6:", when="@3.4:")
         depends_on("python@3.4:", when="@3.0:")
         depends_on("python@2.7,3.4:", when="@2.3:")
+        # https://github.com/pypa/flit/pull/684
+        depends_on("python@:3.13", when="@:3.10")
```
takluyver: I think these range ends are inclusive? If so, I think it should be:


```suggestion
        depends_on("python@:3.13", when="@:3.9")
```

i.e. flit_core v 3.10 should be compatible with Python 3.14.

takluyver (approved):

alalazo (approved):

PR #2798, opened 2025-12-16
CI: Remove redundant sections from pcluster setup

kwryankrattiger:
Hotfix to remove sections of the pcluster build job that should not be necessary/were incorrect post include scope refactor.

### DIFF
--- a/.ci/gitlab/configs/ci.yaml
+++ b/.ci/gitlab/configs/ci.yaml
@@ -12,6 +12,7 @@ ci:
         - cat /proc/meminfo | grep 'MemTotal\|MemFree' || true
       - - touch ${SPACK_USER_CACHE_PATH}/cache/*/*  # bump mtime of cache so it is not invalidated
       - - spack env activate --without-view ${SPACK_CONCRETE_ENV_DIR}
+        - spack repo list
         - spack compiler list
       - - mkdir -p ${SPACK_ARTIFACTS_ROOT}/user_data
         # AWS runners mount E4S public key (verification), UO runners mount public/private (signing/verification)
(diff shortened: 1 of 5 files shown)


.ci/gitlab/scripts/common/clone_spack.sh
haampie, line 1: Add shebang `#!/bin/sh -e`. Can you avoid `bash` if you can write `dash`? See https://www.shellcheck.net/ with the suggested shebang if you're not sure what the difference is.

.ci/gitlab/configs/ci.yaml
```diff
@@ -146,13 +146,7 @@ ci:
...
         - nproc || true
       # Clone the other repo
-      - - mkdir -p .ci/tmp/spack
-        - cd .ci/tmp/spack
-        - git init
-        - git remote add origin https://github.com/${SPACK_CHECKOUT_REPO}.git
-        - git fetch --depth 1 origin ${SPACK_CHECKOUT_VERSION}
-        - git checkout FETCH_HEAD
-        - cd ${CI_PROJECT_DIR}
+      - bash ${SPACK_CI_PACKAGES_ROOT}/.ci/gitlab/scripts/common/clone_spack.sh
```
haampie: Would avoid specifying the shell here.

kwryankrattiger: It shouldn't really matter. I made it executable and removed the shell.

.ci/gitlab/scripts/common/clone_spack.sh
kwryankrattiger: It shouldn't matter, we already implicitly require bash already I believe, but `sh` is fine too. Converted and updated.

kwryankrattiger:
@spackbot run pipeline

haampie (dismissed):

kwryankrattiger:
@spackbot run pipeline

haampie (approved):

PR #2799, opened 2025-12-16
clhep: use the blessed Gitlab instance instead of proj-clhep.web.cern.ch

jmcarcell:
From one of the maintainers, this is the blessed location for sources so I've updated the sha256 hashes (maybe due to a different compression level). In particular version 2.4.7.2 is not available in the old location.


wdconinc:
I think this is hitting a rate limiting trigger on the cern gitlab while verifying all the checksums. When you restart the checksum job, you get different versions that fail. Someone with more patience than me may even get it to pass ;-)

jmcarcell:
I think `url_for_version` was needed. I have built clhep 2.4.7.2 successfully now.

wdconinc (approved):

PR #2800, opened 2025-12-16
Update MUSICA Package Version

boulderdaze:
Update MUSICA package version to 0.14.0


tldahlgren (approved):
Confirmed the sha256.

PR #2801, opened 2025-12-16
demuxlet: Modify LDFLAGS based on htslib dependencies

ebagrenrut:
- Remove `-lcurl` from LDFLAGS when depending on `htslib` sans `libcurl`
- Remove `-lcrypto` from LDFLAGS when depending on `htslib` sans S3 URL support

I found when concretizing a unified environment that contained `demuxlet` and `htslib~libcurl` that `demuxlet` would fail to link because it had a `-lcrypto` in its `LDFLAGS` while the `openssl` package was not installed. The proposed changes ensure that does not happen, and also ensure that `-lcurl` is not in `LDFLAGS` when `htslib~libcurl` is in the graph.

### DIFF
--- a/repos/spack_repo/builtin/packages/demuxlet/package.py
+++ b/repos/spack_repo/builtin/packages/demuxlet/package.py
@@ -34,15 +34,18 @@ class Demuxlet(AutotoolsPackage):
         filter_file("-I ../htslib/", "", "Makefile.am", string=True)
         filter_file("../htslib/libhts.a", "-lhts", "Makefile.am", string=True)
 
-        # Add -ldeflate to LDFLAGS When depending on htslib with libdeflate
+        # Add -ldeflate to LDFLAGS when depending on htslib with libdeflate
         if self.spec.satisfies("^htslib+libdeflate"):
             filter_file("-lcrypto", "-lcrypto -ldeflate", "Makefile.am", string=True)
 
         # Remove -lcurl from LDFLAGS when depending on htslib sans libcurl
-        if not self.spec.satisfies("^htslib+libcurl"):
+        if self.spec.satisfies("^htslib~libcurl"):
             filter_file("-lcurl", "", "Makefile.am", string=True)
 
-        # Remove -lcrypto from LDFLAGS when depending on htslib sans S3 URL support
+        # Remove -lcrypto from LDFLAGS when depending on htslib sans S3 URL support.
+        # htslib's s3 variant is conditional and ~s3 will not be in its spec when
+        # ~libcurl is in its spec, so we must check for the case in which +s3 is not
+        # explicit.
         if not self.spec.satisfies("^htslib+s3"):
             filter_file("-lcrypto", "", "Makefile.am", string=True)
 
--- a/repos/spack_repo/builtin/packages/demuxlet/package.py
+++ b/repos/spack_repo/builtin/packages/demuxlet/package.py
@@ -33,8 +33,18 @@ class Demuxlet(AutotoolsPackage):
         filter_file("-I ../../htslib/htslib", "", "Makefile.am", string=True)
         filter_file("-I ../htslib/", "", "Makefile.am", string=True)
         filter_file("../htslib/libhts.a", "-lhts", "Makefile.am", string=True)
+
+        # Add -ldeflate to LDFLAGS When depending on htslib with libdeflate
         if self.spec.satisfies("^htslib+libdeflate"):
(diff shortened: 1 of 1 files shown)


ebagrenrut:
@snehring

repos/spack_repo/builtin/packages/demuxlet/package.py
```diff
@@ -33,8 +33,18 @@ def patch(self):
         filter_file("-I ../../htslib/htslib", "", "Makefile.am", string=True)
         filter_file("-I ../htslib/", "", "Makefile.am", string=True)
         filter_file("../htslib/libhts.a", "-lhts", "Makefile.am", string=True)
+
+        # Add -ldeflate to LDFLAGS When depending on htslib with libdeflate
         if self.spec.satisfies("^htslib+libdeflate"):
             filter_file("-lcrypto", "-lcrypto -ldeflate", "Makefile.am", string=True)
 
+        # Remove -lcurl from LDFLAGS when depending on htslib sans libcurl
+        if not self.spec.satisfies("^htslib+libcurl"):
```
alalazo:
```suggestion
        if self.spec.satisfies("^htslib~libcurl"):
```

repos/spack_repo/builtin/packages/demuxlet/package.py
```diff
@@ -33,8 +33,18 @@ def patch(self):
...
+            filter_file("-lcurl", "", "Makefile.am", string=True)
+
+        # Remove -lcrypto from LDFLAGS when depending on htslib sans S3 URL support
+        if not self.spec.satisfies("^htslib+s3"):
```
alalazo, line 49:
```suggestion
        if self.spec.satisfies("^htslib~s3"):
```

alalazo (commented):
LGTM, but waiting for maintainers review. Ping me in a few days if you don't get any further comment.

repos/spack_repo/builtin/packages/demuxlet/package.py
ebagrenrut: I actually tried this at first, but I couldn't get it to match when I did not specify the s3 variant one way or another. I wonder if it might be because the s3 variant is only available if the libcurl variant is true? I.e. when I installed `demuxlet^htslib~libcurl`, `self.spec.satisfies("^htslib~s3")` would be `False`. I wonder now, though, if I was still getting an `htslib` that was `+libcurl`. I will do some testing today to see if I can get to the bottom of the issue.

alalazo: Ok, makes sense if the variant is defined conditionally. Thanks for clarifying.

repos/spack_repo/builtin/packages/demuxlet/package.py
ebagrenrut: I've simplified this logic after doing more testing with `htslib~libcurl` in the graph.

alalazo (approved):

repos/spack_repo/builtin/packages/demuxlet/package.py
ebagrenrut: Fixed/addressed.

repos/spack_repo/builtin/packages/demuxlet/package.py
snehring: So we'd want to remove the crypto linker flag when `htslib~libcurl` or `htslib+libcurl~s3`? I'm not sure which representation is more 'readable' that or the current one.

snehring (approved):
lgtm, tested with `htslib~libcurl` and it builds.

repos/spack_repo/builtin/packages/demuxlet/package.py
ebagrenrut: We only need `-lcrypto` if `htslib` is installed with `+s3`. Because the `htslib` `s3` variant is conditional upon `htslib+libcurl`, it does not appear in the graph at all, not even as `~s3`, when `htslib` is `~libcurl`, so we can't depend on `self.spec.satisfies("^htslib~s3")` as a check unless the user were to specifically install `htslib+libcurl~s3`.

snehring: right. I'm implying `htslib~s3` is logically equivalent to `self.spec.satisfies(^htslib~libcurl) or self.spec.satisfies(^htslib+libcurl~s3)` in this case. I'm just not sure if that's better from a readability standpoint vs what you've got currently.

PR #2802, opened 2025-12-16
Add the ATfL package

pawosm-arm:
Also adjust the armpl-gcc package to accommodate our new toolchain.

### DIFF
--- a/repos/spack_repo/builtin/packages/atfl/package.py
+++ b/repos/spack_repo/builtin/packages/atfl/package.py
@@ -1,0 +1,217 @@
+# Copyright Spack Project Developers. See COPYRIGHT file.
+# SPDX-License-Identifier: (Apache-2.0 OR MIT)
+
+from pathlib import Path
+from typing import ClassVar, Dict
+
+from spack_repo.builtin.build_systems.compiler import CompilerPackage
+from spack_repo.builtin.build_systems.generic import Package
+from spack_repo.builtin.packages.llvm.package import LlvmDetection
+
+from spack.package import *
+
+_VERSIONS = {
+    "21.1.1": {
+        "ubuntu22.04": (
+            "8132ef95e4671c20a5f2b21dbe2d7ad8ae16137ea634e3e11096a8b87a3ffeee",
+            "https://developer.arm.com/packages/arm-toolchains%3Aubuntu-22/jammy/arm64/arm-toolchain-for-linux_21.1-81_arm64.deb",
+        ),
+        "ubuntu24.04": (
+            "18f210fb04f27c50af932e1549deeda394a4ff351f9388957de0b096ab5f1db0",
+            "https://developer.arm.com/packages/arm-toolchains%3Aubuntu-24/noble/arm64/arm-toolchain-for-linux_21.1-81_arm64.deb",
+        ),
+        "rhel8": (
+            "c5f0c3f7a25e269160aa3d684cb14edc57252bc38643929c67cda892f1c28dda",
+            "https://developer.arm.com/packages/arm-toolchains%3Arhel-8/el8/aarch64/arm-toolchain-for-linux-21.1-81.aarch64.rpm",
+        ),
+        "rhel9": (
+            "3ead2888f8aa71b79a944270b1f7f723f56a7d161aeb2abaad14b3333f91f0e9",
+            "https://developer.arm.com/packages/arm-toolchains%3Arhel-9/el9/aarch64/arm-toolchain-for-linux-21.1-81.aarch64.rpm",
+        ),
+        "rhel10": (
+            "a2fc22f18b7ee764ca11af48e6215b43ab47597cc67ba63fcccf0e40f9859a76",
+            "https://developer.arm.com/packages/arm-toolchains%3Arhel-10/el10/aarch64/arm-toolchain-for-linux-21.1-81.aarch64.rpm",
+        ),
+        "amzn2023": (
+            "f7bc8c156aaa367dff1d6ced922e90581a04c8bac92e5675cc107eb3cd28428a",
+            "https://developer.arm.com/packages/arm-toolchains%3Aamzn-2023/al2023/aarch64/arm-toolchain-for-linux-21.1-81.aarch64.rpm",
+        ),
+        "sles15": (
+            "8de2e35511d7f59a933c40e2e20ed2e01d018f51caba36dc5e9e0c6da71a3619",
+            "https://developer.arm.com/packages/arm-toolchains%3Asles-15/sl15/aarch64/arm-toolchain-for-linux-21.1-81.aarch64.rpm",
+        ),
+    },
+    "20.1.0": {
+        "ubuntu22.04": (
+            "944cf6420fb7b49c52d6c3d6f139fbb4896073ad401204a0cdf609faea360e73",
+            "https://developer.arm.com/packages/arm-toolchains%3Aubuntu-22/jammy/arm64/arm-toolchain-for-linux_20.1-65_arm64.deb",
+        ),
+        "ubuntu24.04": (
+            "b9f7db08da8d579daad06999a65136e6b449b79c870a16264fbef1a7ab0cbe6b",
+            "https://developer.arm.com/packages/arm-toolchains%3Aubuntu-24/noble/arm64/arm-toolchain-for-linux_20.1-65_arm64.deb",
+        ),
+        "rhel8": (
+            "795e6b74b2b538cc8431ff164c145aa7920ad9394ccf1d544a06667663b70f01",
+            "https://developer.arm.com/packages/arm-toolchains%3Arhel-8/el8/aarch64/arm-toolchain-for-linux-20.1-65.aarch64.rpm",
+        ),
+        "rhel9": (
+            "453fc0f4d62968a833499f40a3d04deddbf292c9ff40fd3d246e84ba5b9d7d7a",
+            "https://developer.arm.com/packages/arm-toolchains%3Arhel-9/el9/aarch64/arm-toolchain-for-linux-20.1-65.aarch64.rpm",
+        ),
+        "amzn2023": (
+            "87b2031290b9a48b0f2c700aff5f38560f7d1cb7d7f0b0f537ce18965260a54a",
+            "https://developer.arm.com/packages/arm-toolchains%3Aamzn-2023/al2023/aarch64/arm-toolchain-for-linux-20.1-65.aarch64.rpm",
+        ),
+        "sles15": (
+            "506a9ddff6d7daf60e1d5b8cfed95f6f25e70db3cc9545d8ea3e9745056740f9",
+            "https://developer.arm.com/packages/arm-toolchains%3Asles-15/sl15/aarch64/arm-toolchain-for-linux-20.1-65.aarch64.rpm",
+        ),
+    },
+}
+
+
+class Atfl(Package, LlvmDetection, CompilerPackage):
+    """Arm Toolchain for Linux (ATfL): LLVM-based AArch64 compilers.
+
+    Installs the distro package (RPM/DEB) from Arm's repos, without root.
+    """
+
+    maintainers("pawosm-arm")
+    homepage = "https://developer.arm.com/documentation/110477"
+
+    _alias: ClassVar[Dict[str, str]] = {"rocky10": "rhel10", "rocky9": "rhel9", "rocky8": "rhel8"}
+    for ver, per_os in _VERSIONS.items():
+        pkg = per_os.get(_alias.get(host_platform().default_os, host_platform().default_os), None)
+        if pkg:
+            version(ver, sha256=pkg[0], url=pkg[1], expand=False)
+
+    # Linux AArch64 only
+    requires("platform=linux", msg="ATfL is only available on Linux")
+    requires("target=aarch64:", msg="ATfL provides AArch64-native compilers")
+
+    provides("c", "cxx")
+    provides("fortran")
+
+    # Optional: allow opt-in to lld via the spec. Default is GNU ld for
+    # maximum compatibility; packages that prefer lld can request `%atfl+lld`.
+    variant("lld", default=False, description="Use lld linker (adds -fuse-ld=lld)")
+
+    # Arm Performance Libraries are required: ATfL's default configuration
+    # links against ArmPL (e.g. -fveclib=ArmPL -> -lamath). Ensure it's
+    # present and wired into the environment so linking always succeeds.
+    depends_on("armpl-gcc~examples", type="run")
+
+    compiler_languages = ["c", "cxx", "fortran"]
+    compiler_version_regex = r"Arm Toolchain for Linux [\d\.]+ [cf]lang version ([\d\.]+)"
+    c_names = ["armclang"]
+    cxx_names = ["armclang++"]
+    fortran_names = ["armflang"]
+    compiler_wrapper_link_paths = {
+        "c": join_path("arm", "armclang"),
+        "cxx": join_path("arm", "armclang++"),
+        "fortran": join_path("arm", "armflang"),
+    }
+    implicit_rpath_libs = ["libclang"]
+    stdcxx_libs = ("-lstdc++",)
+
+    debug_flags = [
+        "-gcodeview",
+        "-gdwarf-2",
+        "-gdwarf-3",
+        "-gdwarf-4",
+        "-gdwarf-5",
+        "-gline-tables-only",
+        "-gmodules",
+        "-g",
+    ]
+
+    opt_flags = ["-O0", "-O1", "-O2", "-O3", "-Ofast", "-Os", "-Oz", "-Og", "-O", "-O4"]
+
+    def archspec_name(self) -> str:
+        """Return the compiler name to use for archspec queries."""
+        return "clang"
+
+    def install(self, spec: Spec, prefix: Prefix) -> None:
+        """Install the package."""
+        archive = self.stage.archive_file
+        extract_dir = join_path(self.stage.path, "extract")
+        mkdir(extract_dir)
+        if archive.endswith(".rpm"):
+            Executable("bsdtar")("-xf", archive, "-C", str(extract_dir))
+        elif archive.endswith(".deb"):
+            Executable("dpkg-deb")("-x", archive, str(extract_dir))
+        else:
+            raise InstallError(f"Unknown archive type: {archive}")
+        src_root = join_path(extract_dir, "opt", "arm", "arm-toolchain-for-linux")
+
+        if not Path(src_root).is_dir():
+            raise InstallError(f"Expected payload at {src_root}")
+
+        # We must remove all of the libamath.* symlinks as they reach /opt
+        for cand in Path(src_root).rglob("libamath.*"):
+            if cand.is_symlink():
+                cand.unlink()
+
+        install_tree(str(src_root), prefix)
+
+        with open(join_path(prefix, "bin", "atfl-performance.cfg"), "a") as f:
+            # Enable lld if requested. Some system builds (e.g. ncurses with
+            # version scripts) are stricter under lld; make this opt-in to
+            # maximize compatibility.
+            if self.spec.satisfies("+lld"):
+                print("-fuse-ld=lld", file=f)
+
+            # ArmPL's libamath must be visible to the compiler.
+            # ArmPL directory layout: <prefix>/armpl_<version>_gcc/lib
+            armpl_dir = join_path(
+                self["armpl-gcc"].prefix, "armpl_" + str(self["armpl-gcc"].version) + "_gcc", "lib"
+            )
+            print(f"-L{armpl_dir} -Wl,-rpath={armpl_dir}", file=f)
+
+            # Ensure ATfL's own Fortran runtime is available to dependents.
+            # At first, Flang was using libFortranRuntime and libFortranDecimal;
+            # later toolchains use libflang_rt.runtime. This avoids per-package
+            # tweaks.
+            fortran_runtime = ""
+            fortran_runtime_paths = set()
+            for cand in Path(prefix).rglob("libFortranRuntime.*"):
+                fortran_runtime = "-lFortranRuntime -lFortranDecimal"
+                fortran_runtime_paths.add(str(cand.parent))
+            for cand in Path(prefix).rglob("libflang_rt.runtime.*"):
+                fortran_runtime = "-lflang_rt.runtime"
+                fortran_runtime_paths.add(str(cand.parent))
+            for rt_path in fortran_runtime_paths:
+                print(f"-L{rt_path} -Wl,-rpath,{rt_path}", file=f)
+            print(f"-Wl,--push-state -Wl,--as-needed {fortran_runtime} -Wl,--pop-state", file=f)
+
+    def _cc_path(self) -> str:
+        return join_path(self.prefix, "bin", "armclang")
+
+    def _cxx_path(self) -> str:
+        return join_path(self.prefix, "bin", "armclang++")
+
+    def _fortran_path(self) -> str:
+        return join_path(self.prefix, "bin", "armflang")
+
+    def _standard_flag(self, *, language: str, standard: str) -> str:
+        flags = {
+            "cxx": {"11": "-std=c++11", "14": "-std=c++14", "17": "-std=c++17"},
+            "c": {"99": "-std=c99", "11": "-std=c11"},
+        }
+        return flags[language][standard]
+
+    def setup_compiler_environment(self, env: EnvironmentModifications) -> None:
+        """Set up the build environment for this compiler package."""
+
+        # Prefer LLVM binutils (needed by LTO).
+        env.set("AR", join_path(self.prefix, "bin", "llvm-ar"))
+        env.set("NM", join_path(self.prefix, "bin", "llvm-nm"))
+        env.set("RANLIB", join_path(self.prefix, "bin", "llvm-ranlib"))
+
+    def setup_run_environment(self, env: EnvironmentModifications) -> None:
+        """Set up the run environment for this package."""
+        self.setup_compiler_environment(env)
+
+    def setup_dependent_build_environment(self, env: EnvironmentModifications, _: Spec) -> None:
+        """Set up the build environment for packages that depend on this one."""
+        self.setup_run_environment(env)
--- a/repos/spack_repo/builtin/packages/acfl/detection_test.yaml
+++ b/repos/spack_repo/builtin/packages/acfl/detection_test.yaml
@@ -81,3 +81,22 @@ paths:
       echo "InstalledDir: /usr/bin"
   platforms: [linux]
   results: []
+
+- layout: # does not detect ATfL
+  - executables:
+    - "bin/armclang"
+    - "bin/armclang++"
+    script: |
+      echo "Arm Toolchain for Linux 20.1 clang version 20.1.0"
+      echo "Target: aarch64--linux-gnu"
+      echo "Thread model: posix"
+      echo "InstalledDir: /opt/arm/arm-toolchain-for-linux-20.1/bin"
+  - executables:
+    - "bin/armflang"
+    script: |
+      echo "Arm Toolchain for Linux 20.1 flang version 20.1.0"
+      echo "Target: aarch64--linux-gnu"
+      echo "Thread model: posix"
+      echo "InstalledDir: /opt/arm/arm-toolchain-for-linux-20.1/bin"
+  platforms: [linux]
+  results: []
--- a/repos/spack_repo/builtin/packages/atfl/detection_test.yaml
+++ b/repos/spack_repo/builtin/packages/atfl/detection_test.yaml
@@ -1,0 +1,53 @@
+paths:
+- layout:
+  - executables:
+    - "bin/armclang"
+    - "bin/armclang++"
+    script: |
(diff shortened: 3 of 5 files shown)


pawosm-arm:
These gitlab-ci failures look like infrastructural issue. Any chances for a review?

pawosm-arm:
`All checks have passed` after the rebase. Can I get my review now?...

pawosm-arm:
@alalazo can I kindly ask you to look into it? I can't edit the reviewers list myself, unfortunately.

alalazo:
What is the difference between `atfl` and `acfl`?

pawosm-arm:
> What is the difference between `atfl` and `acfl`?

ACfL (Arm Compiler for Linux) is our old toolchain, now phased out, while ATfL (Arm Toolchain for Linux) is the new one, a member of a bigger family of the new ATf* toolchains.

repos/spack_repo/builtin/packages/armpl_gcc/package.py
```diff
@@ -417,11 +418,14 @@ class ArmplGcc(Package):
...
-
-    depends_on("gmake", type="build")
+    depends_on("c", type="build", when="+examples")
+    depends_on("fortran", type="build", when="+examples")
+    depends_on("gmake", type="build", when="+examples")
+    requires(
+        "^[virtuals=c,fortran] gcc",
+        when="+examples",
+        msg="armpl-gcc examples require GCC toolchain",
+    )
```
alalazo: You can collect the `when=` condition:
```python
with when("+examples"):
    depends_on("c", type="build")
    ...
```

repos/spack_repo/builtin/packages/armpl_gcc/package.py
```diff
@@ -522,6 +529,8 @@ def setup_run_environment(self, env: EnvironmentModifications) -> None:
 
     @run_after("install")
     def check_install(self):
+        if not self.spec.satisfies("+examples"):
+            return
```
alalazo: This should work too:
```suggestion
@run_after("install", when="+example")
def check_install(self):
```

repos/spack_repo/builtin/packages/atfl/package.py
```diff
@@ -0,0 +1,353 @@
...
+
+from spack.directives import (
+    conflicts,
+    depends_on,
+    maintainers,
+    provides,
+    requires,
+    variant,
+    version,
+)
```
alalazo: These should not be needed. The public API in:
```python
from spack.package import *
```
is importing these names.

repos/spack_repo/builtin/packages/atfl/package.py
```diff
@@ -0,0 +1,353 @@
...
+from spack.error import InstallError
+from spack.llnl.util.filesystem import install_tree
+from spack.package import *
+from spack.platforms import host as host_platform
+from spack.util.executable import Executable
+from spack.version import Version
```
alalazo: Most of these imports are unnecessary too, see https://spack.readthedocs.io/en/latest/package_api.html

repos/spack_repo/builtin/packages/atfl/package.py
```diff
@@ -0,0 +1,353 @@
...
+}
+
+
+class Atfl(Package, CompilerPackage):
+    """Arm Toolchain for Linux (ATfL): LLVM-based AArch64 compilers.
+
+    Installs the distro package (RPM/DEB) from Arm's repos, without root.
+    """
+
+    _name = "atfl"
```
alalazo: Why is this needed?

repos/spack_repo/builtin/packages/atfl/package.py
```diff
@@ -0,0 +1,353 @@
...
+    # Optional: allow opt-in to lld via the spec. Default is GNU ld for
+    # maximum compatibility; packages that prefer lld can request `%atfl+lld`.
+    variant("lld", default=False, description="Use lld linker (adds -fuse-ld=lld)")
+
+    # Arm Performance Libraries are required: ATfL's default configuration
+    # links against ArmPL (e.g. -fveclib=ArmPL -> -lamath). Ensure it's
+    # present and wired into the environment so linking always succeeds.
+    depends_on("armpl-gcc~examples", type="run")
+
+    def install(self, _: str, prefix: str) -> None:
```
alalazo: Hmm, these type hints are incorrect...

alalazo (changes requested):
Started reviewing, still have to finish going through the recipe. I'll post the current comments in the meanwhile.

alalazo:
> ACfL (Arm Compiler for Linux) is our old toolchain, now phased out, while ATfL (Arm Toolchain for Linux) is the new one, a member of a bigger family of the new ATf* toolchains.

Should we deprecate ACfL then? Or maintain both? Also pinging @paolotricerri as a maintainer for the `acfl` package

repos/spack_repo/builtin/packages/atfl/package.py
pawosm-arm: @LukeIreland1 or @robcow-arm do you know?

repos/spack_repo/builtin/packages/atfl/package.py
pawosm-arm: @LukeIreland1 or @robcow-arm FYI

pawosm-arm:
> > ACfL (Arm Compiler for Linux) is our old toolchain, now phased out, while ATfL (Arm Toolchain for Linux) is the new one, a member of a bigger family of the new ATf* toolchains.
> 
> Should we deprecate ACfL then? Or maintain both? Also pinging @paolotricerri as a maintainer for the `acfl` package

As long as it is still downloadable ( https://developer.arm.com/Tools%20and%20Software/Arm%20Compiler%20for%20Linux#Downloads ) IMO it's good to keep it. I assume it will be available there at least for the coming months.

paolotricerri:
I agree with @pawosm-arm as in until acfl can be downloaded and the link works, I would not deprecate the package yet. However, at the next release of ArmPL (the other package I maintain) I will also review our support for acfl.

repos/spack_repo/builtin/packages/atfl/package.py
pawosm-arm: Apparently, not needed. I'll delete it.

repos/spack_repo/builtin/packages/atfl/package.py
pawosm-arm: I'll try to copy the type hints from `go_sh/package.py` hopefully it has them right.

repos/spack_repo/builtin/packages/armpl_gcc/package.py
pawosm-arm: changed

repos/spack_repo/builtin/packages/armpl_gcc/package.py
pawosm-arm: changed

repos/spack_repo/builtin/packages/atfl/package.py
pawosm-arm: removed and replaced with a single import, still works, at least for me locally

repos/spack_repo/builtin/packages/atfl/package.py
pawosm-arm: removed them all. still works for me, at least locally.

repos/spack_repo/builtin/packages/atfl/package.py
pawosm-arm: copied. still works for me, at least locally.

pawosm-arm:
@alalazo can I kindly ask you to look into it again?

repos/spack_repo/builtin/packages/armpl_gcc/package.py
```diff
@@ -520,7 +524,7 @@ def setup_run_environment(self, env: EnvironmentModifications) -> None:
         else:
             env.prepend_path("LD_LIBRARY_PATH", join_path(armpl_dir, "lib"))
 
-    @run_after("install")
+    @run_after("install", when="+example")
```
LukeIreland1: `examples`?

pawosm-arm: I just gave you some `examples`.

Nice catch.

repos/spack_repo/builtin/packages/atfl/package.py
```diff
@@ -0,0 +1,337 @@
...
+    os_key = _alias.get(spack_os, spack_os)
+    latest = max(_VERSIONS, key=Version)
+    for ver, per_os in _VERSIONS.items():
+        pkg = per_os.get(os_key, None)
+        if pkg:
+            version(ver, sha256=pkg[0], url=pkg[1], expand=False, preferred=(ver == latest))
+
+    def archspec_name(self) -> str:
+        """Return the compiler name to use for archspec queries."""
+        return "clang"
```
alalazo: (Minor) Can you put methods after directives, that's what is usually done in other packages?

repos/spack_repo/builtin/packages/atfl/package.py
```diff
@@ -0,0 +1,337 @@
...
+
+    # Linux AArch64 only
+    requires("platform=linux", msg="ATfL is only available on Linux")
+    conflicts("target=x86_64:", msg="ATfL provides AArch64-native compilers")
+    conflicts("target=ppc64:", msg="ATfL provides AArch64-native compilers")
+    conflicts("target=ppc64le:", msg="ATfL provides AArch64-native compilers")
```
alalazo: Would it work to say:
```suggestion
    requires("target=aarch64:", msg="ATfL provides AArch64-native compilers")
```
?

repos/spack_repo/builtin/packages/atfl/package.py
```diff
@@ -0,0 +1,337 @@
...
+    Installs the distro package (RPM/DEB) from Arm's repos, without root.
+    """
+
+    maintainers("pawosm-arm")
+    homepage = "https://developer.arm.com/documentation/110477"
+
+    spack_os = host_platform().default_os
+    _alias: ClassVar[Dict[str, str]] = {"rocky10": "rhel10", "rocky9": "rhel9", "rocky8": "rhel8"}
+    os_key = _alias.get(spack_os, spack_os)
+    latest = max(_VERSIONS, key=Version)
```
alalazo: I think we can skip this, since it should be the default behavior to prefer the latest version

repos/spack_repo/builtin/packages/atfl/package.py
```diff
@@ -0,0 +1,337 @@
...
+class Atfl(Package, CompilerPackage):
+    """Arm Toolchain for Linux (ATfL): LLVM-based AArch64 compilers.
+
+    Installs the distro package (RPM/DEB) from Arm's repos, without root.
+    """
+
+    maintainers("pawosm-arm")
+    homepage = "https://developer.arm.com/documentation/110477"
+
+    spack_os = host_platform().default_os
```
alalazo: Can we remove this from class level, and avoid "polluting" the class scope with temporary variables?

repos/spack_repo/builtin/packages/atfl/package.py
```diff
@@ -0,0 +1,337 @@
...
+        """Install the package."""
+        archive = self.stage.archive_file
+        extract_dir = Path(self.stage.path) / "extract"
+        extract_dir.mkdir(exist_ok=True)
+        if archive.endswith(".rpm"):
+            Executable("bsdtar")("-xf", archive, "-C", str(extract_dir))
+        elif archive.endswith(".deb"):
+            Executable("dpkg-deb")("-x", archive, str(extract_dir))
+        else:
+            raise InstallError(f"Unknown archive type: {archive}")
```
alalazo, line 144: (minor) Do we have any way to download a tarball like `acfl` is doing? That would be more consistent with other packages (I currently see zero uses of `dpkg-deb` and two uses of `bsdtar` from nvidia packages).

repos/spack_repo/builtin/packages/atfl/package.py
```diff
@@ -0,0 +1,337 @@
...
+    def _toolroot(self) -> Path:
+        return Path(self.prefix)
+
+    def _has_amath(self, d: Path) -> bool:
+        return (d / "libamath.a").exists() or any(d.glob("libamath.so*"))
+
+    def _resolve_armpl_spec(self) -> Optional[Path]:
+        if "armpl-gcc" in self.spec:
+            armpl: Spec = self.spec["armpl-gcc"]  # type: ignore[assignment]
+            parent = Path(armpl.prefix)
```
alalazo: This should also work:

```suggestion
            parent = Path(self["armpl-gcc"].prefix)
```

repos/spack_repo/builtin/packages/atfl/package.py
```diff
@@ -0,0 +1,337 @@
...
+            if parent.is_dir():
+                # Typical armpl layout: <prefix>/armpl_<ver>_*/lib
+                for cand in parent.iterdir():
+                    if cand.is_dir() and cand.name.startswith("armpl_"):
+                        libdir = cand / "lib"
+                        if self._has_amath(libdir) or self._has_amath(
+                            libdir / "aarch64-unknown-linux-gnu"
+                        ):
+                            return cand
+                # Fallback: parent already is the root
```
alalazo: (This is more for my knowledge than anything else) Why do we need to search this dynamically? Don't we know the directory structure given the version?

repos/spack_repo/builtin/packages/atfl/package.py
```diff
@@ -0,0 +1,337 @@
...
+            # Last resort: bounded search under the toolroot
+            for hit in list((tr).rglob("libamath.so*")) + list((tr).rglob("libamath.a")):
+                # Make the root be the first ancestor that has a 'lib' directory
+                # so setup_* can do root/'lib' and root/'lib/aarch64-unknown-linux-gnu'.
+                for ancestor in (hit.parent, hit.parent.parent, tr):
+                    if (ancestor.parent / "lib").is_dir() or (ancestor / "lib").is_dir():
+                        return tr
+        return None
+
+    def _find_armpl_root(self) -> Optional[Path]:
```
alalazo: Can you explain why this search function needs to be so complicated?

repos/spack_repo/builtin/packages/atfl/package.py
```diff
@@ -0,0 +1,337 @@
...
+
+    def setup_compiler_environment(self, env: EnvironmentModifications) -> None:
+        """Set up the build environment for this compiler package."""
+        tr = self._toolroot()
+        armpl_root = self._find_armpl_root()
+
+        env.prepend_path("PATH", Path(tr) / "bin")
+        env.prepend_path("CPATH", Path(tr) / "include")
+        env.prepend_path("LIBRARY_PATH", Path(tr) / "lib")
+        env.append_path("LIBRARY_PATH", Path(tr) / "lib" / "aarch64-unknown-linux-gnu")
```
alalazo: Instead of setting these variables, can't we configure the compiler when we install it?

repos/spack_repo/builtin/packages/atfl/package.py
pawosm-arm: Unfortunately, there is no going back to how ACfL have been developed and distributed. I don't want to run into politics here, but in case of the new (ATfL) toolchain, we are targeting specific Linux distributions, providing packages in a format that is native to them, so there are no tarballs expected for final builds of each ATfL version.

alalazo: Fine. If that's the only way the package is distributed, I guess we can't do much about it.

repos/spack_repo/builtin/packages/atfl/package.py
pawosm-arm: According to my colleagues, it is not complicated, it's comprehensive. They are giving me examples of the use cases (e.g., with things provided externally vs. through spack) that would never be exercised by a typical users (me included), but they claim, they do that. What can I say, I wouldn't really want to cripple it if they claim it works...

repos/spack_repo/builtin/packages/atfl/package.py
pawosm-arm: we don't know the version at this point, which is the entire point of the resolver.

repos/spack_repo/builtin/packages/atfl/package.py
pawosm-arm: This sounds awkward. Could you give an alternative as a suggestion? Your suggestion would be really helpful.

repos/spack_repo/builtin/packages/atfl/package.py
alalazo: > According to my colleagues, it is not complicated, it's comprehensive.

To be clear, I didn't mean to imply anything negative, but I thought that - given a version - we'd know the layout of the software. I was wondering in which scenarios this is not true (and if we know that, it's worth a comment in the code).

repos/spack_repo/builtin/packages/atfl/package.py
alalazo: > we don't know the version at this point, which is the entire point of the resolver.

Hmm, I don't follow you on this. If I didn't miss anything, this function is called only by:
- `setup_run_environment`
- `setup_dependent_build_environment`

These methods are always called when we already know a concrete spec (i.e. the solver already produced it). This means that we can inspect the version or other details using:
```python
if self.spec.satisfies("<constraint>"):
    ...
```
Can we simplify the implementation here given the above?

repos/spack_repo/builtin/packages/atfl/package.py
pawosm-arm: I'm genuinely confused, because I'm sure that's exactly what the code is doing. This code is executed when the compiler is installed (or used, or both). LLVM does something identical.

repos/spack_repo/builtin/packages/atfl/package.py
pawosm-arm: I'd happily remove this, but AFAIR @LukeIreland1 observed some inconsistent behavior here, hence this line bas been added. @LukeIreland1 are we ok with removing it?

repos/spack_repo/builtin/packages/atfl/package.py
pawosm-arm: Seems I've been too fast with my comment, I'm sorry for that. Going back to the drawing board.

repos/spack_repo/builtin/packages/atfl/package.py
LukeIreland1: It's been a long time since I added this line, so I can't remember why exactly, but it would have been something along the lines of, the "latest" detection wasn't working in the way I'd expect.

I think it may be worth removing `latest`, and its use, and seeing if the built-in latest resolver works as expected.

repos/spack_repo/builtin/packages/atfl/package.py
pawosm-arm: TBH, @LukeIreland1 I can't find `os_key` (the only user of `spack_os`) used anywhere in spack... and the `spack_os = ...` I presume has been copy-pasted from the old acfl package definition?

pawosm-arm: I meant, this may be something like this:
```
+    _alias: ClassVar[Dict[str, str]] = {"rocky10": "rhel10", "rocky9": "rhel9", "rocky8": "rhel8"}
+    for ver, per_os in _VERSIONS.items():
+        pkg = per_os.get(_alias.get(host_platform().default_os, host_platform().default_os), None)
+        if pkg:
+            version(ver, sha256=pkg[0], url=pkg[1], expand=False)
```

repos/spack_repo/builtin/packages/atfl/package.py
pawosm-arm: yes it would. I've tried it on some x86_64 and it printed:

```
==> Error: failed to concretize `atfl` for the following reasons:
     1. atfl: ATfL provides AArch64-native compilers
```

repos/spack_repo/builtin/packages/atfl/package.py
pawosm-arm: moved down.

repos/spack_repo/builtin/packages/atfl/package.py
pawosm-arm: changed as suggested.

repos/spack_repo/builtin/packages/atfl/package.py
pawosm-arm: changed as shown above.

repos/spack_repo/builtin/packages/atfl/package.py
pawosm-arm: changed as suggested.

repos/spack_repo/builtin/packages/atfl/package.py
pawosm-arm: removed `latest` and its use. seems to still work ok.

repos/spack_repo/builtin/packages/atfl/package.py
pawosm-arm: @alalazo could you give some hint here?

pawosm-arm:
I've addressed some of the comments. The rest will require some more discussion (or just adding comments in the code)

repos/spack_repo/builtin/packages/atfl/package.py
alalazo: This code is executed when the compiler is run, or a dependency needs it. I need to check if having two functions is redundant, but besides that I was wondering if we can dump a configuration file in the installation directory to set the include paths and library paths. 

This is to help usage outside of Spack, which would otherwise require users to set include paths and library paths.

repos/spack_repo/builtin/packages/atfl/package.py
pawosm-arm: I've added the line of comment we've fiercely negotiated offline.

repos/spack_repo/builtin/packages/atfl/package.py
pawosm-arm: `Can we simplify the implementation here given the above?` @LukeIreland1 ?

repos/spack_repo/builtin/packages/atfl/package.py
LukeIreland1: We already ship a modulefile for external users; that’s our recommended outside‑Spack mechanism. If you want a config file in prefix/etc, we can consider that too, but we still need Spack’s env hooks for dependent builds.

repos/spack_repo/builtin/packages/atfl/package.py
LukeIreland1: Well, called by setup_compiler_environment (through _find_armpl_root) which is called by the other two.

We're already using the spec where possible for resolving (hence why it's first in` _find_armpl_root`), but we could go further. Either by using spec.satisfies as @alalazo suggested, or by updating the depends_on("armpl-gcc~examples"... line to just support armpl_gcc >= 24.04, we can probably safely use a template to use the version like this:
root = prefix / f"armpl_{self.spec.version}_gcc"

My original intention was to be flexible, but the last of the resolvers already globs, so I suppose there's a bit of redundancy as things stand.

repos/spack_repo/builtin/packages/atfl/package.py
pawosm-arm: I agree with Luke here, we're giving recommendations on our official pages (https://developer.arm.com/documentation/110477/211/Getting-started?lang=en) and it is to use the environment modules. Users are expecting these to be set, the same policy we had with ACfL, we should not change it now.

pawosm-arm: Note that myself I'm using ATfL without loading modules and that works too. Clang (armclang) and Flang (armflang) knows where its config files, headers and libs are, using relative paths and runtime paths. There's nothing to configure here really, we're adding those env variables to be in line with our own recommendations.

pawosm-arm: I've rewritten this. The whole part is gone.

repos/spack_repo/builtin/packages/atfl/package.py
pawosm-arm: I've rewritten this. The whole part is gone.

pawosm-arm:
Been trying to build SW4 with this and failed. So I decided to rewrite the whole thing. I'll address the style errors tomorrow.

pawosm-arm:
Style errors corrected, @alalazo could you look at this again?

repos/spack_repo/builtin/packages/armpl_gcc/package.py
```diff
@@ -421,11 +422,11 @@ class ArmplGcc(Package):
...
-    depends_on("c", type="build")
-    depends_on("fortran", type="build")
-    requires("^[virtuals=c,fortran] gcc", msg="armpl-gcc is only compatible with the GCC compiler")
-
-    depends_on("gmake", type="build")
+    with when("+examples"):
+        depends_on("c", type="build")
+        depends_on("fortran", type="build")
+        depends_on("gmake", type="build")
+        requires("^[virtuals=c,fortran] gcc", msg="armpl-gcc examples require GCC toolchain")
```
alalazo: This:
```suggestion
        requires("%c,fortran=gcc", msg="armpl-gcc examples require GCC toolchain")
```
should be more precise (it's a direct dependency).

pawosm-arm:
I've moved the `PKG_CONFIG_PATH` change in the `armpl_gcc` package into a separate PR (#3513 ) as it is outside the scope of this PR.

pawosm-arm:
```
ERROR: Job failed (system failure): pod "pipeline/runner-tlzqdu3ar-project-57-concurrent-1-z946dl0q" is disrupted: reason "TerminationByKubelet", message "Pod was terminated in response to imminent node shutdown."
```

pawosm-arm:
@alalazo there were a lot of changes to this PR recently which overlap with your comments. Could you summarize what your concerns would be after these changes?

repos/spack_repo/builtin/packages/armpl_gcc/package.py
alalazo, line 1: Can we extract changes to `armpl_gcc` to a different PR, so they are not conflated?

repos/spack_repo/builtin/packages/atfl/package.py
```diff
@@ -0,0 +1,218 @@
...
+        "-gline-tables-only",
+        "-gmodules",
+        "-g",
+    ]
+
+    opt_flags = ["-O0", "-O1", "-O2", "-O3", "-Ofast", "-Os", "-Oz", "-Og", "-O", "-O4"]
+
+    def archspec_name(self) -> str:
+        """Return the compiler name to use for archspec queries."""
+        return "clang"
```
alalazo, line 132: This means the compiler will use the same version scheme and options as upstream clang. Is it correct?

alalazo (commented):
Mostly LGTM. A minor request to merge changes in behavior for `armpl_gcc` in a separate PR and a question.

repos/spack_repo/builtin/packages/atfl/package.py
pawosm-arm: It will use the same version argument, but not the same version regex:
```
compiler_version_regex = r"Arm Toolchain for Linux [\d\.]+ [cf]lang version ([\d\.]+)"
```
The options, yes, we would wish them to be the same. But we don't have any governance mechanism covering that wish. I doubt it would ever happen that we may diverge, but it's not guaranteed either.
I'd see how much of the things above overlap. I'm considering taking the risk of removal of the things that overlap.

repos/spack_repo/builtin/packages/armpl_gcc/package.py
pawosm-arm: This has been extracted to here: https://github.com/spack/spack-packages/pull/3664

repos/spack_repo/builtin/packages/armpl_gcc/package.py
pawosm-arm: Yes, I've considered this in the new PR, https://github.com/spack/spack-packages/pull/3664

repos/spack_repo/builtin/packages/atfl/package.py
alalazo: > It will use the same version argument, but not the same version regex

That's fine. Point is: if we query archspec for options using clang, then we must be sure that the version we pass is the same as clang - so if versioning starts to be different in the future, we need to revisit this.

pawosm-arm: The `[cf]lang version ([\d\.]+)` version part is always copied from `cmake/Modules/LLVMVersion.cmake`.

pawosm-arm: > I'm considering taking the risk of removal of the things that overlap.

Apparently, only the `compiler_version_argument = "--version"` can be removed as it is being inherited from the `LlvmDetection` class, the rest must be set here, similarly as in other LLVM-derived toolchains (including `apple_clang`).

pawosm-arm:
Ah, I've introduced this change too soon. The #3664 must land first... I'll rebase when it does.

pawosm-arm:
> Ah, I've introduced this change too soon. The #3664 must land first... I'll rebase when it does.

Done.

pawosm-arm:
@alalazo anything else I should do?

alalazo (approved):

alalazo:
@pawosm-arm Nope, LGTM thanks!

pawosm-arm:
@alalazo Thank you :)

PR #2782, opened 2025-12-16
Bump up the version for rocm-7.1.1 release.

srekolam:

PR #2811, opened 2025-12-17
rocprofiler-systems: add boost version constraint because of cmake issue

msimberg:
Adding a version constraint because of this cmake configuration issue reported upstream: https://github.com/ROCm/rocm-systems/issues/2356. rocprofiler-systems with `+internal-dyninst` does not recognize boost because it expects boost system to be found. However, boost's system stub library was removed in 1.89.


afzpatel (approved):
LGTM, @srekolam has added this change to https://github.com/spack/spack-packages/pull/2782 but that PR might take longer to merge.

alalazo (approved):

PR #2816, opened 2025-12-17
hpcviewer: add version 2025.3.1

mwkrentel:
Add version 2025.3.1 which no longer uses Java as a prereq. 
Omit 2025.3.1 for ppc64le where Eclipse requires a much more recent version of glibc.
Drop 2025.2 for MacOS, it doesn't run due to Apple signing issue.

### DIFF
--- a/repos/spack_repo/builtin/packages/hpcviewer/package.py
+++ b/repos/spack_repo/builtin/packages/hpcviewer/package.py
@@ -4,6 +4,7 @@
 
 import os
 import platform
+import shlex
 
 from spack_repo.builtin.build_systems.generic import Package
 
@@ -13,7 +14,7 @@ from spack.package import *
 class Hpcviewer(Package):
     """Binary distribution of hpcviewer and integrated hpctraceviewer for
     the Rice HPCToolkit (Linux x86_64, ppc64le and aarch64, and MacOSX
-    x86_64 and M1/M2).
+    x86_64 and Apple M-series).
 
     Note: hpctoolkit databases are platform independent, so you don't
     need to install hpctoolkit to run the viewers and it's common to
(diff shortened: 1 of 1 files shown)


mwkrentel:
@spackbot fix style

mwkrentel:
I noticed that the style "fix" looks really awful.
```
  viewer_sha = {
      (
          "2025.3.1",
          "aarch64",
      ): "bf9d519cb679bf1ee49da46a6eb3c5f1f223f58e3228c53b9c13d6368dcc16dd",
      ("2025.3.1", "x86_64"): "3ad32708332fc61d53c67d93c17eeec3659e127df14552665f5689ad62773145",
      (
          "2025.2.0",
          "aarch64",
      ): "565bba471273f897c3436dd8b152720e7862a44588204171b8778a2f5b463dfc",
      (
          "2025.2.0",
          "ppc64le",
      ): "79ee51a1d0bc514571dfa7ca70b667e05b935cd0de8bc2a7603b304d2d357adc",
      ("2025.2.0", "x86_64"): "e468637b212c03ebc121f8e2933e52885cd9133216b6b90262ab1441a9158959",
      ("2025.01", "aarch64"): "5cb9c0a9fb15fa9128da0d609554a76541d726737f4bc6b42bee87a2b9fe7e3d",
```

Apparently, when the version number has 3 segments(`2025.3.1`) and the arch name
is `aarch64` or `ppc64le` which are one char longer than `x86_64`, I think the line
is ONE char over the limit.  This looks really awful and inconsistent.

Is there a way to override this?
I thought lines with an obvious hash were exempt from line length.

mwkrentel:
@spackbot restart pipeline

mwkrentel:
Ok, I give up.  How do I restart the pipeline?
It just says "timedout", but I don't think I wrote an infinite loop.

mwkrentel:
@tldahlgren Would you like to review and merge?
It's really just the latest rev and changes to the build system
to support the new format.
Thanks

repos/spack_repo/builtin/packages/hpcviewer/package.py
```diff
@@ -27,10 +28,10 @@ class Hpcviewer(Package):
 
     darwin_sha = {
         (
-            "2025.2.0",
+            "2025.3.1",
             "aarch64",
-        ): "76da2fb0d371be29ee9f4dc090021b095e23ae6851079b5a9d4161401d8c1170",
-        ("2025.2.0", "x86_64"): "7cbb01281b32c7c2b44713c3434d24f0b3c166ba9ae53434b233a150eccd19ab",
+        ): "c87d29f70bc39c23926c001d8393d48d2fdf1ae41da8096a5f2ab40b18d20105",
+        ("2025.3.1", "x86_64"): "6b0f3d8e77fe4e0a7b6a97d022fba5671bc94762ff2d3c71ea4a3ab377d8fff2",
```
becker33, line 34: Is there a good reason that this is removing 2025.2.0 instead of just adding a new entry for the newer version?

mwkrentel: @becker33 Yes!  The reason is that 2025.2.0 doesn't run on Macos due to some
signing issue.  Macos treats it as unauthorized and possibly malware.
I guess we didn't check carefully enough.

This applies only to Macos and the `darwin_sha` list.
We've fix it in 2025.3.1.

I don't suppose there's a way to work around the style issue above?
I think we're right on the border for wrapping lines and some wrap
and some don't and it looks ugly.

Anyway, thanks!  Matt will appreciate this.

becker33 (approved):

PR #2822, opened 2025-12-17
esmf: fix url_for_version

danrosen25:
The url_for_version function needs to be in the Esmf class. Prior to this change the esmf package was using the internal url_for_version, which did not support ESMF versions prior to 8.0.0.  This change makes it possible to install ESMF versions prior to 8.0.0.

**Change Log**
- add @danrosen25 as maintainer
- move url_for_version to appropriate class
- only copy cmake file for esmf v8+

**Testing**
- esmf@8.9.0 installed with gcc-14 (on ubuntu:plucky)
- esmf@7_1_0r installed with gcc-6 (on ubuntu:bionic)

In order to install ESMF 7.1.0 use `spack install --no-checksum esmf@=7_1_0r`. This version of ESMF is not patched and will not work with newer compilers.

### DIFF
--- a/repos/spack_repo/builtin/packages/esmf/package.py
+++ b/repos/spack_repo/builtin/packages/esmf/package.py
@@ -26,7 +26,9 @@ class Esmf(MakefilePackage, PythonExtension):
     url = "https://github.com/esmf-org/esmf/archive/v8.4.1.tar.gz"
     git = "https://github.com/esmf-org/esmf.git"
 
-    maintainers("climbfuji", "jedwards4b", "AlexanderRichert-NOAA", "theurich", "uturuncoglu")
+    maintainers(
+        "climbfuji", "jedwards4b", "AlexanderRichert-NOAA", "theurich", "uturuncoglu", "danrosen25"
+    )
 
     # Develop is a special name for spack and is always considered the newest version
     version("develop", branch="develop")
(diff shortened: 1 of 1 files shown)


theurich (approved):
Makes sense and looks good to me.

climbfuji (approved):

repos/spack_repo/builtin/packages/esmf/package.py
```diff
@@ -139,6 +141,18 @@ def patch(self):
...
+        if version < Version("8.0.0"):
+            # Older ESMF releases had a custom tag format ESMF_x_y_z
+            return "https://github.com/esmf-org/esmf/archive/ESMF_{0}.tar.gz".format(
+                version.underscored
+            )
+        else:
+            # Starting with ESMF 8.0.0 releases are in the form vx.y.z
+            return "https://github.com/esmf-org/esmf/archive/refs/tags/v{0}.tar.gz".format(
+                version.dotted
+            )
```
alalazo, line 161: The oldest version in this recipe is v8.0.1 Is there a reason to keep code to install versions that are not checksummed?

danrosen25: @alalazo 
We had a request to install an older version of [ESMF v7.1.0r](https://github.com/esmf-org/esmf/tree/ESMF_7_1_0r) using spack. This lead me to execute a test installation `spack install --no-checksum esmf@=7_1_0r`, which did not work even though the `url_for_version` is written correctly. This function was just located in the wrong class. Once moved I was able to install ESMF v7.1.0r, but only with a much older gfortran compiler, v6, which matches up to our release testing. Internally, we discussed patching v7.1.0r but it's not our directive to offer LTS for older releases, especially outdated major release.

The short answer is that this is purely a bug fix. It will allow you to install older versions but since they may only work with older compiler versions, which are out of support, we're not supporting these installations. I **CAN** however add `7_1_0r` to a list of deprecated versions if you'd like?

danrosen25: @alalazo I added version `7.1.0r` but it's marked deprecated.

theurich (approved):

alalazo (approved):

PR #2823, opened 2025-12-17
chapel package -- add requires for gcc version

ajpotts:
Add GCC compiler constraints to the Chapel Spack package to prevent unsupported compiler configurations and avoid LLVM/libstdc++ incompatibility issues.

### DIFF
--- a/repos/spack_repo/builtin/packages/chapel/package.py
+++ b/repos/spack_repo/builtin/packages/chapel/package.py
@@ -477,6 +477,12 @@ class Chapel(AutotoolsPackage, CudaPackage, ROCmPackage):
 
     conflicts("platform=windows")  # Support for windows is through WSL only
 
+    requires(
+        "%gcc@9:12",
+        msg="Chapel requires GCC 9–12 due to LLVM and libstdc++ compatibility"
+    )
+
+
     # Ensure GPU support is Sticky: never allow the concretizer to choose this
     variant("rocm", default=False, sticky=True, description="Enable AMD ROCm GPU support")
     variant("cuda", default=False, sticky=True, description="Enable Nvidia CUDA GPU support")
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/chapel/package.py
```diff
@@ -477,6 +477,12 @@ class Chapel(AutotoolsPackage, CudaPackage, ROCmPackage):
 
     conflicts("platform=windows")  # Support for windows is through WSL only
 
+    requires(
+        "%gcc@9:12",
+        msg="Chapel requires GCC 9–12 due to LLVM and libstdc++ compatibility"
+    )
```
bonachea: I'm not sure but I suspect this is Just Plain Wrong. IIUC Chapel does NOT have an unconditional requirement for `gcc`.

Chapel supports builds with a wide variety of compiler families, see the `compiler_map` variable and the corresponding `requires` directive on line 518. The [Chapel documentation for compiler requirements is here](https://chapel-lang.org/docs/main/usingchapel/prereqs.html), but I believe that list is incomplete.

I don't think this spackage currently enforces acceptable compiler version limits, but this seems like the wrong way to do it.

bonachea (changes requested):
I suspect this PR is incorrect as written and could cause regressions on important platforms of interest to Chapel users.

Please provide more detailed information about the problem you are attempting to solve (is there an open issue?), and whether it impacts Chapel builds using non-gcc compilers (e.g. `chapel llvm=none %cce`).

Side-note: there's already an open PR #2749 where the new Chapel release is being integrated and tested.

CC: @e-kayrakli

repos/spack_repo/builtin/packages/chapel/package.py
ajpotts: I’m trying to address an Arkouda-impacting concretization/build issue.  

Using `requires("%gcc@9:12")` without a `when="%gcc"` guard was overly restrictive, so I added the `when` and converted to a `conflicts` instead of `requires` to make it clear that it only applies to situations when `%gcc` is selected.

arifthpe: Could you point us to or otherwise explain that issue?

ajpotts: I'm going to close this temporarily until I have time to reproduce the issue so I can share more details with you.

ajpotts:
Closing until I can reproduce the related issue.

PR #2824, opened 2025-12-18
py-onnxruntime: add v1.23.2, fix Clang build, update dependencies

wdconinc:
Add version 1.23.2 with updated dependencies and fix build error with Clang.

## Clang build error fixed

```
error: too few arguments to function call, expected 3, have 2
  __builtin_ia32_tpause(0x0, __rdtsc() + tpause_spin_delay_cycles);
```

The `__builtin_ia32_tpause` intrinsic has different signatures:
- **GCC**: `__builtin_ia32_tpause(control, uint64_t counter)` [2 args]
- **Clang**: `__builtin_ia32_tpause(control, uint32_t edx, uint32_t eax)` [3 args]

Onnxruntime PR microsoft/onnxruntime#24524 (merged 2025-05-31, first in v1.23.0) called the builtin directly with 2 args, which works with GCC but fails with Clang.

The fix uses the portable `_tpause()` wrapper that both compilers provide. Applied to @1.23: (open-ended) since there's no upstream fix yet.

## Changes

- Add version 1.23.2
- Update abseil-cpp requirement to @20250512.0: for @1.23:
- Add py-pybind11@2.13.6: requirement for @1.21:
- Add flatbuffers~shared with version requirement for @1.18:
- Update protobuf to @3.21.12: for @1.23:
- Add onnx@1.18 and py-onnx@1.18 for @1.23
- Remove duplicate py-onnx@:1.16 constraint
- Add patch for missing cstdint include
- Add fix for __builtin_ia32_tpause Clang incompatibility

## References

- https://github.com/microsoft/onnxruntime/pull/24524
- gcc x86_64-linux-gnu/14/include/waitpkgintrin.h (2-arg builtin)
- llvm-project/clang/lib/Headers/waitpkgintrin.h (3-arg builtin)

### DIFF
--- a/repos/spack_repo/builtin/packages/py_onnxruntime/package.py
+++ b/repos/spack_repo/builtin/packages/py_onnxruntime/package.py
@@ -244,7 +244,9 @@ class PyOnnxruntime(CMakePackage, PythonExtension, ROCmPackage, CudaPackage):
             define("onnxruntime_USE_FULL_PROTOBUF", True),
             define("onnxruntime_DISABLE_CONTRIB_OPS", False),
             # Point to py-onnx for proto files (not in C++ onnx installation)
-            define("onnx_SOURCE_DIR", join_path(self.spec["py-onnx"].prefix, self.site_packages_dir)),
+            define(
+                "onnx_SOURCE_DIR", join_path(self.spec["py-onnx"].prefix, self.site_packages_dir)
+            ),
         ]
 
         if self.spec.satisfies("+cuda"):
--- a/repos/spack_repo/builtin/packages/py_onnxruntime/package.py
+++ b/repos/spack_repo/builtin/packages/py_onnxruntime/package.py
@@ -26,6 +26,7 @@ class PyOnnxruntime(CMakePackage, PythonExtension, ROCmPackage, CudaPackage):
 
     license("MIT")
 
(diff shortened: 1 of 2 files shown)


repos/spack_repo/builtin/packages/py_onnxruntime/package.py
```diff
@@ -191,6 +219,15 @@ def setup_run_environment(self, env: EnvironmentModifications) -> None:
...
 
+    @property
+    def python_lib_dir(self):
+        python_vers_phrase = "python{0}".format(self.spec["python"].version.up_to(2))
+        return join_path("lib", python_vers_phrase)
+
+    @property
+    def site_packages_dir(self):
+        return join_path(self.python_lib_dir, "site-packages")
+
```
tmadlener, line 230: Doesn't `python_platlib` do the same? (I can't seem to find where it is defined to know for sure, but quite a few packages use this to the same effect)

wdconinc: Yeah, this isn't even correct since it now fails for my free-threading python (which uses `python3.14t`).

wdconinc:
Superseded by #3735 (pr17-python-data-ml) which is a strict superset: includes the same fix-tpause-clang.patch, cpuinfo/eigen deps, python_lib_dir property, and additional updates.

wdconinc:
(Sorry, overactive AI agent close this unintentionally.)

PR #68, opened 2025-06-06
gaudi: allow newer fmt for v39

wdconinc:
> [!NOTE]
> This PR was migrated from **https://github.com/spack/spack/pull/49664**.

This PR updates `gaudi` to allow for fmt-11 when gaudi-39 or newer. See https://gitlab.cern.ch/gaudi/Gaudi/-/merge_requests/1627.

PR #2825, opened 2025-12-18
gaudi: workaround test-dependency bug with a `when`

wdconinc:
In #68 we added a more restrictive `fmt@:10` test dependency compared to the build/link/run dependency which is merely `fmt`. Both are declared without a `when` clause. That causes them to be [merged incorrectly](https://github.com/spack/spack/issues/51775) and we end up imposing the restrictive `fmt@:10` as build/link/run/test dependency.

Since dependencies are now [indexed by `when` clause](https://github.com/spack/spack/pull/40326), we can avoid this dependency merging bug by splitting the two problematic `depends_on` directives so they don't have the same `when` clause. This PR does exactly that and imposes the test dependency only for versions of `gaudi` where [the problematic check was included](https://gitlab.cern.ch/gaudi/Gaudi/-/commit/4dc10ab6ad578366d60e361fa8cbd3c54d726d27).

With this change, the reproducer in https://github.com/spack/spack/issues/51775 concretizes without conflicts and indeed gives a `gaudi ^fmt@11` solution.

### DIFF
--- a/repos/spack_repo/builtin/packages/gaudi/package.py
+++ b/repos/spack_repo/builtin/packages/gaudi/package.py
@@ -165,7 +165,7 @@ class Gaudi(CMakePackage, CudaPackage):
     depends_on("fmt@:8", when="@:36.9")
     depends_on("fmt@:10", when="@:38")
     depends_on("fmt@:11", when="@:39")
-    depends_on("fmt@:10", type="test")  # https://gitlab.cern.ch/gaudi/Gaudi/-/issues/345
+    depends_on("fmt@:10", when="@36.15:", type="test")  # https://gitlab.cern.ch/gaudi/Gaudi/-/issues/345
     depends_on("intel-tbb@:2020.3", when="@:37.0")
     depends_on("tbb", when="@37.1:")
     depends_on("uuid")
--- a/repos/spack_repo/builtin/packages/gaudi/package.py
+++ b/repos/spack_repo/builtin/packages/gaudi/package.py
@@ -165,9 +165,11 @@ class Gaudi(CMakePackage, CudaPackage):
     depends_on("fmt@:8", when="@:36.9")
     depends_on("fmt@:10", when="@:38")
     depends_on("fmt@:11", when="@:39")
-    depends_on(
-        "fmt@:10", when="@36.15:", type="test"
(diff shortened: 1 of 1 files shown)


vvolkl (dismissed):

alalazo:
It seems there is a legit build failure https://gitlab.spack.io/spack/spack-packages/-/jobs/19639364

wdconinc:
> It seems there is a legit build failure https://gitlab.spack.io/spack/spack-packages/-/jobs/19639364

~~Oh great, now it seems to ignore the `type="test"` dependency entirely.~~

repos/spack_repo/builtin/packages/gaudi/package.py
```diff
@@ -164,7 +164,9 @@ class Gaudi(CMakePackage, CudaPackage):
     depends_on("fmt@:8", when="@:36.9")
     depends_on("fmt@:10", when="@:38")
     depends_on("fmt@:11", when="@:39")
-    depends_on("fmt@:10", type="test")  # https://gitlab.cern.ch/gaudi/Gaudi/-/issues/345
+    depends_on(
```
jmcarcell: Maybe something like:
```
if self.run_tests:
     depends_on("fmt@:10")
```
? I can't suggest, maybe because of the copilot comment.

wdconinc:
Ah, I think I see it now. We also enable tests when `+examples`, not only when `--test` is passed.

repos/spack_repo/builtin/packages/gaudi/package.py
wdconinc: `self.run_tests` is not available in class definition scope.

wdconinc:
> Ah, I think I see it now. We also enable tests when `+examples`, not only when `--test` is passed.

Good. After 3d679c6 nothing needed to be rebuilt in CI since it again correctly picks up the fmt@10 constraint for the `+examples` imposed by the stacks configuration:
https://github.com/spack/spack-packages/blob/e3cafb574f3448c4bf4ce147cc9d94d04ea53478/stacks/hep/spack.yaml#L94

jmcarcell (approved):

bernhardkaindl (approved):

PR #2826, opened 2025-12-18
b2luigi: new package

CamCoop1:
Adding b2luigi (https://pypi.org/project/b2luigi/) beginning at v1.2.6 in which its dependencies were refactored to only include Spack compatible packages.


CamCoop1:
Pinging 

b2luigi maintainers: @AlexanderHeidelbach @JonasEppelt @GiacomoXT

FCC: @amanmdesai @jmcarcell @kjvbrt

CamCoop1:
@spackbot help

AlexanderHeidelbach:
Dear Spack devs,

is there currently a nice way to automate the version control of packages? From the b2luigi side we try to be up to speed with our user requirements and thus try to have a relativley fast release cycle. 

Going through the documentation and using the GitHub Copilot bot, we did not find any instructions how to automatically add the newest package release to Spack. The best idea seems to be to have an GitHub action that on release creates a PR with the new release version. Is this a recommended procedure or are there better ways?

CamCoop1:
@spackbot request review

repos/spack_repo/builtin/packages/py_b2luigi/package.py
```diff
@@ -0,0 +1,34 @@
...
+    version("1.2.6", sha256="9f3be756f0961ca2241d36d9a9174ea5a23ebd7787cbfa78632047aae25f1202")
+
+    depends_on("python@3.8:3.12", type=("build", "run"))
+    depends_on("py-luigi@3.0.2:", type=("build", "run"))
+    depends_on("py-parse@1.8:", type=("build", "run"))
+    depends_on("py-gitpython@2.1.11:", type=("build", "run"))
+    depends_on("py-colorama@0.3.9:", type=("build", "run"))
+    depends_on("py-cachetools@2.1.1:", type=("build", "run"))
+    depends_on("py-jinja2", type=("build", "run"))
+    depends_on("py-tenacity@8:9", type=("build", "run"))
```
jmcarcell: This means only version 8.X.Y until (and *included*) version 9.X.Y. Is this what you intend to say? I'm saying this because the latest version of `py-tenacity` is 9.1.2. With this I mean you can not know if the version range closes at 9 because there is no version above 9

```suggestion
    depends_on("py-tenacity@8", type=("build", "run"))
```

repos/spack_repo/builtin/packages/py_b2luigi/package.py
```diff
@@ -0,0 +1,34 @@
...
+
+    license("GNU")
+
+    # To be decided
+    # maintainers("github_user1", "github_user2")
+
+    # We start at 1.2.6 as this was the change from retry2->tenacity dependency
+    version("1.2.6", sha256="9f3be756f0961ca2241d36d9a9174ea5a23ebd7787cbfa78632047aae25f1202")
+
+    depends_on("python@3.8:3.12", type=("build", "run"))
```
jmcarcell:
```suggestion
    depends_on("python@3.8:", type=("build", "run"))
```
`pyproject.toml` only says `python>=3.8`

jmcarcell:
> Adding b2luigi ([pypi.org/project/b2luigi](https://pypi.org/project/b2luigi/)) beginning at v1.2.6 in which its dependencies were refactored to only include Spack compatible packages.

I'm not sure what you mean, if a package is not in Spack in can be added


> Dear Spack devs,
> 
> is there currently a nice way to automate the version control of packages? From the b2luigi side we try to be up to speed with our user requirements and thus try to have a relativley fast release cycle.
> 
> Going through the documentation and using the GitHub Copilot bot, we did not find any instructions how to automatically add the newest package release to Spack. The best idea seems to be to have an GitHub action that on release creates a PR with the new release version. Is this a recommended procedure or are there better ways?

I don't think there is an automated way. However something like `spack checksum py-b2luigi` will add the versions so the work that would be needed would be updating the dependencies and make the pull request. You are free to implement whatever you want, though. Updates also don't need to happen for every minor version, it could be done every O(months), but I see `b2luigi` doesn't have that many versions.

repos/spack_repo/builtin/packages/py_b2luigi/package.py
```diff
@@ -0,0 +1,34 @@
...
+    """b2luigi is a helper package constructed around luigi that helps you schedule working
+    packages (so-called tasks) locally or on a batch system. Apart from the very
+    powerful dependency management system by luigi, b2luigi extends the user interface
+    and has a built-in support for the queue systems, e.g. LSF and HTCondor."""
+
+    homepage = "https://github.com/belle2/b2luigi"
+    pypi = "b2luigi/b2luigi-1.2.6.tar.gz"
+
+    license("GNU")
+
```
wdconinc, line 20: I know this is not technically HEP-only, but it is definitely a project from within the HEP  community. Adding a tag helps us (HEP-ecosystem maintainers) keep an eye on which of the `hep`-tagged packages can be updated.

```suggestion

    tags = ["hep"]

```

Automatic CI builds can be added to spack CI as well; see https://github.com/spack/spack-packages/blob/develop/stacks/hep/spack.yaml.

CamCoop1: Thank you @wdconinc I had not come across these tags this is good to know

repos/spack_repo/builtin/packages/py_b2luigi/package.py
CamCoop1: Absolutely, I had not realised this change had taken place yet thank you for picking that up

repos/spack_repo/builtin/packages/py_b2luigi/package.py
CamCoop1: My apologies I was under the impression that this was an exclusive upper range. This change is certainly what we want

CamCoop1:
> > Adding b2luigi ([pypi.org/project/b2luigi](https://pypi.org/project/b2luigi/)) beginning at v1.2.6 in which its dependencies were refactored to only include Spack compatible packages.
> 
> I'm not sure what you mean, if a package is not in Spack in can be added

It was my understanding that for a package to be added into the Spack ecosystem, all of its dependencies must themselves be Spack packages. Prior to `b2luigi v1.2.6` there was one package called `retry2` which was no longer being updated by the maintainer and did not have an associated Spack package. So this dependency was changed to `tenacity` in `v1.2.6` which does have a Spack package. 

Am I then correct in my understanding that any versions of b2luigi `<v1.2.6` would not be compatible in Spack and so we should not be include in this PR?

jmcarcell:
> Am I then correct in my understanding that any versions of b2luigi `<v1.2.6` would not be compatible in Spack and so we should not be include in this PR?

Yes but a recipe for `retry2` can always be added.

CamCoop1:
> Yes but a recipe for `retry2` can always be added.
I think from the b2luigi maintainers side, updating to `tenacity` was beneficial as it is actively being developed and has a very similar API to `retry2` and only required minor updates to the code base. Thats not to say there isn't merit in adding recipes for these older packages with no active development, I just think this was an easy and natural change to make for `b2luigi` in light of us adding it into the Spack ecosystem

CamCoop1:
>  Yes but a recipe for `retry2` can always be added.
 I think from the b2luigi maintainers side, updating to `tenacity` was beneficial as it is actively being developed and has a very similar API to `retry2` and only required minor updates to the code base. Thats not to say there isn't merit in adding recipes for these older packages with no active development, I just think this was an easy and natural change to make for `b2luigi` in light of us adding it into the Spack ecosystem

CamCoop1:
> Yes but a recipe for `retry2` can always be added.


I think from the b2luigi maintainers side, updating to `tenacity` was beneficial as it is actively being developed and has a very similar API to `retry2` and only required minor updates to the code base. Thats not to say there isn't merit in adding recipes for these older packages with no active development, I just think this was an easy and natural change to make for `b2luigi` in light of us adding it into the Spack ecosystem

wdconinc (approved):

PR #2827, opened 2025-12-18
clingo-bootstrap: optimize hash table

haampie:
In clingo@spack the hottest code path is `hash_set.hh`. It uses a bit of
a non-standard hash table implementation, where the container has a
capacity that's a prime number. That's probably to make hash collisions
unlikely, but in practice the hash function is fine, and it's more
efficient to use powers of two so that computing the slot is fast.

Also precomputes the threshold under which growing/rehashing is needed.

All in all this reduce the "self" time:

```
Gringo::Symbol::createFun
from 5.4% to 4.2%
```

```
Clasp::Asp::LogicProgram::simplifyNormal
from 3.8% to 3.6%
```


haampie:
Meh, on Linux the improvement is rather modest, and on macOS I don't see any improvement really. Think I'll just close it.

PR #2835, opened 2025-12-18
Python: more changes for free-threaded support

adamjstewart:
Follow-up to #2632 

There are probably other places we should change too, like `site-packages` becomes `pythonX.Yt`, but we can fix things as we discover bugs.

@BaLinuss @wdconinc @rgommers

### DIFF
--- a/repos/spack_repo/builtin/packages/python/package.py
+++ b/repos/spack_repo/builtin/packages/python/package.py
@@ -306,6 +306,14 @@ class Python(Package):
         else:
             variants += "~pythoncmd"
 
+        if Version(version_str) >= Version("3.13"):
+            for exe in exes:
+                if exe.endswith("t"):
+                    variants += "+freethreading"
+                    break
+            else:
+                variants += "~freethreading"
+
         for module in [
             "readline",
             "sqlite3",
@@ -857,12 +865,18 @@ class Python(Package):
         # installed python, several different commands could be located
         # in the same directory. Be as specific as possible. Search for:
         #
-        # * python3.11
+        # * python3.14t
+        # * python3.14
         # * python3
         # * python
         #
-        # in that order if using python@3.11.0, for example.
-        suffixes = [self.spec.version.up_to(2), self.spec.version.up_to(1), ""]
+        # in that order if using python@3.14.0, for example.
+        suffixes = [
+            self.spec.version.up_to(2) + "t",
+            self.spec.version.up_to(2),
+            self.spec.version.up_to(1),
+            "",
+        ]
         ext = "" if sys.platform != "win32" else ".exe"
         filenames = [f"python{ver}{ext}" for ver in suffixes]
         root = self.prefix.bin if sys.platform != "win32" else self.prefix
--- a/repos/spack_repo/builtin/packages/python/package.py
+++ b/repos/spack_repo/builtin/packages/python/package.py
@@ -872,9 +872,9 @@ class Python(Package):
         #
         # in that order if using python@3.14.0, for example.
         suffixes = [
-            self.spec.version.up_to(2) + "t",
-            self.spec.version.up_to(2),
-            self.spec.version.up_to(1),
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/python/package.py
```diff
@@ -306,6 +306,14 @@ def determine_variants(cls, exes, version_str):
         else:
             variants += "~pythoncmd"
 
+        if Version(version_str) >= Version("3.13"):
+            for exe in exes:
+                if exe.endswith("t"):
+                    variants += "+freethreading"
+                    break
+            else:
+                variants += "~freethreading"
```
adamjstewart, line 315: This is not exactly the recommended approach: https://docs.python.org/3/howto/free-threading-python.html#identifying-free-threaded-python

But maybe it works and is simple? Could also call sysconfig if needed.

rgommers: I can't quite see what `exes` contains, but I guess the assumption is that there's only one `python` package installed? There should be a `python3.14t` or some such name for free-threading indeed. In that case: seems reasonable.

If you'd use `sys._is_gil_enabled()` that would have to be a `subprocess` call I guess, since you want it for the target interpreter and not the one executing this `package.py`?

adamjstewart: My understanding is that `exes` is a list of executables that match `python*` and all report the same version. We could call subprocess if needed, we already do that for imports. I'm wary of `sys._is_gil_enabled()` because it's private and not available in most Python versions.

adamjstewart:
Hmm, not sure if I broke something or CI is just broken...

adamjstewart:
Waiting on #2836

repos/spack_repo/builtin/packages/python/package.py
rgommers: Should work then, so 👍🏼 from me.

wdconinc: There's also `sys.abiflags` which contains `t` when free-threading and is not private and has [been around](https://docs.python.org/3/library/sys.html) for a while.

adamjstewart: True, just one more slow subprocess call if we want to check that. I would be in favor of the current solution until we find a counter-example where it doesn't work.

repos/spack_repo/builtin/packages/python/package.py
```diff
@@ -857,12 +865,18 @@ def command(self):
...
+        # * python3.14t
+        # * python3.14
         # * python3
         # * python
         #
-        # in that order if using python@3.11.0, for example.
-        suffixes = [self.spec.version.up_to(2), self.spec.version.up_to(1), ""]
+        # in that order if using python@3.14.0, for example.
+        suffixes = [
+            str(self.spec.version.up_to(2)) + "t",
```
adamjstewart, line 875: Hmm, technically we should ignore `t` unless `+freethreading`? Not sure if we'll ever encounter a situation where both 3.14 and 3.14t are installed in the same directory. 3.14t adds a 3.14 executable as well, so it would be kinda silly to install both to the same location.

repos/spack_repo/builtin/packages/python/package.py
wdconinc: Yeah, it wasn't as much a suggestion to take that approach here necessarily, but I used it in another context where there was already a python call for the version number, so then it was only a marginally cost.

wdconinc: Yeah, it wasn't as much a suggestion to take that approach here necessarily, but I used it in another context where there was already a python call for the version number, so then it was only a marginally cost.

becker33 (approved):

adamjstewart:
https://github.com/kornia/kornia-rs/issues/609

adamjstewart:
@spackbot run pipeline

PR #2841, opened 2025-12-18
Add s5cmd package

ebagrenrut:
Add `s5cmd`, a very fast tool for managing files locally and in S3

### DIFF
--- a/repos/spack_repo/builtin/packages/s5cmd/package.py
+++ b/repos/spack_repo/builtin/packages/s5cmd/package.py
@@ -1,0 +1,42 @@
+# Copyright Spack Project Developers. See COPYRIGHT file for details.
+#
+# SPDX-License-Identifier: (Apache-2.0 OR MIT)
+from spack_repo.builtin.build_systems.go import GoPackage
+
+from spack.package import *
+
+
+class S5cmd(GoPackage):
+    """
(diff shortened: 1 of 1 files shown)


ebagrenrut:
I'm open to suggestions on how to set the `SHELL` environment variable value temporarily while generating the shell completions files. `s5cmd` currently only looks to the value of the `SHELL` variable in the calling environment to determine which type of completion config to output. I'm effectively doing `os.environ["SHELL"] = "bash"`, but I'm concerned that that causes a side-effect that "leaks" out from the context manager and could be modifying the environment of the python process. It occurs to me that I could also investigate the attributes of `Executable` to see if there's a way to define it with a modified environment.

I tried various appropaches to using `EnvironmentModifications().set()`, but I was unable to get them to work. If there is no elegant, safe way to temporarily change/set `SHELL` only within the scope of the `install_completions` function, then I could also omit the completions entirely...

ebagrenrut:
> It occurs to me that I could also investigate the attributes of `Executable` to see if there's a way to define it with a modified environment.

I instantiated `EnvironmentModifications`, set `SHELL` within it, and then passed that as an `extra_env` parameter when calling the `Executable("s5cmd")` instance. Worked well!

becker33 (approved):

PR #2129, opened 2025-10-27
Add: py-codechecker

Chrismarsh:
- Adds the codechecker package, and all dependent packages
- Updates some dependencies
- Somewhat resolves https://github.com/spack/spack-packages/issues/1852 by conflicting against the broken configuration combination

Issue #1852, opened 2025-10-03
thrift issues


Chrismarsh:
1) `build_system=autotools` requires `+openssl`, as `--without-openssl` doesn't work
```
==> Installing thrift-0.21.0-vowrzyk2nowp6ea6avze46ahqc2bcl2l [45/45]
==> Using cached archive: spack/cache/_source-cache/archive/9a/9a24f3eba9a4ca493602226c16d8c228037db3b9291c6fc4019bfe3bd39fc67c.tar.gz
==> No patches needed for thrift
==> thrift: Executing phase: 'autoreconf'
==> thrift: Executing phase: 'configure'
==> Error: ProcessError: Command exited with status 1:
    'spack/stage/spack-stage-thrift-0.21.0-vowrzyk2nowp6ea6avze46ahqc2bcl2l/spack-src/configure' '--prefix=project/spack/opt/linux-icelake/thrift-0.21.0-vowrzyk2nowp6ea6avze46ahqc2bcl2l' '--enable-shared' '--enable-tests=no' '--with-cpp' '--without-libevent' '--without-zlib' '--without-qt5' '--without-c_glib' '--without-openssl' '--without-java' '--without-kotlin' '--without-erlang' '--without-nodejs' '--without-nodets' '--without-lua' '--with-python' '--without-py3' '--without-perl' '--without-php' '--without-php_extension' '--without-dart' '--without-ruby' '--without-go' '--without-swift' '--without-rs' '--without-cl' '--without-haxe' '--without-netstd' '--without-d'

1 error found in build log:
     113    checking whether yytext is a pointer... yes
     114    checking whether ln -s works... yes
     115    checking for gawk... (cached) gawk
     116    checking for ranlib... (cached) ranlib
     117    checking whether project/spack/opt/linux-icelake/compiler-wrapper-1.0-5txabblu24lka57jpjnn2tuycty4ttwy/libexec/spack/gcc/g++ su
            pports C++11 features with -std=c++11... yes
     118    checking for boostlib >= 1.56.0 (105600)... yes
  >> 119    configure: error: Invalid --with-openssl value
```
2) `build_system=cmake` python bindings aren't created
```
linux-icelake/thrift-0.21.0-sedmqaqah4bvmbwyaprzm4r767zfyy2r/lib took 3m52s
❯ ll
total 1.1M
drwxr-sr-x 4 chm003 eccc_ccrp 4.0K Oct  3 23:30 .
[...]

PR #2846, opened 2025-12-19
thrift: make sure the python variant exists for CMake

jmcarcell:
Currently, compilation with CMake will fail because `py-setuptools` is not found. This happens because the python variant doesn't exist with CMake and `WITH_PYTHON` is set to True. Now, with CMake the python variant has to be set to False because of the conflict.

### DIFF
--- a/repos/spack_repo/builtin/packages/thrift/package.py
+++ b/repos/spack_repo/builtin/packages/thrift/package.py
@@ -52,7 +52,6 @@ class Thrift(CMakePackage, AutotoolsPackage):
         "python",
         default=True,
         description="Build support for python",
-        when="build_system=autotools",
     )
 
     build_system("cmake", "autotools", default="autotools")
--- a/repos/spack_repo/builtin/packages/thrift/package.py
+++ b/repos/spack_repo/builtin/packages/thrift/package.py
@@ -48,11 +48,7 @@ class Thrift(CMakePackage, AutotoolsPackage):
     variant("java", default=False, description="Build support for java")
     variant("javascript", default=False, description="Build Javascript library")
     variant("nodejs", default=False, description="Build NodeJS library")
-    variant(
-        "python",
-        default=True,
-        description="Build support for python",
-    )
+    variant("python", default=True, description="Build support for python")
 
     build_system("cmake", "autotools", default="autotools")
 


thomas-bouvier:
Asking for @Chrismarsh's input as this change was introduced by https://github.com/spack/spack-packages/pull/2129.

Chrismarsh:
When building `thrift build_system=cmake`, I found that even though the python was enabled, the bindings were not created. Reported in https://github.com/spack/spack-packages/issues/1852. This would result in a broken import even though it "seemed to work". Thus the change I made to ensure that the broken variant wasn't surfaced until someone with more time could figure out what the problem is. 

It probably needs to be reported upstream https://issues.apache.org/jira/projects/THRIFT/issues/THRIFT-5909?filter=allopenissues if it isn't fixed in a newer release. I found it quite difficult to determine exactly where and what the problem was as the issues are unlinked from their repo.

From what you're saying, I guess 
https://github.com/spack/spack-packages/blob/develop/repos/spack_repo/builtin/packages/thrift/package.py#L137
is being marked as True even though that variant doesn't exist?

I think this PR should be changed to mark that line as
```
self.define("WITH_PYTHON", False),  # renable once cmake +python is fixed
```

jmcarcell:
We already have a conflict that will set the variant to False if not requested: https://github.com/spack/spack-packages/blob/develop/repos/spack_repo/builtin/packages/thrift/package.py#L107. If 

> I think this PR should be changed to mark that line as
> 
> ```
> self.define("WITH_PYTHON", False),  # renable once cmake +python is fixed
> ```

is added, then two places will need to be modified once python bindings work instead of one.

Chrismarsh:
That link is 404 for me.

 When I submitted the PR, the goal was to remove the broken variant. Are you suggesting to re-add the variant it but conflict it on `build_system=cmake`? That seems reasonable.

jmcarcell:
Sorry, the link is fixed now. 

> When I submitted the PR, the goal was to remove the broken variant. Are you suggesting to re-add the variant it but conflict it on `build_system=cmake`? That seems reasonable.

This is exactly what would happen with this PR.

Chrismarsh:
Ah, I see now. That seems good. I am surprised that with the variant gone a) the define from variant worked (for me) and b) it defaults to True causing the issue. Thanks for catching this

thomas-bouvier (approved):
LGTM, thanks for the clarifications :)

tldahlgren (approved):

thomas-bouvier:
Thanks @tldahlgren !!

PR #2849, opened 2025-12-19
vecgeom: add v2.0.0

sethrj:


wdconinc (changes requested):
This has conflicts.

sethrj:
My bad, I think I rebased off my local 'develop' which was way out of date.

sethrj:
@spackbot fix style

wdconinc (approved):

PR #2857, opened 2025-12-19
ccache: add hidden libfmt dependency

sethrj:
As of [ccache 4.10](https://ccache.dev/releasenotes.html#_build_improvements_6) the `FROM_INTERNET` is replaced by a `DEPS` option, and `libfmt` is used from the system if available.
- Tell spack that ccache no longer vendors libfmt but looks externally
- Fix variables to force external rather than builtin dependencies
- Add missing xxhash build/rt dependency

There are a few vendored dependencies that still use built-in, but since ccache is a build tool and not linked against, this is probably ok. The importance of forcing external xxhash/libfmt was to avoid linking to system libraries, not building in potentially duplicate symbols.


sethrj:
@spackbot fix style

sethrj:
Ping @haampie

haampie (approved):

sethrj:
Thanks @haampie !

PR #2365, opened 2025-11-10
Make llvm path consistent between llvm and llvm-amdgpu in enzyme pkg

chapman39:
This PR makes the llvm pathing consistent between llvm and llvm-amdgpu. This way, specifying llvm-amdgpu in the `spack.yaml` will look like the following.

```yaml
    llvm-amdgpu:
      externals:
      - spec: llvm-amdgpu@6.4.2
        prefix: /opt/rocm-6.4.2/llvm
        extra_attributes:
          compilers:
            c: /opt/rocm-6.4.2/llvm/bin/amdclang
            cxx: /opt/rocm-6.4.2/llvm/bin/amdclang++
            fortran: /opt/rocm-6.4.2/llvm/bin/amdflang
```

PR #2558, opened 2025-11-25
Update radiuss packages with latest releases

adrienbernede:
Updates in Camp, Umpire, CHAI, CARE, RAJA, RAJAPerf.

Essentially new releases and associated dependencies constraints.

PR #2858, opened 2025-12-19
Axom: Update ROCm root var

chapman39:
Similar to https://github.com/spack/spack-packages/pull/2365, this PR updates Axom var `rocm_root` to move a previous directory back from `llvm-amdgpu` prefix. The assumption is the prefix is `/opt/rocm-6.4.2/llvm`.

If other's set the `llvm-amdgpu` prefix differently than below, maybe we can handle both cases.

```yaml
    llvm-amdgpu:
      externals:
      - spec: llvm-amdgpu@6.4.2
        prefix: /opt/rocm-6.4.2/llvm
        extra_attributes:
          compilers:
            c: /opt/rocm-6.4.2/llvm/bin/amdclang
            cxx: /opt/rocm-6.4.2/llvm/bin/amdclang++
            fortran: /opt/rocm-6.4.2/llvm/bin/amdflang
```

Also adds unrelated fix to an incorrect version in MFEM from a previous PR of mine https://github.com/spack/spack-packages/pull/2363#discussion_r2562195648

### DIFF
--- a/repos/spack_repo/builtin/packages/axom/package.py
+++ b/repos/spack_repo/builtin/packages/axom/package.py
@@ -374,7 +374,7 @@ class Axom(CachedCMakePackage, CudaPackage, ROCmPackage):
 
             hip_link_flags = ""
 
-            rocm_root = spec["llvm-amdgpu"].prefix
+            rocm_root = os.path.dirname(spec["llvm-amdgpu"].prefix)
             entries.append(cmake_cache_path("ROCM_ROOT_DIR", rocm_root))
 
             # Recommended MPI flags
--- a/repos/spack_repo/builtin/packages/mfem/package.py
+++ b/repos/spack_repo/builtin/packages/mfem/package.py
@@ -767,7 +767,7 @@ class Mfem(Package, CudaPackage, ROCmPackage):
                     hypre_rocm_libs += hypre["rocsparse"].libs
                 if "^rocrand" in hypre:
                     hypre_rocm_libs += hypre["rocrand"].libs
-                if hypre.satisfies("@2.26.0:"):
+                if hypre.satisfies("@2.29.0:"):
                     if "^rocsolver" in hypre:
                         hypre_rocm_libs += hypre["rocsolver"].libs
                     if "^rocblas" in hypre:
--- a/repos/spack_repo/builtin/packages/mfem/package.py
+++ b/repos/spack_repo/builtin/packages/mfem/package.py
@@ -767,7 +767,7 @@ class Mfem(Package, CudaPackage, ROCmPackage):
                     hypre_rocm_libs += hypre["rocsparse"].libs
                 if "^rocrand" in hypre:
                     hypre_rocm_libs += hypre["rocrand"].libs
-                if hypre.satisfies("@2.39.0:"):
+                if hypre.satisfies("@2.29.0:"):
                     if "^rocsolver" in hypre:
                         hypre_rocm_libs += hypre["rocsolver"].libs
                     if "^rocblas" in hypre:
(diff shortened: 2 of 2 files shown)


chapman39:
@adrienbernede First of all, thank you for updating the Axom Spack package here. I'm curious if you need to set `llvm-amdgpu` prefix like `/opt/rocm-6.4.2` for your use-case, or if `/opt/rocm-6.4.2/llvm` works for you.

I haven't tried in awhile but I think other packages failed when I set the `llvm-amdgpu` prefix to just `/opt/rocm-6.4.2`.

Also @bmhan12 @kennyweiss @white238 for awareness

repos/spack_repo/builtin/packages/mfem/package.py
```diff
@@ -767,7 +767,7 @@ def find_optional_library(name, prefix):
                     hypre_rocm_libs += hypre["rocsparse"].libs
                 if "^rocrand" in hypre:
                     hypre_rocm_libs += hypre["rocrand"].libs
-                if hypre.satisfies("@2.39.0:"):
+                if hypre.satisfies("@2.29.0:"):
```
chapman39, line 770:
```suggestion
                if hypre.satisfies("@2.26.0:"):
```

repos/spack_repo/builtin/packages/mfem/package.py
```diff
@@ -767,7 +767,7 @@ def find_optional_library(name, prefix):
                     hypre_rocm_libs += hypre["rocsparse"].libs
                 if "^rocrand" in hypre:
                     hypre_rocm_libs += hypre["rocrand"].libs
-                if hypre.satisfies("@2.39.0:"):
+                if hypre.satisfies("@2.26.0:"):
```
chapman39:
```suggestion
                if hypre.satisfies("@2.29.0:"):
```

adrienbernede:
@chapman39 this is moving in the opposite direction than https://github.com/spack/spack-packages/pull/1402, as discussed in https://github.com/llnl/axom/discussions/1726#discussion-9196039.

It was a conscious change in https://github.com/spack/spack-packages/pull/2558 (see rocm_path in axom package).

I think because this setting is shared in CachedCMakePackage it may not have shown up in your searches, but merging the current PR would in fact result in inconsistent requirements between several packages regarding the `llvm-amdgpu` prefix.

My experiments have told me that Spack was able to locate correctly llvm-amdgpu from the rocm prefix.

@alecbcs initially required that change to be consistent with spack installed rocm software stack. See https://github.com/spack/spack-packages/pull/1402#discussion_r2395179663.

Paradoxically, your change in https://github.com/spack/spack-packages/pull/2365 looks good to me (and to @alecbcs) because it removes the `llvm` suffix.

chapman39:
@adrienbernede okay. thanks for the info. I will re-attempt to build Axom without this change, and set the prefix to `prefix: /opt/rocm-x.y.z` format, and will get back to you.

v-dobrev (approved):

repos/spack_repo/builtin/packages/axom/package.py
```diff
@@ -374,7 +374,7 @@ def initconfig_hardware_entries(self):
 
             hip_link_flags = ""
 
-            rocm_root = spec["llvm-amdgpu"].prefix
+            rocm_root = os.path.dirname(spec["llvm-amdgpu"].prefix)
```
v-dobrev, line 377: Looking at the `hip` package, it is best to just get the `rocm_root` from the `ROCM_PATH` environment variable set by the `hip` package for dependent packages. Actually, I'm not sure that the dependent environment set by the `hip` package can be accessed here just from `self` in this function -- to reuse the correct logic from the `hip` package, it seems like one can call its `get_paths()` function, using something like `pkg["hip"].get_paths()`, and extract the `"rocm-path"` entry from that. Here's a link to the implementation of this function: https://github.com/spack/spack-packages/blob/2ab337204de4b9f034ec5c6380d734f2c1626b28/repos/spack_repo/builtin/packages/hip/package.py#L290

chapman39:
due to reasons explained by @adrienbernede , im going to close this PR. i was able to get Smith TPLs to build without this branch.

PR #2860, opened 2025-12-20
nlohmann-json: patch method for c++20 module support

wdconinc:
This PR adds to `nlohmann-json` two changes from static to inline to allow the latest release to be used in C++20 modules, see https://github.com/nlohmann/json/pull/4764.

### DIFF
--- a/repos/spack_repo/builtin/packages/nlohmann_json/package.py
+++ b/repos/spack_repo/builtin/packages/nlohmann_json/package.py
@@ -58,6 +58,7 @@ class NlohmannJson(CMakePackage):
     def patch(self):
         if self.spec.satisfies("@3.11.0:3.12.0"):
             # Ensure no TU-local entities for C++20 modules
+            # (https://github.com/nlohmann/json/pull/4764)
             for f in [
                 "include/nlohmann/detail/input/binary_reader.hpp",
                 "single_include/nlohmann/json.hpp",
--- a/repos/spack_repo/builtin/packages/nlohmann_json/package.py
+++ b/repos/spack_repo/builtin/packages/nlohmann_json/package.py
@@ -55,6 +55,25 @@ class NlohmannJson(CMakePackage):
     conflicts("%gcc@:4.8", when="@:3.2.9")
     conflicts("%intel@:16")
 
+    def patch(self):
+        if self.spec.satisfies("@3.11.0:3.12.0"):
+            # Ensure no TU-local entities for C++20 modules
+            for f in [
+                "include/nlohmann/detail/input/binary_reader.hpp",
+                "single_include/nlohmann/json.hpp",
+            ]:
+                filter_file(
+                    "static inline bool little_endianness",
+                    "inline bool little_endianness",
+                    f,
+                    string=True,
+                )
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/nlohmann_json/package.py
```diff
@@ -55,6 +55,25 @@ class NlohmannJson(CMakePackage):
     conflicts("%gcc@:4.8", when="@:3.2.9")
     conflicts("%intel@:16")
 
+    def patch(self):
+        if self.spec.satisfies("@3.11.0:3.12.0"):
```
sethrj, line 59: Maybe better to define the patch routine only when it applies:
```suggestion
    @when("@3.11.0:3.12.0")
    def patch(self):
```
or provide an actual patch file for the diff, or pull in the PR you mention?

Is the motivation a wide range of compatibility with nljson versions?

wdconinc: > Maybe better to define the patch routine only when it applies

I have never really looked into how when decorators apply when they are used with overlapping versions ranges but retain the same function name... Doesn't apply here, but that's why I've been placing the satisfies inside the patch methods.

wdconinc: > Is the motivation a wide range of compatibility with nljson versions?

Motivation is only realistically allowing modules with the last version. But this also applies cleanly to older versions.

wdconinc: > or pull in the PR you mention?

Too much crud included.

sethrj (dismissed):

pearzt:
@spackbot re-run pipeline

repos/spack_repo/builtin/packages/nlohmann_json/package.py
```diff
@@ -55,6 +55,25 @@ class NlohmannJson(CMakePackage):
...
+            # Ensure no TU-local entities for C++20 modules
```
pearzt, line 60: I think this makes it easier to understand where the change is coming from:
```suggestion
            # Ensure no TU-local entities for C++20 modules
            # (https://github.com/nlohmann/json/pull/4764)
```

repos/spack_repo/builtin/packages/nlohmann_json/package.py
pearzt: > I have never really looked into how when decorators apply when they are used with overlapping versions ranges but retain the same function name... Doesn't apply here, but that's why I've been placing the satisfies inside the patch methods.

More out of curiosity than to question your choice, could you elaborate on this? Are you placing the `satisfies` inside the patch method just in case there will be another patch in the future?

wdconinc: Yes, exactly. But I just looked and there are a handful of packages (incl `qt`) with multiple `patch` method that only differ in their decorators, so I guess it works.

pearzt (approved):

sethrj (approved):

PR #2861, opened 2025-12-21
root: use is_system_path from envutil

jmcarcell:
Fix a deprecation warning since https://github.com/spack/spack/pull/50995:

```
==> Warning: /spack-packages/repos/spack_repo/builtin/packages/root/package.py:943: spack.package.is_system_path is deprecated

```

### DIFF
--- a/repos/spack_repo/builtin/packages/root/package.py
+++ b/repos/spack_repo/builtin/packages/root/package.py
@@ -8,9 +8,8 @@ import sys
 
 from spack_repo.builtin.build_systems.cmake import CMakePackage
 
-from spack.package import *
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/root/package.py
```diff
@@ -10,6 +10,8 @@
 
 from spack.package import *
 
+import spack.util.environment as envutil
```
wdconinc:
```suggestion
import spack.util.environment as envutil
from spack.package import *
```

wdconinc (approved):

PR #2862, opened 2025-12-21
cudnn: add through v9.17.0 (support CUDA 13)

wdconinc:
This PR adds `cudnn` through vv9.17.0. All versions listed at https://developer.download.nvidia.com/compute/cudnn/redist/cudnn/linux-x86_64/ and https://developer.download.nvidia.com/compute/cudnn/redist/cudnn/linux-sbsa/ are added. This ensures that there are versions for CUDA 13 available.


tldahlgren (approved):
New version checksums are confirmed.

PR #2863, opened 2025-12-21
HEP: set URL fetch method to curl with retry

wdconinc:
This PR explicitly defines a `url_fetch_strategy` for the HEP stack to ensure retry with continuation to get around the flaky `root` download server (and lack of spack source caching) leading to CI errors, e.g. https://gitlab.spack.io/spack/spack-packages/-/jobs/19672806.

### DIFF
--- a/stacks/hep/spack.yaml
+++ b/stacks/hep/spack.yaml
@@ -7,6 +7,10 @@ spack:
   concretizer:
     static_analysis: true
 
+  config:
+    # ROOT fails downloads frequentlysleads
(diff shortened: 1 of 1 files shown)


stacks/hep/spack.yaml
```diff
@@ -7,6 +7,10 @@ spack:
   concretizer:
     static_analysis: true
 
+  config:
+    # ROOT fails downloads frequentlysleads
```
wdconinc:
```suggestion
    # ROOT fails downloads frequently
```

sethrj (approved):

PR #2874, opened 2025-12-23
py-geopandas: add v1.1.2

adamjstewart:
https://github.com/geopandas/geopandas/releases/tag/v1.1.2

Technically security vulnerability is only 0.8+, but I don't think anyone wants those ancient versions anyway.


adamjstewart:
@spackbot fix style

alalazo (approved):

PR #2875, opened 2025-12-23
root: fix downstream dd4hep build when +rpath

sethrj:
When building
```
dd4hep@1.33+ddalign+ddcad+ddcond+dddetectors+dddigi~ddeve+ddg4+ddrec~debug~doc+edm4hep~geant4units~hepmc3~ipo~lcio~tbb~utilityapps~xercesc
^root@6.38.00~aqua~arrow~cuda~cudnn~cxxmodules~daos~davix~dcache~emacs~examples~fftw~fits~fortran+gdml+geom~geombuilder+gminimal~graphviz+gsl~http~ipo+math+minuit~mlp~opengl~pythia8+python~qt6~r+roofit+root7+rpath~shadow~spectrum~sqlite~ssl~tbb+tiff~tmva~tmva-cpu+tmva-cudnn~tmva-gpu~tmva-pymva~tmva-sofie+tpython+unuran~vc+vdt~veccore~webgui~x+xml~xrootd build_system=cmake build_type=Release cxxstd=20 generator=make patches:=22af347 platform=darwin os=tahoe target=m2 %c,cxx=apple-clang@17.0.0
```
the final plugin build fails with:
```
[9/21] Generating libDDCoreGraphics.components
FAILED: [code=1] DDCore/libDDCoreGraphics.components /private/var/folders/n9/mqnx20b929z469f6p3fbq7c40000gn/T/seth/spack-stage/spack-stage-dd4hep-1.33-z7jy2mn65v76gsa5dyg3dlpkevocmuow/spack-build-z7jy2mn/DDCore/libDDCoreGraphics.components
cd /private/var/folders/n9/mqnx20b929z469f6p3fbq7c40000gn/T/seth/spack-stage/spack-stage-dd4hep-1.33-z7jy2mn65v76gsa5dyg3dlpkevocmuow/spack-build-z7jy2mn/lib && DYLD_LIBRARY_PATH=/private/var/folders/n9/mqnx20b929z469f6p3fbq7c40000gn/T/seth/spack-stage/spack-stage-dd4hep-1.33-z7jy2mn65v76gsa5dyg3dlpkevocmuow/spack-build-z7jy2mn/lib:/private/var/folders/n9/mqnx20b929z469f6p3fbq7c40000gn/T/seth/spack-stage/spack-stage-dd4hep-1.33-z7jy2mn65v76gsa5dyg3dlpkevocmuow/spack-build-z7jy2mn/lib: /private/var/folders/n9/mqnx20b929z469f6p3fbq7c40000gn/T/seth/spack-stage/spack-stage-dd4hep-1.33-z7jy2mn65v76gsa5dyg3dlpkevocmuow/spack-build-z7jy2mn/bin/listcomponents_dd4hep -o libDDCoreGraphics.components libDDCoreGraphics.dylib
Error in <UnknownClass::FindDynamicLibrary>: libRIO[.so | .dll | .dylib | .sl | .dl | .a] does not exist in /opt/spack/opt/spack/tahoe/edm4hep/0.99.4/4lpkeqx/lib:/opt/spack/opt/spack/tahoe/podio/1.6/sj3ympx/lib:/private/var/folders/n9/mqnx20b929z469f6p3fbq7c40000gn/T/seth/spack-stage/spack-stage-dd4hep-1.33-z7jy2mn65v76gsa5dyg3dlpkevocmuow/spack-build-z7jy2mn/lib:/private/var/folders/n9/mqnx20b929z469f6p3fbq7c40000gn/T/seth/spack-stage/spack-stage-dd4hep-1.33-z7jy2mn65v76gsa5dyg3dlpkevocmuow/spack-build-z7jy2mn/lib::/opt/spack/opt/spack/tahoe/edm4hep/0.99.4/4lpkeqx/lib:/opt/spack/opt/spack/tahoe/podio/1.6/sj3ympx/lib:/opt/spack/opt/spack/tahoe/podio/1.6/sj3ympx/lib::.:/usr/local/lib:/usr/X11R6/lib:/usr/lib:/lib:/lib/x86_64-linux-gnu:/usr/local/lib64:/usr/lib64:/lib64:
Error in <UnknownClass::FindDynamicLibrary>: libCling[.so | .dll | .dylib | .sl | .dl | .a] does not exist in /opt/spack/opt/spack/tahoe/edm4hep/0.99.4/4lpkeqx/lib:/opt/spack/opt/spack/tahoe/podio/1.6/sj3ympx/lib:/private/var/folders/n9/mqnx20b929z469f6p3fbq7c40000gn/T/seth/spack-stage/spack-stage-dd4hep-1.33-z7jy2mn65v76gsa5dyg3dlpkevocmuow/spack-build-z7jy2mn/lib:/private/var/folders/n9/mqnx20b929z469f6p3fbq7c40000gn/T/seth/spack-stage/spack-stage-dd4hep-1.33-z7jy2mn65v76gsa5dyg3dlpkevocmuow/spack-build-z7jy2mn/lib::/opt/spack/opt/spack/tahoe/edm4hep/0.99.4/4lpkeqx/lib:/opt/spack/opt/spack/tahoe/podio/1.6/sj3ympx/lib:/opt/spack/opt/spack/tahoe/podio/1.6/sj3ympx/lib::.:/usr/local/lib:/usr/X11R6/lib:/usr/lib:/lib:/lib/x86_64-linux-gnu:/usr/local/lib64:/usr/lib64:/lib64:
Fatal in <TROOT::InitInterpreter>: cannot load symbol dlsym(RTLD_DEFAULT, CreateInterpreter): symbol not found
```

This appears to be because the ROOT lib path needs to be explicitly defined in `ROOT_LIBRARY_PATH`, even for rpath builds. See https://github.com/spack/spack/pull/16899, https://github.com/spack/spack/pull/45109 .

Since Spack no longer builds anything older than ROOT 6.28, I've also eliminated the `root_library_path` property.

### DIFF
--- a/repos/spack_repo/builtin/packages/dd4hep/package.py
+++ b/repos/spack_repo/builtin/packages/dd4hep/package.py
@@ -230,8 +230,10 @@ class Dd4hep(CMakePackage):
         env.set("DD4HEP", self.prefix.examples)
         env.set("DD4hep_DIR", self.prefix)
         env.set("DD4hep_ROOT", self.prefix)
-        if len(self.libs.directories) > 0:
-            env.prepend_path("LD_LIBRARY_PATH", self.libs.directories[0])
+
+    def setup_dependent_run_environment(self, env: EnvironmentModifications) -> None:
+        # Make plugins discoverable to ROOT
+        env.prepend_path("ROOT_LIBRARY_PATH", self.libs.directories[0])
 
     def url_for_version(self, version):
         # dd4hep releases are dashes and padded with a leading zero
--- a/repos/spack_repo/builtin/packages/dd4hep/package.py
+++ b/repos/spack_repo/builtin/packages/dd4hep/package.py
@@ -231,10 +231,6 @@ class Dd4hep(CMakePackage):
         env.set("DD4hep_DIR", self.prefix)
         env.set("DD4hep_ROOT", self.prefix)
 
-    def setup_dependent_run_environment(self, env: EnvironmentModifications) -> None:
-        # Make plugins discoverable to ROOT
-        env.prepend_path("ROOT_LIBRARY_PATH", self.libs.directories[0])
-
     def url_for_version(self, version):
         # dd4hep releases are dashes and padded with a leading zero
         # the patch version is omitted when 0
--- a/repos/spack_repo/builtin/packages/dd4hep/package.py
+++ b/repos/spack_repo/builtin/packages/dd4hep/package.py
@@ -230,6 +230,8 @@ class Dd4hep(CMakePackage):
         env.set("DD4HEP", self.prefix.examples)
         env.set("DD4hep_DIR", self.prefix)
         env.set("DD4hep_ROOT", self.prefix)
+        if len(self.libs.directories) > 0:
+            env.prepend_path("LD_LIBRARY_PATH", self.libs.directories[0])
 
     def url_for_version(self, version):
         # dd4hep releases are dashes and padded with a leading zero
--- a/repos/spack_repo/builtin/packages/root/package.py
+++ b/repos/spack_repo/builtin/packages/root/package.py
@@ -609,7 +609,8 @@ class Root(CMakePackage):
         _add_variant(v, f, "roofit", "+roofit")
         # webui feature renamed to webgui in 6.18
         _add_variant(v, f, ("root7", "webgui"), "+webgui")
-        _add_variant(v, f, "rpath", "+rpath")
+        if Version(version_str) < Version("6.38"):
+            _add_variant(v, f, "rpath", "+rpath")
         _add_variant(v, f, "runtime_cxxmodules", "+cxxmodules")
         _add_variant(v, f, "shadowpw", "+shadow")
         _add_variant(v, f, "spectrum", "+spectrum")
@@ -928,10 +929,6 @@ class Root(CMakePackage):
         # https://github.com/root-project/root/issues/18949
         if "+cxxmodules +vc" in self.spec:
             env.prepend_path("ROOT_INCLUDE_PATH", self.spec["vc"].prefix.include)
-
-        # Also set up include and library paths since downstream packages may
-        # require libCling etc.
-        self.setup_dependent_run_environment(env, dependent_spec)
 
     def setup_dependent_run_environment(self, env: EnvironmentModifications, dependent_spec):
         env.prepend_path("ROOT_INCLUDE_PATH", dependent_spec.prefix.include)
--- a/repos/spack_repo/builtin/packages/root/package.py
+++ b/repos/spack_repo/builtin/packages/root/package.py
@@ -895,6 +895,12 @@ class Root(CMakePackage):
             # warnings when building against ROOT
             env.unset("MACOSX_DEPLOYMENT_TARGET")
 
+    @property
+    def root_library_path(self):
+        # This used to be version-dependent, but such old versions are no
+        # longer supported by Spack
+        return "ROOT_LIBRARY_PATH"
+
     def setup_run_environment(self, env: EnvironmentModifications) -> None:
         env.set("ROOTSYS", self.prefix)
         env.set("ROOT_VERSION", "v{0}".format(self.version.up_to(1)))
@@ -903,10 +909,10 @@ class Root(CMakePackage):
         env.set("CLING_STANDARD_PCH", "none")
         env.set("CPPYY_API_PATH", "none")
         env.set("CPPYY_BACKEND_LIBRARY", self.prefix.lib.root.libcppyy_backend)
-        env.prepend_path("ROOT_LIBRARY_PATH", self.prefix.lib.root)
+        env.prepend_path(self.root_library_path, self.prefix.lib.root)
 
         # https://github.com/root-project/root/issues/18949
-        if "+cxxmodules" in self.spec and "+vc" in self.spec:
+        if "+cxxmodules +vc" in self.spec:
             env.prepend_path("ROOT_INCLUDE_PATH", self.spec["vc"].prefix.include)
 
     def setup_dependent_build_environment(self, env: EnvironmentModifications, dependent_spec):
@@ -936,4 +942,4 @@ class Root(CMakePackage):
         # versions).
         for lib_path in [dependent_spec.prefix.lib, dependent_spec.prefix.lib64]:
             if os.path.exists(lib_path):
-                env.prepend_path("ROOT_LIBRARY_PATH", lib_path)
+                env.prepend_path(self.root_library_path, lib_path)
--- a/repos/spack_repo/builtin/packages/root/package.py
+++ b/repos/spack_repo/builtin/packages/root/package.py
@@ -898,8 +898,9 @@ class Root(CMakePackage):
 
     @property
     def root_library_path(self):
-        # This used to be version-dependent, but such old versions are no
-        # longer supported by Spack
+        # The ROOT_LIBRARY_PATH environment variable was added to ROOT 6.26.
+        # For previous versions (no longer supported by Spack) it was
+        # LD_LIBRARY_PATH.
         return "ROOT_LIBRARY_PATH"
 
     def setup_run_environment(self, env: EnvironmentModifications) -> None:
@@ -926,6 +927,9 @@ class Root(CMakePackage):
         if "platform=darwin" in self.spec:
             # Newer deployment targets cause fatal errors in rootcling
             env.unset("MACOSX_DEPLOYMENT_TARGET")
+
+        # Note that setup_dependent_run_environment will add include and
+        # library path
 
         # https://github.com/root-project/root/issues/18949
         if "+cxxmodules" in self.spec and "+vc" in self.spec:
--- a/repos/spack_repo/builtin/packages/root/package.py
+++ b/repos/spack_repo/builtin/packages/root/package.py
@@ -910,6 +910,7 @@ class Root(CMakePackage):
         env.set("CLING_STANDARD_PCH", "none")
         env.set("CPPYY_API_PATH", "none")
         env.set("CPPYY_BACKEND_LIBRARY", self.prefix.lib.root.libcppyy_backend)
+        # Always define ROOT's library path for downstream usage
         env.prepend_path(self.root_library_path, self.prefix.lib.root)
 
         # https://github.com/root-project/root/issues/18949
@@ -931,7 +932,9 @@ class Root(CMakePackage):
             env.prepend_path("ROOT_INCLUDE_PATH", self.spec["vc"].prefix.include)
 
     def setup_dependent_run_environment(self, env: EnvironmentModifications, dependent_spec):
+        # Set up runtime dependencies *of downstream packages* that use ROOT
         env.prepend_path("ROOT_INCLUDE_PATH", dependent_spec.prefix.include)
+
         # For dependents that build dictionaries, ROOT needs to know where the
         # dictionaries have been installed.  This can be facilitated by
         # automatically prepending dependent package library paths to
--- a/repos/spack_repo/builtin/packages/root/package.py
+++ b/repos/spack_repo/builtin/packages/root/package.py
@@ -914,7 +914,7 @@ class Root(CMakePackage):
         env.prepend_path(self.root_library_path, self.prefix.lib.root)
 
         # https://github.com/root-project/root/issues/18949
-        if "+cxxmodules +vc" in self.spec:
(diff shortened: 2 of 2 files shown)


knoepfel (dismissed):
I just ran into this error this morning.  Thanks for the quick fix.

repos/spack_repo/builtin/packages/root/package.py
```diff
@@ -895,11 +895,6 @@ def add_include_path(dep_name):
             # warnings when building against ROOT
             env.unset("MACOSX_DEPLOYMENT_TARGET")
 
-    @property
-    def root_library_path(self):
```
wdconinc, line 899: This may be relied upon in downstream package repos. May need deprecation somehow?

FYI @jmcarcell

repos/spack_repo/builtin/packages/root/package.py
```diff
@@ -921,17 +915,18 @@ def setup_dependent_build_environment(self, env: EnvironmentModifications, depen
...
             env.unset("MACOSX_DEPLOYMENT_TARGET")
 
         # https://github.com/root-project/root/issues/18949
-        if "+cxxmodules" in self.spec and "+vc" in self.spec:
+        if "+cxxmodules +vc" in self.spec:
             env.prepend_path("ROOT_INCLUDE_PATH", self.spec["vc"].prefix.include)
 
+        # Also set up include and library paths since downstream packages may
+        # require libCling etc.
+        self.setup_dependent_run_environment(env, dependent_spec)
```
wdconinc: This is already called by spack and doesn't need to be called explicitly, right?

https://github.com/spack/spack/blob/f799ac9c71dbcb3f5bbb14613098be77cb204537/lib/spack/spack/build_environment.py#L1070-L1074

https://github.com/spack/spack/blob/f799ac9c71dbcb3f5bbb14613098be77cb204537/lib/spack/spack/modules/common.py#L674-L677

repos/spack_repo/builtin/packages/root/package.py
```diff
@@ -921,17 +915,18 @@ def setup_dependent_build_environment(self, env: EnvironmentModifications, depen
         env.prepend_path("PYTHONPATH", self.prefix.lib.root)
         env.prepend_path("PATH", self.prefix.bin)
         env.append_path("CMAKE_MODULE_PATH", self.prefix.cmake)
-        env.prepend_path("ROOT_INCLUDE_PATH", dependent_spec.prefix.include)
```
wdconinc, line 924: Is this removal intentional?

repos/spack_repo/builtin/packages/root/package.py
sethrj: Right, outside the main spack repo. Will restore.

repos/spack_repo/builtin/packages/root/package.py
sethrj: After starting at the source a bit, I don't think so: if ROOT is set as a build-time (but not run-time) dependency, then `setup_dependent_run_environment` won't be called. So if for some reason you called `rootcling` as part of the build process but didn't need ROOT at runtime, then you definitely need this.

Regardless, it replicates the same behavior as before, except that it checks lib64 as well (but maybe that's not needed anymore since the previous `lib` worked)?

repos/spack_repo/builtin/packages/root/package.py
sethrj: Yes, it's replaced by the call to `setup_dependent_run_environment`

repos/spack_repo/builtin/packages/root/package.py
sethrj: Ah, I forgot about other downstream repos. I'll restore it.

repos/spack_repo/builtin/packages/root/package.py
wdconinc: But if you only declare ROOT as a `build` dependency yet you link against its libraries, isn't that a downstream bug?

sethrj: If a downstream package has a `"build"` dependency on ROOT (by using the `rootcling` executable during its cmake), it shouldn't need `link` or `run` dependency on ROOT. In practice, this would only happen if your rootcling usage were temporary rather than creating a ROOT plugin.

sethrj: I removed the call, and `spack load root` works as does the dd4hep build.

repos/spack_repo/builtin/packages/dd4hep/package.py
```diff
@@ -230,8 +230,6 @@ def setup_run_environment(self, env: EnvironmentModifications) -> None:
         env.set("DD4HEP", self.prefix.examples)
         env.set("DD4hep_DIR", self.prefix)
         env.set("DD4hep_ROOT", self.prefix)
-        if len(self.libs.directories) > 0:
-            env.prepend_path("LD_LIBRARY_PATH", self.libs.directories[0])
```
jmcarcell, line 234: Removing this is not correct. The plugin mechanism that DD4hep uses relies on `LD_LIBRARY_PATH` and unless it is being set by something else many things in DD4hep won't work.

repos/spack_repo/builtin/packages/root/package.py
```diff
@@ -908,11 +910,11 @@ def setup_run_environment(self, env: EnvironmentModifications) -> None:
...
         env.set("CPPYY_API_PATH", "none")
         env.set("CPPYY_BACKEND_LIBRARY", self.prefix.lib.root.libcppyy_backend)
-        if "+rpath" not in self.spec:
-            env.prepend_path(self.root_library_path, self.prefix.lib.root)
+        # Always define ROOT's library path for downstream usage
+        env.prepend_path(self.root_library_path, self.prefix.lib.root)
 
         # https://github.com/root-project/root/issues/18949
-        if "+cxxmodules" in self.spec and "+vc" in self.spec:
+        if "+cxxmodules +vc" in self.spec:
```
jmcarcell:
```suggestion
        if "+cxxmodules+vc" in self.spec:
```
Does self.spec have spaces? If I look at similar patterns in other packages there is no space so this would be False.

repos/spack_repo/builtin/packages/root/package.py
```diff
@@ -921,19 +923,18 @@ def setup_dependent_build_environment(self, env: EnvironmentModifications, depen
...
-        env.prepend_path("ROOT_INCLUDE_PATH", dependent_spec.prefix.include)
-        if "+rpath" not in self.spec:
-            env.prepend_path(self.root_library_path, self.prefix.lib.root)
         if "platform=darwin" in self.spec:
             # Newer deployment targets cause fatal errors in rootcling
             env.unset("MACOSX_DEPLOYMENT_TARGET")
 
         # https://github.com/root-project/root/issues/18949
-        if "+cxxmodules" in self.spec and "+vc" in self.spec:
+        if "+cxxmodules +vc" in self.spec:
```
jmcarcell:
```suggestion
        if "+cxxmodules+vc" in self.spec:
```
Same as above

jmcarcell:
I don't understand well what's going on here. First of all, I have the following builds with ROOT 6.38 with `+rpath` and DD4hep (master, but nothing related to these issues has changed) that work fine without the changes in this PR, on Alma 9 and Ubuntu 24.04. For example, for Alma 9:
```
^dd4hep@de2efeb1c5a37e9eb32928cdb23f58f8a31a6f73=develop+ddalign+ddcad+ddcond+dddetectors+dddigi+ddeve+ddg4+ddrec~debug~doc+edm4hep~geant4units+hepmc3~hepmc3-gz+ipo+lcio~tbb+utilityapps+xercesc build_system=cmake build_type=Release commit=de2efeb1c5a37e9eb32928cdb23f58f8a31a6f73 cxxstd=20 generator=ninja arch=linux-almalinux9-x86_64 %c,cxx=gcc@14.2.0
^root@6.38.00~aqua~arrow~cuda~cudnn+cxxmodules~daos+davix~dcache~emacs+examples+fftw~fits~fortran+gdml+geom~geombuilder+gminimal~graphviz+gsl+http+ipo+math+minuit+mlp+opengl~pythia8+python~qt6+r+roofit+root7+rpath~shadow+spectrum+sqlite+ssl+tbb+tiff+tmva+tmva-cpu+tmva-cudnn~tmva-gpu~tmva-pymva~tmva-sofie+tpython+unuran+vc+vdt~veccore+webgui+x+xml+xrootd build_system=cmake build_type=Release cxxstd=20 generator=make patches:=22af347,97ffa00 arch=linux-almalinux9-x86_64 %c,cxx=gcc@14.2.0
```

Then, if you have enabled `+rpath`, this PR doesn't change anything (removing `LD_LIBRARY_PATH` for DD4hep at runtime doesn't change anything at build time) since the change you are proposing is to do what is done now through `+rpath` always. Or at least I don't see what is different that affects the build.
Other than that, always having the effects of `+rpath` is good, probably most people build with `+rpath` anyway.

repos/spack_repo/builtin/packages/root/package.py
sethrj: `spec` is `Spec` object rather than a string. I think the user of `s in self.spec` is actually the less preferred syntactic sugar for  spec.satisfies(s)`. I usually see spaces between specs because that eliminates potential ambiguities for variants that have `-` in their names (i.e., `+debug-doc` could be enabling debug and disabling docs, or it could be enabling a special variant for debugging documentation. `+debug -doc`, `-doc+debug` and `+debug~doc` are all equivalent.)

repos/spack_repo/builtin/packages/dd4hep/package.py
sethrj: Does DD4hep specifically rely on `LD_LIBRARY_PATH`, even on mac? Does it not rely on ROOT's plugin loading, which uses `ROOT_LIBRARY_PATH`?

Due to `root` being a dependency of `dd4hep`, the `Root.setup_dependent_run_environment` function in `root/package.py` will actually add `(dd4hep install).lib` to the `ROOT_LIBRARY_PATH`. (And before ROOT 6.26, it would add it to `LD_LIBRARY_PATH` instead.) The "dependent run environment" is very subtle.

wdconinc: > Removing this is not correct. The plugin mechanism that DD4hep uses relies on `LD_LIBRARY_PATH` and unless it is being set by something else many things in DD4hep won't work.

If this is a DD4hep requirement, then let DD4hep set it. Spack tries its very best to avoid the need for LD_LIBRARY_PATH by using RPATH, and having some base-level package like ROOT set it for basically every HEP stack is really annoying.

sethrj:
> I don't understand well what's going on here. First of all, I have the following builds with ROOT 6.38 with `+rpath` and DD4hep (master, but nothing related to these issues has changed) that work fine without the changes in this PR, on Alma 9 and Ubuntu 24.04. 

Hi @jmcarcell, related to https://github.com/AIDASoft/DD4hep/pull/1545 , this might be due to assumptions about `X_LIBRARY_PATH` being set across the systems. When you build and run on linux do you have those already set?

repos/spack_repo/builtin/packages/root/package.py
sethrj: This was a best practices change that I've reverted since it's irrelevant to the main goals of this PR.

repos/spack_repo/builtin/packages/dd4hep/package.py
sethrj: The problem is that these systems are so entangled that it's hard to know what part of the stack depends on what envrionment variables: everyone just sets `LD_LIBRARY_PATH`, `DYLD_LIBRARY_PATH`, and `ROOT_LIBRARY_PATH` in their environments and forgets that this has always been a hack.

I see in DD4HEP's source code:
```c++
#if defined( _WIN32 )
          const char* envVar = "PATH";
          const char  sep    = ';';
#elif defined( __APPLE__ )
          const char* envVar = "DYLD_LIBRARY_PATH";
          const char  sep    = ':';
#else
          const char* envVar = "LD_LIBRARY_PATH";
          const char  sep    = ':';
#endif
          char* search_path = ::getenv( envVar );
          logger().debug( std::string( "searching factories in " ) + envVar );
          std::string path = search_path ? std::string(search_path) : "/usr/lib64:/usr/lib:/usr/local/lib";
```
so I suppose this was only correct for the linux case to begin with? I can make that fix a separate PR.

sethrj:
@knoepfel Could you reapprove? I've distilled the changeset to a minimal version. Thanks!

knoepfel (approved):

PR #2876, opened 2025-12-23
py-numpy: add v2.4.0, v2.4.1

adamjstewart:
* https://github.com/numpy/numpy/releases/tag/v2.4.0
* https://github.com/numpy/numpy/releases/tag/v2.4.1

### DIFF
--- a/repos/spack_repo/builtin/packages/py_numpy/package.py
+++ b/repos/spack_repo/builtin/packages/py_numpy/package.py
@@ -23,6 +23,7 @@ class PyNumpy(PythonPackage):
     license("BSD-3-Clause")
 
     version("main", branch="main")
+    version("2.4.0", sha256="6e504f7b16118198f138ef31ba24d985b124c2c469fe8467007cf30fd992f934")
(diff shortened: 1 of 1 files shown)


rgommers:
Note that NumPy 2.4.0 will probably get yanked and a 2.4.1 will come out shortly, due to the `numpy.random` Cython API being broken (xref https://github.com/scipy/scipy/issues/24215). Not that many users of it, but it'll at least fail CI for SciPy. So may want to either skip 2.4.0 completely, or otherwise merge it together with 2.4.1 in a few days so that unpinned deps will pull in 2.4.1.

rgommers:
2.4.1 is out now with a fix, so this can be updated. I'd lean towards skipping 2.4.0 in Spack, but either way works - very few people should be getting 2.4.0 if 2.4.1 is also available.

adamjstewart:
Added 2.4.1. Let's keep 2.4.0, we didn't see any issues in CI with this. Possibly because we don't allow mixing build-time and run-time versions of deps and build everything from source.

rgommers (approved):
LGTM, thanks @adamjstewart

becker33 (approved):

PR #2877, opened 2025-12-23
py_ovito: New package

LydDeb:
Add py_ovito and its dependencies:
* py_pyside6
* py_pyside6_addons
* py_pyside6_essentials
* py_shiboken6

Introduction of the 'versions.json' file to manage wheel URLs based on platform-architecture-cpython combinations


repos/spack_repo/builtin/packages/py_ovito/package.py
```diff
@@ -0,0 +1,57 @@
...
+import json
+import platform
+from pathlib import Path
+
+from spack_repo.builtin.build_systems.python import PythonPackage
+
+from spack.package import *
+
+_dirname = Path(__file__).resolve().parent
+_versions = json.load(open(join_path(_dirname, "versions.json"), "r"))
```
wdconinc, line 15: Can you explain what your motivation is for the json version loading here? It is not a canonical approach at all and should be avoided.

wdconinc (changes requested):

repos/spack_repo/builtin/packages/py_ovito/package.py
LydDeb: With Ovito and its PySide6 dependencies, the following elements had to be considered when defining the source code download URL:
* operating system
* architecture
* Python compatibility

All of this significantly complicated version declaration.
I drew inspiration from the `openjdk` recipe to build a version dictionary and select versions based on the elements listed above.
Not wanting to include this dictionary directly in the recipe, I wrote it in a separate JSON file.

wdconinc: You are downloading the wheels which does indeed multiply the combinatorics. We don't typically do that in spack. For pyside6 the source can be downloaded from code.qt.io (but it seems that ovito may not distribute any source at all). For pyside6 I'd encourage starting from the source.

Even when only binary wheels are available, we don't want to fragment the dicts from the package.py file since it means package files have external dependencies (json files) that aren't manifest or tracked (e.g. checking mtime on a package.py file isn't sufficient to determine if a version was added).

bernhardkaindl: @LydDeb, you need to get approval by Wouter before this PR can be merged, so you should review this review comment by Wouter. I'm converting this PR to draft status until the review comment is resolved to indicate more strongly that it needs work and isn't ready for further review until that's done.

LydDeb: That is understood; indeed, this review entails additional work in writing recipes, as well as an rework of what has already been produced.

bernhardkaindl:
@LydDeb, PySide6 has been merged with #4884, so you can use it for py-ovito now!

PR #2879, opened 2025-12-23
Add standalone distribution to nextflow

ebagrenrut:
- `java` dependencies are now constrained by Nextflow version. I consulted the `build.gradle` for each version tag to determine the minimum `java` for each `version`. It's possible that there are also upper bounds on the supported `java` version, but I did not investigate that (and it's not clear to me how that would be determined).
- An `Nextflow.nf_versions` dictionary now defines distributions by version. In the future, more attributes could be added per version or per distribution per version, if ever desired.
- Nextflow "edge" versions can now also be supported by adding them to the dictionary with `-edge` appended.
- The user selects which distribution they want by the following logic:
    - If they want the release distribution of a stable version, they simply `spack install nextflow@<version>`, e.g. `spack install nextflow@25.10.2`. `spack install nextflow` installs the release distribution of the most recent stable version.
    - If they want the standalone distribution of a stable version, they `spack install nextflow@<version>-standalone`, e.g. `spack install nextflow@25.10.2-standalone`
    - If they want the release distribution of an edge version, should any edge versions be added to the package in the future, they `spack install nextflow@<edge_version>`, e.g. `spack install nextflow@25.12.0-edge`
    - If they want the standalone distribution of an edge version, they `spack install nextflow@<edge_version>-standalone`, e.g. `spack install nextflow@25.12.0-edge-standalone`
- Though I investigated using a `standalone` variant to constrain the versions supported by the package, it is not currently possible to use variants in the context of `version` objects. Manipulating the versions as noted above grants a similar behavior, just without a variant.

@dialvarezs @marcodelapierre

### DIFF
--- a/repos/spack_repo/builtin/packages/nextflow/package.py
+++ b/repos/spack_repo/builtin/packages/nextflow/package.py
@@ -262,7 +262,7 @@ class Nextflow(Package):
         is_deprecated = False
 
         if "deprecated" in attrs:
-            is_deprecated = True
+            is_deprecated = attrs["deprecated"]
 
         # Edge release. Not preferred.
         if "edge" in ver:
--- a/repos/spack_repo/builtin/packages/nextflow/package.py
+++ b/repos/spack_repo/builtin/packages/nextflow/package.py
@@ -301,7 +301,11 @@ class Nextflow(Package):
         ver = str(version).split("-standalone")
         uri = f"https://github.com/nextflow-io/nextflow/releases/download/v{ver[0]}"
         if "standalone" in str(version):
-            return f"{uri}/nextflow-{ver[0]}-dist"
+            # standalone binary name changed with 24.07.0-edge
+            if version < Version("24.07.0"):
+                return f"{uri}/nextflow-{ver[0]}-all"
+            else:
+                return f"{uri}/nextflow-{ver[0]}-dist"
         else:
             return f"{uri}/nextflow"
 
--- a/repos/spack_repo/builtin/packages/nextflow/package.py
+++ b/repos/spack_repo/builtin/packages/nextflow/package.py
@@ -6,142 +6,301 @@ from spack_repo.builtin.build_systems.generic import Package
 
 from spack.package import *
 
+# nextflow distributions are available at
+# https://github.com/nextflow-io/nextflow/releases
+#
+# The "release" distribution of the nextflow binary contains no bundled plugins and is
+# 'nextflow' in the project's release assets. The "standalone" distribution bundles all
+# first-party plugins and is 'nextflow-<version>-dist' in the project's release assets.
+# Only one or the other will be installed, as dictated by the `standalone` variant. At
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/nextflow/package.py
```diff
@@ -11,137 +11,297 @@ class Nextflow(Package):
...
+        if "deprecated" in attrs:
+            is_deprecated = True
+
+        # Edge release. Not preferred.
+        if "edge" in ver:
+            version(
+                ver,
+                sha256=attrs["release"]["sha256"],
+                expand=False,
+                deprecated=is_deprecated,
```
kwryankrattiger, line 274: Make the not preferred explicit. I don't think the default will ever change, but it is makes it clearer compared to the non-edge version.

```suggestion
                preferred=False,
                deprecated=is_deprecated,
```

kwryankrattiger (commented):
LGTM, one nit

repos/spack_repo/builtin/packages/nextflow/package.py
```diff
@@ -11,137 +11,303 @@ class Nextflow(Package):
...
+        ver = str(version).split("-standalone")
+        uri = f"https://github.com/nextflow-io/nextflow/releases/download/v{ver[0]}"
+        if "standalone" in str(version):
+            # standalone binary name changed with 24.07.0-edge
+            if version < Version("24.07.0"):
+                return f"{uri}/nextflow-{ver[0]}-all"
+            else:
+                return f"{uri}/nextflow-{ver[0]}-dist"
+        else:
+            return f"{uri}/nextflow"
```
dialvarezs, line 311:
```suggestion
    def url_for_version(self, version):
        ver = str(version).replace("-standalone", "")
        uri = f"https://github.com/nextflow-io/nextflow/releases/download/v{ver}"
        if "standalone" in str(version):
            # standalone binary name changed with 24.07.0-edge
            if version < Version("24.07.0"):
                return f"{uri}/nextflow-{ver}-all"
            else:
                return f"{uri}/nextflow-{ver}-dist"
        else:
            return f"{uri}/nextflow"
```

dialvarezs (approved):
Hi @ebagrenrut, great job!

I have a minor suggestion to simplify the code.

kwryankrattiger (approved):
LGTM

marcodelapierre (commented):
Have you considered implementing the `standalone` distribution as a Spack variant? I think it might be a more Spack compliant syntax.
E.g. with a variant: `spack install nextflow@25.10.0 +standalone`

marcodelapierre:
> Have you considered implementing the `standalone` distribution as a Spack variant? I think it might be a more Spack compliant syntax. E.g. with a variant: `spack install nextflow@25.10.0 +standalone`

Never mind, I only noticed now your comment on variants.

Good to go for me then.

ebagrenrut:
> LGTM

@kwryankrattiger Please merge this when you get the chance. Thanks!

PR #2880, opened 2025-12-23
strumpack: respect zfp variant

tbennun:
Before this PR, `strumpack~zfp` would still build with ZFP if found.


white238 (approved):
This works around an issue we were hitting in Smith that cause a failure due to strumpack finding zfp in a python install. This at least allows us to turn off the auto-finding zfp but there is still an issue that this package does not send in any specific prefixes and relies on CMake's auto-finding all TPLs of Strumpack. This is out of the scope of this PR and needs to be fixed in Strumpack before here.

For example, this should be setting `ZFP_ROOT` not appending to the `CMAKE_PREFIX_PATH` to allow you to fully control the correct use of ZFP:

https://github.com/pghysels/STRUMPACK/blob/cfba574f60b792433ee87e7e2a874eeb96c74d1f/CMakeLists.txt#L485-L492

repos/spack_repo/builtin/packages/strumpack/package.py
```diff
@@ -133,10 +133,12 @@ def cmake_args(self):
             self.define_from_variant("STRUMPACK_USE_OPENMP", "openmp"),
             self.define_from_variant("STRUMPACK_USE_CUDA", "cuda"),
             self.define_from_variant("STRUMPACK_USE_HIP", "rocm"),
+            self.define_from_variant("STRUMPACK_USE_ZFP", "zfp"),
```
pghysels: This shouldn't be necessary?

pghysels:
> For example, this should be setting `ZFP_ROOT` not appending to the `CMAKE_PREFIX_PATH` to allow you to fully control the correct use of ZFP:
> 
> https://github.com/pghysels/STRUMPACK/blob/cfba574f60b792433ee87e7e2a874eeb96c74d1f/CMakeLists.txt#L485-L492

I believe `CMAKE_PREFIX_PATH` is set here because it is used in `STRUMPACK/cmake/strumpack-config.cmake.in`

pghysels (approved):

repos/spack_repo/builtin/packages/strumpack/package.py
white238: It is there because the `STRUMPACK_USE_ZFP` option is on by default in the strumpack's build system. This is the fix for the issue we ran into where it was finding a zfp inside a python install that was not wanted and caused a failure through Strumpack -> petsc -> slepc in Smith's TPL build. We had `strumpack~zfp` but it found one anyway.

pghysels: I mean that setting `TPL_ENABLE_ZFP` should be enough?

tbennun: Thanks. I will double check that it works without the second config line.

pghysels (approved):

repos/spack_repo/builtin/packages/strumpack/package.py
white238: ah you are probably correct @pghysels . i missed there were both.

alalazo (dismissed):

repos/spack_repo/builtin/packages/strumpack/package.py
alalazo: Let me know if the discussion is resolved here, and if this PR is good to be merged. Thanks!

white238 (approved):

repos/spack_repo/builtin/packages/strumpack/package.py
tbennun: @pghysels it works without this line as well, removed. Thanks again!

@alalazo ready to merge on my end.

alalazo (approved):

PR #2883, opened 2025-12-23
psblas: added new PSBLAS version (v3.9.0)

Cirdans-Home:
This PR adds the new release of PSBLAS (v3.9.0) to the spack package, it also fixes the name of the test folder.


repos/spack_repo/builtin/packages/psblas/package.py
```diff
@@ -67,14 +70,14 @@ class Psblas(AutotoolsPackage):
         "openmp",
         default=False,
         description="Activates OpenMP support",
-        when="@development,3.9.0-rc1",
+        when="@development,3.9.0-rc1,3.9.0-rc3,3.9.0",
```
haampie: If you use `version("develop"` instead of `version("development"` above, then you can simplify these when conditions to:


```suggestion
        when="@3.9.0-rc1:",
```

`develop` is an infinity version, `development` is not a "special" string and not included in that range otherwise.

repos/spack_repo/builtin/packages/psblas/package.py
```diff
@@ -15,20 +16,22 @@ class Psblas(AutotoolsPackage):
...
 
     version("development", branch="development")
     version(
-        "3.9.0-rc1",
-        sha256="7a7091ce52582b6fc442e8793e36461be36c0947272ea803ad72736ec2d56da8",
+        "3.9.0",
+        sha256="ce523ed2d266fb6dc8723c45a2bcbf40a1c67b7ca61cd9408032a73af7cf8605",
         preferred=True,
     )
+    version("3.9.0-rc3", sha256="7744b3a2cb5d2d3b90e740d905590ee9b8d69c02148dce10ee1b8dc7667e53b1")
```
haampie, line 32: Is it necessary to add more RC candidates if you also add 3.9.0 final?

repos/spack_repo/builtin/packages/psblas/package.py
Cirdans-Home: Thank you, I should have fixed this, it also happened in another place so i fixed it by commit, and I did a couple of style adjustments consequently.

repos/spack_repo/builtin/packages/psblas/package.py
Cirdans-Home: We have some users who have written interfaces with that rc3 version and are behind on updating them (they would be broken with the release). I'd just keep it for convenience until they confirm they've upgraded to the release. Hopefully, this is something which will happen soon.

repos/spack_repo/builtin/packages/psblas/package.py
```diff
@@ -4,6 +4,7 @@
 
 from spack_repo.builtin.build_systems.autotools import AutotoolsPackage
 
+# flake8: noqa: F401,F403
```
haampie:
```suggestion
```

Seems unnecessary, otherwise lgtm

haampie:
If you remove that one line I'll merge. Thanks for improving the package generally

Cirdans-Home:
@haampie I accepted the change; it was a line that was there because my version of Python's formatting checker was complaining, and the manual seemed to suggest it was necessary. Thanks for your help!

haampie (approved):

PR #2884, opened 2025-12-23
amg4psblas: new package

Cirdans-Home:
Create new package for the AMG4PSBLAS library, this is a suite of advanced preconditioners to be used in conjunction with the psblas package. To be used, the pull request must follow the pull request [PR2883](https://github.com/spack/spack-packages/pull/2883) for the new version of psblas.


repos/spack_repo/builtin/packages/amg4psblas/package.py
```diff
@@ -0,0 +1,130 @@
...
+    maintainers("pasquadambra", "cirdans-home", "sfilippone")
+
+    # SPDX identifier of the project's license below.
+    license("BSD-3-Clause", checked_by="cirdans-home")
+
+    version("develop", branch="development")
+    version(
+        "1.2.0",
+        sha256="971cac9917a84dad97eccef76feb89b5ea66afa0b80d13f45a62dd5685c01878",
+        preferred=True,
```
haampie: preferred=True is redundant, 1.2.0-rc3 < 1.2.0

repos/spack_repo/builtin/packages/amg4psblas/package.py
```diff
@@ -0,0 +1,130 @@
...
+    conflicts("~mpi", when="+superlu_dist", msg="SuperLU_dist requires MPI support")
+
+    def configure_args(self):
+        args = [f"--prefix={self.prefix}"]
+        args.append(f"--with-psblas={self.spec['psblas'].prefix}")
+        # Check external libraries
+        if "+mumps" in self.spec:
+            args.append(f"--with-mumpsdir={self.spec['mumps'].prefix}")
+            args.append(f"--with-mumpsincdir={self.spec['mumps'].prefix.include}")
+            args.append(f"--with-mumpslibdir={self.spec['mumps'].prefix.lib}")
```
haampie: `.lib` might be `.lib64` on some platforms (we don't normalize).

You could use `self.spec['mumps'].libs.directories[0]`, (... which reminds me that `.directory` would be nicer API if it could assert there's only one such dir)

repos/spack_repo/builtin/packages/amg4psblas/package.py
```diff
@@ -0,0 +1,129 @@
...
+    )
+    variant("openmp", default=False, description="Activate OpenMP support")
+    # Additional configure options
+    variant("ccopt", default="none", description="Additional CCOPT flags")
+    variant("cxxopt", default="none", description="Additional CXXOPT flags")
+    variant("fcopt", default="none", description="Additional FCOPT flags")
+    variant("extra_opt", default="none", description="Additional EXTRA_OPT flags")
+    variant("libs", default="none", description="Additional link flags")
+    variant("clibs", default="none", description="Additional CLIBS flags")
+    variant("flibs", default="none", description="Additional FLIBS flags")
```
haampie, line 53: Is this needed? If just added for completeness, maybe not worth exposing it until someone needs it. There are also automatic `cflags=... cxxflags=...` variants in Spack that enter the compiler wrapper by default.

repos/spack_repo/builtin/packages/amg4psblas/package.py
```diff
@@ -0,0 +1,129 @@
...
+
+    # Dependencies:
+    depends_on("c", type="build")
+    depends_on("cxx", type="build")
+    depends_on("fortran", type="build")
+    # psblas
+    depends_on("psblas@develop", when="@develop")
+    depends_on("psblas@3.9.0", when="@1.2.0")
+    depends_on("psblas@3.9.0-rc3", when="@1.2.0-rc3")
+    depends_on("psblas@:+mpi", when="+mpi")
```
haampie: the `@:` here and below are redundant

```suggestion
    depends_on("psblas+mpi", when="+mpi")
```

repos/spack_repo/builtin/packages/amg4psblas/package.py
```diff
@@ -0,0 +1,129 @@
...
+
+    def configure(self, spec, prefix):
+        configure = Executable("./configure")
+        configure(*self.configure_args())
+
+    def build(self, spec, prefix):
+        make()
+
+    def install(self, spec, prefix):
+        make("install")
```
haampie: These should be inherited, I think they're redundant.

repos/spack_repo/builtin/packages/amg4psblas/package.py
```diff
@@ -0,0 +1,129 @@
...
+
+    def samples(self, spec, prefix):
```
haampie, line 111: You can also do:

```
@run_after("install")
def samples(...):
```

then you don't need to override `phases = ...`

repos/spack_repo/builtin/packages/amg4psblas/package.py
Cirdans-Home: Fixed, thanks

repos/spack_repo/builtin/packages/amg4psblas/package.py
Cirdans-Home: This permits to ask for position independent code (`-fPIC`) and accommodate flags for different level of optimizations, or debug symbols. Possibly to compile against version of BLAS/LAPACK used in PSBLAS that the configure doesn't recognize automatically. I would prefer to have them exposed, even if most users are not going to use them.

repos/spack_repo/builtin/packages/amg4psblas/package.py
Cirdans-Home: Done, and done also for the other external libraries, `umfpack`, `superlu`, `superlu_dist`

repos/spack_repo/builtin/packages/amg4psblas/package.py
Cirdans-Home: Removed, thanks

repos/spack_repo/builtin/packages/amg4psblas/package.py
Cirdans-Home: Removed phases and added the `@run after` bit before the compilation of the samples.

haampie (approved):

PR #2887, opened 2025-12-24
Logic to install plugins for cmake netcdf-c builds

vanderwb:
While plugin installation support is provided for the autotools build method, this was apparently not implemented for CMake builds of `netcdf-c`. This PR aims to rectify this inconsistency. As with many configurations, the process for turning plugin install on changed in 4.9.3.

From 4.9.0 to 4.9.2, setting `PLUGIN_INSTALL_DIR` to the desired directory is all that is required (assuming we are building a shared library).

Meanwhile, in 4.9.3, a new variable is introduced and documented which serves the same purpose - `NETCDF_WITH_PLUGIN_DIR`. However, setting this doesn't seem to be sufficient as there is no logic in the top-level CMakeLists.txt to set `ENABLE_PLUGIN_INSTALL`, which is required for plugin libraries to be installed in the `plugins` subdirectory. This seems like a bug, but we can work around it by setting this manually.

### DIFF
--- a/repos/spack_repo/builtin/packages/netcdf_c/package.py
+++ b/repos/spack_repo/builtin/packages/netcdf_c/package.py
@@ -394,15 +394,18 @@ class CMakeBuilder(AnyBuilder, cmake.CMakeBuilder):
         else:
             base_cmake_args.append(self.define(nc + "FIND_SHARED_LIBS", False))
 
-        if self.spec.satisfies("@4.9.0:+shared"):
-            # The plugins are not built when the shared libraries are disabled:
-            if self.spec.satisfies("@4.9.3:"):
-                # This toggle is not defined in the top-level CMake parameters but is still
-                # used by the plugin config; so we work around this bug for now
-                base_cmake_args.append(self.define("ENABLE_PLUGIN_INSTALL", True))
-                base_cmake_args.append(self.define("NETCDF_WITH_PLUGIN_DIR", self.prefix.plugins))
-            else:
-                base_cmake_args.append(self.define("PLUGIN_INSTALL_DIR", self.prefix.plugins))
+        # The plugins are not built when the shared libraries are disabled:
+        if self.spec.satisfies("@4.9.3:+shared"):
+            # This toggle is not defined in the top-level CMake parameters but is still
+            # used by the plugin config; so we work around this bug for now
+            base_cmake_args.extend(
+                [
+                    self.define("ENABLE_PLUGIN_INSTALL", True),
+                    self.define("NETCDF_WITH_PLUGIN_DIR", self.prefix.plugins),
+                ]
+            )
+        elif self.spec.satisfies("@4.9.0:+shared"):
+            base_cmake_args.append(self.define("PLUGIN_INSTALL_DIR", self.prefix.plugins))
         return base_cmake_args
 
     @run_after("install")
--- a/repos/spack_repo/builtin/packages/netcdf_c/package.py
+++ b/repos/spack_repo/builtin/packages/netcdf_c/package.py
@@ -393,6 +393,16 @@ class CMakeBuilder(AnyBuilder, cmake.CMakeBuilder):
             base_cmake_args.append(self.define(nc + "FIND_SHARED_LIBS", True))
         else:
             base_cmake_args.append(self.define(nc + "FIND_SHARED_LIBS", False))
+
+        if self.spec.satisfies("@4.9.0:+shared"):
+            # The plugins are not built when the shared libraries are disabled:
+            if self.spec.satisfies("@4.9.3:"):
+                # This toggle is not defined in the top-level CMake parameters but is still
+                # used by the plugin config; so we work around this bug for now
+                base_cmake_args.append(self.define("ENABLE_PLUGIN_INSTALL", True))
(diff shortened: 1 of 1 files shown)


vanderwb:
@WardF and/or @skosukhin - just sending you a ping on this since I submitted it over the holidays. The issue I'm attempting to fix here has lead to problems for downstream packages: we initially saw it with `nccmp` as it produces a "filters not found" error when using CMake builds of netcdf-c in its current form.

vanderwb:
@WardF and/or @skosukhin - sending another ping for a review on this PR. Thanks!

skosukhin:
Sorry for not reacting earlier. I neither maintain nor review the CMake part of this package. I'd recommend pinging people who worked on that before.

vanderwb:
@johnwparent - I believe you added the CMake build for netcdf-c and also set it to be the default builder. Can you take a look at this PR when you have a moment? Thanks!

skosukhin:
> also set it to be the default builder

I thought I missed something. [CMake is the default build system on Windows only](https://github.com/spack/spack-packages/blob/b0240d741183660bc0d14573074652c6f87f8151/repos/spack_repo/builtin/packages/netcdf_c/package.py#L315). Personally, I don't recommend it in any other case.

vanderwb:
Sorry @skosukhin - you're right, of course:

```
default_build_system = "cmake" if sys.platform == "win32" else "autotools"
```

And your recommendation makes sense. That said, I still think this should work when CMake is used and would be happy to see the above issue fixed.

repos/spack_repo/builtin/packages/netcdf_c/package.py
```diff
@@ -393,6 +393,16 @@ def cmake_args(self):
             base_cmake_args.append(self.define(nc + "FIND_SHARED_LIBS", True))
         else:
             base_cmake_args.append(self.define(nc + "FIND_SHARED_LIBS", False))
+
+        if self.spec.satisfies("@4.9.0:+shared"):
```
johnwparent: Lets remove the extra level of branching.
```
if self.spec.satisfies("@4.9.3:+shared"):
   ....
elif self.spec.satisfies("@4.9.0:+shared"):
   ...
```

This way if we hit the first branch, we avoid a second satisfaction check.

repos/spack_repo/builtin/packages/netcdf_c/package.py
```diff
@@ -393,6 +393,16 @@ def cmake_args(self):
...
+            # The plugins are not built when the shared libraries are disabled:
+            if self.spec.satisfies("@4.9.3:"):
+                # This toggle is not defined in the top-level CMake parameters but is still
+                # used by the plugin config; so we work around this bug for now
+                base_cmake_args.append(self.define("ENABLE_PLUGIN_INSTALL", True))
+                base_cmake_args.append(self.define("NETCDF_WITH_PLUGIN_DIR", self.prefix.plugins))
```
johnwparent:
```suggestion
base_cmake_args.extend([
    self.define("ENABLE_PLUGIN_INSTALL", True),
    self.define("NETCDF_WITH_PLUGIN_DIR", self.prefix.plugins),
])
```

johnwparent (commented):
Just a couple of very minor code org requests. Looks good otherwise.

repos/spack_repo/builtin/packages/netcdf_c/package.py
johnwparent: FWIW I would not use this suggestion directly, seems like GH chewed up the formatting

repos/spack_repo/builtin/packages/netcdf_c/package.py
vanderwb: Thanks! Done.

repos/spack_repo/builtin/packages/netcdf_c/package.py
vanderwb: Done. Tweaked the formatting slightly based on a `spack style` suggestion.

vanderwb:
@johnwparent - this should be ready for a second look whenever you have time. Thanks!

johnwparent (approved):
Looks great, thanks for adding this!
I'm approving and will merge later today, but another thought is, should we gate this behind a variant, or is the plugin usage popular enough to always enable it? I'm not too familiar with netcdf plugins so I'll leave that entirely to you.

skosukhin:
I don't know how it is implement in CMake, but with Autotools, it's not possible to build with shared libraries enabled and no plugins at all. And some other plugins can be [disabled with variants](https://github.com/vanderwb/spack-packages/blob/444353ebf83068a1fe68654779a48184deadb364/repos/spack_repo/builtin/packages/netcdf_c/package.py#L159).

vanderwb:
Well, it obviously is configurable with the CMake build, but I think the base set of plugins should be expected since they are built always with autotools as Sergey mentioned. I noticed the CMake build was not including them when a dependent package - `nccmp` - was breaking for me at runtime because it couldn't find the filters "default" filter plugins.

That said, ensuring that the variants supported by the autotools build like blosc are also configurable when using CMake is a good idea, IMO. I'd be in favor of submitting another PR to address that, so that this change goes through now, but I could also look into it and potentially add support in this PR if you prefer @johnwparent.

PR #2888, opened 2025-12-24
py-tilelang: Add torch-c-dlpack-ext to requirements

wu-kan:
See <https://github.com/tile-ai/tilelang/pull/1403>

### DIFF
--- a/repos/spack_repo/builtin/packages/py_torch_c_dlpack_ext/package.py
+++ b/repos/spack_repo/builtin/packages/py_torch_c_dlpack_ext/package.py
@@ -20,11 +20,13 @@ class PyTorchCDlpackExt(PythonPackage):
     depends_on("py-torch")
 
     with default_args(type="build"):
+        depends_on("c")
+        depends_on("cxx")
+        depends_on("cmake")
+        depends_on("ninja")
         depends_on("py-setuptools@61.0:")
         depends_on("py-apache-tvm-ffi@0.1.1:")
         depends_on("py-pybind11")
-        depends_on("cmake")
-        depends_on("ninja")
 
     def setup_build_environment(self, env: EnvironmentModifications) -> None:
         env.prepend_path("CPLUS_INCLUDE_PATH", self.spec["py-pybind11"].prefix.include)
--- a/repos/spack_repo/builtin/packages/py_torch_c_dlpack_ext/package.py
+++ b/repos/spack_repo/builtin/packages/py_torch_c_dlpack_ext/package.py
@@ -1,0 +1,28 @@
+# Copyright Spack Project Developers. See COPYRIGHT file for details.
+#
+# SPDX-License-Identifier: (Apache-2.0 OR MIT)
+
+from spack_repo.builtin.build_systems.python import PythonPackage
+
+from spack.package import *
+
+
+class PyTorchCDlpackExt(PythonPackage):
+    """torch c dlpack ext"""
+
+    homepage = "https://pypi.org/project/torch-c-dlpack-ext/"
+    pypi = "torch-c-dlpack-ext/torch_c_dlpack_ext-0.1.4.tar.gz"
+
+    license("Apache-2.0")
+
+    version("0.1.4", sha256="ad292d17e285ab9523940e51e87d21ffce4982ce8beb46fb18b5c2b4760a1a10")
+
+    depends_on("py-torch")
+
+    with default_args(type="build"):
+        depends_on("py-setuptools@61.0:")
+        depends_on("py-apache-tvm-ffi@0.1.1:")
+        depends_on("py-pybind11")
+    
+    def setup_build_environment(self, env: EnvironmentModifications) -> None:
+        env.prepend_path("CPLUS_INCLUDE_PATH", self.spec["py-pybind11"].prefix.include)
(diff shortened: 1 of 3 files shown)


tldahlgren (commented):
Confirmed the version sha256s and the homepage and dependencies of the new `py-torch-c-dlpack-ext`.

@adamjstewart As the `py-torch` maintainer, would you be interested and available to comment on the new `py-torch-c-dlpack-ext` package?

repos/spack_repo/builtin/packages/py_torch_c_dlpack_ext/package.py
```diff
@@ -0,0 +1,32 @@
...
+    """torch c dlpack ext"""
+
+    homepage = "https://pypi.org/project/torch-c-dlpack-ext/"
+    pypi = "torch-c-dlpack-ext/torch_c_dlpack_ext-0.1.4.tar.gz"
+
+    license("Apache-2.0")
+
+    version("0.1.4", sha256="ad292d17e285ab9523940e51e87d21ffce4982ce8beb46fb18b5c2b4760a1a10")
+
+    depends_on("py-torch")
```
adamjstewart: Presumably needed at run-time?

repos/spack_repo/builtin/packages/py_torch_c_dlpack_ext/package.py
```diff
@@ -0,0 +1,32 @@
...
+
+    with default_args(type="build"):
+        depends_on("c")
+        depends_on("cxx")
+        depends_on("cmake")
+        depends_on("ninja")
```
adamjstewart: I see nothing in the source code to suggest that any C or C++ compiler, cmake, ninja, or pybind11 are used. I'm not even convinced that apache-tvm-ffi is needed. Why were these added?

adamjstewart (changes requested):

repos/spack_repo/builtin/packages/py_torch_c_dlpack_ext/package.py
wu-kan: <https://github.com/apache/tvm-ffi/blob/v0.1.4/addons/torch_c_dlpack_ext/pyproject.toml#L23>

repos/spack_repo/builtin/packages/py_torch_c_dlpack_ext/package.py
wu-kan: In <https://github.com/apache/tvm-ffi/blob/v0.1.4/addons/torch_c_dlpack_ext/README.md>

> ... provides an Ahead-Of-Time (AOT) compiled module to support faster DLPack conversion in DLPack v1.2. By default, tvm-ffi will JIT-compile a version of this functionality during loading, and use a safe-fallback if JIT-compilation fails. Installing this wheel allows users to avoid this JIT compilation overhead and also avoid the cases where the user environment does not necessarily have a compiler toolchain to run JIT-compilation.

`apache-tvm-ffi` id added here <https://github.com/apache/tvm-ffi/blob/v0.1.4/addons/torch_c_dlpack_ext/build_backend.py#L55>.

`torch_c_dlpack_ext` is compiled by `apache-tvm-ffi` <https://github.com/apache/tvm-ffi/blob/v0.1.4/python/tvm_ffi/_optional_torch_c_dlpack.py#L119>

cpp source <https://github.com/apache/tvm-ffi/blob/v0.1.4/python/tvm_ffi/utils/_build_optional_torch_c_dlpack.py#L40>

`py-bind11` comes form headers of torch

repos/spack_repo/builtin/packages/py_torch_c_dlpack_ext/package.py
adamjstewart: Yes, this means it's needed at run-time. The current type is build/link, not run. I usually use build/run for these kinds of deps just in case. In this case, it looks like torch should also be a link dep.

repos/spack_repo/builtin/packages/py_torch_c_dlpack_ext/package.py
adamjstewart: If it's JIT, does the compile happen when you install torch-c-dlpack-ext, or when you run torch-c-dlpack-ext for the first time? I don't see any C source, only C++. Also still nothing to suggest that cmake/ninja are needed during build time.

torch already has a build/link/run dep on pybind11. Does this not get pulled in with your link dep on torch?

repos/spack_repo/builtin/packages/py_torch_c_dlpack_ext/package.py
wu-kan: Done, add link dep

repos/spack_repo/builtin/packages/py_torch_c_dlpack_ext/package.py
wu-kan: Compile happens at the install time. Removed dep on `c`, `cmake` and `pybind11`.
`ninja` is needed when `^ py-apache-tvm-ffi@:0.1.3` or `platform=windows`(<https://github.com/apache/tvm-ffi/pull/272>)

adamjstewart: Hmm, maybe `py-apache-tvm-ffi` should have run-time deps on these instead of `py-torch-c-dlpack-ext` having build-time deps? It's hard to tell, I guess the headers are embedded in `py-torch-c-dlpack-ext` so the deps are specific to that package. It sounds like ninja should go in `py-apache-tvm-ffi` though.

wu-kan: Done, move ninja to runtime deps of py-apache-tvm-ffi

adamjstewart (approved):
Looks better now.

PR #2905, opened 2025-12-29
ufs-weather-model: misc. updates

AlexanderRichert-NOAA:
This PR:
 - enables oneapi support for ufs-weather-model, and corrects the error logic when a supported compiler is not found
 - sets CMAKE_MODULE_PATH
 - sets an upper limit on bacio version, as ufsatm needs bacio_4
 - add cxx dependency (fails without it)


AlexanderRichert-NOAA:
@spackbot fix style

tldahlgren (approved):

PR #2906, opened 2025-12-29
ESMF: Set CMAKE_MODULE_PATH in dependent build environment

AlexanderRichert-NOAA:
This PR sets adds <esmf prefix>/cmake to CMAKE_MODULE_PATH so that dependents (namely, ufs-weather-model) can find it via CMake/find_package(). Setting ESMF_ROOT would also work for ufs-weather-model specifically, but this feels like the more general/safe solution.


AlexanderRichert-NOAA:
doesn't work

PR #2907, opened 2025-12-29
podio, edm4hep, dd4hep: conflicts ^python +freethreading

wdconinc:
This PR marks explicit conflicts between `podio`, `edm4hep`, and `dd4hep` and `^python +freethreading`. To be resolved in next releases (https://github.com/AIDASoft/podio/pull/911, https://github.com/key4hep/EDM4hep/pull/469, https://github.com/AIDASoft/DD4hep/pull/1547).

### DIFF
--- a/repos/spack_repo/builtin/packages/dd4hep/package.py
+++ b/repos/spack_repo/builtin/packages/dd4hep/package.py
@@ -176,6 +176,11 @@ class Dd4hep(CMakePackage):
     # See https://github.com/AIDASoft/DD4hep/issues/1210
     conflicts("^root@6.31.1:", when="@:1.27")
 
+    # See https://github.com/AIDASoft/DD4hep/pull/1547
+    conflicts(
+        "^python +freethreading", when="@:1.34", msg="python free-threading requires dd4hep@1.35:"
+    )
+
     @property
     def libs(self):
         # We need to override libs here, because we don't build a libdd4hep so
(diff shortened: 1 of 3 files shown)


alalazo (approved):

tmadlener (commented):
This comes a bit too late for now, but I think all the conflicts could be removed and replaced by explicitly injecting the python prefix via cmake, all the packages have corresponding cmake variables that should allow for this

 -`podio_PYTHON_INSTALLDIR`
- `EDM4HEP_PYTHON_INSTALLDIR`
- `DD4HEP_PYTHON_INSTALL_DIR`

All of them are populated to some more (or less as we have seen) reasonable default, but they can all be either specified as relative (to the `CMAKE_INSTALL_PREFIX`) or absolute paths and the corresponding python files will end up there.

wdconinc:
> This comes a bit too late for now, but I think all the conflicts could be removed and replaced by explicitly injecting the python prefix via cmake, all the packages have corresponding cmake variables that should allow for this
> 
> -`podio_PYTHON_INSTALLDIR`
> - `EDM4HEP_PYTHON_INSTALLDIR`
> - `DD4HEP_PYTHON_INSTALL_DIR`
> 
> All of them are populated to some more (or less as we have seen) reasonable default, but they can all be either specified as relative (to the `CMAKE_INSTALL_PREFIX`) or absolute paths and the corresponding python files will end up there.

True, but determining the target directory from spack is not trivial either and requires spinning up a subprocess of the target python (since the python being used by spack itself is not the relevant one). That complexity is avoided with a conflict that only restricts until the next versions are released.

PR #2913, opened 2025-12-30
dd4hep: use dyld path for macos

sethrj:
The plugin path used for dd4hep is platform dependent, and the existing value is hardcoded for linux.

### DIFF
--- a/repos/spack_repo/builtin/packages/dd4hep/package.py
+++ b/repos/spack_repo/builtin/packages/dd4hep/package.py
@@ -231,7 +231,13 @@ class Dd4hep(CMakePackage):
         env.set("DD4hep_DIR", self.prefix)
         env.set("DD4hep_ROOT", self.prefix)
         if len(self.libs.directories) > 0:
-            env.prepend_path("LD_LIBRARY_PATH", self.libs.directories[0])
+            # Plugin lookup mechanism is system-dependent
+            libvar = (
+                "DYLD_LIBRARY_PATH"
+                if self.spec.satisfies("platform=darwin")
+                else "LD_LIBRARY_PATH"
+            )
+            env.prepend_path(libvar, self.libs.directories[0])
 
     def url_for_version(self, version):
         # dd4hep releases are dashes and padded with a leading zero


wdconinc (approved):

sethrj:
Thanks @wdconinc !

paulgessinger:
Hi @sethrj, I just realized this is a bit problematic: what happens is that this essentially puts all libs in a view on `DYLD_LIBRARY_PATH`, which the macos dl loader puts first in the search path. What happens then is that there's collateral pollution, e.g. libiconv's dylib ends up in `DYLD_LIBRARY_PATH`. That's a problem when e.g. cmake from spack links against the system provided libiconv (no idea if that's intentional or not), but the loader resolves to the view's dylib, and that then causes symbol lookup problems.

That's not ideal. Can we have the package not modify `DYLD_LIBRARY_PATH`?

jmcarcell:
Doesn't that apply also to any package modifying `LD_LIBRARY_PATH` on Linux and `DYLD_LIBRARY_PATH` if you are in a situation with two libiconv?

wdconinc:
> That's not ideal. Can we have the package not modify `DYLD_LIBRARY_PATH`?

@paulgessinger Are you here because of https://github.com/AIDASoft/DD4hep/pull/1664 :-) ? The goal there is indeed to replace that `dd4hep/package.py` block by `DD4HEP_LIBRARY_PATH` prepends for exactly the reason you outline (also on linux).

> Doesn't that apply also to any package modifying LD_LIBRARY_PATH on Linux and DYLD_LIBRARY_PATH if you are in a situation with two libiconv?

Yes. IMHO no application should be requiring global variable pollution for routine operation in prefixed installs in 2026, and application specific dl-loadable (non-rpath-linked) plugins should not be stored in plain `lib/` (where pollution may occur) but may be better placed in `lib/<application>/` or `lib/<application>/plugins/`, and be found there using a mechanism limited to that application (like dedicated env vars or configs).

sethrj:
@paulgessinger Yes, it is indeed problematic and does break things as iconv, but I think it was necessary at the time because of DD4HEP's loading policy. `DYLD_LIBRARY_PATH` is worse than `LD_LIBRARY_PATH` because, as you discovered, hardwired shared link paths like `/usr/lib/libiconv.dylib` are actually *overridden* by the variable rather than supplemented.

I 100% agree with @wdconinc here.

paulgessinger:
@wdconinc No I hadn't seen the dd4hep issue. Indeed that specific variable never really worked for me. I have a workaround on my end (exclude libiconv from the view) but anything more robust is preferable.

PR #2921, opened 2025-12-31
Update glow

ebagrenrut:
- Add `glow` 2.1.1
- Set `go` version dependencies based on contents of `go.mod` at each tag in the upstream project
- Ensure the `glow` version is set and reported by `glow --version` by passing the `-X main.Version` ldflag. Previously, the version was reported as "version unknown".

/cc @harrysharma1 @alecbcs

### DIFF
--- a/repos/spack_repo/builtin/packages/glow/package.py
+++ b/repos/spack_repo/builtin/packages/glow/package.py
@@ -23,6 +23,7 @@ class Glow(GoPackage):
 
     license("MIT")
 
+    version("2.1.1", sha256="f13e1d6be1ab4baf725a7fedc4cd240fc7e5c7276af2d92f199e590e1ef33967")
     version("1.5.1", sha256="b4ecf269b7c6447e19591b1d23f398ef2b38a6a75be68458390b42d3efc44b92")
     version("1.5.0", sha256="66f2a876eba15d71cfd08b56667fb07e1d49d383aa17d31696a39e794e23ba92")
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/glow/package.py
```diff
@@ -34,6 +35,25 @@ class Glow(GoPackage):
...
+    depends_on("go@1.13:")
+
+    @property
+    def build_args(self):
+        return [
+            "-p",
+            str(self.module.make_jobs),
+            "-modcacherw",
+            "-ldflags",
+            f"-s -w -X main.Version={self.version}",
```
becker33, line 52: Should this be a standard component of `GoPackage`? If that's necessary to make `foo --version` work we should add it to the defaults.

ebagrenrut: At least amongst the packages I'm adding, and others I've fiddled with, it does seem that there's not _really_ a standard way to set the version. Some try to grab it out of a git tag (which of course leads to problems), and not all even specify it using the same variable or within the same scope/package.

ebagrenrut: I _do_ think that what _would_ be more helpful is a way to add more arguments to `ldflags` without having to reproduce all of `build_args`.

becker33: We could add a `ldflags` property that can be overridden? But it might be confusing with the `c/cxx/fortran` ldflags that we support

We could namespace it further to help with that if that's useful

becker33 (approved):

Issue #1942, opened 2025-10-10
Installation issue: p4est


msimberg:
### Steps to reproduce the issue

```console
$ spack spec -I p4est@2.8.7 ~mpi ~openmp
 -   vlznman  p4est@2.8.7~mpi~openmp build_system=autotools platform=linux os=sles15 target=zen2 %c=gcc@14.3.0
[^]  toutdiv      ^compiler-wrapper@1.0 build_system=generic platform=linux os=sles15 target=zen2
[e]  pvkn5vv      ^gcc@14.3.0~binutils+bootstrap~graphite~mold~nvptx~piclibs~profiled~strip build_system=autotools build_type=RelWithDebInfo languages:='c,c++,fortran' platform=linux os=sles15 target=zen2
[^]  zracf7z      ^gcc-runtime@14.3.0 build_system=generic platform=linux os=sles15 target=zen2
[e]  55tksgk      ^glibc@2.31 build_system=autotools platform=linux os=sles15 target=zen2
[^]  it24lin      ^gmake@4.4.1~guile build_system=generic platform=linux os=sles15 target=zen2 %c=gcc@14.3.0
[^]  j5db5a7      ^libtool@2.4.7 build_system=autotools platform=linux os=sles15 target=zen2 %c=gcc@12.3.0
[^]  hizu6da          ^findutils@4.10.0 build_system=autotools patches:=440b954 platform=linux os=sles15 target=zen2 %c=gcc@12.3.0
[^]  tb7ebpu              ^gettext@0.23.1+bzip2+curses+git~libunistring+libxml2+pic+shared+tar+xz build_system=autotools platform=linux os=sles15 target=zen2 %c,cxx=gcc@12.3.0
[^]  ihl64hw                  ^bzip2@1.0.8~debug~pic+shared build_system=generic platform=linux os=sles15 target=zen2 %c=gcc@12.3.0
[^]  7dvwukd                  ^libiconv@1.18 build_system=autotools libs:=shared,static platform=linux os=sles15 target=zen2 %c=gcc@12.3.0
[^]  ms2ywb3                  ^libxml2@2.13.5~http+pic~python+shared build_system=autotools platform=linux os=sles15 target=zen2 %c=gcc@12.3.0
[^]  x5hkpgx                      ^pkgconf@2.3.0 build_system=autotools platform=linux os=sles15 target=zen2 %c=gcc@12.3.0
[...]

PR #2926, opened 2026-01-01
p4est: add explicit -lm to fix link failures with newer toolchains

mathsen:
Some example targets (e.g. v4l2) use math functions such as sin(), but libm was not listed explicitly on the link line. Older toolchains linked this transitively, but newer GCC/binutils reject this with “DSO missing from command line”. Add -lm explicitly to make the link robust and portable.

This should also fix this [issue](https://github.com/spack/spack-packages/issues/1942).


tldahlgren:
@davydden Any objections to this PR being merged?

tldahlgren (approved):

alecbcs:
Merging after >1 month with no response from @davydden.

PR #2928, opened 2026-01-02
openfoam: update to v2512

raphStrickling:
+ Make latest OpenFOAM v2512 release available ([release notes](https://www.openfoam.com/news/main-news/openfoam-v2512))
+ Update URL following the recent OpenFOAM migration to GitLab ([OpenFOAM news](https://www.openfoam.com/news/main-news/openfoam-v2512/usability#gitlab))

### DIFF
--- a/repos/spack_repo/builtin/packages/openfoam/common/spack-Allwmake
+++ b/repos/spack_repo/builtin/packages/openfoam/common/spack-Allwmake
@@ -19,6 +19,13 @@ then
     ./Allwmake-spack $@   # Pass arguments
 else
     ./Allwmake $@         # Pass arguments
+
+    # Build any additional plugins (if the sources exist)
+    if [ -x ./Allwmake-plugins ]
+    then
+        ./Allwmake-plugins -k $@  # Pass arguments
+    fi
+
     ##echo "Disabled build of openfoam"   # When testing environment only
 
     # Generate manpages
--- a/repos/spack_repo/builtin/packages/openfoam/package.py
+++ b/repos/spack_repo/builtin/packages/openfoam/package.py
@@ -424,13 +424,13 @@ class Openfoam(Package):
     patch("1612-spack-patches.patch", when="@1612")
     # kahip patch (wmake)
     patch(
-        "https://develop.openfoam.com/Development/openfoam/commit/8831dfc58b0295d0d301a78341dd6f4599073d45.patch",
+        "https://gitlab.com/openfoam/core/openfoam/commit/8831dfc58b0295d0d301a78341dd6f4599073d45.patch",
         when="@1806",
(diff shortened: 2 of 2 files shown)


olesenm:
Hi @raphStrickling - you were quick (thanks)!
I started this too, but holidays got in the way. In any case, could you also work in parts of the following patch too?

- ensures that older hot-fix patches still make sense
- build the "plugins" if available.  In an early version we migrated some of the code from "modules/" to "plugins/", but still tend to package and compile cfmesh. This patch adds it back in, and makes the build consistent with ubuntu packs, rpm packages and easybuild.

[0001-openfoam-update-git-repo-location-build-optional-plu.patch](https://github.com/user-attachments/files/24468309/0001-openfoam-update-git-repo-location-build-optional-plu.patch)

raphStrickling:
Hi @olesenm, thanks for your feedback!

I updated the URL to the hot-fix patches as you suggested.

Regarding the plugin builds, I added a variant that allows users to specify which plugins to build (by `spack install openfoam@2512 plugins=cfmesh,avalanche`). I suggest to get the source code from git from version 2512 onwards, as this makes it easier for Spack to pull the requested plugins via git submodules.
The plugins variant is set to "none" by default, so cfMesh will not build by default. This is done to ensure backward compatibility, as this plugin has not been available in older versions. I could not find a way to conditionally set the default based on the specified version. So if cfMesh is required, it must be specified with `plugins=cfmesh`.
I wasn’t able to get the research plugin to work. I don’t believe this is a Spack issue. It seems that the Allwmake-plugins script from the OpenFOAM repository may not be compatible with the directory structure of the research repository. Please correct me if I’m mistaken.
Similarly, the data-community and turbulence-community plugins are not supported as of now, since these repos contain submodules themselves, which would require recursively pulling the submodules based on the `plugins` specification.

olesenm:
> Hi @olesenm, thanks for your feedback!
> 
> I updated the URL to the hot-fix patches as you suggested.
> 
> Regarding the plugin builds, I added a variant that allows users to specify which plugins to build (by `spack install openfoam@2512 plugins=cfmesh,avalanche`). I suggest to get the source code from git from version 2512 onwards, as this makes it easier for Spack to pull the requested plugins via git submodules. The plugins variant is set to "none" by default, so cfMesh will not build by default.

Sounds reasonable.

> I wasn’t able to get the research plugin to work.

The "research" plugin is a sandbox area for trying ideas (some of which may not entirely work). Not entirely surprised if it's slightly broken.

> community and turbulence-community plugins are not supported as of now, since these repos contain submodules themselves, which would require recursively pulling the submodules based on the `plugins` specification.

Yup. Some of the data-driven bits have some very adventurous submodules, with questionable code quality.

olesenm (approved):

olesenm (approved):

alalazo (approved):
Approved by maintainer already

PR #2935, opened 2026-01-05
[libvdwxc] Fix the autotools mpi detection

mtaillefumier:
Fix the mpi test in configure.ac


RMeli:
Worth applying [MR 25](https://gitlab.com/libvdwxc/libvdwxc/-/merge_requests/25/diffs) in full?

If yes, you can apply it directly from GitLab, without the need for an additional file:
```
https://gitlab.com/libvdwxc/libvdwxc/-/merge_requests/25.diff
```

mtaillefumier:
thanks for the link. I will modify the PR

repos/spack_repo/builtin/packages/libvdwxc/package.py
```diff
@@ -71,3 +71,10 @@ def configure_args(self):
     # The relevant upstream fix for the m4 would be:
     # https://gitlab.com/libvdwxc/libvdwxc/-/commit/9340f857515c4a2e56d2aa7cf3a21c41ba8559c3.diff
     patch("fftw-detection.patch", when="@:0.4.0")
+
+    # fix a the mpi detection error
```
RMeli:
```suggestion
    # fix a mpi detection error
```
Small nitpick, then LGTM

RMeli (dismissed):

RMeli (approved):

lang-m:
This MR breaks compilation of libvdwxc@0.5.0. The diff from MR25 cannot be applied in full because `newrelease.py` is not part of the release tarball.

@RMeli Is it possible to only use a subset from the diff, doesn't seem to be supported by Spack's `patch`? Assuming no, would it be preferable to add a single patch file here or to apply patches of all relevant commits in MR25?

RMeli:
> @RMeli Is it possible to only use a subset from the diff, doesn't seem to be supported by Spack's patch? Assuming no, would it be preferable to add a single patch file here or to apply patches of all relevant commits in MR25?

Yes, this is what @mtaillefumier did initially. I suggested to use the MR patch because it's usually cleaner, but did not test it myself. ~I'll revert this PR, and then~ *We can* add back b3885041a805845a6914b2e32f2c7f39cedfe809.

mtaillefumier:
@RMeli I will take care of it.

RMeli:
@lang-m shoukd be fixed in https://github.com/spack/spack-packages/pull/2977

PR #2936, opened 2026-01-05
openblas: skip tests during build

sethrj:
This works around an openblas test failure on macos ARM architectures by omitting tests during builds: https://github.com/OpenMathLib/OpenBLAS/issues/5459


mathomp4 (approved):

mathomp4:
@sethrj I assume this is independent of #2890 by @wdconinc ?

sethrj:
@mathomp4 correct; this one only affects the makefile build system.

mathomp4 (approved):

rtrigg:
I tried to reproduce this on macOS Sequoia M1 with apple-clang 17.0.0 / gfortran 15.2.0, and openblas@0.3.28 failed during the build when linking test binaries (dblat1, sblat1) with LTO errors:

```
ld: -lto_library library filename must be 'libLTO.dylib'
make: *** [Makefile:171: tests] Error 2
```

So, different error than the `cblas_sgemm` accuracy failure, but with the PR applied, the build did complete successfully.

albestro (dismissed):
not tested, but for what concerns `nvpl_blas`, it lgtm

sethrj:
whoops actually I just realized I accidentally made that part of the same PR; I can split it out.

alalazo (changes requested):
https://gitlab.spack.io/spack/spack-packages/-/pipelines/1391893

sethrj:
@alalazo Indeed I noticed the last set of openblas builds fail as well. Some subtlety of makefiles is causing a race condition in the build (which claims to be parallel-safe) and I'm trying to find away around it. (I don't see the issue on macos.) Giving it another shot.

sethrj:
@spackbot fix style

sethrj:
@spackbot fix style

sethrj:
@alalazo The CI finally seems to be fixed (ignoring some spurious runner failures, that is). Thanks for your patience!

sethrj:
@alalazo This is good to go please.

alalazo (approved):

repos/spack_repo/builtin/packages/openblas/package.py
```diff
@@ -615,9 +616,17 @@ def make_defs(self):
...
+        """Override 'make all' with sequential builds due to race conditions."""
+        # Due to the verbosity of the command line and number of object files
+        # created, we suppress makefile command echoing via `-s`.
+        args = ["-s"] + self.make_defs
+
+        # Make each target sequentially
+        with working_dir(self.build_directory):
+            for target in ["libs", "netlib", "shared"]:
+                tty.info("Building target", target)
+                make(*(args + [target]))
```
alalazo, line 629: One question though: would it be possible to recover previous behavior `if self.run_tests is True`?

sethrj: It should, yes: just after this, there remains:
```
    @run_after("build")
    @on_package_attributes(run_tests=True)
    def check_build(self):
        make("tests", *self.make_defs, parallel=False)
```
which should build the only remaining target.

PR #2946, opened 2026-01-06
orca: add versions 6.1.1 and avx2-6.1.1

klartz:
Add the latest ORCA versions: 6.1.1 and avx2-6.1.1.

Issue [#47489](https://github.com/spack/spack/pull/47489) discusses a change in ORCA’s filename scheme compared to earlier releases, and suggests treating avx2-6.0.1 as a temporary edge case until a later release clarified which scheme would be used going forward.
Based on the currently available releases, it seems that ORCA is continuing to use the avx2-6.0.1 filename scheme:

    avx2-6.0.0    orca_6_0_0_linux_x86-64_avx2_shared_openmpi416.tar.xz
    avx2-6.0.1    orca_6_0_1_linux_x86-64_shared_openmpi416_avx2.tar.xz
    avx2-6.1.1    orca_6_1_1_linux_x86-64_shared_openmpi418_avx2.tar.xz

So I’ve also modified the URL construction logic to treat avx2-6.0.0 as the edge case instead, and to use the newer filename scheme for later versions.

### DIFF
--- a/repos/spack_repo/builtin/packages/orca/package.py
+++ b/repos/spack_repo/builtin/packages/orca/package.py
@@ -25,6 +25,10 @@ class Orca(Package):
     license("LGPL-2.1-or-later")
 
     version(
+        "avx2-6.1.1", sha256="5eaf676f9711a38835d609264321a30266b487b65477547802dedee982bc82d5"
+    )
+    version("6.1.1", sha256="a0bc1d6d2c3c00620367bbc5dbf2b3a7018abc92d1ff65f06cec46f75350b9be")
+    version(
         "avx2-6.0.1", sha256="f31f98256a0c6727b6ddfe50aa3ac64c45549981138d670a57e90114b4b9c9d2"
     )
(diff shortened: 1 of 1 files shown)


tldahlgren (commented):
I cannot confirm the version sha256s since this is a manual download.

@snehring ?

snehring:
I had gone through the effort of doing this for 6.1.0 and then apparently never actually commited and made a PR. Whoops.

This looks good to me. The checksums match what I've downloaded.

snehring (approved):

snehring:
> I had gone through the effort of doing this for 6.1.0 and then apparently never actually commited and made a PR. Whoops.

ah, actually it was in #1957

klartz:
Oops, my bad. I hadn't seen that PR.

snehring:
All good, it's stale so this one can supersede it.

tldahlgren (approved):

PR #2957, opened 2026-01-07
ci: remove sha256 long args

haampie:
Also remove a script that was used to install a newer make, which is redundant.


kwryankrattiger (approved):
LGTM

PR #2963, opened 2026-01-07
surfer: add v0.5.0 and v0.4.0

davekeeshan:
Add v0.5.0 and v.0.4.0

Also upreved the versions of rust, not all versions of rust are available


tldahlgren (approved):
Confirmed both version's commit+tag.

PR #2971, opened 2026-01-08
adios2: update sz dependency in newer adios

vicentebolea:


vicentebolea:
@eisenhauer current adios2 sz3 is for sz3 master, 3.1 and 3.2 tags fails to build. Also lets get rid of SZ2 support and make ADIOS2_USE_SZ to refer to SZ3

eisenhauer:
New PR in ADIOS should work with SZ 3.1, 3.2 and master.  (Versions 3.3.0, 3.3.1 and 3.3.2 are released, but aren't in spack yet.). Lets ping @pnorbert about the USE_SZ change...   The change to SZ3 is a clean break (like adios to adios2) on the side of the maintainers.  I wouldn't want to try to paper over that...

vicentebolea:
> The change to SZ3 is a clean break (like adios to adios2) on the side of the maintainers. I wouldn't want to try to paper over that...

That makes sense to keep maintaining both variants (at least temporarly), find it reflected in the last push of this PR

vicentebolea:
@ax3l ready for review :)

repos/spack_repo/builtin/packages/adios2/package.py
```diff
@@ -74,7 +74,8 @@ class Adios2(CMakePackage, CudaPackage, ROCmPackage):
     variant("bzip2", default=True, description="Enable BZip2 compression")
     variant("zfp", default=True, description="Enable ZFP compression")
     variant("png", default=True, description="Enable PNG compression")
-    variant("sz", default=True, description="Enable SZ compression")
+    variant("sz", default=True, description="Enable SZ2 compression")
+    variant("sz3", default=True, when="@2.12:", description="Enable SZ3 compression")
```
tldahlgren, line 78: Can both of these options be enabled at the same time?

If not, you could add a conflict (see https://spack.readthedocs.io/en/latest/packaging_guide_creation.html#conflicts).

tldahlgren (commented):
See comment for question.

repos/spack_repo/builtin/packages/adios2/package.py
vicentebolea: The can be enabled at the same time. No conflict needed

tldahlgren (approved):
LGTM.

Ping me when it's okay to merge this PR.

vicentebolea:
@tldahlgren from my side it is ready. Thanks!!

PR #2977, opened 2026-01-08
fix libvdwxc configure

simonpintarelli:
Follow-up for #2935. Tested on my computer ;) 

The patch from https://github.com/spack/spack-packages/pull/2935#issuecomment-3718062193 was aimed at the gitlab tarball and didn't apply.


RMeli:
Would b3885041a805845a6914b2e32f2c7f39cedfe809 (local file patch as originally done) be a better approach, so it does not have to rely on your GitHub profile?

simonpintarelli:
> Would [b388504](https://github.com/spack/spack-packages/commit/b3885041a805845a6914b2e32f2c7f39cedfe809) (local file patch as originally done) be a better approach, so it does not have to rely on your GitHub profile?

Yes, indeed.

mtaillefumier (approved):

mtaillefumier:
@simonpintarelli you can add me as reviewer if I am not already included

RMeli (approved):
Thanks guys!

PR #2982, opened 2026-01-08
malt: Add 1.4.0, 1.4.1, 1.5.0 and mark some old deprecated

svalat:
Add new versions of MALT : 
 - 1.4.0
 - 1.4.1
 - 1.5.0

Since 1.4.0 don't need anymore deps on NodeJS.

File extension has changed to `.xz`in 1.5.0 => I don't know if there is a more generic way to handle that gap ?


RikkiButler20 (approved):
LGTM

PR #2984, opened 2026-01-09
Support alsa plugins for sound using pulseaudio/alsa-lib

biddisco:
This adds a new alsa-plugins package and upgrades the existing alsa-lib and pulseaudio packages so that they work with applications such as musescore https://github.com/musescore/MuseScore


repos/spack_repo/builtin/packages/alsa_lib/package.py
```diff
@@ -13,19 +13,27 @@ class AlsaLib(AutotoolsPackage):
     space library that developers compile ALSA applications against."""
 
     homepage = "https://www.alsa-project.org"
-    url = "ftp://ftp.alsa-project.org/pub/lib/alsa-lib-1.2.3.2.tar.bz2"
+    url = "ftp://ftp.alsa-project.org/pub/lib/alsa-lib-1.2.15.2.tar.bz2"
+    git = "https://github.com/alsa-project/alsa-lib.git"
 
     license("LGPL-2.1-or-later")
 
+    version("1.2.15.2", tag="v1.2.15.2", commit="63a981865a1c7d9501ae556e28ae3bb53d015b61")
```
wdconinc: Is there a particular reason to use a tag here? That prevents caching the source.

repos/spack_repo/builtin/packages/alsa_lib/package.py
```diff
@@ -13,19 +13,27 @@ class AlsaLib(AutotoolsPackage):
     space library that developers compile ALSA applications against."""
 
     homepage = "https://www.alsa-project.org"
-    url = "ftp://ftp.alsa-project.org/pub/lib/alsa-lib-1.2.3.2.tar.bz2"
+    url = "ftp://ftp.alsa-project.org/pub/lib/alsa-lib-1.2.15.2.tar.bz2"
```
wdconinc, line 16: Just as an FYI, it is not necessary to update the `url` each time a new version is added. Spack determines this from older urls automatically.

repos/spack_repo/builtin/packages/alsa_lib/package.py
```diff
@@ -13,19 +13,27 @@ class AlsaLib(AutotoolsPackage):
...
 
     variant("python", default=False, description="enable python")
+    variant("plugins", default=False, description="enable plugins")
 
     patch("python.patch", when="@1.1.4:1.1.5 +python")
 
     depends_on("c", type="build")  # generated
+    depends_on("autoconf", type="build", when="@1.2.15.2 build_system=autotools")
+    depends_on("automake", type="build", when="@1.2.15.2 build_system=autotools")
+    depends_on("libtool", type="build", when="@1.2.15.2 build_system=autotools")
```
wdconinc: This seems surprising. Why would only this particular version depend on these packages? And what other option for `build_system` is there?

repos/spack_repo/builtin/packages/alsa_plugins/package.py
```diff
@@ -0,0 +1,57 @@
...
+    homepage = "https://github.com/alsa-project/alsa-plugins"
+    url = "https://github.com/alsa-project/alsa-plugins"
+    git = "https://github.com/alsa-project/alsa-plugins"
+
+    # notify when the package is updated.
+    maintainers("biddisco")
+
+    license("LGPL-2.1-or-later", checked_by="biddisco")
+
+    version("1.2.12", tag="v1.2.12", commit="52574cb5ccbb8b546df2759e4b341a20332269b6")
```
wdconinc, line 24: Same question here about using a git tag instead of a download.

repos/spack_repo/builtin/packages/alsa_plugins/package.py
```diff
@@ -0,0 +1,57 @@
...
+
+    depends_on("c", type="build")
+    depends_on("autoconf", type="build")
+    depends_on("automake", type="build")
+    depends_on("libtool", type="build")
+
+    depends_on("alsa-lib")
+    depends_on("pulseaudio +alsa", when="+pulseaudio")
+
+    variant("pulseaudio", default=True, description="Enable pulseaudio support")
```
wdconinc: Often it is easier to set the default as False, except for really indispensable functionality.

repos/spack_repo/builtin/packages/alsa_plugins/package.py
```diff
@@ -0,0 +1,57 @@
...
+
+    conflicts("platform=darwin", msg="ALSA (+plugins) only works for Linux")
+
+    def configure_args(self):
+        args = []
+        if "+pulseaudio" in self.spec:
+            args.append("--enable-pulseaudio")
```
wdconinc: Suggestion to use `self.enable_or_disable` to avoid automatically finding a system pulseaudio.

wdconinc: e.g. https://github.com/spack/spack-packages/blob/544ce9ee44831906ea6463f80bfec6f19a4fc017/repos/spack_repo/builtin/packages/lzo/package.py#L37

repos/spack_repo/builtin/packages/alsa_plugins/package.py
```diff
@@ -0,0 +1,57 @@
...
+        return args
+
+    def setup_build_environment(self, env):
+        env.prepend_path("PKG_CONFIG_PATH", self.spec["alsa-lib"].prefix.lib64.pkgconfig)
```
wdconinc: Surprised that this is necessary. Normally this should be found automatically. Any reason you could identify for why it isn't?

repos/spack_repo/builtin/packages/alsa_plugins/package.py
```diff
@@ -0,0 +1,57 @@
...
+        # 1. ALSA plugin directory
+        alsa_plugin_dir = os.path.join(self.prefix.lib, "alsa-lib")
+        if os.path.exists(alsa_plugin_dir):
+            env.prepend_path("ALSA_PLUGIN_DIR", alsa_plugin_dir)
+
+        # 2. PulseAudio libraries (needed for dlopen dependencies)
+        if "pulseaudio" in self.spec:
+            pulseaudio_libdir = os.path.join(self.spec["pulseaudio"].prefix.lib64, "pulseaudio")
+            if os.path.exists(pulseaudio_libdir):
+                env.prepend_path("LD_LIBRARY_PATH", pulseaudio_libdir)
```
wdconinc: Is there no other variable that can be used and that alsa respects? LD_LIBRARY_PATH is a big hammer and this can lead to significant side effects (even though they are mitigated here by adding `pulseaudio` to the path components).

repos/spack_repo/builtin/packages/pulseaudio/package.py
```diff
@@ -58,6 +73,9 @@ class Pulseaudio(AutotoolsPackage):
     depends_on("speexdsp@1.2:")
     depends_on("m4", type="build")
 
+    #    if spec.satisfies("%gcc@8:"):
+    os.environ["LDFLAGS"] = "-Wl,--copy-dt-needed-entries"
```
wdconinc: This is not canonical in spack. Modifications to the environment are not intended to leak outside the spack invocation (this does), and we have `setup_build_environment` or `flag_handler` for this.

repos/spack_repo/builtin/packages/pulseaudio/package.py
```diff
@@ -58,6 +73,9 @@ class Pulseaudio(AutotoolsPackage):
...
+
     def configure_args(self):
```
wdconinc: It is safer to add the configure_args and meson_args to separate builders, e.g. https://github.com/spack/spack-packages/blob/544ce9ee44831906ea6463f80bfec6f19a4fc017/repos/spack_repo/builtin/packages/dbus/package.py#L80

repos/spack_repo/builtin/packages/pulseaudio/package.py
```diff
@@ -97,3 +115,36 @@ def configure_args(self):
...
+            "-Ddefault_library=shared",
+            "-Dprefix={0}".format(self.prefix),
+            "-Dlibdir={0}/lib64".format(self.prefix),
+            "-Dbashcompletiondir={0}/share/bash-completion/completions".format(self.prefix),
+            "-Dsystemduserunitdir={0}/lib/systemd/user".format(self.prefix),
+        ]
+
+    def setup_build_environment(self, env):
+        if self.spec.satisfies("build_system=meson"):
+            env.append_flags("CPPFLAGS", "-I{0}".format(self.spec["libiconv"].prefix.include))
```
wdconinc, line 82: Should this be CXXFLAGS? CPP is for the preprocessor.

I am also surprised these are necessary if you depend on libiconv. Why does the build not find the right include directory?

repos/spack_repo/builtin/packages/pulseaudio/package.py
```diff
@@ -97,3 +115,36 @@ def configure_args(self):
...
+            "-Ddoxygen=false",
+            "-Dbluez5=disabled",
+            "-Dx11=disabled",
+            "-Dtests=false",
+            "-Ddefault_library=shared",
+            "-Dprefix={0}".format(self.prefix),
+            "-Dlibdir={0}/lib64".format(self.prefix),
+            "-Dbashcompletiondir={0}/share/bash-completion/completions".format(self.prefix),
+            "-Dsystemduserunitdir={0}/lib/systemd/user".format(self.prefix),
+        ]
```
wdconinc: We typically try to make sure that two build systems of the same package at least in spirit behave the same way. In the case of configure, you have enable dbus and glib2, but that's not explicitly added for meson. Can you verify that?

repos/spack_repo/builtin/packages/pulseaudio/package.py
```diff
@@ -97,3 +115,36 @@ def configure_args(self):
...
+            env.append_flags("CPPFLAGS", "-I{0}".format(self.spec["libiconv"].prefix.include))
+            env.append_flags("LDFLAGS", "-L{0} -liconv".format(self.spec["libiconv"].prefix.lib))
+
+    def setup_run_environment(self, env):
+        # Always add ALSA plugin directory
+        env.prepend_path("ALSA_PLUGIN_DIR", join_path(self.prefix.lib, "alsa-lib"))
+
+        # If pulseaudio is in the spec, make sure its libraries are on the runtime linker path
+        if self.spec.satisfies("^pulseaudio"):
+            # 1️⃣ Standard lib64 dir
```
wdconinc: Try to stick to regular ascii charsets.

repos/spack_repo/builtin/packages/pulseaudio/package.py
```diff
@@ -97,3 +115,36 @@ def configure_args(self):
...
+            env.prepend_path("LD_LIBRARY_PATH", self.spec["pulseaudio"].prefix.lib64)
+
+            # 2️⃣ Also check for a pulseaudio subdirectory (some builds install here)
+            pulseaudio_libdir = join_path(self.spec["pulseaudio"].prefix.lib64, "pulseaudio")
+            if os.path.exists(pulseaudio_libdir):
+                env.prepend_path("LD_LIBRARY_PATH", pulseaudio_libdir)
```
wdconinc: Same comment as earlier about LD_LIBRARY_PATH.

repos/spack_repo/builtin/packages/alsa_plugins/package.py
biddisco: They were missing before, and since I added them, I made them from the newer version, but I'll get rid of the version dependency

repos/spack_repo/builtin/packages/alsa_plugins/package.py
biddisco: ok, defaulted to False

repos/spack_repo/builtin/packages/alsa_plugins/package.py
biddisco: I could not find another way to get it running without the user having to manually add ENV vars or other tweaks, however, I will add a comment to the code here and look at it again to try to find a better way.

repos/spack_repo/builtin/packages/alsa_plugins/package.py
biddisco: apologies, that comment was related to the previous question about `type="build", when="@1.2.15.2 build_system=autotools")` - 

tags removed and replaced with urls

repos/spack_repo/builtin/packages/alsa_plugins/package.py
biddisco: I have removed it. Seems to still build. Not sure why It was there.

repos/spack_repo/builtin/packages/pulseaudio/package.py
biddisco: fixed

repos/spack_repo/builtin/packages/pulseaudio/package.py
biddisco: good catch

repos/spack_repo/builtin/packages/pulseaudio/package.py
biddisco: I'm not expert at meson setup, but the meson build has 'auto' configuration for dbus/glib etc and I figured I would set the minimum I needed to get a working build and let the build system choose anything else.

I'd prefer to not touch all the options until I find a situation where a build fails or a user asks for some adjustment.

repos/spack_repo/builtin/packages/pulseaudio/package.py
biddisco: interesting. I tried removing it and things still work. I will comment it out, but leave a comment in case I need to revisit this. I will try the other one. When I first tried to get builds and tests running I added a lot of flags and perhaps some can be removed now that other things are fixed and working.

repos/spack_repo/builtin/packages/alsa_plugins/package.py
biddisco: ok, I double checked and when I remove this runtime lib path, the als-plugins are not correctly loaded up as they depend on some ancillary plugin helper libraries that are not picked up correctly
```
[pid 2643253] openat(AT_FDCWD, "/etc/ld.so.cache", O_RDONLY|O_CLOEXEC) = 18
ALSA lib dlmisc.c:339:(snd_dlobj_cache_get0) [error.core] Cannot open shared library libasound_module_pcm_pulse.so (libpulsecommon-17.0.so: cannot open shared object file: No such file or directory)
```
I will leave in the `LD_LIBRARY_PATH` setting, but keep experimenting

biddisco:
I have rebased this onto latest develop (at the time of writing), and fixed (I think) all of the review comments that I can address immediately.
Thank you for taking the time to check this. I hope it is now ok/better.

biddisco:
it seems that actually, removing some run time env vars has broken the alsa-plugin load. I will recheck everything

biddisco:
The runtime paths for alsa-plugins loading are now functioning as expected.

repos/spack_repo/builtin/packages/pulseaudio/package.py
```diff
@@ -58,42 +72,64 @@ class Pulseaudio(AutotoolsPackage):
...
+                "--disable-systemd-login",
+                "--disable-udev",
+                "--disable-waveout",
+                "--enable-dbus",
+                "--enable-glib2",
+                "--with-database=gdbm",
+                "--with-systemduserunitdir=no",
+                "CXXFLAGS={0}".format(self.spec["libtool"].headers.cpp_flags),
+                "LDFLAGS={0}".format(self.spec["libtool"].libs.search_flags),
+                "--libdir={0}".format(self.prefix.lib64),
```
wdconinc: The use of `lib64` here seems like a host operating system specific detail.

repos/spack_repo/builtin/packages/pulseaudio/package.py
```diff
@@ -58,42 +72,64 @@ class Pulseaudio(AutotoolsPackage):
...
+                "-Ddatabase=gdbm",
+                "-Ddoxygen=false",
+                "-Dbluez5=disabled",
+                "-Dx11=disabled",
+                "-Dtests=false",
+                "-Ddefault_library=shared",
+                "-Dprefix={0}".format(self.prefix),
+                "-Dlibdir={0}/lib64".format(self.prefix),
+                "-Dbashcompletiondir={0}/share/bash-completion/completions".format(self.prefix),
+                "-Dsystemduserunitdir={0}/lib/systemd/user".format(self.prefix),
```
wdconinc: The use of `lib64` in libdir, but then `lib` in systemduserunitdir seems inconsistent, and host OS specific.

wdconinc: I think `prefix` is likely already added automatically for meson packages.

wdconinc (commented):
Some more minor comments.

biddisco:
ok, I have removed any deliberate paths set with {var}/lib or {var}/lib64 and instead used `self.prefix.lib` so that spack can choose based on it's better knowledge. 
Also added one other path fix because after a spack update, the install failed due to an install rule being incorrect (bash completions or something being installed into system path instead of spack install location). 
Thanks for reviewing this.

wdconinc (commented):
Making good progress. I am wondering if/why this works. Most packages with multiple build systems have the builder classes at file level, not class level. Also, setup_build_environment is then a function inside the builder class, not the top level package class. I think right now it may be that setup_build_environment is ignored. Best to conform to the canonical way. Sorry for not catching this on the previous review.

Other suggestions is to use f-strings where the argument does not include any quotes, e.g. self.prefix.lib, but not necessarily for self.spec["libiconv"].prefix.lib.

biddisco:
Fixed the setup_build_environment, used a common base class for both builders.
Also tested the autoconf build on version @13 and had to fix one type conversion error that newer gcc's didn't like

wdconinc (approved):

PR #2985, opened 2026-01-09
AOCC: add v5.1.0

amd-toolchain-support:
What’s new in AOCC 5.1 (https://www.amd.com/en/developer/aocc.html)

- Based on LLVM 17.0.6 release (llvm.org, Nov 2023)
- Zen5 tuned AOCL-LibM 5.2 ([AMD Math Library](https://www.amd.com/en/developer/aocl/libm.html))
- Bug fixes impacting performance of C, C++, and Fortran applications
- Deprecated the `-finline-aggressive` option. Use the `-inline-threshold` option instead
- Deprecated the `-freroll-loops` option


RikkiButler20 (approved):
Looks good to me.

PR #2988, opened 2026-01-09
py-ray: add v2.53.0

thomas-bouvier:


repos/spack_repo/builtin/packages/py_ray/package.py
```diff
@@ -16,43 +16,61 @@ class PyRay(PythonPackage):
...
+    depends_on("py-click@7:", when="@2.53.0:", type=("build", "run"))
     depends_on("py-click@7:8.0.4", when="@2.0.1", type=("build", "run"))
     depends_on("py-filelock", type=("build", "run"))
     depends_on("py-grpcio@1.32:1.43.0", when="@2.0.1 ^python@:3.9", type=("build", "run"))
     depends_on("py-grpcio@1.42:1.43.0", when="@2.0.1 ^python@3.10:", type=("build", "run"))
     depends_on("py-jsonschema", type=("build", "run"))
     depends_on("py-msgpack@1", type=("build", "run"))
-    depends_on("py-numpy@1.19.3:", type=("build", "run"))
+    depends_on("py-numpy@1.19.3:", when="@2.0.1 ^python@3.9:", type=("build", "run"))
+    depends_on("py-numpy@1.16:", when="@2.0.1 ^python@:3.8", type=("build", "run"))
```
tldahlgren: I'm not seeing the python version constraint in https://github.com/ray-project/ray/blob/ray-2.0.1/python/requirements.txt.

repos/spack_repo/builtin/packages/py_ray/package.py
```diff
@@ -16,43 +16,61 @@ class PyRay(PythonPackage):
...
 
     with when("+default"):
         depends_on("py-aiohttp@3.7:", type=("build", "run"))
         depends_on("py-aiohttp-cors", type=("build", "run"))
         depends_on("py-colorful", type=("build", "run"))
-        depends_on("py-py-spy@0.2:", type=("build", "run"))
-        depends_on("py-gpustat@1:", type=("build", "run"))
+        depends_on("py-py-spy@0.4:", when="@2.53.0: ^python@3.12:", type=("build", "run"))
+        depends_on("py-py-spy@0.2:", when="@2.53.0: ^python@:3.11", type=("build", "run"))
+        depends_on("py-py-spy@0.2:", when="@2.0.1", type=("build", "run"))
```
tldahlgren: Is there a reason these two can't be combined as the default when the dependency condition above is not met?

```suggestion
        depends_on("py-py-spy@0.2:", when="@2.0.1:", type=("build", "run"))
```

repos/spack_repo/builtin/packages/py_ray/package.py
```diff
@@ -16,43 +16,61 @@ class PyRay(PythonPackage):
...
+        depends_on("py-grpcio@1.42:", when="@2.53.0: ^python@3.10:", type=("build", "run"))
+        depends_on("py-grpcio@1.32:", when="@2.53.0: ^python@:3.9", type=("build", "run"))
+        depends_on("py-gpustat@1:", when="@2.0.1", type=("build", "run"))
         depends_on("py-opencensus", type=("build", "run"))
-        depends_on("py-pydantic", type=("build", "run"))
-        depends_on("py-prometheus-client@0.7.1:0.13", type=("build", "run"))
+        depends_on("py-pydantic@1,2.12:2", type=("build", "run"))
```
tldahlgren, line 53: This doesn't match the requirement for `0.2.1` at https://github.com/ray-project/ray/blob/03b6bc7b5a305877501110ec04710a9c57011479/python/requirements.txt#L24.

repos/spack_repo/builtin/packages/py_ray/package.py
```diff
@@ -16,43 +16,61 @@ class PyRay(PythonPackage):
 
     license("Apache-2.0")
 
+    version("2.53.0", sha256="bb2e1393e0617b2edbdbc793718a5dbe98d5024e9f2ab06b33ecc524b02c9e0e")
     version("2.0.1", sha256="b8b2f0a99d2ac4c001ff11c78b4521b217e2a02df95fb6270fd621412143f28b")
 
     variant("default", default=False, description="Install default extras")
 
     depends_on("c", type="build")
     depends_on("cxx", type="build")
```
tldahlgren, line 28: Why is there no dependency on `py-aiosignal` for `2.0.1` per `requirements.txt` (see https://github.com/ray-project/ray/blob/03b6bc7b5a305877501110ec04710a9c57011479/python/requirements.txt#L9)?

(Besides the fact the package would need to be added to Spack.)

tldahlgren: 🤔 Perhaps not needed based on https://github.com/ray-project/ray/blob/0de211850589aea71f842873bc32574c702ab492/python/requirements_compiled.txt#L71,

tldahlgren: I'm also not seeing a requirement on `copy-cuda12x` for non-darwin systems for `2.53.0` per https://github.com/ray-project/ray/blob/0de211850589aea71f842873bc32574c702ab492/python/requirements.txt#L11.

repos/spack_repo/builtin/packages/py_ray/package.py
tldahlgren: There should be a dependency on `py-numpy@1.20:` for `2.53.0` per https://github.com/ray-project/ray/blob/0de211850589aea71f842873bc32574c702ab492/python/requirements.txt#L25.

tldahlgren (changes requested):
The `py-ray` dependencies are rather challenging to confirm given the number and organization of dependencies across the two versions. I've noted a few "inconsistencies" but it would be easier to manager/review with a different organization, such as use of the `when` context (see the example in the docs at https://spack.readthedocs.io/en/latest/packaging_guide_creation.html#conditional-dependencies).

tldahlgren:
Also, especially given there is no maintainer on either package, kindly paste the results of `spack debug report` in a comment for the platform(s) you've been able to successfully built the packages (see https://spack.readthedocs.io/en/latest/roles_and_responsibilities.html#package-contributors).

repos/spack_repo/builtin/packages/py_ray/package.py
thomas-bouvier: I looked at `setup.py` for both versions:

I see this for version 2.0.1: https://github.com/ray-project/ray/blob/ray-2.0.1/python/setup.py#L294-L315

And numpy seems to be optional (variant +data) for version 2.53.0: https://github.com/ray-project/ray/blob/ray-2.53.0/python/setup.py#L226-L246

repos/spack_repo/builtin/packages/py_ray/package.py
thomas-bouvier: No reason, fixing it ;)

repos/spack_repo/builtin/packages/py_ray/package.py
thomas-bouvier: I followed this: https://github.com/ray-project/ray/blob/ray-2.53.0/python/setup.py#L230. I am not sure what dependencies we should consider.

repos/spack_repo/builtin/packages/py_ray/package.py
thomas-bouvier: Adding `py-aiosignal`

thomas-bouvier:
So in summary I am not sure where to look for dependencies. Could be https://github.com/ray-project/ray/blob/ray-2.53.0/python/setup.py#L402-L412 and https://github.com/ray-project/ray/blob/ray-2.53.0/python/setup.py#L224-L286, or https://github.com/ray-project/ray/blob/ray-2.53.0/python/requirements.txt. I think the latter accounts for all variants being installed, which is not what we aim for in the current PR.

thomas-bouvier:
This is the machine on which I tested the package:

```
* **Spack:** 1.2.0.dev0 (https://github.com/spack/spack/commit/6e390cb55bc9bd9591cf8d4b4042178e210ee1bd)
* **Builtin repo:** https://github.com/spack/spack-packages/commit/7fa4170e9cec36794bce59a0ea2e62af204cee39
* **Python:** 3.12.3
* **Platform:** linux-ubuntu24.04-skylake
```

thomas-bouvier:
According to adamjstewart on Slack:

> The source of truth should go pyproject.toml -> setup.cfg -> setup.py -> requirements.txt in order from best to worst. Basically, trust what package managers like pip/uv use, don't trust other things. Requirements.txt often includes pinned versions or extra deps for CI testing, not for actual usage.

So I think we should consider `setup.py` in this case @tldahlgren

tldahlgren (approved):
I agree the presence of different sources of requirements for `ray` makes identifying dependencies challenging.

You provided information on the system you've successfully built the packages on so I'm going to approve and merge.

thomas-bouvier:
Thank you @tldahlgren !

PR #2989, opened 2026-01-09
AOCL: add version 5.2

amd-toolchain-support:
Summary of changes in PR:
- Add new version 5.2 for AOCL libraries (refer https://www.amd.com/en/developer/aocl.html)
- Add a new AOCL library `aocl-dlp` (refer https://www.amd.com/en/developer/aocl/dlp.html)

> AOCL-DLP (Deep Learning Primitives) is a high-performance library that provides optimized deep learning primitives for AMD processors
- Fix `aocl-utils` configuration in `amdlibm` recipe for cmake build support
- Update `amdlibm` recipe to use `cmake` as default build_system for `amdlibm@5.2`

### DIFF
--- a/repos/spack_repo/builtin/packages/amdlibm/package.py
+++ b/repos/spack_repo/builtin/packages/amdlibm/package.py
@@ -102,8 +102,8 @@ class Amdlibm(SConsPackage, CMakePackage):
         spec = self.spec
 
         args = [
-            self.define("LIBAOCLUTILS_INCLUDE_PATH", spec["aocl-utils"].prefix.include),
-            f"-DLIBAOCLUTILS_LIBRARY_PATH={spec['aocl-utils'].libs}",
+            self.define("AOCL_UTILS_INCLUDE_DIR", spec["aocl-utils"].prefix.include),
+            f"-DAOCL_UTILS_LIB={spec['aocl-utils'].libs}",
         ]
         return args
 
--- a/repos/spack_repo/builtin/packages/amdlibm/package.py
+++ b/repos/spack_repo/builtin/packages/amdlibm/package.py
@@ -46,7 +46,15 @@ class Amdlibm(SConsPackage, CMakePackage):
     variant("verbose", default=False, description="Building with verbosity", when="@:4.1")
 
     # Build system
-    build_system(conditional("cmake", when="@5.1:"), "scons", default="scons")
+    # - For version 5.2: both 'cmake' and 'scons' are supported, with 'cmake' as default.
+    # - For version 5.1: both 'scons' and 'cmake' are supported, with 'scons' as default.
+    # - For versions below 5.1: only 'scons' is supported.
+    with when("@5.2"):
+        build_system("cmake", "scons", default="cmake")
+
+    with when("@:5.1"):
+        build_system("scons", conditional("cmake", when="@5.1:"), default="scons")
+
     # Mandatory dependencies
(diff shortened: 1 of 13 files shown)


repos/spack_repo/builtin/packages/amdlibm/package.py
```diff
@@ -101,8 +110,8 @@ def cmake_args(self):
         spec = self.spec
 
         args = [
-            self.define("LIBAOCLUTILS_INCLUDE_PATH", spec["aocl-utils"].prefix.include),
-            f"-DLIBAOCLUTILS_LIBRARY_PATH={spec['aocl-utils'].libs}",
+            self.define("AOCL_UTILS_INCLUDE_DIR", spec["aocl-utils"].prefix.include),
+            f"-DAOCL_UTILS_LIB={spec['aocl-utils'].libs}",
```
becker33, line 114: Are these changes appropriate for old versions as well? This looks like it needs a version guard.

repos/spack_repo/builtin/packages/aocl_dlp/package.py
```diff
@@ -0,0 +1,64 @@
...
+    version("5.2", sha256="1eec26eeaf427cb2377ec21415ddce6e1bc62d4eab8ec51630a9c02711019c1c")
+
+    # Feature toggles mapping directly to AOCL-DLP CMake options
+    variant("benchmarks", default=True, description="Build benchmarks (BUILD_BENCHMARKS)")
+    variant("tests", default=True, description="Enable tests (BUILD_TESTING)")
+    variant(
+        "ctest", default=False, description="Set DLP_CTEST_DISABLED=ON to skip CTest invocation"
+    )
+    variant("examples", default=True, description="Build examples (BUILD_EXAMPLES)")
+    variant("shared", default=True, description="Build shared libraries (BUILD_SHARED_LIBS)")
```
becker33: These descriptions are overly repetitive, please make them less so.

For the ctest variable, please make the logic clearer (I assume `+ctest` sets `DLP_CTEST_DISABLED=False`, but the description is unclear. Something simpler like `enable ctest` would be clearer in this case.

becker33 (changes requested):

repos/spack_repo/builtin/packages/aocl_dlp/package.py
ningli-amd: Agreed. There's no need to have those CMake variables corresponding to these variants in the descriptions. 

> I assume +ctest sets DLP_CTEST_DISABLED=False

Yes you are correct on this. The description is for end users so just say 'Enable ctest' is good enough.

repos/spack_repo/builtin/packages/amdlibm/package.py
amd-toolchain-support: `amdlibm` has supported cmake builds since version `5.1`, and the current changes address an existing issue in `v5.1`, so a version guard isn't necessary.

repos/spack_repo/builtin/packages/aocl_dlp/package.py
amd-toolchain-support: variants description has been updated through https://github.com/spack/spack-packages/pull/2989/commits/6e21c3e0f74c333b96c5d9adde5be6e4cae37119 to make it clear.

becker33 (approved):

PR #2992, opened 2026-01-09
Update MET: set LD_LIBRARY_PATH to find Python libraries at runtime, add variant shared-intel, find g2c libraries

climbfuji:
## Description

This PR updates the MET package from the JCSDA spack-stack-dev fork. This package is being used extensively by the spack-stack partners and verified to work.

Notes.
1. To fix the missing Python libraries in the MET executables, we initially tried to use RPATH (the recommended way) using the commented-out logic in this PR. This, however, led to segfaults for many platforms/compilers. Thus, we went for the LD_LIBRARY_PATH route.
2. The `shared-intel` variant allows users to link against `libintlc.so` instead of the legacy `libirc.so`, which can also lead to segmentation faults because of missing symbols (there is a history for this in the old spack packages repository).

### DIFF
--- a/repos/spack_repo/builtin/packages/met/package.py
+++ b/repos/spack_repo/builtin/packages/met/package.py
@@ -42,6 +42,15 @@ class Met(AutotoolsPackage):
     variant("modis", default=False, description="Enable compilation of modis")
     variant("graphics", default=False, description="Enable compilation of mode_graphics")
 
+    # JCSDA fork only
+    variant(
+        "shared-intel",
+        default=False,
+        when="%oneapi",
+        description="Enable linking to shared intel libraries (libintlc instead of libirc)",
+    )
+
+    depends_on("c", type="build")  # generated
     depends_on("cxx", type="build")  # generated
     depends_on("fortran", type="build")  # generated
 
@@ -66,6 +75,7 @@ class Met(AutotoolsPackage):
     depends_on("py-numpy", when="+python", type=("build", "run"))
     depends_on("py-xarray", when="+python", type=("build", "run"))
     depends_on("py-pandas", when="+python", type=("build", "run"))
+    depends_on("patchelf@0.13:", when="platform=linux", type="build")
 
     patch("openmp_shape_patch.patch", when="@10.1.0")
 
@@ -96,6 +106,8 @@ class Met(AutotoolsPackage):
     def setup_build_environment(self, env: EnvironmentModifications) -> None:
         spec = self.spec
         cppflags = []
+        fflags = []
+        fcflags = []
         ldflags = []
         libs = []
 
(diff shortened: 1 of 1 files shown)


climbfuji:
@spackbot fix style

repos/spack_repo/builtin/packages/met/package.py
```diff
@@ -42,6 +42,14 @@ class Met(AutotoolsPackage):
     variant("modis", default=False, description="Enable compilation of modis")
     variant("graphics", default=False, description="Enable compilation of mode_graphics")
 
+    variant(
```
AlexanderRichert-NOAA, line 45: Could/should this be set through cflags/fflags variant?

climbfuji: How does this work?

AlexanderRichert-NOAA: I guess it strikes me as not ideal, though not necessarily a dealbreaker, to do this through a variant. Possible solutions:
 - fix the underlying compiler issue
 - short of that, the recipe would ideally make this automatic based on compiler/version/etc. (`if spec.satisfies(...): flags.append(...)`; in any case I think the `spec.satisfies("%oneapi")` is redundant based on how the variant is defined)
 - a user could spec `met fflags='-shared-intel'`
 - a user could put `-shared-intel` in their flags for whatever compiler package.
 
It would help to get input from a Spack dev here, or if there's precedent for this in other packages that would be useful information.

climbfuji: There is a bit of history behind these flags. Intel says "libirc.so" is deprecated **but** at the same time insists on it remaining the default, because there are many customers who use the legacy libraries. I told them that these can cause segmentation faults due to missing symbols, but they didn't want to change it.

My position therefore is that if Intel doesn't want the default to be anything else than the old, semi-broken libraries, then Spack shouldn't do that either. That basically rules out your first two suggestions.

It's ok to set
```
packages:
  met:
      require:
      - fflags='shared-intel'
```
for met and other packages that need them (e.g. parallel-netcdf), but the problem is that it's much more visible/better documented with the variant. How would a user otherwise know? It took us weeks to come up with this solution.

AlexanderRichert-NOAA: Yeah I take your point. I don't want to overthink it, so I'll approve and let the Spack devs chime in if there are better approaches.

AlexanderRichert-NOAA (approved):

repos/spack_repo/builtin/packages/met/package.py
climbfuji: Thanks! I'll ping one of the spack developers for advice. I do agree that the additional `spec.satisfies("%oneapi")` is redundant and I will remove that.

climbfuji: @becker33 Sorry to bother you, but do you have any feedback on the discussion above regarding declaring a variant `shared-intel` versus other ways to set these flags?

repos/spack_repo/builtin/packages/met/package.py
```diff
@@ -133,7 +149,11 @@ def setup_build_environment(self, env: EnvironmentModifications) -> None:
 
         if "+grib2" in spec:
             g2c = spec["g2c"]
-            env.set("MET_GRIB2CLIB", g2c.libs.directories[0])
+            shared_g2c = True if "+shared" in g2c else False
```
tldahlgren, line 152: Nit: Is `shared_g2c = "+shared" in g2c` not sufficient?

tldahlgren (approved):

repos/spack_repo/builtin/packages/met/package.py
climbfuji: I think so, yes. Do you want me to push the change before merging?

tldahlgren: It's not necessary.

tldahlgren: There was at least one comment that needed to be resolved. That was not one I initiated.

repos/spack_repo/builtin/packages/met/package.py
climbfuji: I'll go ahead and resolve this, @AlexanderRichert-NOAA approved the changes as they are now.

PR #2994, opened 2026-01-09
vtk-m: add conflict for Kokkos 5

rbberger:
Kokkos 5 requires C++20. vtk-m fails to compile with it. I was considering explicitly directing people to use viskores, but I don't know too much about it and how compatible they are.

> [!NOTE]
> Viskores has the appropriate changes for C++20, so should probably be used moving forward. See this [changelog](https://github.com/Viskores/viskores/blob/c854a593a47fc4e55e35090f18de9187aeac39b1/docs/changelog/memory-order-constants.md).

### DIFF
--- a/repos/spack_repo/builtin/packages/vtk_m/package.py
+++ b/repos/spack_repo/builtin/packages/vtk_m/package.py
@@ -107,6 +107,7 @@ class VtkM(CMakePackage, CudaPackage, ROCmPackage):
     # VTK-m uses the default Kokkos backend
     depends_on("kokkos", when="+kokkos")
     depends_on("kokkos@3.7:3.9", when="@2.0 +kokkos")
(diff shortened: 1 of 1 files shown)


tpadioleau (approved):
Looks good to me.

I wanted to consider `depends_on("kokkos@:4", when="+kokkos")` but I am not sure how to interpret `depends_on("kokkos@3.7:3.9", when="@2.0 +kokkos")`, why is it specific to `2.0.x` ?

kmorel (approved):

kmorel:
I agree with your assessment. Applications still using VTK-m would have to be updated to Viskores, but this will be expected going forward. I don't see a reason to patch VTK-m to use C++20 unless there is a compelling use case for that.

johnwparent (approved):

PR #2941, opened 2026-01-05
realm: new package

rbberger:
Extracted from #1231 without explicit changes to the GASNet Spack package. We can get this one merged without waiting on GASNet changes. This one also defaults to using NVCC as cuda compiler. Not adding any released version, since there is only a RC tag at the moment.

@muraj @elliottslaughter

PR #2997, opened 2026-01-10
legion: add new releases, add realm as new dependency

rbberger:
Extracted from #1231 without explicit changes to the GASNet Spack package. Based on #2941.

### DIFF
--- a/repos/spack_repo/builtin/packages/legion/package.py
+++ b/repos/spack_repo/builtin/packages/legion/package.py
@@ -66,6 +66,9 @@ class Legion(CMakePackage, CudaPackage, ROCmPackage):
     depends_on("realm+rocm", when="@25.09.0: +rocm")
     depends_on("realm+cuda_unsupported_compiler", when="@25.09.0: +cuda_unsupported_compiler")
 
+    for d in range(1, 10):
+        depends_on(f"realm max_dims={d}", when=f"max_dims={d}")
+
     # force same compiler as kokkos if static build
     depends_on("kokkos %gcc", when="+kokkos~shared %gcc")
     depends_on("kokkos %clang", when="+kokkos~shared %clang")
--- a/repos/spack_repo/builtin/packages/legion/package.py
+++ b/repos/spack_repo/builtin/packages/legion/package.py
@@ -11,7 +11,7 @@ from spack_repo.builtin.build_systems.rocm import ROCmPackage
 from spack.package import *
 
 
-class Legion(CMakePackage, ROCmPackage):
+class Legion(CMakePackage, CudaPackage, ROCmPackage):
     """Legion is a data-centric parallel programming system for writing
     portable high performance programs targeted at distributed heterogeneous
     architectures. Legion presents abstractions which allow programmers to
@@ -30,8 +30,11 @@ class Legion(CMakePackage, ROCmPackage):
 
     license("Apache-2.0")
 
-    maintainers("pmccormick", "streichler", "elliottslaughter")
+    maintainers("pmccormick", "streichler", "elliottslaughter", "rbberger")
     tags = ["e4s"]
+    version("25.12.0", tag="legion-25.12.0", commit="6f710cb46590b04ad299362819fdecb3a4e429ca")
+    version("25.09.0", tag="legion-25.09.0", commit="8759d840099a138b5f395e86c841848520b34b73")
+    version("25.06.0", tag="legion-25.06.0", commit="d8e35c48d089014b0f764181b7b90278a7558b21")
     version("25.03.0", tag="legion-25.03.0", commit="04716e3b3686d4af71e6a4398dfbe8cd869c057b")
     version("24.12.0", tag="legion-24.12.0", commit="2f087ebe433a19f9a3abd05382f951027933bad9")
     version("24.09.0", tag="legion-24.09.0", commit="4a03402467547b99530042cfe234ceec2cd31b2e")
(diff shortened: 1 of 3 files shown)


tldahlgren (dismissed):
Confirmed the new version commit+tags.

@elliottslaughter @pmccormick @streichler Any objections to this PR being merged?

tldahlgren:
> Confirmed the new version commit+tags.
> 
> @elliottslaughter @pmccormick @streichler Any objections to this PR being merged?

Spoke too soon. Looks like there are CI failures with the package, e.g., https://gitlab.spack.io/spack/spack-packages/-/jobs/20470139.

elliottslaughter:
My main feedback is that the the Realm version is unpinned. The only version that the Legion developers are willing to promise to work with is the one specifically pinned in the Legion repository. As of Legion 25.09.0 and 25.12.0 that version is https://github.com/StanfordLegion/realm/commit/7ef8789fc0a40abf6774152d4061420cff2e307f. The version in current master is https://github.com/StanfordLegion/realm/commit/0ef7edc8c012d4ab6a50805c044cec8a8edeae33 and I'd expect that before our next release we'll probably bump it at least once more.

Other versions of Realm may work but you're on your own in terms of ensuring compatibility.

mpokorny:
Echoing @elliottslaughter above, is there some way to add default Realm versions for Legion releases? It's not clear to me how to do that in Spack when the Realm version included by Legion doesn't really have a version number associated with it (as far as Spack can tell). To use this package currently, I check out the Legion source tree in a local "develop" tree, and symlink the Realm source in that tree to create what looks to Spack like a Realm develop tree. I haven't encountered any problems using that method, but it's pretty clumsy.

mpokorny:
I have a proposal for your consideration. I'm going to assume that the Realm package eventually gets versions beyond "main", but this proposal doesn't require it. For argument's sake, assume that Realm has versions 1.0 and 1.1, as both inherent Realm versions and Realm Spack package versions. Now assume that Legion version 2028.03.0 uses a git commit somewhere between Realm 1.0 and 1.1, say with a commit sha of "deadbeef". My proposal starts with creating a Spack package version for Realm like
```python
version("1.0-lg2028.03.0", commit="deadbeef")
```
I believe that adding the "-lg2028..." suffix to the version number puts that version between 1.0 and 1.1 according to  Spack version number ordering. Additionally, if Realm does not add version numbers but may eventually do so, I believe that a standalone "lg2028..." version number would precede any numeric version number as well as "main". We could of course do this in Spack without adding tags to Realm's repository. How much the version ordering matters really depends on how we might use those numbers in the Legion package, as I show below.

Next, for the Legion package, we add a boolean variant, say `safe_realm`, which defaults to `True`.
```python
variant("safe_realm", default=True, description="Restrict Realm dependency to the known working version")

depends_on("realm") # option A
depends_on("realm@1.0-lg2028.03.0:", when="@2028.03.0 ~safe_realm") # option B

depends_on("realm@1.0-lg2028.03.0", when="@2028.03.0 +safe_realm") # always include this
```
Which option is favored may depend on whether there is (or will be) a total ordering of the Realm package version numbers, including the versions used by Legion. Of course, that pattern would have to be extended for different Legion package versions.

My motivation is to make it easier for Legion users to install a specific version of Legion with the approved Realm version for the release. I would guess this is the primary, or even sole, use case for the Realm package, as I know of no other Soack packages that depend on Realm. It is important to make this use case work reliably. While there may be other use cases for a Realm dependency, and I think this proposal can support them, given the current situation "fixing" those (speculative) cases is less urgent.

Comments?

mpokorny:
All the above aside, this package is missing the propagation of `max_dims` to Realm.

rbberger:
@elliottslaughter to make any meaningful solution, similar to what @mpokorny proposes, it would be good to know what versioning scheme Realm will be targeting. Is it going to follow the `YY.MM.XX` scheme or something else?

rbberger:
@mpokorny Added a few updates. This now adds two realm versions that match what is bundled in legion. It's not perfect, but allows the maximum flexibility while providing meaningful defaults.
- `legion@25.09.0`, `legion@25.12.0` and `legion@26.03.0` will pull in the latest 'safe' commit from realm `26.03.0-legion` as of today.
- if you want to use the "safe" realm commit for 25.09, 25.12 all you need to do is add the requirement of `^realm@25` to your config
- newer commits in `realm@main` can still be used for all versions.

The main question here is whether the selected versioning scheme in Realm matches what they will use in the end. Otherwise we would have to fix these versions later on to not interfere with the regular version selection logic.

elliottslaughter:
Looks like they followed `YY.MM.X` except that the fields are not zero-padded. This is the only current RC I'm aware of: https://github.com/StanfordLegion/realm/tree/v25.6.1-rc.4

Hopefully we'll get actual Realm releases at some point, but until then this seems reasonable to me.

repos/spack_repo/builtin/packages/legion/package.py
```diff
@@ -52,55 +56,85 @@ class Legion(CMakePackage, ROCmPackage):
...
+    depends_on("realm+hwloc", when="@25.09.0: +hwloc")
+    depends_on("realm+openmp", when="@25.09.0: +openmp")
+    depends_on("realm+cuda", when="@25.09.0: +cuda")
+    depends_on("realm+rocm", when="@25.09.0: +rocm")
+    depends_on("realm+cuda_unsupported_compiler", when="@25.09.0: +cuda_unsupported_compiler")
+
+    depends_on("realm@26:", when="@26:")
+    depends_on("realm@25:", when="@25.09.0:")
+
+    for d in range(9):
```
mpokorny:
```python
for d in range(1, 10):
```

mpokorny (commented):
I like what you've done, with one correction.

I agree that we should understand the version numbering Realm intends to use, if any.

rbberger:
@elliottslaughter With #1409 merged, we now have usable gasnet spack package. Can we proceed with merging this one? Realm doesn't force any particular gasnet configuration based on the conduit selection like legion did, but this is a good start.

elliottslaughter:
Looks good to me, thanks.

junghans (dismissed):

eugeneswalker (approved):

PR #2999, opened 2026-01-10
frobby: add v0.9.7

d-torrance:
With the new version, we can use a tarball instead of a git tag and simplify "make install".


RikkiButler20 (approved):
Looks good to me.

PR #3000, opened 2026-01-11
doxygen: add v1.16.0

mcmehrtens:
* Add Doxygen v1.16.0 and v1.16.1 to the Doxygen Spack package
* Maintainers: @sethrj


sethrj:
@mcmehrtens it looks like 1.16.1 was release this morning to fix a few regressions; could you please add it instead/in addition? https://www.doxygen.nl/manual/changelog.html

sethrj (approved):
Checksums are valid. Thanks!

PR #3011, opened 2026-01-12
bufr_query: add v0.0.5

mathomp4:
[bufr-query v0.0.5](https://github.com/NOAA-EMC/bufr-query/releases/tag/v0.0.5) was released. First one in over a year. Let's add it to spack!

NOTE: I'm fairly certain [from the diffs](https://github.com/NOAA-EMC/bufr-query/compare/v0.0.4...v0.0.5) that it doesn't have any new dependencies, but I'll ask @rmclaren to confirm here.

### DIFF
--- a/repos/spack_repo/builtin/packages/bufr_query/package.py
+++ b/repos/spack_repo/builtin/packages/bufr_query/package.py
@@ -20,6 +20,7 @@ class BufrQuery(CMakePackage, PythonExtension):
 
     license("Apache-2.0", checked_by="srherbener")
 
+    version("0.0.5", sha256="72cb82d3837aac0dbe2c514b9883b4a19aad2ab7c6969c760e1b549ab4ed4831")
     version("0.0.4", sha256="cc21a298c03ee3a8938823301606e91c7b321005da284ebf2c9b25044bfcbad8")
     version("0.0.3", sha256="f2952a190cc1d7714a3bfe481fb1545459639ba304fc31b941062b471dea1d41")
     version("0.0.2", sha256="b87a128246e79e3c76e3158d89823e2ae38e9ee1a5a81b6f7b423837bdb93a1f")
--- a/repos/spack_repo/builtin/packages/bufr_query/package.py
+++ b/repos/spack_repo/builtin/packages/bufr_query/package.py
@@ -37,12 +37,10 @@ class BufrQuery(CMakePackage, PythonExtension):
     depends_on("eckit@1.24.4:", type=("build", "run"))
     depends_on("eigen@3:", type=("build", "run"))
     depends_on("gsl-lite", type=("build", "run"))
-
-    # If using netcdf-c 4.9.3 or later, you must use bufr-query 0.0.5 or later
+    depends_on("netcdf-c", type=("build", "run"))
+    # bufr-query versions up to 0.0.4 are incompatible with netcdf-c 4.9.3 and later
     # https://github.com/NOAA-EMC/bufr-query/issues/58
-    depends_on("netcdf-c@4.9.3:", when="@0.0.5:", type=("build", "run"))
     depends_on("netcdf-c@:4.9.2", when="@:0.0.4", type=("build", "run"))
-
     depends_on("netcdf-cxx4", type=("build", "run"))
     depends_on("bufr", type=("build", "run"))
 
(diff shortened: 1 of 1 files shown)


rmclaren (approved):

rmclaren:
@mathomp4 It does use WXFlow for logging, let me. It also has a dependency on pvlib (only needed when computing solar angles). We are trying to get rid of that one...

mathomp4:
> @mathomp4 It does use WXFlow for logging, let me. It also has a dependency on pvlib (only needed when computing solar angles). We are trying to get rid of that one...

Ohhh. I didn't know it had a Python variant. Yeah, I'd have never noticed that. Maybe we need to add some bits for that.

rmclaren:
I'm thinking of not worying about pvlib for now. You can always install it manually if you actually do need it.

mathomp4:
> I'm thinking of not worying about pvlib for now. You can always install it manually if you actually do need it.

Okay. At the moment spack seems to only care about `py-pybind11`:
```python
    with when("+python"):
        extends("python")
        depends_on("py-pybind11", type="build")
```

rmclaren:
> > I'm thinking of not worying about pvlib for now. You can always install it manually if you actually do need it.
> 
> Okay. At the moment spack seems to only care about `py-pybind11`:
> 
> ```python
>     with when("+python"):
>         extends("python")
>         depends_on("py-pybind11", type="build")
> ```

You will get an python import error if wxflow is not available (when using obsbuilder). py-wxflow is avaliable via spack stack already so I suppose that will be OK (?).

mathomp4:
> You will get an python import error if wxflow is not available (when using obsbuilder). py-wxflow is avaliable via spack stack already so I suppose that will be OK (?).

Hmm. Okay. Let me do some test builds to make sure i get the dependencies right.

srherbener (approved):
This looks good to me, but I haven't tested it.  I'm assuming that the CI tests will check the hash for the version 0.0.5 tag.

mathomp4:
I'm converting to draft. I'm having issues building on macOS. I think I know the fixes, but I don't want this merged until then.

mathomp4:
Okay. This was fun. I ask @srherbener and @rmclaren to look it over. I had to do a few more fixes.

First, this package somehow missed the sweep for:
```python
    depends_on("c", type="build")
```
and spack v1. Not sure how, but it did. Without this, building with apple-clang on my macbook was going nuts.

Second, because of https://github.com/NOAA-EMC/bufr-query/issues/58, I've restricted 0.0.4 and older to `netcdf-c` 4.9.2 or older. 

Third, I found that I had to fiddle with the flags being passed in to also pass in the `netcdf-cxx4` include dir. At least, I needed it for 0.0.5, but it didn't seem to hurt 0.0.4.

Finally, I added the `py-wxflow` dependency for 0.0.5.

With these changes, I can build `bufr-query` 0.0.4 and 0.0.5 on my macbook. I don't have a Linux box readily available at the moment, but I should tomorrow.

rmclaren:
Thanks for finding that @mathomp4! I barely even remember that change we made for the newer versions of netcdf... Thanks for testing it so thouroughly.

rmclaren (approved):

mathomp4:
> Thanks for finding that @mathomp4! I barely even remember that change we made for the newer versions of netcdf... Thanks for testing it so thouroughly.

I've learned to always try building my PRs here. Too many times I had to make a second one with "Fix what I broke in number XYZ" :)

srherbener (approved):
Thanks for contributing this, and taking the time to test it thoroughly!

tldahlgren (approved):
Confirmed the version sha256.

rmclaren:
@mathomp4 Quick question. Under which conditions did you find that you needed the following:

```cmake
# Join the include directories from both netcdf-c and netcdf-cxx4
# Spack's .headers.include_flags produces strings like "-I/path/to/inc"
nc_includes = self.spec["netcdf-c"].headers.include_flags
ncxx_includes = self.spec["netcdf-cxx4"].headers.include_flags

combined_flags = f"{nc_includes} {ncxx_includes}"

args.extend(
    [
        self.define("CMAKE_C_FLAGS", combined_flags),
        self.define("CMAKE_CXX_FLAGS", combined_flags),
    ]
)
```

I haven't encountered a situation where I have needed to do this... and someone wrote a bug for it https://github.com/NOAA-EMC/bufr-query/issues/148#issuecomment-3750913467.

mathomp4:
@rmclaren When I was trying to build `bufr-query` 0.0.5 on my mac, it hit a point where it failed to build. It was a `#include <netcdf>` like like maybe this one:

https://github.com/NOAA-EMC/bufr-query/blob/563d7b3c4b5ce6cc679ee45547edd1f0bd4615ef/core/src/encoders/netcdf/Encoder.cpp#L12

in `Encoder.cpp`. Maybe CMake saw the file was C++ and needed the include files from `netcdf-cxx4`?

Note: I was building via Spack. So maybe it didn't have access to the same `FindNetCDF.cmake` some builds outside of spack might have? 

It seemed necessary for 0.0.5.

PR #2679, opened 2025-12-05
FFTW: check if neon is in float_simd_features before removing

gardner48:

PR #3019, opened 2026-01-13
fftw: add check before removing 'neon' list item

kuberry:
In SHA `d13c881b1a3` on 12/3/25, the following was added:
```
         # GCC on apple silicon does not support Neon intrinsics
         if self.spec.satisfies("platform=darwin target=aarch64: %gcc"):
                 float_simd_features.remove("neon")
```
It is possible that "neon" is not in float_simd_features. Adding:
```
             if "neon" in float_simd_features:
```

### DIFF
--- a/repos/spack_repo/builtin/packages/fftw/package.py
+++ b/repos/spack_repo/builtin/packages/fftw/package.py
@@ -158,7 +158,8 @@ class FftwBase(AutotoolsPackage):
 
         # GCC on apple silicon does not support Neon intrinsics
         if self.spec.satisfies("platform=darwin target=aarch64: %gcc"):
-            float_simd_features.remove("neon")
+            if "neon" in float_simd_features:
+                float_simd_features.remove("neon")
 
         simd_options = []
         for feature in simd_features:


kuberry:
Currently, the package logic regarding `neon` looks like:

```
simd_features = ["sse2", "avx", "avx2", "avx512", "avx-128-fma", "kcvi", "vsx", "asimd"]
float_simd_features = ["altivec", "sse", "neon"]
```
Note: "asimd" in simd_features and "neon" in float_simd_features.

```
# armv8 introduced asimd to replace neon feature
if "asimd" in spec.target:
    float_simd_features.remove("neon")
```
For aarch64 macs, something like target=m1 will likely be used, which means that `asimd` will be in spec.target (removing neon).

```
# GCC on apple silicon does not support Neon intrinsics
if self.spec.satisfies("platform=darwin target=aarch64: %gcc"):
        float_simd_features.remove("neon")
```
without a check, this will attempt to remove neon again on a mac.

gardner48:
This is also addressed in https://github.com/spack/spack-packages/pull/2679

kuberry:
Thank you @gardner48. I kicked off the pipeline again on https://github.com/spack/spack-packages/pull/2679.

kuberry:
The purpose of this PR is solved by #2679

PR #3022, opened 2026-01-13
wps: add 4.6.0

emwjacobson:
Added a new version of WPS.

Pinging maintainer @MichaelLaufer

### DIFF
--- a/repos/spack_repo/builtin/packages/wps/package.py
+++ b/repos/spack_repo/builtin/packages/wps/package.py
@@ -17,6 +17,7 @@ class Wps(Package):
     url = "https://github.com/wrf-model/WPS/archive/v4.2.tar.gz"
     maintainers("MichaelLaufer")
 
+    version("4.6.0", sha256="ca7bbfc6c28a107c6eb00ded70e693f5c9a3926ecde7656f49e306c9eb9a309b")
     version("4.5", sha256="3cb29107f85b81af05b7dd494e5d4a13cf1b33b87d7e2911a64a7abc59bf55f3")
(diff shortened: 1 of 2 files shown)


repos/spack_repo/builtin/packages/wps/package.py
```diff
@@ -38,7 +39,8 @@ class Wps(Package):
     patch("patches/4.2/preamble.patch", when="@4.2:")
     patch("patches/4.3/arch.configure.defaults.patch", when="@=4.3")
     patch("patches/4.3.1/arch.configure.defaults.patch", when="@4.3.1")
-    patch("patches/4.4/configure.patch", when="@4.4:")
+    patch("patches/4.4/configure.patch", when="@4.4:4.5")
+    patch("patches/4.6/configure.patch", when="@4.6:")
```
alalazo, line 43: Can you add a brief comment above these `patch` directives explaining:
2. Why it is needed
3. Where it was taken from

?

emwjacobson: Being honest I just followed the pattern of the other existing patches. Prior to me adding 4.6.0, these patches already existed I believe to handle netcdf-c and netcdf-fortan being separate packages - I just added a new one for 4.6.0 which needed to be slightly different.

If the existing comment of `# These patches deal with netcdf & netcdf-fortran being two diff things` is not sufficient, maybe @MichaelLaufer (the original maintainer) could comment more on them?

bernhardkaindl (approved):
Hi @emwjacobson! I think your answer to the review comment is reasonable, and I don't know what more to do or to answer.

PR #3027, opened 2026-01-14
python@3.14 +optimizations: --tail-call-interp if possible

haampie:
Should be in addition to `--computed-gotos` (still used in the regex engine).


adamjstewart (approved):

haampie:
Everything builds except one py-tensorflow due to system failure, I'm merging it.

PR #3028, opened 2026-01-14
cuda: add v13.1.0 and v13.1.1

stephenswat:
This commit adds versions 13.1.0 and 13.1.1 of the CUDA package.

### DIFF
--- a/repos/spack_repo/builtin/build_systems/cuda.py
+++ b/repos/spack_repo/builtin/build_systems/cuda.py
@@ -211,7 +211,8 @@ class CudaPackage(PackageBase):
         conflicts("%gcc@13:", when="+cuda ^cuda@:12.3")
         conflicts("%gcc@14:", when="+cuda ^cuda@:12.6")
         conflicts("%gcc@15:", when="+cuda ^cuda@:12.9")
-        conflicts("%gcc@16:", when="+cuda ^cuda@:13.0")
+        conflicts("%gcc@16:", when="+cuda ^cuda@:13.1")
+        conflicts("%gcc@15:", when="+cuda ^cuda@13.1:")
         conflicts("%clang@12:", when="+cuda ^cuda@:11.4.0")
         conflicts("%clang@13:", when="+cuda ^cuda@:11.5")
         conflicts("%clang@14:", when="+cuda ^cuda@:11.7")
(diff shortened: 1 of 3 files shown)


pauleonix:
Seems like PAPI is incompatible with the newest CUPTI included with CUDA 13.1. CUPTI [introduced](https://docs.nvidia.com/cupti/release-notes/release-notes.html#updates-in-cuda-13-1) a C++/C23 `bool` in their "C-based interface" ([`bAllowDeviceLevelCounters`](https://docs.nvidia.com/cupti/api/structCUpti__Profiler__GetCounterAvailability__Params.html#_CPPv4N44CUpti_Profiler_GetCounterAvailability_Params25bAllowDeviceLevelCountersE)).

Edit: I reported this to NVIDIA.

pauleonix (approved):
CUDA changes LGTM

pauleonix (changes requested):
Forgot that the conflicts for GCC and Clang as host compilers need to be updated for 13.1 as well in https://github.com/spack/spack-packages/blob/b59cd7b52ee6e71ae894af18671273958f9819cd/repos/spack_repo/builtin/build_systems/cuda.py#L203-L224

I took a look at https://docs.nvidia.com/cuda/cuda-installation-guide-linux/index.html#host-compiler-support-policy and very surprisingly there actually seems to be step back in host compiler support from gcc@:15 and clang@:20 to gcc@:14 and clang@:19 (see https://docs.nvidia.com/cuda/archive/13.0.0/cuda-installation-guide-linux/index.html#host-compiler-support-policy for comparison). Given that this was not mentioned in the release notes I wonder if the docs are actually correct :grimacing:

Could you test locally if e.g. `cuda@13.1 ~allow-unsupported-compilers ^gcc@15` installs fine?

stephenswat:
I added the incompatibility in PAPI and I updated to build system to meet the compiler requirements specified in the CUDA installation guide.

pauleonix (approved):
LGTM

alalazo (approved):

pauleonix:
It seems like the host compiler support was just a documentation issue as it now shows support for gcc@15 and clang@21. So currently the conflicts are a bit more conservative than needed.

PR #3034, opened 2026-01-14
llvm: fix penalties

haampie:
In the past we avoided conditional variants in LLVM because people found the
tri-state annoying (+foo, ~foo, no variant foo at all).

For example, setting `~lua` in config would imply the existence of the lua
variant, which would trigger the condition `@11:+lldb`, ultimately enabling
`+lldb`, even though the point of adding it was to disable a bunch of
variants.

However, there's a better reason (imo) to return to conditional variants,
because if you put `~lldb` in config now, then due to the conflict on `+lua`,
you take the penalty of the non-default variant `~lua`, which as a user is
confusing, as lua doesn't have anything to do with the spec you're interested
in anymore.

In short: toggling off a feature should not result in concretization penalties
because it also implies turning off sub-features.

In the worst case this can even lead to non-determinism, because the
concretizer can choose between disabling sub-features vs actual features.

So, bring back conditional variants again.

### DIFF
--- a/repos/spack_repo/builtin/packages/llvm/package.py
+++ b/repos/spack_repo/builtin/packages/llvm/package.py
@@ -155,16 +155,23 @@ class Llvm(CMakePackage, CudaPackage, LlvmDetection, CompilerPackage):
         "or as a project (with the compiler in use)",
     )
 
-    variant("offload", default=True, when="@19:", description="Build the Offload subproject")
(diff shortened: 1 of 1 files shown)


haampie:
Sigh, we do have specs like `~offload~lldb` which immediately run into the original problem again of concretization errors because `~lldb` implies the absense of `~offload`. I'll close for now :(

PR #3037, opened 2026-01-14
exago: Avoid cloning spack as submodule when building

### LINKED ISSUE #185: ExaGO should only clone select submodules in Spack
While I long had a dream to have `submodules=False` for a new release of ExaGO to avoid cloning Spack each time the source code is downloaded (see #38), I recently, through @larenspear's question on Spack Slack, learned that you can override the submodules function entirely to only clone select submodules - https://github.com/spack/spack-packages/blob/1b897a6a4d5dee4d6b08f22317dbcfa25265df0a/repos/spack_repo/builtin/packages/dakota/package.py#L11-L18.

Happy to PR this fix into the package myself, but figured it would be good to decide first if this approach should be taken, or if #38 should live on.

The fix would be simple enough:

```python
def submodules(package):
    submodules = []
    submodules.append("tpl/pybind11")
    submodules.append("tpl/spdlog")
    submodules.append("tpl/toml11")

    return submodules
```

Should speed up fresh ExaGO builds / CI a little on the Spack side...

cameronrutherford:
Closes https://github.com/pnnl/ExaGO/issues/185

I did this in the GitHub web editor so formatting might be off.

### DIFF
--- a/repos/spack_repo/builtin/packages/exago/package.py
+++ b/repos/spack_repo/builtin/packages/exago/package.py
@@ -10,6 +10,13 @@ from spack_repo.builtin.build_systems.rocm import ROCmPackage
 
 from spack.package import *
 
+def submodules(package):
+    submodules = []
+    submodules.append("tpl/pybind11")
+    submodules.append("tpl/spdlog")
+    submodules.append("tpl/toml11")
+
+    return submodules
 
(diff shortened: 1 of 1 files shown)


cameronrutherford:
@spackbot fix style

alalazo (approved):

PR #3038, opened 2026-01-14
Palace: accumulated fixes and updates for develop

Sbozzolo:
The most recent development version of Palace introduces multiple changes in the build system, reflected in this updated spack package.py

This commit also introduces various fixes for released versions and uncommon build configurations (e.g., static builds).


repos/spack_repo/builtin/packages/palace/package.py
```diff
@@ -129,14 +129,52 @@ class Palace(CMakePackage, CudaPackage, ROCmPackage):
...
+
+    with when("@0.16:"):
+        # +lapack means: use external lapack
+        depends_on(
+            "mfem+mpi+metis+lapack cxxstd=17 commit=bf1b25d82e232909e1d3aa1f416dbdae9931a550",
+            patches=[
+                "patch_mesh_vis_dev.diff",
+                "patch_par_tet_mesh_fix_dev.diff",
+                "patch_gmsh_parser_performance.diff",
+                "patch_race_condition_fix.diff",
```
cameronrutherford: This may become more complicated as we iterate.

Each mfem commit will require slightly different patchfiles in some cases, and so we will eventually have to have patchfiles maintained for each commit of mfem we depend on here.

I raised this on slack, but perhaps this will be solved in future by having `commit` as the name of the subfolder that we then store each commit itself withihn.

I don't think we need to resolve that in this PR until we start having different mfem commits we depend on.

cameronrutherford (approved):
Just a small comment. Since we have been testing this in our repo, LGTM, but a fresh set of eyes might catch something we missed.

repos/spack_repo/builtin/packages/palace/package.py
alalazo: Is there any chance that these changes can be upstreamed?

Sbozzolo: Most of them are already upstreamed, but Palace currently requires an older version that does not include them. We will be updating MFEM in Palace in the near future.

repos/spack_repo/builtin/packages/palace/package.py
```diff
@@ -129,14 +129,51 @@ class Palace(CMakePackage, CudaPackage, ROCmPackage):
     depends_on("metis~int64", when="~int64")
 
     conflicts("^hypre+int64", msg="Palace uses HYPRE's mixedint option for 64 bit integers")
-    depends_on("hypre@:2")  # MFEM 4.8 is incompatible with hypre v3+ and we haven't updated yet
+    depends_on("hypre@:2", when="@:0.15.0")
+    depends_on("hypre@3:")
```
cameronrutherford:
```suggestion
    depends_on("hypre@3:", when="@0.15.0:")
```

I think based on CI we need something like this. I might be wrong with how versions range syntax works...

Sbozzolo: It worked with 0.16 :)

cameronrutherford (approved):
Glad it was a small fix to get GitLab CI to pass!

repos/spack_repo/builtin/packages/palace/package.py
```diff
@@ -129,14 +129,51 @@ class Palace(CMakePackage, CudaPackage, ROCmPackage):
...
+                "patch_mesh_vis_dev.diff",
+                "patch_par_tet_mesh_fix_dev.diff",
+                "patch_gmsh_parser_performance.diff",
+                "patch_race_condition_fix.diff",
+            ],
+        )
+        depends_on("mfem+shared", when="+shared")
+        depends_on("mfem~shared", when="~shared")
+        depends_on("mfem+openmp", when="+openmp")
+        depends_on("mfem+threadsafe", when="+openmp")
```
alalazo, line 162: (Minor, can also be done in a following PR) You can group dependency constraints by their when clause:
```suggestion
        depends_on("mfem+openmp+threadsafe", when="+openmp")
```

Sbozzolo:
I added commit a01979b6714638fe4dcec11de57ef2fd35dfae3d because I found that patches were not correctly found on my local (dirty) spack installation. This solution seems more robust.

cameronrutherford:
> I added commit a01979b6714638fe4dcec11de57ef2fd35dfae3d because I found that patches were not correctly found on my local (dirty) spack installation. This solution seems more robust.

Nice catch! I assume this is referenced in the spack docs somewhere, but this is new to me.

cameronrutherford (approved):

hughcars (approved):

hughcars:
We are looking to get off of the manual commit sha, and will be updating that in a future PR to either the commit sha for mfem 4.9, or mfem 4.9 once it is merged to the recipe https://github.com/spack/spack-packages/pull/3040. At that point the patches can be viewed as patch release changes on top of each mfem minor version release, and this will become a bit saner to manage.

cameronrutherford (approved):

alalazo (approved):
Merging this, with the hope to reduce patches size in the future 😬

PR #1557, opened 2025-09-18
llvm-amdgpu provides fortran via amdflang

scheibelp:
Without this, `llvm-amdgpu` cannot be used as a fortran compiler, e.g. `spack spec foo %fortran=llvm-amdgpu`

`amdflang` is available for all versions that are described in the package.

PR #3040, opened 2026-01-15
mfem: add v4.9

v-dobrev:
Adding `mfem` v4.9 with a relevant patch and some tweaks.

This change is now reverted: ~~Also updates the `llvm-amdgpu` package to provide `"fortran"` starting with v6 instead of v7.~~

### DIFF
--- a/repos/spack_repo/builtin/packages/llvm_amdgpu/package.py
+++ b/repos/spack_repo/builtin/packages/llvm_amdgpu/package.py
@@ -67,7 +67,7 @@ class LlvmAmdgpu(CMakePackage, LlvmDetection, CompilerPackage):
     version("5.7.0", sha256="4abdf00b297a77c5886cedb37e63acda2ba11cb9f4c0a64e133b05800aadfcf0")
 
     provides("c", "cxx")
-    provides("fortran", when="@7.0:")
+    provides("fortran", when="@6:")
 
     variant(
         "rocm-device-libs",
--- a/repos/spack_repo/builtin/packages/llvm_amdgpu/package.py
+++ b/repos/spack_repo/builtin/packages/llvm_amdgpu/package.py
@@ -67,7 +67,7 @@ class LlvmAmdgpu(CMakePackage, LlvmDetection, CompilerPackage):
     version("5.7.0", sha256="4abdf00b297a77c5886cedb37e63acda2ba11cb9f4c0a64e133b05800aadfcf0")
 
     provides("c", "cxx")
-    provides("fortran", when="@6:")
+    provides("fortran", when="@7.0:")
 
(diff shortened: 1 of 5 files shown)


rbberger (changes requested):
@v-dobrev please see https://github.com/spack/spack-packages/pull/1557 and revert your llvm_amdgpu change. It's wrong. Even if this was the case, it shouldn't be mixed with a mfem change.

tzanio (dismissed):
Thank you, @v-dobrev !

v-dobrev:
> @v-dobrev please see #1557 and revert your llvm_amdgpu change. It's wrong. Even if this was the case, it shouldn't be mixed with a mfem change.

Thanks for the feedback, @rbberger! I know that when configured as external package, `llvm-amdgpu` actually can be configured to provide `"fortran"` (that's how I was testing), so I think it is natural to support that case. What do you think?

rbberger:
@v-dobrev the ROCm stack maintainers explicitly removed that option for older versions to be consistent with their packaging at the time. You can appeal to them. That being said with externals you can also add a dependency to the "correct" Fortran provider for `:6`.

afzpatel:
Spack-built llvm-amdgpu doesn't provide Fortran prior to ROCm 7.0, so adding `provides("fortran", when="@6:")` is incorrect.  

I prefer the approach taken in this PR, although there are a few CI build failures: https://github.com/spack/spack-packages/pull/2499

v-dobrev:
@rbberger, I reverted the change in `llvm-amdgpu` as you requested.

As @afzpatel pointed out, #2499 is the better solution.

rbberger (dismissed):

v-dobrev:
@rbberger, what happened here? Why did you force-push to my branch?

rbberger:
@v-dobrev rebased onto latest develop to rerun pipeline (it was still failing).

v-dobrev:
> @v-dobrev rebased onto latest develop to rerun pipeline (it was still failing).

Okay, I see. I don't like re-basing my branches for various reasons, so please next time either use merge update or let me know and I'll do it.

v-dobrev:
One of the failures in the Gitlab CI pipelines is that `visit` is using c++14 standard and `mfem` now requires c++17.

@cyrush, should we just limit `visit` to use `mfem` <= 4.8 or can we tell `visit` to switch to c++17 if using `mfem` >= 4.9?

cyrush:
Limiting the current visit release to mfem 4.8 sounds like the easiest path.

When we release our next visit version we may move to c++17

rbberger:
@v-dobrev please go ahead and add that limit.

rbberger (approved):

PR #3042, opened 2026-01-15
py-keras: add v3.13.1

adamjstewart:
Honestly struggling to keep up with which versions of which deps are supported when. There are several different files to check and they all disagree. Based on https://github.com/keras-team/keras/issues/21926, I don't think we need to be so strict with pinned dependencies. @eugeneswalker this may help bring macOS CI back online.

https://github.com/keras-team/keras/releases/tag/v3.13.1


eugeneswalker (approved):
Thanks!

PR #3050, opened 2026-01-16
git-lfs: add 3.7.1

mathomp4:
git-lfs older than 3.7.1 apparently has a bad security issue:

https://github.com/git-lfs/git-lfs/security/advisories/GHSA-6pvw-g552-53c5

and per that page:

> This problem exists in all versions since 0.5.2 and is patched in v3.7.1. All users should upgrade to v3.7.1.

ETA: Also. Should we *deprecate* every version but 3.7.1 given that CVE?

### DIFF
--- a/repos/spack_repo/builtin/packages/git_lfs/package.py
+++ b/repos/spack_repo/builtin/packages/git_lfs/package.py
@@ -27,6 +27,7 @@ class GitLfs(MakefilePackage):
 
     license("MIT")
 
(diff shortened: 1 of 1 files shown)


sethrj:
>  if symbolic or hard links exist which collide with the paths of files tracked by Git LFS.

This seems pretty minor, I would just stick with adding the new version

sethrj (approved):
verified shasum

PR #3052, opened 2026-01-16
py-dask: fix bounds of python dep

thomas-bouvier:
Python version bounds seem important to Dask. For instance, Python 3.14 breaks `py-dask@2025.7.0`.

### DIFF
--- a/repos/spack_repo/builtin/packages/py_dask/package.py
+++ b/repos/spack_repo/builtin/packages/py_dask/package.py
@@ -49,7 +49,7 @@ class PyDask(PythonPackage):
     depends_on("python@3.8:", type=("build", "run"), when="@:2023.4.1")
 
     # python@3.14 breaks py-dask@:2025.7.0
-    depends_on("py-python@:3.13", type=("build", "run"), when="@:2025.7.0")
+    depends_on("python@:3.13", type=("build", "run"), when="@:2025.7.0")
 
     depends_on("py-setuptools", type="build")
     depends_on("py-setuptools@62.6:", type="build", when="@2023.4.1:")
--- a/repos/spack_repo/builtin/packages/py_dask/package.py
+++ b/repos/spack_repo/builtin/packages/py_dask/package.py
@@ -49,7 +49,7 @@ class PyDask(PythonPackage):
     depends_on("python@3.8:", type=("build", "run"), when="@:2023.4.1")
 
     # python@3.14 breaks py-dask@:2025.7.0
-    depends_on("python@:3.13", type=("build", "run"), when="@:2025.7.0")
+    depends_on("python@:3.13", type=("build", "run"), when="@:2025.12")
 
     depends_on("py-setuptools", type="build")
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/py_dask/package.py
```diff
@@ -44,9 +44,10 @@ class PyDask(PythonPackage):
 
     conflicts("~array", when="@2023.8: +dataframe", msg="From 2023.8, +dataframe requires +array")
 
+    depends_on("python@3.10:3.13", type=("build", "run"), when="@2024.12.0:2025.7.0")
+    depends_on("python@3.10:3.12", type=("build", "run"), when="@2024.8.1:2024.11.2")
+    depends_on("python@3.9:3.12", type=("build", "run"), when="@2024.7.1")
     depends_on("python@3.8:", type=("build", "run"), when="@:2023.4.1")
-    depends_on("python@3.9:", type=("build", "run"), when="@2024.7.1")
-    depends_on("python@3.10:", type=("build", "run"), when="@2024.8.1:")
```
wdconinc: Since lower and upper limits are typically changed for independent reasons, it may make more sense to specify the lower and upper limits independently as well instead of tying them together.

thomas-bouvier: Fixed in last commit!

repos/spack_repo/builtin/packages/py_dask/package.py
```diff
@@ -44,9 +44,12 @@ class PyDask(PythonPackage):
...
     conflicts("~array", when="@2023.8: +dataframe", msg="From 2023.8, +dataframe requires +array")
 
-    depends_on("python@3.8:", type=("build", "run"), when="@:2023.4.1")
-    depends_on("python@3.9:", type=("build", "run"), when="@2024.7.1")
     depends_on("python@3.10:", type=("build", "run"), when="@2024.8.1:")
+    depends_on("python@3.9:", type=("build", "run"), when="@2024.7.1")
+    depends_on("python@3.8:", type=("build", "run"), when="@:2023.4.1")
+
+    # python@3.14 breaks py-dask@:2025.7.0
+    depends_on("python@:3.13", type=("build", "run"), when="@:2025.7.0")
```
wdconinc:
```suggestion
    depends_on("python@:3.13", type=("build", "run"), when="@:2025.12")
```
https://pypi.org/project/dask/2026.1.0/ is first one that marks 3.14 as supported.

thomas-bouvier: It’s unclear to me whether we should evaluate these bounds based on their original context outside Spack or consider only the versions integrated into Spack, which currently max out at 2025.7.0.

wdconinc: If we are going to add version bounds and we know what the correct value is, we should add the correct value even if the version is not (yet) in spack.

haampie (approved):

PR #3056, opened 2026-01-18
Update TFEL and MGIS packages

thelfer:


repos/spack_repo/builtin/packages/mgis/package.py
```diff
@@ -37,45 +38,85 @@ class Mgis(CMakePackage):
 
     # released version
     version(
-        "3.0.1",
-        sha256="fb9a7f5008a43c70bdb1c4b80f32f7fd3e4274c912b93c36af7011d3c4f93039",
+        "3.1.0",
+        sha256="61afae1a367dbb150b24ca85f042efb15a77184a54a746f11c08d9b7cb9e94f3",
+        url="https://github.com/thelfer/MFrontGenericInterfaceSupport/archive/MFrontGenericInterfaceSupport-3.1.tar.gz",
         preferred=True,
```
alalazo: No need to have `preferred=True` since v3.1.0 would be preferred anyhow

repos/spack_repo/builtin/packages/mgis/package.py
```diff
@@ -37,45 +38,85 @@ class Mgis(CMakePackage):
...
-    version("2.0", sha256="cb427d77f2c79423e969815b948a8b44da33a4370d1760e8c1e22a569f3585e2")
+    version("2.2.0", sha256="b3776d7b3a534ca626525a42b97665f7660ae2b28ea57b3f53fd7e8538da1ceb")
+    version("2.1.1", sha256="3fb5500cdb855543403028e28b6418913b3067ab2509d254022a234ea59ed4a8")
+    version("2.1.0", sha256="f5b556aab130da0c423f395fe4c35d6bf509dd8fc958242f2e37ea788464aea9")
+    version("2.0.1", sha256="9850aa177d4a6e43faa1434968d2daad65aab2a9b7d64fa4eea51bd25c3d2c5c")
+    version(
+        "2.0.0",
+        sha256="cb427d77f2c79423e969815b948a8b44da33a4370d1760e8c1e22a569f3585e2",
+        url="https://github.com/thelfer/MFrontGenericInterfaceSupport/archive/MFrontGenericInterfaceSupport-2.0.tar.gz",
+    )
```
alalazo: Why are we retroactively changing these versions (e.g. from v2.2 to v2.2.0)?

repos/spack_repo/builtin/packages/tfel/package.py
```diff
@@ -51,26 +52,32 @@ class Tfel(CMakePackage):
 
     # released versions
     version(
-        "5.0.1",
-        sha256="820b2f9d54e237b2c2d9d6f06aaf7d4a1d3f34fb373e6549bee4fd8b55ecfad1",
+        "5.1.0",
+        sha256="1afd98200de332e97e86d109ce0e1aaa8f18cc6c6c81daec3218809509cdfad7",
         preferred=True,
```
alalazo: Same as above

alalazo (changes requested):
See comments, in particular about changing the versions. Also, I noticed that:
```
$ cat repos/spack_repo/builtin/packages/tfel/package.py | grep version\( | wc -l
92
```
Is it worth starting to deprecated older versions, and maybe keep non-deprecated only the latest patch release of the major that are not under active development?

repos/spack_repo/builtin/packages/mgis/package.py
thelfer: done

repos/spack_repo/builtin/packages/tfel/package.py
thelfer: done

repos/spack_repo/builtin/packages/mgis/package.py
thelfer: Semantic versioning did not exist/was not as popular as today when a I started TFEL twenty years ago. Changing versioning on the TFEL or MGIS side would be cumbersome for my end-users.

But on the `spack` side, I feel like that a bit of consistency would not hurt. If you are afraid of backward compatibility, shall we add both `2.0` and `2.0.0` ? (IMHO, that's ugly)

thelfer:
> See comments, in particular about changing the versions. Also, I noticed that:
> 
> ```
> $ cat repos/spack_repo/builtin/packages/tfel/package.py | grep version\( | wc -l
> 92
> ```
> 
> Is it worth starting to deprecated older versions, and maybe keep non-deprecated only the latest patch release of the major that are not under active development?

Definitively. I keep all the versions for two reasons:

- `TFEL` is used in safety critical studies, so we must be able to roll-back to previous versions in case of trouble
- We may want to benchmark different versions

> maybe keep non-deprecated only the latest patch release of the major that are not under active development?

That's a very nice suggestion. Let me do this

thelfer:
@alalazo It is better now ?

thelfer:
@alalazo Is there something more I can do ?

repos/spack_repo/builtin/packages/mgis/package.py
```diff
@@ -36,46 +37,120 @@ class Mgis(CMakePackage):
     version("rliv-1.0", branch="rliv-1.0")
 
     # released version
+    version(
+        "3.1.0",
+        sha256="61afae1a367dbb150b24ca85f042efb15a77184a54a746f11c08d9b7cb9e94f3",
+        url="https://github.com/thelfer/MFrontGenericInterfaceSupport/archive/MFrontGenericInterfaceSupport-3.1.tar.gz",
```
alalazo, line 43: You can group the non-deprecated versions at the top, then use a context manager like:
```python
    ...
```
for the deprecated versions. That will save a few lines. You can look at `zig` or `clingo` to check a package doing the same thing.

repos/spack_repo/builtin/packages/mgis/package.py
alalazo: Up to you, but I would respect the versions as they are declared upstream.

repos/spack_repo/builtin/packages/mgis/package.py
```diff
@@ -36,46 +37,120 @@ class Mgis(CMakePackage):
...
+        deprecated=True,
+    )
 
     # variants
     variant("c", default=True, description="Enables c bindings")
     variant("fortran", default=True, description="Enables fortran bindings")
     variant("python", default=True, description="Enables python bindings")
     variant("static", default=False, description="Enables static libraries")
 
+    with when("@3.1:") or when("rliv-3.1:"):
```
alalazo: 1. `or` doesn't work with `when`
2. You cannot check a version without the `@` sigil (second when)
3. `rliv-3.1:` contains all numeric versions, so it's likely not what you meant 🙂 

I think you want to have something like:

```suggestion
    with when("@3.1:,rliv-3.1"):
```

repos/spack_repo/builtin/packages/mgis/package.py
```diff
@@ -104,7 +179,12 @@ class Mgis(CMakePackage):
 
     def patch(self):
         """Fix the test suite to use the PYTHONPATH provided by the spack buildenv"""
-        filter_file("tests/;", "tests:", "bindings/python/tests/CMakeLists.txt")
+        filter_file("tests/;", "tests:")
```
alalazo: Without a file this is a no-op

repos/spack_repo/builtin/packages/mgis/package.py
```diff
@@ -104,7 +179,12 @@ class Mgis(CMakePackage):
...
+        if "+python" in self.spec:
```
alalazo: If you use inverse logic, you save one indentation:

```suggestion
        if not self.spec.satisfies("+python"):
            return
```

repos/spack_repo/builtin/packages/mgis/package.py
```diff
@@ -104,7 +179,12 @@ class Mgis(CMakePackage):
...
+            if when("@3.0.1") or when("@rliv-3.0"):
+                filter_file("tests/;", "bindings/python/boost/tests/CMakeLists.txt")
```
alalazo: The call to `filter_file` seems wrong. If I look at the implementation this is a no-op: https://spack.readthedocs.io/en/latest/package_api.html#spack.package.filter_file

repos/spack_repo/builtin/packages/mgis/package.py
```diff
@@ -120,6 +200,15 @@ def cmake_args(self):
...
         args.append("-Denable-doxygen-doc=OFF")
+        args.append("-DTFEL_DIR={0}/share/tfel/cmake".format(self.spec["tfel"].prefix))
+        args.append("-Denable-parallel-stl-algorithms=OFF")
+
+        if "+openmp" in self.spec:
+            args.append("-Denable-openmp=ON")
+        if "~mgis-function" in self.spec:
+            args.append("-Denable-mgis-function=OFF")
+        if "+exception" in self.spec:
+            args.append("-Denable-exception=ON")
```
alalazo: Here you can use `self.define_from_variant`

repos/spack_repo/builtin/packages/tfel/package.py
alalazo, line 1: Same comment on:
```python
    ...
```

alalazo (changes requested):

repos/spack_repo/builtin/packages/mgis/package.py
thelfer: done. This is so much better. Thanks !

repos/spack_repo/builtin/packages/mgis/package.py
thelfer: done !

repos/spack_repo/builtin/packages/tfel/package.py
thelfer: done !

repos/spack_repo/builtin/packages/mgis/package.py
thelfer: I removed the `patch` method

repos/spack_repo/builtin/packages/mgis/package.py
thelfer: done

repos/spack_repo/builtin/packages/mgis/package.py
thelfer: I remove the `patch` method, I am not sure of what it does

repos/spack_repo/builtin/packages/mgis/package.py
thelfer: done !

thelfer:
@alalazo I think that I have taken into account your advices. Thanks for your time and patience.

repos/spack_repo/builtin/packages/mgis/package.py
```diff
@@ -37,45 +38,87 @@ class Mgis(CMakePackage):
...
+            url="https://github.com/thelfer/MFrontGenericInterfaceSupport/archive/MFrontGenericInterfaceSupport-1.0.tar.gz",
+        )
 
     # variants
     variant("c", default=True, description="Enables c bindings")
     variant("fortran", default=True, description="Enables fortran bindings")
     variant("python", default=True, description="Enables python bindings")
     variant("static", default=False, description="Enables static libraries")
 
+    with when("@3.1:,rliv-3.1:"):
```
alalazo:
```suggestion
    with when("@3.1:,rliv-3.1"):
```

repos/spack_repo/builtin/packages/mgis/package.py
alalazo: Before it was looking for files with a `tests/;` match and was substituting that with `tests:` (i.e. removing the slash and turning `;` into`:`)

alalazo (approved):

PR #3061, opened 2026-01-19
ninja: Fix supports_jobserver for version 1.13

vbrunini:
So that builds using ninja will correctly use the jobserver when it is enabled in parallel builds.

### DIFF
--- a/repos/spack_repo/builtin/packages/ninja/package.py
+++ b/repos/spack_repo/builtin/packages/ninja/package.py
@@ -38,7 +38,10 @@ class Ninja(Package):
     version("1.7.2", sha256="2edda0a5421ace3cf428309211270772dd35a91af60c96f93f90df6bc41b16d9")
     version("1.6.0", sha256="b43e88fb068fe4d92a3dfd9eb4d19755dae5c33415db2e9b7b61b4659009cde7")
     version(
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/ninja/package.py
```diff
@@ -112,5 +112,5 @@ def setup_dependent_package(self, module, dspec):
         module.ninja = MakeExecutable(
             which_string(name, path=[self.spec.prefix.bin], required=True),
             jobs=determine_number_of_jobs(parallel=dspec.package.parallel),
-            supports_jobserver=self.spec.version == ver("kitware"),
+            supports_jobserver=self.spec.satisfies("@1.13:"),
```
haampie: This does not include `kitware`, but I also don't think anybody should be using that fork now.
```suggestion
            supports_jobserver=self.spec.satisfies("@kitware,1.13:"),
```

haampie: Can you deprecate `kitware` too?

vbrunini: Ah, I forgot that it's only the "special" named versions that are treated as infinity versions. Yeah I can fix the check and deprecate `kitware`.

haampie (approved):

PR #3064, opened 2026-01-20
acts: remove "build" type from some dependencies

stephenswat:
The ACTS package currently has a problem where, upon installation, Spack may select different Python versions for the ACTS build itself and for the installation of some of its Python package dependencies. Because package installation are bound to a specific Python version, this then causes an import error and, in turn, a failure in the ACTS build. This commit works around this issue by removing the "build" dependency type specification from Python and several Python packages.

### DIFF
--- a/repos/spack_repo/builtin/packages/acts/package.py
+++ b/repos/spack_repo/builtin/packages/acts/package.py
@@ -468,18 +468,24 @@ class Acts(CMakePackage, CudaPackage):
     depends_on("python", when="+python")
     depends_on("python@3.8:", when="+python @19.11:19")
     depends_on("python@3.8:", when="+python @21:")
-    depends_on("python@3.12:", when="@44:", type="build")
-    depends_on("py-numpy @2.2", when="@44:", type="build")
+    # NOTE: Python and many of the Python packages we depend on are build
+    # dependencies only, but marking them as such allows Spack to pick up
+    # different Python versions for e.g. the ACTS build and the numpy
+    # installation which, in turn, causes the ACTS build to fail. Until a more
+    # robust solution is available we pretend that these packages are also
+    # run- and link-time dependencies.
+    depends_on("python@3.12:", when="@44:")
+    depends_on("py-numpy @2.2", when="@44:")
     depends_on("py-onnxruntime@:1.12", when="+onnx @:23.2")
     depends_on("py-onnxruntime@1.12:", when="+onnx @23.3:")
-    depends_on("py-particle @0.24", when="@44:", type="build")
+    depends_on("py-particle @0.24", when="@44:")
     depends_on("py-pybind11 @2.6.2:", when="+python @18:")
     depends_on("py-pybind11 @2.13.1:", when="+python @36:")
     depends_on("py-pytest", when="+python +unit_tests")
-    depends_on("py-setuptools", when="@44:44.1.0", type="build")
-    depends_on("py-sympy @1.13", when="@44:", type="build")
+    depends_on("py-setuptools", when="@44:44.1.0")
+    depends_on("py-sympy @1.13", when="@44:")
     # TODO: Clarify version on next release
-    depends_on("py-hatchling", when="@44.1.1:", type="build")
+    depends_on("py-hatchling", when="@44.1.1:")
     depends_on("py-torch", when="+gnn +torch")
 
     with when("+root"):


tldahlgren (approved):

tldahlgren:
@stephenswat @wdconinc Any objections to my merging this PR without waiting for further comment or approval?

wdconinc:
> @stephenswat @wdconinc Any objections to my merging this PR without waiting for further comment or approval?

This was flagged for discussion by @stephenswat in https://spackpm.slack.com/archives/C08Q62S7XEX/p1768903203166009. I would keep it open in case anyone there wants to comment and suggest another solution, or until @stephenswat indicates it's indeed good as a short-term fix (I'm ok with it).

stephenswat:
At least in the short term this fix works for me; I suggest we can look for a better long-term fix, which better matches the semantics of these dependencies, in the future.

stephenswat:
Tagging @paulgessinger who reported this issue.

paulgessinger:
Can we get this in?

wdconinc (approved):

wdconinc:
@stephenswat Since this is arguably a bug, would you be able to report it (at spack/spack) for tracking and resolving?

repos/spack_repo/builtin/packages/acts/package.py
```diff
@@ -468,18 +468,24 @@ class Acts(CMakePackage, CudaPackage):
...
     depends_on("python@3.8:", when="+python @19.11:19")
     depends_on("python@3.8:", when="+python @21:")
-    depends_on("python@3.12:", when="@44:", type="build")
-    depends_on("py-numpy @2.2", when="@44:", type="build")
+    # NOTE: Python and many of the Python packages we depend on are build
+    # dependencies only, but marking them as such allows Spack to pick up
+    # different Python versions for e.g. the ACTS build and the numpy
+    # installation which, in turn, causes the ACTS build to fail. Until a more
+    # robust solution is available we pretend that these packages are also
+    # run- and link-time dependencies.
```
wdconinc, line 476: @stephenswat Is there a PR in acts to track here for the "more robust solution"? This approach is starting to limit us in terms of upgrading numpy independent of when acts may choose to upgrade numpy.

paulgessinger: We don't actually depend on any specific version of numpy in ACTS. That's the fundamental problem here. 

In Python land you don't generally put in upper limits in Python version constraints. In the internal uv based build, the numpy version we happen to have locked to is perfectly compatible with newer python versions. 

However, in Spack land, it is common to specify upper limits. The solution in my opinion is to switch off supplying numpy externally through Spack here, or not bounding the version of numpy. 

The whole point of me adding this system that installs these codegen dependencies during the build using uv is to shield the internal version used from external build environment. 

For the Spack package, we ended up suppressing this mechanism and this breaks here.

wdconinc: I don't see shielding the versions used for codegen from the rest of the environment as being in conflict with what spack does, when you provide these build dependencies as `type="build"`. With type=build you allow spack to pick whatever versions, regardless of what is constrained elsewhere in the run-time tree.

The reference to "causes the ACTS build to fail" in the comment seems to indicate that there is some crosstalk from the codegen that leaves observable effects from what the internal versions are (symbols in binaries). Is that a correct interpretation?

paulgessinger: I don't believe so. There should be no trace whatsoever in the installed binaries. We generate cpp code from Python that does not depend on any Python libraries. 

Iirc the problem before was that numpy would be installed for some Python version, and the ACTS codegen step would use a different Python version (both from Spack), which numpy was not available for.

wdconinc: > Iirc the problem before was that numpy would be installed for some Python version, and the ACTS codegen step would use a different Python version (both from Spack), which numpy was not available for.

Ok, I found the slack messages about this. It seems it was possibly a concretization bug in spack, so let me see if that might have been fixed...

wdconinc: It turns out (slack) that the issue that caused the build dependency concretization behavior here may have been solved in https://github.com/spack/spack/pull/51915. That PR was backported into spack v1.0.3, released today. Can we see if that update indeed resolves this issue and consider making these dependencies build-type again?

PR #3066, opened 2026-01-20
mochi-margo: added version 0.22.0 and missing cmake args

mdorier:
This PR adds margo version 0.22.0 and adds missing "ENABLE_PLUMBER" argument in cmake args when `+hwloc` is specified. Also ENABLE_PLUMBER requires libfabric, which mercury might not have been built with, so an explicit dependency on libfabric is required.


tldahlgren (approved):
Confirmed the sha256.

@mdorier @carns @fbudin69500 Any objections to merging this PR as is?

carns:
looks good to me :+1:

PR #2916, opened 2025-12-30
Fixes for CARE and dependencies

dtaller:

PR #2856, opened 2025-12-19
Synchronize RADIUSS packages with all local additions

davidbeckingsale:

PR #3067, opened 2026-01-20
Synchronize RADIUSS packages with all local additions and add 2025.12 releases.

adrienbernede:
(Supersedes #2856 and adds latest releases.)

This branch aims at pushing to spack-packages the remaining changes that were maintained locally in llnl/radiuss-spack-configs. The goal being to now maintain the packages directly in spack/spack-packages.

- [X] cleaned deprecated workaround in the local packages
- [X] added the latest releases (2025.12)
- [X] vetted both in Spack CI (here) and in each projects CI

> Questions that should not delay the review but should trigger discussions:
> - Do we need to create a cxxstd variant for Umpire and RAJAPerf packages (see #2916) ?
> - Do we need to add the option to CHAI as well ?


repos/spack_repo/builtin/packages/blt/package.py
```diff
@@ -76,6 +76,13 @@ class Blt(Package):
     version("0.2.5", sha256="3a000f60194e47b3e5623cc528cbcaf88f7fea4d9620b3c7446ff6658dc582a5")
     version("0.2.0", sha256="c0cadf1269c2feb189e398a356e3c49170bc832df95e5564e32bdbb1eb0fa1b3")
 
+    # https://github.com/google/googletest/pull/4798
```
alecbcs, line 79:
```suggestion
    # https://github.com/google/googletest/pull/4798
    # Make BLT compatible with OpenAPI 2025.2
```

davidbeckingsale (approved):

repos/spack_repo/builtin/packages/umpire/package.py
```diff
@@ -231,7 +245,7 @@ class Umpire(CachedCMakePackage, CudaPackage, ROCmPackage):
     variant("backtrace", default=False, description="Enable backtrace tools")
     variant("dev_benchmarks", default=False, description="Enable developer benchmarks")
     variant("device_alloc", default=False, description="Enable DeviceAllocator")
-    variant("werror", default=False, description="Enable warnings as errors")
+    variant("werror", default=True, description="Enable warnings as errors")
```
alecbcs: Why are we enabling this variant by default now?

alecbcs (commented):
A couple quick questions and suggestions. Otherwise looks pretty good to me.

@davidbeckingsale @adayton1 @kab163 as the package maintainers do you all have any thoughts before we merge?

repos/spack_repo/builtin/packages/umpire/package.py
adrienbernede: @davidbeckingsale, although relevant for our CI setup it is likely too restrictive for general usage, and a potential breaking change in situations we don't cover. Do you agree?

davidbeckingsale: Agree!

davidbeckingsale:
```suggestion
    variant("werror", default=False, description="Enable warnings as errors")
```

davidbeckingsale (approved):

repos/spack_repo/builtin/packages/caliper/package.py
```diff
@@ -95,7 +97,9 @@ class Caliper(CachedCMakePackage, CudaPackage, ROCmPackage):
 
     conflicts("+rocm+cuda")
     # Legacy nvtx is only supported until cuda@12.8, newer cuda only provides nvtx3.
-    conflicts("^cuda@12.9:", "@:2.12.1")
+    conflicts("^cuda@12.9:", "@:2.13.1")
```
alecbcs:
```suggestion
    depends_on("cuda@:12.8", when="@:2.13.0 ^cuda")
```

We can make this easier to read / understand if we apply the constraints with a `depends_on()` directive with the qualifier of `^cuda` to limit the constraint to only when the package depends on `cuda`.

repos/spack_repo/builtin/packages/caliper/package.py
```diff
@@ -95,7 +97,9 @@ class Caliper(CachedCMakePackage, CudaPackage, ROCmPackage):
...
+    # rocprofiler-sdk is only supported since rocm@6.2.4.
+    conflicts("^llvm-amdgpu@:6.2.3", "@2.14: +rocm")
```
alecbcs:
```suggestion
    depends_on("llvm-amdgpu@6.2.4:", when="@2.1.4: +rocm")
```
Same thing here. I think swapping this to a `depends_on()` directive helps with readability.

alecbcs (commented):
One more quick suggestion. Didn't realize we could express `depends_on()` directives like this with `when="^dependency"` but it actually works very nicely and I find to be more readable.

adrienbernede:
@alecbcs I agree that `depends_on` is preferred. IIRC, it is also better in terms of performance since conflict are evaluated later on.
However, I will need to test those changes just to make sure they don't break concretization for any reason.

repos/spack_repo/builtin/packages/caliper/package.py
adrienbernede: I think the equivalent is
`depends_on("cuda@:12.8", when="@:2.13.1 ^cuda")`
Because the conflict also applied to 2.13.1, correct ?

repos/spack_repo/builtin/packages/caliper/package.py
adrienbernede:
```suggestion
    depends_on("llvm-amdgpu@6.2.4:", when="@2.14: +rocm")
```

repos/spack_repo/builtin/packages/caliper/package.py
adrienbernede:
```suggestion
    depends_on("cuda@:12.8", when="@:2.13.1 ^cuda")
```

davidbeckingsale (approved):

repos/spack_repo/builtin/packages/caliper/package.py
```diff
@@ -95,7 +97,9 @@ class Caliper(CachedCMakePackage, CudaPackage, ROCmPackage):
 
     conflicts("+rocm+cuda")
     # Legacy nvtx is only supported until cuda@12.8, newer cuda only provides nvtx3.
-    conflicts("^cuda@12.9:", "@:2.12.1")
+    depends_on("cuda@:12.8", when="@:2.13.1 ^cuda")
```
alecbcs:
```suggestion
    depends_on("cuda@:12.8", when="@:2.13.1 +cuda")
```

I'm not sure why the audit is tagging this here, but if we change to `+cuda` it works the same. I'll dig into the audit check and figure it out after this PR.

repos/spack_repo/builtin/packages/caliper/package.py
```diff
@@ -154,11 +158,54 @@ def initconfig_hardware_entries(self):
...
+                cuda_archs = ";".join(spec.variants["cuda_arch"].value)
+                entries.append(cmake_cache_string("CMAKE_CUDA_ARCHITECTURES", cuda_archs))
+
+            # gcc-toolchain support
+            gcc_toolchain_regex = re.compile(".*gcc-toolchain.*")
+            using_toolchain = list(
+                filter(gcc_toolchain_regex.match, spec.compiler_flags["cxxflags"])
+            )
+            if using_toolchain:
+                cuda_flags.append("-Xcompiler {}".format(using_toolchain[0]))
```
alecbcs:
```suggestion
                cuda_flags.append(f"-Xcompiler {using_toolchain[0]}")
```

Swapping to an f-string here to make it a bit more readable.

repos/spack_repo/builtin/packages/caliper/package.py
```diff
@@ -154,11 +158,54 @@ def initconfig_hardware_entries(self):
...
+
+            if using_toolchain:
+                gcc_prefix = using_toolchain[0]
+                entries.append(
+                    cmake_cache_string("HIP_CLANG_FLAGS", "--gcc-toolchain={0}".format(gcc_prefix))
+                )
+                entries.append(
+                    cmake_cache_string(
+                        "CMAKE_EXE_LINKER_FLAGS",
+                        hip_link_flags + " -Wl,-rpath={0}/lib64".format(gcc_prefix),
```
alecbcs:
```suggestion
                        hip_link_flags + f" -Wl,-rpath={gcc_prefix}/lib64",
```

Same here.

repos/spack_repo/builtin/packages/caliper/package.py
```diff
@@ -154,11 +158,54 @@ def initconfig_hardware_entries(self):
...
+                    )
+                )
+            else:
+                entries.append(
+                    cmake_cache_string(
+                        "CMAKE_EXE_LINKER_FLAGS", "-Wl,-rpath={0}/llvm/lib/".format(rocm_root)
```
alecbcs:
```suggestion
                        "CMAKE_EXE_LINKER_FLAGS", f"-Wl,-rpath={rocm_root}/llvm/lib/"
```

And here.

alecbcs (approved):
Looks good to me. There are a few f-strings I'd like to fix, but will do that in a follow up PR.

davidbeckingsale (approved):

PR #3071, opened 2026-01-20
binutils: add v2.45.1

alalazo:


tldahlgren (approved):
Confirmed the version sha256.

PR #3073, opened 2026-01-20
updated Adiak to be a CachedCMakePackage 

LinaMuryanto:
Updating Adiak to be CachedCMakePackage so that it can find correct mpi executable.


repos/spack_repo/builtin/packages/adiak/package.py
```diff
@@ -1,15 +1,14 @@
-# Copyright Spack Project Developers. See COPYRIGHT file for details.
+# Copyright 2013-2021 Lawrence Livermore National Security, LLC and other
+# Spack Project Developers. See the top-level COPYRIGHT file for details.
```
tldahlgren: The copyright form is out of date. Please use/restore the newer one:

```suggestion
# Copyright Spack Project Developers. See COPYRIGHT file for details.
```

repos/spack_repo/builtin/packages/adiak/package.py
```diff
@@ -21,51 +20,41 @@ class Adiak(CMakePackage):
 
     variant("mpi", default=True, description="Build with MPI support")
     variant("shared", default=True, description="Build dynamic libraries")
-    variant("python", default=False, when="@0.5.0:", description="Build Python bindings")
```
tldahlgren, line 24: Are python bindings no longer supported? If not since which version?

repos/spack_repo/builtin/packages/adiak/package.py
```diff
@@ -21,51 +20,41 @@ class Adiak(CMakePackage):
...
 
     license("MIT")
-
```
tldahlgren, line 27: Is there a reason to drop the `master` branch?

repos/spack_repo/builtin/packages/adiak/package.py
```diff
@@ -21,51 +20,41 @@ class Adiak(CMakePackage):
...
     def cmake_args(self):
         args = []
-        if self.spec.satisfies("+mpi"):
-            args.append(f"-DMPI_CXX_COMPILER={self.spec['mpi'].mpicxx}")
-            args.append(f"-DMPI_C_COMPILER={self.spec['mpi'].mpicc}")
-            args.append("-DENABLE_MPI=ON")
-        else:
-            args.append("-DENABLE_MPI=OFF")
 
+        if self.spec.satisfies("+mpi"):
```
tldahlgren: Are the other settings added when this variant is enabled incorrect or no longer relevant? If the latter, from what version on?

tldahlgren (changes requested):
The new copyright needs to be restored. There are a number of options that have been removed related to `python` and `mpi`. Were they the source of issues or are they not relevant from some version on?

@daboehme @mplegendre Pinging you here since you can't be added directly as Reviewers.

repos/spack_repo/builtin/packages/adiak/package.py
LinaMuryanto: will add this back. My mistakes. I think I was not comparing with the latest develop branch.

daboehme: The Python bindings are supported and the variant should remain

repos/spack_repo/builtin/packages/adiak/package.py
daboehme: The master branch should also remain

repos/spack_repo/builtin/packages/adiak/package.py
daboehme: I'm not sure why this was changed. The python and mpi variants are independent of each other and both should remain. In both cases we want the existing behavior (i.e. add `-DENABLE_MPI=On` if and only if the mpi variant is active and `-DENABLE_PYTHON_BINDINGS=On` if and only if the python variant is active).

repos/spack_repo/builtin/packages/adiak/package.py
```diff
@@ -21,51 +20,41 @@ class Adiak(CMakePackage):
...
             args.append("-DBUILD_SHARED_LIBS=OFF")
 
-        if self.spec.satisfies("+python"):
-            args.append("-DENABLE_PYTHON_BINDINGS=ON")
-            pybind11_cmake = os.path.join(
-                self.spec["py-pybind11"].prefix, "pybind11", "share", "cmake", "pybind11"
-            )
-            args.append(f"-Dpybind11_DIR={pybind11_cmake}")
-
-        args.append("-DENABLE_TESTS=OFF")
```
daboehme, line 70: Let's also keep the `-DENABLE_TESTS=Off` arg.

daboehme (changes requested):
I'm generally fine with moving to CachedCMakePackage if that helps with the MPI support. We would like to keep the python variant though.

repos/spack_repo/builtin/packages/adiak/package.py
LinaMuryanto: @daboehme @tldahlgren 
my mistakes, I did not compare to the package.py in the latest 'develop' of spack/spack-packages.
Updating it and testing it again..Will let you know when ready for full review again. Thanks

repos/spack_repo/builtin/packages/adiak/package.py
LinaMuryanto: @daboehme I am adding 'tests' variant.

LinaMuryanto:
@daboehme when adiak is built with +mpi+python, does it depend on mpi4py?

daboehme:
> @daboehme when adiak is built with +mpi+python, does it depend on mpi4py?

Yes in that case it depends on mpi4py

LinaMuryanto:
> > @daboehme when adiak is built with +mpi+python, does it depend on mpi4py?
> 
> Yes in that case it depends on mpi4py

I have tested my changes:
`
    Start 1: blt_gtest_smoke
    Start 2: blt_mpi_smoke
    Start 3: test_adiak
    Start 4: test_python_serial
    Start 5: test_python_mpi
1/5 Test #1: blt_gtest_smoke ..................   Passed    0.02 sec
2/5 Test #3: test_adiak .......................   Passed    0.12 sec
3/5 Test #4: test_python_serial ...............   Passed    0.18 sec
4/5 Test #5: test_python_mpi ..................   Passed    1.81 sec
5/5 Test #2: blt_mpi_smoke ....................   Passed    1.86 sec

100% tests passed, 0 tests failed out of 5
`

@daboehme  with +mpi and +python, if it depends on mpi4py, should we have depends_on("mpi4py") in the package.py?
but the tests passed when I tested it with +mpi and +python, and I do not see py-mpi4py being installed in my spack env.

daboehme:
> @daboehme with +mpi and +python, if it depends on mpi4py, should we have depends_on("mpi4py") in the package.py? but the tests passed when I tested it with +mpi and +python, and I do not see py-mpi4py being installed in my spack env.

I think it does make sense to add the mpi4py dependency with +mpi+python. Adiak checks at runtime if mpi4py is present and works around it if not, so the tests still pass, but we do lose some functionality.

daboehme (approved):

LinaMuryanto:
I have tested this update on toss4-cpu.

tldahlgren (approved):
LGTM.

PR #3084, opened 2026-01-21
PyTorch: add v2.10.0

adamjstewart:
* https://github.com/pytorch/pytorch/releases/tag/v2.10.0
* https://github.com/pytorch/audio/releases/tag/v2.10.0
* https://github.com/pytorch/vision/releases/tag/v0.25.0

This is the final torchaudio release. Therefore, I've removed upper bounds to try to allow support for newer versions of Python/PyTorch. We can add them on an as-needed basis.

### DIFF
--- a/repos/spack_repo/builtin/packages/nvtx/package.py
+++ b/repos/spack_repo/builtin/packages/nvtx/package.py
@@ -19,6 +19,7 @@ class Nvtx(Package, PythonExtension):
     license("Apache-2.0")
 
     version("develop", branch="dev")
+    version("3.3.0", sha256="67d0cda2f9d19a89684592dab40c0bf2c2b13d5d588e51392076c0890a64b6c0")
     version("3.2.1", sha256="737c3035f0e43a2252e7cd94c3f26e11e169f624236efe31794f044ce44a70af")
     version("3.1.0", sha256="dc4e4a227d04d3da46ad920dfee5f7599ac8d6b2ee1809c9067110fb1cc71ced")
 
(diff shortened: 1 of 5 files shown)


repos/spack_repo/builtin/packages/nvtx/package.py
```diff
@@ -19,6 +19,7 @@ class Nvtx(Package, PythonExtension):
     license("Apache-2.0")
 
     version("develop", branch="dev")
+    version("3.3.0", sha256="67d0cda2f9d19a89684592dab40c0bf2c2b13d5d588e51392076c0890a64b6c0")
     version("3.2.1", sha256="737c3035f0e43a2252e7cd94c3f26e11e169f624236efe31794f044ce44a70af")
```
thomas-bouvier, line 23: The following `"Cython==3.0.11"` appeared at version v3.2.1, do we care?

repos/spack_repo/builtin/packages/py_torchaudio/package.py
thomas-bouvier, line 68: Just asking, why did we choose to keep version bounds tied together here? Someone advised my to decouple them on another PR.

thomas-bouvier (approved):
LGTM

repos/spack_repo/builtin/packages/nvtx/package.py
adamjstewart: Most likely future-proofing the code. It builds with the latest cython, so no reason to overconstrain.

repos/spack_repo/builtin/packages/py_torchaudio/package.py
adamjstewart: Above, Python versions are pinned based on which wheels are available. Most libraries that use the CPython API tend to be incompatible with newer versions of Python.

Below, PyTorch versions are pinned because PyTorch doesn't make any guarantees about compatibility of domain libraries other than the one released on the same day.

tldahlgren (commented):
Confirmed the commits[+tags] and checksum.

tldahlgren (approved):

PR #3085, opened 2026-01-21
CMake: conflict older CMake with newer curl

qwertos:

### DIFF
--- a/repos/spack_repo/builtin/packages/cmake/package.py
+++ b/repos/spack_repo/builtin/packages/cmake/package.py
@@ -175,7 +175,7 @@ class Cmake(Package):
     depends_on("curl")
 
     # https://gitlab.kitware.com/cmake/cmake/-/merge_requests/11134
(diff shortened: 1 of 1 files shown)


johnwparent (dismissed):
LGTM, thanks!

repos/spack_repo/builtin/packages/cmake/package.py
```diff
@@ -174,6 +174,9 @@ class Cmake(Package):
     depends_on("curl@:8.15", when="@:3.25")
     depends_on("curl")
 
+    # https://gitlab.kitware.com/cmake/cmake/-/merge_requests/11134
+    conflicts("curl@8.17:", when="@:3.30")
```
johnwparent: Apologies, approved too soon, looks like this extends to curl 8.16 as well: https://github.com/curl/curl/commit/1a12663d06d9c42fd40a8655d5247f7afeeb8e1e

johnwparent (approved):

PR #3095, opened 2026-01-22
Add HDF5 version 2.0.0.

lrknox:


hyoklee (approved):

lkurz (changes requested):
Larry - should probably also use "list_url" attribute since homepage and download url are at different locations.

brtnfld (approved):

gheber:
I don't think this is a good idea (using the GitHub CI/CD artifact link). As far as I can tell, GitHub guarantees checksums only for *uploaded* artifacts. See https://github.blog/changelog/2025-03-18-github-actions-now-supports-a-digest-for-validating-your-artifacts-at-runtime/ and https://docs.github.com/en/actions/tutorials/store-and-share-data#downloading-or-deleting-artifacts.

lrknox:
> I don't think this is a good idea (using the GitHub CI/CD artifact link). As far as I can tell, GitHub guarantees checksums only for _uploaded_ artifacts. See https://github.blog/changelog/2025-03-18-github-actions-now-supports-a-digest-for-validating-your-artifacts-at-runtime/ and https://docs.github.com/en/actions/tutorials/store-and-share-data#downloading-or-deleting-artifacts.

Our release-files.yml workflow downloads the source and binary artifacts, runs sha256sum on them, storing the checksum in hdf5-2.0.0.sha256sums.txt for the 2.0.0 release, and then uploads the files to https://github.com/HDFGroup/hdf5/releases using softprops/action-gh-release.  So the release Assets are uploaded files, and the checksums should be guaranteed.

The file that didn't have the correct checksum was one of those copied to S3 for archive purposes.  The step that we are missing and need to add is to calculate the sha256 checksum on the copied files to be archived and check the calculated  checksum against the corresponding one in the hdf5-2.0.0.sha256sums.txt file.

hyoklee:
https://github.blog/changelog/2025-10-28-immutable-releases-are-now-generally-available/

repos/spack_repo/builtin/packages/hdf5/package.py
```diff
@@ -39,13 +39,21 @@ class Hdf5(CMakePackage):
...
 
-    # Odd versions are considered experimental releases
-    # Even versions are maintenance versions
+    # Beginning with version 2.0.0, the versioning policy follows the MAJOR.MINOR.PATCH
+    # Semantic Versioning Specification. (See https://semver.org/)
+    version(
+        "2.0.0",
+        sha256="f4c2edc5668fb846627182708dbe1e16c60c467e63177a75b0b9f12c19d7efed",
+        url="https://github.com/HDFGroup/hdf5/releases/download/2.0.0/hdf5-2.0.0.tar.gz",
+        preferred=True,
```
haampie:
```suggestion
```

haampie:
Given that the tar file is bit by bit identical to the one on the HDF5 website and does not require some special token with time stamp in the URL, I'd say the current URL is best.

I've taken the liberty to update `url_for_version` and simplify, and remove `preferred=True` which is redundant on the latest version.

haampie (dismissed):

haampie:
@lrknox looks like CI is failing. `hdf5-vol-log` is using removed API, so a forward bound is enough, which I've pushed.

For `openpmd-api` I'm not sure:

```
CMake Error at CMakeLists.txt:358 (message):
  Found MPI but only serial version of HDF5.  Either set openPMD_USE_MPI=OFF
  to disable MPI or set openPMD_USE_HDF5=OFF to disable HDF5 or provide a
  parallel install of HDF5.
  If you manually installed a version of HDF5 in a non-default path, add its
  installation prefix to the environment variable CMAKE_PREFIX_PATH to find
  it: https://cmake.org/cmake/help/latest/envvar/CMAKE_PREFIX_PATH.html
```

is that an issue with `hdf5@2`, `cmake` modules or `openpmd-api`?


Then VTK:

```
     117    -- Found HDF5: hdf5-shared (found version "2.0.0") found components
            : C HL
  >> 118    CMake Error at CMake/vtkModule.cmake:6153 (message):
     119      The variable `HDF5_IS_PARALLEL` was expected to have been availab
            le, but
     120      was not defined.
     121    Call Stack (most recent call first):
     122      CMake/vtkModule.cmake:5861 (vtk_module_third_party_external)
     123      ThirdParty/hdf5/CMakeLists.txt:3 (vtk_module_third_party)
```

hyoklee (approved):

lrknox:
Subsequent commits have added later versions to the hdf5 package.py file.  New changes to add version HDF5 2.1.1 are now in PR #4822;  closing this one.

PR #3096, opened 2026-01-22
ratel - new version, and include FC for UHyper support

jeremylt:
We added prototype support for UHyper files, so adding the Fortran dependency now.

Oddly, we conditionally build the Fortran components only if FC is detected, but that seems to somehow be broken with Spack, as our makefile detects an FC, but the Fortran linker is not properly configured?
```
     133    gfortran -fPIC  -c -o build/src/materials/elasticity_uhyper.o /tmp/jeth8984/spack
            -stage/spack-stage-ratel-develop-x6kcut44qj6uc6r5sb7cklfyqiv5sspa/spack-src/src/m
            aterials/elasticity_uhyper.for
     134    /projects/ratel/spack/opt/spack/linux-zen2/compiler-wrapper-1.0-wcszu4iq7tvlglvhw
            ta6h3mbjufzgppq/libexec/spack/gcc/gfortran: 1: eval: SPACK_FC_LINKER_ARG: ERROR: 
            LINKER ARG WAS NOT SET, MAYBE THE PACKAGE DOES NOT DEPEND ON FC?
```


jeremylt:
I'm honestly not sure who is supposed to review these?

wdconinc:
> I'm honestly not sure who is supposed to review these?

Sorry this sometimes takes a while. If you are in the spack slack workspace, it often is useful to post the PRs that need attention in the #pull-requests channel.

wdconinc (approved):
Maintainer submitted. Looks good to me.

PR #3099, opened 2026-01-22
lmod: add v9.0.5

drkrynstrng:
Adding latest Lmod release v9.0.5.

### DIFF
--- a/repos/spack_repo/builtin/packages/lmod/package.py
+++ b/repos/spack_repo/builtin/packages/lmod/package.py
@@ -1,6 +1,7 @@
 # Copyright Spack Project Developers. See COPYRIGHT file for details.
 #
 # SPDX-License-Identifier: (Apache-2.0 OR MIT)
+
 from glob import glob
 
 from spack_repo.builtin.build_systems.autotools import AutotoolsPackage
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/lmod/package.py
```diff
@@ -58,27 +60,21 @@ class Lmod(AutotoolsPackage):
...
-
     variant("auto_swap", default=True, description="Auto swapping of compilers, etc.")
     variant(
         "redirect", default=False, description="Redirect messages to stdout (instead of stderr)"
     )
 
+    depends_on("c", type="build")
+    depends_on("pkgconfig", type="build")
+    depends_on("procps", type="build")
+    depends_on("sed", type="build")
```
haampie, line 71: Used to be `platform=darwin`. I think it's fine either way, but unclear if you removed this intentionally

haampie (approved):

repos/spack_repo/builtin/packages/lmod/package.py
drkrynstrng: Some version of sed is required as a build dependency on all platforms, so I removed the platform constraint. I think BSD sed works now though.

alecbcs (approved):
Looks good to me. Thanks @drkrynstrng!

PR #3100, opened 2026-01-22
lua: add v5.4.8

drkrynstrng:
Adding Lua v5.4.8. Also removing the unneeded cxx dependency and adding the license (https://lua.org/license.html).


drkrynstrng:
@spackbot run pipeline

rbberger (approved):

PR #3101, opened 2026-01-22
libxc: add fhc variant

nolta:
Vasp & Quantum Espresso need libxc compiled with `--disable-fhc`.

@RMeli

### DIFF
--- a/repos/spack_repo/builtin/packages/libxc/package.py
+++ b/repos/spack_repo/builtin/packages/libxc/package.py
@@ -46,6 +46,7 @@ class Libxc(AutotoolsPackage, CudaPackage, CMakePackage):
     variant("shared", default=True, description="Build shared libraries")
     variant("kxc", default=False, when="@5:", description="Build with third derivatives")
     variant("lxc", default=False, when="@5:", description="Build with fourth derivatives")
+    variant("fhc", default=True, when="@5.2:", description="Enforce Fermi hole curvature")
     variant(
         "fortran",
         default=True,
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/libxc/package.py
```diff
@@ -46,6 +46,7 @@ class Libxc(AutotoolsPackage, CudaPackage, CMakePackage):
     variant("shared", default=True, description="Build shared libraries")
     variant("kxc", default=False, when="@5:", description="Build with third derivatives")
     variant("lxc", default=False, when="@5:", description="Build with fourth derivatives")
+    variant("fhc", default=True, when="@5.2:", description="Enforce Fermi hole curvature")
```
RMeli, line 49: From the release notes of 7.0.0 I see
> If Libxc is compiled with the –disable-fhc flag, XC_FLAGS_ENFORCE_FHC defaults to false, and the library works exactly as before.

I'm wondering if this should default to `False`, so that the old behavior is preserved across versions (and users can then explicitly turn the variant on, if needed)?

RMeli:
```suggestion
    variant("fhc", default=False, when="@5.2:", description="Enforce Fermi hole curvature")
```

RMeli (commented):
Small question about the default for the `fhc` variant, otherwise LGTM.

repos/spack_repo/builtin/packages/libxc/package.py
nolta: I believe `True` was the implicit default before this patch, so if that's true it preserves the old behavior. If we change it to `False` we'd have to update all the dependents to `depends_on("libxc+fhc", ...)`.

RMeli (approved):

PR #3102, opened 2026-01-22
vim: add v9.1.2103

drkrynstrng:
Adding Vim v9.1.2103.

### DIFF
--- a/repos/spack_repo/builtin/packages/vim/package.py
+++ b/repos/spack_repo/builtin/packages/vim/package.py
@@ -17,7 +17,7 @@ class Vim(AutotoolsPackage):
     """
 
     homepage = "https://www.vim.org"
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/vim/package.py
```diff
@@ -17,14 +17,15 @@ class Vim(AutotoolsPackage):
     """
 
     homepage = "https://www.vim.org"
-    url = "https://github.com/vim/vim/archive/v8.1.0338.tar.gz"
+    url = "https://github.com/vim/vim/archive/v9.1.2103.tar.gz"
```
sethrj, line 20: For future reference, the URL field serves as a template and doesn't need to be updated with new versions.

sethrj (approved):
Checksum matches. Thanks!

RikkiButler20 (approved):
LGTM

PR #3106, opened 2026-01-23
gcc: default to +binutils

haampie:
GCC has a conflict on `~binutils` for RHEL, where the system
assembler and linker are ancient. However

1. Spack has no clue whether or not the system toolchain is used, so the
   conflict can be too restrictive
2. In general when building a compiler, it makes sense to also build the
   assembler and linker; we wanna avoid system dependencies.

So, make things less restrictive by dropping the conflict, and instead
use better defaults.


michaelkuhn (dismissed):
LGTM

michaelkuhn:
@spackbot run pipeline

michaelkuhn:
@spackbot run pipeline

PR #3111, opened 2026-01-23
spiner: add v1.7.0

rbberger:
- [x] Requires #3110

PR #3113, opened 2026-01-23
singularity-eos: add v1.11.0

rbberger:
removes mpark-variant dependency. variant is now provided by ports-of-call.

- Requires: #3111 
- Requires: #3110


tldahlgren (approved):
Confirmed the version sha256.

PR #3115, opened 2026-01-24
screen: add v5.0.1

drkrynstrng:
Adding Screen v5.0.1. Also deprecating all older versions because of various CVEs over the years.


drkrynstrng:
@spackbot fix style

tldahlgren (approved):
Confirmed the new (and re-confirmed all affected) sha56s.

PR #3117, opened 2026-01-24
tmux: add v3.6a

drkrynstrng:
Adding tmux v3.6a. Also making some minor style changes.


tldahlgren (approved):
Confirmed new version sha256.

PR #3121, opened 2026-01-24
py-cartopy: relax Python constraints

adamjstewart:
Most recent version builds just fine with Python 3.14, there just aren't official wheels yet.


tldahlgren (approved):
Confirmed the branch.

PR #3122, opened 2026-01-24
py-docstring-parser: add v0.17.0

adamjstewart:
Previous version uses ast imports removed in Python 3.14.


tldahlgren (approved):
Confirmed the version sha256.

PR #3123, opened 2026-01-24
py-jedi: relax Python requirements

adamjstewart:
I don't use jedi, but I want to be able to build ipython with Python 3.14. IPython seems to work fine with this jedi, so perhaps the requirements are too strict?

### DIFF
--- a/repos/spack_repo/builtin/packages/py_jedi/package.py
+++ b/repos/spack_repo/builtin/packages/py_jedi/package.py
@@ -29,6 +29,9 @@ class PyJedi(PythonPackage):
         depends_on("python@:3.10", when="@:0.18.1")
         depends_on("python@:3.11", when="@:0.19.0")
         depends_on("python@:3.12", when="@:0.19.1")
+        # Although v0.19.2 does not explicitly support Python v3.14, it does work well enough 
+        # according to the project maintainer that we've chosen to drop the upperbound constraint
+        # https://github.com/ipython/ipython/issues/15117#issuecomment-3799451766
         # depends_on("python@:3.13", when="@0.19.2:")
 
         depends_on("py-setuptools")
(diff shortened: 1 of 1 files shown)


alecbcs:
Looks like https://github.com/davidhalter/jedi/issues/2064 is an open issue for Python 3.14 compatibility in Jedi with a few known bugs. 

I'm tempted to wait a little to see if there's any progress there before removing the constraint here. I'll revisit this PR at the end of this week and test out using v0.19.2 as my language server backend for v3.14 for a couple days to see if I run into any issues. 

If there's no progress on that issue and I discover no issues, I think it'll be fine to merge this sometime next week.

repos/spack_repo/builtin/packages/py_jedi/package.py
```diff
@@ -29,7 +29,7 @@ class PyJedi(PythonPackage):
         depends_on("python@:3.10", when="@:0.18.1")
         depends_on("python@:3.11", when="@:0.19.0")
         depends_on("python@:3.12", when="@:0.19.1")
-        depends_on("python@:3.13", when="@0.19.2:")
+        # depends_on("python@:3.13", when="@0.19.2:")
```
alecbcs, line 35:
```suggestion
        # Although v0.19.2 does not explicitly support Python v3.14, it does work well enough 
        # according to the project maintainer that we've chosen to drop the upperbound constraint
        # https://github.com/ipython/ipython/issues/15117#issuecomment-3799451766
        # depends_on("python@:3.13", when="@0.19.2:")
```

Adding comment as background on decision.

alecbcs (changes requested):
One quick suggestion to add some background to the commented out directive.

Based on David's [comments here](https://github.com/ipython/ipython/issues/15117#issuecomment-3799451766) I believe we should be good to merge.

alecbcs:
@spackbot fix style

alecbcs (approved):

PR #3130, opened 2026-01-25
py-pydantic: add v1.10.26

adamjstewart:
Silences Python 3.14 warnings


tldahlgren (approved):
Confirmed the version sha256.

Issue #3016, opened 2026-01-13
[mpas-model] dependencies question


teabagk7:
https://github.com/spack/spack-packages/blob/205253f318ee2d257b733b2584e219d8bd0064a5/repos/spack_repo/builtin/packages/mpas_model/package.py#L62-L67

@t-brown Would it be reasonable to add metis as a dependency, as it is required for the runs?

teabagk7:
@t-brown ?

t-brown:
Sorry, I've been thinking about it. I am inclined to say "no", as in you don't have to use Metis to generate the graph. You also only need to create them once. However, I do know where you are coming from. So undecided.

Since you opened the issue, you feel strongly for it to be added as a dependency. Do you want to open a PR with it included? I'd be fine approving it.

teabagk7:
Thanks for your reply. I’ll create PR, no problem!

PR #3136, opened 2026-01-26
mpas-model: add metis to dependencies

teabagk7:
Adding Metis to the dependencies of the mpas-model, as discussed here:
https://github.com/spack/spack-packages/issues/3016

### DIFF
--- a/repos/spack_repo/builtin/packages/mpas_model/package.py
+++ b/repos/spack_repo/builtin/packages/mpas_model/package.py
@@ -64,6 +64,7 @@ class MpasModel(MakefilePackage):
     depends_on("fortran", type="build")  # generated
 
     depends_on("mpi")
+    depends_on("metis", type="run")
     depends_on("parallelio")
 
     conflicts(


t-brown (approved):
LGTM. Thanks!

repos/spack_repo/builtin/packages/mpas_model/package.py
```diff
@@ -64,6 +64,7 @@ class MpasModel(MakefilePackage):
     depends_on("fortran", type="build")  # generated
 
     depends_on("mpi")
+    depends_on("metis", type="run")
```
tldahlgren, line 67: Does this dependency apply to all of the listed versions or for a specific version on?

tldahlgren (commented):
See comment for question.

repos/spack_repo/builtin/packages/mpas_model/package.py
teabagk7: MPAS uses an unstructured mesh and needs a graph partition for multi-rank MPI.

If the Spack package is intended to deliver a working MPAS MPI workflow out of the box, then `metis` as a dependency makes sense for all versions, because users will immediately need it to partition meshes for N>1.

teabagk7: @tldahlgren, Could we go ahead and merge this PR if there are no concerns?

t-brown (approved):

tldahlgren (approved):

PR #3138, opened 2026-01-26
tree-sitter: add cli tool to the installation

albestro:
According to https://tree-sitter.github.io/tree-sitter/5-implementation.html#implementation

> Tree-sitter consists of two components: a C library (libtree-sitter), and a command-line tool (the tree-sitter CLI).

Before this PR just the library was built and the CLI tool wasn't.

Here the instruction on how to build it https://tree-sitter.github.io/tree-sitter/6-contributing.html#building, for which `rust` is required.

Given https://tree-sitter.github.io/tree-sitter/creating-parsers/1-getting-started.html#dependencies, I also updated the other dependencies, namely adding `node-js` and dropping the `# generated` comment for `c`/`cxx`.

About `cxx`, I'm not sure if it should be there, because AFAIK the library is written in `c`, so it might be not needed at all.

TODO
- [ ] how to deal with `+cli` and not-fully-supported systems with version 0.25 (https://github.com/spack/spack-packages/pull/3138#issuecomment-3806671487)

### DIFF
--- a/repos/spack_repo/builtin/packages/tree_sitter/package.py
+++ b/repos/spack_repo/builtin/packages/tree_sitter/package.py
@@ -20,6 +20,7 @@ class TreeSitter(MakefilePackage):
 
     license("MIT")
 
+    version("0.26.3", sha256="7f4a7cf0a2cd217444063fe2a4d800bc9d21ed609badc2ac20c0841d67166550")
     version("0.26.2", sha256="3cda4166a049fc736326941d6f20783b698518b0f80d8735c7754a6b2d173d9a")
     version("0.25.3", sha256="862fac52653bc7bc9d2cd0630483e6bdf3d02bcd23da956ca32663c4798a93e3")
     version("0.25.2", sha256="26791f69182192fef179cd58501c3226011158823557a86fe42682cb4a138523")
--- a/repos/spack_repo/builtin/packages/tree_sitter/package.py
+++ b/repos/spack_repo/builtin/packages/tree_sitter/package.py
@@ -55,25 +55,34 @@ class TreeSitter(MakefilePackage):
 
     depends_on("node-js", type="run")
 
-    def edit(self, spec, prefix):
-        env["PREFIX"] = prefix
+    # https://github.com/tree-sitter/tree-sitter/pull/5226
+    patch(
+        "https://github.com/tree-sitter/tree-sitter/pull/5226.patch",
+        sha256="4fe83cf4fa00b3a54cd495d4b6a921ae70adc5bc8f49aeaf0c2218ec0e96f6b3",
+        when="@0.26:0.26.3",
+    )
 
-        # Starting from 0.25.0 endianness is taken into account using system headers
+    def setup_build_environment(self, env):
+        env.set("PREFIX", self.prefix)
+
+        # Starting from 0.25 endianness is taken into account using system headers
         #   https://github.com/tree-sitter/tree-sitter/pull/3740
         # but GLIBC provides them according to some defines that changed over time.
         #   https://www.sourceware.org/glibc/wiki/Release/2.20#Deprecation_of__BSD_SOURCE_and__SVID_SOURCE_feature_macros
-        if spec.satisfies("@0.25: ^glibc@:2.19"):
-            filter_file("-D_DEFAULT_SOURCE", "-D_BSD_SOURCE", "Makefile")
+        if self.spec.satisfies("@0.25 ^glibc@:2.19"):
+            env.append_flags("CFLAGS", "-D_BSD_SOURCE")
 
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/tree_sitter/package.py
```diff
@@ -60,3 +64,24 @@ def edit(self, spec, prefix):
         #   https://www.sourceware.org/glibc/wiki/Release/2.20#Deprecation_of__BSD_SOURCE_and__SVID_SOURCE_feature_macros
         if spec.satisfies("@0.25: ^glibc@:2.19"):
             filter_file("-D_DEFAULT_SOURCE", "-D_BSD_SOURCE", "Makefile")
+
+    def build(self, spec, prefix):
+        super().build(spec, prefix)
+
+        from spack.package import Executable
```
haampie: this can be removed, because it's part of `from spack.package import *`

albestro: Fixed that! 😉

albestro:
CI reports a problem https://gitlab.spack.io/spack/spack-packages/-/jobs/19967359#L819 that sounds similar to https://github.com/spack/spack-packages/blob/bd8ffd885d061b0e2c395692e8ce9dd6aca05b97/repos/spack_repo/builtin/packages/tree_sitter/package.py#L57-L62 but I don't know out of my mind how a similar fix should be applied to the rust build as well.

albestro:
I replicated the problem as in CI, so with centos7 on x86_64.

The problem seems a [missing `_BSD_SOURCE` definition](https://github.com/tree-sitter/tree-sitter/pull/5219) in the build script of the rust binding used by the cli tool, which has been [backported to 0.26](https://github.com/tree-sitter/tree-sitter/pull/5226) (on branch `release-0.26`) but not on previous versions. So, **I'm going to add a patch for 0.26, but I wouldn't bother adding a patch for 0.25 (and before)**.

I don't know this project in details, but AFAICT it might be worth considering the cli tool an optional, hence adding a variant for it. With a `variant("cli", default=false)` the CI would pass correctly, but we would still let a user build `+cli` also for "not-fully-supported" versions (e.g. 0.25 on BSD would not work, but on a GNU should work).

If we go this way, we might think of adding either:
- a conflict for 0.25 for specific platforms (I don't know if and how we can identify them in spack)
- more easily have the variant conditionally available just from 0.26 on (by applying the patch)

I wouldn't go down the path of separating the library `tree-sitter` from the tool `tree-sitter-cli` in two different spack packages (considering that they belong to the same repo); but if you think it is better, we can go that way.

### Extra dependency

By replicating the problem, I realized that there is also an additional hidden dependency on `libclang`

- [bindgen dependency of tree-sitter/lib/binding-rust](https://github.com/tree-sitter/tree-sitter/blob/4f8983b89f0e891085f758930cf7238f1275ea4d/lib/Cargo.toml#L63)
- [bindgen crate requires libclang](https://github.com/rust-lang/rust-bindgen/blob/74a5f30c14769891e12de53ba275233ef2616819/README.md?plain=1#L74)

So, for building the `cli` also `llvm` is a dependency. I'm going to add that too.

albestro:
I've not added the `llvm` dependency for `rust-bindgen` yet, but I'm going to check if

- https://github.com/rust-lang/rust-bindgen/blob/main/book/src/requirements.md#clang
- https://github.com/rust-lang/rust-bindgen/blob/main/book/src/requirements.md#from-source

can be represented with something like

```
depends_on("libllvm@9:", type="build")
```

repos/spack_repo/builtin/packages/tree_sitter/package.py
```diff
@@ -48,15 +49,43 @@ class TreeSitter(MakefilePackage):
     version("0.20.2", sha256="2a0445f8172bbf83db005aedb4e893d394e2b7b33251badd3c94c2c5cc37c403")
     version("0.20.1", sha256="12a3f7206af3028dbe8a0de50d8ebd6d7010bf762db918acae76fc7585f1258d")
 
-    depends_on("c", type="build")  # generated
-    depends_on("cxx", type="build")  # generated
+    depends_on("c", type=["build", "run"])
+    depends_on("cxx", type=["build", "run"])
```
albestro: verify if it can be dropped

```suggestion
```

albestro: AFAIK it is a C library, and at runtime `tree-sitter-cli` needs a C compiler.

haampie (approved):

alalazo:
Ok, this one is still blocked on a bug we'll submit a fix for shortly

albestro:
@spackbot run pipeline

PR #3139, opened 2026-01-26
ci: try out lazy directives

haampie:
Not meant to be merged, but for testing spack/spack


haampie:
e4s generate goes in `11 minutes 36 seconds` looks good.

still have to look if there's any diff in hashes.

PR #3143, opened 2026-01-26
PDI: add missing numpy dependency

tpadioleau:
Following https://github.com/pdidev/pdi/issues/616, without this, `import pdi` fails in a Spack environment containing only `pdi+python`.


tldahlgren (approved):
Waiting on @jbigot

jbigot (approved):
Good catch, thanks.

PR #3119, opened 2026-01-24
node-js: fix Apple Clang/Python supported versions

adamjstewart:

PR #3144, opened 2026-01-26
Update Developer Tools Stack (MacOS aarch64)

alecbcs:
Add additional popular CLI tools, reorganize, and clean up stack comments a bit since it was initially created.

TODO:
- [x] test if commented out specs build after recent updates
- [x] discuss re-oganization with @trws

### DIFF
--- a/stacks/developer-tools-darwin/spack.yaml
+++ b/stacks/developer-tools-darwin/spack.yaml
@@ -16,7 +16,7 @@ spack:
       # editors
     - emacs+json~native+treesitter   # TODO +native not supported until gcc builds on darwin
     - nano
-    - neovim
+    # - neovim
     - py-pynvim
 
       # runtimes and language compilers
@@ -24,16 +24,15 @@ spack:
     - go
     - libtool
     - llvm+link_llvm_dylib+lld~lldb~polly+python
-    - node-js
+    # - node-js # TODO: re-enable after spack/spack-packages#3119 merges
     - npm
     - python
     - rust+dev
-    - zig
 
       # cli dev tools
     - bfs
     - ccache
-    - difftastic
+    # - difftastic # TODO investigate linker error here
     - eza
     - fd
     - flex
@@ -59,7 +58,7 @@ spack:
       # styling and lints
     - aspell
     - astyle
-    # - cppcheck # need to resolve build error in ci context
+    # - cppcheck # TODO: investigate why this is not reproducible outside of runner
     - py-black
     - py-codespell
     - py-flake8
--- a/stacks/developer-tools-darwin/spack.yaml
+++ b/stacks/developer-tools-darwin/spack.yaml
@@ -53,7 +53,6 @@ spack:
     - sed
     - tmux
     - tree
-    - trip
     - which
     - zoxide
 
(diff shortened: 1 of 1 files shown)


zackgalbreath:
We got most of the way there, but [five packages failed to build](https://gitlab.spack.io/spack/spack-packages/-/pipelines/1401890/failures) in the `develop-tools` stack.

alecbcs:
Yep planning to work on reproducing and fix each of the builds later today.

alecbcs:
@cosmicexplorer found when testing here that the constraint bounds on `node-js` missed `v24.2.0` and thus we were concretizing to the latest possible python when Spack fell back the most general dependency spec. To prevent this case from happening again in the future, I think it's probably best if we drop the lower bound version constraints entirely and just rely on Spack layering the constraints instead.

Edit: fixed by @adamjstewart in #3119

trws:
This looks pretty good to me.  Did you test to see if the neovim dependency can build on the newer version by any chance?  If so that would be nice, if not it might be good to have that documented.

trws:
I suppose my main question on the reorg (didn't occur to me until just now) is would you apply it to the other instances of the devtools stack? I don't have a copy set up on my loft, so  apologies if this is already done or something, just occurred to me my main hope was to keep it relatively easy to move the stacks forward together.

alecbcs:
Yep! I was going to tackle each as a separate PR, but I'm definitely planning to make them all consistent.

alecbcs:
> This looks pretty good to me. Did you test to see if the neovim dependency can build on the newer version by any chance? If so that would be nice, if not it might be good to have that documented.

Unfortunately I did try uncommenting the neovim spec and it still failed to build. I might do a deep dive in another PR to try and get it working though.

trws:
Darn, thanks for trying it!

trws (dismissed):

repos/spack_repo/builtin/packages/node_js/package.py
```diff
@@ -77,25 +77,25 @@ class NodeJs(Package):
...
+    depends_on("python@:3.10", when="@:19.0.1", type="build")
+    depends_on("python@:3.11", when="@:20.10.0", type="build")
+    depends_on("python@:3.12", when="@:20.15.1", type="build")
+    depends_on("python@:3.13", when="@:20.18.3", type="build")
+    depends_on("python@:3.11", when="@:21.1.0", type="build")
+    depends_on("python@:3.12", when="@:22.2.0", type="build")
+    depends_on("python@:3.13", when="@:22.16.0", type="build")
+    depends_on("python@3.8:3.13", when="@:23", type="build")
+    depends_on("python@3.9:3.13", when="@:24", type="build")
+    depends_on("python@3.9:3.14", when="@:25", type="build")
```
adamjstewart: This change is not correct, as multiple version ranges overlap and conflict with each other. I'm guessing #3119 is what you're looking for?

adamjstewart (changes requested):

repos/spack_repo/builtin/packages/node_js/package.py
alecbcs: Yep, that's better. I'll drop the patch here and rebase when your PR merges.

alecbcs:
@adamjstewart I've rebased this PR now that #3119 is merged. I'll need you to approve before we can proceed since you requested changes.

adamjstewart (approved):

kwryankrattiger (approved):

alecbcs:
Rebase after #3548 is merged due to failing check on develop.

zackgalbreath (approved):

PR #3149, opened 2026-01-26
gha: fix typo

haampie:
Fix a typo and bashism.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -80,7 +80,7 @@ jobs:
     steps:
     - name: Success
       run: |
-        if [ "${{ needs.prechecks.result }}" == "failure" ] || [ "${{ needs.prechecks.result }}" == "canceled" ]; then
+        if [ "${{ needs.prechecks.result }}" = "failure" ] || [ "${{ needs.prechecks.result }}" = "cancelled" ]; then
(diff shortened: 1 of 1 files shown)


tldahlgren:
> Fix a typo and bashism.

By bashism is it safe to assume you mean to switch to a more shell compatible, POSIX compliant equality check?

tldahlgren:
> Fix a typo and bashism.

What is the best way to check that the new spelling is what is returned?  Even Github isn't consistent in e.g., https://docs.github.com/en/actions/reference/workflows-and-actions/expressions#always through https://docs.github.com/en/actions/reference/workflows-and-actions/expressions#cancelled.

tldahlgren (commented):
See my prior comments (and GitHub link) for questions.

alecbcs:
I swapped to a more generic version of this all test for a couple other projects. Might come in handy here too.
https://github.com/llnl/hubcast/blob/531bd31dff7bc1fe9d79097e66bf5830a941c4e5/.github/workflows/ci.yml#L87-L105

alalazo (approved):
Same as we merged in `spack/spack`

alalazo:
Merging since it prevents that a change goes accidentally in if a job is canceled

PR #3152, opened 2026-01-27
add python-mumps recipe 

williampiat3:

### DIFF
--- a/repos/spack_repo/builtin/packages/mumps/package.py
+++ b/repos/spack_repo/builtin/packages/mumps/package.py
@@ -478,18 +478,20 @@ class Mumps(Package):
                 with open(join_path(pkg_path, f"{char}mumps.pc"), "w") as f:
                     desc = f"The {parallel_desc} {precision_desc[char]} MUMPS library {ord_desc}"
                     f.write(
-                    "\n".join([
-                    f"prefix={self.prefix}",
-                    "exec_prefix=${prefix}",
-                    "includedir=${prefix}/include",
-                    "libdir=${exec_prefix}/lib",
-                    "",
-                    f"Name: {char}mumps",
-                    f"Description: {desc}",
-                    f"Version: {self.version}",
-                    "Cflags: -I${includedir}",
-                    f"Libs: -L${{libdir}} -l{char}mumps"
-                    ])
+                        "\n".join(
+                            [
+                                f"prefix={self.prefix}",
+                                "exec_prefix=${prefix}",
+                                "includedir=${prefix}/include",
+                                "libdir=${exec_prefix}/lib",
+                                "",
+                                f"Name: {char}mumps",
+                                f"Description: {desc}",
+                                f"Version: {self.version}",
+                                "Cflags: -I${includedir}",
+                                f"Libs: -L${{libdir}} -l{char}mumps",
+                            ]
+                        )
                     )
 
     @property
--- a/repos/spack_repo/builtin/packages/mumps/package.py
+++ b/repos/spack_repo/builtin/packages/mumps/package.py
@@ -477,17 +477,18 @@ class Mumps(Package):
             ):
                 with open(join_path(pkg_path, f"{char}mumps.pc"), "w") as f:
                     f.write(
-                        f"""prefix={self.prefix}
-                exec_prefix=${{prefix}}
-                includedir=${{prefix}}/include
-                libdir=${{exec_prefix}}/lib
-
-                Name: {char}mumps
-                Description: The {parallel_desc} {precision_desc[char]} MUMPS library {ord_desc}
-                Version: {self.version}
-                Cflags: -I${{includedir}}
-                Libs: -L${{libdir}} -l{char}mumps
-                """
+                "\n".join([
+                f"prefix={self.prefix}",
+                f"exec_prefix=${{prefix}}",
+                f"includedir=${{prefix}}/include",
+                f"libdir=${{exec_prefix}}/lib",
+                "",
+                f"Name: {char}mumps",
+                f"Description: The {parallel_desc} {precision_desc[char]} MUMPS library {ord_desc}",
+                f"Version: {self.version}",
+                f"Cflags: -I${{includedir}}",
+                f"Libs: -L${{libdir}} -l{char}mumps"
+                ])
                     )
 
     @property
--- a/repos/spack_repo/builtin/packages/mumps/package.py
+++ b/repos/spack_repo/builtin/packages/mumps/package.py
@@ -476,21 +476,20 @@ class Mumps(Package):
                 or ("+complex" in self.spec and "+double" in self.spec and char == "z")
             ):
                 with open(join_path(pkg_path, f"{char}mumps.pc"), "w") as f:
-                
-                desc = f"The {parallel_desc} {precision_desc[char]} MUMPS library {ord_desc}"
+                    desc = f"The {parallel_desc} {precision_desc[char]} MUMPS library {ord_desc}"
                     f.write(
-                "\n".join([
-                f"prefix={self.prefix}",
-                "exec_prefix=${prefix}",
-                "includedir=${prefix}/include",
-                "libdir=${exec_prefix}/lib",
-                "",
-                f"Name: {char}mumps",
-                f"Description: {desc}",
-                f"Version: {self.version}",
-                "Cflags: -I${includedir}",
-                f"Libs: -L${{libdir}} -l{char}mumps"
-                ])
+                    "\n".join([
+                    f"prefix={self.prefix}",
+                    "exec_prefix=${prefix}",
+                    "includedir=${prefix}/include",
+                    "libdir=${exec_prefix}/lib",
+                    "",
+                    f"Name: {char}mumps",
+                    f"Description: {desc}",
+                    f"Version: {self.version}",
+                    "Cflags: -I${includedir}",
+                    f"Libs: -L${{libdir}} -l{char}mumps"
+                    ])
                     )
 
     @property
--- a/repos/spack_repo/builtin/packages/mumps/package.py
+++ b/repos/spack_repo/builtin/packages/mumps/package.py
@@ -66,6 +66,7 @@ class Mumps(Package):
         description="Allow BLAS calls in OpenMP regions "
         + "(warning: might not be supported by all multithread BLAS)",
     )
+    variant("pkgconfig", default=False, description="Create unofficial pkgconfig files")
 
     depends_on("c", type="build")  # generated
     depends_on("fortran", type="build")  # generated
@@ -175,7 +176,9 @@ class Mumps(Package):
             )
 
             orderings.append("-Dmetis")
-
+        if "+pkgconfig" in self.spec:
+            # Keeping orderings in case we have to create pkg_config files
+            self.orderings = orderings
         makefile_conf.append("ORDERINGSF = %s" % (" ".join(orderings)))
 
         # Determine which compiler suite we are using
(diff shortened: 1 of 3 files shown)


williampiat3:
this is the missing recipe for using python-mumps in our spack workflows

jcortial-safran:
I use https://people.freedesktop.org/~dbn/pkg-config-guide.html as reference.

A couple remarks:
- The generated files are weirdly formatted
- It should be possible to factor them better using ${prefix} and ${exec-prefix} to define includedir and libdir
- Only trivial information is included in the generated file. For instance, no flags relative to dependencies (such as scotch) are added. Is this actually useful ?

williampiat3:
Thank you for your remarks, for context I tried to replicate the pkg_config in the conda recipe (https://github.com/conda-forge/mumps-feedstock/blob/main/recipe/make_pkg_config.py) This reference may be plagued with the same issues that you are raising, thus: 

* I changed the description of the pkg_config to include all orderings compiled in the spec
* I changed the includedir and libdir as requested
* I am not sure if adding flags to dependencies is useful, as far as my testing goes I haven't felt it was needed. But I am not an expert.

These pkg_config files are necessary for the MesonBuilder to detect MUMPS: meson detects packages using pkg_config files

repos/spack_repo/builtin/packages/py_python_mumps/package.py
```diff
@@ -0,0 +1,57 @@
+import os
```
tldahlgren: Please add the standard package copyright "header":


```suggestion
# Copyright Spack Project Developers. See COPYRIGHT file for details.
#
# SPDX-License-Identifier: (Apache-2.0 OR MIT

import os
```

repos/spack_repo/builtin/packages/py_python_mumps/package.py
```diff
@@ -0,0 +1,57 @@
...
+from spack_repo.builtin.build_systems.python import PythonPackage
+
+from spack.package import *
+
+
+class PyPythonMumps(PythonPackage):
+    """Python wrapper for the MUMPS solver (python_mumps)."""
+
+    homepage = "https://gitlab.kwant-project.org/kwant/python-mumps"
+    url = "https://pypi.org/packages/source/p/python-mumps/python_mumps-0.0.6.tar.gz"
```
tldahlgren: Why specify the URL instead of using the `pypi` property (https://spack.readthedocs.io/en/latest/build_systems/pythonpackage.html#id2)?

repos/spack_repo/builtin/packages/py_python_mumps/package.py
```diff
@@ -0,0 +1,57 @@
...
+    variant("mpi", default=True, description="Whether to have MPI support on python-mumps or not")
+
+    # build dependencies
+    with default_args(type="build"):
+        depends_on("meson")
+        depends_on("ninja")
+        depends_on("py-meson-python")
+        depends_on("py-setuptools")
+        depends_on("py-setuptools-scm")
+        depends_on("py-cython")
```
tldahlgren: According to https://gitlab.kwant-project.org/kwant/python-mumps/-/blob/v0.0.6/pyproject.toml?ref_type=tags, half of these need lower bounds:


```suggestion
        depends_on("meson@1.8:")
        depends_on("ninja")
        depends_on("py-meson-python@0.18:")
        depends_on("py-setuptools")
        depends_on("py-setuptools-scm")
        depends_on("py-cython@3.1.1:")
```

tldahlgren: Where do the dependencies on `py-setuptools` and `ninja` come from?  The former is a dependency of `py-setuptools-scm` but I'm not seeing the latter except as a `pixi` dependency.

repos/spack_repo/builtin/packages/py_python_mumps/package.py
```diff
@@ -0,0 +1,57 @@
...
+    homepage = "https://gitlab.kwant-project.org/kwant/python-mumps"
+    url = "https://pypi.org/packages/source/p/python-mumps/python_mumps-0.0.6.tar.gz"
+
+    maintainers = ["williampiat3"]
+
+    version("0.0.6", sha256="58c33104f77c448e127e9e6da316f71dfb1a17719ecf022634669d513306b1fa")
+
+    variant("mpi", default=True, description="Whether to have MPI support on python-mumps or not")
+
+    # build dependencies
```
tldahlgren, line 25: Looks like there is a dependency on `python@3.12:` (see https://gitlab.kwant-project.org/kwant/python-mumps/-/blob/v0.0.6/pyproject.toml?ref_type=tags#L30).

repos/spack_repo/builtin/packages/py_python_mumps/package.py
```diff
@@ -0,0 +1,57 @@
...
+        depends_on("py-meson-python")
+        depends_on("py-setuptools")
+        depends_on("py-setuptools-scm")
+        depends_on("py-cython")
+
+    # Python dependencies
+    with default_args(type=("build", "run")):
+        depends_on("py-numpy")
+        depends_on("py-mpi4py", when="+mpi")
+        depends_on("py-scipy")
```
tldahlgren: Looks like the `numpy` dependency should be `py-numpy@2:` and `scipy` should be `py-scipy@1.13:`.  See https://gitlab.kwant-project.org/kwant/python-mumps/-/blob/v0.0.6/pyproject.toml?ref_type=tags#L31.

Where is `py-mpi4py` coming from? If it's from the `pixi` dependency, then it is also constrained to `py-mpi4py@4.1:4` along with a couple of others above it in https://gitlab.kwant-project.org/kwant/python-mumps/-/blob/v0.0.6/pyproject.toml?ref_type=tags#L135.

repos/spack_repo/builtin/packages/py_python_mumps/package.py
```diff
@@ -0,0 +1,57 @@
+import os
+
+from spack_repo.builtin.build_systems.meson import MesonBuilder
+from spack_repo.builtin.build_systems.python import PythonPackage
+
+from spack.package import *
+
+
+class PyPythonMumps(PythonPackage):
```
tldahlgren, line 11: From the dependencies on `meson`, should this inherit from `MesonPackage` instead (https://spack.readthedocs.io/en/latest/build_systems/mesonpackage.html)?

tldahlgren (dismissed):
See comments for missing copyright, dependency constraints, and questions.

repos/spack_repo/builtin/packages/py_python_mumps/package.py
williampiat3: `ninja` is a meson builder dependency, see discussion below about MesonPackage

repos/spack_repo/builtin/packages/py_python_mumps/package.py
williampiat3: There is a dependency on `python@3.12`: however it may be a little bit too restrictive (see discussions here: https://github.com/conda-forge/python-mumps-feedstock/pull/30). In a few words, `python@3.10:` are working therefore I will add `python@3.10:` rather than `python@3.12`:

repos/spack_repo/builtin/packages/py_python_mumps/package.py
williampiat3: mpi4py comes from the variant when mumps is built using MPI. python-mumps provides an interface for mumps compiled with MPI. The couple of others you are referring to are conda-dependent packages and are not applicable here

repos/spack_repo/builtin/packages/py_python_mumps/package.py
williampiat3: I used the PythonPackage inheritance naively because this is a python package and I thought this was the only option in that case.
I thought about inheriting from both MesonPackage and PythonPackage. However they both have an `install` phase. Not knowing what would be the outcome and to avoid having the hassle of handling the double inheritance I used directly the MesonBuilder within the recipe since only the `setup` phase from the MesonBuilder is needed.
I am open to suggestions on how to make it  either a MesonPackage or a MesonPackage,PythonPackage, but I would like to keep the recipe as simple as possible

repos/spack_repo/builtin/packages/py_python_mumps/package.py
williampiat3: Fixed

williampiat3:
After a discussion with @jcortial-safran, the upcoming Mumps version should provide pkgconfig files directly without the need of creating them within the spack install. Therefore I will mark this PR as a draft since python-mumps is not about having all versions available but rather having mumps functionnalities in python.
Nonetheless, all your remarks @tldahlgren are still relevant since they refer to python-mumps config only and I do want to make this available via spack.

williampiat3:
Reopening the PR in preparation of latest mumps release (the release doesn't seem to provide pkgconfig files as we though it would)

repos/spack_repo/builtin/packages/mumps/package.py
```diff
@@ -441,6 +444,52 @@ def install(self, spec, prefix):
...
+                        f"""prefix={self.prefix}
+                exec_prefix=${{prefix}}
+                includedir=${{prefix}}/include
+                libdir=${{exec_prefix}}/lib
+
+                Name: {char}mumps
+                Description: The {parallel_desc} {precision_desc[char]} MUMPS library {ord_desc}
+                Version: {self.version}
+                Cflags: -I${{includedir}}
+                Libs: -L${{libdir}} -l{char}mumps
```
jcortial-safran: This creates spurious spaces in the generated files. Please change to get rid of those.

jcortial-safran (changes requested):

repos/spack_repo/builtin/packages/mumps/package.py
williampiat3: fixed

jcortial-safran (approved):

jcortial-safran (approved):

williampiat3:
@tldahlgren, by any chance, would you have some time to review this PR? We would like to move forward with it
Thank you very much for your time.

jcortial-safran (approved):

williampiat3:
@spackbot fix style

jcortial-safran (approved):

repos/spack_repo/builtin/packages/py_python_mumps/package.py
```diff
@@ -0,0 +1,69 @@
...
+        depends_on("py-numpy@2:")
+        depends_on("py-mpi4py@4.1:4", when="+mpi")
+        depends_on("py-scipy@1.13:")
+
+    # Optional/test deps
+    depends_on("py-pytest", type="test")
+
+    # External solver
+    depends_on("mumps+float+complex+double+metis+scotch+pkgconfig", when="+mpi")
+    depends_on("mumps~mpi+float+complex+double+metis+scotch+pkgconfig", when="~mpi")
```
becker33:
```suggestion
    # External solver
    depends_on("mumps+float+complex+double+metis+scotch+pkgconfig")
    depends_on("mumps+mpi", when="+mpi")
    depends_on("mumps~mpi", when="~mpi")
```
This might be easier to read -- the constraints are accumulated, so these have identical semantics

repos/spack_repo/builtin/packages/py_python_mumps/package.py
```diff
@@ -0,0 +1,69 @@
...
+
+    patch("patch_meson_build.patch")
+
+    def setup_build_environment(self, env):
+        env.set("CC", self.compiler.cc)
+        env.set("CXX", self.compiler.cxx)
```
becker33: Is this necessary, it bypasses the Spack compiler wrapper so we really shouldn't be doing this.

repos/spack_repo/builtin/packages/py_python_mumps/package.py
```diff
@@ -0,0 +1,69 @@
...
+
+    @run_before("install")
+    def setup_meson(self) -> None:
```
becker33: Why isn't this a `MesonPackage`?

jcortial-safran (approved):

becker33 (changes requested):

repos/spack_repo/builtin/packages/py_python_mumps/package.py
williampiat3: Noted! thank you

repos/spack_repo/builtin/packages/py_python_mumps/package.py
williampiat3: They are necessary in conjunction with line 29 and 30:

Initialy I did not have these lines and neither the `depends_on('c')` and `depends_on('cxx')` and it worked.

 However when using `spack find` the package was registered as _no_arch_. 

To make the compiler dependency explicit in the spec I added  `depends_on('c')` and `depends_on('cxx')` . Sadly, this resulted in the recipe failing due to meson not finding the compilers. 

I therefore added the lines you mention to have a working recipe with proper compiler tagging in the spec

This was a workaround to have a working recipe and a correct compiler spec on the package. I am open to any suggestions on how to make this reflect the proper compiler spec without these lines

williampiat3:
The last change should be enough: I rely completely on meson-py for building the package, no need to call the initial meson setup step

williampiat3:
@spackbot fix style

jcortial-safran (approved):

repos/spack_repo/builtin/packages/py_python_mumps/package.py
becker33: What was the error you saw when you built with the c and cxx dependencies but without these lines?

becker33 (approved):

PR #3153, opened 2026-01-27
py-rioxarray: add v0.21.0

adamjstewart:
https://github.com/corteva/rioxarray/releases/tag/0.21.0


repos/spack_repo/builtin/packages/py_rioxarray/package.py
```diff
@@ -35,9 +36,9 @@ class PyRioxarray(PythonPackage):
         depends_on("py-rasterio@1.4.3:", when="@0.19:")
         depends_on("py-rasterio@1.3:", when="@0.16:")
         depends_on("py-rasterio")
-        depends_on("py-xarray@2024.7:", when="@0.19:")
-        depends_on("py-xarray@2022.3:", when="@0.16:")
-        depends_on("py-xarray@0.17:")
+        depends_on("py-xarray@2024.7:2025.11", when="@0.19:")
+        depends_on("py-xarray@2022.3:2025.11", when="@0.16:")
+        depends_on("py-xarray@0.17:2025.11")
```
alalazo, line 41: Can we just add a separate upper bound:
```suggestion
        depends_on("py-xarray@2024.7:", when="@0.19:")
        depends_on("py-xarray@2022.3:", when="@0.16:")
        depends_on("py-xarray@0.17:")
        depends_on("py-xarray@:2025.11")
```
?

adamjstewart: We _can_, although I personally find this more confusing because I have to scroll to the bottom of the list before I can see an upper bound that affects the most recent version. I expect this upper bound to eventually go away anyway.

alalazo (approved):
Fine. The two things are semantically equivalent, and if this one's preferred by the maintainer it's ✔️

PR #3154, opened 2026-01-27
kokkos: bring back deprecated_code variant

rbberger:


tpadioleau (commented):
Actually the CMake defaults are to disable 4 and enable 5 in Kokkos 5, https://github.com/kokkos/kokkos/blob/a5d374ba179921b9805008dda06a4b690cfbdd1a/cmake/kokkos_enable_options.cmake#L41-L42. Note that the documentation seems wrong, https://kokkos.org/kokkos-core-wiki/get-started/configuration-guide.html#general-options.

I wonder if giving the control only over the current major version would be enough ? `deprecated_code` controls `Kokkos_ENABLE_DEPRECATED_CODE_X` for `Kokkos@X` ?

rbberger:
@tpadioleau I'll do whatever you prefer. It was my understanding that people were asking for X-1 deprecated code.

rbberger:
It would also be informative if Kokkos plans to always have both X and X-1 depreciated code

rbberger:
@tpadioleau changed the default to match the CMake defaults. I guess we wait until the next WG meeting to decide if we want the simpler variant?

cedricchevalier19:
During the WG meeting we say we want to keep the logic simple, so using a boolean variant is the group choice.

repos/spack_repo/builtin/packages/kokkos/package.py
```diff
@@ -404,6 +411,9 @@ def cmake_args(self):
             from_variant("Kokkos_ENABLE_COMPILE_AS_CMAKE_LANGUAGE", "cmake_lang"),
         ]
 
+        if spec.satisfies("@5: +deprecated_code"):
+            options.append(self.define("Kokkos_ENABLE_DEPRECATED_CODE_4", True))
```
tpadioleau: Do you know a way to get the major version ? Something like `self.define(f"Kokkos_ENABLE_DEPRECATED_CODE_{self.major_version}", True)` ?

rbberger: Did the generic thing.

repos/spack_repo/builtin/packages/kokkos/package.py
```diff
@@ -404,6 +412,14 @@ def cmake_args(self):
             from_variant("Kokkos_ENABLE_COMPILE_AS_CMAKE_LANGUAGE", "cmake_lang"),
         ]
 
+        if spec.satisfies("@5: +deprecated_code"):
+            if spec.version == Version("develop"):
+                highest = max(v for v in self.versions if not v.isdevelop())
+                major_version = int(str(highest.up_to(1)))
+            else:
+                major_version = int(str(spec.version.up_to(1)))
+            options.append(self.define(f"Kokkos_ENABLE_DEPRECATED_CODE_{major_version-1}", True))
```
tpadioleau:
```suggestion
            options.append(self.define(f"Kokkos_ENABLE_DEPRECATED_CODE_{major_version}", True))
```

I think we can only disable the deprecation for the current major version as there is no guarantee of the existence of the CMake option for the prior version.

Otherwise seems good to me

repos/spack_repo/builtin/packages/kokkos/package.py
```diff
@@ -300,6 +300,14 @@ class Kokkos(CMakePackage, CudaPackage, ROCmPackage):
         variant("cxxstd", default="17", values=("14", "17", "20"), when="@3")
         variant("cxxstd", default="17", values=("17", "20", "23"), when="@4")
         variant("cxxstd", default="20", values=("20", "23"), when="@5:")
+
+    variant(
+        "deprecated_code",
+        default=False,
```
tpadioleau:
```suggestion
        default=True,
```

We need to enable it to keep the current behavior.

repos/spack_repo/builtin/packages/kokkos/package.py
rbberger: I thought the reason we're doing this is to bring back X-1, since users were asking for it. Current version deprecated code will just issue a warning.
 https://github.com/kokkos/kokkos/blob/22ced87d9d0ced21f334280f9399eabe75805e87/cmake/kokkos_enable_options.cmake#L41-L42
DEPRECATED_CODE_5 is always on by default. The only reason people would be asking for this option was to bring back DEPRECATED_CODE_4. If that is going away, then sure, what you are proposing is the only meaningful resolution.

rbberger: I pushed the change, which is fine by me.

tldahlgren (dismissed):

tldahlgren:
@tpadioleau @rbberger Is this PR okay to merge now?

rbberger:
@cedricchevalier19 since I missed that WG meeting you mentioned, can you give the final approval for merge?

repos/spack_repo/builtin/packages/kokkos/package.py
```diff
@@ -404,6 +412,14 @@ def cmake_args(self):
             from_variant("Kokkos_ENABLE_COMPILE_AS_CMAKE_LANGUAGE", "cmake_lang"),
         ]
 
+        if spec.satisfies("@5: +deprecated_code"):
+            if spec.version == Version("develop"):
+                highest = max(v for v in self.versions if not v.isdevelop())
+                major_version = int(str(highest.up_to(1)))
+            else:
+                major_version = int(str(spec.version.up_to(1)))
+            options.append(self.define(f"Kokkos_ENABLE_DEPRECATED_CODE_{major_version}", True))
```
tpadioleau:
```suggestion
        if spec.satisfies("@5:"):
            if spec.version == Version("develop"):
                highest = max(v for v in self.versions if not v.isdevelop())
                major_version = int(str(highest.up_to(1)))
            else:
                major_version = int(str(spec.version.up_to(1)))
            options.append(from_variant(f"Kokkos_ENABLE_DEPRECATED_CODE_{major_version}", "deprecated_code"))
```
Sorry missed that in the previous review.

rbberger: Good catch!

rbberger:
@spackbot fix style

repos/spack_repo/builtin/packages/kokkos/package.py
tpadioleau: > I thought the reason we're doing this is to bring back X-1, since users were asking for it. Current version deprecated code will just issue a warning. https://github.com/kokkos/kokkos/blob/22ced87d9d0ced21f334280f9399eabe75805e87/cmake/kokkos_enable_options.cmake#L41-L42 DEPRECATED_CODE_5 is always on by default. The only reason people would be asking for this option was to bring back DEPRECATED_CODE_4. If that is going away, then sure, what you are proposing is the only meaningful resolution.

Sorry if there was a misunderstanding. I hope I understood correctly the request of the user. I understood the user was using Kokkos without deprecated code outside of Spack but then using Spack there was no way to disable it.

tpadioleau (approved):

cedricchevalier19 (approved):

tldahlgren (approved):
Thanks.

PR #3159, opened 2026-01-27
pipx build system

ebagrenrut:
This introduces a new build system based on [pipx](https://pipx.pypa.io/stable/). pipx is used to install Python applications (typically CLI applications) into isolated environments. It creates symlinks for the select cli tools that can be put into the user's PATH. It's convenient to have such a build system in Spack because there are occassionally times when one is developing a Python application using a Spack environment, for which they may be leveraging several Spack-provided Python packages, but they may also need to have a Python-based cli tool in their Spack environment that could have Python package dependencies that conflict with the dependencies of the application they're developing.

We developed this build system at my site for just this reason: We had occassions in which we wanted Python-based cli tools (`black`, `flake8`, `isort`, `poetry`, etc) in our Spack envs which also needed conflicting Python dependencies from Spack-provided `py-*` packages.

What `pipx` is _not_ good for is providing Python packages in general. It does not add Python packages to a site-packages directory like `pip` does, nor would one want to put on `PYTHONPATH` any libraries that `pipx` might have installed into an isolated virtual env. It's strictly good for Python-based tools/utilities that need to have an isolated run-time environment.

This PR:

- Updates the `pipx` package in spack-packages -- which was woefully out-of-date and seemed to be abandoned -- so that it can be used as the basis for a build system;
- Adds the `pipx` build system (`PipxPackage` and `PipxBuilder` classes)
- Adds some Python-based utilities that we have found convenient to include in our Spack envs without conflicting with any of the `py-*` packages in the envs.

I understand that the approach here might not be generalizable to _any_ alternative Python build system. I think something that needs to be discussed (in comments here, or in the Spack Slack) is how (if?) we should prevent Python-based packages from colliding. For example, Spack might happily install `py-black` and `pipx-black` in the same environment, but both of these packages provide the `black` command, so there is a collision. However, these packages do not provide equivalent functionality, either (`py-black` exposes Black's libraries via PYTHONPATH, while `pipx-black` doesn't and shouldn't), so using a "providers" approach may not be appropriate. Such a question might need to be worked out for _any_ alternative Python package build systems that may be introduced to Spack.

@adamjstewart You asked, so here it is :)


ebagrenrut:
@spackbot fix style

repos/spack_repo/builtin/build_systems/pipx.py
```diff
@@ -0,0 +1,111 @@
+# SPDX-License-Identifier: (Sana-Proprietary)
```
adamjstewart: This needs to change. @alalazo @tgamblin I'm surprised CI didn't catch this, I thought we had license checks for all files?

adamjstewart (changes requested):
Instead of having a new package class and duplicating all Python tools, why not add `PythonPipxBuilder` to `repos/spack_repo/builtin/build_systems/python.py` and allow all `PythonPackage`s to be built with either `build_system=python_pip` or `build_system=python_pipx`?

@alalazo I'm wondering if we should rename these to `build_system=pip` and `build_system=pipx`. Technically both tools can compile non-Python extensions, and I think it's already clear that they are mostly associated with Python installation. I'm also still dreaming of adding build/installer support in addition to pip, but currently there are too many dependencies.

repos/spack_repo/builtin/build_systems/pipx.py
ebagrenrut: Thanks for catching that! Certainly, I intended to replace any headers with the standard Spack header.

ebagrenrut:
> Instead of having a new package class and duplicating all Python tools, why not add `PythonPipxBuilder` to `repos/spack_repo/builtin/build_systems/python.py` and allow all `PythonPackage`s to be built with either `build_system=python_pip` or `build_system=python_pipx`?

I like the idea and will work on it. Though any Python package _can_ (probably) be installed with `pipx`, I think it would only make sense in the case in which the package can be useful when isolated in its own virtual env. Hopefully it's not too confusing for package maintainers.

When I originally developed the build system, it was in an internal Spack fork in which I didn't want to inadvertantly break any existing python_pip builds. Then when the spack-packages project was separated from the main Spack project, it was convenient to leave the pipx build system separate because it could easily live in our custom, internal repo without impacting any existing builds _and_ I didn't need to keep it in sync with python_pip at all.

ebagrenrut:
@adamjstewart A bit of a wrinkle: The `pipx` package is itself a `PythonPackage`. In `PythonPipxBuilder`, I need to set dependencies such that they depend on `pipx`, which creates a circular dependency (in which `pipx` _could_ depends on `pipx`). I'm open to suggestions. One option is that I could make the `pipx` package more generic and have it only depend on `python` and then in its `install` method call something like `spec["python"].command("-m", "pip". "install", "pipx")`.

My WIP is here, if you're interested in taking a look: https://github.com/ebagrenrut/spack-packages/blob/integrate-pipx-build_system-with-python/repos/spack_repo/builtin/build_systems/python.py

adamjstewart:
I think we can override the `build_system` variant in `pipx` to only support `build_system=python_pip`? Is it possible to install pipx without pip? We can't remove other dependencies, as Spack packages won't install on air-gapped networks unless we explicitly manage dependencies.

ebagrenrut:
I actually tried to set `build_system("python_pip")` [within pipx](https://github.com/ebagrenrut/spack-packages/blob/4a8fd4cb8890927cb9e6d0f4c29f40bc69427a48/repos/spack_repo/builtin/packages/pipx/package.py#L26), but I still get this when trying to use the repo:
```
Error: cannot load package 'pipx' from the 'builtin' repository: Package 'pipx' cannot depend on itself.
```
This is uncharted territory for me, though, so it's possible that I haven't coded things up properly. It doesn't look to me like any other package classes support multiple build systems, so I'm not exactly sure if I'm doing it correctly.

With respect to installing pipx without pip, I could refactor the `pipx` package to simply grab the pipx zippapp from GitHub and make it the `pipx` executable. This may only work on *nix platforms, though.

adamjstewart:
@alalazo will be the expert on this. I imagine it's the same situation for build/installer. I think we would have to convert all dependencies of pipx/build/installer to not be PythonPackage, which would be annoying.

ebagrenrut:
I _have_ written at least one other package that extends multiple package classes and then uses `build_system()` to select between the build_system based on the package's version. So I wonder if I could have distinct `PythonPipxPackage` and `PythonPipxBuilder` classes and then any given Python-based package that wanted to support both installers could extend both `PythonPackage` and `PythonPipxPackage` classes, add a `pipx` variant, and do something like:

```
build_system(conditional("python_pipx", when="+pipx"), "python_pip", default="python_pip")
```

Too ugly?

Perhaps by trying to integrate multiple build systems in `build_systems/python.py`, we are just moving ugliness out to the packages...

adamjstewart:
Moving ugliness out of the 3k Python packages is an honorable endeavor 🙂

ebagrenrut:
@spackbot fix style

ebagrenrut:
@adamjstewart OK, what I've now done is:

- Removed `pipx`'s dependency on `python_pip` by installing its zipapp.
- Added `PipxPythonBuilder` and `PipxPythonPackage` classes to `build_systems/python.py`. `PipxPythonPackage` extends `PythonPackage` in an attempt to avoid code duplication with respect to fetching tarballs from PyPI.

As far as I could tell, one can't define multiple build_systems in a package class; at least I could not figure out how to do so. 

I'm still working on a modification to an existing package to see if I can make this work by using the `build_system` variant at build time. I _think_ that such a package would need to extend `PipxPythonPackage` and then use `when`'s and `self.spec.satisfies` logic to do the appropriate thing depending on which build_system the user chooses.

ebagrenrut:
@adamjstewart I now have something that seems to actually work, but I'm not sure it's as elegant as you had hoped :) I've also adapted the `py-cutadapt` package to support a `build_system=python_pipx` variant to demonstrate that it works. Please let me know what you think once you get a chance to review it.

adamjstewart:
I would prefer not to change the base class for all 3k Python packages (of course not all make sense). We also can't remove required dependencies from `pipx`, this package will no longer install correctly on air-gapped networks where it used to work. Waiting for @alalazo to review since he designed most of this infrastructure.

ebagrenrut:
> I would prefer not to change the base class for all 3k Python packages (of course not all make sense).

Understood. I think I can make `PythonPackage` work as the sole base class if it sets:
```
build_system("python_pip", "python_pipx", default="python_pip")
```

Every package class that extends `PythonPackage` would then support both. However, every such package would still need to be modified to Do The Right Thing when `build_system=python_pipx` is specified by the user. `pipx`, like `cargo` and `go`, is going to pull all of its dependencies from a remote registry (pypi, in this case), so it won't work in an air-gapped case (at least I assume that `cargo` and `go` don't work in the air-gapped case), and all of the `py-*` dependencies are thus unnecessary when `build_system=pipx`.

There's also still the issue that having this in `PythonPackage` seems to result in `pipx` having a dependency on itself, even when I try to force `pipx` to only support `python_pip` build system by adding `build_system("python_pip")` to it. I may just not be aware of how to force a single build system on a package, so perhaps @alalazo has a better idea there.

ebagrenrut:
Closing this as I don't think there's sufficient interest. I will, however, be opening a PR that modernizes the `py-pipx` package, as I think that will still be useful.

PR #3163, opened 2026-01-28
Add py-pygithub@2.8.1

climbfuji:
## Description

All in the title. Built and tested with GCC 13.3.1 and Intel oneAPI 2025.3.0 on AlmaLinux 9 running under WSL2.


repos/spack_repo/builtin/packages/py_pygithub/package.py
```diff
@@ -15,20 +15,34 @@ class PyPygithub(PythonPackage):
...
+    depends_on("py-deprecated", type=("build", "run"), when="@:2.1.1")
+
+    # DetailedHTTPError: GET
+    # https://files.pythonhosted.org/packages/source/P/PyGithub/PyGithub-2.8.1.tar.gz
+    # returned 404: Not Found
+    def url_for_version(self, version):
+        if version == Version("2.8.1"):
+            return "https://files.pythonhosted.org/packages/c1/74/e560bdeffea72ecb26cff27f0fad548bbff5ecc51d6a155311ea7f9e4c4c/pygithub-2.8.1.tar.gz"
+        else:
+            return f"https://files.pythonhosted.org/packages/source/P/PyGithub/PyGithub-{version}.tar.gz"
```
tldahlgren: There are a couple of ways to handle PyPi changes like this. In this case it is from upper/mixed case package URL paths to lower case paths. Regardless, using the version with the hash is not the way to go.
One option is to give older versions precedence with a change such as:
  
```suggestion
    def url_for_version(self, version):
        name = "pygithub" if version >= Version("2.4") else "PyGithub"
        return f"https://files.pythonhosted.org/packages/source/{name[0]}/{name}/{name}-{version}.tar.gz"
```

tldahlgren: If you only make the suggested change then `spack checksum` will only show versions up to `2.3.0`.

I suggest you also change the `pypi` property to the newer archive form:

```
    pypi = "pygithub/pygithub-2.8.1.tar.gz"
```

With the suggestion, `spack checksum` will list all versions from `2.4.0` on.

tldahlgren: If you also give the older `version` directives version-specific URLs AND keep  (the suggested)`url_for_version`, then `spack checksum` will show the simplified (i.e., no hashes) URLs for ALL available versions of the package, from `0.0.0` through `2.8.1`.

In other words, this occurs if the URL directives are:

```
    version("2.8.1", sha256="341b7c78521cb07324ff670afd1baa2bf5c286f8d9fd302c1798ba594a5400c9")
    version("2.1.1", sha256="ecf12c2809c44147bce63b047b3d2e9dac8a41b63e90fcb263c703f64936b97c", url=f"https://files.pythonhosted.org/packages/source/P/PyGithub/pygithub-2.1.1.tar.gz")
    version("1.59.1", sha256="c44e3a121c15bf9d3a5cc98d94c9a047a5132a9b01d22264627f58ade9ddc217", url=f"https://files.pythonhosted.org/packages/source/P/PyGithub/PyGithub-1.59.1.tar.gz")
    version("1.55", sha256="1bbfff9372047ff3f21d5cd8e07720f3dbfdaf6462fcaed9d815f528f1ba7283", url=f"https://files.pythonhosted.org/packages/source/P/PyGithub/PyGithub-1.55.tar.gz")
```

repos/spack_repo/builtin/packages/py_pygithub/package.py
```diff
@@ -15,20 +15,34 @@ class PyPygithub(PythonPackage):
...
+    depends_on("py-setuptools@64:", type="build", when="@2.8.1:")
     depends_on("py-setuptools-scm", type="build", when="@1.58.1:")
+    depends_on("py-setuptools-scm@7:", type="build", when="@2.8.1:")
     depends_on("py-pynacl@1.4.0:", type=("build", "run"))
     depends_on("py-python-dateutil", type=("build", "run"), when="@2.1.0:")
     depends_on("py-requests@2.14.0:", type=("build", "run"))
     depends_on("py-pyjwt@2.4.0:", type=("build", "run"))
     depends_on("py-pyjwt@2.4.0: +crypto", type=("build", "run"), when="@1.58.1:")
     depends_on("py-typing-extensions@4:", type=("build", "run"), when="@2.1.0:")
+    depends_on("py-typing-extensions@4.5:", type=("build", "run"), when="@2.8.1:")
```
tldahlgren, line 48: Can you move this new dependency to appear before the one conditioned on `@2.1.0:`?

tldahlgren (changes requested):
Confirmed the version checksum and dependencies; however, there are some suggestions to make adding additional versions easier.

URLs with hashes do not allow for Spack substitutions when newer versions are added. Refer to the suggested changes to use more standard `pypi` URLs. There are additional suggestions that can help improve `spack checksum` outputs as well.

Also, it is generally recommended to have dependencies conditioned on newer versions listed before those of older versions.

repos/spack_repo/builtin/packages/py_pygithub/package.py
climbfuji: Thanks for these suggestions, I'll work on that now.

climbfuji:
@spackbot fix style

tldahlgren (approved):
LGTM. Thanks.

PR #3164, opened 2026-01-28
py-pyvips: add c dependency

haampie:
Their C sources were written as Python strings, but a C compiler is still needed.


thomas-bouvier (approved):

thomas-bouvier:
Thank you!!

PR #3168, opened 2026-01-28
mimalloc: add 3.2.6

msimberg:


RMeli (dismissed):
LGTM, but out of curiosity, any reason for not adding `3.2.7`? 

https://github.com/microsoft/mimalloc/releases/tag/v3.2.7

msimberg:
That's a good question, and no good reason! Just the bad reason that I somehow missed that version. Thanks for spotting that. Let me add that as well.

RMeli (approved):

PR #3170, opened 2026-01-28
Add latest releases of gcovr

tvoskui:
Adding gcovr versions 8.4, 8.5, and 8.6 to the package.py

Per release notes, version 8.6 no longer supports Python 3.9

https://gcovr.com/en/stable/changelog.html


RikkiButler20 (approved):
LGTM
Verified sha256 & dependency

PR #3171, opened 2026-01-28
koliop: new package

tpadioleau:
hep @etiennemlb, do you mind review the package ?

### DIFF
--- a/repos/spack_repo/builtin/packages/koliop/package.py
+++ b/repos/spack_repo/builtin/packages/koliop/package.py
@@ -20,6 +20,7 @@ class Koliop(CMakePackage):
     license("MIT", checked_by="tpadioleau")
 
     version("master", branch="master", no_cache=True)
+    version("0.1.1", sha256="abf5a7187067a452a6a5e45dcd6502ef077f77a591839adf8d21ed4e8b9987f7")
     version("0.1.0", sha256="511df587fba11c16e728d1ccdc68c9004dcb0cc87c548e955599d69bf68e6642")
 
     depends_on("cxx", type="build")
--- a/repos/spack_repo/builtin/packages/koliop/package.py
+++ b/repos/spack_repo/builtin/packages/koliop/package.py
@@ -20,6 +20,7 @@ class Koliop(CMakePackage):
     license("MIT", checked_by="tpadioleau")
 
     version("master", branch="master", no_cache=True)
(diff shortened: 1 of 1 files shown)


etiennemlb:
LGTM, just, I'm not sure what the Spack policy on this is, but I would have called it libkoliop instead of koliop. A matter of taste only I think.

tpadioleau:
> LGTM, just, I'm not sure what the Spack policy on this is, but I would have called it libkoliop instead of koliop. A matter of taste only I think.

As you want, but i am not sure to know when to use koliop and libkoliop ? Do you have a rationale ?

etiennemlb:
When it is a lib, call it a lib, when it is not, dont use the lib. But thats being pedantic and this is not a widely followed rule anyway (spack list | grep "^lib" | wc -l == 346).

tpadioleau:
> When it is a lib, call it a lib, when it is not, dont use the lib. But thats being pedantic and this is not a widely followed rule anyway (spack list | grep "^lib" | wc -l == 346).

I meant compared to what we did in the CMake counterpart.

Anyway, i am fine with both. I let you do the change if necessary, otherwise i let you approve :)

etiennemlb (approved):
LGTM

repos/spack_repo/builtin/packages/koliop/package.py
```diff
@@ -0,0 +1,52 @@
...
+            self.define("koliop_ASSERT_ENABLED", True),
+            self.define("koliop_ASSUME_INPUT_BUFFERS_ARE_DEVICE_COMPATIBLE", True),
+            self.define("koliop_BUILD_FORTRAN_INTERFACE", False),
+            self.define("koliop_BUILD_TESTING", False),
+            self.define("koliop_CONSTANT_MEMORY_USAGE", False),
+            self.define("koliop_ENABLE_Kokkos", "SYSTEM"),
+            self.define("koliop_ENABLE_KokkosKernels", "SYSTEM"),
+            self.define("koliop_ENABLE_LTO", False),
+            self.define("koliop_FENCE_ON_OPERATOR_EXIT", False),
+            self.define("koliop_UTILITY_FLATTEN_EXPECT_POW2_EXTENTS", False),
```
becker33, line 44: Several of these are great candidates for future PRs to expose them as variants. Some of them less so.

becker33 (approved):

PR #3172, opened 2026-01-28
umpire + camp: add camp@2025.12.0 and add upper bound constraint for umpire and CUDA13

albestro:
Changes:
- add missing version for camp
- add upper bound for CUDA13 constraint in umpire

### DIFF
--- a/repos/spack_repo/builtin/packages/camp/package.py
+++ b/repos/spack_repo/builtin/packages/camp/package.py
@@ -25,6 +25,7 @@ class Camp(CMakePackage, CudaPackage, ROCmPackage):
     license("BSD-3-Clause")
 
     version("main", branch="main", submodules=False)
+    version("2025.12.0", sha256="e0e89649fc02beda44e28e2a750ecec1751e00f83498fec9762ebcf5ae66a8f7")
     version(
         "2025.09.2",
         tag="v2025.09.2",
--- a/repos/spack_repo/builtin/packages/camp/package.py
+++ b/repos/spack_repo/builtin/packages/camp/package.py
@@ -25,7 +25,12 @@ class Camp(CMakePackage, CudaPackage, ROCmPackage):
     license("BSD-3-Clause")
 
     version("main", branch="main", submodules=False)
-    version("2025.12.0", sha256="e0e89649fc02beda44e28e2a750ecec1751e00f83498fec9762ebcf5ae66a8f7")
+    version(
+        "2025.12.0",
+        tag="v2025.12.0",
+        commit="a8caefa9f4c811b1a114b4ed2c9b681d40f12325",
+        submodules=False,
+    )
     version(
         "2025.09.2",
         tag="v2025.09.2",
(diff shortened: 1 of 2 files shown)


repos/spack_repo/builtin/packages/umpire/package.py
```diff
@@ -265,6 +271,7 @@ class Umpire(CachedCMakePackage, CudaPackage, ROCmPackage):
     depends_on("camp+openmp", when="+openmp")
     depends_on("camp~cuda", when="~cuda")
     depends_on("camp~rocm", when="~rocm")
+    depends_on("camp@2025.12:", when="@2025.12:")
```
albestro, line 282: I was not able to find evidence of the explicit need for this, so that's the main question for reviewers.

repos/spack_repo/builtin/packages/camp/package.py
```diff
@@ -25,6 +25,7 @@ class Camp(CMakePackage, CudaPackage, ROCmPackage):
     license("BSD-3-Clause")
 
     version("main", branch="main", submodules=False)
+    version("2025.12.0", sha256="e0e89649fc02beda44e28e2a750ecec1751e00f83498fec9762ebcf5ae66a8f7")
```
msimberg: Any particular reason to use this style to specify the version compared to what was used for previous versions?

I prefer this style, but don't know if it has any effect on the downloaded archive... `submodules=False` on the other releases so this should also be ok.

msimberg (dismissed):
Looks good to me in general, but would like to hear from umpire developers before merging.

davidbeckingsale:
We have these versions added in #3067 which should be merged very soon

albestro:
> We have these versions added in #3067 which should be merged very soon

Thanks @davidbeckingsale, I haven't noticed that. I see it has been merged, and I've just rebased. There are still a couple of things that might be missing from that PR, so I'm not going to close this.

Waiting for your review 😉 and thanks in advance!

(I'm going to update the title and first message of this PR to adapt to the remaining changes after rebase)

albestro:
ping @davidbeckingsale

tldahlgren (dismissed):
Confirmed version sha256.

@adrienbernede @kab163 @trws @msimberg @davidbeckingsale Any reason to not merge this PR as is?

repos/spack_repo/builtin/packages/camp/package.py
adrienbernede: I would like the version to be added the same way as the previous ones.
```suggestion
    version(
        "2025.12.0",
        tag="v2025.12.0",
        commit="a8caefa9f4c811b1a114b4ed2c9b681d40f12325",
        submodules=False,
    )
```

repos/spack_repo/builtin/packages/umpire/package.py
adrienbernede: Although it may not be a strict requirement, we prefer to ensure coherency between the versions of those software.

adrienbernede (changes requested):
Thank you for those changes, they all make sense and should not be delayed anymore.
I am only requesting that we keep adding new versions the same way we used to.

repos/spack_repo/builtin/packages/camp/package.py
albestro: ops, sorry @msimberg. I completely missed this.

adrienbernede (approved):

tldahlgren (approved):

PR #3175, opened 2026-01-28
stacks: use environment with groups

alalazo:
Meant to try https://github.com/spack/spack/pull/51891

<details>
<summary>Pipelines</summary>

893f78329cfa126b6c8180d865b3267ef12f2f12: https://gitlab.spack.io/spack/spack-packages/-/pipelines/1414270

</details>


stacks/tutorial/spack.yaml
```diff
@@ -47,19 +39,42 @@ spack:
...
-      - ['%gcc@12']
   specs:
-  - $gcc_system_packages
+  - group: compiler
+    specs:
+    - gcc@12
+
+  - group: "gcc@11 specs"
+    specs:
+    - $gcc_system_packages
```
haampie: maybe inline those? the group already has a name

stacks/tutorial/spack.yaml
```diff
@@ -47,19 +39,42 @@ spack:
...
       - [^openblas, ^netlib-lapack]
-      - ['%gcc@12']
     - matrix:
       - [py-scipy ^openblas, armadillo ^openblas, netlib-lapack, openmpi, mpich, elpa ^mpich]
-      - ['%gcc@12']
   specs:
-  - $gcc_system_packages
+  - group: compiler
+    specs:
+    - gcc@12
```
haampie:
```suggestion
    - gcc@12.3
```

was on purpose

alalazo: Wasn't the purpose to have the same version that we would get from:
```
$ apt install gcc-12
```
on Ubuntu 22.04? Do we still need that now that compilers are dependencies?

alecbcs: As far as I'm aware we'd used apt before compilers as dependencies to bootstrap gcc@12 for the environment. With groups we can drop the usage of `apt` and use a spack build `gcc@12` to more closely mirror the steps in the tutorial.

stacks/radiuss/spack.yaml
```diff
@@ -9,43 +9,75 @@ spack:
...
       - target=x86_64_v3
-      - ~cuda
-      - ~rocm
     mpi:
-      require: openmpi
+      require:
+      - openmpi
 
   concretizer:
     unify: false
```
alalazo, line 17: @adrienbernede Is there any reason why we use `unify:false` for `radiuss` on the `cpu`? Can it be changed to `unify:true`?

adrienbernede: Here are my concerns:

This stack is not meant to be built together (e.g. as dependencies of a unique project), so it may over-constrain the specs and fail.

Could this result in building only older versions of certain packages ?
E.g. if A@latest requires B@latest-1, then B@latest will not be built in the stack, correct ?

alalazo: > Could this result in building only older versions of certain packages ?

Yes, that might happen. In the case you make there will be just `B@<latest-1>`. I'll keep `unify:false` then.

alalazo: Another question. Do you have any concern if the same specs that are now in `radiuss-cuda` would be built in a single pipeline together with the specs of `radiuss` (that's what this PR is experimenting with)?

adrienbernede: My only concern is that it will not help to decipher where a failing job comes from. What are the benefits of merging the pipelines ?

alalazo: Less duplicate builds, less runners used, faster pipelines. Also, since the `spack.yaml` is unique also more consistency in the configuration for the common parts.

adrienbernede: OK, I vote the benefits outweighs the drawbacks :)

adrienbernede: To make it clear:
Yes to merging pipelines, but no to unify:true to ensure we build the latest release of each package.

stacks/ml-linux-x86_64/spack.yaml
adamjstewart, line 1: Currently don't see a huge advantage since we still end up duplicating all config. What I was hoping for in https://github.com/spack/spack/issues/32893 and https://github.com/spack/spack/issues/32894 was the ability to create a cross-product/matrix of:

- packages
- variants
- arches

Ideally, these would also reuse shared dependencies instead of building the same non-GPU Python libs 3x.

alalazo: > Currently don't see a huge advantage since we still end up duplicating all config.

Where do you see that? Currently we are _deduplicating_ configuration, since the common part is shared at the top level, and the override per group is just for the parts that should be different among stack.

As for builds, the current situation on develop is:
- https://gitlab.spack.io/spack/spack-packages/-/pipelines/1408020

where for the ml-*-x86 stacks we build:
- 441 concrete specs for the `cpu` stack 
- 454 concrete specs for the `cuda` stack
- 228 concrete specs for the `rocm` stack

In this PR the unification of these 3 stacks:
- https://gitlab.spack.io/spack/spack-packages/-/jobs/20029384

gives 540 concrete specs to build, which seems a nice improvement over the sum of the three above :slightly_smiling_face:

alalazo: Also, not sure if that was wanted, but if we aim for a shorter configuration, we can use `definitions:` for the specs in the `ml` environment and share those definitions across `cpu`, `cuda` and `rocm` builds.

adamjstewart: Thanks for the clarification! If we can also deduplicate the package list via definitions, then this sounds like exactly what I'm looking for. My brain mostly revolves around GitHub Actions, so a matrix-based design like that is ideal, but definitions could also work for https://github.com/spack/spack/issues/32894. I tried this a year ago in https://github.com/spack/spack/pull/47464 but couldn't get it working.

adamjstewart, line 1: Can we change the directory to `stacks/ml-linux-x86_64` or even `stacks/ml/linux/x86_64`? It is no longer CPU-specific.

.ci/gitlab/.gitlab-ci.yml
adamjstewart: Same here, would like to remove "CPU` from this section to make it clear it's a single combined stack now.

stacks/ml-linux-x86_64-cpu/spack.yaml
```diff
@@ -18,23 +16,18 @@ spack:
       require: "%c,cxx=clang"
     py-tensorflow:
       require: "%c,cxx=clang"
-    py-torch:
-      require: "%c,cxx=gcc"
```
adamjstewart: Why do jaxlib and tensorflow stay but torch moves?

alalazo: That's because `py-torch` needs other changes for `cuda` and `rocm`, so I put everything in one place (instead of collecting `%c,cxx=gcc` in the main environment scope)

alalazo:
Closing. This PR has been resubmitted as multiple PRs, one per stack.

PR #3178, opened 2026-01-29
gawk: add v5.3.2

drkrynstrng:
Adding gawk v5.3.2. The libsigsegv dependency is no longer needed starting with v5.3.0.

### DIFF
--- a/repos/spack_repo/builtin/packages/gawk/package.py
+++ b/repos/spack_repo/builtin/packages/gawk/package.py
@@ -1,6 +1,7 @@
 # Copyright Spack Project Developers. See COPYRIGHT file for details.
 #
 # SPDX-License-Identifier: (Apache-2.0 OR MIT)
+
 import re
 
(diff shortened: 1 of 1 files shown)


haampie (dismissed):

repos/spack_repo/builtin/packages/gawk/package.py
```diff
@@ -43,10 +45,9 @@ class Gawk(AutotoolsPackage, GNUMirrorPackage):
 
     variant("nls", default=False, description="Enable Native Language Support")
 
-    depends_on("c", type="build")  # generated
-
+    depends_on("c", type="build")
     depends_on("gettext", when="+nls")
-    depends_on("libsigsegv")
+    depends_on("libsigsegv", when="@:5.2.2")
```
haampie:
```suggestion
    depends_on("libsigsegv", when="@:5.2")
```

FYI: that'd be more correct, would they ever release version 5.2.3.

haampie (approved):

PR #3187, opened 2026-01-29
py-jaxlib: better default decisions

alalazo:
This PR just reworks the `requires` directive so that an unconstrained concretization of `py-jaxlib` will not fall back to use v0.7.0 on `darwin` to avoid the penalty of using the second best required compiler.

### DIFF
--- a/repos/spack_repo/builtin/packages/nmad/package.py
+++ b/repos/spack_repo/builtin/packages/nmad/package.py
@@ -3,8 +3,9 @@
 # SPDX-License-Identifier: (Apache-2.0 OR MIT)
 
 from spack_repo.builtin.build_systems.autotools import AutotoolsPackage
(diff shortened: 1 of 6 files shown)


alalazo:
@spackbot fix style

adamjstewart:
This is an interesting bug. Yeah, JAX should always use Apple Clang on Darwin and LLVM on Linux. Not sure if LLVM would work on Darwin, but most things don't work on Darwin.

adamjstewart (approved):

alalazo:
Hmm, I didn't double check the changes by `spackbot` but it shouldn't have changed files that were not touched by the PR :disappointed:

adamjstewart:
We'll switch to ruff soon anyway, I wouldn't worry too much.

PR #3188, opened 2026-01-29
Remove '-v' flag from vcheck mode determination

sibcse:
In most compilers, the `-v` flag merely enables verbose output, which is different from `--version` or similar. However, the compiler wrapper switches into 'vcheck' mode if this flag is present, and does not apply any extra search paths or libraries as it should. Thus, if any package (e.g., in a Makefile) introduces a `-v` flag to the compilation or linking commands, it fails.

This PR simply removes the `-v` from the list of flags that are used to determine the 'vcheck' mode.

### DIFF
--- a/repos/spack_repo/builtin/packages/compiler_wrapper/cc.sh
+++ b/repos/spack_repo/builtin/packages/compiler_wrapper/cc.sh
@@ -357,7 +357,7 @@ esac
 if [ -z "$mode" ] || [ "$mode" = ld ]; then
     for arg in "$@"; do
         case $arg in
-            -v|-V|--version|-dumpversion)
+            -V|--version|-dumpversion)
                 mode=vcheck
                 break
                 ;;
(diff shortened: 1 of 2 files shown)


haampie:
Thanks! I think this is reasonable, but I vaguely recall that there was a reason to special-case `-v`, namely to prevent an exit code 1 that happened because Spack's compiler flags were added. Let's see what CI thinks of it in practice.

haampie:
This is what I had in mind: https://github.com/llvm/llvm-project/issues/116799

```
clang -v -Wl,-rpath,foo
```

This invokes the linker and errors. Spack adds `-Wl,-rpath,...`

sibcse:
Ok, thanks for having a look. That does look suboptimal, and several CI tests seem to fail... although, not all of those failures seem related to llvm/clang? What's your suggestion?

alalazo:
This PR needs to be ported to https://github.com/spack/compiler-wrapper

PR #3189, opened 2026-01-29
trilinos: allow %llvm (with fc=flang)

KineticTheory:
The current recipe for `trilinos` assumes that `%llvm` flavors will use `fc=gfortran`.  This small change allows compiles with `flang`.

### DIFF
--- a/repos/spack_repo/builtin/packages/trilinos/package.py
+++ b/repos/spack_repo/builtin/packages/trilinos/package.py
@@ -592,7 +592,7 @@ class Trilinos(CMakePackage, CudaPackage, ROCmPackage):
                 flags.append("-Wl,-undefined,dynamic_lookup")
 
             # Fortran lib (assumes clang is built with gfortran!)
-            if spec.satisfies("+fortran") and (
+            if spec.satisfies("+fortran %fortran=gcc") and (
                 spec.satisfies("%gcc")
                 or spec.satisfies("%clang")
                 or spec.satisfies("%apple-clang")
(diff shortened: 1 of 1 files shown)


KineticTheory:
@rbberger Here is the change you helped me with.

repos/spack_repo/builtin/packages/trilinos/package.py
```diff
@@ -592,7 +592,7 @@ def flag_handler(self, name, flags):
                 flags.append("-Wl,-undefined,dynamic_lookup")
 
             # Fortran lib (assumes clang is built with gfortran!)
-            if spec.satisfies("+fortran") and (
+            if spec.satisfies("+fortran %fortran=gcc") and (
                 spec.satisfies("%gcc")
                 or spec.satisfies("%clang")
                 or spec.satisfies("%apple-clang")
```
rbberger: You should be able to remove all other conditions since the first one covers all of them.

rbberger:
LGTM. I'll let the Trilinos folks chime in before merging.

kuberry:
I can confirm that on a mac using apple-clang and gfortran, this works. The logic that is being replaced precedes `%fortran=` existing and seems like an appropriate use.

rbberger (approved):

rbberger:
@kuberry @ccober6 @cgcgcg @rppawlo @psakievich I'm still missing an approval from one of the maintainers.

cgcgcg (approved):

PR #3195, opened 2026-01-29
node-js: add v24.13.0, deprecate v15 -> 23.11.1

emwjacobson:
Previously when attempting to install node-js 24.2.0 it could select python up to version 3.14 as node-js@24.2.0 was outside of any explicit `depends_on` statement for python. This would cause it to fall back to the generic `depends_on("python", type="build")` statement, which doesn't constrain the version.

Due to python3.14 removing "FancyURLOpener" from urllib [[1]](https://docs.python.org/3/whatsnew/3.14.html#id12), this would cause node-js@24.2.0 to fail to compile.

This update fixes the versioning for node-js 24 versions, constraining 24.0.0:24.2.0 up to python3.13 while allowing 24.3.0:24 to up to python 3.14. Additionally, since python3.9 is EOL, and building nodejs requires "a supported version of Python" [[2]](https://github.com/nodejs/node/blob/main/BUILDING.md#prerequisites) I thought it would also make sense to bump node-js@25's minimum version to python3.10.

Finally, I added the latest 24.13.0 version of node-js, as that's the latest LTS.

@cosmicexplorer is the listed maintainer, so obligatory mention. Comments, questions, concerns?


emwjacobson:
Latest commit/merge pulls in the new `depends_on` from `develop`, where it looks like the versioning was addressed since I originally opened this.

Regardless, this still marks the deprecated versions as deprecated and brings in the latest 24 LTS.

alecbcs:
@emwjacobson were the non-LTS deprecations due to a CVE or just because they are now EOL?

emwjacobson:
Just because they are now EOL. Is that not an appropriate use of marking them deprecated?

alecbcs:
No that's okay, if it was due to a CVE I just typically try to list the ID of the CVE as a comment.

alecbcs (approved):
Looks good to me. Thanks @emwjacobson!

PR #3197, opened 2026-01-29
CMakePackage: Add option to get cmake args from dependencies

kwryankrattiger:
This feature is not complete but it is a minimal implementation to provided the ability for packages to define hints for cmake. Not all hints can be passed as environment variables and cmake argument injection is not possible via existing mechanisms.

As a demonstration of this feature the python hint flags are converted from the ad hoc method in the CMakePackage to using the package callback.

Flags can be disabled using a common class list attribute `disable_cmake_hints_from`.

Extracted from #2057

### DIFF
--- a/.ci/gitlab/configs/ci.yaml
+++ b/.ci/gitlab/configs/ci.yaml
@@ -27,7 +27,7 @@ ci:
       after_script:
       - - cat /proc/loadavg || true
         - cat /proc/meminfo | grep 'MemTotal\|MemFree' || true
-      - - time ./spack python ${SPACK_CI_PACKAGES_ROOT}/.ci/gitlab/scripts/common/aggregate_package_logs.spack.py
+      - - time ./spack/bin/spack python ${SPACK_CI_PACKAGES_ROOT}/.ci/gitlab/scripts/common/aggregate_package_logs.spack.py
           --prefix /home/software/spack:${CI_PROJECT_DIR}/opt/spack
           --log install_times.json
           ${SPACK_ARTIFACTS_ROOT}/user_data/install_times.json || true
--- a/.ci/gitlab/configs/ci.yaml
+++ b/.ci/gitlab/configs/ci.yaml
@@ -27,7 +27,7 @@ ci:
       after_script:
       - - cat /proc/loadavg || true
         - cat /proc/meminfo | grep 'MemTotal\|MemFree' || true
-      - - time ./spack/bin/spack python ${SPACK_CI_PACKAGES_ROOT}/.ci/gitlab/scripts/common/aggregate_package_logs.spack.py
+      - - time ./spack python ${SPACK_CI_PACKAGES_ROOT}/.ci/gitlab/scripts/common/aggregate_package_logs.spack.py
           --prefix /home/software/spack:${CI_PROJECT_DIR}/opt/spack
           --log install_times.json
           ${SPACK_ARTIFACTS_ROOT}/user_data/install_times.json || true
--- a/repos/spack_repo/builtin/build_systems/cmake.py
+++ b/repos/spack_repo/builtin/build_systems/cmake.py
@@ -42,24 +42,6 @@ def _extract_primary_generator(generator):
     return _primary_generator_extractor.match(generator).group(1)
 
 
-def _maybe_set_python_hints(pkg: PackageBase, args: List[str]) -> None:
-    """Set the PYTHON_EXECUTABLE, Python_EXECUTABLE, and Python3_EXECUTABLE CMake variables
-    if the package has Python as build or link dep and ``find_python_hints`` is set to True. See
-    ``find_python_hints`` for context."""
-    if not getattr(pkg, "find_python_hints", False) or not pkg.spec.dependencies(
-        "python", deptype=("build", "link")
-    ):
-        return
-    python_executable = pkg.spec["python"].command.path
-    args.extend(
-        [
-            define("PYTHON_EXECUTABLE", python_executable),
-            define("Python_EXECUTABLE", python_executable),
-            define("Python3_EXECUTABLE", python_executable),
-        ]
-    )
-
-
 def _supports_compilation_databases(pkg: PackageBase) -> bool:
     """Check if this package (and CMake) can support compilation databases."""
 
@@ -172,19 +154,15 @@ class CMakePackage(PackageBase):
     https://cmake.org/cmake/help/latest/
     """
 
+    #: Disable adding extra CMake args provided by listed depenedencies
+    disable_cmake_hints_from: List[str] = []
+
     #: This attribute is used in UI queries that need to know the build
     #: system base class
     build_system_class = "CMakePackage"
 
     #: Legacy buildsystem attribute used to deserialize and install old specs
     default_buildsystem = "cmake"
-
-    #: When this package depends on Python and ``find_python_hints`` is set to True, pass the
-    #: defines {Python3,Python,PYTHON}_EXECUTABLE explicitly, so that CMake locates the right
-    #: Python in its builtin FindPython3, FindPython, and FindPythonInterp modules. Spack does
-    #: CMake's job because CMake's modules by default only search for Python versions known at the
-    #: time of release.
-    find_python_hints = True
 
     build_system("cmake")
 
@@ -405,7 +383,34 @@ class CMakeBuilder(BuilderWithDefaults):
             )
 
         _conditional_cmake_defaults(pkg, args)
-        _maybe_set_python_hints(pkg, args)
+
+        # Append extra hint/option arguments from dependencies
+        disable_cmake_hints_from = getattr(pkg, "disable_cmake_hints_from", [])
+        for dep in pkg.spec.edges_to_dependencies():
+            # Skip disabled dependency cmake args
+            # Consider virtual names as well package names
+            disabled_for_pkg = dep.spec.package.name in disable_cmake_hints_from
+            disabled_for_virtual = False
+            if dep.virtuals:
+                disabled_for_virtual = any((v in disable_cmake_hints_from for v in dep.virtuals))
+
+            if disabled_for_pkg or disabled_for_virtual:
+                continue
+
+            # TODO: Consider properties from virtual packages like Mpi
+            packages = [dep.spec.package]
+            for dep_pkg in packages:
+                if not hasattr(dep_pkg, "dependent_cmake_args"):
+                    continue
+                try:
+                    # Try calling as a function first
+                    args.extend(dep_pkg.dependent_cmake_args(pkg))
+                except TypeError:
+                    # Fallback to calling as a str/list attribute
+                    dep_args = dep_pkg.dependent_cmake_args
+                    if isinstance(dep_args, str):
+                        dep_args = [dep_args]
+                    args.extend(dep_args)
 
         return args
 
--- a/repos/spack_repo/builtin/build_systems/cmake.py
+++ b/repos/spack_repo/builtin/build_systems/cmake.py
@@ -385,8 +385,14 @@ class CMakeBuilder(BuilderWithDefaults):
         _conditional_cmake_defaults(pkg, args)
 
         # Append extra hint/option arguments from dependencies
+        # TODO: Consider properties from virtual packages like Mpi
         disable_cmake_hints_from = getattr(pkg, "disable_cmake_hints_from", [])
         for dep in pkg.spec.edges_to_dependencies():
+            # Skip packages without the callback
+            dep_pkg = dep.spec.package
+            if not hasattr(dep_pkg, "dependent_cmake_args"):
+                continue
+
             # Skip disabled dependency cmake args
             # Consider virtual names as well package names
             disabled_for_pkg = dep.spec.package.name in disable_cmake_hints_from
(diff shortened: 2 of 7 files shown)


adamjstewart:
Still don't love that it's opt-out instead of opt-in, but I understand why (warnings harm no one, and fixes help everyone). I agree it makes more sense to move this into the Python package.

haampie:
What I don't like about this conceptually is that the python package imports the cmake module.

Once we do content hashing properly and follow imports, it would mean that Python's hash is affected by changes to the `cmake` module. That feels wrong to me. I prefer the status quo of having this logic in `cmake.py`.

Secondly, you've put this under `python`, but in reality `pkg.spec["python"].command` picks up `python-venv`'s package. I think the logic is better kept in `cmake.py` where it chooses `python`'s or `python-venv`'s interpreter path. Putting it in the `python` package seems to suggest this is about `python`'s interpreter, but it's not.

repos/spack_repo/builtin/build_systems/cmake.py
```diff
@@ -405,7 +383,34 @@ def std_args(pkg: PackageBase, generator: Optional[str] = None) -> List[str]:
...
+
+            # TODO: Consider properties from virtual packages like Mpi
+            packages = [dep.spec.package]
+            for dep_pkg in packages:
+                if not hasattr(dep_pkg, "dependent_cmake_args"):
+                    continue
+                try:
+                    # Try calling as a function first
+                    args.extend(dep_pkg.dependent_cmake_args(pkg))
+                except TypeError:
```
alalazo: Why are we allowing it to be a list of strings or a string? Since this will not be used much I'd just allow the callback method.

repos/spack_repo/builtin/build_systems/cmake.py
```diff
@@ -405,7 +383,34 @@ def std_args(pkg: PackageBase, generator: Optional[str] = None) -> List[str]:
             )
 
         _conditional_cmake_defaults(pkg, args)
-        _maybe_set_python_hints(pkg, args)
+
+        # Append extra hint/option arguments from dependencies
+        disable_cmake_hints_from = getattr(pkg, "disable_cmake_hints_from", [])
+        for dep in pkg.spec.edges_to_dependencies():
+            # Skip disabled dependency cmake args
```
alalazo, line 396: (minor) I expect most functions not to implement this callback. So I think the first early "continue" should be along the lines of:
```python
if not hasattr(dep.spec.package, "dependent_cmake_args"):
    continue
```

Also, we shouldn't introduce the more general structure for virtuals below:
```python
packages = [dep.spec.package]
for dep_pkg in packages:
    ...
```
if we are not going to use it.

repos/spack_repo/builtin/packages/python/package.py
```diff
@@ -13,6 +13,7 @@
 from shutil import copy
 from typing import Dict, List
 
+from spack_repo.builtin.build_systems.cmake import define as cmdefine
```
alalazo: This is really a convenience function, so I guess we may avoid it to drop the dependency.

repos/spack_repo/builtin/build_systems/cmake.py
alalazo, line 1: Another thing missing here is some unit test for the feature, since it's going to be triggered for each and every CMake package.

alalazo (commented):
I have a few comments on the code, on top of the `python-venv` vs. `python` issue Harmen pointed out.[^1] 

Sorry to be pickier than for other PRs, but this is going to be triggered most of the times and is in a code path that is difficult to remove after we merge the PR.

[^1]: The `cmdefine` issue can be solved by not using the convenience function imo.

repos/spack_repo/builtin/build_systems/cmake.py
kwryankrattiger: I thought it might be convenient, but can drop since it isn't currently used anywhere.

repos/spack_repo/builtin/build_systems/cmake.py
kwryankrattiger: The virtuals structure doesn't/can't work at all with the current package API afaict. I can remove it and add it back later if needed.

repos/spack_repo/builtin/packages/python/package.py
kwryankrattiger: yeah, I didn't think about how we would be using this in the package hash. The only reason to include it was to take advantage of the helper function for consistency with other cmake option builders.

repos/spack_repo/builtin/build_systems/cmake.py
kwryankrattiger: Are you okay with allowing disabling flags by virtual name?

kwryankrattiger:
> Secondly, you've put this under python, but in reality pkg.spec["python"].command picks up python-venv's package. I think the logic is better kept in cmake.py where it chooses python's or python-venv's interpreter path. Putting it in the python package seems to suggest this is about python's interpreter, but it's not.

So what is the correct behavior here? I would think that `pkg.spec["python"].command` is what we want for configuring the dependent.

repos/spack_repo/builtin/build_systems/cmake.py
alalazo: > Are you okay with allowing disabling flags by virtual name?

Yes, that would be fine with me - but that can also be added when / if we hook `mpi` in? Right now I would only see Python and `boost` using this feature, right?

kwryankrattiger:
@haampie are you okay if we just put a comment next to the python setup that says something along the lines of

```
# this can redirect to python-venv if pkg extends python
```

repos/spack_repo/builtin/build_systems/cmake.py
kwryankrattiger: And mesa GL (next).

For MPI, without the virtual handling, we could make a `MPIPackage` class that is inherited by all of the MPI providers, and then they have the method by inheritance rather than virtual provider interface magic.

haampie:
> I would think that `pkg.spec["python"].command` is what we want for configuring the dependent.

Sure, but it's very confusing to use `pkg.spec["python"].command` in a method of `class Python`. Everyone reading that would expect `self.command` (used 22 times in the same class).

That tells me that `python` is not a good use case for this feature right now, so I would undo that change.

Maybe you could write it as

```python
class Python:
    def dependent_cmake_args(self, dependent_spec: Spec) -> List[str]:
        if dependent_spec.dependencies("python-venv"):
            return []
        return [... self.command ...]

...

class PythonVenv:
    def dependent_cmake_args(self, dependent_spec: Spec) -> List[str]:
        return [... self.command ...]
```

but that's also not satisfactory cause packages would have to opt out of two packages that provide hints instead of one.

repos/spack_repo/builtin/packages/python/package.py
```diff
@@ -1324,6 +1324,16 @@ def setup_dependent_package(self, module, dependent_spec):
         module.python_platlib = join_path(dependent_spec.prefix, self.platlib)
         module.python_purelib = join_path(dependent_spec.prefix, self.purelib)
 
+    def dependent_cmake_args(self, pkg: PackageBase) -> List[str]:
```
haampie: `dependent_pkg` at the very least. It's also inconsistent with `dependent_spec` used in `setup_dependent_*_environment`. The lack of `setup_*` prefix is also somewhat inconsistent?

kwryankrattiger: So to be clear here, you are asking to pass the spec instead of the package? I am fine with that, I just don't want to make the wrong change.

haampie: Yeah, just asking for consistency with https://spack.readthedocs.io/en/latest/packaging_guide_build.html#package-specific-environment-variables and https://spack.readthedocs.io/en/latest/packaging_guide_build.html#setting-package-module-variables

kwryankrattiger:
> That tells me that python is not a good use case for this feature right now, so I would undo that change.

I disagree on a couple points here.

Personally I don't think it is that confusing. It is reasonably well documented at this point that python has some special handling already and the comment links to the PR that has additional details about why this is the case. I spent a couple of seconds reading the PR and understood what was done well enough to understand why it was written that way.

We could also write it is `self.spec.command` which is what we do for MPI providers for similar cases.

I really don't like package specific code inside of the `CMakePackage`. The people maintaining that code are not the same people maintaining the oddities of python or other packages and how they are modeled in Spack. This moves the responsibility and context of python changes into python and out of the build system.

repos/spack_repo/builtin/build_systems/cmake.py
alalazo: Current implementation looks good to me. Closing this thread.

repos/spack_repo/builtin/packages/python/package.py
```diff
@@ -1324,6 +1324,16 @@ def setup_dependent_package(self, module, dependent_spec):
         module.python_platlib = join_path(dependent_spec.prefix, self.platlib)
         module.python_purelib = join_path(dependent_spec.prefix, self.purelib)
 
+    def dependent_cmake_args(self, dependenct_spec: Spec) -> List[str]:
```
alalazo: (minor) typo

```suggestion
    def dependent_cmake_args(self, dependent_spec: Spec) -> List[str]:
```
and same in the body of the method

.ci/gitlab/configs/ci.yaml
```diff
@@ -27,7 +27,7 @@ ci:
       after_script:
       - - cat /proc/loadavg || true
         - cat /proc/meminfo | grep 'MemTotal\|MemFree' || true
-      - - time ./spack/bin/spack python ${SPACK_CI_PACKAGES_ROOT}/.ci/gitlab/scripts/common/aggregate_package_logs.spack.py
+      - - time ./spack python ${SPACK_CI_PACKAGES_ROOT}/.ci/gitlab/scripts/common/aggregate_package_logs.spack.py
```
alalazo: 🤔

kwryankrattiger: blah, yeah, streams got crossed and bad revert

alalazo (dismissed):
Current state LGTM. From my point of view, we can merge as soon as @haampie ✔️ too

repos/spack_repo/builtin/build_systems/cmake.py
```diff
@@ -172,20 +154,16 @@ class CMakePackage(PackageBase):
     https://cmake.org/cmake/help/latest/
     """
 
+    #: Disable adding extra CMake args provided by listed depenedencies
```
haampie:
```suggestion
    #: List of package and virtual names for which CMake argument injection should be disabled
```

repos/spack_repo/builtin/build_systems/cmake.py
```diff
@@ -405,7 +383,27 @@ def std_args(pkg: PackageBase, generator: Optional[str] = None) -> List[str]:
...
+        disable_cmake_hints_from = getattr(pkg, "disable_cmake_hints_from", [])
+        for dep in pkg.spec.edges_to_dependencies():
+            # Skip packages without the callback
+            dep_pkg = dep.spec.package
+            if not hasattr(dep_pkg, "dependent_cmake_args"):
+                continue
+
+            # Skip disabled dependency cmake args
+            # Consider virtual names as well package names
+            disabled_for_pkg = dep.spec.package.name in disable_cmake_hints_from
```
haampie:
```suggestion
            disabled_for_pkg = dep_pkg.name in disable_cmake_hints_from
```

haampie: Well, can you just simplify to:

```python
if dep_pkg.name in disable_cmake_hints_from:
    continue

if any(v in disable_cmake_hints_from for v in dep.virtuals):
    continue
```

haampie (dismissed):
Still not a convinced that `python` is currently the best example given the interaction with `python-venv`, but I can live with it. Few small requests.

repos/spack_repo/builtin/build_systems/cmake.py
```diff
@@ -405,7 +383,27 @@ def std_args(pkg: PackageBase, generator: Optional[str] = None) -> List[str]:
             )
 
         _conditional_cmake_defaults(pkg, args)
-        _maybe_set_python_hints(pkg, args)
+
+        # Append extra hint/option arguments from dependencies
+        # TODO: Consider properties from virtual packages like Mpi
+        disable_cmake_hints_from = getattr(pkg, "disable_cmake_hints_from", [])
```
haampie, line 389: Just use `pkg.disable_cmake_hints_from`

kwryankrattiger: We can't necessarily assume that. `pkg` is a `PackageBase` and `std_args` is a static method. We don't know if it is actually a `CMakePackage` or not in this context. In practice I think it always will be, but the API does not necessarily require it.

haampie: Hm, fair. I think we may need `std_args` in Python packages that wrap a cmake build with whatever the latest popular build back end is.

repos/spack_repo/builtin/build_systems/cmake.py
```diff
@@ -405,7 +383,28 @@ def std_args(pkg: PackageBase, generator: Optional[str] = None) -> List[str]:
...
+            dep_pkg = dep.spec.package
+            if not hasattr(dep_pkg, "dependent_cmake_args"):
+                continue
+
+            # Skip disabled dependency cmake args
+            if dep_pkg.name in disable_cmake_hints_from:
+                continue
+
+            # Skip disabled virtual dependency cmake args
+            if any((v in disable_cmake_hints_from for v in dep.virtuals)):
```
haampie:
```suggestion
            if any(v in disable_cmake_hints_from for v in dep.virtuals):
```

alalazo (approved):

alalazo:
@spackbot run pipeline

kwryankrattiger:
CI Failures are unrelated to this change. This change is blocking progress on other work @alalazo @haampie are we okay to force merge this to unblock things?

alalazo:
Yeah, I'll go on force merging this - there are just a handful of known flaky jobs that went :x:

PR #3198, opened 2026-01-29
Cabana: Update arborx dependency conditions in package.py

JStewart28:
ArborX updated its API in version 2.0. Older versions of Cabana, which use the older ArborX API, still request the newest version of ArborX.


streeve (approved):
Note https://github.com/ECP-copa/Cabana/pull/814 (where we would have merged a long time ago but seem to have been stuck on a CI failure)

junghans (approved):

PR #3200, opened 2026-01-30
HDF5: Remove Java directory from source builds if we're not using it

tgamblin:
Some sites use Tenable, and Tenable isn't very smart. It will unfortunately flag the `java` subdirectory in HDF5 source checkouts, as a path like this:

```
spack-stage-hdf5-1.14.6-ll3jck7sgauntx7ryvol5cz5it3hruge/spack-src/java
```

looks like an old version of the Java browser plugin to Tenable. This is not Spack's fault or HDF5's fault -- Tenable should be smarter.

Still, we can at least help Spack users while we wait for Tenable to get smarter. The HDF5 build doesn't use this directory when Java is not enabled, and I suspect that most users are not building the Java bindings. So we can just remove the directory, so that old builds don't get flagged by Tenable.

### DIFF
--- a/repos/spack_repo/builtin/packages/hdf5/package.py
+++ b/repos/spack_repo/builtin/packages/hdf5/package.py
@@ -312,6 +312,12 @@ class Hdf5(CMakePackage):
                 "testpar/CMakeTests.cmake",
             )
 
+        # Remove the java subdirectory from source builds that aren't going to use it.
+        # The Java subdirectory unfortunately looks like a Java browser plugin to Tenable,
+        # and if it sits around it will cause users' machines to be erroneously flagged.
+        if not self.spec.satisfies("+java"):
+            shutil.rmtree("java")
+
     # The parallel compiler wrappers (i.e. h5pcc, h5pfc, etc.) reference MPI
     # compiler wrappers and do not need to be changed.
     # These do not exist on Windows.


tgamblin:
Maybe @cyrush, @alecbcs, or @becker33  can comment on this as I think it will be controversial.  Question is whether we should really be putting fixes in packages for IT tools used at different sites. Generally I hate this. I just think it will help users like @cyrush.

becker33 (changes requested):
I don't think this is an appropriate change to be pushing the maintenance burden onto Spack package maintainers.

tgamblin:
Others?

alecbcs:
I staged every concrete spec in my "default" programming environment on my laptop. Out of 166 specs in the concrete environment 10 of them have `java/` sub-directories in their src repositories that trigger tenable.

```
$ fd --type=directory "java"
spack-stage-libjpeg-turbo-3.0.4-os5br3mnhtufaf37og6blwgd7yfi34l3/spack-src/java/
spack-stage-berkeley-db-18.1.40-tm5l7psfhf7zaowle4d5auecqc27l22p/spack-src/examples/java/
spack-stage-mergiraf-0.13.0-m6lnpouzkh2p2ps5d7usouqoozjv74n6/spack-src/examples/java/
spack-stage-gettext-0.23.1-cgtakyzigrmj6utqshxu4gwqlwwmoqq3/spack-src/gettext-runtime/intl-java/
spack-stage-bison-3.8.2-hdz33ugyh7uazp6nkzzo3ocfhuonkum6/spack-src/examples/java/
spack-stage-git-2.45.3-w4xvkcqsgn3eflxfppduf7hwuzvl5qmg/spack-src/t/t4034/java/
spack-stage-emacs-30.2-xpgppr3cc3syydqc6cz4wygfklgaxeoi/spack-src/java/
spack-stage-python-3.13.11-sydj5zpbmj5qd2b7rgud2plqk2yxfmk7/spack-src/Android/testbed/app/src/main/java/
spack-stage-cmake-3.31.9-5regqpyv7dhsbonlqlfcuphxhbrv3zu6/spack-src/Tests/CoberturaCoverage/src/main/java/
spack-stage-cmake-3.31.9-5regqpyv7dhsbonlqlfcuphxhbrv3zu6/spack-src/Tests/JacocoCoverage/Coverage/src/main/java/
spack-stage-cmake-3.31.9-5regqpyv7dhsbonlqlfcuphxhbrv3zu6/spack-src/Tests/Java/
spack-stage-rust-1.92.0-idobdxtuoa47qeru6emywkb7wxm2x6y2/spack-src/vendor/capstone-sys-0.17.0/capstone/bindings/java/
```

I agree with @becker33 I think this would be a lot of effort to generalize this over the 8000+ packages in Spack.

tgamblin:
One other thought. We could:

1. Make the default stage directory name a configurable thing. 
2. Take the version out of the default stage directory name.

We currently pull the format from `spack.stage.compute_stage_name()`, and it includes the package version by default. IMO it doesn't really need to.  most people don't look at the stage directories -- we could just call it `"{name}-{hash}"` instead of `"{name}-{version}-{hash}"`.

This would fix the issue generally because tenable wouldn't see the version in the path. It would also make the default stage directory path a bit shorter, and since it's configurable you could change it back if you really need a descriptive stage name, e.g., I sometimes want to stage lots of versions of something to compare them.

This approach doesn't require any maintenance effort from packagers, and I think it fixes all the cases @alecbcs mentioned.  Thoughts?

cyrush:
I appreciate the proposed solution, but I don't think we should take such measures in the spack package. 

Also, I value the fact that the version name is in the source dirs. 

We know this instance of attack of the bad heuristic, but who knows what the future holds. 

I don't want to set a precedent for this being the way we live our lives.

alecbcs:
For Spack users globally who may be less connected to their IT department or security environment rules, I think the idea of a configurable stage format is a good one we should implement. I concur with Cyrus though, I find the version in the path to be helpful when debugging and think we should maintain the existing format as the default.

Continuing Cyrus' last point, I'm also going to continue to push on this specific issue with IT departments we interact with. It's hilarious that the combination of a semantic version and the word `java` in a path cause them to believe an old version of java is installed. Without pushing back and forcing them to buy/demand better software from their vendors I fear their next check, might be even more harmful for Spack / developers in general. (Like that time I couldn't install anything written in Go because they flagged executables with fewer than 3 character names.)

tgamblin:
Closing this as we do not plan to implement it.  We may make stage names configurable as in https://github.com/spack/spack-packages/pull/3200#issuecomment-3821996900, which would at least give affected people an easy out.

PR #3201, opened 2026-01-30
Add libsanitizer variant to GCC package

hseara:
Recent glibc releases (e.g. 2.42 on Arch Linux) expose build failures in older GCC versions due to libsanitizer incompatibilities. While multiple patches exist to address specific glibc changes, they are not always available or complete for all supported GCC releases.

This change introduces a +libsanitizer variant (enabled by default) to the GCC package, allowing users to explicitly disable libsanitizer when building older GCC versions on newer platforms. This provides a practical workaround for systems where libsanitizer currently prevents successful builds, without changing default behavior.

The variant maps directly to --enable-libsanitizer / --disable-libsanitizer at configure time and is intended to ease portability until upstream or Spack-side patches fully address these issues.


michaelkuhn:
@spackbot run pipeline

michaelkuhn (approved):
Works for me (Fedora 43, GCC 15).

PR #3210, opened 2026-01-30
fms: add 2025.04, 2026.01

mathomp4:
I noticed FMS released 2026.01 recently and that 2025.04 wasn't in here.

cc: @rem1776 in case FMS 2026 has any sort of "extra" dependencies I'm not aware of.

### DIFF
--- a/repos/spack_repo/builtin/packages/fms/package.py
+++ b/repos/spack_repo/builtin/packages/fms/package.py
@@ -20,6 +20,8 @@ class Fms(CMakePackage):
     license("LGPL-3.0-or-later")
 
     maintainers("AlexanderRichert-NOAA", "Hang-Lei-NOAA", "edwardhartnett", "rem1776", "climbfuji")
+    version("2026.01", sha256="aa481afbd4df3d09e42114a74ec5dc21bb85644335fbea378f48d0c1f22866d0")
+    version("2025.04", sha256="d9d91a8594c65d9d51928a9f4f4ec6ffaaa0d1d00a752ca32f2ea1f91a70d3f0")
     version("2025.03", sha256="17bdbac86bd3b1e5c1c3a5496c1495e8badcd5f671c659b17921532c990f014c")
     version(
         "2025.02.01", sha256="a481a0d49cf4ca06476849600c2da41fb25053e930a8fabe2add09c3d2fbee9c"
--- a/repos/spack_repo/builtin/packages/fms/package.py
+++ b/repos/spack_repo/builtin/packages/fms/package.py
@@ -101,8 +101,7 @@ class Fms(CMakePackage):
 
     variant(
         "precision",
-        values=any_combination_of("32", "64")
-        .with_default("none"),
+        values=any_combination_of("32", "64").with_default("none"),
         description="Build a version of the library with mixed precision ('none'), or 32 and/or 64 bit reals",
         when="@2025.04:",
     )
--- a/repos/spack_repo/builtin/packages/fms/package.py
+++ b/repos/spack_repo/builtin/packages/fms/package.py
@@ -103,11 +103,20 @@ class Fms(CMakePackage):
     # and allow back specified precision variants for 2026.0x and later.
     # variant(
     # "precision",
-    # values=any_combination_of("32", "64").with_default("none"),
-    # description="Build a version of the library with mixed precision ('none'), or 32 and/or 64 bit reals",
+    # values=any_combination_of("32", "64", "mixed").with_default("mixed"),
+    # description="Build a version of the library with mixed precision ('mixed' or 'none'), or 32 and/or 64 bit reals",
     # when="@2026.0x:",
     # )
 
+    variant(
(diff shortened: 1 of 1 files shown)


climbfuji (approved):
Defer to @rem1776 for changes to build variants.

mathomp4:
> Defer to @rem1776 for changes to build variants.

Ohh. Maybe there are some. I just saw that the libyaml handling in FMS 2026.01 CMake is  [different](https://github.com/NOAA-GFDL/FMS/compare/2025.04...2026.01#diff-1e7de1ae2d059d21e1dd75d5812d5a34b0222cef273b7c3a2af62eb747f9d20a) 

I'm not too conversant on PkgConfig and CMake. Let me do a quick test build and see.

---

Oh, and note to @climbfuji (and @rem1776) I hope to soon figure out a good patch to work around https://github.com/NOAA-GFDL/FMS/issues/1809 for ifx until a fix can go in in 2026.02+.

I think it's pretty simple. Just if ifx 2025.3 remove that one line in a cmake file.

But for now, I wanted to get this in.

rem1776:
@mathomp4 @climbfuji

In 2025.04 there were some slight changes to the real precision option.

We didn't modify the existing 32/64 bit options, but now if neither are specified it will build with r8 defaults and exclude the `r4/r8` suffix in the library name. I don't think that will break anything with the spack configuration as is.

For both versions there shouldn't be any changes in dependencies, we just modified the cmake build to use pkg-config to find libyaml in the 2026.01 release. I hope that is not problem for the spack build, usually just comes down to if the PKG_CONFIG_PATH is set properly.

rem1776 (approved):

mathomp4:
@rem1776 Uh oh. I get a lot of:

```
  >> 49      CMake Error at CMakeLists.txt:615 (target_link_libraries):
     50        Target "testLibs" links to:
     51
     52          FMS::fms
     53
     54        but the target was not found.  Possible reasons include:
     55
     56          * There is a typo in the target name.
     57          * A find_package call is missing for an IMPORTED target.
     58          * An ALIAS target is missing.
```
for various targets.

I'm drafting this for now until I can figure it out.

mathomp4:
Update. I can get `precision=none` to work (aka the mixed-precision).

But when I try and do any of:

* `spack install fms@2026.01 precision=32`
* `spack install fms@2026.01 precision=64`
* `spack install fms@2026.01 precision=32,64`

they all fail like:
```
  >> 48      CMake Error at CMakeLists.txt:615 (target_link_libraries):
     49        Target "testLibs" links to:
     50
     51          FMS::fms
     52
     53        but the target was not found.  Possible reasons include:
     54
     55          * There is a typo in the target name.
     56          * A find_package call is missing for an IMPORTED target.
     57          * An ALIAS target is missing.
```

I try this some more soon.

mathomp4:
Oh. wait. @rem1776 This might be broken until https://github.com/NOAA-GFDL/FMS/pull/1787 is merged in, right?

So maybe for 2025.04 and 2026.01 I should say only "precision: none" is allowed?

And, huh, maybe that should be better expressed as "precision: mixed"?

rem1776:
@mathomp4 Yeah that PR should fix it. none/mixed sounds like a good option to use in the meantime.

mathomp4:
Well, since only "one" precision is possible for current FMS, I just removed the `variant` bit.  The annoying thing is that that means `spack info fms` doesn't really tell users why they don't have that option anymore...

I might see if @tldahlgren or someone from Spack has better thoughts.

tldahlgren:
> Well, since only "one" precision is possible for current FMS, I just removed the `variant` bit. The annoying thing is that that means `spack info fms` doesn't really tell users why they don't have that option anymore...
> 
> I might see if @tldahlgren or someone from Spack has better thoughts.

I can't think of anything you aren't already doing here.

tldahlgren (dismissed):
Confirmed the new version sha56s.

tldahlgren:
Ugh. At this point, IIRC (@becker33), we don't support version ranges on variants.

tldahlgren:
> Ugh. At this point, IIRC (@becker33), we don't support version ranges on variants.

I'm not sure if adding `conflicts` is an option or worth it.

mathomp4:
I'm redrafting at the moment. Some people have said `mixed` would be better. One moment...

mathomp4:
@tldahlgren (and @becker33 ?) Can you take another look with a Spack eye? A user thought `mixed` is better than `none`.

Of course, when we can use `any_combination_of` we will get `none` for "free" but for now we don't need it it seems.

repos/spack_repo/builtin/packages/fms/package.py
becker33, line 134: Because you added the constraint on the variant above, you need to remove this conflict.

becker33 (changes requested):
One request and a style fix and this should be good

mathomp4:
I'm closing this in favor of #4737

PR #3222, opened 2026-02-02
acts: add v45.0.0

stephenswat:
This commit adds version 45.0.0 of the ACTS package. Encodes a new dependency on EDM4hep version 1.0, and disallows this version of EDM4hep for older ACTS versions. Also encodes a newer dependency on nlohmann JSON.

### DIFF
--- a/repos/spack_repo/builtin/packages/acts/package.py
+++ b/repos/spack_repo/builtin/packages/acts/package.py
@@ -50,6 +50,7 @@ class Acts(CMakePackage, CudaPackage):
 
     # Supported Acts versions
     version("main", branch="main")
+    version("45.0.0", commit="92ab57740f8e875555ea28f542844ac1eb5db65b")
     version("44.4.0", commit="a05c35a14b39a461925d11de12ccd2da5e38b3d1")
     version("44.3.0", commit="d4c630145d5050dd2edc58f1de0c872caff23dd8")
     version("44.2.0", commit="c3d440eb1e441fcd15995b8af87ea1497e0cc126")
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/acts/package.py
wdconinc, line 423: New minimum.

repos/spack_repo/builtin/packages/acts/package.py
```diff
@@ -434,6 +435,8 @@ class Acts(CMakePackage, CudaPackage):
     depends_on("edm4hep @0.4.1:", when="+edm4hep")
     depends_on("edm4hep @0.7:", when="@25: +edm4hep")
     depends_on("edm4hep @0.10.5:", when="@42: +edm4hep")
+    depends_on("edm4hep @:0", when="@:44 +edm4hep")
+    depends_on("edm4hep @1:", when="@45: +edm4hep")
```
wdconinc: There is still fallback support for older versions. We may want to propagate that support so users can decouple edm4hep upgrades from acts upgrades within a limited timeframe. Right now it would not be possible to validate an acts upgrade without adding an edm4hep upgrade at the same time.

repos/spack_repo/builtin/packages/acts/package.py
stephenswat: Added.

repos/spack_repo/builtin/packages/acts/package.py
stephenswat: I removed the `depends_on("edm4hep @1:", when="@45: +edm4hep")` line.

wdconinc (approved):

PR #3225, opened 2026-02-02
py-numpy: fix amdblis detection logic

amd-toolchain-support:
The py-numpy and py-scipy packages in Spack use the blas_lapack_pkg_config function defined in the numpy recipe for BLAS/LAPACK detection. This function returns the names of the BLAS and LAPACK libraries that pkg-config should search for.

As of version 5.1, the amdblis package, when built in multi-threaded mode, defines its pkg-config metadata in blis-mt.pc instead of blis.pc. Consequently, the library name mapping in the above function needs to be updated accordingly.

This PR adds the necessary logic to handle this change.

### DIFF
--- a/repos/spack_repo/builtin/packages/py_numpy/package.py
+++ b/repos/spack_repo/builtin/packages/py_numpy/package.py
@@ -305,9 +305,9 @@ class PyNumpy(PythonPackage):
         # Handle AMD BLIS: use multithreaded pkg-config name for @5.1: when threading is enabled
         if spec["blas"].name == "amdblis":
             blas = "blis"
-            if spec["amdblis"].satisfies('@5.1:') and (
-                spec["amdblis"].satisfies('threads=openmp') or
-                spec["amdblis"].satisfies('threads=pthreads')
+            if spec["amdblis"].satisfies("@5.1:") and (
+                spec["amdblis"].satisfies("threads=openmp")
+                or spec["amdblis"].satisfies("threads=pthreads")
             ):
                 blas = "blis-mt"
 
--- a/repos/spack_repo/builtin/packages/py_numpy/package.py
+++ b/repos/spack_repo/builtin/packages/py_numpy/package.py
@@ -302,10 +302,10 @@ class PyNumpy(PythonPackage):
         if spec["blas"].name == ["blis"]:
             blas = "blis"
 
-        # Handle AMD BLIS: use multithreaded pkg-config name when OpenMP is enabled
+        # Handle AMD BLIS: use multithreaded pkg-config name for @5.1: when OpenMP is enabled
         if spec["blas"].name == "amdblis":
             blas = "blis"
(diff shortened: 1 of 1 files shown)


becker33 (approved):

becker33 (dismissed):

amd-toolchain-support:
The updates now add a version guard for the multithreaded pkg-config name in amdblis.

repos/spack_repo/builtin/packages/py_numpy/package.py
```diff
@@ -299,9 +299,15 @@ def blas_lapack_pkg_config(self) -> Tuple[str, str, str]:
         if spec["lapack"].name in ["intel-mkl", "intel-parallel-studio", "intel-oneapi-mkl"]:
             lapack = self._blas_lapack_pkg_config_mkl(spec["lapack"])
 
-        if spec["blas"].name in ["blis", "amdblis"]:
+        if spec["blas"].name == ["blis"]:
```
adamjstewart: This is broken, either use `in` or remove the square brackets.

adamjstewart (changes requested):

rgommers (commented):
I'd like to check whether gating this on `threads=openmp` is correct. It looks like there are three build modes:

- single-threaded, uses `blis.pc`
- pthreads, uses `blis-mt.pc`  
- openmp, uses `blis-mt.pc`  

See:

https://github.com/amd/blis/blob/9734fc18cc51abd06b99862820d7ebc3e1c35e51/CMakeLists.txt#L240

https://github.com/amd/blis/blob/9734fc18cc51abd06b99862820d7ebc3e1c35e51/CMakeLists.txt#L1310-L1316

So the question is whether there's a `threads=pthreads` build mode to take into account.

amd-toolchain-support:
> I'd like to check whether gating this on `threads=openmp` is correct. It looks like there are three build modes:
> 
> * single-threaded, uses `blis.pc`
> * pthreads, uses `blis-mt.pc`
> * openmp, uses `blis-mt.pc`
> 
> See:
> 
> https://github.com/amd/blis/blob/9734fc18cc51abd06b99862820d7ebc3e1c35e51/CMakeLists.txt#L240
> 
> https://github.com/amd/blis/blob/9734fc18cc51abd06b99862820d7ebc3e1c35e51/CMakeLists.txt#L1310-L1316
> 
> So the question is whether there's a `threads=pthreads` build mode to take into account.

Thanks @rgommers , you are right that `threads=pthreads` build mode also needs to be included. Corresponding changes have been pushed.

repos/spack_repo/builtin/packages/py_numpy/package.py
amd-toolchain-support: Thanks @adamjstewart, recommended changes are pushed with recent commit.

rgommers (approved):
LGTM now, thanks @amd-toolchain-support

adamjstewart (approved):

PR #3229, opened 2026-02-02
Tfmodisco: new package

Markus92:
This package adds the py-modisco package, which implements the TF-MoDISco algorithm ("Transcription Factor Motif Discovery from Importance Scores").

It has some dependencies I also added. Especially the `leidenalg` stuff was interesting to debug to say the least. I cherrypicked two patches: 

1. First one is for `libleidenalg < 0.11` so it does correct C includes (it mixed up quotes and brackets a bit).
2. Secons is for `igraph`. I took a patch from upstream for libxml >= 2.12 support. It is included in the package tree as the original commit fixes both source code as well as modifies the changelog, but the patch does not apply to the older `CHANGELOG.md` files. I just took that part manually out of the patch file. Maybe in a future version of Spack we can pass something like `patch(..., exclude="CHANGELOG.md") but this was easiest for now.

The other igraph modifications are follows: given that the original package.py kinda hardcodes OpenBLAS in the cmake arguments, I made that requirement a bit more explicit. In theory it should work with other BLAS/Lapack implementations, but I didn't have any handy to test them. Second, the requirements are set for the build system but it just doesn't make sense to me, as the requirements are for both CMake and Autotools. CMake is the recommended build system, so maybe it's time to remove Autotools later.


Markus92:
@spackbot fix style

repos/spack_repo/builtin/packages/igraph/package.py
```diff
@@ -31,14 +31,20 @@ class Igraph(CMakePackage, AutotoolsPackage):
...
+    depends_on("python", type="build")
+
+    # with when("build_system=cmake"):
+    depends_on("arpack-ng")
+    depends_on("blas")
+    depends_on("glpk+gmp@4.57:")
+    depends_on("gmp")
+
+    # igraph needs blas and lapack
+    depends_on("openblas")
```
becker33:
```suggestion
    depends_on("lapack")
```
From the prior commits it doesn't look like there's a restriction on which blas/lapack to use.

repos/spack_repo/builtin/packages/igraph/package.py
becker33, line 22: If we deprecated the `0.7` release we can get rid of the autotools build soon -- it doesn't look like anything is relying on it, but I'll leave that up to you whether to bother.

repos/spack_repo/builtin/packages/libleidenalg/package.py
```diff
@@ -0,0 +1,33 @@
...
+    maintainers("Markus92")
+
+    license("GPL-3.0-or-later", checked_by="Markus92")
+
+    version("0.10.0", sha256="ae265fb718e2233bfd01e3bc9679d9bed53a182e4cb13dbb12b49e6e92105cc7")
+
+    depends_on("c", type="build")
+    depends_on("cxx", type="build")
+
+    depends_on("igraph@0.10", when="@0.10:0.11")
```
becker33, line 27:
```suggestion
    depends_on("igraph")
    depends_on("igraph@0.10", when="@0.10:0.11")
```
I think a catchall dependency here will make new versions more likely to work when they're added.

repos/spack_repo/builtin/packages/py_leidenalg/package.py
```diff
@@ -0,0 +1,42 @@
...
+
+    version("0.10.0", sha256="a3829dceb3d9198a0489891d471f2abe8ffd5a3e8ddc8c298543d16caa8d7f19")
+
+    depends_on("c", type=("build"))
+    depends_on("cxx", type=("build"))
+
+    depends_on("python@3.8:", type=("build", "run"))
+    depends_on("py-setuptools@45:", type=("build"))
+    depends_on("py-igraph@0.10.0:0.10", type=("build", "run"))
+    depends_on("igraph@0.10.0:0.10", type=("build"))
```
becker33:
```suggestion
    depends_on("py-igraph@0.10", type=("build", "run"))
    depends_on("igraph@0.10", type=("build"))
```
This is equivalent and simpler.

repos/spack_repo/builtin/packages/py_modisco/package.py
```diff
@@ -0,0 +1,38 @@
...
+    depends_on("py-scikit-learn@1.0.2:", type=("build", "run"))
+    depends_on("py-leidenalg@0.8.10:", type=("build", "run"))
+    depends_on("py-igraph@0.9.11:", type=("build", "run"))
+    depends_on("py-tqdm@4.38.0:", type=("build", "run"))
+    depends_on("py-pandas@1.4.3:", type=("build", "run"))
+    depends_on("py-logomaker@0.8:", type=("build", "run"))
+    depends_on("py-h5py@3.7.0:", type=("build", "run"))
+    depends_on("py-hdf5plugin", type=("build", "run"))
+    depends_on("py-memelite", type=("build", "run"))
+    depends_on("py-jinja2", type=("build", "run"))
```
becker33: [optional]
```suggestion
    with default_args(type=("build", "run")):
        depends_on("py-numpy@1.21.5:")
        depends_on("py-scipy@1.6.2:")
        depends_on("py-numba@0.53.1:")
        depends_on("py-scikit-learn@1.0.2:")
        depends_on("py-leidenalg@0.8.10:")
        depends_on("py-igraph@0.9.11:")
        depends_on("py-tqdm@4.38.0:")
        depends_on("py-pandas@1.4.3:")
        depends_on("py-logomaker@0.8:")
        depends_on("py-h5py@3.7.0:")
        depends_on("py-hdf5plugin")
        depends_on("py-memelite")
        depends_on("py-jinja2")
```

becker33 (changes requested):

repos/spack_repo/builtin/packages/igraph/package.py
Markus92: Thanks for this! It will definitely need both `blas` and `lapack`. I'll see if I can test this on a system with a non-openblas implementation.

repos/spack_repo/builtin/packages/py_modisco/package.py
Markus92: I like it :) This was written as part of internal training on packaging for Spack. Maybe I can make a PR to update the default Python template?

repos/spack_repo/builtin/packages/igraph/package.py
Markus92: I've tested with openblas and intel mkl, builds/links fine with both so this specific requirement can be removed. Thanks for the suggestion!

becker33 (changes requested):
Lines 39 and 44 are duplicates now.

becker33 (approved):

PR #3231, opened 2026-02-02
bart: update to version 0.9.00

Markus92:
Adds bart version 0.9.00

Updated the patch for the makefile and also taking the new naming of the environment variable into account.

### DIFF
--- a/repos/spack_repo/builtin/packages/bart/package.py
+++ b/repos/spack_repo/builtin/packages/bart/package.py
@@ -41,7 +41,7 @@ class Bart(MakefilePackage, CudaPackage, PythonExtension):
 
     # patch to set path to bart
     patch("bart_path-0.5.00.patch", when="@0.5.00")
-    patch("bart_path-0.6.00.patch", when="@0.6.00")
+    patch("bart_path-0.6.00.patch", when="@0.6.00:0.7")
     patch("bart_path-0.9.00.patch", when="@0.9.00")
 
     depends_on("libpng")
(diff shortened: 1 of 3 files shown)


Markus92:
@spackbot fix style

repos/spack_repo/builtin/packages/bart/package.py
```diff
@@ -32,11 +36,13 @@ class Bart(MakefilePackage, CudaPackage):
...
     # patch to fix Makefile for openblas and cuda
     patch("Makefile.patch", when="@:0.6.00")
-    patch("Makefile-0.7.00.patch", when="@0.7.00:")
+    patch("Makefile-0.7.00.patch", when="@0.7.00")
+    patch("Makefile-0.9.00.patch", when="@0.9.00")
 
     # patch to set path to bart
     patch("bart_path-0.5.00.patch", when="@0.5.00")
-    patch("bart_path-0.6.00.patch", when="@0.6.00:")
+    patch("bart_path-0.6.00.patch", when="@0.6.00")
```
becker33: Shouldn't this be `when="@0.6:0.7"` to preserve the previous behavior?

becker33 (changes requested):

repos/spack_repo/builtin/packages/bart/package.py
Markus92: You are right, must've missed that one. Put it back, thanks!

becker33 (approved):

PR #3234, opened 2026-02-03
py-scipy: add lowerbound for meson version

haampie:
Meson is pulled in through `py-python-meson`, but without any version restrictions.

Just add it as a direct dep with the bounds from https://github.com/scipy/scipy/blob/main/meson.build.

### DIFF
--- a/repos/spack_repo/builtin/packages/py_meson_python/package.py
+++ b/repos/spack_repo/builtin/packages/py_meson_python/package.py
@@ -18,6 +18,7 @@ class PyMesonPython(PythonPackage):
 
     license("MIT")
 
+    version("0.19.0", sha256="9959d198aa69b57fcfd354a34518c6f795b781a73ed0656f4d01660160cc2553")
     version("0.18.0", sha256="c56a99ec9df669a40662fe46960321af6e4b14106c14db228709c1628e23848d")
     version("0.16.0", sha256="9068c17e36c89d6c7ff709fffb2a8a9925e8cd0b02629728e5ceaf2ec505cb5f")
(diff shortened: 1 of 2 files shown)


adamjstewart:
Seems fine with me. Can you add a `# meson.build` comment on the line before so I know which file to check when adding new versions?

rgommers (approved):
LGTM, thanks.

adamjstewart (dismissed):

adamjstewart:
@eli-schwartz do you want to give this a quick review before I merge?

adamjstewart (approved):

eli-schwartz (approved):

eli-schwartz:
This seems reasonable to me. Although meson is depended on indirectly via py-meson-python, the version constraint is only a minimum "to ensure the build backend wiring is compatible". It is entirely normal for projects to depend on a newer version in meson.build.

Although pip install in an isolated build venv generally installs the latest version, and you can "get away with" not requiring a suitable version (probably), this is not reliable or robust when the version from an existing environment may be used, and users for example decide that they wish to upgrade scipy, and only scipy. Specifying the direct build constraints is therefore necessary in order to prevent build failures; this PR prevents build failures, so :+1:

haampie:
> Although pip install in an isolated build venv generally installs the latest version

Yup, this is what you get when you haven't bumped `spack-packages` in a while, and when spack reuses old stuff ;)

haampie:
@spackbot run pipeline

haampie:
Apparently this change causes the tutorial pipeline to get stuck concretizing for 1h+, six times in a row :)

That should not be a blocker, it's just that that stack is configured incorrectly. So would suggest to force merge this after #3255 is closed.

PR #3235, opened 2026-02-03
nccl: fix cuda version requirements

Markus92:
Got some compilation errors with nccl 2.29 and cuda 11.8, which prompted me to go over all the release notes to add proper CUDA versioning.

Didn't do every single minor version (some versions claim they support CUDA 11.0, 11.4 and 12.0, but how about the rest?) but some others advertise it's all based on major version. This should tighten it down considerably though.


msimberg (approved):
Looks good to me, thanks!

I think approximately-correct version ranges is already a nice improvement. The ranges can always be further tightened later.

PR #3237, opened 2026-02-03
memtailor: add v1.2

d-torrance:
Also update the version where we require googletest < 1.17.0. Although 1.1 built with the C++17 flags that googletest now requires, it didn't add them during the configure stage, so detection of googletest failed.


RikkiButler20 (approved):
Looks good to me

PR #3239, opened 2026-02-03
Precice: Added gcc@:9.2 conflict

emwjacobson:
Since version 3.0.0, gcc 9.3: has been a requirement.
https://github.com/precice/precice/blob/v3.0.0/cmake/CheckSTL.cmake#L5

Maintainers are @MakisH and @fsimonis. Let me know any comments/questions/concerns, though I'd imagine this would be a fairly easy fix...

### DIFF
--- a/repos/spack_repo/builtin/packages/precice/package.py
+++ b/repos/spack_repo/builtin/packages/precice/package.py
@@ -125,6 +125,7 @@ class Precice(CMakePackage):
     conflicts("%apple-clang@:5")
     conflicts("%clang@:3.7")
     conflicts("%intel@:16")
+    conflicts("%gcc@:9.2", when="@3.0.0:")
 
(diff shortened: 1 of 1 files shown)


fsimonis (approved):
This is correct. Thanks for the fix!
 Our baselines generally need some restructuring.

repos/spack_repo/builtin/packages/precice/package.py
MakisH, line 123: I guess then we can already remove this comment, which will be always outdated:


```suggestion
```

We already require C++17, and we are about to move the requirement to C++20.

fsimonis: Out of scope, though. Let's clean this up in one go at a later time.

MakisH (approved):

becker33 (approved):

PR #3245, opened 2026-02-04
py-matplotlib: relax py-meson-python constraints

adamjstewart:


rgommers (approved):
Thanks @adamjstewart, that looks correct indeed:

https://github.com/matplotlib/matplotlib/blob/eb7c4cebdc589f894c165de1852f63791e0edb3f/pyproject.toml#L50

I quickly refreshed my memory - the reason 0.17.0 had an issue was because of the symlink support introduced in that release, which didn't quite agree with `matplotlib`'s use of symlinks (xref https://github.com/mesonbuild/meson-python/pull/713). The fix for that went into `meson-python` 0.18.0

becker33 (approved):

PR #3249, opened 2026-02-04
ci: disable tutorial pipeline

haampie:
There are three issues with the pipeline:

* `x` and `x cflags=-O3` are unified, but the tutorial expects these to
  be concretized separately
* The concretization time is very long, and sometimes times out after 1h
* It no longer works with `spack@develop` due to changes to compiler
  mixing.

All these issues are caused by `unify: when_possible` in an attempt to
bootstrap GCC 12.

Soon enough there'll be a proper, faster way to do this in Spack. For
now, disable this stack so we can bump Spack.


alalazo (approved):

becker33:
I think we need to revert this -- this is going to lead to a very broken tutorial in the long run, instead of a tutorial that has a single spec we're not happy with.

PR #3251, opened 2026-02-04
neovim: Fix utf8proc dependency

RemiLacroix-IDRIS:
Version 2.10.0 is needed for neovim 0.11+ to compile.

### DIFF
--- a/repos/spack_repo/builtin/packages/neovim/package.py
+++ b/repos/spack_repo/builtin/packages/neovim/package.py
@@ -110,7 +110,7 @@ class Neovim(CMakePackage):
         depends_on("tree-sitter@0.20.9:0.25")
     with when("@0.11:"):
         depends_on("cmake@3.16:", type="build")
-        depends_on("utf8proc", type="link")
+        depends_on("utf8proc@2.10.0", type="link")
         depends_on("tree-sitter@0.25", type="link", when="@:0.11.5")
         depends_on("tree-sitter@0.25:", type="link")
 


repos/spack_repo/builtin/packages/neovim/package.py
```diff
@@ -110,7 +110,7 @@ class Neovim(CMakePackage):
         depends_on("tree-sitter@0.20.9:0.25")
     with when("@0.11:"):
         depends_on("cmake@3.16:", type="build")
-        depends_on("utf8proc", type="link")
+        depends_on("utf8proc@2.10.0", type="link")
```
alecbcs, line 113:
```suggestion
        depends_on("utf8proc@2.10.0:", type="link")
```

Unless I'm misunderstanding. We likely want to leave this as an open-ended range to allow newer versions of `utf8proc` to be used as dependencies.

alecbcs (changes requested):
@RemiLacroix-IDRIS one quick suggestion, otherwise this looks good to me.

Note: @trws and @albestro it looks like this constraint on `utf8proc` is present in the [cmake.deps/deps.txt](https://github.com/neovim/neovim/blob/v0.11.0/cmake.deps/deps.txt#L34) file for v0.11.0.

repos/spack_repo/builtin/packages/neovim/package.py
albestro: I'd agree with you in general, but with neovim it is probably better to be strict and require an exact version.

I'm saying this because in the past I've seen that they pin specific version of dependencies to avoid problems, and they seems to support just the version they build with their cmake-based build system.

So, I wouldn't go hard on this, despite seems a good idea in general.

albestro (approved):
Thanks @RemiLacroix-IDRIS!

repos/spack_repo/builtin/packages/neovim/package.py
RemiLacroix-IDRIS: TBH, it was actually a typo, I meant to use the open range as suggested by @albestro but it does match what they were using in their deps file at the time neovim v0.11.6 was released.

I am good with both way, just let me know whatever you prefer.

alecbcs: I'll defer to @albestro here since he's the package maintainer.

alecbcs (approved):

PR #3254, opened 2026-02-04
Update byobu and fix newt

ebagrenrut:
- Update/fix `byobu`
    - Added the latest `byobu` release, 6.13
    - Enabled Python support for `byobu`, which allows the `byobu-config`
      command to work. If the user does not care about `byobu-config` and
      really does not want `byobu` to have a dependency on `python`, they
      can build with `~python`.
    - Added several missing build- and run-time dependencies. The `byobu`
      `configure` script and `Makefile` rely on a bunch of external
      utilities (`awk`, `find`, `grep`, several from `coreutils`, `man`, and
      `sed`). I suspect that these went unnoticed previously as builds were
      happening on systems/in environments with a bunch of things already
      installed, but in a container with a minimal image, I was unable to
      build and run `byobu`.
    - `url_for_version` is defined so that the source for `byobu` 6+ can be
      downloaded from GitHub. The source for 5.x is still only in LaunchPad.

- Update/fix `newt`
    - Add the latest `newt` release, 0.52.25
    - Add a `python` variant, default to `False`, that, when `True`, causes
      `newt`'s `snack` Python module to be built and installed. This module
      is needed by `byobu+python`. Having reviewed the `newt` source for
      past versions, it was clear that making it properly build `snack` for
      versions 0.52.20 and earlier was going to require a lot of
      hacking/patching of `newt`'s source as those versions look only in
      `/usr` for Python. This issue was fixed in `newt` 0.52.21, which added
      a `--with-python` configure option that can be used to provide the
      python command to be used.
    - Added a dependency on `gettext`. `newt` needs `libintl`, and if we do
      not provide it from Spack, the build will try to get it from the OS.
      If it doesn't find any provider of `libintl`, the `configure` fails. A
      `setup_build_environment` function is now defined to ensure that
      `configure` will find the Spack-provided `gettext`.
    - A `setup_dependent_run_environment` function is now defined that adds
      the directory containing the `snack` module to `PYTHONPATH`. This
      allows dependent packages that need `snack` to find it. This function
      is only called if `newt`'s `python` variant is `True`.
    - Include a patch for `newt` 0.52.25 (and possibly newer, should there be 
      another release), that forces the installation of `snack` into `newt`'s 
      `PREFIX` when the `python` variant is `True`. The build for 0.52.25 was 
      changed to find the `site-packages` directory of the provided Python 
      and install `snack` there. This is rather un-Spacky, so the patch reverts 
      this behavior.

@matthiasdiener Please let me know if you're still interested in maintaining the `byobu` Spack package.

### DIFF
--- a/repos/spack_repo/builtin/packages/byobu/package.py
+++ b/repos/spack_repo/builtin/packages/byobu/package.py
@@ -13,7 +13,7 @@ class Byobu(AutotoolsPackage):
 
     homepage = "https://www.byobu.org/"
 
-    maintainers("matthiasdiener")
+    maintainers("ebagrenrut")
 
     license("GPL-3.0-or-later")
 
--- a/repos/spack_repo/builtin/packages/byobu/package.py
+++ b/repos/spack_repo/builtin/packages/byobu/package.py
@@ -11,16 +11,53 @@ from spack.package import *
 class Byobu(AutotoolsPackage):
     """Byobu: Text-based window manager and terminal multiplexer."""
 
-    homepage = "https://www.byobu.co/"
-    url = "https://launchpad.net/byobu/trunk/5.123/+download/byobu_5.123.orig.tar.gz"
+    homepage = "https://www.byobu.org/"
 
     maintainers("matthiasdiener")
 
     license("GPL-3.0-or-later")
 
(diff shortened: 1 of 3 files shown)


matthiasdiener (approved):
Thanks @ebagrenrut ! I currently don't have time to maintain these packages, feel free to add yourself as a maintainer.

becker33 (approved):

PR #3253, opened 2026-02-04
ci: bump spack commit

haampie:
Should include a few more performance improvements

PR #3255, opened 2026-02-05
Revert "ci: disable tutorial pipeline (#3249)"

becker33:
This reverts commits c3ce4ac100fe65aa817a09e4fe000d427e1ce155 and 8d0701f2a54ac40f42a8f36335ddc00fb5d000bb.

~~We need the tutorial pipeline to keep running in CI. Yes, it elides the differences between two of the zlib-ng specs and that's not ideal. Yes, it will be better once we merge the concretization groups PR. But we still need to keep building it for now so that errors don't accumulate.~~

The tutorial stack as used in the tutorial is not currently buildable in CI, but the old version with an external gcc is buildable to ensure things don't break.

### DIFF
--- a/.ci/gitlab/configs/linux/ci.yaml
+++ b/.ci/gitlab/configs/linux/ci.yaml
@@ -4,7 +4,7 @@ ci:
       before_script-:
       # Test package relocation on linux using a modified prefix
       # This is not well supported on MacOS (https://github.com/spack/spack/issues/37162)
-      - - spack config add "config:install_tree:projections:${SPACK_JOB_SPEC_PKG_NAME}/${SPACK_JOB_SPEC_DAG_HASH}:'morepadding/{architecture.platform}-{architecture.target}/{name}-{version}-{hash}'"
+      - - spack config add "config:install_tree:projections:${SPACK_JOB_SPEC_PKG_NAME}:'morepadding/{architecture.platform}-{architecture.target}/{name}-{version}-{hash}'"
   - match_behavior: first
     submapping:
     - match:
--- a/.ci/gitlab/.gitlab-ci.yml
+++ b/.ci/gitlab/.gitlab-ci.yml
@@ -590,7 +590,7 @@ tools-sdk-build:
 
 tutorial-generate:
   extends: [ ".tutorial", ".generate-x86_64"]
-  image: ghcr.io/spack/tutorial-ubuntu-22.04:main
+  image: ghcr.io/spack/tutorial-ubuntu-22.04:v2023-10-30
 
 tutorial-build:
   extends: [ ".tutorial", ".build" ]
--- a/.ci/gitlab/.gitlab-ci.yml
+++ b/.ci/gitlab/.gitlab-ci.yml
@@ -10,7 +10,7 @@ variables:
   PROTECTED_MIRROR_FETCH_DOMAIN: "s3://spack-binaries"
   PROTECTED_MIRROR_PUSH_DOMAIN: "s3://spack-binaries"
   SPACK_CI_DISABLE_STACKS:
-    value: /^.*(cray|darwin|tutorial).*$/
+    value: /^.*(cray|darwin).*$/
     expand: false
 
 default:
--- a/.ci/gitlab/configs/win64/ci.yaml
+++ b/.ci/gitlab/configs/win64/ci.yaml
@@ -19,7 +19,7 @@ ci:
 
       script::
       - spack.ps1 env activate --without-view ${SPACK_CONCRETE_ENV_DIR}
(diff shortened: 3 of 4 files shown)


alecbcs (dismissed):
I agree with Greg's assessment here. Although it's not building a perfect tutorial stack, it's useful to continue building these packages to in the interim to ensure they continue working for the tutorial.

kwryankrattiger (dismissed):
Agreed. I was a bit surprised to see it disabled

tgamblin (dismissed):

alalazo:
For context, this was disabled because the packages were built with a mix of `gcc` and `clang` in the `netlib-scalapack` matrix due to `unify:when_possible`. This results in pipelines that are always :x: in `spack/spack` and in specs that are not what they meant to be for the tutorial.

The `tutorial` pipeline is fixed in https://github.com/spack/spack-packages/pull/3175 but that PR needs https://github.com/spack/spack/pull/51891 to work

alalazo:
@kwryankrattiger Is there a way to mark a pipeline `xfail` so that it doesn't result in a :x: pushed to PR status?

haampie:
@becker33 I would not merge this because it blocks bumping the spack commit.

haampie (changes requested):
I agree with @alalazo's `xfail` suggestion. Greg's reason to revert skips over the other reason this was merged.

I think it's more important to use latest Spack in this repo (#3253), especially after concretizer related changes, than it is to build the tutorial stack. It's not for lack of trying to make both work: https://spackpm.slack.com/archives/CJUC2P5M2/p1769612032952209, it's just practically impossible (either the concretizer gets stuck for ~2h, or it runs out of memory, or it produces incorrect results).

haampie:
Hm, no, I take back that `xfail` is a good solution, disabling it is best.

For example the tutorial pipeline continues to block improvements to packages: https://github.com/spack/spack-packages/pull/3234. Concretization consistently times out after 1h+ there.

tgamblin:
> For context, this was disabled because the packages were built with a mix of gcc and clang in the netlib-scalapack matrix due to unify:when_possible. This results in pipelines that are always ❌ in spack/spack and in specs that are not what they meant to be for the tutorial.

This is a way better reason to disable and/or `xfail` the pipeline than the reasons outlined in #3249.  Or maybe bullet 3 in #3249 is trying to address this?  It's not as clear there as how @alalazo explains it above. I don't think others got the full context either.

I am ok with keeping this disabled *if* we can:
1. Keep the tutorial alive in #3175 and keep rebasing that (via https://github.com/spack/spack/pull/51891) on `spack-packages/develop` until we get this fixed.
2. Prioritize https://github.com/spack/spack/pull/51891.

Is that workable?  I agree that it is important to keep the tutorial alive -- updating it before conferences is important.  Thinking particularly for ISC -- if we do not reinstate the pipeline well before that, then getting the tutorial working for ISC is going to be painful.

For users who want to do the tutorial, the build cache will live on, while we work out the kinks here -- in that people doing the tutorial via our containers can still rely on the binaries -- right?

haampie:
Sounds reasonable to me if the tutorial stack is only tested in #3175 where it actually concretizes to specs used in the tutorial ;)

Then #3234 and #3253, which would be blocked by the revert, can be merged after this PR is closed/resolved.

becker33:
As suggested in our slack conversation, I am also reverting the changes to the tutorial pipeline to use `when_possible`

This will result in a stack that is not usable in the tutorial, but will at least check the relevant specs/builds until we can refactor this after https://github.com/spack/spack/pull/51891 goes in.

haampie (approved):

haampie:
> Duration: 2 minutes 42 seconds 

:)

alecbcs:
@spackbot run pipeline

alalazo:
I'd still be for reverting the revert, unless somebody wants to fix `silo` in [this pipeline](https://gitlab.spack.io/spack/spack-packages/-/pipelines/1415779)

.ci/gitlab/configs/linux/ci.yaml
```diff
@@ -4,7 +4,7 @@ ci:
       before_script-:
       # Test package relocation on linux using a modified prefix
       # This is not well supported on MacOS (https://github.com/spack/spack/issues/37162)
-      - - spack config add "config:install_tree:projections:${SPACK_JOB_SPEC_PKG_NAME}/${SPACK_JOB_SPEC_DAG_HASH}:'morepadding/{architecture.platform}-{architecture.target}/{name}-{version}-{hash}'"
+      - - spack config add "config:install_tree:projections:${SPACK_JOB_SPEC_PKG_NAME}:'morepadding/{architecture.platform}-{architecture.target}/{name}-{version}-{hash}'"
```
alalazo, line 7: This introduces bugs when we have multiple instances of the same package in the DAG. Just hit one in a branch with multiple `gcc-runtime`.

alalazo: I'll re-submit this change in a PR.

PR #3259, opened 2026-02-05
cmake: fix a few constraints in dependents

haampie:
Fixes a couple oversights that may or may not prevent unified concretization of
build dependencies in the Windows stack.

### DIFF
--- a/repos/spack_repo/builtin/packages/curl/package.py
+++ b/repos/spack_repo/builtin/packages/curl/package.py
@@ -91,7 +91,7 @@ class Curl(NMakePackage, AutotoolsPackage, CMakePackage):
 
     # CMake 4.0: is not compatible with CMake systems requiring
     # 3.0, which curl@7.63 requires
-    depends_on("cmake@:3", when="build_system=cmake @:7.63")
+    depends_on("cmake@:3", when="build_system=cmake @:7.63", type="build")
 
     depends_on("gnutls@3.6.5:", when="tls=gnutls @8.18:")
     depends_on("gnutls", when="tls=gnutls")
--- a/repos/spack_repo/builtin/packages/curl/package.py
+++ b/repos/spack_repo/builtin/packages/curl/package.py
@@ -91,7 +91,7 @@ class Curl(NMakePackage, AutotoolsPackage, CMakePackage):
 
     # CMake 4.0: is not compatible with CMake systems requiring
     # 3.0, which curl@7.63 requires
-    depends_on("cmake@:3", when="build_system=cmake @:7.63", type="build")
+    depends_on("cmake@:3", type="build", when="build_system=cmake @:7.63")
 
     depends_on("gnutls@3.6.5:", when="tls=gnutls @8.18:")
     depends_on("gnutls", when="tls=gnutls")
--- a/repos/spack_repo/builtin/packages/scitokens_cpp/package.py
+++ b/repos/spack_repo/builtin/packages/scitokens_cpp/package.py
@@ -43,7 +43,7 @@ class ScitokensCpp(CMakePackage):
     depends_on("cxx", type="build")  # generated
 
     depends_on("cmake@2.6:", type="build")
-    depends_on("cmake@3.10:", when="@0.7.1:", type="build")
+    depends_on("cmake@3.10:", type="build", when="@0.7.1:")
     depends_on("openssl")
     depends_on("sqlite")
     depends_on("curl")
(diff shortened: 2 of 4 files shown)


kwryankrattiger (dismissed):
Lgtm

.ci/env
```diff
@@ -1,2 +1,2 @@
-SPACK_CHECKOUT_VERSION=c3567d3b12c57f28590a00160c82ec060f7c7ba4
+SPACK_CHECKOUT_VERSION=686c31bd2e92b7de30360833e094766212372572
```
alalazo: Just putting a placeholder to _not_ merge this while it points to a PR

alalazo (changes requested):

.ci/env
kwryankrattiger: Hm. It would be nice if we used `refs/pull/1818181/head` instead of an opaque hash. That makes it easier to tell where this comes from.

haampie:
`when="%cxx=msvc %fortran=msvc"` seems to be triggering a bug in the concretizer.

Now testing with the fix included from https://github.com/spack/spack/pull/51923.

Edit: seems to work, reverting that...

repos/spack_repo/builtin/packages/curl/package.py
```diff
@@ -91,7 +91,7 @@ class Curl(NMakePackage, AutotoolsPackage, CMakePackage):
 
     # CMake 4.0: is not compatible with CMake systems requiring
     # 3.0, which curl@7.63 requires
-    depends_on("cmake@:3", when="build_system=cmake @:7.63")
+    depends_on("cmake@:3", when="build_system=cmake @:7.63", type="build")
```
alecbcs:
```suggestion
    depends_on("cmake@:3", type="build", when="build_system=cmake @:7.63")
```

Swap added `type` into the middle to follow convention.

repos/spack_repo/builtin/packages/scitokens_cpp/package.py
```diff
@@ -42,8 +42,8 @@ class ScitokensCpp(CMakePackage):
 
     depends_on("cxx", type="build")  # generated
 
-    depends_on("cmake@2.6:")
-    depends_on("cmake@3.10:", when="@0.7.1:")
+    depends_on("cmake@2.6:", type="build")
+    depends_on("cmake@3.10:", when="@0.7.1:", type="build")
```
alecbcs:
```suggestion
    depends_on("cmake@2.6:", type="build")
    depends_on("cmake@3.10:", type="build", when="@0.7.1:")
```

Same suggestion here to follow convention.

haampie: Just push these tiny things next time please :)

alecbcs: I appreciate the suggestion, but I do not think pushing changes like this directly is the right move in general.

I wanted to give feedback on your PR, and if I push directly then I block myself from reviewing it, since I cannot approve a PR when the latest commit is my own.

I also think it is better to notify the author of any requested change, no matter how big or small, through review feedback so they don't accidentally force push and overwrite something they weren't aware of. For smaller items, that's exactly what suggestions are for, since they make it as easy as a single click to accept and commit the change.

alecbcs (dismissed):
Looks good to me. Thanks for the quick turnaround @haampie :)

@alalazo I'll wait on your final review before we merge since you previously requested changes. 
(And thus I'd have to force merge in the GitHub interface.)

haampie:
Rebased this to include the latest Spack version to deal with the concretization issue...

alalazo (dismissed):

haampie:
@johnwparent for your information: the last commit removes your constraint that requires `cmake@4.1:` on windows when `cxx,fortran=msvc`. The reason for this:

* Previously this constraint was ignored by the concretizer due to a bug, so after removal the status quo is maintained.
* If the constraint is respected, it makes concretization of `paraview` impossible because it requires both external `cmake@3` and `cmake@4.1:` with `paraview %cxx,fortran=msvc`

Maybe you could bump external CMake to a newer version in the relevant image and add back the constraint?

IMO the change shouldn't be controversial given that the conflict was ignored from the start and paraview built fine.

johnwparent:
There's also always the potential that updating the image might cause other CI failures, so if this is an immediate need, it makes sense to just merge without trying to update the CI image.

---

I don't _love_ actively introducing a broken relationship between CMake and `%fortran=msvc` but as Harmen points out, it's a no-op functionally so 🤷‍♂️ .

We also must not actually be using the fortran compiler with CMake in CI, since otherwise I have no idea how that's working... maybe there's a lower bound where oneAPI + MSVC + CMake work again...

The correct solution, as Harmen points out, is to bump the CMake version in the CI image to post 4.1 (and for me to investigate whether there is a lower bound for this conflict) but that may take some time if the image updates introduces other failures. 


Thanks for the heads up @haampie

haampie:
Just commented out the constraint so that we don't forget about it

alalazo (approved):

PR #3260, opened 2026-02-05
gsibec: add v1.4.0

mathomp4:
This adds [`gsibec`](https://github.com/GEOS-ESM/GSIbec/) 1.4.0 which also now depends on `ip`

ETA: I also added v1.4.1 which is now a v-tag so I had to fiddle with the package.


alecbcs (approved):
Looks good to me. Thanks @mathomp4!

Merging after >1 without hearing back from @climbfuji.

climbfuji (commented):
Retrospective approval!

PR #3261, opened 2026-02-05
vasp: fix compilation with intel-oneapi-mkl as fftw provider

klartz:
Compilation failed with `Cannot open include file 'fftw3.f'` because the MKL FFTW headers are not in the default include path. Adds `$MKLROOT/include/fftw` to the list of include directories to resolve the issue.


robertexternbrink:
fixed by #3333

tldahlgren:
@snehring 

> fixed by #3333

@klartz Can you confirm that PR resolves the issue? If so, then this PR can be closed.

snehring:
It does look like #3333 fixes this.

klartz:
I recompiled with #3333 and it does indeed fix the issue, so this PR can be closed. Sorry for the late reply.

tldahlgren:
Closing since #3333 resolves the issue.

PR #3262, opened 2026-02-05
vasp: fix make_include filename for intel-oneapi compilers

klartz:
When building with intel-oneapi compilers, the GNU fallback logic generated an incorrect make_include filename due to an extra dot.


tldahlgren:
@snehring

snehring (approved):
lgtm

tldahlgren (approved):

PR #3266, opened 2026-02-05
hdf5-vol-log: add variant for hdf5 examples

qtpowell:
Adds a patch and a variant to be able to disable the hdf5 examples. These examples are normally downloaded during build, which causes errors when building on compute nodes at facilities such as Frontier.

### DIFF
--- a/repos/spack_repo/builtin/packages/hdf5_vol_log/package.py
+++ b/repos/spack_repo/builtin/packages/hdf5_vol_log/package.py
@@ -22,6 +22,8 @@ class Hdf5VolLog(AutotoolsPackage):
 
     version("1.4.0", tag="logvol.1.4.0", commit="786d2cc4da8b4a0827ee00b1b0ab3968ef942f99")
 
+    variant("hdf5_examples", default=True, description="Enable HDF5 examples", when="@1.4.0")
+
     depends_on("c", type="build")  # generated
     depends_on("cxx", type="build")  # generated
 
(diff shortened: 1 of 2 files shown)


hyoklee (approved):

hyoklee (approved):

qtpowell:
@kwryankrattiger  can you review?

repos/spack_repo/builtin/packages/hdf5_vol_log/package.py
```diff
@@ -22,6 +22,8 @@ class Hdf5VolLog(AutotoolsPackage):
 
     version("1.4.0", tag="logvol.1.4.0", commit="786d2cc4da8b4a0827ee00b1b0ab3968ef942f99")
 
+    variant("hdf5_examples", default=True, description="Enable HDF5 examples", when="@1.4.0")
```
hyoklee, line 25: @qtpowell , shouldn't it be False by default?

qtpowell: I don't think so, since the current behavior without the variant includes these examples.

hyoklee (approved):

repos/spack_repo/builtin/packages/hdf5_vol_log/package.py
kwryankrattiger: The convention for an examples variant is just `examples`

```suggestion
    variant("examples", default=True, description="Enable HDF5 examples", when="@1.4.0")
```

qtpowell: This variant only controls a subset of the examples under the subdirectory `examples/hdf5_examples`, the other examples are untouched.

kwryankrattiger: I see your point. I suppose I am fine with the name for now then and if we end up revisiting this we can discuss again.

kwryankrattiger (approved):

PR #3267, opened 2026-02-05
yosys: add v0.62

davekeeshan:
yosys: add v0.62

https://github.com/YosysHQ/yosys/releases/tag/v0.62


RikkiButler20 (approved):
Verified sha256. LGTM

PR #3273, opened 2026-02-05
CoinHSL add C and Fortran language dependencies.

nkoukpaizan:
This add C and Fortran language dependencies to the CoinHSL package. 

Without this, the Spack build environment compilers may not be able to run simple executables when meson checks for the compilers.


AndrewLister-STFC (approved):
Thanks!

alalazo (approved):

PR #3292, opened 2026-02-07
mfem: fix early return with openmp

Sbozzolo:
Applies https://github.com/mfem/mfem/pull/5224

The MFEM package.py includes a patch, `mfem-4.9.patch` that contains OpenMP code that is not strictly conforming to specification, possibly resulting in compilation errors.

### DIFF
--- a/repos/spack_repo/builtin/packages/mfem/mfem-4.9.patch
+++ b/repos/spack_repo/builtin/packages/mfem/mfem-4.9.patch
@@ -1,5 +1,5 @@
 diff --git a/fem/fe/fe_base.cpp b/fem/fe/fe_base.cpp
-index 535fdaca8b..281de35fac 100644
+index 535fdaca8b..b951c9bdd1 100644
 --- a/fem/fe/fe_base.cpp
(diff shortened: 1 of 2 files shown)


v-dobrev:
Thanks @Sbozzolo! What compiler do you see this issue with? I thought, I tested the patch (with `+openmp`) but either I did not or I just tested it with the ROCm clang and that worked fine.

Sbozzolo:
It is failing with gcc 13.3.0 in our [CI](https://github.com/awslabs/palace/actions/runs/21770113362/job/62815564179).

v-dobrev (dismissed):
Looks good, thanks @Sbozzolo!

It turns out Clang does not have an issue with the original patch, just GCC. In any case, this fix is good for both.

alalazo (changes requested):
We should add a comment above the patch directive referring to this patch and describe what it does, where it was taken from etc.

v-dobrev (approved):

alalazo (approved):

Issue #2514, opened 2025-11-20
Installation issue: eigen


msparapa:
### Steps to reproduce the issue

```console
$ spack spec -I eigen@5.0.0
 -   eigen@5.0.0~ipo~nightly~rocm build_system=cmake build_type=Release generator=make platform=linux os=rhel8 target=skylake_avx512 %c,cxx,fortran=gcc@8.5.0
[+]      ^cmake@3.31.9~doc+ncurses+ownlibs~qtgui build_system=generic build_type=Release platform=linux os=rhel8 target=skylake_avx512 %c,cxx=gcc@8.5.0
[+]          ^curl@8.15.0~gssapi~ldap~libidn2~librtmp~libssh~libssh2+nghttp2 build_system=autotools libs:=shared,static tls:=openssl platform=linux os=rhel8 target=skylake_avx512 %c,cxx=gcc@8.5.0
[+]              ^nghttp2@1.67.1 build_system=autotools platform=linux os=rhel8 target=skylake_avx512 %c,cxx=gcc@8.5.0
[+]                  ^diffutils@3.12 build_system=autotools platform=linux os=rhel8 target=skylake_avx512 %c=gcc@8.5.0
[+]                      ^libiconv@1.18 build_system=autotools libs:=shared,static platform=linux os=rhel8 target=skylake_avx512 %c=gcc@8.5.0
[+]              ^openssl@3.6.0~docs+shared build_system=generic certs=mozilla platform=linux os=rhel8 target=skylake_avx512 %c,cxx=gcc@8.5.0
[+]                  ^ca-certificates-mozilla@2025-08-12 build_system=generic platform=linux os=rhel8 target=skylake_avx512
[+]              ^perl@5.42.0+cpanm+opcode+open+shared+threads build_system=generic platform=linux os=rhel8 target=skylake_avx512 %c=gcc@8.5.0
[+]                  ^berkeley-db@18.1.40+cxx~docs+stl build_system=autotools patches:=26090f4,b231fcc platform=linux os=rhel8 target=skylake_avx512 %c,cxx=gcc@8.5.0
[+]                  ^bzip2@1.0.8~debug~pic+shared build_system=generic platform=linux os=rhel8 target=skylake_avx512 %c=gcc@8.5.0
[+]                  ^gdbm@1.25 build_system=autotools platform=linux os=rhel8 target=skylake_avx512 %c=gcc@8.5.0
[...]

PR #3293, opened 2026-02-07
eigen: Make dependence on fortran optional

jcortial-safran:
Eigen's dependence on Fortran is optional. In particular, Fortran is required to build eigen-based implementations for LAPACK and, when @:3, for BLAS as well.

When @:3.4.0, eigen detects if Fortran is present, and it automatically enables ot disables BLAS and LAPACK based on that.

When @3.41:, eigen exposes 2 CMake variables, EIGEN_BUILD_BLAS and EIGEN_BUILD_LAPACK, to let the user control whether to build BLAS and LAPACK. These variables are ON by default when eigen is the top level project.

So I added 2 variants, blas and lapack, that control the above CMake variables, and that are only available when @:3.4.1. I made thme *False* by default to make them mimick the behavior of eigen @3.4.0 under recent spack versions. My understanding is that when Fortran is not explicitly required, CMake does find it (or should not find it...) and thus BLAS and LAPACK are disabled. 

I also tried to make +lapack dependent on +blas.

Hopefully this PR offers a workaround for #2514.

### DIFF
--- a/repos/spack_repo/builtin/packages/eigen/package.py
+++ b/repos/spack_repo/builtin/packages/eigen/package.py
@@ -48,7 +48,9 @@ class Eigen(CMakePackage, ROCmPackage):
     version("3.2.5", sha256="8068bd528a2ff3885eb55225c27237cf5cda834355599f05c2c85345db8338b4")
 
     variant("blas", description="Build BLAS implementation", when="@3.4.1:", default=False)
-    variant("lapack", description="Build LAPACK implementation", when="@3.4.1: +blas", default=False)
+    variant(
+        "lapack", description="Build LAPACK implementation", when="@3.4.1: +blas", default=False
+    )
     variant("nightly", description="run Nightly test", default=False)
 
     # TODO: https://eigen.tuxfamily.org/dox/TopicUsingBlasLapack.html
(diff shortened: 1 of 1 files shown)


jcortial-safran:
@spackbot fix style

HaoZeke (approved):
Thanks @jcortial-safran. LGTM

repos/spack_repo/builtin/packages/eigen/package.py
```diff
@@ -88,7 +94,8 @@ class Eigen(CMakePackage, ROCmPackage):
 
     depends_on("c", type="build")
     depends_on("cxx", type="build")
-    depends_on("fortran", type="build", when="@3.4.1:")
+    depends_on("fortran", type="build", when="@3.4.1: +lapack")
+    depends_on("fortran", type="build", when="@3.4.1:3 +blas")
```
alalazo, line 98: This means that `eigen@5+blas~lapack` does not depend on `fortran`. Is it correct?

jcortial-safran: Correct.

alalazo (approved):

PR #3294, opened 2026-02-08
mpich: add v5.0.0

rbberger:


alalazo (changes requested):
There are failures in pipelines that need to be investigated. Most likely `pruners-ninja` and maybe other packages can't yet use `mpich@5`

raffenet:
> There are failures in pipelines that need to be investigated. Most likely `pruners-ninja` and maybe other packages can't yet use `mpich@5`

```
  >> 169    ninj.cpp:1302:92: error: 'MPIO_Request' has not been declared
```

It's related to the removal of the definition of `MPIO_Request` from the standard header files. To my understanding, that is an old, non-standard way of using MPI I/O interfaces. The downstream code should be updated to use `MPI_Request` instead.

raffenet:
> > There are failures in pipelines that need to be investigated. Most likely `pruners-ninja` and maybe other packages can't yet use `mpich@5`
> 
> ```
>   >> 169    ninj.cpp:1302:92: error: 'MPIO_Request' has not been declared
> ```
> 
> It's related to the removal of the definition of `MPIO_Request` from the standard header files. To my understanding, that is an old, non-standard way of using MPI I/O interfaces. The downstream code should be updated to use `MPI_Request` instead.

https://github.com/PRUNERS/NINJA/pull/5

rbberger:
https://github.com/spack/spack-packages/pull/3398

rbberger:
@spackbot re-run pipeline.

rbberger:
@alalazo The CI for this one finally passed.

alalazo (approved):

PR #3301, opened 2026-02-08
guile: add 3.0.11 (and 3.0.10)

jagot:
https://www.gnu.org/software/guile/download/

### DIFF
--- a/repos/spack_repo/builtin/packages/guile/package.py
+++ b/repos/spack_repo/builtin/packages/guile/package.py
@@ -87,7 +87,7 @@ class Guile(AutotoolsPackage, GNUMirrorPackage):
         else:
             config_args.append("--without-libreadline-prefix")
 
-        if not spec.satisfies("+readline"):
+        if not spec.satisfies("+gmp"):
             config_args.append("--enable-mini-gmp")
 
         if spec.satisfies("+shared"):
--- a/repos/spack_repo/builtin/packages/guile/package.py
+++ b/repos/spack_repo/builtin/packages/guile/package.py
@@ -50,6 +50,7 @@ class Guile(AutotoolsPackage, GNUMirrorPackage):
     depends_on("libtool@1.5.6:", type="link")  # links to libltdl.so
     depends_on("libunistring@0.9.3:")
     depends_on("libffi")
+    depends_on("libxcrypt", type="link")
     depends_on("readline", when="+readline")
     depends_on("pkgconfig", type="build")
     depends_on("gperf", when="@3:")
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/guile/package.py
```diff
@@ -82,4 +87,12 @@ def configure_args(self):
         else:
             config_args.append("--without-libreadline-prefix")
 
+        if not spec.satisfies("+readline"):
```
haampie: did you mean to check the `gmp` variant here?

jagot: Right, thanks for catching that!

haampie (dismissed):

jagot:
I should probably mention that this builds and works as expected on Linux. On macOS, it builds and the interpreter seems to run fine, but if I embed Guile in a multithreaded C++ application, it crashes with an error message: `Too many root sets`, which is a message from the garbage collector `bdw-gc`.  I have no such problems if I link against the same version of Guile installed using Homebrew.

haampie:
OK, maybe worth checking if they have any different config options or patches.

jagot:
Yes of course, but I was not able to figure out the difference yet, so would appreciate if someone else could have a go as well. In any case, I need this on Linux, where it does work.

haampie:
Can you still fix

```
    lib/libguile-3.0.so.1.7.1
            libcrypt.so.1 => not found
```

add `depends_on("libxcrypt", type="link")`

Applies to guile 2 as well.

repos/spack_repo/builtin/packages/guile/package.py
```diff
@@ -28,26 +30,29 @@ class Guile(AutotoolsPackage, GNUMirrorPackage):
...
 
     depends_on("c", type="build")  # generated
 
     depends_on("bdw-gc@7.0: threads=none", when="threads=none")
     depends_on("bdw-gc@7.0: threads=posix", when="threads=posix")
     depends_on("bdw-gc@7.0: threads=dgux386", when="threads=dgux386")
-    depends_on("gmp@4.2:")
+    depends_on("gmp@4.2:", when="+gmp")
     depends_on("gettext")
     depends_on("libtool@1.5.6:", type="link")  # links to libltdl.so
```
wdconinc, line 50: @jagot Here is a code suggestion based on @haampie's comment. If you apply it, we can merge this PR.
```suggestion
    depends_on("libtool@1.5.6:", type="link")  # links to libltdl.so
    depends_on("libxcrypt", type="link")
```

jagot: Thanks, I had already committed this to my fork, but forgot to push the change.

wdconinc (approved):
Thanks. Looks fine to me.

PR #3305, opened 2026-02-08
rocksdb: add 10.10.1 and reduce package size

michaelkuhn:
The rocksdb package is huge by default. Two changes reduce its size:
1. Do not build the static library by default. This decreases the size from 1.4 GB to 300 MB.
2. Introduce a strip variant that is on by default and strips the libraries. This decreases the sizes of the static+shared and shared builds to 50 MB and 20 MB, respectively.

```
$ du -sh opt/spack/linux-skylake/rocksdb-10.10.1-*
17M	opt/spack/linux-skylake/rocksdb-10.10.1-krej4blwryb5eixb6y6obza6xav7pqyu
48M	opt/spack/linux-skylake/rocksdb-10.10.1-ncospq5pojyhbwvo2jjqbmcth4r4bnlj
1.4G	opt/spack/linux-skylake/rocksdb-10.10.1-ru567lgykpxksilmc62q7f35d7kvwapl
322M	opt/spack/linux-skylake/rocksdb-10.10.1-vuy2e3hkrmbq2apwlzk7pv3a2fawwrq2
```

### DIFF
--- a/repos/spack_repo/builtin/packages/rocksdb/package.py
+++ b/repos/spack_repo/builtin/packages/rocksdb/package.py
@@ -17,6 +17,7 @@ class Rocksdb(MakefilePackage):
     license("Apache-2.0 OR GPL-2.0-only")
 
     version("master", git=git, branch="master", submodules=True)
+    version("10.10.1", sha256="df2ff348f3fac8578fd4b727eee7267aaf90cd403c99b55e898d1db63fa8cff5")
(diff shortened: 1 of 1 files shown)


haampie:
In the past strip + patchelf has produced broken binaries. Is it possible to disable debug symbols instead?

michaelkuhn:
I've pushed an updated version that disables the debug flag in the Makefile. Size reduction is the same.

michaelkuhn:
@haampie Ping? :slightly_smiling_face:

haampie (approved):

PR #3310, opened 2026-02-09
curl: 8.18

haampie:
See https://daniel.haxx.se/blog/2026/01/07/curl-8-18-0/

Not deprecating older versions because they're still needed by some.


haampie:
@spackbot run pipeline

haampie:
@spackbot run pipeline

alecbcs (approved):
Looks good to me. I agree let's merge this and then do a follow up to see how we can push dependents to get them off of insecure versions.

PR #3323, opened 2026-02-09
julia: add v1.11.9

m-fila:


haampie:
LGTM.

As for your other PR about 1.12, seems like they backported the fix for the 7z issue we see in CI as part of https://github.com/JuliaLang/julia/pull/60612. So, I think we can revisit that with the 1.12.5 release, or you can add a patch over there.

haampie (approved):

m-fila:
Thanks. In the 1.12 PR I was indeed a bit confused by the relationship between `p7ip` and `7zip`, and didn't have much time to check out what is actually going on

PR #3332, opened 2026-02-09
silo: replace broken patch

haampie:
Use the minimal diff instead of all the accumulated commits from https://github.com/llnl/Silo/pull/515, which does not apply. Expand to full commit in URL.


alalazo (approved):
`tutorial` stack passed

PR #3336, opened 2026-02-09
freefem: add develop version, superlu variant, and update lapack dependency

mohd-afeef-badri:
This PR updates the `FreeFem` package by:

- adding a `develop` version from the upstream Git repository to make recent upstream developments available (the latest released version in Spack, 4.14, is about two years old)
- introducing a `+superlu` variant (enabled by default) to allow users to explicitly disable SuperLU support
- switching the dependency from `netlib-lapack` to the generic `lapack` virtual dependency, since `FreeFem` is typically used together with `PETSc`, which depends on the `lapack` virtual and not specifically on `netlib-lapack` (https://packages.spack.io/package.html?name=petsc), avoiding unnecessary concretization conflicts
- adding the LGPL-3.0-only license metadata


corentin-dev:
I'm sorry, I won't be able to review this PR, and would like to step down as I'm not using FreeFEM (and won't probably in the foreseeable future).

Do you wish to take over?

mohd-afeef-badri:
@corentin-dev sure I can take over, I plan to use FreeFEM for another package of mine as a dependency. I will gladly take over. I will see if others can join, I know some other people from the community.

Although, I guess you will still need to review this one.

alalazo (approved):
Thanks. Feel free to submit another Pr and list yourself as a maintainer.

PR #3342, opened 2026-02-09
py-numcodecs: add 0.16.5

Chrismarsh:
- adds version 0.16.5 (deps didn't seem to change)
- requires `@0.16.3:` when `%gcc@15:` due to https://github.com/zarr-developers/numcodecs/issues/750, https://github.com/zarr-developers/numcodecs/releases/tag/v0.16.3

### DIFF
--- a/repos/spack_repo/builtin/packages/py_numcodecs/package.py
+++ b/repos/spack_repo/builtin/packages/py_numcodecs/package.py
@@ -23,6 +23,7 @@ class PyNumcodecs(PythonPackage):
     license("MIT")
 
     version("main", branch="main", submodules=True)
+    version("0.16.5", sha256="0d0fb60852f84c0bd9543cc4d2ab9eefd37fc8efcc410acd4777e62a1d300318")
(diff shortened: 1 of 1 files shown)


RikkiButler20 (approved):
LGTM

repos/spack_repo/builtin/packages/py_numcodecs/package.py
```diff
@@ -68,6 +69,9 @@ class PyNumcodecs(PythonPackage):
 
     patch("apple-clang-12.patch", when="@:0.13 %apple-clang@12:")
 
+    # https://github.com/zarr-developers/numcodecs/releases/tag/v0.16.3
+    requires("@0.16.3:", when="%gcc@15:")
```
alecbcs, line 73:
```suggestion
    # https://github.com/zarr-developers/numcodecs/releases/tag/v0.16.3
    depends_on("gcc@15:", when="@0.16.3: %c=gcc")
```

Note for the future since this already merged. Instead of using a `requires()` directive here I'd recommend using `depends_on()` with a self referential when condition. I find the latter to be a bit easier to read and understand and better models how we actually think about the compiler dependency.

PR #3343, opened 2026-02-09
crlf -> lf

haampie:
For whatever reason, these files were contributed with `\r\n`.


alecbcs:
We should add a style check inside of CI to prevent this from happening in the future. Additionally spackbot could be made to fix this in its automatic style fix.

haampie:
I believe it can be configured in `.gitattributes`.

Edit: done that. Sticking to `*.py` and not `*.patch` because those patch files already exist and would change checksum, causing lookup errors for existing concrete specs.

alecbcs (approved):
Looks good to me. Thanks for adding the gitattributes!

PR #3345, opened 2026-02-10
eigen: deprecate older versions

jcortial-safran:
I propose to deprecate older versions of eigen, namely @:3.3, except the most recent releases of each series 3.3.x and 3.2.x.

This will allow to discard a few patches and make the recipe more maintainable as soon as the deprecated versions are removed.


HaoZeke (commented):
~Maybe more administrative, but do know if there are packages which pin to a specific `3.3.x`? If not I'm happy with this.~

Actually works for me.

HaoZeke (approved):

rbberger (approved):

PR #3351, opened 2026-02-10
GDAL: add v3.12.2

adamjstewart:
https://github.com/OSGeo/gdal/blob/v3.12.2/NEWS.md


RikkiButler20 (approved):
Verified sha256. LGTM

PR #3352, opened 2026-02-10
Keras: wider backend support window

adamjstewart:
https://github.com/keras-team/keras#backend-compatibility-table was recently added and finally documents which versions of these backends are actually supported.


alecbcs (approved):
Looks good to me. Thanks!

PR #3355, opened 2026-02-10
ROCm: remove CRLF line endings

afzpatel:
I've verified with `file` that these patches no long display `with CRLF line terminators`.


haampie (approved):
Thanks! It does influence their content hash. Could be that Spack still has some cache invalidation issue when you update the contents of a patch without renaming the file. In that case users should do `spack clean -m`.

PR #3357, opened 2026-02-10
task: add v3.4.2, modify URL to point at GitHub

taliaferro:
This introduces three changes to the `task` package:

1. Add a new checksum for the latest version, Taskwarrior 3.4.0
2. Add `depends_on("c")` -- Taskwarrior is written in C++ and Rust, but without this set, the `cmake` phase fails with an error complaining that the `SPACK_CC` environment variable is unset.
3. Update the download URL 

Prior versions of Taskwarrior were released as source tarballs on `taskwarrior.org`, which was the original download URL in the package. Starting with release 2.5.2 the developers began publishing it in parallel through GitHub releases, but continued to redirect appropriately from taskwarrior.org. Starting with `task@3.3.0`, however, they seem to have stopped adding redirects for new versions -- GitHub now appears to be the only place to get new releases. With this in mind, I've updated the `url` property on the package to point to the GitHub release (thankfully, the checksums for old versions of the packages are the same for both). For the two versions which predate the earliest GitHub release, I've set the `url=` property separately on those versions.

### DIFF
--- a/repos/spack_repo/builtin/packages/task/package.py
+++ b/repos/spack_repo/builtin/packages/task/package.py
@@ -11,15 +11,17 @@ class Task(CMakePackage):
     """Feature-rich console based todo list manager"""
 
     homepage = "https://www.taskwarrior.org"
-    url = "https://taskwarrior.org/download/task-2.4.4.tar.gz"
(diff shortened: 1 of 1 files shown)


taliaferro:
@spackbot fix style please

alecbcs (approved):
Looks good to me. Thanks @taliaferro!

PR #3358, opened 2026-02-10
timew: add v1.9.1 and c language dependency

taliaferro:
This adds a checksum for `timew@1.9.1`, released last August. In the process of getting it to build, I also noticed that (much like Taskwarrior) it would no longer build with new versions of Spack because it was missing `depends("c")`. (the `cmake` phase would fail, complaining that the `SPACK_CC` environment variable was unset.)


alecbcs (approved):
Looks good to me. Thanks @taliaferro 🎉

PR #3021, opened 2026-01-13
trilinos: add 20 as a cxxstd value 

kuberry:
Beginning with `trilinos@17`, `cxxstd=20` will be required. 

This sets `cxxstd=20` as the default and adds a conflict for `cxxstd=17` from 17 onward.

EDIT: This sets `cxxstd=17` (consistent with `trilinos@16.x`, but adds the option of 20 and marks it as a conflict with anything before `trilinos@17`

PR #3361, opened 2026-02-11
trilinos: add v17.0.0

kuberry:
The cxxstd conflicts have already been registered in #3021, so that 17 will be used for :16 and 20 will be used for 17:.


cgcgcg (approved):

repos/spack_repo/builtin/packages/trilinos/package.py
```diff
@@ -340,6 +341,7 @@ class Trilinos(CMakePackage, CudaPackage, ROCmPackage):
     conflicts("+intrepid", when="~sacado")
     conflicts("+minitensor", when="~boost")
     conflicts("+phalanx", when="~sacado")
+    conflicts("+stk", when="~gtest")
```
kuberry: `+stk` (even with the default ~test) enables `STKDoc_tests`, which requires gtest, resulting in the following error:
```sh
-- ERROR: Did not find a lib in the lib set "gtest googletest" for the TPL 'gtest'!
-- ERROR: Could not find the libraries for the TPL 'gtest'!
-- TIP: If the TPL 'gtest' is on your system then you can set:
     -Dgtest_LIBRARY_DIRS='<dir0>;<dir1>;...'
   to point to the directories where these libraries may be found.
   Or, just set:
     -DTPL_gtest_LIBRARIES='<path-to-libs0>;<path-to-libs1>;...'
   to point to the full paths for the libraries which will
   bypass any search for libraries and these libraries will be used without
   question in the build.  (But this will result in a build-time error
   if not all of the necessary symbols are found.)
-- ERROR: Failed finding all of the parts of TPL 'gtest' (see above), Aborting!

-- NOTE: The find module file for this failed TPL 'gtest' is:
     /var/folders/zk/1zr9bdg53s3_d7qxmv2m5f80002l2h/T/pakuber/spack-stage/spack-stage-trilinos-17.0.0-yy64b34lvenax3qepvvtz4gike33sof6/spack-src/cmake/TPLs/FindTPLgtest.cmake
   which is pointed to in the file:
     /var/folders/zk/1zr9bdg53s3_d7qxmv2m5f80002l2h/T/pakuber/spack-stage/spack-stage-trilinos-17.0.0-yy64b34lvenax3qepvvtz4gike33sof6/spack-src/TPLsList.cmake

TIP: One way to get past the configure failure for the
TPL 'gtest' is to simply disable it with:
  -DTPL_ENABLE_gtest=OFF
which will disable it and will recursively disable all of the
downstream packages that have required dependencies on it, including
the package 'STKDoc_tests' which triggered its enable.
When you reconfigure, just grep the cmake stdout for 'gtest'
and then follow the disables that occur as a result to see what impact
this TPL disable has on the configuration of Trilinos.

CMake Error at cmake/tribits/core/package_arch/TribitsProcessEnabledTpls.cmake:248 (message):
  ERROR: TPL_gtest_NOT_FOUND=TRUE, aborting!
```
This line addresses the stk dependency on gtest. If anyone on the STK team is interested in having a look, I'm curious why building without tests would invoke a test dependency.

repos/spack_repo/builtin/packages/trilinos/package.py
```diff
@@ -416,6 +418,8 @@ class Trilinos(CMakePackage, CudaPackage, ROCmPackage):
         msg="Only serial builds are supported on Windows currently.",
     )
 
+    conflicts("+test", when="@17: ~gtest")
```
kuberry: From 17 onward, a vendored version of gtest is no longer provided with Trilinos. A large number of packages having testing requirements that need gtest.

repos/spack_repo/builtin/packages/trilinos/package.py
```diff
@@ -992,6 +999,10 @@ def define_tpl(trilinos_name, spack_name, have_dep):
                 ]
             )
 
+        if spec.satisfies("@17: +gtest"):
```
kuberry, line 1037: `define_tpl` for gtest couldn't be used. `define_tpl` does allow to specify that the spec is called 'googletest', however the library name is not lgoogletest.so or lgoogletest.dylib, which means that the googletest package requires a custom `def libs` function for googletest's package.py if we would want to use `define_tpl`.

repos/spack_repo/builtin/packages/trilinos/package.py
cgcgcg: We could be a bit more surgical here and only enable it if one of the enabled packages needs it. Then again, it's a really lightweight dependency.

repos/spack_repo/builtin/packages/trilinos/package.py
```diff
@@ -995,6 +999,10 @@ def define_tpl(trilinos_name, spack_name, have_dep):
...
+            options.append(define_tpl_enable("GTEST", True))
```
cgcgcg: Should that be "gtest" instead of "GTEST"?

repos/spack_repo/builtin/packages/trilinos/package.py
cgcgcg: @alanw0

alanw0: I'll take a look. I agree that `~test` should turn off all the tests...

repos/spack_repo/builtin/packages/trilinos/package.py
kuberry: It is really lightweight. It would require adding quite a few conflicts as well:
https://github.com/trilinos/Trilinos/pull/14719/changes
At least compadre, minitensor, stk, rol, percept, krino... I thought it might be safer to just enable it whenever `+test` so that we don't miss one.

repos/spack_repo/builtin/packages/trilinos/package.py
kuberry: I changed it to 'gtest' just now. It works for me either way, but I was unclear about the capitalization.

repos/spack_repo/builtin/packages/trilinos/package.py
kuberry: In the last round of CI failures, it looks like the packages I mentioned above (minitensor, percept, etc..). fail if they do not also also have a registered dependence on gtest in Spack. 

This is strange, because these are registered as dependencies in tribits with lines like:
TEST_REQUIRED_TPLS gtest
which to my understanding should not be invoked when tests are off (which I can confirm `'-DTrilinos_ENABLE_TESTS:BOOL=OFF'` is being passed to CMake). 

@bartlettroscoe, am I correct in my understanding of this (namely that test optional and required deps shouldn't be required when Trilinos_ENABLE_TESTS:BOOL=OFF)?

@alanw0 I believe this is the same issue affecting STKDoc_tests

bartlettroscoe: > @bartlettroscoe, am I correct in my understanding of this (namely that test optional and required deps shouldn't be required when Trilinos_ENABLE_TESTS:BOOL=OFF)?

@kuberry, yes, but there is an old defect in TriBITS that will enabled optional test dependencies even if tests are disabled.  See:

* https://github.com/TriBITSPub/TriBITS/issues/56
* https://github.com/trilinos/Trilinos/issues/14817

The workaround is to change to an optional TPL dependency and put in the logic to handle that.

kuberry: In the most recent commit, I required gtest for all of the packages listing gtest as a test dependency in https://github.com/trilinos/Trilinos/pull/14719/changes. This alleviated the need to add gtest as a dependency for +tests (because the packages that need it already have registered conflicts).

kuberry:
It looks the current issue in 17 holding this PR back is HIP related:
```sh
     3726    In file included from /tmp/root/spack-stage/spack-stage-trilinos-1
             7.0.0-yzugdaa34n3jtxn7crbhiljtiwxtxcnt/spack-src/packages/stk/stk_
             ngp_test/stk_ngp_test/GlobalReporter.cpp:2:
  >> 3727    /tmp/root/spack-stage/spack-stage-trilinos-17.0.0-yzugdaa34n3jtxn7
             crbhiljtiwxtxcnt/spack-src/packages/stk/stk_ngp_test/stk_ngp_test/
             GlobalReporter.hpp:11:2: error: "Kokkos_ENABLE_HIP_RELOCATABLE_DEV
             ICE_CODE is required for HIP build"
     3728       11 | #error "Kokkos_ENABLE_HIP_RELOCATABLE_DEVICE_CODE is requi
             red for HIP build"
```

cgcgcg:
@alanw0 Is that an issue with the CMake logic? Should stk_ngp_test be disabled if RDC isn't enabled?

alanw0:
> @alanw0 Is that an issue with the CMake logic? Should stk_ngp_test be disabled if RDC isn't enabled?

We need to push a stk update to trilinos. We recently removed (in our 'home' repo) the RDC requirement for current ROCM versions. It was needed for older versions.
stk_ngp_test should be enabled iff tests are enabled. If it is being enabled when tests are not enabled I'll need to look into that.

cgcgcg:
@alanw0 Thanks!

kuberry:
When we enable STK with `+stk`, it will enable all PT packages in the STK folder (including stk_ngp_test), so I wasn't surprised to see stk_ngp_test being enabled.
```sh
2026-02-15T19:18:51.987297Z 01O -- Setting subpackage enable Trilinos_ENABLE_STKNGP_TEST=ON because parent package Trilinos_ENABLE_STK=ON
```

In one of the failing builds, we have:
```sh
-DKokkos_ENABLE_ROCM:BOOL=OFF
-DKokkos_ENABLE_HIP:BOOL=ON
-DKokkos_ENABLE_HIP_RELOCATABLE_DEVICE_CODE:BOOL=OFF
```

In the package logic, we have:
depends_on("kokkos+hip_relocatable_device_code~shared", when="+rocm_rdc")
So evidently `+rocm_rdc` isn't being enabled or isn't doing anything.

For `+rocm_rdc` to do anything, it would have to be through:
```
            if "+rocm" in spec:
                options.extend(
                    [
                        define_kok_enable("ROCM", False),
                        define_kok_enable("HIP", True),
                        define_kok_enable("HIP_RELOCATABLE_DEVICE_CODE", "rocm_rdc"),
                    ]
                )
```
I.e. we would need the spec to be `+rocm` and `+rocm_rdc`. It looks like it is `+rocm` and `~rocm_rdc`.

From:
https://github.com/spack/spack-packages/blob/develop/stacks/e4s-rocm-external/spack.yaml
it looks like the spec is:
```sh
  - trilinos +amesos +amesos2 +anasazi +aztec +belos +boost +epetra +epetraext +ifpack ~ifpack2 +intrepid +intrepid2 +isorropia +kokkos +ml +minitensor +muelu +nox +piro +phalanx +rol +rythmos +sacado +stk +shards +shylu ~stokhos +stratimikos +teko +tempus +tpetra +trilinoscouplings +zoltan +zoltan2 +superlu-dist gotype=long_long +rocm amdgpu_target=gfx90a
```
without specifying `+rocm_rdc` (which is false by default).

I'm wondering why we are just now running into this (recent addition to STK?).

alanw0:
I've added the patch to require RDC only for rocm versions earlier than 6.3. #14964

repos/spack_repo/builtin/packages/trilinos/package.py
```diff
@@ -572,6 +586,13 @@ class Trilinos(CMakePackage, CudaPackage, ROCmPackage):
     # https://github.com/spack/spack-packages/pull/2931#issuecomment-3756434768
     patch("16-2-stk_destructor_kokkos_decoration.patch", when="@=16.2.0 +stk")
 
+    # https://github.com/spack/spack-packages/pull/3361
+    patch(
+        "https://github.com/trilinos/Trilinos/commit/b36622c86f3adfef3ec4cf6e40f9645099e11a18.patch?full_index=1",
+        sha256="7a07f769aedf6433e440e0ce666c66d2721bfa8b99b18d8e3e95e1303878778d",
+        when="@17.0.0 +stk",
```
kuberry: I just pushed a change from @17.0.0 to @=17.0.0. The pipeline failed previously with https://gitlab.spack.io/spack/spack-packages/-/jobs/20616032:
```sh
     3669    In file included from /tmp/root/spack-stage/spack-stage-trilinos-1
             7.0.0-ypz3wsi6l2ai7cm3w2x2dcbsq7xp2uun/spack-src/packages/stk/stk_
             ngp_test/stk_ngp_test/GlobalReporter.cpp:2:
  >> 3670    /tmp/root/spack-stage/spack-stage-trilinos-17.0.0-ypz3wsi6l2ai7cm3
             w2x2dcbsq7xp2uun/spack-src/packages/stk/stk_ngp_test/stk_ngp_test/
             GlobalReporter.hpp:12:2: error: "Kokkos_ENABLE_HIP_RELOCATABLE_DEV
             ICE_CODE is required for HIP build"
     3671       12 | #error "Kokkos_ENABLE_HIP_RELOCATABLE_DEVICE_CODE is requi
             red for HIP build"
```
The line number (12) reflects the patch being applied from: https://github.com/trilinos/Trilinos/pull/14964/changes

It looks like ROCM 7.1.1 is being used, and this section of code should not be entered in STK unless ROCM<6.3...

kuberry: Conditional include of rocm_version.h in https://github.com/trilinos/Trilinos/pull/15011 combined with STK patch in https://github.com/trilinos/Trilinos/pull/14964/changes addresses STK issues. Marking as resolved.

eugeneswalker (dismissed):

kuberry:
@spackbot re-run pipeline

kuberry:
@spackbot re-run pipeline

kuberry:
@spackbot re-run pipeline

kuberry:
@spackbot re-run pipeline

kuberry:
@alanw0, this is looking like an STK issue with code in stk/stk_util/stk_util/diag/Timer.hpp, marked as deprecated. 
https://gitlab.spack.io/spack/spack-packages/-/jobs/20918908

Any suggestions on how we patch this to remove this deprecated call?

alanw0:
> @alanw0, this is looking like an STK issue with code in stk/stk_util/stk_util/diag/Timer.hpp, marked as deprecated. https://gitlab.spack.io/spack/spack-packages/-/jobs/20918908
> 
> Any suggestions on how we patch this to remove this deprecated call?

Is there a flag being used that makes deprecations into errors instead of warnings? Aren't deprecations usually supposed to just be warnings?
One approach might be to activate the macro STK_HIDE_DEPRECATED_CODE.
If necessary, we can push through a commit to remove the deprecated code.

cgcgcg:
What's the best way to move on here? The pipeline fails on HIP, but that's a known issue that has (afaik) absolutely nothing to do with Trilinos. Can we override the CI and merge?

kuberry:
@spackbot re-run pipeline

ccober6:
@sebrowne,  can you do a review of this, and let's see if we can get it merged.

ccober6:
I looks like I cannot add you as a reviewer. Not in an appropriate group?

sebrowne (approved):
Trilinos changes look good to me.  I really don't like the change in enablement logic around `gtest` and the Trilinos packages because of the aforementioned TriBITS bug, but I think it's probably the best incremental solution as far as practicality goes.

eugeneswalker (approved):

kshea21 (approved):
Looks good to me, thanks everyone!

PR #3364, opened 2026-02-11
py-tensorflow: env.set(..., str)

haampie:


adamjstewart (approved):
I thought we fixed all these when I added type hints?

haampie:
The type of `spec["dep"]` is some opaque wrapper :(

PR #3365, opened 2026-02-11
py-pillow: add v12.1.1

adamjstewart:
https://pillow.readthedocs.io/en/stable/releasenotes/12.1.1.html

Not sure if we should deprecate older versions? Also wondering if we should deprecate py-pillow-simd, which now has multiple CVEs piling up...


RikkiButler20 (approved):
Verified sha256. LGTM

PR #3370, opened 2026-02-11
py-torchmetrics: does not support setuptools 82+

adamjstewart:
Another package that will break with setuptools 82.


alecbcs (approved):
Looks good to me. Thanks!

PR #3371, opened 2026-02-11
Fix flag handler condition for GCC 15 compatibility

MBlaschek:
this also happens for earlier versions, 1.23.


michaelkuhn (changes requested):
This breaks gdbm@1.24, since it also needs the flag.

haampie:
`@:1.24` includes 1.23

MBlaschek:
yes you are right! Sorry for wasting your time on this!

PR #3372, opened 2026-02-11
ci: bump spack

haampie:
Ensures build environments are unified, should help with py-setuptools
etc.

Also allows us to merge #3259  afterwards.


haampie:
fyi @adamjstewart

haampie:
The only rebuild in e4s (except for a known instability with boost) is:

```
nrm/ehovmht
```

This is actually a case where it fixes the issue it's meant to fix :)

Previously it used `py-setuptools@80.9.0` as a build dep. Now it uses `py-setuptools@59.4.0` as a build dep.

That's correct because of the runtime dependency:

```
nrm -> py-pyzmq -> py-gevent -> py-setuptools@=59.4.0
```

alalazo (approved):

PR #3373, opened 2026-02-11
Use more padding only for the spec being built

alalazo:
In some DAGs there could be more instances of the same package. We want to use "morepadding" only for the path of the spec being built, not for the other dependencies.


alalazo:
Example where this may cause failures because of  two `gcc-runtimes`: https://gitlab.spack.io/spack/spack-packages/-/jobs/20380889

kwryankrattiger (approved):
Thanks!

haampie:
build failure unrelated

PR #3374, opened 2026-02-11
ci: sprinkle `target=`

haampie:
package specific requirements override `all`, leading to other arch
choises among some packages

### DIFF
--- a/stacks/e4s-neoverse-v2/spack.yaml
+++ b/stacks/e4s-neoverse-v2/spack.yaml
@@ -45,7 +45,7 @@ spack:
     paraview:
       require: "+examples target=neoverse_v2 %gcc"
     boost:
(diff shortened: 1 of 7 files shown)


adamjstewart:
Do we consider this a bug? Yes, we should have the ability to override `all` for package-specific config, but if the specs don't collide, why not append instead of overwrite?

haampie:
Because that's hard to define for the use cases of "all clang except package foo with gcc". %clang and %gcc do not conflict. It gets too subtle.

I agree current semantics are suboptimal :). Maybe just needs something more ergonomic for `target`. Compiler requirements don't make sense in `all` anyways. In the future 'target' also doesn't make sense cause it wouldn't apply to scripts/data packages.

eugeneswalker (approved):

PR #3379, opened 2026-02-11
py-fenics-dolfinx 0.10: bump version, refinements

jhale:
Depends on #3369


repos/spack_repo/builtin/packages/py_fenics_dolfinx/package.py
```diff
@@ -82,11 +87,12 @@ class PyFenicsDolfinx(PythonPackage):
         depends_on("py-petsc4py", type=("build", "run"))
         depends_on("py-slepc4py", type=("build", "run"))
 
-    depends_on("py-cffi@:1.16", type=("build", "run"))
```
jhale, line 76: This was added due to upstream bugs and is now unnecessary

repos/spack_repo/builtin/packages/py_fenics_dolfinx/package.py
```diff
@@ -82,11 +87,12 @@ class PyFenicsDolfinx(PythonPackage):
...
+    depends_on("py-cffi", type=("build", "run"))
 
+    depends_on("py-nanobind@2.5:", when="@0.10:", type="build")
     depends_on("py-nanobind@2:", when="@0.9:", type="build")
     depends_on("py-nanobind@1.8:1.9", when="@0.8", type="build")
-    depends_on("py-scikit-build-core@0.10: +pyproject", when="@0.10:", type="build")
```
jhale, line 80: Checked, this lower bound on scikit-build-core goes back to 0.9

repos/spack_repo/builtin/packages/py_fenics_dolfinx/package.py
```diff
@@ -74,6 +77,8 @@ class PyFenicsDolfinx(PythonPackage):
     depends_on("py-numpy@1.21:", type=("build", "run"))
     depends_on("py-mpi4py", type=("build", "run"))
 
+    conflicts("~petsc4py", when="@:0.8", msg="+petsc4py is required for versions 0.8 and lower")
```
jhale: petsc4py was only an optional dependency from 0.9 and up.

repos/spack_repo/builtin/packages/py_fenics_dolfinx/package.py
```diff
@@ -74,6 +77,8 @@ class PyFenicsDolfinx(PythonPackage):
...
+    conflicts("~petsc4py", when="+slepc4py", msg="+slepc4py requires +petsc4py")
```
jhale, line 74: A user cannot force +slepc4py and also try and force ~petsc4py - niche but correct.

repos/spack_repo/builtin/packages/py_fenics_dolfinx/package.py
jhale: Will change this into a straightforward depend.

jhale: Done.

jhale:
@chrisrichardson Can you review this PR for me? Thanks.

chrisrichardson (approved):

chrisrichardson (approved):

repos/spack_repo/builtin/packages/py_fenics_dolfinx/package.py
```diff
@@ -19,6 +19,9 @@ class PyFenicsDolfinx(PythonPackage):
     license("LGPL-3.0-or-later")
 
     version("main", branch="main", no_cache=True)
+    version(
+        "0.10.0.post4", sha256="3f827a88ab52843fbd7a5cc7814ecba165bdec65fd10df05eb031c286e8cd605"
+    )
```
alalazo, line 24: There is already post5, should we update the PR?

alalazo (commented):
One comment. I'm also curious why the package is versioned `.postN` but never bumps the patch release, if anyone knows

jhale:
Yes - the postx changes were related to packaging and test errors, rather than library fixes. If a user has 0.10.0 installed then it is as good as a post release. That said, I see no reason not to use the 'best' release here.

repos/spack_repo/builtin/packages/py_fenics_dolfinx/package.py
jhale: Sorry I should have explained that; adding post5 would require a PR on fenics-dolfinx - there is no functional difference except unit testing, so post4 is OK with us. We will be releasing 0.11 soon in any case.

jhale:
@alalazo Could we get this merged? Thanks.

becker33 (approved):

PR #3383, opened 2026-02-12
[hypre]: 64-bit integer support

victorapm:
- Add logic in `hypre` for not allowing building with full 64-bit integers support and external BLAS/LAPACK (which is 32-bit by default). Note that building `hypre +lapack +mixedint` is still allowed
- Update `petsc +hypre+int64` so it asks for `hypre+mixedint` instead of `hypre+int64`. The new method is the preferred way of getting 64-bit integers support in hypre

Related to issue https://github.com/hypre-space/hypre/issues/1467


victorapm:
@jczhang07 @balay could you take a look at the petsc changes? Thanks!

repos/spack_repo/builtin/packages/petsc/package.py
```diff
@@ -320,7 +320,9 @@ class Petsc(Package, CudaPackage, ROCmPackage):
     depends_on("hypre~fortran", when="+hypre~fortran")
     depends_on("hypre+complex", when="+hypre+complex")
     depends_on("hypre~complex", when="+hypre~complex")
-    depends_on("hypre+int64", when="+hypre+int64")
+    depends_on("hypre+mixedint+cuda", when="+hypre+int64+cuda")
+    depends_on("hypre+mixedint+rocm", when="+hypre+int64+rocm")
+    depends_on("hypre+mixedint", when="+hypre+int64")
```
jczhang07, line 325: @victorapm In petsc's Hypre configure file `hypre.py`, where `self.getDefaultIndexSize()` returns petsc's index size, we have
```python
    if self.getDefaultIndexSize() == 64:
      if cudabuild or hipbuild or syclbuild: # HYPRE 2.23 supports only mixedint configurations with CUDA/HIP/SYCL
        args.append('--enable-bigint=no --enable-mixedint=yes')
      else:
        args.append('--enable-bigint')
```
Is it correct?  If so, shouldn't you have  

`depends_on("hypre+int64", when="+hypre+int64")`

victorapm: Hi Junchao, maybe we need to update that in PETSc... We generally recommend mixedint to get 64-bit support in hypre. bigint was the old way (that promotes all integers to 64-bits)

jczhang07: change into this?

```python
    if self.getDefaultIndexSize() == 64:
        args.append('--enable-bigint=no --enable-mixedint=yes')
```

victorapm: Right, you can even omit `--enable-bigint=no` since that's the default already

jczhang07: OK, let me revise `petsc.py`.

jczhang07 (approved):

becker33 (approved):

PR #3384, opened 2026-02-12
NAMD: add 3.0.2

alephpiece:
Added NAMD 3.0.2:
- Unset `PLATFORM` to avoid leaking external `PLATFORM` into NAMD's HIP make logic, which could incorrectly inject CUDA defines/headers in ROCm-only builds.
- Add a patch to override `HIPARCH`.
- Add a patch to fix missing `CudaLocalRecord::num_inline_peer` error.


jcphill (approved):
Shouldn't break anything.

jcphill (approved):

alalazo (approved):
Maintainers approved.

PR #3385, opened 2026-02-12
fix(hdf5): remove LD_LIBRARY_PATH setup in run environment

### LINKED ISSUE #3380: hdf5 setting `LD_LIBRARY_PATH`
I noticed that the hdf5 package is setting `LD_LIBRARY_PATH` https://github.com/spack/spack-packages/blob/c1e5ffc3e3cbdc73d5b7320f30382ab4879c88de/repos/spack_repo/builtin/packages/hdf5/package.py#L514-L518

As a consequence an environment view inherits this `LD_LIBRARY_PATH` pointing to the environment view `lib` directory. This exposes all libraries of the view and in my case makes some executables segfault (`nano`, `htop`, ...).

To reproduce:
```
spack env create ./local-env
spack env activate ./local-env
spack add hdf5
spack install
echo $LD_LIBRARY_PATH
```
It should show a path containing `local-env/.spack-env/view/lib`.

I was wondering if we could find an alternative approach ? cc @brtnfld @byrnHDF @gheber @hyoklee @lkurz @lrknox

hyoklee:
Removed unnecessary LD_LIBRARY_PATH setup in run environment.

close #3380

### DIFF
--- a/repos/spack_repo/builtin/packages/hdf5/package.py
+++ b/repos/spack_repo/builtin/packages/hdf5/package.py
@@ -512,10 +512,7 @@ class Hdf5(CMakePackage):
         env.set("SZIP_INSTALL", self.spec["szip"].prefix)
 
     def setup_run_environment(self, env: EnvironmentModifications) -> None:
-        # According to related github posts and problems running test_install
-        # as a stand-alone test, it appears the lib path must be added to
-        # LD_LIBRARY_PATH.
-        env.append_path("LD_LIBRARY_PATH", self.prefix.lib)
+        pass
 
     def cmake_args(self):
         spec = self.spec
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/hdf5/package.py
```diff
@@ -512,10 +512,7 @@ def setup_build_environment(self, env: EnvironmentModifications) -> None:
         env.set("SZIP_INSTALL", self.spec["szip"].prefix)
 
     def setup_run_environment(self, env: EnvironmentModifications) -> None:
```
haampie: Can you remove the function altogether?

haampie (approved):

haampie:
Failures seem unrelated, let's double check:

@spackbot run pipeline

PR #3386, opened 2026-02-12
rccl: add patch to limit max memory usage, reduce number of targets

haampie:
ref https://github.com/ROCm/rocm-systems/pull/3231

Cause we don't use cgroups v2, the computation of "available memory" doesn't
work, and we end up with up to 16 jobs, with user reports claiming 32GB per job
is needed.

Edit: in practice it's less and depends on number of GPU architectures selected.
After #3418, 8GB per link job seems reasonable.

This patch allows us to set a maximum on the amount of memory used, provided
an accurate estimate for memory per job is given. Currently they are hard-coded
at 32GB and 8GB respectively, so we link with at most 4 jobs (or fewer if less
memory is available on the system)

### DIFF
--- a/repos/spack_repo/builtin/packages/rccl/memory-3231.patch
+++ b/repos/spack_repo/builtin/packages/rccl/memory-3231.patch
@@ -1,0 +1,45 @@
+diff --git a/CMakeLists.txt b/CMakeLists.txt
+index cb5d34e1caf..86b97a29c93 100644
+--- a/CMakeLists.txt
++++ b/CMakeLists.txt
+@@ -1364,24 +1364,35 @@ execute_process(
+   COMMAND bash "-c" "cat /sys/fs/cgroup/memory.max"
+   OUTPUT_VARIABLE memory_max_string)
+ if (${memory_max_string} MATCHES "^[0-9]+")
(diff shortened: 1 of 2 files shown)


haampie:
@afzpatel just FYI, `rccl` managed to make a node with 2TB of memory go OOM :) can you potentially convince the relevant people to prioritize less resource intensive builds? It's nice the ecosystem is open source, but when nobody can build it... Limiting this to 32GB of memory, which requires a patch, we cannot build it cause it takes 6h+.

What's the idea then? You have to have 256GB of memory to build it in "reasonable" time?

Edit: seems like we only build for one amd gpu target in the rest of the stack, so applying that to rccl too

haampie:
With 1 target peak memory is apparently 21.7GB and 3 jobs were used. I guess we could do 8GB per job per arch or something?

zackgalbreath (approved):

haampie:
Giving @afzpatel @renjithravindrankannath and @srekolam an opportunity to look at this.

srekolam (approved):

haampie:
Rebased this after #3418 was extracted and merged.

afzpatel (approved):

PR #3389, opened 2026-02-12
openblas: add variant static

mkrack:
Variant `+static` is needed for statically linked CP2K builds


mathomp4 (approved):

mkrack:
It's not clear to me if some tests fail because of the new variant `static` or the new version `0.3.31`. Maybe, it is better to add the new variant and the new version in two steps?

RikkiButler20 (dismissed):
LGTM

RikkiButler20:
@spackbot re-run pipeline

mkrack:
Add only new variant `static`  for the time being and postpone the version update.

PR #3390, opened 2026-02-12
gnu packages: bump

haampie:
Bump 22 gnu packages (builds for me with missing_library_policy: error)

Edit: interesting moment for https://www.gnu.org/ to be down.

### DIFF
--- a/repos/spack_repo/builtin/packages/aspell/package.py
+++ b/repos/spack_repo/builtin/packages/aspell/package.py
@@ -22,12 +22,14 @@ class Aspell(AutotoolsPackage, GNUMirrorPackage):
 
     license("LGPL-2.1-or-later")
 
+    version("0.60.8.2", sha256="57fe4863eae6048f72245a8575b44b718fb85ca14b9f8c0afc41b254dfd76919")
     version("0.60.8.1", sha256="d6da12b34d42d457fa604e435ad484a74b2effcd120ff40acd6bb3fb2887d21b")
     version("0.60.8", sha256="f9b77e515334a751b2e60daab5db23499e26c9209f5e7b7443b05235ad0226f2")
     version("0.60.6.1", sha256="f52583a83a63633701c5f71db3dc40aab87b7f76b29723aeb27941eff42df6e1")
 
-    depends_on("c", type="build")  # generated
-    depends_on("cxx", type="build")  # generated
+    depends_on("c", type="build")
+    depends_on("cxx", type="build")
+    depends_on("ncurses", type="link")
 
     patch("fix_cpp.patch")
     patch("issue-519.patch", when="@:0.60.6.1")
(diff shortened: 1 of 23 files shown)


repos/spack_repo/builtin/packages/aspell/package.py
```diff
@@ -22,12 +22,14 @@ class Aspell(AutotoolsPackage, GNUMirrorPackage):
...
+    version("0.60.8.2", sha256="57fe4863eae6048f72245a8575b44b718fb85ca14b9f8c0afc41b254dfd76919")
     version("0.60.8.1", sha256="d6da12b34d42d457fa604e435ad484a74b2effcd120ff40acd6bb3fb2887d21b")
     version("0.60.8", sha256="f9b77e515334a751b2e60daab5db23499e26c9209f5e7b7443b05235ad0226f2")
     version("0.60.6.1", sha256="f52583a83a63633701c5f71db3dc40aab87b7f76b29723aeb27941eff42df6e1")
 
-    depends_on("c", type="build")  # generated
-    depends_on("cxx", type="build")  # generated
+    depends_on("c", type="build")
+    depends_on("cxx", type="build")
+    depends_on("ncurses", type="link")
```
alecbcs, line 32: Why are we adding a link dependency on ncurses here?

haampie:
```
$ readelf -d ./bin/aspell | grep NEEDED
 0x0000000000000001 (NEEDED)             Shared library: [libaspell.so.15]
 0x0000000000000001 (NEEDED)             Shared library: [libncursesw.so.6]
 0x0000000000000001 (NEEDED)             Shared library: [libtinfo.so.6]
 0x0000000000000001 (NEEDED)             Shared library: [libstdc++.so.6]
 0x0000000000000001 (NEEDED)             Shared library: [libc.so.6]
```

alecbcs: Gotcha. I know we put some logic into Spack a while back to detect when we were bleeding through system dependencies. Is that able to catch this? I hadn't seen it pop up as a warning when building `aspell` on Linux.

haampie: Yeah, `config:missing_library_policy:error` for the build, and `spack verify libraries` to do the same after the build in case you hadn't turned it on.

ELF platforms only though, so doesn't work on macOS.

alecbcs: Awesome. I'll make sure to enable that for my builds to catch this stuff in the future.

haampie: I thought we ran this in CI too, but maybe not consistently

haampie:
Build failure of adios2 is unrelated and fixed by https://github.com/spack/spack-packages/pull/3416. If that's all I'd suggest force merging this.

alecbcs (approved):
Looks good to me. Thanks @haampie!

PR #3393, opened 2026-02-12
build_systems: use %pkg in when conditions

haampie:
Good style to refer to direct deps by `%` instead of `^` in Spack v1.x.

Also extract the 72 `+cuda`s to the common `with when(...)` bit.


haampie:
CI failure is unrelated: rccl wants more than 2TB of memory (ref #3386), can be force merged.

rbberger (approved):

alalazo:
I'm going to check if it has any effect on concretizer performance before merging.

alalazo:
Confirmed it doesn't have noticeable effects:

[radiuss.develop.csv](https://github.com/user-attachments/files/25484983/radiuss.develop.csv)
[radiuss.pr.csv](https://github.com/user-attachments/files/25484984/radiuss.pr.csv)

(Removed `py-cinemasci` from the plot cause it's just noise)

comparison

PR #3394, opened 2026-02-12
py-tensorboard: setuptools 82+ not yet compatible

adamjstewart:


alecbcs (approved):

PR #3396, opened 2026-02-12
py-language-data: new package

qwertos:

### DIFF
--- a/repos/spack_repo/builtin/packages/py_language_data/package.py
+++ b/repos/spack_repo/builtin/packages/py_language_data/package.py
@@ -1,0 +1,21 @@
+# Copyright Spack Project Developers. See COPYRIGHT file for details.
+#
+# SPDX-License-Identifier: (Apache-2.0 OR MIT)
+
+from spack_repo.builtin.build_systems.python import PythonPackage
+from spack.package import *
+
+
+class PyLanguageData(PythonPackage):
+    """Supplementary data about languages used by the langcodes module"""
+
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/py_language_data/package.py
```diff
@@ -0,0 +1,22 @@
...
+
+    homepage = "https://github.com/georgkrause/language_data"
+    pypi = "language_data/language_data-1.3.0.tar.gz"
+
+    license("MIT")
+
+    version("1.3.0", sha256="7600ef8aa39555145d06c89f0c324bf7dab834ea0b0a439d8243762e3ebad7ec")
+
+    depends_on("py-setuptools@60:", type="build")
+    depends_on("py-setuptools@8:", type="build")
```
tldahlgren: Per https://github.com/georgkrause/language_data/blob/v1.3.0/pyproject.toml#L38:


```suggestion
    depends_on("py-setuptools-scm@8:", type="build")
```

tldahlgren (changes requested):
Confirmed the homepage and version sha256. There's at least one dependency change needed.

Although they don't explicitly specify a dependency on `python@3.9:`, it is implied in `classifiers`.

repos/spack_repo/builtin/packages/py_language_data/package.py
qwertos: Good catch! Fixed!

tldahlgren (approved):

PR #3397, opened 2026-02-12
py-thinc: added versions 8.1.10 and 8.3.10; and increased readability of complex dependencies

qwertos:

### DIFF
--- a/repos/spack_repo/builtin/packages/py_thinc/package.py
+++ b/repos/spack_repo/builtin/packages/py_thinc/package.py
@@ -70,5 +70,5 @@ class PyThinc(PythonPackage):
         depends_on("py-contextvars@2.4:2", when="@8.1.10:^python@3.6")
 
         # Historical
-        depends_on("py-plac@0.9.6:1.1")
-        depends_on("py-tqdm@4.10:4")
+        depends_on("py-plac@0.9.6:1.1", when="@:7.4.1")
+        depends_on("py-tqdm@4.10:4", when="@:7.4.1")
--- a/repos/spack_repo/builtin/packages/py_thinc/package.py
+++ b/repos/spack_repo/builtin/packages/py_thinc/package.py
@@ -16,6 +16,7 @@ class PyThinc(PythonPackage):
 
     license("MIT")
 
+    version("8.3.10", sha256="5a75109f4ee1c968fc055ce651a17cb44b23b000d9e95f04a4d047ab3cb3e34e")
     version("8.1.10", sha256="6c4a48d7da07e044e84a68cbb9b22f32f8490995a2bab0bfc60e412d14afb991")
     version("7.4.1", sha256="0139fa84dc9b8d88af15e648fc4ae13d899b8b5e49cb26a8f4a0604ee9ad8a9e")
     version("7.4.0", sha256="523e9be1bfaa3ed1d03d406ce451b6b4793a9719d5b83d2ea6b3398b96bc58b8")
@@ -23,13 +24,18 @@ class PyThinc(PythonPackage):
     depends_on("c", type="build")  # generated
     depends_on("cxx", type="build")  # generated
 
-    depends_on("python@3.6:", type=("build", "run"), when="@8.1.10:")
+    depends_on("python@3.6:", type=("build", "run"), when="@8.1.10")
+    depends_on("python@3.10:3.14", type=("build", "run"), when="@8.3.10:")
+
     depends_on("py-setuptools", type="build")
-    depends_on("py-cython@0.25:2", type="build", when="@8.1.10:")
+
+    depends_on("py-cython@0.25:2", type="build", when="@8.1.10")
+    depends_on("py-cython@3", type="build", when="@8.3.10:")
 
     with default_args(type=("build", "run")):
         depends_on("py-blis@0.4.0:0.4", when="@:7.4.1")
-        depends_on("py-blis@0.7.8:0.7", when="@8.1.10:")
+        depends_on("py-blis@0.7.8:0.7", when="@8.1.10")
+        depends_on("py-blis@1.3", when="@8.3.10:")
 
         depends_on("py-murmurhash@0.28:1.0", when="@:7.4.1")
         depends_on("py-murmurhash@1.0.2:1.0", when="@8.1.10:")
@@ -52,8 +58,11 @@ class PyThinc(PythonPackage):
 
         depends_on("py-numpy@1.7:", when="@:7.4.1")
         depends_on("py-numpy@1.15:", when="@8.1.10:")
+        depends_on("py-numpy@2:", when="@8.1.10:")
 
-        depends_on("py-pydantic@1.7.4:1.7,1.9:1.10", when="@8.1.10:")
+        depends_on("py-pydantic@1.7.4:1.7,1.9:1.10", when="@8.1.10")
+        depends_on("py-pydantic@2", when="@8.3.10:")
+
         depends_on("py-packaging@20:", when="@8.1.10:")
 
         depends_on("py-dataclasses@0.6:0", when="@8.1.10:^python@:3.6")
--- a/repos/spack_repo/builtin/packages/py_thinc/package.py
+++ b/repos/spack_repo/builtin/packages/py_thinc/package.py
@@ -16,20 +16,50 @@ class PyThinc(PythonPackage):
 
     license("MIT")
 
+    version("8.1.10", sha256="6c4a48d7da07e044e84a68cbb9b22f32f8490995a2bab0bfc60e412d14afb991")
     version("7.4.1", sha256="0139fa84dc9b8d88af15e648fc4ae13d899b8b5e49cb26a8f4a0604ee9ad8a9e")
     version("7.4.0", sha256="523e9be1bfaa3ed1d03d406ce451b6b4793a9719d5b83d2ea6b3398b96bc58b8")
 
     depends_on("c", type="build")  # generated
     depends_on("cxx", type="build")  # generated
 
+    depends_on("python@3.6:", type=("build", "run"), when="@8.1.10:")
     depends_on("py-setuptools", type="build")
-    depends_on("py-murmurhash@0.28:1.0", type=("build", "run"))
-    depends_on("py-cymem@2.0.2:2.0", type=("build", "run"))
-    depends_on("py-preshed@1.0.1:3.0", type=("build", "run"))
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/py_thinc/package.py
```diff
@@ -16,20 +16,61 @@ class PyThinc(PythonPackage):
...
-    depends_on("py-blis@0.4.0:0.4", type=("build", "run"))
-    depends_on("py-wasabi@0.0.9:1.0", type=("build", "run"))
-    depends_on("py-srsly@0.0.6:1.0", type=("build", "run"))
-    depends_on("py-catalogue@0.0.7:1.0", type=("build", "run"))
-    depends_on("py-numpy@1.7:", type=("build", "run"))
-    depends_on("py-plac@0.9.6:1.1", type=("build", "run"))
-    depends_on("py-tqdm@4.10:4", type=("build", "run"))
+
+    depends_on("py-cython@0.25:2", type="build", when="@8.1.10")
+    depends_on("py-cython@3", type="build", when="@8.3.10:")
```
tldahlgren, line 35: Can you reverse the order so the dependencies of newer versions are listed before those of older?

Same below.

repos/spack_repo/builtin/packages/py_thinc/package.py
```diff
@@ -16,20 +16,61 @@ class PyThinc(PythonPackage):
...
+        depends_on("py-srsly@2.4:2", when="@8.1.10:")
+
+        depends_on("py-catalogue@0.0.7:1.0", when="@:7.4.1")
+        depends_on("py-catalogue@2.0.4:2.0", when="@8.1.10:")
+
+        depends_on("py-confection@0.0.1:0", when="@8.1.10:")
+
+        depends_on("py-numpy@1.7:", when="@:7.4.1")
+        depends_on("py-numpy@1.15:", when="@8.1.10:")
+        depends_on("py-numpy@2:", when="@8.1.10:")
```
tldahlgren: Where is this coming from for `8.1.10`? I'm not seeing a requirement for `py-numpy@2:`.

But I do see it for `8.3.x`: https://github.com/explosion/thinc/blob/v8.3.x/pyproject.toml#L9

repos/spack_repo/builtin/packages/py_thinc/package.py
```diff
@@ -16,20 +16,61 @@ class PyThinc(PythonPackage):
 
     license("MIT")
 
+    version("8.3.10", sha256="5a75109f4ee1c968fc055ce651a17cb44b23b000d9e95f04a4d047ab3cb3e34e")
+    version("8.1.10", sha256="6c4a48d7da07e044e84a68cbb9b22f32f8490995a2bab0bfc60e412d14afb991")
```
tldahlgren, line 20: Note there is a newer patch release: `8.1.12`.

repos/spack_repo/builtin/packages/py_thinc/package.py
```diff
@@ -16,20 +16,61 @@ class PyThinc(PythonPackage):
...
-    depends_on("py-preshed@1.0.1:3.0", type=("build", "run"))
-    depends_on("py-blis@0.4.0:0.4", type=("build", "run"))
-    depends_on("py-wasabi@0.0.9:1.0", type=("build", "run"))
-    depends_on("py-srsly@0.0.6:1.0", type=("build", "run"))
-    depends_on("py-catalogue@0.0.7:1.0", type=("build", "run"))
-    depends_on("py-numpy@1.7:", type=("build", "run"))
-    depends_on("py-plac@0.9.6:1.1", type=("build", "run"))
-    depends_on("py-tqdm@4.10:4", type=("build", "run"))
+
+    depends_on("py-cython@0.25:2", type="build", when="@8.1.10")
```
tldahlgren, line 34:
```suggestion
    depends_on("py-cython@0.25:2", type="build", when="@8.1.10:")
```

repos/spack_repo/builtin/packages/py_thinc/package.py
```diff
@@ -16,20 +16,61 @@ class PyThinc(PythonPackage):
...
+    depends_on("py-cython@3", type="build", when="@8.3.10:")
+
+    with default_args(type=("build", "run")):
+        depends_on("py-blis@0.4.0:0.4", when="@:7.4.1")
+        depends_on("py-blis@0.7.8:0.7", when="@8.1.10")
```
tldahlgren, line 39:
```suggestion
        depends_on("py-blis@0.7.8:0.7", when="@8.1.10:")
```

repos/spack_repo/builtin/packages/py_thinc/package.py
tldahlgren: There are some python version-dependent lower bounds for `py-numpy` in, e.g., https://github.com/explosion/thinc/blob/v8.1.12/pyproject.toml#L9.

repos/spack_repo/builtin/packages/py_thinc/package.py
```diff
@@ -16,20 +16,61 @@ class PyThinc(PythonPackage):
...
+        depends_on("py-catalogue@0.0.7:1.0", when="@:7.4.1")
+        depends_on("py-catalogue@2.0.4:2.0", when="@8.1.10:")
+
+        depends_on("py-confection@0.0.1:0", when="@8.1.10:")
+
+        depends_on("py-numpy@1.7:", when="@:7.4.1")
+        depends_on("py-numpy@1.15:", when="@8.1.10:")
+        depends_on("py-numpy@2:", when="@8.1.10:")
+
+        depends_on("py-pydantic@1.7.4:1.7,1.9:1.10", when="@8.1.10")
```
tldahlgren, line 65:
```suggestion
        depends_on("py-pydantic@1.7.4:1.7,1.9:1.10", when="@8.1.10:")
```

tldahlgren (commented):
Confirmed the new version sha256s and most dependencies. (Note there are newer (patch) releases.) 

It appears there's a missing dependency on `py-ml-datasets@0.2` (see https://github.com/explosion/thinc/blob/v8.3.x/requirements.txt#L10 and https://github.com/explosion/thinc/blob/v8.1.10/requirements.txt#L10).

Ideally dependencies would be listed with those for newer versions appearing first, and not only for readability. (See comment.)

See comments for more suggestions.

tldahlgren (changes requested):
See comments.

repos/spack_repo/builtin/packages/py_thinc/package.py
qwertos: Please see conversation at https://github.com/spack/spack-packages/pull/2536#discussion_r2562078593

repos/spack_repo/builtin/packages/py_thinc/package.py
qwertos: Are you asking me to add that version to this PR?

tldahlgren: Up to you. Sometimes ppl add every version to packages even when older, especially patch, releases don't necessarily add functional value.

repos/spack_repo/builtin/packages/py_thinc/package.py
qwertos: This change would seem to conflict with line 35. Can you explain what is intended here?

repos/spack_repo/builtin/packages/py_thinc/package.py
qwertos: Good catch here! Will fix.

qwertos: Good catch here! Will fix.

repos/spack_repo/builtin/packages/py_thinc/package.py
qwertos: This change would seem to conflict with line 66. Can you explain what is intended here?

repos/spack_repo/builtin/packages/py_thinc/package.py
qwertos: This isn't an issue for [8.1.10](https://github.com/explosion/thinc/blob/v8.1.10/pyproject.toml).

repos/spack_repo/builtin/packages/py_thinc/package.py
qwertos: This change would seem to conflict with line 40. Can you explain what is intended here?

repos/spack_repo/builtin/packages/py_thinc/package.py
tldahlgren: There isn't a problem with overlap. The concretizer handles it.

What adding the colon allows is it ensures the dependency applies should someone decide (someday) to add a patch release between it and the next minor release.

qwertos: AFAIK that's only true if the dependency version ranges don't have an upper bound. Here, the concretizer would have to find a version of py-pydantic that is both <= 1.10 and of major version 2; an impossible task.

tldahlgren: I thought your question was why I suggested appending the colon to the `when` condition. If someone were to add the version `8.1.12` then, without the colon or any other changes here, this package would *not* have a dependency on `py-pydantic`. In other words, with the colon, there would not be a need to make any changes to the `py-pydantic` dependency.

qwertos: Maybe I could do something like `@8.1.10:8.1` but I would not feel comfortable making an intentional declaration like that without going through all published versions within that range and confirming it's accurate. It very quickly can get to the point where I may as well add every version (at least the versions between the oldest and newest `version()` directives) published on pypi since I'd be doing that work anyway. At the end of the day, IMHO it's the responsibility of the person who would add `version("8.1.12")` to make sure that the `depends_on()` directives are accurate for that version. And if those dependencies get missed, that's the value of having peer checkers.

repos/spack_repo/builtin/packages/py_thinc/package.py
tldahlgren: The concretizer will pick the most relevant version range. But I understand your hesitation.

repos/spack_repo/builtin/packages/py_thinc/package.py
tldahlgren: I will not push this further. We can leave it for later if it becomes relevant.

tldahlgren (approved):
Thanks.

PR #3399, opened 2026-02-12
cudnn: add rpath to helper libraries for cudnn 8

Markus92:
CuDNN versions prior to 7 shipped with one library: `libcudnn.so`. This made linking easy as there is only one thing to link to.

CuDNN 8 changed this: the CuDNN library was split into 6 libraries and a shim `libcudnn.so`. To make sure `libcudnn.so` can find it's helper libraries, it has its rpath set to `$ORIGIN`. As all libs come into the same folder, this works fine for the most part. However, the individual split libraries do not have an rpath set, so they cannot see each other. When using pytorch this can lead to library not found errors during runtime, as the libraries do link to each other: e.g. `libcudnn_adv_infer.so` links to `libcudnn_ops_infer.so.8`

CuDNN 9 has the rpath set to `$ORIGIN` in all helper libraries off-the-shelf, so is not affected by this behavior. This PR adds that behavior to CuDNN 8.

### DIFF
--- a/repos/spack_repo/builtin/packages/cudnn/package.py
+++ b/repos/spack_repo/builtin/packages/cudnn/package.py
@@ -454,13 +454,12 @@ class Cudnn(Package):
     # and also not an issue in cudnn@9 comes with RPATH set to $ORIGIN
     # in all those helper libraries.
     # Note that libcudnn.so does come with the RPATH set.
-    @run_after("install")
+    @run_after("install", when="@8")
     def ensure_rpaths(self):
-        if self.spec.satisfies("@8"):
-            patchelf = which("patchelf")
-            for path in Path(self.prefix).rglob("*.so.*"):
-                if not path.is_symlink() and not path.name.startswith("libcudnn.so"):
-                    patchelf("--set-rpath", "$ORIGIN", str(path))
+        patchelf = which("patchelf")
+        for path in Path(self.prefix.lib).rglob("lib*.so.*"):
+            if not path.is_symlink() and not path.name.startswith("libcudnn.so"):
+                patchelf("--set-rpath", "$ORIGIN", str(path))
 
     # contains precompiled binaries without rpaths
     unresolved_libraries = ["*"]
--- a/repos/spack_repo/builtin/packages/cudnn/package.py
+++ b/repos/spack_repo/builtin/packages/cudnn/package.py
@@ -4,6 +4,7 @@
 
 import os
 import platform
+from pathlib import Path
 
 from spack_repo.builtin.build_systems.generic import Package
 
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/cudnn/package.py
```diff
@@ -445,5 +448,19 @@ def install(self, spec, prefix):
...
+    # issues if the LD_LIBRARY_PATH is not set at runtime.
+    # This is not an issue in cudnn@:7 as those only have one libcudnn.so file,
+    # and also not an issue in cudnn@9 comes with RPATH set to $ORIGIN
+    # in all those helper libraries.
+    # Note that libcudnn.so does come with the RPATH set.
+    @run_after("install")
+    def ensure_rpaths(self):
+        if self.spec.satisfies("@8"):
+            patchelf = which("patchelf")
+            for path in Path(self.prefix).rglob("*.so.*"):
```
haampie: maybe restrict to a lib dir and update to `lib*.so.*`?

repos/spack_repo/builtin/packages/cudnn/package.py
```diff
@@ -445,5 +448,19 @@ def install(self, spec, prefix):
             if os.path.isdir(target_include) and not os.path.isdir(prefix.include):
                 symlink(target_include, prefix.include)
 
+    # CuDNN 8 does not have rpaths set in the helper libraries, which can cause
+    # issues if the LD_LIBRARY_PATH is not set at runtime.
+    # This is not an issue in cudnn@:7 as those only have one libcudnn.so file,
+    # and also not an issue in cudnn@9 comes with RPATH set to $ORIGIN
+    # in all those helper libraries.
+    # Note that libcudnn.so does come with the RPATH set.
+    @run_after("install")
```
haampie:
```suggestion
    @run_after("install", when="@8")
```

saves a level of indent

Markus92: Didn't know that was possible, thanks!

repos/spack_repo/builtin/packages/cudnn/package.py
Markus92: That's probably better. Appreciate it

Markus92:
Thanks for the comments, @haampie ! Took care of them hopefully

haampie (approved):

haampie:
Would be good to test drive a bit, cause patchelf's mechanism for growing the dynamic section is a rather involved operation. Plus look if the file size is much larger than before (probably won't matter cause of how big the rest of cuda is).

I don't have an nvidia gpu to test on though

Markus92:
File sizes differ a bit but not much, 7mb on a 2.3gb package for CuDNN 8.9.7

Some simple tests seem to indicate it works (simple MNIST example). I'll deploy it across our users and report back if it fixes their issues.

haampie:
sounds good to me

PR #3404, opened 2026-02-13
munge: fix CVE-2026-25506

tgamblin:
Munge recently had [CVE-2026-25506](https://nvd.nist.gov/vuln/detail/CVE-2026-25506) reported. The CVE is no longer under embargo and is fixed in Munge `0.5.18`.

This adds Munge `0.5.18` along with patches for prior versions that fix the CVE.

### DIFF
--- a/repos/spack_repo/builtin/packages/munge/package.py
+++ b/repos/spack_repo/builtin/packages/munge/package.py
@@ -18,6 +18,7 @@ class Munge(AutotoolsPackage):
 
     license("LGPL-3.0-only")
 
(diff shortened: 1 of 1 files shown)


alalazo (approved):
Should we also deprecate versions from e.g. 2020 and before (that would leave 2 versions non -deprecated)?

tgamblin:
> Should we also deprecate versions from e.g. 2020 and before (that would leave 2 versions non -deprecated)?

We can do that later. The intent of this one is to give people who may have been using this munge package un-compromised versions of it, providing the patch with least disruption and allowing a clean backport to past package release (@kwryankrattiger @becker33 FYI).

I don't think updates like this for CVEs *and* deprecations should be addressed at the same time if it can be avoided.

PR #3405, opened 2026-02-13
miniforge3: add 26.1.1-3

afuetterer:
This PR adds 26.1.0-0 of miniforge3.

Question: Is the defintion of versions in the dict necessary? miniforge comes via a shell script from GitHub, the exact URL is constructed via `url_for_version`, e.g.

- Miniforge3-26.1.0-0-Linux-aarch64.sh
- Miniforge3-26.1.0-0-Linux-ppc64le.sh
- Miniforge3-26.1.0-0-Linux-x86_64.sh
- Miniforge3-26.1.0-0-MacOSX-arm64.sh
- Miniforge3-26.1.0-0-MacOSX-x86_64.sh
Ref: https://github.com/conda-forge/miniforge/releases/tag/26.1.0-0

With this approach there is no way of deprecating older versions, right?

What do you think?

### DIFF
--- a/repos/spack_repo/builtin/packages/miniforge3/package.py
+++ b/repos/spack_repo/builtin/packages/miniforge3/package.py
@@ -36,6 +36,12 @@ _versions = {
 }
 
 
+def _should_deprecate_version(version: str) -> bool:
+    major_version_to_deprecate = 20
+    major_version = int(version.split(".")[0])
+    return major_version < major_version_to_deprecate
+
+
 class Miniforge3(Package):
     """Miniforge3 is a minimal installer for conda and mamba specific to conda-forge."""
 
@@ -49,7 +55,7 @@ class Miniforge3(Package):
         key = f"{platform.system()}-{platform.machine()}"
         pkg = packages.get(key)
         if pkg:
-            version(ver, sha256=pkg[0], expand=False)
+            version(ver, sha256=pkg[0], expand=False, deprecated=_should_deprecate_version(ver))
 
     variant("mamba", default=True, description="Enable mamba support.")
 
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/miniforge3/package.py
alalazo: A way to deprecate could be:
```suggestion
            version(ver, sha256=pkg[0], expand=False, deprecated=<some expression>)
```
The expression imo should be the simplest thing that works, given the deprecation pattern.

afuetterer: Thanks for the suggestion, let me try that.

I think if a version is from 2020 it can be deprecated?
https://github.com/conda-forge/miniforge/releases/tag/4.8.3-4

afuetterer: Maybe something like this:

```python
    for _version, packages in _versions.items():
        key = f"{platform.system()}-{platform.machine()}"
        pkg = packages.get(key)
        if pkg:
            major = int(_version.split(".")[0])
            version(_version, sha256=pkg[0], expand=False, deprecated=major<20)
```

I need some help with the expression here. It works for the given examples, but feels a bit fragile.

afuetterer: Not sure how to proceed. Should I close this?

afuetterer:
New release: https://github.com/conda-forge/miniforge/releases/tag/26.1.1-3

afuetterer:
I bumped the version to reflect the newer release.
I also added a test to verify installation.

I added a helper function to deprecate versions as @alalazo suggested. I chose 20 as a threshold here. Or should versions 24 and 25 also be considered deprecated?

repos/spack_repo/builtin/packages/miniforge3/package.py
```diff
@@ -82,3 +92,17 @@ def setup_run_environment(self, env: EnvironmentModifications) -> None:
         if "+mamba" in self.spec:
             filename = self.prefix.etc.join("profile.d").join("mamba.sh")
             env.extend(EnvironmentModifications.from_sourcing_file(filename))
+
+    @run_after("install")
```
afuetterer, line 96: This is the first test I wrote for a spack package. I followed the docs here: https://spack.readthedocs.io/en/latest/packaging_guide_testing.html#adding-installation-phase-tests

Is this alright?

alalazo (approved):

PR #3407, opened 2026-02-13
SIRIUS: patch hipDeviceProp_t.uuid for ROCm<6

alephpiece:
`hipDeviceProp_t.uuid` is defined on ROCm 6.0 and later, which causes compilation to fail on older ROCm versions (e.g., ROCm 5.x). Patch `acc.hpp` to switch to using `hipUUID` and the `DeviceGetUuid` API directly for the ROCm backend, safely resolving the build issue for ROCm < 6.0.


hainest:
@alephpiece Have you successfully tested this patch?

alephpiece:
This patch has been merged into the upstream SIRIUS, so I'll close this PR.

PR #3408, opened 2026-02-13
rocm.py: add a few missing amd_gpu_target values

haampie:
taken from https://llvm.org/docs/AMDGPUUsage.html

... would be nice to handle these variants better and treat AMD GPU targets like the CPU `target=...`, especially cause of the variants `xnack`, `wavefrontsize64`, `cumode`, leading to a combinatorial explosion of `<target-id> ::== <processor> ( "+" <target-feature> )*`, plus the fact that generic targets exist now.


alecbcs (approved):
Looks good to me. Thanks!

PR #3411, opened 2026-02-13
py-pip: 26.0.1

haampie:
Extracted from mass bump PR because of build failures, let's see...

Edit: cause we have a handful `--global-option`s that need to be migrated.


bernhardkaindl:
Seen in the conflict editor: 26.1.2 and 26.1 have been merged already,
and as this PR is old, it should be obsolete and not be needed anymore.

PR #3418, opened 2026-02-14
ci: single target for rccl in ci

haampie:
Same as https://github.com/spack/spack-packages/pull/3386 but without the
patch, so this can be merged earlier.


haampie:
this commit was cherry-picked from an approved pr and should fix ci which is currently doing ~3 iterations to build rccl in 4h, failing repeatedly.

PR #3423, opened 2026-02-14
repo: inline filter_file(..., **kwargs)

haampie:
Type checkers take issue with `kwargs = {...}` + `f(**kwargs)` due to type
erasure. The pattern is often somewhat redundant and seems to have been
copied to match style of other packages.

* Inline `**kwargs`
* Remove kwargs that repeat defaults

Before: 1762 diagnostics, after: 1525 diagnostics; from `ty check --extra-search-path ~/spack/lib/spack --extra-search-path repos --ignore unresolved-reference --output-format concise`

<details>

<summary>was done automatically</summary>

```
#!/usr/bin/env python3

import ast
import sys

filter_file_defaults = {
    "string":False,
    "backup":False,
    "ignore_absent":False,
    "start_at":None,
    "stop_at":None,
    "encoding":"utf-8",
}



def check_filter_file_calls(filename):
    with open(filename, "r") as f:
        lines = f.readlines()

    tree = ast.parse("".join(lines), filename=filename)

    for node in ast.walk(tree):
        # find the last assignment to `kwargs` in the current scope
        if isinstance(node, ast.Assign) and isinstance(node.value, ast.Dict):
            for target in node.targets:
                if isinstance(target, ast.Name) and target.id == "kwargs":
                    kwargs_value = {}
                    for key, value in zip(node.value.keys, node.value.values):
                        if isinstance(key, ast.Constant):
                            kwargs_value[key.value] = ast.literal_eval(value)
                    print(f"{filename}:{node.lineno}:{node.col_offset}:{node.end_col_offset}: kwargs = {kwargs_value}")
                    # remove all defaults from kwargs_value
                    for k in filter_file_defaults:
                        if k in kwargs_value and kwargs_value[k] == filter_file_defaults[k]:
                            del kwargs_value[k]

        # either the funcion call is `filter_file(..., **kwargs)`
        # or its obj.filter(..., **kwargs)

        if not isinstance(node, ast.Call):
            continue
        
        if (
            getattr(node.func, "id", None) == "filter_file"
            or isinstance(node.func, ast.Attribute) and node.func.attr == "filter"
        ):
            if node.keywords:
                for kw in node.keywords:
                    if isinstance(kw.value, ast.Name) and kw.value.id == "kwargs":
                        if not kwargs_value:
                            # kwargs is empty, we delete this part from the function call
                            lines[kw.lineno - 1] = lines[kw.lineno - 1][:kw.col_offset] + lines[kw.lineno - 1][kw.end_col_offset:]
                        else:
                            # we inline the kwargs values directly into the function call
                            lines[kw.lineno - 1] = lines[kw.lineno - 1][:kw.col_offset] + ", ".join(f"{k}={repr(v)}" for k, v in kwargs_value.items()) + lines[kw.lineno - 1][kw.end_col_offset:]
                        print(lines[kw.lineno - 1], end="")

    with open(filename, "w") as f:
        f.writelines(lines)

            

if __name__ == "__main__":
    for filename in sys.argv[1:]:
        check_filter_file_calls(filename)
```
</details>


climbfuji (approved):
I mainly looked at eccodes. Thanks for cleaning this up.

adamjstewart:
Which type checkers take issue with this? I use ty with kwargs every day.

bernhardkaindl:
> Which type checkers take issue with this? I use ty with kwargs every day.

haampie stuck it at the end of PR description, Copilot just didn't "see" it it seems:

> Before: 1762 diagnostics, after: 1525 diagnostics; from ty check --extra-search-path ~/spack/lib/spack --extra-search-path repos --ignore unresolved-reference --output-format concise

He did it programmatically using the script he attached in the PR description.

adamjstewart:
Ah, I see it now. Looks like a bug in ty. Worth reporting to them.

bernhardkaindl:
> Ah, I see it now. Looks like a bug in ty. Worth reporting to them.

@adamjstewart, @haampie:

Agreed, the latest `ty` seems to have fixed or disabled this warning by default:

```py
ty check --extra-search-path ../spack --extra-search-path repos --ignore unresolved-reference --ignore unresolved-import --ignore too-many-positional
-arguments repos/spack_repo/builtin/packages/eccodes/package.py;ty --version
ty 0.0.54
error[invalid-argument-type]: Argument expression after ** must be a mapping type
   --> repos/spack_repo/builtin/packages/eccodes/package.py:155:65
    |
155 |             conflicts("extra_definitions={0}".format(center), **kwargs)
    |                                                                 ^^^^^^ Found `(dict[str, str] & ~AlwaysFalsy) | (list[dict[str, str]] & ~AlwaysFalsy)`
    |

error[invalid-argument-type]: Argument expression after ** must be a mapping type
   --> repos/spack_repo/builtin/packages/eccodes/package.py:161:19
    |
161 |                 **kwargs,
    |                   ^^^^^^ Found `str | dict[str, str]`
    |

Found 2 diagnostics
ty 0.0.54
```

It does not warn about this construct in `eccodes/package.py` by default (the warning above is on a different construct than the one patched in this PR in the same recipe).

```py
        kwargs = {"string": False, "backup": False, "ignore_absent": True}

        # Return the kind and not the size:
        filter_file(
            r"(^\s*kind_of_double\s*=\s*)(\d{1,2})(\s*$)",
            "\\1kind(real\\2)\\3",
            "fortran/grib_types.f90",
            **kwargs,
        )
        filter_file(
            r"(^\s*kind_of_\w+\s*=\s*)(\d{1,2})(\s*$)",
            "\\1kind(x\\2)\\3",
            "fortran/grib_types.f90",
            **kwargs,
        )
```

It looks like many changes made by haampie's script simplify the code, which looks good a first sight.

But in many other places where identical kwargs are reused, this PR duplicates `kwargs` arguments many times, which I think is undesirable.

Therefore, I think we may delay want to do this PR to more careful with so many changes until further review, convert this PR into draft state and stop the Gitlab CI running pipeline and trim it. I think those places where we're sure that the changes are
- correct,
- beneficial and
- make the code smaller and better
are good candidates to improve like this PR does.

PR #3424, opened 2026-02-14
magma: add -Xfatbin -compress-all flags when compiling with cuda

Markus92:
When compiling magma with a lot of cuda architectures (we got 6 in our heterogenous cluster!) the library can get over 2GB of code causing linker errors:

```
lib/libmagma.so: PC-relative offset overflow in PLT entry for `_Z24herk_template_batched_tnI6float2Li8ELi12ELi24ELi24ELi8ELi1ELi8ELi12ELi8ELi12ELi1ELi0EEv12magma_uplo_tiiPKPKT_iiiS6_iiiPPS2_iiiS2_S2_iP11magma_queue'
collect2: error: ld returned 1 exit status
make[2]: *** [CMakeFiles/magma.dir/build.make:23918: lib/libmagma.so] Error 1
make[2]: Leaving directory '/tmp/spackadmin/spack-stage/spack-stage-magma-2.9.0
```

The people over at Nix had the same [issue](https://github.com/NixOS/nixpkgs/pull/220402
) and looked at how PyTorch [solved it](https://github.com/pytorch/pytorch/blob/fe05266fda4f908130dea7cbac37e9264c0429a2/CMakeLists.txt#L548), as PyTorch ships with a dozen architectures for their conda builds.

This PR adds these flags for the CUDA builds and this fixed my linker issues.

### DIFF
--- a/repos/spack_repo/builtin/packages/magma/package.py
+++ b/repos/spack_repo/builtin/packages/magma/package.py
@@ -159,6 +159,7 @@ class Magma(CMakePackage, CudaPackage, ROCmPackage):
             options.append(define("GPU_TARGET", capabilities))
             archs = ";".join("%s" % i for i in cuda_arch)
             options.append(define("CMAKE_CUDA_ARCHITECTURES", archs))
+            options.append(define("CMAKE_CUDA_FLAGS", " -Xfatbin -compress-all"))
 
         if "@2.5.0" in spec:
             options.append(define("MAGMA_SPARSE", False))


tldahlgren (commented):
@G-Ragghianti @luszczek Any objections to the change in this PR?

G-Ragghianti:
I checked with the MAGMA developers.  This looks OK.

G-Ragghianti (approved):

luszczek:
This fix introduces another variable to the build. There is already `CUDA_NVCC_FLAGS` used earlier in the file and now we're adding `CMAKE_CUDA_FLAGS`. The new *fat binary* flags are specific to NVIDIA's `nvcc` so I would add them there. CMake supported non-`nvcc` builds on CUDA platforms for a while, so adding the new flags to `nvcc`-specific flags to a generic CUDA flags might cause problems. I'm thinking especially MAGMA's HIP builds that may target CUDA behind the scenes. But for NVIDIA-only builds the current fix works despite proliferating CMake and environment variables.

Markus92:
@luszczek  not sure why, but blindly replacing `CMAKE_CUDA_FLAGS` by `CMAKE_NVCC_FLAGS` does not result in a functioning build, seems the flag doesn't get propagated:

```
CMake Warning:
  Manually-specified variables were not used by the project:                                                                                                                                                                         
    BLAS_LIBRARIES
    CMAKE_NVCC_FLAGS
    CMAKE_POLICY_DEFAULT_CMP0090

```

This does almost make me wonder if the other `CMAKE_NVCC_FLAGS` is actually functioning at all, but I don't have a Cray system to test that on.

luszczek:
I doubt `CMAKE_NVCC_FLAGS` has a meaning anywhere including Cray. CMake does not mention this variable in its documentation. I would advise to remove it. My internet search indicates confusion about its use. Two Spack pakages that use this variable are MAGMA and heFFTe, bot of which are from overlapping development team. Other packages that use this variable don't mention Cray's compute environment.

tldahlgren (approved):

PR #3437, opened 2026-02-16
bats: add 1.13.0

afuetterer:
This PR adds version 1.13.0 of bats-core.

Ref: https://github.com/bats-core/bats-core/releases/tag/v1.13.0

Built tested locally, worked fine.

### DIFF
--- a/repos/spack_repo/builtin/packages/bats/package.py
+++ b/repos/spack_repo/builtin/packages/bats/package.py
@@ -11,10 +11,11 @@ class Bats(Package):
     """Bats is a TAP-compliant testing framework for Bash."""
 
     homepage = "https://github.com/bats-core/bats-core"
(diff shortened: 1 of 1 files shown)


alecbcs (approved):
Looks good to me. Thanks @afuetterer!

---

Super small note, but for the future you don't actually need to update the url property when adding a new version.

(The url property in the package class serves as a template that Spack uses to construct download URLs for all versions. Spack automatically substitutes the version number from the template URL, so it doesn't need to point to the latest version.)

afuetterer:
> Super small note, but for the future you don't actually need to update the url property when adding a new version.
> 
> (The url property in the package class serves as a template that Spack uses to construct download URLs for all versions. Spack automatically substitutes the version number from the template URL, so it doesn't need to point to the latest version.)

Actually, I was unsure about this. Some contributors updated the property as well, so I did the same. Thanks for letting me know, will stop doing that.

PR #3439, opened 2026-02-16
melissa & py-melissa: version 2.4.1, deprecating older versions

abhishek1297:
### Updates:
- Added `v2.4.1` (keeping as preferred)
- Removed `develop` as it will no longer support separate installation (`melissa` & `py-melissa-core`)
- Deprecated older versions
- Removed upper bound for `mpi4py` `@2.4.1`
- Removed `setup_build_environment` (sanity additions related to `PKG_CONFIG_PATH`)

### DIFF
--- a/repos/spack_repo/builtin/packages/melissa/package.py
+++ b/repos/spack_repo/builtin/packages/melissa/package.py
@@ -18,13 +18,18 @@ class Melissa(CMakePackage):
     # attention: Git**Hub**.com accounts
     maintainers("abhishek1297", "raffino")
 
-    version("develop", branch="develop", preferred=True)
-    version("2.3.0", sha256="f356e05082e4bb26a210cd11ccfa78a783ebe07be2bd75d5e51ed10da3b58997")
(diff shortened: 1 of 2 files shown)


repos/spack_repo/builtin/packages/melissa/package.py
```diff
@@ -43,10 +48,6 @@ class Melissa(CMakePackage):
         depends_on("libzmq@4.2:4")
         depends_on("mpi")
 
-    def setup_build_environment(self, env):
-        zmq_pkgconfig_path = self.spec["libzmq"].prefix.lib.pkgconfig
-        env.prepend_path("PKG_CONFIG_PATH", zmq_pkgconfig_path)
-
```
becker33, line 49: Should this really be removed, or should it be executed for the older versions still if someone builds a deprecated version?

abhishek1297: This was just a sanity addition. At some point building containers, I had an issue where PKG_CONFIG_PATH was not having the prefix, for whatever reason. But, spack already adds it. This is redundant and can be removed.

becker33 (approved):

PR #3446, opened 2026-02-17
emacs: improve platform conditionals

alecbcs:
Removed a dependency on `sys` by swapping to `self.spec.satisfies()` as is currently recommended in the docs.

### DIFF
--- a/repos/spack_repo/builtin/packages/emacs/package.py
+++ b/repos/spack_repo/builtin/packages/emacs/package.py
@@ -3,7 +3,6 @@
 # SPDX-License-Identifier: (Apache-2.0 OR MIT)
 
 import os
-import sys
 
 from spack_repo.builtin.build_systems.autotools import AutotoolsPackage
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/emacs/package.py
```diff
@@ -119,7 +118,7 @@ class Emacs(AutotoolsPackage, GNUMirrorPackage):
     patch("disable-posix-spawn-macos.patch", when="@28:30.2 platform=darwin os=sequoia")
     patch("disable-posix-spawn-macos.patch", when="@28:30.2 platform=darwin os=sonoma")
 
-    def configure_args(self):
+    def configure_args(self) -> list:
```
alalazo:
```suggestion
    def configure_args(self):
```

alalazo (commented):
I'd just remove the partial type hint in `package.py`

alalazo (approved):

PR #3447, opened 2026-02-17
hdf5: update configure args to use self.define()

alecbcs:
A couple quick fixes to update HDF5 to use `self.define()` when building configure args.


alalazo (approved):

alalazo:
Force merging, since pipelines are :heavy_check_mark: https://gitlab.spack.io/spack/spack-packages/-/pipelines/1433097 but Gitlab never updated status here

PR #3449, opened 2026-02-17
acts: add v45.1.0 and v45.1.1

stephenswat:
This commit adds two new versions of the ACTS package: v45.1.0 and v45.1.1.

### DIFF
--- a/repos/spack_repo/builtin/packages/acts/package.py
+++ b/repos/spack_repo/builtin/packages/acts/package.py
@@ -50,6 +50,8 @@ class Acts(CMakePackage, CudaPackage):
 
     # Supported Acts versions
     version("main", branch="main")
+    version("45.1.1", commit="da50efc7b15cad8fdc5e194719c72d7d8b706823")
+    version("45.1.0", commit="061a9d87b0fc07b554ec0b3849e875cf964f8323")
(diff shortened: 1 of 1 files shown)


wdconinc:
Do we need to change anything in how the ROOT dependencies are applied? Examples does not depend on ROOT anymore. Also, vdt is now a dependency when ROOT.

stephenswat:
> Do we need to change anything in how the ROOT dependencies are applied? Examples does not depend on ROOT anymore. Also, vdt is now a dependency when ROOT.

Looking at this, I will suggest a small rewrite of the way variants are handled in this package and submit a new PR; stay tuned.

wdconinc:
> > Do we need to change anything in how the ROOT dependencies are applied? Examples does not depend on ROOT anymore. Also, vdt is now a dependency when ROOT.
> 
> Looking at this, I will suggest a small rewrite of the way variants are handled in this package and submit a new PR; stay tuned.

New PR to replace this, or new PR after this is merged?

wdconinc (approved):

stephenswat:
> New PR to replace this, or new PR after this is merged?

See #3455; I redid the way dependencies between variants (especially the `examples` variant) are handled so that I don't need to now copy the `examples` variant declaration again; instead I could encode the removal of the `root` requirement by updating one of the `requires` statements.

PR #3453, opened 2026-02-17
llvm: fewer targets by default

haampie:
This is about 1000 fewer source files to build and a reasonable default IMO: x86/aarch64 for the CPU, amdgpu/nvptx for the GPU.

On the fence about webassembly / bpf, but I think it's fine like this.

### DIFF
--- a/repos/spack_repo/builtin/packages/llvm/package.py
+++ b/repos/spack_repo/builtin/packages/llvm/package.py
@@ -211,7 +211,7 @@ class Llvm(CMakePackage, CudaPackage, LlvmDetection, CompilerPackage):
     )
     variant(
         "targets",
-        default="all",
+        default="aarch64,amdgpu,nvptx,x86",
         description=(
             "What targets to build. Spack's target family is always added "
             "(e.g. X86 is automatically enabled when targeting znver2)."


rbberger (approved):
I usually set it to this subset in my production environments.

alecbcs:
@trws as the package maintainer do you have any thoughts here? If you're alright with this, I'm happy to merge.

trws:
I remember arguing against this change a long, long time ago but can no longer remember well why.  The only real downside I can see is if this is built on a non-arm, non-x86 system with the default value it will make a non-functional compiler for the host.

alecbcs:
No matter the targets given in the variant, the package always additionally adds the host architecture. I believe this was originally done for `targets=none` still building a valid installation on a users system, but it works nicely for this case too where we provide minimal cross compilation support with few architectures.

haampie:
Yeah, we automatically add the host arch:

https://github.com/spack/spack-packages/blob/73ba82410fb8fab46070310dd63b7ae79c48521c/repos/spack_repo/builtin/packages/llvm/package.py#L1288-L1302

trws:
Ah cool, I missed it when I scanned back over the package, LGTM then. :+1:

PR #3454, opened 2026-02-17
llvm: ~gold by default

haampie:
The gold linker is no longer maintained, it's replaced by lld and
considered deprecated.

### DIFF
--- a/repos/spack_repo/builtin/packages/llvm/package.py
+++ b/repos/spack_repo/builtin/packages/llvm/package.py
@@ -3,7 +3,6 @@
 # SPDX-License-Identifier: (Apache-2.0 OR MIT)
 import os
 import re
-import sys
 
 from spack_repo.builtin.build_systems.cmake import CMakePackage, generator
 from spack_repo.builtin.build_systems.compiler import CompilerPackage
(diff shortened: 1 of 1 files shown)


alecbcs:
Should we add a `when` condition here to disable gold by default for newer versions of LLVM but leave it on for older versions?

haampie:
It's not so much an llvm thing, it's that binutils dropped support https://lists.gnu.org/archive/html/info-gnu/2025-02/msg00001.html a year ago. It's for a niche use case of `clang -flto -fuse-ld=gold`. People should use `-fuse-ld=lld` also for older versions of llvm instead.

trws:
It's unfortunate, but it's probably the right answer.  The downside is there's no good way to use LTO on the LLVM build without it AFAIK.  Is there LTO support in bfd yet or anything else we can use?

haampie:
Is the use case building runtimes with LTO enabled using just-built clang? Or is it about libLLVM?

GCC's lto-plugin is supported by binutils `ld`, since GCC 4.6.0 according to https://github.com/gcc-mirror/gcc/commit/1cd0b7167ed9b43b189c39549d8165a020aee9b6

trws:
The runtimes are part of it, but it's really about building the clang/llvm/libllvm itself with LTO.  It sounds like any vaguely recent gcc doesn't have the problem though, so maybe we just do this in general or make it that `+gold` is enabled when compiling with gcc<4.6?  Once you have the new clang the new lld is good to go, but I never managed to find a way to build lld without already having built all of libllvm other than doing the full bootstrap build.

trws:
Fair enough, then I suppose I'm ok with this.

haampie:
> when compiling with gcc<4.6

Last release: April 12, 2013. I think it's fine to drop it then :p

rbberger (approved):

PR #3455, opened 2026-02-17
acts: improve variant dependency handling

stephenswat:
The ACTS package currently defines a lot of variants with `when` clauses based on other variants. This makes it difficult to add and remove requirements in a granular way as new versions get added: it requires the re-declaration of the variant multiple times. In order to streamline this process, I migrated the dependencies between variants to using `requires` statements instead.

### DIFF
--- a/repos/spack_repo/builtin/packages/acts/package.py
+++ b/repos/spack_repo/builtin/packages/acts/package.py
@@ -237,24 +237,12 @@ class Acts(CMakePackage, CudaPackage):
     }
     variant("cxxstd", default="17", when="@:35", **_cxxstd_common)
     variant("cxxstd", default="20", when="@36:", **_cxxstd_common)
-    variant(
-        "examples",
-        default=False,
-        description="Build the examples",
-        when="@0.23:16 +digitization +fatras +identification +json +root",
-    )
-    variant(
-        "examples",
-        default=False,
-        description="Build the examples",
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/acts/package.py
```diff
@@ -237,24 +237,12 @@ class Acts(CMakePackage, CudaPackage):
...
-        default=False,
-        description="Build the examples",
-        when="@35: +fatras +json +root",
-    )
+    variant("examples", default=False, description="Build the examples", when="@0.23:")
+    requires("+digitization", when="@:16 +examples")
+    requires("+identification", when="@:34 +examples")
+    requires("+root", when="+examples")
+    requires("+fatras", when="+examples")
+    requires("+json", when="+examples")
```
wdconinc: If you want to reduce duplication even more:
```suggestion
    with when("+examples"):
        requires("+digitization", when="@:16")
        requires("+identification", when="@:34")
        requires("+root")
        requires("+fatras")
        requires("+json")
```

wdconinc (dismissed):
This looks good! Thanks!

repos/spack_repo/builtin/packages/acts/package.py
stephenswat: Good one, thank you!

wdconinc (approved):

PR #3465, opened 2026-02-17
r-shinyfiletree: new package

qwertos:


wdconinc (approved):
Looks good. Not on cran.

PR #3470, opened 2026-02-18
openPMD-api: Fix pybind11 3.0.2

ax3l:
Hotfix: Python property lifetime fix & pybind11 3.0.2 support

### DIFF
--- a/repos/spack_repo/builtin/packages/openpmd_api/package.py
+++ b/repos/spack_repo/builtin/packages/openpmd_api/package.py
@@ -124,6 +124,13 @@ class OpenpmdApi(CMakePackage):
         when="@0.16.0",
     )
 
+    # Python property lifetime fix & pybind11 3.0.2 support
+    patch(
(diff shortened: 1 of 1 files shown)


haampie:
:wave: @ax3l, the new `find_package(pybind11 3.0.0 CONFIG REQUIRED)` bit means you have to add

```python
depends_on("py-pybind11@3:", type="link", when="@0.17")
```

too, right?

(And maybe `type="link"` could be `type="build"` for pybind11, I think it's header only and CMake should be able to find it with build type)

bernhardkaindl:
@ax3l I see you sent a 👍 for https://github.com/spack/spack-packages/pull/3470#issuecomment-3919203276

ax3l:
Oh I thought I fixed that already

ax3l:
PR #5300 fixes the pybind11 dep, we need both

bernhardkaindl:
@ax3l Thanks for the updated info! As you're the experts, assume that this is ready for merge right after merging #5300.

bernhardkaindl (approved):

PR #3471, opened 2026-02-18
binutils: gold was removed

haampie:


repos/spack_repo/builtin/packages/binutils/package.py
alecbcs:
```suggestion
    # The gold linker was removed in v2.44.
    variant(
        "gold", 
        default=False, 
        when="@:2.43 +ld", 
        description="Build the gold linker.",
    )
```

If we limit the variant by updating the `when=` condition here, it'll limit the availability of the variant same as the conflicts directive, but will also update the output of the package info to reflect the change.

repos/spack_repo/builtin/packages/binutils/package.py
```diff
@@ -127,6 +127,7 @@ class Binutils(AutotoolsPackage, GNUMirrorPackage):
             when="@2.37:2.40,2.42:",
             msg="https://github.com/spack/spack/issues/35817",
         )
+    conflicts("+gold", when="@2.44:", msg="Gold was removed")
```
alecbcs:
```suggestion
```

alecbcs (changes requested):
Quick nitpick to make the output of `spack info` more accurate. 
Thanks for catching that this was removed @haampie 😃

repos/spack_repo/builtin/packages/binutils/package.py
haampie: Conditional variants are annoying, `~gold` in people's config implies `@:2.43` then, so I'd rather not.

alecbcs: Since `+gold` had previously defaulted to False, I think it'll be fairly rare for someone to have `~gold` in their config. I agree conditional variants are annoying (and we could likely do some work to improve handling cases like this where someone has an existing configuration to disable a variant that is no longer available) but in general I think this is a better way to model a variant instead of keeping it around and making it unusable to a growing number of versions that will incorrectly report the option when running `spack info`.

alecbcs (approved):
Looks good to me. Thanks!

PR #3479, opened 2026-02-18
dyninst: patch missing cstdint includes

pearzt:
Building DynInst with GCC 15 currently fails as `cstdint` [is no longer included by default](https://gcc.gnu.org/gcc-15/porting_to.html#header-dep-changes). With this PR, I propose to add a patch to pulls in a commit from `master` that adds the missing include.


pearzt:
Ah, opening this PR was a bit premature, as there seems to be another build problem with GCC 15. I will update this when I have a fix.

pearzt:
@spackbot fix style

pearzt:
@hainest Friendly ping:)

hainest (approved):
Sorry, this slipped by me.

alalazo (approved):

PR #3481, opened 2026-02-18
abacus: utilize join_path()

alecbcs:
Switch to using `join_path()` instead of appending strings to follow newer packaging conventions.

### DIFF
--- a/repos/spack_repo/builtin/packages/abacus/package.py
+++ b/repos/spack_repo/builtin/packages/abacus/package.py
@@ -16,11 +16,11 @@ class Abacus(MakefilePackage):
     for large-scale electronic-structure simulations
     from first principles"""
 
-    maintainers("bitllion")
-
     homepage = "http://abacus.ustc.edu.cn/"
-    git = "https://github.com/abacusmodeling/abacus-develop.git"
     url = "https://github.com/abacusmodeling/abacus-develop/archive/refs/tags/v2.2.1.tar.gz"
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/abacus/package.py
```diff
@@ -91,11 +91,11 @@ def edit(self, spec, prefix):
                     pass
                 else:
                     lineList.append(line)
-        with open(self.build_directory + "/Makefile.system", "w") as f:
+        with open(join_path(self.build_directory, "Makefile.system"), "w") as f:
             for i in lineList:
                 f.write(i)
 
-        with open(self.build_directory + "/Makefile.system", "a") as f:
+        with open(join_path(self.build_directory, "Makefile.system"), "a") as f:
```
rbberger, line 98: You could just extract those 3 calls into a variable. Also, the last append can be merged into the previous with block after the for loop.

alecbcs: I agree with all of these notes, but I think they're out of scope for this PR. I'm not a maintainer of abacus and don't feel comfortable making changes to the install logic without being able to use and verify the resulting binary. Just trying to bring some of these packages up to date with modern packaging standards to help give folks a larger base of examples that are using best practices.

rbberger: Understood.

rbberger (approved):

PR #3482, opened 2026-02-18
abyss: use spec.satisfies() for direct MPI dependency

alecbcs:
Update Abyss to use `spec.satisfies()` to standardize on querying direct dependency.

### DIFF
--- a/repos/spack_repo/builtin/packages/abyss/package.py
+++ b/repos/spack_repo/builtin/packages/abyss/package.py
@@ -73,8 +73,11 @@ class Abyss(AutotoolsPackage):
             f"--with-sqlite={self.spec['sqlite'].prefix}",
             f"--with-mpi={self.spec['mpi'].prefix}",
         ]
+
         if maxk:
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/abyss/package.py
```diff
@@ -73,8 +73,11 @@ def configure_args(self):
             f"--with-sqlite={self.spec['sqlite'].prefix}",
             f"--with-mpi={self.spec['mpi'].prefix}",
         ]
+
         if maxk:
             args.append(f"--enable-maxk={maxk}")
-        if self.spec["mpi"].name == "mpich":
+
+        if self.spec.satisfies("%mpich"):
```
alalazo: Nitpick, but this checks the dependency is to satisfy a `mpi` request:
```suggestion
        if self.spec.satisfies("%mpi=mpich"):
```

alalazo (approved):

PR #3487, opened 2026-02-19
gsibec: add 1.4.2

mathomp4:
This adds gsibec v1.4.2 which has fixes for GNU compilation as well as tentative flags for `ifx` that match `ifort` flags.

ETA: It also adds a conflict for v1.4.1 and gcc@10 since that prompted v1.4.2

### DIFF
--- a/repos/spack_repo/builtin/packages/gsibec/package.py
+++ b/repos/spack_repo/builtin/packages/gsibec/package.py
@@ -22,6 +22,7 @@ class Gsibec(CMakePackage):
     license("Apache-2.0")
 
     version("develop", branch="develop")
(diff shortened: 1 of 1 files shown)


climbfuji (commented):
Since we know that 1.4.1 doesn't build with gcc@10+, should we add a conflict?

mathomp4:
> Since we know that 1.4.1 doesn't build with gcc@10+, should we add a conflict?

Oh. Yeah. Smart. Let me add that...

climbfuji (approved):

RikkiButler20 (approved):
LGTM

PR #3488, opened 2026-02-19
hdf-eos2: add flags for gcc@14

mathomp4:
In test builds of spack-stack with `gcc@14` (see https://github.com/JCSDA/spack-stack/pull/1908) it was found this package needed an extra flag to build. It needed the first two flags already here, but I needed an extra. Not sure why I'm different...


climbfuji (approved):

RikkiButler20 (approved):
Looks good to me

PR #3490, opened 2026-02-19
py-h5py: add flags for gcc@14

mathomp4:
In test builds of spack-stack with `gcc@14` (see https://github.com/JCSDA/spack-stack/pull/1908) it was found this package needed extra flags to build.


takluyver:
Thanks, make sense. I assume this is compiler warnings being upgraded to errors; we have made a bit of progress recently on cleaning up some warnings, but I don't know if all the relevant ones are gone.

climbfuji (approved):

mathomp4:
> Thanks, make sense. I assume this is compiler warnings being upgraded to errors; we have made a bit of progress recently on cleaning up some warnings, but I don't know if all the relevant ones are gone.

@takluyver Yes. I was building with `gcc@14` and it looks like things that used to be warnings are now errors. This just moves those back to warnings.

alalazo (approved):

PR #3491, opened 2026-02-19
resmap: new package

dacolombo:
## Review status

As per https://github.com/spack/spack-packages/pull/3491#discussion_r3069657429, this installs binaries which is not compatible with the package not being limited to a single target architecture. As the review says, the package should be installed from source if at all possible.

### DIFF
--- a/repos/spack_repo/builtin/packages/resmap/package.py
+++ b/repos/spack_repo/builtin/packages/resmap/package.py
@@ -5,8 +5,9 @@ from spack.package import *
 
 
 class Resmap(Package):
-    """ResMap (Resolution Map) is an easy to use software package for computing the local resolution of
-    3D density maps studied in structural biology, primarily electron cryo-microscopy (cryo-EM)."""
+    """ResMap (Resolution Map) is an easy to use software package for computing the
+    local resolution of 3D density maps studied in structural biology, primarily
+    electron cryo-microscopy (cryo-EM)."""
 
     homepage = "https://resmap.sourceforge.net/"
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/resmap/package.py
```diff
@@ -0,0 +1,24 @@
...
+from spack.package import *
+
+
+class Resmap(Package):
+    """ResMap (Resolution Map) is an easy to use software package for computing the
+    local resolution of 3D density maps studied in structural biology, primarily
+    electron cryo-microscopy (cryo-EM)."""
+
+    homepage = "https://resmap.sourceforge.net/"
+    url = "https://sourceforge.net/projects/resmap/files/ResMap-1.1.4-linux64/download"
```
wdconinc, line 13: Preference in spack (used on windows, linux and mac) is to build from source instead of downloading binaries, i.e. http://sourceforge.net/projects/resmap/files/src

bernhardkaindl:
@dacolombo, as per https://github.com/spack/spack-packages/pull/3491#discussion_r3069657429, this installs binaries which is not compatible with the package not being limited to a single target architecture. As the review says, the package should be installed from source if at all possible.

bernhardkaindl (changes requested):
See https://github.com/spack/spack-packages/pull/3491/changes#r3069657429 (Summary: no binaries please when source is possible)

dacolombo:
@bernhardkaindl thanks for pointing that out, I will try the compilation from source as soon as possible.

PR #3492, opened 2026-02-19
precice: package baseline cleanup

fsimonis:
This PR is a package cleanup of the preCICE package, as a row of edge cases for versions have accumulated in an unstructured way.
I know it's not good practice, but I also fixed some incorrect version ranges and added missing restrictions of the preCICE baseline versions. My goal is not to spam too many PRs here.

### DIFF
--- a/repos/spack_repo/builtin/packages/precice/package.py
+++ b/repos/spack_repo/builtin/packages/precice/package.py
@@ -202,15 +202,13 @@ class Precice(CMakePackage):
         else:
             cmake_args.append("-DPYTHON=OFF")
 
-        # The xSDK installation policies were implemented after 1.5.2.
-        # The TPL arguments were removed in 3.0.0.
+        # The xSDK installation policies were implemented after 1.5.2 and removed in 3.0.0
         if spec.satisfies("@1.6:3"):
             cmake_args.extend([
                 "-DTPL_ENABLE_BOOST:BOOL=ON",
                 "-DTPL_ENABLE_EIGEN3:BOOL=ON",
                 "-DTPL_ENABLE_LIBXML2:BOOL=ON",
                 self.define_from_variant("TPL_ENABLE_PETSC", "petsc"),
-                self.define_from_variant("TPL_ENABLE_PYTHON", "python")
             ])
 
         # Release options
--- a/repos/spack_repo/builtin/packages/precice/package.py
+++ b/repos/spack_repo/builtin/packages/precice/package.py
@@ -120,10 +120,17 @@ class Precice(CMakePackage):
     depends_on("py-numpy@1.17:", when="+python", type=("build", "run"))
     depends_on("py-numpy@1.21.5:", when="+python@3.2:", type=("build", "run"))
 
-    # We require C++14 compiler support
     conflicts("%gcc@:4")
-    conflicts("%apple-clang@:5")
+    conflicts("%gcc@:9.2.99", when="@1.4:")
+    conflicts("%gcc@:11.1.99", when="@2.4:")
+    conflicts("%gcc@:13.1.99", when="@3.2:")
+
     conflicts("%clang@:3.7")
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/precice/package.py
```diff
@@ -101,32 +112,51 @@ class Precice(CMakePackage):
...
     conflicts("%gcc@:4")
-    conflicts("%apple-clang@:5")
+    conflicts("%gcc@:9.2.99", when="@1.4:")
+    conflicts("%gcc@:11.1.99", when="@2.4:")
+    conflicts("%gcc@:13.1.99", when="@3.2:")
+
     conflicts("%clang@:3.7")
+    conflicts("%clang@:9.99", when="@1.4:")
+    conflicts("%clang@:13.99", when="@2.4:")
+    conflicts("%clang@:17.99", when="@3.2:")
```
fsimonis, line 152: These conflicts mark unsupported compiler versions.
Lower versions may work though.

Is this too strict?

fsimonis:
@spackbot fix style

repos/spack_repo/builtin/packages/precice/package.py
```diff
@@ -101,32 +112,51 @@ class Precice(CMakePackage):
...
+
+    conflicts("%apple-clang@:5")
     conflicts("%intel@:16")
     conflicts("%gcc@:9.2", when="@3.0.0:")
 
+    conflicts("-shared", when="@3:", msg="The shared library version is unsupported since 3.0.0")
```
MakisH: This is meant to restrict the **static** preCICE library, no?

MakisH (commented):
Looks good overall, I don't see anything strange here, besides one comment.

I am not sure if the one or the other structure is better, but you clearly have a good reason advocating for this change.

repos/spack_repo/builtin/packages/precice/package.py
fsimonis: `-shared` means that it conflicts with not being build as shared, yes

MakisH: But the message suggests that shared library builds are not supported since 3.0.0, which is contradicting. How about:

```suggestion
    conflicts("-shared", when="@3:", msg="Since v3.0.0, only shared library builds are supported.")
```

fsimonis: Ah. Yes, you are right. Thanks!

MakisH (approved):

fsimonis:
@spackbot fix style

repos/spack_repo/builtin/packages/precice/package.py
bernhardkaindl: @fsimonis, you should not need to append `.99` - Spac takes an incomplete version as also inclusive any `.*` version after e.g. `:9.2`.

So for gcc, these should mean that 
when using gcc,
- At least 9.3 or newer needed for `@1.4:` (any 9.2.x version or lower is no-go)
- At least 11.2 or newer needed for `@2.4`  ...
- At least 13.2 or newer needed for `@3.4`  ...

```suggestion
    conflicts("%gcc@:9.2", when="@1.4:")
    conflicts("%gcc@:11.1", when="@2.4:")
    conflicts("%gcc@:13.1", when="@3.2:")

    conflicts("%clang@:3.7")
    conflicts("%clang@:9", when="@1.4:")
    conflicts("%clang@:13", when="@2.4:")
    conflicts("%clang@:17", when="@3.2:")
```

bernhardkaindl:
@fsimonis, do you still plan to fix the remaining issues found in review?

As far as I can see, you didn't update this PR since March 2026.

bernhardkaindl (changes requested):

fsimonis:
@bernhardkaindl Hi, yes, I plan to get back to this soonish.
I admit that this PR should have been a draft.

PR #3493, opened 2026-02-19
amber: add v24

dacolombo:
Edit: Changes have been suggested by GitHub Copilot-based review.

### DIFF
--- a/repos/spack_repo/builtin/packages/amber/package.py
+++ b/repos/spack_repo/builtin/packages/amber/package.py
@@ -230,12 +230,12 @@ class Amber(Package, CudaPackage):
             filter_file(
                 f"cmake \$AMBER_PREFIX/amber{self.version}_src",
                 "cmake $AMBER_PREFIX/spack-src",
-                "build/run_cmake"
+                "build/run_cmake",
             )
             filter_file(
                 f"-DCMAKE_INSTALL_PREFIX=\$AMBER_PREFIX/amber{self.version}",
                 f"-DCMAKE_INSTALL_PREFIX={prefix}",
-                "build/run_cmake"
+                "build/run_cmake",
             )
 
             # CUDA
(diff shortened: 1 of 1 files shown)


hainest:
@hseara If/when you have a moment, could you take a look at this?

bernhardkaindl:
@dacolombo, could you review the added review comments?

bernhardkaindl (changes requested):
Review the review comments below https://github.com/spack/spack-packages/pull/3493#pullrequestreview-4585555626

bernhardkaindl:
@dacolombo, another ping for reviewing the review comments below https://github.com/spack/spack-packages/pull/3493#pullrequestreview-4585555626

repos/spack_repo/builtin/packages/amber/package.py
bernhardkaindl, line 106: @dacolombo, as said correctly in https://github.com/spack/spack-packages/pull/3493/changes#r3486272217, these variants doe not appear to be checked in the cmake-specific branch for v24, so they do nothing for v24.

If they are not supported for v24, these varians should be moved below a:

```suggestion
with default_args(when="@:20"):
   variant("openmp", description="Use OpenMP pragmas to parallelize", default=False)
   variant("x11", description="Build programs that require X11", default=False)
   variant("update", description="Update the sources prior compilation", default=False)
```
This would remove theose versions for v24, which might be not what you want, but do implement them, you'd have to implement theim in the v24 branch.

If v24 does autodetection (which I don't know) it might still autodetect OpenMP and X11 when development libraries are provided by the host. To support those in a clean on/off way with the CMake build system, you should check those variants in the CMake case or limit theim to the autoconf-based versions like there.

However, using the version to decide that an not the `build_system` is a bit of a kludge.

The correct way to a add CMake-built versions to a package would be to have a class inherited from CMakePackage to build the newer CMake build. That would cleanly separate the CMake build into another class in the same file (only the CMake-specific will be there). See other pacakge which do that. E.g. there are a number of pacakges which use MakefilePackage and CMakePackage and use version constrains to enforce constrains on when `build_system=cmake` is used.

bernhardkaindl (changes requested):
The review comment https://github.com/spack/spack-packages/pull/3493/changes#r3486272217 seems to be true, please review.

PR #3494, opened 2026-02-19
cups,cargo_c,ffmpeg,fltk,gettext,ghostscript,wxwidgets,qt: Fixes for libraries pulling in unintended external dependencies

Markus92:
I've started building some of my environments with `missing-libs: error` and this exposed a few issues. Our build server runs a full Gnome desktop environment (don't ask why) so there's a lot of cruft that can get pulled in accidentally.

Some packages make it extremely difficult to disable automatic detection (looking at you, `fltk` and `wxwidgets`) as they don't always honor the disable option. They might need some patches, but those are beyond the scope of this PR for now. Setting the default variants to True at least makes it so the right stuff gets pulled in. If your build environment doesn't include 1900 RPMs this will probably work a lot better.

### DIFF
--- a/repos/spack_repo/builtin/packages/cargo_c/package.py
+++ b/repos/spack_repo/builtin/packages/cargo_c/package.py
@@ -21,3 +21,4 @@ class CargoC(CargoPackage):
 
     depends_on("rust@1.89:", type="build", when="@0.10.17:")
     depends_on("rust@1.88:", type="build", when="@0.10.16:")
+    depends_on("openssl")
--- a/repos/spack_repo/builtin/packages/cups/package.py
+++ b/repos/spack_repo/builtin/packages/cups/package.py
@@ -29,7 +29,8 @@ class Cups(AutotoolsPackage):
     depends_on("pkgconfig", type="build")
 
     depends_on("gnutls")
+    depends_on("libxcrypt")
 
     def configure_args(self):
-        args = ["--enable-gnutls", "--with-components=core"]
+        args = ["--with-tls=gnutls", "--with-components=core"]
         return args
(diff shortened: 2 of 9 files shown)


haampie:
If you can't prevent an optional dependency from being used if detected, it's fine to make it unconditional dependency. Effectively unused variants give false hope :p. Not necessarily a blocker, but personally I would only add variants if they influence the build system.

Oh, and for autotools you can sometimes also use `ac_cv_*` internal variables to prevent auto-detection if there's no `--without-*` or `--disable-*` flag.

Markus92:
Did some more digging on this and found some good options. For `fltk`, it seems the package mostly dates to 2016! so I updated it to just use the `AutotoolsPackage` builder and that fixed pretty much all of it. WxWidgets behaves a bit better now, too.

Markus92:
@spackbot fix style

bernhardkaindl:
Hi @Markus92, I guess it would be good if you could open individual PRs for unrelated changes/packages.

Markus92:
@bernhardkaindl you can close this PR then. That's a TON of work for super minor changes.

bernhardkaindl:
My concern is mainly about reviewability rather than the size of the individual changes. For example, the wxwidgets change
```py
if self.spec.satisfies("+tiff"):
```
would actually always disable TIFF support because the variant is named libtiff instead: https://github.com/spack/spack-packages/pull/3494#discussion_r3566524584

This is exactly the kind of issue that's easier to spot when a PR focuses on a single package.

Before a PR is merged, ideally the maintainer of each affected package should have a reasonable opportunity to review the changes for their package. When a single PR touches nine unrelated packages, it becomes much more difficult for both package maintainers and the reviewers responsible for merging PRs to ensure that every change has received appropriate review.

Keeping unrelated changes in separate PRs reduces the chance that small package-specific issues slip through, makes review more efficient, and ultimately helps avoid regressions for users. That's the main reason I suggested splitting the changes, not because the changes themselves are large or complex.

bernhardkaindl (changes requested):
I'm marking this as Changes requested because I believe these unrelated package changes should be submitted as separate PRs.

I appreciate the work that went into these updates, but splitting them up will make review much easier for the individual package maintainers and reduce the chance of package-specific issues being overlooked.

PR #3495, opened 2026-02-20
fix: libseccomp+python build

cmelone:
libseccomp's python extension Makefile does a check for out of source builds, but fails when srcdir and builddir are different strings but point to the same underlying directory (ie `/dir` and `.`).

```
[ /var/tmp/melone1/spack-stage/spack-stage-libseccomp-2.6.0-w24sgwofjl4355bnza4sfvikvesxqawt/spack-src/src/python = . ] || cp /var/tmp/melone1/spack-stage/spack-stage-libseccomp-2.6.0-w24sgwofjl4355bnza4sfvikvesxqawt/spack-src/src/python/seccomp.pyx .
cp: '/var/tmp/melone1/spack-stage/spack-stage-libseccomp-2.6.0-w24sgwofjl4355bnza4sfvikvesxqawt/spack-src/src/python/seccomp.pyx' and './seccomp.pyx' are the same file
```

This adds a patch that checks if the file to be copied already exists in the builddir before attempting to copy it over.

### DIFF
--- a/repos/spack_repo/builtin/packages/libseccomp/package.py
+++ b/repos/spack_repo/builtin/packages/libseccomp/package.py
@@ -35,6 +35,7 @@ class Libseccomp(AutotoolsPackage, PythonExtension):
         # https://github.com/seccomp/libseccomp/commit/afbde6ddaec7c58c3b281d43b0b287269ffca9bd
         depends_on("python@:3.11", type=("run", "link", "build"), when="@:2.5")
         depends_on("py-setuptools", type="build", when="@2.6:")
+        # upstream PR: https://github.com/seccomp/libseccomp/pull/482
         patch("fix-pyx-copy.patch", when="@2.6")
 
     def configure_args(self):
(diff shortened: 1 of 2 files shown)


repos/spack_repo/builtin/packages/libseccomp/package.py
```diff
@@ -35,6 +35,7 @@ class Libseccomp(AutotoolsPackage, PythonExtension):
         # https://github.com/seccomp/libseccomp/commit/afbde6ddaec7c58c3b281d43b0b287269ffca9bd
         depends_on("python@:3.11", type=("run", "link", "build"), when="@:2.5")
         depends_on("py-setuptools", type="build", when="@2.6:")
+        patch("fix-pyx-copy.patch", when="@2.6")
```
kwryankrattiger, line 39: Is there an upstream PR for this?

cmelone: good point, forgot to do that

becker33 (dismissed):

repos/spack_repo/builtin/packages/libseccomp/package.py
kwryankrattiger: If you can comment a link to the upstream PR for reference when it is up, then I will approve

kwryankrattiger (changes requested):
Link to upstream PR

kwryankrattiger (approved):

PR #3498, opened 2026-02-20
metkit: add dependency for C compiler

sibcse:
Metkit needs a C compiler, otherwise the build fails with ```[spack cc]: Error: SPACK_CC_* variables not set: this usually means a missing `depends_on("c", type="build")` in package.py.```


alalazo (approved):
Confirmed:

https://github.com/ecmwf/metkit/blob/7616a8a8ae53669b5a0870f7d0dad2f7c4b92814/CMakeLists.txt#L15

PR #3508, opened 2026-02-23
hwloc: add v2.13.0

rbberger:

### DIFF
--- a/repos/spack_repo/builtin/packages/hwloc/package.py
+++ b/repos/spack_repo/builtin/packages/hwloc/package.py
@@ -39,6 +39,7 @@ class Hwloc(AutotoolsPackage, CudaPackage, ROCmPackage):
     executables = ["^hwloc-bind$"]
 
     version("master", branch="master")
+    version("2.13.0", sha256="1514a5253f0a5c23bc006d3bdd30a6f6125c9a8dc9b5fa4984913d1fff45315d")
     version("2.12.2", sha256="ff7d309fdff7ceddfe15c1e79eaff25f3126a134f29f44d4e85571f187a6bab8")
     version("2.11.1", sha256="9f320925cfd0daeaf3a3d724c93e127ecac63750c623654dca0298504aac4c2c")
     version("2.10.0", sha256="c7fd8a1404a9719c76aadc642864b9f77aed1dc1fc8882d6af861a9260ba240d")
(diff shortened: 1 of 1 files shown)


rbberger:
@bgoglin looks like `_wrefresh` is defined in `libncurses`, not `libtinfo`. So it appears to me it's missing from the link line.

bgoglin:
> @bgoglin looks like `_wrefresh` is defined in `libncurses`, not `libtinfo`. So it appears to me it's missing from the link line.

@sthibaul This is related to f4ceffd23be46c8585010742aaf1dec553e53bdf ?

sthibaul:
It's probably related, yes. but we do test linking curses before reverting to tinfo/termcap, so @rbberger  I wonder in which case you are seeing an issue (OS ? toolchain ?).
Possibly a solution is to check in hwloc_internal.m4, after deciding which library to be used, for the presence of refresh(), and call it only when it is indeed available.

rbberger:
@sthibaul See the CI failures, like https://gitlab.spack.io/spack/spack-packages/-/jobs/20696760.
```
checking termcap support using ncurses and ... no
checking termcap support using ncurses and -lncursesw... no
checking termcap support using ncurses and -lncurses... no
checking termcap support using ncurses and -ltinfo... yes
```
and
```
/usr/bin/ld: lstopo_no_graphics-lstopo-ascii.o: in function `output
            _ascii':
  >> 659    /tmp/root/spack-stage/spack-stage-hwloc-2.13.0-5jigahsf3hy6dgzsbgd2
            wjukdw3g57o4/spack-src/utils/lstopo/lstopo-ascii.c:610: undefined r
            eference to `wrefresh'
  >> 660    collect2: error: ld returned 1 exit status
```
With `^ncurses+termlib`, the link line only adds `-lm -ltinfo` instead of `-lm -lncurses -ltinfo`.

sthibaul:
> With ^ncurses+termlib

Ok, that's an odd case, but why not.

Checking again, it seems we don't actually need the refresh() call because in the end we only call tinfo-only functions, which do not cache anything, so the fflush is all we need to get the color definitions out before feeding the pager.

I have pushed removing the refresh call.

rbberger:
@spackbot run pipeline

haampie (approved):

PR #3509, opened 2026-02-23
`[needs-conflicts-with-ROCm]` mesa: add newer versions up to v26.0

rbberger:
This expands on #3499, but also adds new versions and works through the various issues due to deprecations/removals. Most notably, osmesa was removed completely in 26.0. So this also removes `osmesa` variant altogether.

### DIFF
--- a/repos/spack_repo/builtin/packages/mesa/package.py
+++ b/repos/spack_repo/builtin/packages/mesa/package.py
@@ -22,11 +22,12 @@ class Mesa(MesonPackage):
     license("MIT AND SGI-B-2.0 AND BSL-1.0")
 
     version("main", branch="main")
-    version(
-        "25.0.5",
-        sha256="c0d245dea0aa4b49f74b3d474b16542e4a8799791cd33d676c69f650ad4378d0",
-        preferred=True,
-    )
+    version("26.0.0", sha256="2a44e98e64d5c36cec64633de2d0ec7eff64703ee25b35364ba8fcaa84f33f72")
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/mesa/package.py
```diff
@@ -138,7 +141,7 @@ class Mesa(MesonPackage):
 
     # Require at least 1 front-end
     # TODO: Add egl to this conflict once made available
-    conflicts("~osmesa ~glx")
+    conflicts("@:25.0 ~osmesa ~glx")
```
wdconinc, line 144: Does this mean that glx is now required?

rbberger: No. I think this needs to now account for llvmpipe/softpipe being a frontend. Since we're always enabling at least softpipe we should always have at least one, making the conflict for enforcing at least one frontend pointless, unless we add dedicated variants.

v-dobrev:
Looking at the CI failures, I see that Mesa 26.0.0 has some incompatibility issues with ROCm LLVM, e.g.:
* when compiling `src/gallium/auxiliary/gallivm/lp_bld_debug.cpp`: it is using `LLVMDIBuilderCreatePointerType` which has a different signature in the ROCm version of LLVM (starting with ROCm v5.7.0) compared to the main LLVM.
* when compiling `src/gallium/auxiliary/gallivm/lp_bld_nir_soa.c`: similar issue with `LLVMDIBuilderCreateAutoVariable` -- the signature of this function in ROCm was changed in ROCm v5.7.0 too.

PR #3515, opened 2026-02-23
Add dependencies to CUDA package

RMeli:
This PR adds some assumed dependencies to the CUDA package and a pin to `libxml2` so that it can be built on Ubuntu 25.10.


haampie (approved):

haampie:
Looks like the CI issues are spurious

sethrj:
@RMeli  The correct way to pin versions for operating systems is within your build environment, not within the package recipe. It looks like this change predated #4562 , and at this point was listed incorrectly as a runtime dependency (see #4911), so next time you should need only run `spack external find --not-buildable libxml` before installing CUDA, and the installed version shouldn't depend on it.

PR #3518, opened 2026-02-23
py-bibtexparser: new package

qwertos:


manuelakuhn (approved):
LGTM and build fine on rocklinux 8 with gcc@15

PR #3525, opened 2026-02-23
lftp: add v4.9.3

drkrynstrng:
Adding LFTP v4.9.3.


RikkiButler20 (approved):
Verified checksum. LGTM

PR #3526, opened 2026-02-23
Bump repo package api version to 2.4

kwryankrattiger:
Since there is no mechanism to restrict usage of features from a newer API version it is not possible to enforce older API versions on develop.

### DIFF
--- a/repos/spack_repo/builtin/repo.yaml
+++ b/repos/spack_repo/builtin/repo.yaml
@@ -1,3 +1,3 @@
 repo:
   namespace: builtin
-  api: v2.2
+  api: v2.4


haampie:
Are you aware of anything that's currently being used that requires the version bump? Would be nice to delay the version bump to maximize compatibility

kwryankrattiger:
> Are you aware of anything that's currently being used that requires the version bump? Would be nice to delay the version bump to maximize compatibility

I haven't gone through with a fine tooth comb. It is mostly because we make this claim and have no testing to guarantee it is true. I agree with your sentiment but it makes me nervous to make this claim without something stronger in place to enforce it. It is completely untested.

The tradeoff of doing this is saying that the new repo needs at least the Spack package API the repo was tested against when the packages repo was released. I don't think that is overly restrictive, that packages repo _should_ continue to be usable by newer versions of Spack. We are still guaranteeing backwards compatibility to users. Forward compat seems like a happy accident at best.

haampie:
Alternatively we could run `mypy`/`ty` with Spack v1.0.

Or better since we cannot error on type issues right now: error on non-empty diff between `ty` against Spack v1.0 vs Spack v1.2

haampie:
Hm, diffing isn't entirely feasible cause of unrelated changes upstream:

https://gist.github.com/haampie/79b31d829143ec0a843d0c35eed2c923

kwryankrattiger:
> Alternatively we could run mypy/ty with Spack v1.0.

I am fine with that, I was thinking a similar sort of thing along with a `spack audit packages` using previous versions as well. But that is really only a solution going forward, not so much for what is there now. As it stands we don't have `mypy` or `ty` checks for packages and the lift to deploy those is going to require a fair amount of updates in packages from what I have heard from those who have tried.

So, my proposal here is to bump the repo package API version to match the API version of the release they were tested against, and then add better checking for the repo against older versions of the API to ensure we are adhering to what is in the config going forward.

If that means we can drop the API version back to 2.2 then that is great. It is _probably_ safe to say if the head of develop with these new checks satisfies the package API 2.2, then all of the previous states also satisfy the package API 2.2. But I don't think we should be asserting things we haven't verify it explicitly.

haampie:
Actually the diff from `ty` is rather small after I added a typehint for `PackageBase.spec` on spack develop. I've updated the gist above.

```
-Found 3617 diagnostics
+Found 3743 diagnostics
```

The remaining diff is mostly the same issue about the type of variant values.

I think it should be feasible to handle this with static analysis.

kwryankrattiger:
> I think it should be feasible to handle this with static analysis.

Yeah that is what we need. Again my only qualm is that we haven't had that and we are making changes (minimal I know) to `package_base.py` in order to be able to run the analysis with fewer errors which means it is not really feasible to verify as it stands.

We could backport that fix, and any others needed, to a 1.0.5.  But considering we just released a 1.0.4 I don't  know how soon that would happen. I am fine to hold off on this if we have the bandwidth to add and backport these changes, but as we go back through and release patch releases I think it would be better to reflect the tested state of things.

haampie:
That fix only applied to develop

haampie:
From the `ty` diff I don't see a good reason to merge this. Maybe you can target the release branch?

kwryankrattiger:
Closing after offline conversation

PR #3529, opened 2026-02-24
llvm-amdgpu: remove fortran build dependency

rbberger:
Correct me if I'm wrong, but I don't see a Fortran dependency in LLVM either. So this seems like a misdetection.


srekolam:
there is amdflang which is generated inside the /bin folder

rbberger:
@srekolam yeah, but it's not the question whether it produces a Fortran compiler, but whether it needs a **Fortran compiler dependency** to build at all. I recently saw a build of `llvm_amdgpu %c,cxx,fortran=gcc` which struck me as odd, knowing how vanilla LLVM builds.

haampie (approved):
Yeah, `depends_on("fortran", ...)` was auto-generated based on the source files. Fortran files were likely part of the test suite.

afzpatel (approved):

PR #3530, opened 2026-02-24
alps: add v2.3.4, modify maintainers & license

Ooolab:
In this updated spack package, we added support for newer versions of Boost and macOS26.

### DIFF
--- a/repos/spack_repo/builtin/packages/alps/package.py
+++ b/repos/spack_repo/builtin/packages/alps/package.py
@@ -5,7 +5,7 @@
 from spack_repo.builtin.build_systems.cmake import CMakePackage
 
 from spack.package import *
-
+import os
 
 class Alps(CMakePackage):
     """
@@ -15,15 +15,14 @@ class Alps(CMakePackage):
     """
 
     homepage = "https://github.com/ALPSim/ALPS"
-    url = "https://github.com/ALPSim/ALPS/archive/refs/tags/v2.3.3-beta.5.tar.gz"
+    url = "https://github.com/ALPSim/ALPS/archive/refs/tags/v2.3.4.tar.gz"
 
-    maintainers("Sinan81")
+    maintainers("Ooolab","egull")
 
-    license("BSL-1.0", checked_by="Sinan81")
+    license("MIT", checked_by="Ooolab")
 
-    version(
-        "2.3.3-beta.6", sha256="eb0c8115b034dd7a9dd585d277c4f86904ba374cdbdd130545aca1c432583b68"
-    )
+    version("2.3.4", sha256="bc65453bb88cc42de77c2963a44966fb3c92e80ff8e8e29e80fb87694fba41d7")
+    version("2.3.3-beta.6", sha256="eb0c8115b034dd7a9dd585d277c4f86904ba374cdbdd130545aca1c432583b68")
 
     depends_on("c", type="build")
     depends_on("cxx", type="build")
@@ -31,10 +30,7 @@ class Alps(CMakePackage):
 
     # update version constraint on every boost release after providing version & checksum info
     # in resources dictionary below
-    depends_on(
-        "boost@:1.87 +chrono +date_time +filesystem +iostreams +mpi +numpy"
-        "+program_options +python +regex +serialization +system +test +thread +timer"
-    )
+    depends_on("boost@1.80:")  # Just for headers. Note that the checksums are listed below
     depends_on("fftw")
     depends_on("hdf5 ~mpi+hl")
     depends_on("lapack")
@@ -42,18 +38,21 @@ class Alps(CMakePackage):
     depends_on("py-numpy", type=("build", "run"))
     depends_on("py-scipy", type=("build", "run"))
     depends_on("py-matplotlib", type=("build", "run"))
-
+    depends_on("mpi")
+    
     extends("python")
 
     # https://github.com/ALPSim/ALPS/issues/9
-    conflicts(
-        "%gcc@14", when="@:2.3.3-beta.6", msg="use gcc older than version 14 or else build fails"
-    )
+    #conflicts(
+    #    "%gcc@14", when="@:2.3.3-beta.6", msg="use gcc older than version 14 or else build fails"
+    #)
 
     # See https://github.com/ALPSim/ALPS/issues/6#issuecomment-2604912169
     # for why this is needed
     for boost_version, boost_checksum in (
         # boost version, shasum
+        ("1.89.0", "85a33fa22621b4f314f8e85e1a5e2a9363d22e4f4992925d4bb3bc631b5a0c7a"),
+        ("1.88.0", "46d9d2c06637b219270877c9e16155cbd015b6dc84349af064c088e9b5b12f7b"),
         ("1.87.0", "af57be25cb4c4f4b413ed692fe378affb4352ea50fbe294a11ef548f4d527d89"),
         ("1.86.0", "1bed88e40401b2cb7a1f76d4bab499e352fa4d0c5f31c0dbae64e24d34d7513b"),
         ("1.85.0", "7009fe1faa1697476bdc7027703a2badb84e849b7b0baad5086b087b971f8617"),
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/alps/package.py
```diff
@@ -15,45 +17,44 @@ class Alps(CMakePackage):
     """
 
     homepage = "https://github.com/ALPSim/ALPS"
-    url = "https://github.com/ALPSim/ALPS/archive/refs/tags/v2.3.3-beta.5.tar.gz"
 
-    maintainers("Sinan81")
+    maintainers("Ooolab", "egull")
```
Sinan81: This usually means maintainer for the spack package file, as opposed to the maintainer for the software itself. So, I would recommend keeping me as maintainer alongside the others listed here. But it's up to you.

repos/spack_repo/builtin/packages/alps/package.py
```diff
@@ -15,45 +17,44 @@ class Alps(CMakePackage):
...
 
-    license("BSL-1.0", checked_by="Sinan81")
+    license("MIT", checked_by="Ooolab")
 
     version(
-        "2.3.3-beta.6", sha256="eb0c8115b034dd7a9dd585d277c4f86904ba374cdbdd130545aca1c432583b68"
+        "2.3.4-beta.2",
+        sha256="ca2e1307630e6fccac279ab7711036f7c6dee43c386fd6f24cfc77c86a3c7f1c",
+        url="https://github.com/ALPSim/ALPS/archive/refs/tags/v2.3.4-beta.2.tar.gz",
```
Sinan81: as far as I can tell, there's no need for a different url for every version here. Spack automatically converts the deleted `url = ...` statement to the correct one for a given version. When the url pattern changes between versions (which's not the case here I think), in Spack we usually use `url_for_version` method to handle different patterns.

repos/spack_repo/builtin/packages/alps/package.py
```diff
@@ -15,45 +17,44 @@ class Alps(CMakePackage):
...
-    )
+    depends_on("boost@1.80:")  # Just for headers. Note that the checksums are listed below
     depends_on("fftw")
     depends_on("hdf5 ~mpi+hl")
     depends_on("lapack")
     depends_on("python", type=("build", "link", "run"))
     depends_on("py-numpy", type=("build", "run"))
     depends_on("py-scipy", type=("build", "run"))
     depends_on("py-matplotlib", type=("build", "run"))
+    depends_on("mpi")
```
Sinan81: ALPS can be built without MPI right? if so, you might want to add a variant like:
```
variant("mpi", default=False,  description="Build with MPI support")
```
afterwards, do like:
```
```

repos/spack_repo/builtin/packages/alps/package.py
```diff
@@ -15,45 +17,44 @@ class Alps(CMakePackage):
...
-
-    # update version constraint on every boost release after providing version & checksum info
-    # in resources dictionary below
-    depends_on(
-        "boost@:1.87 +chrono +date_time +filesystem +iostreams +mpi +numpy"
-        "+program_options +python +regex +serialization +system +test +thread +timer"
-    )
+    depends_on("boost@1.80:")  # Just for headers. Note that the checksums are listed below
     depends_on("fftw")
     depends_on("hdf5 ~mpi+hl")
```
Sinan81: if building with mpi and and also want to use hdf5 built with mpi, you could add the following
```
depends_on("hdf5+mpi+hl", when="+mpi")
```

repos/spack_repo/builtin/packages/alps/package.py
```diff
@@ -74,27 +75,126 @@ class Alps(CMakePackage):
...
+
+        # Tell CMake to compile boost from source
+        args.append("-DBoost_USE_STATIC_LIBS=ON")
+
+        args.append("-DBoost_USE_STATIC_RUNTIME=OFF")
+
+        # Add MPI support - just specify compilers (minimal change)
+        args.append("-DMPI_CXX_COMPILER={0}".format(self.spec["mpi"].mpicxx))
+
+        args.append("-DMPI_C_COMPILER={0}".format(self.spec["mpi"].mpicc))
```
Sinan81: if defining an mpi variant mentioned above, then these should be added only if "+mpi" variant is selected.

repos/spack_repo/builtin/packages/alps/package.py
```diff
@@ -74,27 +75,126 @@ class Alps(CMakePackage):
...
+        env.set("BOOST_ROOT", boost_src_dir)
+        env.set("Boost_SRC_DIR", boost_src_dir)
+
+        # Include paths for compilation
+        env.append_path("CPLUS_INCLUDE_PATH", self.spec["python"].headers.directories[0])
+
+        # For MPI - set compiler wrappers
+        env.set("MPI_CXX", self.spec["mpi"].mpicxx)
+        env.set("MPI_CC", self.spec["mpi"].mpicc)
+        env.set("MPICXX", self.spec["mpi"].mpicxx)
```
Sinan81: same as above.

Sinan81:
Happy to see that the developers of ALPS are now maintaining the spack package file. Also, it's nice that boost dependency is simplified.

repos/spack_repo/builtin/packages/alps/package.py
Ooolab: Thank you for a fast review! I have updated the package according to your suggestions. Thank you for your continued support for ALPS!

repos/spack_repo/builtin/packages/alps/package.py
Ooolab: It has been updated.

repos/spack_repo/builtin/packages/alps/package.py
Ooolab: added conditional MPI support through the variant "mpi"

Sinan81 (approved):
Looks good. approving.

Ooolab:
There is a pipeline error: Status: Failed
ml-linux-aarch64-cpu-build

Is this preventing the merging? No sure what caused this. Any suggestions? Thanks!

Sinan81:
> There is a pipeline error: Status: Failed ml-linux-aarch64-cpu-build
> 
> Is this preventing the merging? No sure what caused this. Any suggestions? Thanks!

it must be an unrelated bug. Should not prevent the merge.

Sinan81 (approved):

Sinan81:
I don't have permissions to merge. So, we're waiting for one of core spack developers to merge this.

repos/spack_repo/builtin/packages/alps/package.py
```diff
@@ -15,45 +17,42 @@ class Alps(CMakePackage):
...
     depends_on("cxx", type="build")
     depends_on("fortran", type="build")
-
-    # update version constraint on every boost release after providing version & checksum info
-    # in resources dictionary below
-    depends_on(
-        "boost@:1.87 +chrono +date_time +filesystem +iostreams +mpi +numpy"
-        "+program_options +python +regex +serialization +system +test +thread +timer"
-    )
+    depends_on("boost@1.80:")  # Just for headers. Note that the checksums are listed below
```
Sinan81: if boost is only needed for headers, then probably we should make this a build only dependency like
```
depends_on("boost@1.80:", type="build")
```
default dependency is build & link.
This can be improved later.

Sinan81:
pinging @alalazo for help with merge since he reviewed the previous alps MR.

Sinan81 (approved):

repos/spack_repo/builtin/packages/alps/package.py
Ooolab: Will take note of this and improve it in the next version.
Thanks!

repos/spack_repo/builtin/packages/alps/package.py
```diff
@@ -15,45 +17,42 @@ class Alps(CMakePackage):
     """
 
     homepage = "https://github.com/ALPSim/ALPS"
-    url = "https://github.com/ALPSim/ALPS/archive/refs/tags/v2.3.3-beta.5.tar.gz"
+    url = "https://github.com/ALPSim/ALPS/archive/refs/tags/v2.3.4-beta.2.tar.gz"
 
-    maintainers("Sinan81")
+    maintainers("Ooolab", "egull", "Sinan81")
```
alalazo, line 22: @egull Are you fine with this?

repos/spack_repo/builtin/packages/alps/package.py
```diff
@@ -15,45 +17,42 @@ class Alps(CMakePackage):
...
 
-    license("BSL-1.0", checked_by="Sinan81")
+    license("MIT", checked_by="Ooolab")
```
alalazo: Why did this change? I guess it shouldn't for older versions?

repos/spack_repo/builtin/packages/alps/package.py
```diff
@@ -15,45 +17,42 @@ class Alps(CMakePackage):
...
-    depends_on("hdf5 ~mpi+hl")
     depends_on("lapack")
     depends_on("python", type=("build", "link", "run"))
     depends_on("py-numpy", type=("build", "run"))
     depends_on("py-scipy", type=("build", "run"))
     depends_on("py-matplotlib", type=("build", "run"))
+    depends_on("mpi", when="+mpi")
+    depends_on("hdf5+mpi+hl", when="+mpi")
+    depends_on("hdf5~mpi+hl", when="~mpi")
+    depends_on("zlib")
```
alalazo:
```suggestion
    depends_on("zlib-api")
```

repos/spack_repo/builtin/packages/alps/package.py
```diff
@@ -74,27 +73,129 @@ class Alps(CMakePackage):
             placement="boost_source_files",
         )
 
+    # Patch for >=Boost 1.88.0 compatibility
+    def patch(self):
+        # Only apply patch for Boost versions greater than 1.87
+        # Check if boost dependency is specified and get its version
+        if "boost" in self.spec:
```
alalazo: Invert the logic to avoid excessive nesting. For instance:

```suggestion
        if "boost" not in self.spec:
            return
```

repos/spack_repo/builtin/packages/alps/package.py
```diff
@@ -74,27 +73,129 @@ class Alps(CMakePackage):
...
+                    "src/alps/numeric/matrix/matrix_element_iterator.hpp",
+                )
+
     def cmake_args(self):
         args = []
-        # Boost_ROOT_DIR option is replaced by Boost_SRC_DIR as of 2.3.3-beta.6
+        # This will ensure availability of the cstddef header on darwin (mac)
+        cstdlibstr = ""
+        if self.spec.satisfies("platform=darwin"):
+            cstdlibstr = " -stdlib=libc++"
```
alalazo: Hmm, what about stuff using `%gcc`? Would it work?

Sinan81 (approved):

Ooolab:
We would appreciate your review @alalazo of ALPS. Thanks!

repos/spack_repo/builtin/packages/alps/package.py
```diff
@@ -74,27 +73,129 @@ class Alps(CMakePackage):
...
+                + cstdlibstr
             )
         )
-        args.append(
-            "-DBoost_SRC_DIR={0}".format(join_path(self.stage.source_path, "boost_source_files"))
-        )
+
+        # Point to boost source files that need to be compiled
+        boost_src_dir = os.path.join(self.stage.source_path, "boost_source_files")
+        args.append("-DBoost_SRC_DIR={0}".format(boost_src_dir))
```
alalazo: Here and below you can use the more idiomatic `self.define` (you'll find plenty of examples in other packages)

repos/spack_repo/builtin/packages/alps/package.py
```diff
@@ -74,27 +73,129 @@ class Alps(CMakePackage):
...
+
+        # Tell CMake to compile boost from source
+        args.append("-DBoost_USE_STATIC_LIBS=ON")
+
+        args.append("-DBoost_USE_STATIC_RUNTIME=OFF")
+
+        # Add MPI support - just specify compilers (minimal change)
+        if "+mpi" in self.spec:
```
alalazo:
```suggestion
        if self.spec.satisfies("+mpi"):
```

repos/spack_repo/builtin/packages/alps/package.py
```diff
@@ -74,27 +73,129 @@ class Alps(CMakePackage):
...
+
+        # Tell CMake to use Spack’s HDF5 explicitly
+        args.append("-DHDF5_DIR={0}".format(self.spec["hdf5"].prefix))
+
+        # Add Python support
+        args.append("-DPYTHON_EXECUTABLE={0}".format(self.spec["python"].command.path))
+
+        args.append("-DPYTHON_INCLUDE_DIR={0}".format(self.spec["python"].headers.directories[0]))
+
+        args.append("-DPYTHON_LIBRARY={0}".format(self.spec["python"].libs[0]))
```
alalazo: Why do you need this? These variables should be added by `python` itself, see:

Image

repos/spack_repo/builtin/packages/alps/package.py
alalazo: Also, specifically for Boost, is it a vendored library? Can we configure `alps` so it reuses a Spack compiled version of Boost?

repos/spack_repo/builtin/packages/alps/package.py
```diff
@@ -74,27 +73,129 @@ class Alps(CMakePackage):
...
-    @run_after("install")
-    def relocate_python_stuff(self):
-        pyalps_dir = join_path(python_platlib, "pyalps")
-        with working_dir(self.prefix):
-            copy_tree("pyalps", pyalps_dir)
-        with working_dir(self.prefix.lib):
-            copy_tree("pyalps", pyalps_dir)
-            # in pip installed pyalps package, xml dir is provided under platlib/pyalps
-            copy_tree("xml", join_path(pyalps_dir, "xml"))
+    def setup_build_environment(self, env):
```
alalazo, line 175: Why are all these needed, and not managed by Cmake?

alalazo (changes requested):

repos/spack_repo/builtin/packages/alps/package.py
Ooolab: This has been included in the updated package.

repos/spack_repo/builtin/packages/alps/package.py
Ooolab: We changed to MIT license so that it is maximally non-restrictive.

repos/spack_repo/builtin/packages/alps/package.py
Ooolab: updated to include this change. Thanks.

repos/spack_repo/builtin/packages/alps/package.py
Ooolab: updated to include this change. thanks.

repos/spack_repo/builtin/packages/alps/package.py
Ooolab: Yes. They worked. This header problem is only specific to Darwin in Mac.

repos/spack_repo/builtin/packages/alps/package.py
Ooolab: updated to include the change. thanks for the suggestions, which make the package more concise.

repos/spack_repo/builtin/packages/alps/package.py
Ooolab: updated to include the change.

repos/spack_repo/builtin/packages/alps/package.py
Ooolab: Indeed, these are not required. We have removed them.

repos/spack_repo/builtin/packages/alps/package.py
Ooolab: ALPS need the source files of Boost to compile together and link to. This is how the original ALPS dependence on Boost was developed.

repos/spack_repo/builtin/packages/alps/package.py
Ooolab: It seems so, possibly due to the Boost source files being compiled and linked to later in the installation process.

repos/spack_repo/builtin/packages/alps/package.py
Ooolab: @egull Could you comment on this? Thanks.

repos/spack_repo/builtin/packages/alps/package.py
egull: @egull here. Yes, I agree with being a maintainer.

repos/spack_repo/builtin/packages/alps/package.py
egull: Totally ok by me. Go ahead, happy to be part on it.

repos/spack_repo/builtin/packages/alps/package.py
Sinan81: I think it would be useful if spack had a generic package option `+src` for installing source files at prefix, so that one could do `depends_on("boost+src")` etc. I vaguely remember this coming up in a few other occasions.

Sinan81 (approved):

Sinan81:
I think it would be useful to merge this sooner that later. Boost situation can be studied and improved in the next iteration, if any.

repos/spack_repo/builtin/packages/alps/package.py
alalazo: Sure, but what about the previous versions? Note that `license` has a `when=` argument, so you can still say older versions use BSL-1.0 Or are you saying that 2.3.3 ended up with MIT?

repos/spack_repo/builtin/packages/alps/package.py
alalazo: Question is why the sources are needed instead of just headers + libraries? I guess we can postpone this discussion if it needs changes to be upstreamed first.

Sinan81: yes, I'm inclined to postpone it.

Ooolab:
Updated license information; Would appreciate merging the pull request. Thanks!

alalazo (approved):

PR #3534, opened 2026-02-24
proj: fix bug when curl is deactivated

jcortial-safran:

### DIFF
--- a/repos/spack_repo/builtin/packages/proj/package.py
+++ b/repos/spack_repo/builtin/packages/proj/package.py
@@ -176,7 +176,7 @@ class CMakeBuilder(AnyBuilder, cmake.CMakeBuilder):
         else:
             test_flag = "PROJ4_TESTS"
         args.append(self.define(test_flag, self.pkg.run_tests))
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/proj/package.py
```diff
@@ -176,7 +176,7 @@ def cmake_args(self):
         else:
             test_flag = "PROJ4_TESTS"
         args.append(self.define(test_flag, self.pkg.run_tests))
-        if self.spec["curl"].satisfies("build_system=cmake"):
+        if self.spec.satisfies("+curl") and self.spec["curl"].satisfies("build_system=cmake"):
```
haampie: You can simplify this to

```suggestion
        if self.spec.satisfies("%curl build_system=cmake"):
```

the `%` means direct dependency in Spack v1.x

adamjstewart (approved):

jcortial-safran:
Unless I am mistaken, the CI failure is not linked to the content of the PR. Is it possible to retry it ?

PR #3538, opened 2026-02-24
netcdf-c: Many changes to make the build more configurable

dtaller:
## Pull request overview

This PR updates the Spack `netcdf-c` package recipe to make the build more configurable and adjust CMake configuration knobs, primarily around documentation generation, tests/examples toggles, and NCZarr enablement.

**Changes:**
- Added a `+doc` variant intended to control building/installing man-page documentation, and made `m4` conditional on `+doc` for some cases.
- Updated CMake arguments to tie tests/examples to Spack’s `run_tests` and added a CMake toggle for NCZarr based on the `nczarr_zip` variant.

# Current state:
- Breaks building `@4.4~doc` on host without m4: Configure fails with the error that m4 is required, so the change to add `+doc` needs a revert
- `ENABLE_NCZARR` is being toggled via the `nczarr_zip` variant, but `nczarr_zip` is documented (and used elsewhere in this package) as controlling only the zipfile storage backend via `libzip`. This also makes CMake builds behave differently than Autotools builds, which unconditionally enable NCZarr for `@4.8.1:` (see `configure_args()` around the `--enable-nczarr` logic). Consider either (a) introducing a separate `+nczarr` variant to control the core NCZarr feature for both build systems, or (b) leaving core NCZarr enabled and only using `+nczarr_zip` to control whether `libzip` is used.
- see the latest review comments by Copilot at the end for details

-> @dtaller


repos/spack_repo/builtin/packages/netcdf_c/package.py
```diff
@@ -374,9 +376,13 @@ def cmake_args(self):
             self.define(nc + "ENABLE_NETCDF_4", True),
             self.define_from_variant(nc + "ENABLE_DAP", "dap"),
             self.define_from_variant(nc + "ENABLE_HDF4", "hdf4"),
+            self.define("ENABLE_EXAMPLES", False),
```
adayton1: Are there standard variant names for examples and tests?

adayton1 (approved):

repos/spack_repo/builtin/packages/netcdf_c/package.py
dtaller: I could add a variant "examples" if you would like? A lot of packages have a variant called examples.
I just hard-coded it to off because when you build netcdf as part of very large project you typically do not care about building examples, but I could add that variant

dtaller: Added examples variant

skosukhin: If I remember correctly, the examples are not installed. This means that if we add the examples variant, we can have two different specs (`+example` and `~examples`) delivering identical installations. We should avoid that.

When building with Autotools, the examples are built only with `make check` (they are treated as extra test programs). There is no standard special target in CMake to build test programs, and I agree that we should skip building the examples if they are neither installed nor run as tests. If CMake runs the examples as tests, you can disable them by default and enable only when Spack is requested to run the tests, e.g. see [`eccodes`](https://github.com/spack/spack-packages/blob/ceae411aa6294f84ac7b2f356148724bd406f120/repos/spack_repo/builtin/packages/eccodes/package.py#L335).

dtaller: Removed the examples variant, set ENABLE_EXAMPLES defined based on self.run_tests as in eccodes.

repos/spack_repo/builtin/packages/netcdf_c/package.py
```diff
@@ -374,9 +376,13 @@ def cmake_args(self):
             self.define(nc + "ENABLE_NETCDF_4", True),
             self.define_from_variant(nc + "ENABLE_DAP", "dap"),
             self.define_from_variant(nc + "ENABLE_HDF4", "hdf4"),
+            self.define("ENABLE_EXAMPLES", self.run_tests),
+            self.define("ENABLE_TESTS", False),
```
skosukhin: I guess the following makes sense as well:
```suggestion
            self.define("ENABLE_TESTS", self.run_tests),
```

repos/spack_repo/builtin/packages/netcdf_c/package.py
```diff
@@ -374,9 +376,13 @@ def cmake_args(self):
...
+            self.define("ENABLE_FILTER_TESTING", False),
```
skosukhin:
```suggestion
            self.define("ENABLE_FILTER_TESTING", self.run_tests),
```
This one might be trickier, though. I don't know how it is implemented in CMake, but in Autotools, there is a subset of filters that can be disabled only with `~shared`. So, we might need something like:
```python
self.define("ENABLE_FILTER_TESTING",
  all(
    self.run_tests,
    self.spec.satisfies("@4.9.3:+shared"),  # see below in the code
    <there is at least one plugin to test>,
)
```

skosukhin (changes requested):

repos/spack_repo/builtin/packages/netcdf_c/package.py
dtaller: Made this change

repos/spack_repo/builtin/packages/netcdf_c/package.py
dtaller: I moved this option down to where we enable plugins. If I understood your comment correctly, then ENABLE_FILTER_TESTING should be defined as self.run_tests when plugins are enabled and False when plugins are disabled, so I moved this option into the block where we enable plugins.

Please let me know if I am not understanding this correctly. I appreciate your help.

repos/spack_repo/builtin/packages/netcdf_c/package.py
```diff
@@ -402,10 +407,14 @@ def cmake_args(self):
                 [
                     self.define("ENABLE_PLUGIN_INSTALL", True),
                     self.define("NETCDF_WITH_PLUGIN_DIR", self.prefix.plugins),
+                    self.define("ENABLE_FILTER_TESTING", self.run_tests),
                 ]
             )
         elif self.spec.satisfies("@4.9.0:+shared"):
             base_cmake_args.append(self.define("PLUGIN_INSTALL_DIR", self.prefix.plugins))
+            base_cmake_args.append(self.define("ENABLE_FILTER_TESTING", self.run_tests))
```
skosukhin: I would `extend` instead of appending twice (see above).

repos/spack_repo/builtin/packages/netcdf_c/package.py
```diff
@@ -402,10 +407,14 @@ def cmake_args(self):
...
+        else:
+            base_cmake_args.append(self.define("ENABLE_FILTER_TESTING", False))
```
skosukhin: I would double-check if this has any effect: the CMake scripts might never even check `ENABLE_FILTER_TESTING` when `~shared` (or when `ENABLE_PLUGIN_INSTALL` is `False` or something like that). Actually, this is relevant for the cases above as well.

dtaller: I actually removed the logic for ENABLE_FILTER_TESTING . Yes, you are right, the CMake system will disable this anyway in my case where shared and other features are disabled. I was a bit too aggressive in adding the flag here.

repos/spack_repo/builtin/packages/netcdf_c/package.py
dtaller: I actually removed the logic for ENABLE_FILTER_TESTING . Yes, you are right, the CMake system will disable this anyway in my case where shared and other features are disabled. I was a bit too aggressive in adding the flag here.

skosukhin (approved):

repos/spack_repo/builtin/packages/netcdf_c/package.py
```diff
@@ -374,9 +376,12 @@ def cmake_args(self):
             self.define(nc + "ENABLE_NETCDF_4", True),
             self.define_from_variant(nc + "ENABLE_DAP", "dap"),
             self.define_from_variant(nc + "ENABLE_HDF4", "hdf4"),
+            self.define("ENABLE_EXAMPLES", self.run_tests),
+            self.define("ENABLE_TESTS", self.run_tests),
```
wdconinc: This currently fails in ci because self here is a builder, not a package.
```suggestion
            self.define("ENABLE_TESTS", self.pkg.run_tests),
```
(Same for line above)

Ref: https://github.com/spack/spack-packages/blob/997b2ceb769f2da277d39363c04fd82164ca6ced/repos/spack_repo/builtin/packages/jsoncpp/package.py#L82

dtaller: Fixed

wdconinc:
@skosukhin If you approve, I can mark all comments as resolved and merge this.

wdconinc (dismissed):

repos/spack_repo/builtin/packages/netcdf_c/package.py
```diff
@@ -193,10 +195,10 @@ class NetcdfC(CMakePackage, AutotoolsPackage):
             # It is possible to install the package with CMake and without M4 on a non-Windows
             # platform but some of the man files will not be installed in that case (even if they
             # are in the release tarball):
-            depends_on("m4", type="build", when="build_system=cmake")
+            depends_on("m4", type="build", when="+doc build_system=cmake")
             # Apart from the redundant configure-time check, which we suppress below, M4 is not
             # needed when building with Autotools if the man files are in the release tarball:
-            depends_on("m4", type="build", when="@:4.4 build_system=autotools")
+            depends_on("m4", type="build", when="@:4.4 +doc build_system=autotools")
```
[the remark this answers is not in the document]
bernhardkaindl: This review comment is correct: I confirmed by building @4.4~doc on host without m4:
Configure fails with the error that m4 is required, so the change to add `+doc` needs a revert.

repos/spack_repo/builtin/packages/netcdf_c/package.py
```diff
@@ -374,9 +376,12 @@ def cmake_args(self):
             self.define(nc + "ENABLE_NETCDF_4", True),
             self.define_from_variant(nc + "ENABLE_DAP", "dap"),
             self.define_from_variant(nc + "ENABLE_HDF4", "hdf4"),
+            self.define("ENABLE_EXAMPLES", self.pkg.run_tests),
+            self.define("ENABLE_TESTS", self.pkg.run_tests),
             self.define(nc + "ENABLE_PARALLEL_TESTS", False),
             self.define_from_variant(nc + "ENABLE_FSYNC", "fsync"),
             self.define(nc + "ENABLE_LARGE_FILE_SUPPORT", True),
+            self.define_from_variant("ENABLE_NCZARR", "nczarr_zip"),
             self.define_from_variant("NETCDF_ENABLE_LOGGING", "logging"),
```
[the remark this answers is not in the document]
bernhardkaindl: @dtaller, can you check and fix (or maybe revert this change?) unless needed?

From the looks of it, this review comment appears to be correct, I'd suggest to revert/remove this change or create a dedicated variant if you want a variant for completely disabling `nczarr`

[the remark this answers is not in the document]
dtaller: Added +nczarr variant. I added the variant because excluding nczarr helped me avoid a linking error, and I do not need this anyway.

repos/spack_repo/builtin/packages/netcdf_c/package.py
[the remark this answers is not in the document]
dtaller: Reverted the +doc

repos/spack_repo/builtin/packages/netcdf_c/package.py
```diff
@@ -160,6 +160,8 @@ class NetcdfC(CMakePackage, AutotoolsPackage):
     variant("blosc", default=True, description="Enable Blosc compression plugin")
     variant("zstd", default=True, description="Enable Zstandard compression plugin")
 
+    variant("doc", default=True, description="build documentation man files")
```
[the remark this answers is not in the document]
skosukhin: This variant does not affect Autotools. Keeping it this way is misleading.

repos/spack_repo/builtin/packages/netcdf_c/package.py
```diff
@@ -466,7 +472,8 @@ def configure_args(self):
         # NCZarr was added in version 4.8.0 as an experimental feature and became a supported one
         # in version 4.8.1:
         if self.spec.satisfies("@4.8.1:"):
-            config_args.append("--enable-nczarr")
+            if self.spec.satisfies("+nczarr"):
```
skosukhin, line 475: Similar to HDF5, we do not support installations without NCZarr. Please revert.

skosukhin (changes requested):

bernhardkaindl:
@dtaller JFYI, there are new review comments for the change to review.

bernhardkaindl:
It looks like PRs have some issues with the Gitlab Pipeline, and the current state which has change requests is running a pipeline which (and it also already has many fails), so I'd like to convert this PR into draft status to stop the pipelines and not spend resources and money on running the GitLab pipeline until the review from the maintainers is green.

dtaller:
Hello, sorry for the delayed reply. Currently, out main code that uses netcdf disables NCZarr, and if NCZarr is enabled then it fails to compile. I will need to figure out how to change our main code build system to allow NCZarr enabled before returning to this, sorry about that

bernhardkaindl:
> Hello, sorry for the delayed reply. Currently, out main code that uses netcdf disables NCZarr, and if NCZarr is enabled then it fails to compile. I will need to figure out how to change our main code build system to allow NCZarr enabled before returning to this, sorry about that

@dtaller You don't have to to all at once, not all in a single PR.

About https://github.com/spack/spack-packages/pull/3538#discussion_r3508112767:

What about not changing NCZarr in this PR (it has many other changes), so the changes wanted get into spack, and then you change your main code build system to allow NCZarr enabled?

PR #3539, opened 2026-02-25
exodusii: Add missing releases

gsjaardema:
The seacas package which exodusII is a subset of has two releases which were not propogated over to exodusii package.  This fixes that.


tldahlgren (approved):
Version sha256s have been confirmed.

gsjaardema:
@tldahlgren Thanks.

PR #3541, opened 2026-02-25
root: add v6.38.02

jmcarcell:


RikkiButler20 (approved):
Verified sha256. LGTM

Issue #3400, opened 2026-02-12
Issue with ROOT and web-based display


jmcarcell:
After `civetweb` was added for ROOT 6.38.0 (https://github.com/spack/spack-packages/pull/2589), `civetweb` was added as a dependency. When trying to run something that requires a web browser, we have been finding crashes (https://github.com/key4hep/key4hep-spack/issues/848). I tried to change the options that are used to build civetweb (https://github.com/root-project/root/pull/14194#issuecomment-3847590289) but it still crashed. The workaround I did was to use the builtin civetweb from ROOT, and that works fine.

My setup: ROOT 6.38.0 on Alma 9 with GCC 14 and Ubuntu 24 with GCC 13 (system compiler).

PR #3542, opened 2026-02-25
root: enable builtin_civetweb until issues are fixed

jmcarcell:
There are crashes when running the browser when using an external civetweb (see
https://github.com/spack/spack-packages/issues/3400) and this feature seems not
to have been extensively tested. I tried disabling and enabling different options for civetweb but crashes were still happening both on Alma 9 and Ubuntu 24. The maintainer seems to only have tried this with the civetweb that comes in Fedora.

### DIFF
--- a/repos/spack_repo/builtin/packages/root/package.py
+++ b/repos/spack_repo/builtin/packages/root/package.py
@@ -437,7 +437,8 @@ class Root(CMakePackage):
 
     # Optional dependencies
     depends_on("arrow", when="+arrow")
-    depends_on("civetweb +shared", when="+http")
+    # Commented out until it is fixed and builtin_civetweb can be turned off
+    # depends_on("civetweb +shared", when="+http")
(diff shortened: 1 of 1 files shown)


wdconinc:
I think this is a bit of an unfortunate band-aid, and would turn civetweb into the ONLY feature builtin we use (llvm being the other one due to extensive patches).

Has there been any attempt at bisecting this between civetweb 1.16.0 and https://github.com/civetweb/civetweb/tree/d5321963b1d0bc953101de91f8588bf83db73bf5 (which was what was current when https://github.com/root-project/root/commit/d14759bc1679fec7c21d1ce96a01f5482ddd67a1 happened)? I can't reproduce this, or I would be able to help.

If anything, you should be able to verify that using `civetweb@git.d5321963b1d0bc953101de91f8588bf83db73bf5=1.16.0` results in a working version (and if it doesn't then something else must be going on).

PR #3543, opened 2026-02-25
icon: add single-precision variants

huppd:
This PR adds two variants to the ICON package: single-precision and single-precision-ecrad.


huppd:
@spackbot fix style

skosukhin:
There are now such configure options in the version of ICON that can currently be installed with Spack.

huppd:
> There are now such configure options in the version of ICON that can currently be installed with Spack.

Thanks for raising this point. I agree that `single-percision-ecrad` was not there yet in 2025.04. But should be available in 2025.10-2. Is it OK if I add this version here?

huppd:
@spackbot fix style

repos/spack_repo/builtin/packages/icon/package.py
```diff
@@ -102,6 +105,15 @@ class Icon(AutotoolsPackage):
 
     # Optimization Features:
     variant("mixed-precision", default=False, description="Enable mixed-precision dynamical core")
+    variant(
+        "single-precision", default=False, description="Enable single-precision", when="@2025.04:"
```
skosukhin: If nothing has changed over the last year or two, the semantics of `when` in a `variant` is rather confusing, and I don't remember a single case when it is used correctly.

When I say `icon~single-precision`, I mean "I don't want ICON to be built with the single-precision enabled", and even versions that are older than `2025.04` can deliver that. However, with `when="@2025.04:"` in the `variant`, `icon~single-precision` means "I want version 2025.04 or newer, and I don't want it to be built with the single-precision enabled".

So, instead, we need a `conflicts` for `+single-precision`, `when` the version is too old (how I usually do it). A `requires` for a newer version, `when` `+single-precision` might work as well.

repos/spack_repo/builtin/packages/icon/package.py
```diff
@@ -24,6 +24,9 @@ class Icon(AutotoolsPackage):
 
     license("BSD-3-Clause", checked_by="skosukhin")
 
+    version(
```
skosukhin, line 27: Thanks!

huppd:
@spackbot fix style

repos/spack_repo/builtin/packages/icon/package.py
huppd: Thanks, I like that, so now one gets also a nice error message, when using an too old version.

skosukhin (approved):

Try2Code:
thx for the work @huppd

Try2Code (approved):

huppd:
Thanks a lot for the review! I was wondering if there is anything left for me to do to get this merged?

alalazo (approved):
Maintainers already approved

huppd:
Thanks @alalazo for setting the PR to automerge, unfortunately the ci/gitlab-ci seems to be stuck.

alalazo:
@spackbot run pipeline

PR #3546, opened 2026-02-25
memtailor: add v1.3

d-torrance:


RikkiButler20 (approved):
Verified sha256. LGTM

PR #3555, opened 2026-02-26
lua-luaposix: add v36.3

drkrynstrng:
Adding luaposix v36.3. Also adding Lua version constraints and updating the url template.

### DIFF
--- a/repos/spack_repo/builtin/packages/lua_luaposix/package.py
+++ b/repos/spack_repo/builtin/packages/lua_luaposix/package.py
@@ -8,23 +8,23 @@ from spack.package import *
 
 
 class LuaLuaposix(LuaPackage):
-    """Lua posix bindings, including ncurses"""
+    """Lua POSIX bindings, including ncurses"""
 
-    homepage = "https://github.com/luaposix/luaposix/"
-    url = "https://github.com/luaposix/luaposix/archive/release-v33.4.0.tar.gz"
+    homepage = "https://github.com/luaposix/luaposix"
+    url = "https://github.com/luaposix/luaposix/archive/refs/tags/v36.3.tar.gz"
 
     license("MIT")
 
-    version(
-        "35.0",
-        sha256="a4edf2f715feff65acb009e8d1689e57ec665eb79bc36a6649fae55eafd56809",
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/lua_luaposix/package.py
```diff
@@ -8,23 +8,23 @@
...
-    version(
-        "35.0",
-        sha256="a4edf2f715feff65acb009e8d1689e57ec665eb79bc36a6649fae55eafd56809",
-        url="https://github.com/luaposix/luaposix/archive/refs/tags/v35.0.tar.gz",
-    )
+    version("36.3", sha256="82cd9a96c41a4a3205c050206f0564ff4456f773a8f9ffc9235ff8f1907ca5e6")
     version("36.1", sha256="5e48e8ea5adfd8958bcdbaba132f5766473afedecc5c14ae3593464a5463a616")
+    version("35.0", sha256="a4edf2f715feff65acb009e8d1689e57ec665eb79bc36a6649fae55eafd56809")
     version("33.4.0", sha256="e66262f5b7fe1c32c65f17a5ef5ffb31c4d1877019b4870a5d373e2ab6526a21")
     version("33.2.1", sha256="4fb34dfea67f4cf3194cdecc6614c9aea67edc3c4093d34137669ea869c358e1")
```
haampie:
```suggestion
    def url_for_version(self, version):
        if version >= Version("35.0"):
            return f"https://github.com/luaposix/luaposix/archive/refs/tags/v{version}.tar.gz"
        return f"https://github.com/luaposix/luaposix/archive/release-v{version}.tar.gz"
```

the shasums for older versions are different after changing the URL. I would use url_for_version.

haampie (changes requested):

repos/spack_repo/builtin/packages/lua_luaposix/package.py
drkrynstrng: Hmm, it worked for me even after clearing all cache files. But yes, checking manually I see that the checksums changed.

drkrynstrng:
@spackbot fix style

haampie (approved):

PR #3563, opened 2026-02-26
py-yarl: add cython constraint for PEP 703 compatibility

LydDeb:
The `freethreading_compatible` option appears in `yarl`'s `pyproject.toml` file starting with version 1.21.
This option is only supported from `cython` version 3.1 onwards.

### DIFF
--- a/repos/spack_repo/builtin/packages/py_yarl/package.py
+++ b/repos/spack_repo/builtin/packages/py_yarl/package.py
@@ -31,6 +31,9 @@ class PyYarl(PythonPackage):
         depends_on("py-setuptools@40:", when="@1.7.2:")
         depends_on("py-setuptools")
         depends_on("py-tomli", when="@1.9.3: ^python@:3.10")
+        # The freethreading_compatible directive is supported only by Cython >= 3.1
+        # cf. PEP 703 and https://docs.python.org/3/howto/free-threading-python.html
+        depends_on("py-cython@3.1:", when="@1.21:")
         # requires https://github.com/cython/cython/commit/ea38521bf59edef9e6d22cbabf44229848091a76
         depends_on("py-cython@3:", when="@1.15.4:")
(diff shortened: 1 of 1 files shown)


manuelakuhn (commented):
As far as I can tell, this is already present in `@1.20:`, 
see https://github.com/aio-libs/yarl/blob/v1.20.0/pyproject.toml#L67

Other than that this looks good to me.

LydDeb:
> As far as I can tell, this is already present in `@1.20:`, see https://github.com/aio-libs/yarl/blob/v1.20.0/pyproject.toml#L67
> 
> Other than that this looks good to me.

I see what you mean. Indeed, in version 1.20.0 there is an `enable = ["cpython-freethreading"]` in the `[tool.cibuildwheel]` block. (cf. https://github.com/aio-libs/yarl/blob/2127e8dee07ecc92d2d98b708ff23fb80f121ecb/pyproject.toml#L67)

Similarly, it is also present in version 1.21.0 in the same block. (cf. https://github.com/aio-libs/yarl/blob/d6d21162bae44cf0b256b9ae202509d71382dbdc/pyproject.toml#L68)

However, this configuration seems to only concern the CI/CD part of the yarl project. And, from what I understand, it's the addition of `freethreading_compatible = "True"` in the `[tool.local.cythonize.kwargs.directive]` block that causes an error, during an installation with Spack, if the Cython version is too old (cf. https://github.com/aio-libs/yarl/blob/d6d21162bae44cf0b256b9ae202509d71382dbdc/pyproject.toml#L52)

Should I still extend the constraint?

manuelakuhn:
Ok, thanks for the explanation. I am fine if with the constraints as is if you add a comment where this comes from.

manuelakuhn (approved):
Thanks. LGTM and builds fine on rocky 8 with gcc@15

PR #3568, opened 2026-02-26
py-setuptools-scm: forward compat with setuptools

haampie:
Fix CI on develop: old py-setuptools-scm needs old py-setuptools


alalazo (approved):
Force merging since CI is currently broken on this

PR #3571, opened 2026-02-26
stacks: rework e4s to enforce a single AMD architecture

alalazo:
Currently, the AMD software built in the `e4s` pipelines is causing a lot of latency for other pipelines to run. One possible cause is the fact that some nodes in the DAG get `amdgpu_target:=auto` and thus build for a huge number of targets (thanks @haampie for noticing that). 

See for instance:
- https://gitlab.spack.io/spack/spack-packages/-/jobs/20751824

that gets to 99% completion in ~4hrs then fails. These build failures are retried 3 times, amounting to ~12hrs waiting time for the PRs that are gated on `develop` pipelines to finish.

In this PR we try to pin the entire `e4s` stack to build a single AMD target, and see it that helps with the build times and failure rates.


eugeneswalker (dismissed):

eugeneswalker (dismissed):

alalazo:
@spackbot run pipeline

stacks/e4s/spack.yaml
```diff
@@ -44,36 +52,88 @@ spack:
     rocblas:
       require: ~hipblaslt #hangs in ci
     trilinos:
```
kuberry, line 54: Are these variant choices preferred or required? I would expect to see a required set of variants with a preferred set of `~` variants (prefer not to build them, but if they're already built or must be build then we allow it).

alalazo: These are preferred, to try to reduce the number of different `trilinos` configuration we build in the public `e4s` stack[^1].

The change I did in `xyce` was to remove what is a hard requirement on `trilinos`, but seems just to be build preferences for stuff that you don't want `trilinos` to build for a single:
```console
$ spack install xyce
```
It turns out though that imposing those restrictions in an environment where we already build `trilinos` with those variants on causes yet another build of `trilinos` 🙂 

[^1]: Didn't turn out great, I reduced by 2 I think so far

kuberry: Changes to the Trilinos dependency in Xyce look good to me.

eugeneswalker:
This looks great.

One question before we merge: why change to `~pymi_static_tpls` on Xyce? That that variant was explicitly enabled says to me it was important, as we normally do not explicitly enable variants unless a code team requests it or we find it necessary for a use case.

alalazo:
> why change to ~pymi_static_tpls on Xyce?

This stems from 2 places:
1. First, there's the removal of some "artificial" constraint in `xyce` that lead to more builds of `trilinos` (thread in https://github.com/spack/spack-packages/pull/3571#discussion_r2883432276)
2. At that point `xyce` failed to build because it couldn't find some static libraries. I didn't investigate further what was misconfigured there, but just noticed that the `~pymi_static_tpls` seemed not to be needed in pipelines according to its own description.

The `pymi_static_tpls` variant is defined like:
```python
    # Downstream dynamic library symbols from pip-installed numpy and other                                                                                                                                                                                                      
    # pip-installed python packages can cause conflicts. This is most often                                                                                                                                                                                                      
    # seen with blas symbols from numpy, and building blas static resolves                                                                                                                                                                                                       
    # this issue.                                                                                                                                                                                                                                                                
    variant(
        "pymi_static_tpls",
        default=True,
        sticky=True,
        when="+pymi",
        description="Require static blas build for PyMi",
    )
```
but we don't `pip` install our `py-numpy`. Instead, we ensure it's built with a consistent stack.

eugeneswalker (approved):

PR #3581, opened 2026-02-27
py-metaphlan: added version 4.2.4, fix py-cmseq requirement

V-Karch:
Adding a new version for `py-metaphlan` and making sure it builds uses `py-cmseq` and `py-hclust2` properly

### DIFF
--- a/repos/spack_repo/builtin/packages/py_metaphlan/package.py
+++ b/repos/spack_repo/builtin/packages/py_metaphlan/package.py
@@ -40,4 +40,3 @@ class PyMetaphlan(PythonPackage):
     depends_on("blast-plus@2.6:", type=("build", "run"))
     depends_on("raxml@8.2.10:", type=("build", "run"))
     depends_on("samtools@1.9:", type=("build", "run"))
-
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/py_metaphlan/package.py
RikkiButler20: Is the py-hclust2 dependency needed for v4.2.4?

V-Karch: @RikkiButler20 I don't believe so, It wasn't in the `setup.py` from `py-metaphlan@4.2.4`

RikkiButler20:
```suggestion
    depends_on("py-hclust2", type=("build", "run"), when="@4.0.2")
```
`py-clust2` is only a dependency for v4.0.2
After this fix, everything looks good.

RikkiButler20 (changes requested):

RikkiButler20 (approved):
LGTM. Thank you.

PR #2710, opened 2025-12-09
glib: ensure no system glib is used

Chrismarsh:
This fixes https://github.com/spack/spack-packages/issues/2706.

Although `gobject-introspection` correctly uses the `glib-bootstrap`, `glib`'s meson build seems to pick up system glib (which may be too old), and fails to build.

Checking what pkg-config thinks the glib version is
```
$ spack load gobject-introspection
$ spack load pkgconfig
$ pkg-config --cflags glib-2.0
-I/usr/include/glib-2.0 -I/usr/lib64/glib-2.0/include -I/usr/include -I/usr/include/sysprof-4 -pthread
```

This also adds a depdency on `cmake` as the meson build explicitly looks for it and will fall back to a system cmake if a spack cmake is not provided:
```
Found CMake: /usr/bin/cmake (3.26.5)
Run-time dependency gobject-introspection-1.0 found: NO (tried cmake)
```

/cc @michaelkuhn

Issue #2706, opened 2025-12-09
Installation issue: glib@2.86


Chrismarsh:
### Steps to reproduce the issue

```console
$ spack install glib@2.86
```

### Error message

https://github.com/spack/spack-packages/pull/1625 Does not seem to have fully resolved all the sisues
```
Found CMake: /usr/bin/cmake (3.26.5)
Run-time dependency gobject-introspection-1.0 found: NO (tried cmake)

../spack-src/meson.build:2699:24: ERROR: Dependency lookup for gobject-introspection-1.0 with method 'pkgconfig' failed: Could not generate cflags for gobject-introspection-1.0:
Package dependency requirement 'glib-2.0 >= 2.82.0' could not be satisfied.
Package 'glib-2.0' has version '2.68.4', required version is '>= 2.82.0'
Package dependency requirement 'gobject-2.0 >= 2.82.0' could not be satisfied.
Package 'gobject-2.0' has version '2.68.4', required version is '>= 2.82.0'
```

also: this appears to need a build cmake dependency (`Found CMake: /usr/bin/cmake (3.26.5)`) 

### Information on your system

```
 spack debug report
* **Spack:** 1.2.0.dev0 (https://github.com/spack/spack/commit/499a1b5494fc8c2d7bfe8bd74e84033ae690ca17)
* **Builtin repo:** https://github.com/spack/spack-packages/commit/e6dbe840bfaecfee7e85f45f08d762b7c12f053d
* **Python:** 3.9.18
* **Platform:** linux-rhel9-sapphirerapids

```

### Additional information

@michaelkuhn 

### General information

- [x] I have run `spack debug report` and reported the version of Spack/Python/Platform
- [x] I have run `spack maintainers <name-of-the-package>` and **@mentioned** any maintainers
- [x] I have uploaded the build log and environment files
- [x] I have searched the issues of this repo and believe this is not a duplicate

Chrismarsh:
Looks like it is maybe this issue
https://bugs.gentoo.org/951487

```
$  cd gobject-introspection-1.86.0-p62irw2q64jk4yebdbp2tvvfi4x5sgxq/lib/pkgconfig
$ cat gobject-introspection-1.0.pc
<...>
Name: gobject-introspection
Description: GObject Introspection
Version: 1.86.0
Requires: glib-2.0 >= 2.82.0, gobject-2.0 >= 2.82.0

```

Chrismarsh:
I think system paths are sneaking in:
```
[...]

PR #3584, opened 2026-02-27
Bugfix: glib detection for glib when building with gobject-introspection

kwryankrattiger:
Path ordering when an system installed glib pkg config is in the same path as an external dependency results in failure to find the correct `glib-2.0` package.

The `gobject-introspection` package handles a transitive build dependency on `glib-2.0` by appending `glib-bootstrap` to the `PKG_CONFIG_PATH`. 

_Most_ of the cases in Spack already have a glib dependency which should be sufficient for using `gobject-introspection` for builds. Two cases not covered are `babl` and `glib`. `babl` just gets the `glib` package directly and `glib` needs to use the `glib-bootstrap` to avoid recursive dependencies.

### DIFF
--- a/repos/spack_repo/builtin/packages/glib/package.py
+++ b/repos/spack_repo/builtin/packages/glib/package.py
@@ -97,6 +97,10 @@ class Glib(MesonPackage):
         depends_on("meson@0.48.0:")
         depends_on("pkgconfig")
         depends_on("gobject-introspection@1.80:", when="+introspection")
+        # glib-bootstrap is needed in the PKG_CONFIG_PATH during build to
+        # configure gobject-instrospection. After that glib itself can be
+        # used as the glib implementation.
+        depends_on("glib-bootstrap", type="build", when="%gobject-introspection@1.80:")
 
     depends_on("libffi")
     depends_on("zlib-api")
--- a/repos/spack_repo/builtin/packages/gobject_introspection/package.py
+++ b/repos/spack_repo/builtin/packages/gobject_introspection/package.py
@@ -158,9 +158,6 @@ class BuildEnvironment:
         env.prepend_path("GI_TYPELIB_PATH", join_path(self.prefix.lib, "girepository-1.0"))
         env.set("GI_SCANNER_DISABLE_CACHE", "1")
 
-        if self.spec.satisfies("^glib-bootstrap"):
-            env.append_path("PKG_CONFIG_PATH", self.spec["glib-bootstrap"].prefix.lib.pkgconfig)
-
 
 class AutotoolsBuilder(BuildEnvironment, autotools.AutotoolsBuilder):
     @run_before("build")
--- a/repos/spack_repo/builtin/packages/babl/package.py
+++ b/repos/spack_repo/builtin/packages/babl/package.py
@@ -37,6 +37,7 @@ class Babl(MesonPackage):
     depends_on("pkgconfig", type="build")
     depends_on("lcms")
     depends_on("gobject-introspection")
+    depends_on("glib", type="build")
 
     def setup_dependent_build_environment(
         self, env: EnvironmentModifications, dependent_spec: Spec


repos/spack_repo/builtin/packages/glib/package.py
```diff
@@ -97,6 +97,11 @@ class Glib(MesonPackage):
         depends_on("meson@0.48.0:")
         depends_on("pkgconfig")
         depends_on("gobject-introspection@1.80:", when="+introspection")
+        depends_on(
+            "glib-bootstrap",
+            type="build",
+            when="+introspection ^gobject-introspection@1.80:"
```
haampie:
```suggestion
            when="%gobject-introspection"
```

reduces duplication a bit

repos/spack_repo/builtin/packages/glib/package.py
```diff
@@ -97,6 +97,11 @@ class Glib(MesonPackage):
         depends_on("meson@0.48.0:")
         depends_on("pkgconfig")
         depends_on("gobject-introspection@1.80:", when="+introspection")
+        depends_on(
+            "glib-bootstrap",
+            type="build",
```
haampie:
```suggestion
```

kwryankrattiger: We only want to being it into pkg-config for the build, the runtime should not include it I think. It wasn't added there previously, and it provides a conflicting package.

repos/spack_repo/builtin/packages/glib/package.py
kwryankrattiger: We only want this if gobject-introspection is going to need the bootstrap, which is only in the newer versions it seems.

We could drop the variant I suppose.

repos/spack_repo/builtin/packages/glib/package.py
kwryankrattiger: I added a comment about why it is build only.

repos/spack_repo/builtin/packages/gobject_introspection/package.py
```diff
@@ -158,9 +158,6 @@ def setup_dependent_build_environment(
         env.prepend_path("GI_TYPELIB_PATH", join_path(self.prefix.lib, "girepository-1.0"))
         env.set("GI_SCANNER_DISABLE_CACHE", "1")
 
-        if self.spec.satisfies("^glib-bootstrap"):
-            env.append_path("PKG_CONFIG_PATH", self.spec["glib-bootstrap"].prefix.lib.pkgconfig)
-
```
michaelkuhn, line 163: What about other packages depending on gobject-introspection? Don't they also need glib-bootstrap in `PKG_CONFIG_PATH`?

michaelkuhn: Sorry, I missed the third paragraph in your initial comment, which already answers this.

michaelkuhn (approved):
I still think this is probably an underlying problem in Spack's handling of external packages, but it seems to be a relatively harmless workaround.

repos/spack_repo/builtin/packages/gobject_introspection/package.py
kwryankrattiger: They already have a `glib` dependency which provides the `glib-2.0` package that pkgconfig is looking for. See how I fix `babl` by adding `glib` instead of `glib-bootstrap`.

My understanding from these packages is the only purpose of having `glib-bootstrap` is to have a package that provides a `glib-2.0` that can be used to build `glib +introspection`.

kwryankrattiger:
> I still think this is probably an underlying problem in Spack's handling of external packages, but it seems to be a relatively harmless workaround.

The issue here was that the package adding paths to `PKG_CONFIG_PATH` did not respect topological ordering or external/built distinction. From the package recipe we can handle external more or less (I had done this in the previous revision), but it is not possible to handle topological order insertion correctly.

You can still manufacture cases that fail by installing into a system root with another `glib-2.0` in it or by having all external dependencies from multiple roots with multiple `glib-2.0` pkgconfig files. But at that point all of Spack's ability to distinguish what is what is already pretty broken so idk if it actually matters that much. Whenever you are using externals you need to be cognoscente of where those paths are pointing to and what else is being pulled in implicitly.

Chrismarsh:
I believe this is the same problem I ran into here https://github.com/spack/spack-packages/issues/2706 and tried to fix here https://github.com/spack/spack-packages/pull/2710

I had a really hard time reproducing it other than in an empty spack env performing a complex set of builds. A retry would then pass without issue. I did have a `/usr`  external. 

This PR looks to use the same solution as me, which solved my problem, so I'm happy about that. I will close https://github.com/spack/spack-packages/pull/2710. It may be out of scope here, but I think the cmake addition should be added to avoid future issues https://github.com/spack/spack-packages/pull/2710#discussion_r2748217783

/cc @becker33

haampie (approved):

PR #3593, opened 2026-03-01
libxslt: depends on cmake >= 3.18 since version 1.1.38

cessenat:
libxslt complains if cmake is lower than 3.18, and that from version 1.1.38 on.


wdconinc (approved):
Confirmed. Thanks.

PR #3594, opened 2026-03-01
qt_base: depends on cmake>=3.19 since 6.8.1 and  >= 3.22 since 6.9.0

cessenat:
qt_base: depends on cmake >= 3.19 since version 6.8.1 and cmake >= 3.22 since 6.9.0


wdconinc (approved):
Thanks for catching this!

PR #3597, opened 2026-03-02
gocryptfs: add v2.6.1

robertexternbrink:
apptainer cannot be built with gocryptfs 2.5.1 anymore, adding 2.6.1 fixes this.

go 1.19 is still the minimal version required.

@snehring please review


alecbcs (changes requested):
@brinkcoder version addition looks good to me. Can you also add the dependency constraint into the apptainer package?

snehring:
checksum checks out, concur with Alec, apptainer recipe should be updated to reflect this restriction.

repos/spack_repo/builtin/packages/apptainer/package.py
```diff
@@ -69,7 +69,7 @@ class Apptainer(SingularityBase):
     depends_on("go@1.20:", when="@1.3:")
     depends_on("go@1.22.7:", when="@1.4:")
     depends_on("go@1.23.6:", when="@1.4.1:")
-    depends_on("gocryptfs@2.4:", type="run", when="@1.3:")
+    depends_on("gocryptfs@2.6.1:", type="run", when="@1.3:")
```
alecbcs:
```suggestion
    depends_on("gocryptfs@2.4:", type="run", when="@1.3:")
    depends_on("gocryptfs@2.6.1:", type="run", when="@1.4.4:")
```

I'm not exactly sure when the newer dependency on `gocryptfs` is enforced in apptainer, but we should add another `depends_on()` directive with the updated constraint instead of modifying the existing directive. (If we simply modify the existing directive, we'll retroactively update the dependency for older versions of apptainer which would invalidate people's older installations.)

alecbcs (changes requested):

repos/spack_repo/builtin/packages/apptainer/package.py
robertexternbrink: I was not able to build apptainer down to version 1.3.2 with gocryptfs@2.5.1 but I will follow your suggestion.

snehring:
@brinkcoder I just got around to testing this. What issue were you seeing when attempting to build apptainer with gocryptfs @2.4.0 or @2.5.1? I'm able to build ```apptainer@1.4.4 ^gocryptfs@2.4``` successfully.

snehring:
#3655 _might_ be relevant here I hit that when I was testing.

snehring:
yeah, I bet that's it. 2.6.1 falls back to the openssl-less build successfully when cgo_enabled=0 whereas 2.4 and 2.5.1 don't. This changed in https://github.com/rfjakob/gocryptfs/commit/2ebd0d754b8ee46e6c65e90e1d1e13491b03b7b5

robertexternbrink:
Yes, it is #3655 
So I roll back the dependency for apptainer and just add gocryptfs@2.6.1
Sorry for the noise

snehring:
Yeah I think that's the way to go. Sorry to make you change that only to change it back!

snehring (approved):

alecbcs (approved):
Looks good to me. Thanks!

@snehring thanks for tracking down the issue to `cgo_enabled`

Issue #3569, opened 2026-02-26
Installation issue: PDI along with NumPy


tpadioleau:
### Steps to reproduce the issue

Concretize the following environment.

```yaml
spack:
  concretizer:
    reuse: false
    unify: true
  repos:
    builtin:
      commit: 9629cca2ad70ff3b1b63143a0c12368931e6388a
  specs:
    - 'pdi@1.10 +python'
    - 'py-numpy@2'
```

### Error message

<details><summary>==> Error: failed to concretize `pdi@1.10:1+python`, `py-numpy@2` for the following reasons:</summary>

<pre>
     1. Cannot satisfy 'py-numpy@2' 1(1.26.4)
     2. Cannot satisfy 'py-numpy@2' 4(1.26.4)
        required because py-numpy@2 requested explicitly 
     3. Cannot satisfy 'py-numpy@1.21.5:1' and 'py-numpy@2'
        required because pdi depends on py-numpy@1.21.5:1 when @1.10.0:+python 
          required because pdi@1.10:1+python requested explicitly 
        required because py-numpy@2 requested explicitly 
     4. Cannot satisfy 'py-numpy@1.17.4:1' and 'py-numpy@2'
        required because pdi depends on py-numpy@1.17.4:1 when +python 
          required because pdi@1.10:1+python requested explicitly 
        required because py-numpy@2 requested explicitly . You could consider setting `concretizer:unify` to `when_possible` or `false` to allow multiple versions of some packages.
</pre></details>

### Information on your system

* **Spack:** 1.1.1
* **Builtin repo:** https://github.com/spack/spack-packages/commit/9629cca2ad70ff3b1b63143a0c12368931e6388a
* **Python:** 3.8.10
* **Platform:** linux-ubuntu20.04-icelake

### Additional information

@jbigot

### General information

- [x] I have run `spack debug report` and reported the version of Spack/Python/Platform
- [x] I have run `spack maintainers <name-of-the-package>` and **@mentioned** any maintainers
- [x] I have uploaded the build log and environment files
- [x] I have searched the issues of this repo and believe this is not a duplicate

tpadioleau:
solved by https://github.com/spack/spack-packages/pull/3603

PR #3603, opened 2026-03-02
Update PDI dependencies to numpy:2 that is now tested in CI

jbigot:
Accept numpy v2 as we now test against it in the CI


tpadioleau:
closes #3569

alalazo (approved):

PR #3609, opened 2026-03-02
Acts: deprecation removals

kwryankrattiger:
Remove versions deprecated prior to v2025.11 in preparation for release.


stephenswat:
Please note that with the removal of these versions, a lot of the dependencies and variants can also be cleaned up. Would you want to take care of that or would you prefer if I follow up on this?

kwryankrattiger:
> Please note that with the removal of these versions, a lot of the dependencies and variants can also be cleaned up. Would you want to take care of that or would you prefer if I follow up on this?

It doesn't look too bad at a glance I can take a stab at it.

stephenswat (approved):
Gorgeous, thank you!

wdconinc (approved):

PR #3622, opened 2026-03-02
qgis: add 3.44

Chrismarsh:
- adds the last of the qgis 3.x line, `qgis@3.44`
- updated patch for py sip due to changed layout of toml file
- adds a new version of `py-psycopg2` to allow for `python@3.14` support with qgis
- bounds the `libspatialindex` version due to an upstream API change that broke qgis

### DIFF
--- a/repos/spack_repo/builtin/packages/qgis/package.py
+++ b/repos/spack_repo/builtin/packages/qgis/package.py
@@ -235,8 +235,8 @@ class Qgis(CMakePackage):
         sip_inc_dir = join_path(self["qscintilla"].module.python_platlib, pyqtx, "bindings")
         sip_line = f'sip-include-dirs = ["{sip_inc_dir}"]'
 
-        # Recent QGis have added a @sipabi@ that expands to the [tool.sip.project] header. 
-        # The simple patch append approach  results in a duplicate header which causes a 
+        # Recent QGis have added a @sipabi@ that expands to the [tool.sip.project] header.
+        # The simple patch append approach  results in a duplicate header which causes a
         # build failure. The middle case is probably overkill but it is  kept as a just
         # in (edge) case
 
--- a/repos/spack_repo/builtin/packages/qgis/package.py
+++ b/repos/spack_repo/builtin/packages/qgis/package.py
@@ -134,6 +134,7 @@ class Qgis(CMakePackage):
     depends_on("libspatialite@4.2.0:")
     depends_on("libzip")
     depends_on("libtasn1")
+
     depends_on("proj@4.4.0:")
     depends_on("proj@4.9.3:", when="@3.8.2:")
     depends_on("proj@7.2:", when="@3.28:")
@@ -141,6 +142,8 @@ class Qgis(CMakePackage):
     # fails to build with proj 9.4+ until the backported patch in 3.34.5
     # https://github.com/qgis/QGIS/pull/56761
     depends_on("proj@:9.3", when="@:3.34.4")
+    depends_on("proj@:9.5", when="@3.34.5:3.34.15")
+
     depends_on("py-psycopg2", type=("build", "run"))  # TODO: is build dependency necessary?
     depends_on("py-pyqt4", when="@2")
     depends_on("py-pyqt5@5.3:", when="@3")
--- a/repos/spack_repo/builtin/packages/qgis/package.py
+++ b/repos/spack_repo/builtin/packages/qgis/package.py
@@ -235,7 +235,7 @@ class Qgis(CMakePackage):
         sip_inc_dir = join_path(self["qscintilla"].module.python_platlib, pyqtx, "bindings")
         sip_line = f'sip-include-dirs = ["{sip_inc_dir}"]'
 
-        # Recent QGis have added a @sipabi@ that expands to the [tool.sip.project] header.
+        # QGis 3.42.0 added a @sipabi@ that expands to the [tool.sip.project] header.
         # The simple patch append approach  results in a duplicate header which causes a
         # build failure. The middle case is probably overkill but it is  kept as a just
         # in (edge) case
--- a/repos/spack_repo/builtin/packages/qgis/package.py
+++ b/repos/spack_repo/builtin/packages/qgis/package.py
@@ -6,6 +6,8 @@ from spack_repo.builtin.build_systems.cmake import CMakePackage
 
 from spack.package import *
 
+from pathlib import Path
+import re
 
 class Qgis(CMakePackage):
     """QGIS is a free and open-source cross-platform desktop geographic
@@ -14,12 +16,13 @@ class Qgis(CMakePackage):
     """
 
     homepage = "https://qgis.org"
-    url = "https://qgis.org/downloads/qgis-3.8.1.tar.bz2"
+    url = "https://qgis.org/downloads/qgis-3.44.7.tar.bz2"
 
     maintainers("adamjstewart", "Sinan81", "Chrismarsh")
 
     license("GPL-2.0-or-later")
 
+    version("3.44.7", sha256="1ab06f40600c84e928b4fe22a66997d80973201b10769e7a636e5be83459b814")
     # Prefer latest LTR
     version(
         "3.40.6",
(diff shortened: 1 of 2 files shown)


Chrismarsh:
@adamjstewart @Sinan81

Chrismarsh:
@spackbot fix style

Chrismarsh:
@spackbot fix style

adamjstewart (commented):
New patch syntax is rather complex, I'm hoping we can either:

1. overwrite the entire file every time,
2. use `filter_file` to replace existing lines, and/or
3. use `if self.spec.satisfies("@X.Y:Z"):` to make things conditional on version

Honestly, @Sinan81 wrote most of this file, so I'm hoping he can provide a more useful review.

Sinan81:
will provide a review very soon.

repos/spack_repo/builtin/packages/qgis/package.py
```diff
@@ -223,9 +229,35 @@ def fix_qsci_sip(self):
...
 
+        pp = Path("python/gui/pyproject.toml.in")
+        txt = pp.read_text(encoding="utf-8")
+
         sip_inc_dir = join_path(self["qscintilla"].module.python_platlib, pyqtx, "bindings")
-        with open(join_path("python", "gui", "pyproject.toml.in"), "a") as tomlfile:
-            tomlfile.write(f'\n[tool.sip.project]\nsip-include-dirs = ["{sip_inc_dir}"]\n')
+        sip_line = f'sip-include-dirs = ["{sip_inc_dir}"]'
+
+        # Recent QGis have added a @sipabi@ that expands to the [tool.sip.project] header.
```
Sinan81: I would simply specify here that `@sipapi@` was added as of version 3.42.0 as can be seen via doing blame on github repo then click on the commit that introduced this. Apart from that it looks good to me.

Sinan81 (changes requested):

Sinan81:
> New patch syntax is rather complex, I'm hoping we can either:
> 
> 1. overwrite the entire file every time,
> 2. use `filter_file` to replace existing lines, and/or
> 3. use `if self.spec.satisfies("@X.Y:Z"):` to make things conditional on version
> 
> Honestly, @Sinan81 wrote most of this file, so I'm hoping he can provide a more useful review.

The current logic seems fine to me. I don't find it convoluted.

Sinan81:
Built default version 3.40 and also 3.44 from scratch. looks good. will approve right after the above comment is addressed.

repos/spack_repo/builtin/packages/qgis/package.py
Chrismarsh: Good idea, thanks

Chrismarsh:
One weird thing was that spack wouldn't find a new version until I updated the url. But the url scheme hasn't changed (as far as I could tell??), so this isn't a case where `url_for_version` is required. Any ideas?

Sinan81:
> One weird thing was that spack wouldn't find a new version until I updated the url. But the url scheme hasn't changed (as far as I could tell??), so this isn't a case where `url_for_version` is required. Any ideas?

no idea. Definitely, there's something weird going on. I did a test where I restored original url statement, and tried installing version 3.44.8 (yep, it's released half a day ago lol). 
initially, `spack install ...` just hung, Then I tried again with debug option. it worked fine. Then I cancelled the install and tried again, twice, where spack couldn't fetch the file from qgis downloads webpage. This could simply be a glitch with qgis webpage. 

We can keep the url statement as it is, and merge.

Sinan81 (approved):

Chrismarsh:
@Sinan81, sorry I just invalided your review as I added a comment to describe the url behaviour. @haampie pointed out to me on Slack that it is because qgis uses a json to list the urls which the `checksum` crawler won't discover.

Sinan81 (approved):

Chrismarsh:
It looks like the just-added 3.44.8 is a LTR. Given this package `prefers` the LTR releases, I think we should add 3.44.8 to this PR (kinda while we are here). So I will do that

Chrismarsh:
Well this isn't clear to me at all:

- QGis website gives 3.44.8 for the LTR version on windows(`https://download.qgis.org/downloads/QGIS-OSGeo4W-3.44.8-1.msi`)
- 3.44.8 added the LTR badge on the splash (https://github.com/qgis/QGIS/compare/final-3_44_7%5E...final-3_44_8#diff-905d384c32f55ba6626d1043f78a3f3873589174e703f10b1a0702dce1ad3ee2)
-  Roadmap seems to suggest 3.44.8 is actually a LR and 3.44.12 is now considered the LTR? (https://qgis.org/resources/roadmap/)
But 3.44 is a LTR branch (https://github.com/qgis/QGIS-Website/issues/755) so maybe it's fine to have 3.44.8 listed as the `preferred=True` in spack until 3.44.12 is released?

Sinan81:
yes, it's pretty confusing. as far as I can tell 3.44.8 is an LTR. I found the following explanation

> In the first four months after its release, a new LTR is also the current LR. In this phase, the new LTR doesn’t replace the previous LTR in the LTR repositories. This happens as soon as a new LR is released.

in https://qgis.org/resources/roadmap/

Sinan81:
>so maybe it's fine to have 3.44.8 listed as the preferred=True in spack until 3.44.12 is released?

by now, I'm convinced that 3.44.8 is an LTR. please go ahead and make it prefferred.

Sinan81:
Could we also add a version constraint for proj like:
```
depends_on("proj@:9.5", when="@3.34.5:3.34.15")
```
only tested this on 3.34.15 but I'm sure older versions will fail to build with proj@9.6:

repos/spack_repo/builtin/packages/qgis/package.py
```diff
@@ -14,18 +16,22 @@ class Qgis(CMakePackage):
...
 
     license("GPL-2.0-or-later")
 
     # Prefer latest LTR
-    version(
-        "3.40.6",
-        sha256="dd68d39a2a29326031195bed2125e8b0fa7112fe9ee74d5f9850d06b02cef6a8",
+    version("3.44.8",
+        sha256="146e197f34f1f9ede8cfdf5b9cc4d76667771720a302172c32d4117367356e96",
         preferred=True,
```
adamjstewart, line 32: Spack already prefers the newest version by default, maybe we could remove `preferred=True` since there are no non-LTR versions?

Sinan81: well. version 4.0.0 is just released. we might add a non-LTR 4.x.x version to the package file soon. So, I'm inclined to keep the preferred thing.

Sinan81 (approved):

repos/spack_repo/builtin/packages/qgis/package.py
Chrismarsh: What is the motivation of preferring the LTR anyway? I feel like user who wants a LTR would know what version they wanted anyway?

Sinan81: If version() had a comment option (so that LTR version can be indicated as a comment and appeared in `spack info qgis` ) I would be ok with removing preferred option. Of course, most users can do `spack edit qgis` however here I think we're targeting clueless users and/or setting a sensible default.

Sinan81:
@spackbot fix style

repos/spack_repo/builtin/packages/qgis/package.py
Sinan81: If both of you want to remove it, please go ahead.

Sinan81 (approved):

repos/spack_repo/builtin/packages/qgis/package.py
Chrismarsh: I have no strong preference either way. I just wasn't sure if there was a specific reason you'd added that preference. I find QGIS generally pretty stable on non LTR. However, I think we leave it as is for the moment, as it's a sane default for when we add Qgis 4.x. I'm unsure how much will break with that version.

adamjstewart: I'm fine either way, if no one has strong preferences then let's keep the status quo.

adamjstewart (approved):

Sinan81:
@Chrismarsh thanks for keeping the spack package up to date!

PR #3634, opened 2026-03-03
cached cmake: check %cuda not ^cuda

alalazo:
If cuda is not a dependency of the root, but is used by any transitive link/run dependency, then errors like [this one](https://gitlab.spack.io/spack/spack-packages/-/jobs/20855918) are triggered.


rbberger (approved):
Seems like there might be more of these floating around for MPI as well.

PR #3637, opened 2026-03-03
py-illumina-utils: added version 2.14

V-Karch:
Adding a new version for `py-illumina-utils`
Url format changed on pypi so I added a `url=`

Alternatively if it's better for the style we could have a `url_for_version` but I'm not sure it's necessary


RikkiButler20 (approved):
LGTM. Thanks

PR #3641, opened 2026-03-04
dd4hep and dependencies: add cxxstd=23

paulgessinger:
Except for xerces-c, I set lower version bounds based on which versions I tested.

### DIFF
--- a/repos/spack_repo/builtin/packages/clhep/package.py
+++ b/repos/spack_repo/builtin/packages/clhep/package.py
@@ -61,7 +61,8 @@ class Clhep(CMakePackage):
             "14",
             conditional("17", when="@2.3.4.3:"),
             conditional("20", when="@2.4.6.4:"),
-            "23",
+            # Assuming versions not supporting C++20 do not support C++23
+            conditional("23", when="@2.4.6.4:"),
         ),
         multi=False,
         description="Use the specified C++ standard when building.",
--- a/repos/spack_repo/builtin/packages/clhep/package.py
+++ b/repos/spack_repo/builtin/packages/clhep/package.py
@@ -61,6 +61,7 @@ class Clhep(CMakePackage):
             "14",
             conditional("17", when="@2.3.4.3:"),
             conditional("20", when="@2.4.6.4:"),
+            conditional("23", when="@2.4.7.2:"),
         ),
         multi=False,
         description="Use the specified C++ standard when building.",
--- a/repos/spack_repo/builtin/packages/clhep/package.py
+++ b/repos/spack_repo/builtin/packages/clhep/package.py
@@ -61,7 +61,7 @@ class Clhep(CMakePackage):
             "14",
             conditional("17", when="@2.3.4.3:"),
             conditional("20", when="@2.4.6.4:"),
-            conditional("23", when="@2.4.7.2:"),
+            "23",
         ),
         multi=False,
         description="Use the specified C++ standard when building.",
--- a/repos/spack_repo/builtin/packages/geant4/package.py
+++ b/repos/spack_repo/builtin/packages/geant4/package.py
@@ -61,7 +61,7 @@ class Geant4(CMakePackage):
         conditional("11", "14", when="@:10"),
         conditional("17", when="@10.4.1:"),
         conditional("20", when="@10.7.0:"),
-        conditional("23"),
+        conditional("23", when="@11:"),
     )
     variant(
         "cxxstd",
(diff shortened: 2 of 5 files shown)


sethrj (changes requested):
I think version requirements need to be "accurate", not just a sampled range. Sampling newer versions leads to artificially tight constraints that could unnecessarily prevent a solution. It's confusing to maintainers because it's impossible to differentiate in the code from "I tried version X and assume newer versions work too" from "the code [project.toml, cmakelists] explicitly specifies versions X through Y are required".

If you've established any of these versions as lower bounds (e.g. release notes saying "now supports C++23" or older versions giving a hardcoded list of options) then please leave them, otherwise remove the version requirements.

Thanks for the update!

paulgessinger:
@sethrj I don't think this is general practice. I can drop the version constraints altogether.

repos/spack_repo/builtin/packages/geant4/package.py
```diff
@@ -61,6 +61,7 @@ class Geant4(CMakePackage):
         conditional("11", "14", when="@:10"),
         conditional("17", when="@10.4.1:"),
         conditional("20", when="@10.7.0:"),
+        "23",
```
drbenmorgan: Following up on @sethrj's comment, the constraints on Geant4's C++23 support are:

- Geant4 >= 11.0.0
- CMake >= 3.20.0 (I'm a bit out the loop, so not sure if Spack's C++23 checks enforce this)

paulgessinger: I couldn't find anything specifying this, but I'll put `@11:` in as a constraint then.

repos/spack_repo/builtin/packages/vecgeom/package.py
```diff
@@ -61,7 +61,12 @@ class Vecgeom(CMakePackage, CudaPackage):
     )
     version("1.1.20", sha256="e1c75e480fc72bca8f8072ea00320878a9ae375eed7401628b15cddd097ed7fd")
 
-    _cxxstd_values = (conditional("11", "14", when="@:1.1"), "17", conditional("20", when="@1.2:"))
+    _cxxstd_values = (
+        conditional("11", "14", when="@:1.1"),
+        "17",
+        conditional("20", when="@1.2:"),
+        "23",
```
sethrj: If 20 is prohibited before 1.2, then 23 will be as well.

repos/spack_repo/builtin/packages/clhep/package.py
```diff
@@ -61,7 +61,7 @@ class Clhep(CMakePackage):
             "14",
             conditional("17", when="@2.3.4.3:"),
             conditional("20", when="@2.4.6.4:"),
-            conditional("23", when="@2.4.7.2:"),
+            "23",
```
sethrj: To revise my earlier statement: even if we don't know whether `2.4.7.0` works or not, we know that since `20` is incompatible before `2.4.6.x` then  `23` will be incompatible. Omitting this could lead to weird concretization errors that select cxxstd=23 for old versions.

sethrj (changes requested):
Sorry for the confusion and back-and-forth, but I think we need to put *some* constraints if the older versions are similarly constrained. I've tried to express this in a documentation update: https://github.com/spack/spack/pull/52022


> Always specify version ranges with an open-world assumption:
> 
> - all "ground truths" about exclusions and inclusions (e.g., versions with features added or removed) must satisfy the range, and
> - no potential but unknown versions are excluded from the range.
> 
> This practice avoids overconstraining version ranges, which can lead to concretization errors, and ensures that every version in a package is *meaningful* and not just *incidental*.
> If and when future versions are known incompatible, the version range should be constrained with an upper bound.

jmcarcell:
Good from the DD4hep side, it has been built with C++23 for the LCG stacks with GCC 15 which have been around for several months (close to one year probably).

paulgessinger:
@sethrj happy to change this but I'm not sure I quite understand what I'm supposed to put in now. Copy the constraints from what was in place for the `20` value (assuming if versions weren't `20` compatible they can't be `23` compatible)?

sethrj:
@paulgessinger Yes, exactly. That's the only way to express what we know:
- Versions older than X will *not* support 23
- Some newish version Y *will*  support 23
- Versions newer than Y are assumed to support 23

While not excluding what we don't know:
- Whether versions between X and Y support 23

paulgessinger:
@sethrj added version bounds back to clhep and vecgeom.

sethrj (approved):
Thanks!

PR #3645, opened 2026-03-04
DDC: Add version 0.11.0

tpadioleau:
Also fixes the upper bounds for Kokkos and Kokkos Kernels.

https://github.com/CExA-project/ddc/releases/tag/v0.11.0

### DIFF
--- a/repos/spack_repo/builtin/packages/ddc/package.py
+++ b/repos/spack_repo/builtin/packages/ddc/package.py
@@ -21,6 +21,7 @@ class Ddc(CMakePackage):
     license("MIT", checked_by="tpadioleau")
 
     version("main", branch="main", no_cache=True)
+    version("0.11.0", sha256="c3ee616cc6cbbf417dade247cd49805b1a5422b4ac3539cb954f608b8ea27cf4")
     version("0.10.0", sha256="0ab717a21c641b59af8119ff665c0322498fcccf5f49b2c3a2746eecbf1a4964")
     version("0.9.0", sha256="e975a19f2d8e4fc668ab7628e145b927987812496c94b384ee9e72d054711078")
     version("0.8.0", sha256="6c6d28f1d406e1417021f88d748829cae0afce2cb3714cf82fd3f4cd3b7b91b4")
@@ -42,9 +43,12 @@ class Ddc(CMakePackage):
     depends_on("cmake@3.22:3", type="build", when="@:0.8")
     depends_on("cmake@3.25:4", type="build", when="@0.9:")
 
-    depends_on("kokkos@4.4.1:4")
+    depends_on("kokkos@4.4.1:4", when="@:0.8")
+    depends_on("kokkos@4.4.1:5", when="@0.9:")
 
     with when("+fft"):
+        depends_on("kokkos-fft@0.3:0", when="@:0.10")
+        depends_on("kokkos-fft@0.3:1", when="@0.11:")
         for variant, backend in [
             ("~openmp", "host_backend=fftw-serial"),
             ("+openmp", "host_backend=fftw-openmp"),
@@ -52,11 +56,12 @@ class Ddc(CMakePackage):
             ("+rocm", "device_backend=hipfft"),
             ("+sycl", "device_backend=onemkl"),
         ]:
-            depends_on(f"kokkos-fft@0.3:0 {backend}", when=f"^kokkos {variant}")
+            depends_on(f"kokkos-fft {backend}", when=f"^kokkos {variant}")
 
     with when("+splines"):
         depends_on("ginkgo@1.8:1")
-        depends_on("kokkos-kernels@4.5.1:4")
+        depends_on("kokkos-kernels@4.5.1:5", when="@:0.8")
+        depends_on("kokkos-kernels@4.5.1:5", when="@0.9:")
         depends_on("lapack")
 
         for arch in CudaPackage.cuda_arch_values:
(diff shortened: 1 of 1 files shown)


tpadioleau:
cc @tretre91

repos/spack_repo/builtin/packages/ddc/package.py
```diff
@@ -42,21 +43,25 @@ class Ddc(CMakePackage):
     depends_on("cmake@3.22:3", type="build", when="@:0.8")
     depends_on("cmake@3.25:4", type="build", when="@0.9:")
 
-    depends_on("kokkos@4.4.1:4")
+    depends_on("kokkos@4.4.1:4", when="@:0.8")
+    depends_on("kokkos@4.4.1:5", when="@0.9:")
 
     with when("+fft"):
+        depends_on("kokkos-fft@0.3:0", when="@:0.10")
+        depends_on("kokkos-fft@0.3:1", when="@0.11:")
```
alalazo: Do you want to separate lower bounds from upper bounds? That would be:
```suggestion
        depends_on("kokkos-fft@0.3:")
        depends_on("kokkos-fft@:0", when="@:0.10")
        depends_on("kokkos-fft@:1", when="@:0.11")
```
Similarly for other constraints. That is usually easier to maintain in time.

repos/spack_repo/builtin/packages/ddc/package.py
```diff
@@ -42,21 +43,25 @@ class Ddc(CMakePackage):
...
             ("+rocm", "device_backend=hipfft"),
             ("+sycl", "device_backend=onemkl"),
         ]:
-            depends_on(f"kokkos-fft@0.3:0 {backend}", when=f"^kokkos {variant}")
+            depends_on(f"kokkos-fft {backend}", when=f"^kokkos {variant}")
 
     with when("+splines"):
         depends_on("ginkgo@1.8:1")
-        depends_on("kokkos-kernels@4.5.1:4")
+        depends_on("kokkos-kernels@4.5.1:5", when="@:0.8")
```
tretre91:
```suggestion
        depends_on("kokkos-kernels@4.5.1:4", when="@:0.8")
```

repos/spack_repo/builtin/packages/ddc/package.py
tretre91: I agree it feels a bit easier to read

If we were to bump the lower bound in a future release, for example suppose v0.13 only supports kokkos-ftt >= 1.0, would it look something like this?
```py
depends_on("kokkos-fft@0.3:", when="@:0.12")
depends_on("kokkos-fft@1.0:", when="@:0.13")

depends_on("kokkos-fft@:0", when="@:0.10")
depends_on("kokkos-fft@:1", when="@:0.13")
```
instead of
```py
depends_on("kokkos-fft@0.3:0", when="@:0.10")
depends_on("kokkos-fft@0.3:1", when="@0.11:0.12")
depends_on("kokkos-fft@1.0:1", when="@0.13:")
```

repos/spack_repo/builtin/packages/ddc/package.py
tpadioleau: thx!

repos/spack_repo/builtin/packages/ddc/package.py
tpadioleau: We are open to suggestions about maintainability so we can give it a try.

@tretre91 What about something like
```python
depends_on("kokkos-fft@0.3:")
depends_on("kokkos-fft@1:", when="@0.13:")

depends_on("kokkos-fft@:0", when="@:0.10")
depends_on("kokkos-fft@:1")
```

Then if we add v0.14 that does not update the requirements, no need to change the upper bound in the clause `when`.

tretre91: Yes, we can go for something like this

repos/spack_repo/builtin/packages/ddc/package.py
```diff
@@ -39,24 +40,34 @@ class Ddc(CMakePackage):
 
     depends_on("cxx", type="build")
 
-    depends_on("cmake@3.22:3", type="build", when="@:0.8")
-    depends_on("cmake@3.25:4", type="build", when="@0.9:")
+    depends_on("cmake@3.25:", type="build", when="@0.9:")
+    depends_on("cmake@3.22:", type="build")
+    depends_on("cmake@:4", type="build")
+    depends_on("cmake@:3", type="build", when="@:0.8")
```
tretre91, line 46: I would put it the other way around, the oldest versions first. Same for the other dependencies
```suggestion
    depends_on("cmake@3.22:", type="build")
    depends_on("cmake@3.25:", type="build", when="@0.9:")
    depends_on("cmake@:3", type="build", when="@:0.8")
    depends_on("cmake@:4", type="build")
```

tpadioleau: I was wondering what would be best, I followed the list of DDC versions we declare.

tpadioleau:
Actually now that DDC is not header-only anymore, the lapacke.h needs to be found to compile. The script `FindLAPACKE.cmake` fails to find it and compilation fails.

tpadioleau:
@alalazo I have two questions:
- what do you think of the current separation of lower and upper bounds ?
- is it possible to add this package in an existing stack ? If so, which one ?

alalazo (approved):
LGTM. As for the pipeline, can this be added to `e4s`? @eugeneswalker  Would that be possible?

repos/spack_repo/builtin/packages/ddc/package.py
alalazo: Let me know when you want to merge

tpadioleau:
@alalazo ok for me

repos/spack_repo/builtin/packages/ddc/package.py
tpadioleau: @tretre91 do you find it blocking or we can merge ?

tpadioleau:
> LGTM. As for the pipeline, can this be added to `e4s`? @eugeneswalker Would that be possible?

Thank you! I will open a PR as soon as we fix the build with CUDA.

repos/spack_repo/builtin/packages/ddc/package.py
tretre91: It's not blocking

tretre91 (approved):
LGTM. The order of the `depends_on` can be discussed or changed later if needed

alalazo (approved):

PR #3646, opened 2026-03-04
qt: add v5.15.18

wdconinc:
This PR adds `qt`, v5.15.18, [diff](https://github.com/qt/qtbase/compare/v5.15.17-lts-lgpl...v5.15.18-lts-lgpl), [release notes](https://code.qt.io/cgit/qt/qtreleasenotes.git/tree/qt/5.15.18/release-note.md). Bugfix/maintenance release. Did not notice any major dependency changes (as often is the case, the vendored dependencies are updated, but this does not imply an updated requirement). This package is built in CI.

Note: I have not built this locally yet but it will get built in CI.

### DIFF
--- a/repos/spack_repo/builtin/packages/qt/package.py
+++ b/repos/spack_repo/builtin/packages/qt/package.py
@@ -32,6 +32,7 @@ class Qt(Package):
 
     license("LGPL-3.0-only")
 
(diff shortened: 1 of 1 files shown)


RikkiButler20 (approved):
Looks good to me.

PR #3650, opened 2026-03-04
ci: include remote file for per-package resource requests

zackgalbreath:
The existing `linux/ci.yaml` only specifies resource requests for a small subset of the packages we build in CI. The practical consequence of this is that we overprovision cloud runners for the majority of GitLab CI rebuild jobs.

This commit chooses to host more complete resource usage data in a separate repository to avoid bloating the history of spack-packages over time.

### DIFF
--- a/.ci/gitlab/include.yaml
+++ b/.ci/gitlab/include.yaml
@@ -41,5 +41,13 @@ include:
   - path: configs/linux
     when: platform == "linux"
 
+  # === Per-package resource requests (Only used for Linux on Kubernetes) ===
+  - when: platform == "linux"
+    git: https://gitlab.spack.io/spack/resource-requests
+    branch: main
+    paths:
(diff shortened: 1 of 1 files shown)


zackgalbreath:
Evidence that this change works as intended:
https://gitlab.spack.io/spack/spack-packages/-/jobs/20872636#L5

Notice that this package is tagged as `small` and it is reporting a non-default resource request (0.25G instead of 16G). This is despite the fact that `perl-www-robotrules` is not explicitly listed in [linux/ci.yaml](https://github.com/spack/spack-packages/blob/3ba9f2c4d8411e5a1fca4b43a105ebb1afbbad31/.ci/gitlab/configs/linux/ci.yaml).

kwryankrattiger (approved):
LGTM

PR #3651, opened 2026-03-04
compadre: add v1.7.0

kuberry:


repos/spack_repo/builtin/packages/compadre/package.py
```diff
@@ -68,22 +74,25 @@ def cmake_args(self):
...
         )
 
         if spec.variants["debug"].value == "0":
-            if spec.satisfies("^kokkos~cuda"):
-                options.append(
-                    "-DCMAKE_CXX_FLAGS:STRING=%s"
-                    % "' -Ofast -funroll-loops -march=native -mtune=native '"
-                )
+            if spec.satisfies("@:1.6"):
+                if spec.satisfies("^kokkos~cuda"):
```
alalazo: We can invert to avoid excessive nesting:
```suggestion
if spec.satisfies("@1.7:"):
    <else-branch>
elif spec.satisfies("^kokkos~cuda"):
    <...>
```

repos/spack_repo/builtin/packages/compadre/package.py
```diff
@@ -68,22 +74,25 @@ def cmake_args(self):
...
+                    )
+            else:
+                options.append("-DCMAKE_BUILD_TYPE:STRING=Release")
             options.append("-DCompadre_DEBUG:BOOL=OFF")
         else:
-            options.append("-DCMAKE_CXX_FLAGS:STRING='-g -O0'")
-            options.append("-DCMAKE_BUILD_TYPE:STRING=DEBUG")
+            if spec.satisfies("@:1.6"):
+                options.append("-DCMAKE_CXX_FLAGS:STRING='-g -O0'")
+            options.append("-DCMAKE_BUILD_TYPE:STRING=Debug")
```
alalazo: The `debug` variant is highly overlapping and in conflict with the `build_type` variant from `CMakePackage`. Can we fix this?

alalazo (changes requested):

repos/spack_repo/builtin/packages/compadre/package.py
alalazo: I'm mainly concerned about the comment :point_up:

kuberry: The debug variant controlled only the CMake variable Compadre_DEBUG prior to 1.7.0, which had logic that was too independent of the CMAKE_BUILD_TYPE. This has been fixed in 1.7.0, which is why we can now rely on only setting CMAKE_BUILD_TYPE.

alalazo: One idea may be to add:
```
requires("build_type=debug", when="debug=1")
requires("build_type=debug", when="debug=2")
```
among the directives at the top, and let the `CMakePackage` base class deal with `-DCMAKE_BUILD_TYPE`.

repos/spack_repo/builtin/packages/compadre/package.py
kuberry: Thank you for this. I've refactored the logic.

repos/spack_repo/builtin/packages/compadre/package.py
kuberry: I've added build_type as a variant with a default of RelWithDebInfo and rewritten the logic to only be based on build_type. Also, I deprecated previous versions and will remove their conditional dependencies and configuration options in a future release.

kuberry:
The recipe here is not sufficient for compiling on CUDA and HIP. I'm going to rework it and then update the PR.

kuberry:
@spackbot re-run pipeline

alalazo (approved):

PR #3196, opened 2026-01-29
GoPackage: disable CGO support by default for better portability

alecbcs:
Fixes #3194.

Updates the `GoPackage` builder to disable CGO support by default to improve the portability of go programs by reducing C dependencies. Exposes a `cgo_enabled` property to packages to enable CGO functionality selectively.

PR #3655, opened 2026-03-04
gocryptfs: set cgo_enabled = True

snehring:
#3196 disabled cgo by default, gocryptfs needs cgo to call out to openssl routines.


alecbcs (approved):

PR #3660, opened 2026-03-05
All CI jobs should use the same misc cache

scheibelp:
At one point the misc cache was updated to be "per-spack": placed in a directory that was based on a hash of the spack prefix itself. This was intended to help different clones of spack avoid conflicting with one another.

It caused a slowdown with CI though, so was reverted in https://github.com/spack/spack/pull/50920 (https://github.com/spack/spack/pull/50919 also has a useful explanation of the issue).

https://github.com/spack/spack/pull/47615 intends to reintroduce this as default behavior, this is my attempt to get all pipelines to override that default and use a shared misc cache (and continue to do so after that PR merges).

### DIFF
--- a/.ci/gitlab/config.yaml
+++ b/.ci/gitlab/config.yaml
@@ -1,0 +1,2 @@
+config:
+  misc_cache: "$user_cache_path/cache"


haampie:
This shouldn't be necessary:

https://github.com/spack/spack-packages/blob/ddc6fd99ca03708d58f1746ed613e34ea4a0c813/.ci/gitlab/.gitlab-ci.yml#L5

https://github.com/spack/spack-packages/blob/ddc6fd99ca03708d58f1746ed613e34ea4a0c813/.ci/gitlab/.gitlab-ci.yml#L115-L121

(the artifacts: ... bit can probably be dropped now that creating the misc cache is much faster)

scheibelp:
> This shouldn't be necessary:
> SPACK_USER_CACHE_PATH: "${CI_PROJECT_DIR}/tmp/_user_cache/" 

`misc_cache` was (and will be in https://github.com/spack/spack/pull/47615) defined as a subdir of `SPACK_USER_CACHE_PATH`/`$user_cache_path`, e.g. before it was 

```
# In etc/spack/defaults/base/config.yaml
  misc_cache: $user_cache_path/cache
```

and is intended to be 

```
# In etc/spack/defaults/base/config.yaml
  misc_cache: $user_cache_path/$spack_instance_id/cache
  # actually it will be
  # misc_cache: $state_home/$spack_instance_id/cache
  # but $user_cache_path/$state_home are aliases of one another in that PR
```

`SPACK_USER_CACHE_PATH` is a base dir, so setting it would not avoid this problem.

>  artifacts: 
>   paths: 

If I understand, these are read paths ("go to the listed locations and bundle them up"), without this PR, there will be errors like https://gitlab.spack.io/spack/spack/-/jobs/20317105 (on https://github.com/spack/spack/pull/47615/commits/898e9321c04474c509ddbe2ef0b6ad7cfb90c2b5 in https://github.com/spack/spack/pull/47615)

becker33 (approved):

PR #3672, opened 2026-03-05
omegah: add exodus variant

Sichao25:
This PR adds exodus variant to omega_h package.

### DIFF
--- a/repos/spack_repo/builtin/packages/omega_h/package.py
+++ b/repos/spack_repo/builtin/packages/omega_h/package.py
@@ -68,6 +68,7 @@ class OmegaH(CMakePackage, CudaPackage):
     variant("mpi", default=True, description="Activates MPI support")
     variant("zlib", default=True, description="Activates ZLib support")
     variant("trilinos", default=True, description="Use Teuchos and Kokkos")
+    variant("exodus", default=False, description="Use SEACASExodus")
     variant("throw", default=False, description="Errors throw exceptions instead of abort")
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/omega_h/package.py
```diff
@@ -68,6 +68,7 @@ class OmegaH(CMakePackage, CudaPackage):
     variant("mpi", default=True, description="Activates MPI support")
     variant("zlib", default=True, description="Activates ZLib support")
     variant("trilinos", default=True, description="Use Teuchos and Kokkos")
+    variant("exodus", default=False, description="Use SEACASExodus")
```
cwsmith:
```suggestion
    variant("exodus", default=False, description="Enable use of ExodusII meshes")
```
I'm not sure many people will know what 'SEACAS' is.

cwsmith: We may also want to set a min required version.  I'm thinking 10.8.6-scorec as it will likely work and isn't just the latest.  I'm sure support goes back further than that... but I don't think users will care if it is exact.

cwsmith (changes requested):
Looks good. I left a few comments below.

Sichao25:
@spackbot fix style

cwsmith (approved):
LGTM.  Thank you.

alalazo:
@spackbot run pipeline

alalazo (approved):

PR #3676, opened 2026-03-05
Add style tools to the developer stack

johnwparent:
Add all of the tools required to run Spack style checks as well as ruff and ty, which will be added to the style checks in upcoming PRs, to the spack developer build stacks for binary mirror population.

Motivation: 

Support bootstrapping style checks from the binary mirror for a faster, ruff/ty enabled dev bootstrap.

Ruff and ty are going to be added to Spack style. Ruff and ty builds from source require a build of rust, which is expensive from both a time and disc space perspective. Bootstrapping from source, as we currently do with the style checks when dependencies are unresolved, would be a poor ux if users had to wait for a build of rust (which builds llvm) to run spack style, and for users with limited home drive storage, infeasible for a simple style check. 

Ruff and ty, will instead need to be pulled from the public binary mirror during a dev bootstrap to support the style command. To do so, we need to add them to the stacks, and what better stack than the dev stack. 

As dev boostrap is now incorporating the public binary mirror, we should also populate it with the other dev bootstrap so we can avoid more from source dev bootstrapping operations.

### DIFF
--- a/repos/spack_repo/builtin/packages/py_ruff/package.py
+++ b/repos/spack_repo/builtin/packages/py_ruff/package.py
@@ -33,10 +33,12 @@ class PyRuff(PythonPackage):
     version("0.3.0", sha256="0886184ba2618d815067cf43e005388967b67ab9c80df52b32ec1152ab49f53a")
     version("0.1.6", sha256="1b09f29b16c6ead5ea6b097ef2764b42372aebe363722f1605ecbcd2b9207184")
 
-    depends_on("c", type="build")
-    depends_on("gmake", type="build")
+    
 
     with default_args(type="build"):
+        depends_on("c")
+        depends_on("gmake")
+
         depends_on("py-maturin@1.9:1", when="@0.12.7:")
         depends_on("py-maturin@1")
 
--- a/repos/spack_repo/builtin/packages/py_ruff/package.py
+++ b/repos/spack_repo/builtin/packages/py_ruff/package.py
@@ -33,8 +33,6 @@ class PyRuff(PythonPackage):
     version("0.3.0", sha256="0886184ba2618d815067cf43e005388967b67ab9c80df52b32ec1152ab49f53a")
     version("0.1.6", sha256="1b09f29b16c6ead5ea6b097ef2764b42372aebe363722f1605ecbcd2b9207184")
 
-    
-
     with default_args(type="build"):
         depends_on("c")
         depends_on("gmake")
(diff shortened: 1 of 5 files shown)


trws:
I should have thought to put all of these in here to begin with, and am all for it.

johnwparent:
> I should have thought to put all of these in here to begin with, and am all for it.

I think we're actually going to drop support for isort, black, flake and mypy in favor of ruff and ty, so I plan on pulling all but ty, ruff, and pytest out of this stack, unless you think it would be valuable to still provide these tools? (in which case I'll remove the strict version requirements and put them elsewhere in this yaml)

johnwparent:
Gitlab failures are legit ty failures, investigating now.

adamjstewart (dismissed):
Do we have any mechanism to bump the version pins over time? Maybe we should always use the latest version (I already updated ty to 0.0.20).

johnwparent:
@spackbot fix style

johnwparent:
> Do we have any mechanism to bump the version pins over time? Maybe we should always use the latest version (I already updated ty to 0.0.20).

Nothing automated or mechanical, I chose the current version pins based on the latest available versions of each (looks like I caught you between adding 0.015 and 0.016 for ty), so I'll update them to the latest available now. 

I think if they're updated in the package recipe, we should be bumping the stack versions as well, and then bumping the bootstrap specs once there's a package repo release with the new version. Probably best enforced by policy.

johnwparent:
@spackbot fix style

adamjstewart:
> I think if they're updated in the package recipe, we should be bumping the stack versions as well

If you don't pin the versions, this is exactly what will happen automatically 🙂

johnwparent:
> > I think if they're updated in the package recipe, we should be bumping the stack versions as well
> 
> If you don't pin the versions, this is exactly what will happen automatically 🙂

Right, but given we're using this stack to bootstrap, and the ruff/ty bootstrap will be cache-only, I think these should be set explicitly (at least for now) so we have a better guarantee of bootstrap success and an expectation of ruff/ty behavior. I could be wrong though.

kwryankrattiger:
> Do we have any mechanism to bump the version pins over time? Maybe we should always use the latest version (I already updated ty to 0.0.20).

I think for this stack in particular we want to keep versions pinned. The primary purpose of this is to populate a bootstrap cache which needs to follow what we are using for style in Spack itself. And we will want to time releases with updates to that, which may not always go to the next latest release.

kwryankrattiger:
Version pinning aside, is there any opposition to merge this so these can be in the packages release?

adamjstewart (approved):

PR #3677, opened 2026-03-06
ww3: New package with most default switches working

rsoutelino:
Added spack package for WW3 wave model. 

- Tested several available default switches with intel oneapi 2025
- Ran spack style tests and unit tests, all passed.


repos/spack_repo/builtin/packages/ww3/package.py
```diff
@@ -0,0 +1,126 @@
+from spack_repo.builtin.build_systems.cmake import CMakePackage
```
wdconinc, line 1: Missing spdx header (see other pkgs).

repos/spack_repo/builtin/packages/ww3/package.py
```diff
@@ -0,0 +1,126 @@
...
+class Ww3(CMakePackage):
+    """WAVEWATCH III (WW3) is a community wave modeling framework that includes
+    the latest scientific advancements in the field of wind-wave modeling and dynamics.
+
+    Example installation:
+        spack install ww3@develop switch=Ifremer1 %intel-oneapi-compilers@2025.3.2
+    """
+
+    homepage = "https://github.com/NOAA-EMC/WW3"
+    url = "https://github.com/NOAA-EMC/WW3"
```
wdconinc: This url normally points to a tarball, e.g. https://github.com/NOAA-EMC/WW3/archive/refs/tags/6.07.1.tar.gz

repos/spack_repo/builtin/packages/ww3/package.py
```diff
@@ -0,0 +1,126 @@
...
+    git = "https://github.com/metocean/WW3.git"
```
wdconinc: We can't use a non-standard git repository as a reference location for source of a different project that is still maintained and active, for reasons of supply chain security and trademark/representation.

```suggestion
    git = "https://github.com/NOAA-EMC/WW3.git"
```

There are ways for users to specify a different git repository as a local attribute.

repos/spack_repo/builtin/packages/ww3/package.py
```diff
@@ -0,0 +1,126 @@
...
+
+    maintainers("rsoutelino")
+
+    license("LGPL-3.0-only", checked_by="rsoutelino")
+
+    # Development version - use with: spack install ww3@develop
+    version("develop", branch="develop")
```
wdconinc, line 23: One way to get a custom repository is:

```suggestion
    version("develop", branch="develop")
    version("develop-metocean", branch="develop", git="https://github.com/metocean/WW3.git")
```
(suitably reflowed with spack style)

repos/spack_repo/builtin/packages/ww3/package.py
```diff
@@ -0,0 +1,126 @@
...
+
+    depends_on("c", type="build")
+    depends_on("fortran", type="build")
```
wdconinc, line 67: Add CMake min version too.

repos/spack_repo/builtin/packages/ww3/package.py
```diff
@@ -0,0 +1,126 @@
...
+    # depends_on("bacio", when="switch=NCEP_glwu")
+    # depends_on("w3emc", when="switch=NCEP_glwu")
+
+    # depends_on("scotch", when="switch=NCEP_rwps")
+
+    # ESMF - Required when MULTI_ESMF in switch
+    depends_on("esmf", when="switch=multi_esmf")
+
+    # WW3 configuration variants
+    variant(
```
wdconinc: Variants typically defined before depends_on so the logical flow allows dependencies to use variants defined before (for a python class it doesn't matter but it helps the humans who, for now at least, have to read this file).

wdconinc (changes requested):
Looks good. Changes requested on the git url since we can't ship with a default url pointing to a downstream project. Provided a suggestion for how to include your metocean repo as a version too. Would suggest adding one tagged version in addition to the moving target branches.

repos/spack_repo/builtin/packages/ww3/package.py
rsoutelino: Thanks. That was actually a leftover from some local tests, I meant to replace `metocean` by `NOAA-EMC` but forgot.

repos/spack_repo/builtin/packages/ww3/package.py
rsoutelino: This import was somehow not working in my spack version, but looks OK now.

repos/spack_repo/builtin/packages/ww3/package.py
rsoutelino: Yes, I was trying to do that at the start, but the main developers of WW3 stopped creating git tags, and hence the tar balls are not available for newer versions. I thought I'd use the git attribute to let people choose via commit hash or installing `develop` version. I'm happy to go with the approach you recommend.

repos/spack_repo/builtin/packages/ww3/package.py
rsoutelino: As above, that `metocean` repo was left there unintentionally; it was just one of my local tests. We can keep that out so it stays clean and simple. Thanks!

repos/spack_repo/builtin/packages/ww3/package.py
rsoutelino: Sounds good, thanks!

repos/spack_repo/builtin/packages/ww3/package.py
rsoutelino: Ok, will add 3.19 as per: https://github.com/NOAA-EMC/WW3/blob/develop/CMakeLists.txt#L4

repos/spack_repo/builtin/packages/ww3/package.py
rsoutelino: The last one is outdated, from 2019:

https://github.com/NOAA-EMC/WW3/tags

I've changed the url to the latest tarball, as I guess people can always install the `develop` version, but I was wondering that can get confusing to people, as in the wave modelling community, it's common to refer to ww3 version 7.x, etc. They stopped tagging in git, but I think they are stating the version in other artifacts, such as https://github.com/NOAA-EMC/WW3/blob/develop/VERSION. 

Happy to any suggestions here :). Thanks

wdconinc: I'd suggest defining versions by git commit then, but keeping some kind of 'sensible' versioning scheme that allows for the main developers to start releasing tags again. E.g. based on `git describe --tags` you could get:
```py
    version("6.07.1-390-g3e591453", commit="3e591453c9d95bb837f87b578e10f79f9c5f0952")
```
Then if `6.08` were to ever get released, it will be more recent than any versions added here. Also ordering between consecutive commits is ensured by the (in this case) serial number 390.

rsoutelino: Thanks for the suggestion. Sounds like a good idea. I added the latest version defined in their VERSION artifact plus git commit hash.

repos/spack_repo/builtin/packages/ww3/package.py
```diff
@@ -0,0 +1,129 @@
...
+    # NOTE: NCEP switches are not currently supported due to CMake target export
+    # issues in Spack's NCEPLIBS packages (g2, g2c, bacio, w3emc). The packages
+    # export incorrect target names that don't match what WW3 expects. This should
+    # be reported to WW3 developers and/or fixed in Spack's NCEPLIBS packages.
+    # depends_on("g2c", when="switch=NCEP_glwu")
+    # depends_on("g2", when="switch=NCEP_glwu")
+    # depends_on("bacio", when="switch=NCEP_glwu")
+    # depends_on("w3emc", when="switch=NCEP_glwu")
+
+    # depends_on("scotch", when="switch=NCEP_rwps")
```
bernhardkaindl, line 109: @rsoutelino This does not look like it should for a package that is ready for review.

Do you plan to finish the recipe? As far as I'm aware merged spack recipes don't contain such disabled code.

rsoutelino: Thanks @bernhardkaindl. No, sorry, those are just leftovers notes. They can be deleted

bernhardkaindl: > Thanks @bernhardkaindl. No, sorry, those are just leftovers notes. They can be deleted

@rsoutelino, I shouldn't remove them because if I do, I can't merge as the last pusher of the branch.

It would be better if you could clean them up and then request a re-review of wdconinc, because as he requested changes of this PR before. If you can't make progress after cleaning everything up and wdconinc doesn't re-review, your last resort would be opening a new PR, but you need to cleanup the PR yourself first.

bernhardkaindl:
@rsoutelino, can you apply the cleanup changes requested yourself?

repos/spack_repo/builtin/packages/ww3/package.py
rsoutelino: @bernhardkaindl Yes, I'll clean them up. Thanks

PR #3692, opened 2026-03-06
chapel: add v2.8

e-kayrakli:
This is a WIP PR to add the upcoming Chapel 2.8 release, which is scheduled for March 12th.

Changes include:
- Add 2.8
- Deprecate 2.5
- Add a new version of an existing patch for 2.8
- Update LLVM dependencies

### DIFF
--- a/repos/spack_repo/builtin/packages/chapel/package.py
+++ b/repos/spack_repo/builtin/packages/chapel/package.py
@@ -63,11 +63,17 @@ class Chapel(AutotoolsPackage, CudaPackage, ROCmPackage):
 
     version("main", branch="main")
 
+    version(
+        "2.8.0",
+        url="https://chapel-lang.org/tmp/chapel-2.8.0.tar.gz",
+        sha256="3e5e8da6c56"
+    )
+
     version("2.7.0", sha256="5e3269babdae334c80fc3f25114698fdfe53e84ea06626af22d2b54eeb75bee6")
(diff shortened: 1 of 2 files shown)


repos/spack_repo/builtin/packages/chapel/package.py
```diff
@@ -63,11 +63,17 @@ class Chapel(AutotoolsPackage, CudaPackage, ROCmPackage):
 
     version("main", branch="main")
 
+    version(
+        "2.8.0",
+        url="https://chapel-lang.org/tmp/chapel-2.8.0.tar.gz",
```
bonachea: @e-kayrakli This looks like a temporary URL.

If this URL is not the persistent/archival location of the release, then I suggest we demote this to a Draft PR to ensure it doesn't get merged with a URL that is expected to stop working soon.

bonachea (changes requested):
Thanks for creating this PR, but I'm guessing it's not yet ready for merge?

repos/spack_repo/builtin/packages/chapel/package.py
arifthpe: > I suggest we demote this to a Draft PR to ensure it doesn't get merged with a URL that is expected to stop working soon

This is my fault, I remember having this conversation with you previously @bonachea but I just checked and our internal release process docs still say to _not_ use a draft PR, will go update that now.

e-kayrakli:
Closing in favor of https://github.com/spack/spack-packages/pull/3836

PR #3697, opened 2026-03-07
py-pillow(-simd): deprecate versions due to CVEs

adamjstewart:
Not sure why pillow has so many more CVEs than the average library. I believe almost all of these are severe enough to warrant deprecation. This will eventually allow us to remove py-pillow-simd (abandoned) and the pil provider.


adamjstewart:
@spackbot fix style

haampie (approved):

PR #3698, opened 2026-03-07
TensorFlow: add v2.21.0

adamjstewart:
https://github.com/tensorflow/tensorflow/releases/tag/v2.21.0

### DIFF
--- a/repos/spack_repo/builtin/packages/py_tensorflow/package.py
+++ b/repos/spack_repo/builtin/packages/py_tensorflow/package.py
@@ -182,6 +182,8 @@ class PyTensorflow(Package, CudaPackage, ROCmPackage, PythonExtension):
         depends_on("patchelf", when="@2.13: platform=linux")
         # https://github.com/tensorflow/tensorflow/issues/60179#issuecomment-1491238631
         depends_on("coreutils", when="@2.13: platform=darwin")
+        # XLA requires xxd now
+        depends_on("xxd-standalone", when="@2.21:")
 
         depends_on("swig")
         depends_on("py-pip")
--- a/repos/spack_repo/builtin/packages/py_tensorflow/package.py
+++ b/repos/spack_repo/builtin/packages/py_tensorflow/package.py
@@ -48,6 +48,7 @@ class PyTensorflow(Package, CudaPackage, ROCmPackage, PythonExtension):
     maintainers("adamjstewart", "aweits")
     tags = ["e4s"]
 
+    version("2.21.0", sha256="ef3568bb4865d6c1b2564fb5689c19b6b9a5311572cd1f2ff9198636a8520921")
     version(
         "2.20.0-rocm-enhanced",
         sha256="1db75eb24f617ac0b1aea417c294cbdf98ec7ede3cb2957e07c1e9f8eefa8713",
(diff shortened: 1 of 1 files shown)


adamjstewart:
https://github.com/tensorflow/tensorflow/issues/111876

haampie (dismissed):

bernhardkaindl:
As CI passed for #5424 two weeks ago which is very similar, the CI failure should hopefully not be a permanent issue. I tried GitHub's rebase variant, but it failed with a conflict between head and base, so I had to resort to a normal merge of develop to retry CI with the current CI code base.

adamjstewart:
CI is passing now because the failure happened on macOS and our darwin runners seem to be offline? So as soon as we re-add darwin tests, we would hit this failure again. I would hold off on merging until we get macOS tests to pass.

adamjstewart:
I'm actually considering moving forward with this anyway, it lets us build newer tensorboard and therefore newer setuptools in CI. Can anyone review?

wdconinc:
So the idea here would be `conflicts("platform=macos", when="@2.21:")`?

adamjstewart:
I would rather not resort to that, but yeah...

repos/spack_repo/builtin/packages/py_tensorflow/package.py
wdconinc, line 925: @adamjstewart This addition appears to make my build of py-tensorflow succeed (on my wife's Mac, apple-m3, so I can't as easily debug during the week when she needs it).
```suggestion
        if spec.satisfies("platform=darwin") and platform.machine() == "arm64":
            filter_file(r"build:opt --host_copt=-mcpu=apple-m[0-9]+", "", ".tf_configure.bazelrc")

```
Before this I got the same error on the host tools architecture. It does not seem like TF has any support yet for defining what architecture is used for the host tools, so suppressing the mcpu seems to be the best way forward here. And that's a spack-packages addition that doesn't really help much in the TF bug report...

adamjstewart: Will try a couple combinations of this to see if I can make it even simpler, but I like this workaround for now.

repos/spack_repo/builtin/packages/py_tensorflow/package.py
```diff
@@ -912,8 +918,10 @@ def post_configure_fixes(self):
             after = spec["llvm-amdgpu"].prefix.bin.clang
             filter_file(before, after, ".bazelrc")
 
-        filter_file("build:opt --copt=-march=native", "", ".tf_configure.bazelrc")
-        filter_file("build:opt --host_copt=-march=native", "", ".tf_configure.bazelrc")
+        # Support for host_copt customization on macOS arm64 seems to be broken?
+        # https://github.com/tensorflow/tensorflow/issues/111876
+        if spec.satisfies("@2.21: platform=darwin target=aarch64:"):
+            filter_file("build:opt --host_copt=.*", "", ".tf_configure.bazelrc")
```
adamjstewart, line 924: @wdconinc see what you think of this.

1. The former `filter_file` calls never did anything because we set `CC_OPT_FLAGS`, which controls these. It will never contain `-march=native` now that archspec is used to set these flags.
2. `--copt` seems fine, only `--host_copt` breaks on macOS arm64
3. 2.21 is problematic, previous versions seemed to work fine

This is still a workaround until we learn why `--host_copt` isn't working, updated my bug report to try to get to the bottom of this now that I know where most of these flags are coming from.

wdconinc: I think this looks reasonable (but I can't test until the weekend).

I think the issue is that `--host_platform` still points to the bazel default (which appears to be x86_64). Injecting the following in the `.tf_configure.bazelrc` may also work:
```
build --host_platform=@build_bazel_apple_support//platforms:darwin_arm64
```
But it should likely be `configure.py` that injects this rather than our `package.py`. And adding a line is tricky since I don't know if it is order dependent in bazelrc files. I'd put that right before the original `--host_copt` by hand, and try it.

bernhardkaindl (approved):

PR #3699, opened 2026-03-07
PyTorch: deprecate versions affected by CVEs

adamjstewart:


adamjstewart:
@alalazo @kwryankrattiger this should cut the number of PyTorch builds in half, didn't realize torchdata required such an old version.

haampie (approved):

PR #3702, opened 2026-03-07
pandas: deprecate versions affected by CVEs

adamjstewart:


rgommers (commented):
That doesn't seem right to me, the CVE has already been rejected and no longer has a description, so 99.x% chance it was just nonsense: https://www.cve.org/CVERecord?id=CVE-2024-42992

PR #3703, opened 2026-03-07
OpenCV: deprecate versions affected by CVEs

adamjstewart:


adamjstewart:
@spackbot fix style

haampie:
Two issues:

**1. Duplicate CVE comment on version 4.1.2** — both lines reference CVE-2019-5064. The second should be CVE-2019-5063 (companion heap overflow via XML, same affected range 4.0.0–4.1.x, fixed in 4.2.0).

**2. Missing `version("4.12.0", ...)` entry** — `"4.12.0"` is added to `contrib_vers` but there's no corresponding `version()` declaration with a sha256. After deprecating 4.10.0, 4.9.0 becomes the latest available, leaving the CVE-2025-53644 fix (4.12.0) unpackaged. The PR should add the 4.12.0 version entry.

haampie (approved):

PR #3704, opened 2026-03-07
py-horovod: deprecate versions affected by CVEs

adamjstewart:


haampie:
Thanks for the PR! One issue: per NVD, CVE-2024-10190 affects Horovod **up to and including v0.28.1 with no documented fix** (CPE: `horovod:horovod:*:*:*:*:*:*:*:*`, versions ≤ 0.28.1 — https://nvd.nist.gov/vuln/detail/CVE-2024-10190). The PR deprecates only 0.24.0–0.24.2, but 0.24.3, 0.25.0, 0.26.0, and 0.27.0 are also in the affected range. If 0.24.3 contains a backported fix not yet reflected in NVD, a note/reference would help; otherwise those versions should be deprecated under this CVE too. (CVE-2022-0315 looks correct.)

adamjstewart:
@thomas-bouvier should we deprecate all versions of horovod until a fix is released?

haampie:
Probably not?

thomas-bouvier:
It’s unlikely we’ll see a new Horovod release, since most (if not all) of its features are now natively supported in PyTorch.

I am not sure about Spack's depreciation policy. All I can say is that Horovod was widely used at some point and we have to keep some versions in the repo ;)

adamjstewart:
Was this fixed in master? Given that the CVE in question is Critical (9.8/10) and there haven't been any new horovod releases in almost 3 years, I'm inclined to deprecate the entire package recipe. It won't be removed for some time (~6 months?) and people can always install an older version of spack-packages if they need horovod. What do others think?

tldahlgren (approved):

eugeneswalker:
If the package is not being actively maintained, and its features are natively supported in pytorch, and no new releases have happened for ~3 years-- it seems reasonable to deprecate it. People who want to use it with Spack could still do it. My two cents.

PR #3705, opened 2026-03-07
Python: add new versions

adamjstewart:
Basically all versions other than the latest patch release have a CVE: https://www.cvedetails.com/version-list/10210/18230/1/Python-Python.html


adamjstewart:
@spackbot fix style

haampie (approved):

PR #3713, opened 2026-03-07
libjpeg: add v10

haampie:
Add libjpeg v10 (higher bit depth support prep, new lossless negate in jpegtran, no dependency changes). Built and installed on darwin-m4 / apple-clang@17.


RikkiButler20 (approved):
Looks good to me

PR #3714, opened 2026-03-07
swig: add v4.4.1

haampie:
Add swig 4.4.1 (Python 3.13 fixes, no dependency changes). Drop `preferred=True` from 4.1.1 and update `sourceforge_mirror_path`. Built and installed on darwin-m4 / apple-clang@17.


haampie:
Two follow-up commits:

`swig`: added `preferred=True` to all standard (non-fortran) versions. Without this, a constraint like `swig@:4.2` resolves to `4.1.1-fortran` instead of `4.1.1` because Spack's version ordering treats the `-fortran` suffix as greater than the bare numeric version. The fortran fork should only be chosen when explicitly requested.

`libpressio`: capped `swig@:4.2` for `+python` builds. swig 4.3.0 changed `SWIG_Python_AppendOutput` to a 3-arg function (swig/swig@cd39cf13); libpressio's bundled `numpy.i` still uses the old 2-arg form and hasn't been fixed upstream.

sethrj:
@haampie FWIW the `-fortran` for a version has the same capabilities as that version, plus the ability to compile fortran: I think it's OK if 4.1.1-fortran is preferred over 4.1.1

sethrj (approved):
Verified installation of 4.4.1 as default and successful install of swig@4.4.1 build_system=autotools platform=darwin os=tahoe target=m2 %c,cxx=apple-clang@17.0.0 .

PR #3715, opened 2026-03-07
boost: add v1.90.0

haampie:
Add boost 1.90.0 and add `openmethod` variant for the new header-only OpenMethod library (requires `@1.90.0:`). Built and installed on darwin-m4 / apple-clang@17.


hainest (approved):
SHA confirmed.

hainest:
@spackbot re-run pipeline, please

hainest (approved):

tldahlgren (approved):
Confirmed the sha256 and use of lib option.

PR #3716, opened 2026-03-07
protobuf, abseil-cpp, py-grpcio: add v34.0, v20260107.1, v1.78.1

haampie:
- **protobuf**: add v34.0; add `conflicts("%gcc@:12", when="@34: +shared")`
- **abseil-cpp**: add v20260107.1; [require C++17 for `@20250512:`](https://github.com/abseil/abseil-cpp/releases/tag/20250512.0); fix `cxxstd=11` condition (`@:2022` → `@:20220623` — old bound never matched since versions are date-based); update default to `17`
- **py-grpcio**: add v1.78.1; [require `py-setuptools@77.0.1:` for `@1.78:`](https://github.com/grpc/grpc/blob/v1.78.1/pyproject.toml); patch missing includes for abseil 20260107+ on `@1.71:1.76`

The GCC ≤12 conflict is needed because protobuf 34.0 mixes `__attribute__((visibility("default")))` with `[[gnu::warn_unused]]` in shared builds, which older GCC can't parse ([GCC#69585](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=69585), fixed in GCC 13). Introduced in [protobuf@10af44b](https://github.com/protocolbuffers/protobuf/commit/10af44bb0b8e2428b072c06ca0cee7ff0b38c624), reported upstream as [protobuf#26383](https://github.com/protocolbuffers/protobuf/issues/26383). Minimal reproducer:

```cpp
// g++-12 -std=c++17 -c test.cpp  → FAILS
// g++-13 -std=c++17 -c test.cpp  → OK
class __attribute__((visibility("default"))) [[gnu::warn_unused]] Foo {
public:
    int x;
};
```

Verified: fails with GCC 11.5 and 12.4 `+shared`, passes with `~shared` and with GCC 13.3.

Tested on darwin/aarch64 (apple-clang@17) and linux/zen2 (GCC 11/12/13).


hyoklee (approved):

haampie:
Two follow-up fixes added after discovering issues when `abseil-cpp@20260107.1` is used in C++17 mode on GCC:

**protobuf:** Added `depends_on("abseil-cpp@:20240722", when="@22:")` — similar to [protocolbuffers/protobuf#25701](https://github.com/protocolbuffers/protobuf/issues/25701), mixing `ABSL_ATTRIBUTE_WARN_UNUSED` (which expands to `[[gnu::warn_unused]]` in C++17) with `PROTOBUF_EXPORT` (`__attribute__((visibility(...)))`) in class-head declarations causes a GCC build failure. Not reproducible on darwin/apple-clang. Capped until upstream fixes it.

**py-grpcio:** `grpcio 1.75.0` fails to build with `abseil-cpp@20260107.1` because `absl/strings/string_view.h` no longer transitively includes `<string>` (abseil now aliases `absl::string_view` to `std::string_view`), leaving `std::to_string` undeclared in `memory_request.h`. Backported [grpc/grpc#41351](https://github.com/grpc/grpc/pull/41351) as a patch for `@1.71:1.76`. Also added `py-grpcio@1.78.1` which already contains the fix and requires `setuptools@77.0.1:`.

haampie:
not a bot: looks like the issue is with the cxxstd=17 default in abseil, also older versions of abseil with new protobuf fail.

hyoklee (approved):

adamjstewart:
Changes to py-grpcio look good, didn't review the rest. I was about to open a PR with the same changes but will wait for this one instead.

haampie:
CI failed for py-tensorflow in the e4s stack. py-tensorflow has `conflicts("%gcc@:11", when="@2.17:")`, so with gcc@11.4 the newest usable version is `@2.16.2`, which pins through to py-protobuf `@4.25.3` → protobuf `@25.3`.

The breakage comes from abseil-cpp `@20260107.1`: `absl/strings/string_view.h` used to transitively include `<cstring>`, but after `absl::string_view` became a direct alias for `std::string_view` that transitive include disappeared, exposing a missing `#include <cstring>` in protobuf's bundled `third_party/utf8_range/utf8_validity.cc`. The memcpy was removed in `@26+` ([9c7d2b9](https://github.com/protocolbuffers/protobuf/commit/9c7d2b9d83e0ad4e8ffeb8535d8c0a31a0a6a0b8)) so this only affects `@22:25`. The patch here adds the explicit include.

haampie:
Passing in CI now, Windows failure is unrelated.

hyoklee (approved):

alalazo (approved):

PR #3741, opened 2026-03-08
.github/skills: add general package-update, py-package-update, and pr-review skills

wdconinc:
Adds a Copilot skill documenting the bulk package-update workflow:

- Phase-by-phase instructions (inventory → dependency analysis → apply changes → PR grouping → CI check → PR submission)
- PR description template with AI attribution
- Policy: **do not close existing PRs** from other authors — post a comment linking to the new PR instead
- Conventions reference table and common pitfalls

Derived from the workflow used in the March 2026 bulk refresh of 237+ packages across 22 PRs.

---
> ⚠️ This PR was prepared with AI assistance from GitHub Copilot.
> It will only be marked ready for review after human review of all changes.

### DIFF
--- a/.github/skills/package-update.md
+++ b/.github/skills/package-update.md
@@ -1,0 +1,352 @@
+# Skill: General Package Version Update (Bulk)
+
+Perform a bulk update of many packages across one or more ecosystems in this Spack repository.
+This skill covers the full workflow: inventorying locally modified packages, verifying dependency
+changes introduced by new versions, grouping the changes into coherent pull requests, and
+submitting all PRs as drafts for human review.
+
+Use the more focused [`py-package-update`](py-package-update.md) skill for smaller, targeted
+updates of Python packages.
+
+## Usage
+
+Invoke this skill when the working tree already contains many modified `package.py` files that
+need to be committed and submitted, or when you want to refresh a broad set of packages to their
+latest upstream releases:
+
+```
+Update all locally modified packages and open PRs
+Update all packages in the HEP stack to their latest versions
+```
+
+## Step-by-Step Instructions
+
+---
+
+### Phase 1 — Inventory the modified packages
+
+#### 1a. List all modified package files
+
+```sh
+git diff --name-only HEAD | grep 'packages/' | sort
+```
+
+Also check for new (untracked) package directories that should be included:
+
+```sh
+git status --short | grep '^\?\?' | grep 'packages/'
+```
+
+Note any new packages (entirely new directories), new patch files, and packages with
+structural changes beyond just version bumps.
+
+#### 1b. Classify each package
+
+For each modified file record:
+
+- Package name (Spack name with hyphens)
+- Highest new version being added
+- Whether the change is: version-only, dependency change, build-system change, new package,
+  or new patch file
+
+---
+
+### Phase 2 — Verify dependency changes for each package
+
+For every modified package, determine what dependency changes the **new versions** introduce.
+This requires comparing the upstream build specification of the new version against the
+previously packaged version.
+
+#### 2a. Locate the upstream source
+
+The upstream source is identified by the `pypi`, `url`, or `git` attribute in `package.py`.
+
+- **PyPI packages** (`pypi =`): fetch `https://pypi.org/pypi/<dist-name>/json` to get the
+  release list. Download and inspect the `pyproject.toml`, `setup.cfg`, or `setup.py` for the
+  new version from the sdist tarball link under `releases[<version>]`.
+- **GitHub-hosted packages** (`url =` or `git =`): use the GitHub API or `git ls-remote` to
+  check for new tags. Inspect the relevant build config files at the tagged commit.
+- **Autotools/CMake packages**: inspect `configure.ac`, `CMakeLists.txt`, and any
+  `*Config.cmake.in` or `*-config.cmake.in` files for `find_package`, `pkg_check_modules`, and
+  minimum version requirements.
+
+Use parallel sub-agents (one per ecosystem or logical group) to speed up the analysis phase.
+
+#### 2b. Inspect the upstream build config at the new version
+
+| Build system | Files to inspect | What to look for |
+|---|---|---|
+| Python/PyPI | `pyproject.toml`, `setup.cfg`, `setup.py` | `requires-python`, `dependencies`, `[build-system].requires` |
+| CMake | `CMakeLists.txt`, `cmake/Find*.cmake` | `find_package(Foo VERSION X.Y REQUIRED)` |
+| Autotools | `configure.ac`, `Makefile.am` | `AC_CHECK_LIB`, `PKG_CHECK_MODULES`, version constraints |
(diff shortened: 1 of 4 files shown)


.github/skills/package-update.md
```diff
@@ -0,0 +1,518 @@
...
+> # Correct:
+> version("1.3.0", sha256="abc123...")
+> ```
+>
+> A quick way to remove all such comments in bulk before committing:
+> ```sh
+> sed -i 's/  # FIXME$//' repos/spack_repo/builtin/packages/<pkg>/package.py
+> ```
+
+Do not manually compute checksums. Do not modify `url`, `homepage`, `pypi`, `git`, `maintainers`,
```
cmelone:
```suggestion
Do not manually compute checksums. Do not modify `url`, `homepage`, `pypi`, `git`, `maintainers`, `license`
```

.github/skills/package-update.md
```diff
@@ -0,0 +1,518 @@
...
+    build_system("cmake", default="cmake", when="@2:")
+```
+
+Keep both base classes active so existing older versions continue to build.
+
+#### 3c. New packages
+
+If a new upstream dependency does not yet exist in the repository, create its `package.py` before
+adding the `depends_on(...)` reference. Follow the [`py-package-update`](py-package-update.md)
+conventions for Python packages.
```
cmelone:
````suggestion
If an upstream dependency does not yet exist in the repository, create the package before adding the `depends_on(...)` reference.
The `spack create` command will generate a recipe scaffold.
```
spack create <url> --name <pkg_name> --skip-editor
``` 
Parse the output to extract the created file path; read the file and add packaging details. Follow the [`py-package-update`](py-package-update.md) conventions for Python packages.
````

File: .github/skills/package-update.md [hunk 3]
```diff
@@ -0,0 +1,518 @@
...
+Before writing PR descriptions, determine which new versions are already in the binary cache
+(`cache.spack.io`) and which will need new CI builds. This helps reviewers set expectations.
+
+For each new version of a package, check:
+
+```sh
+curl -s "https://cache.spack.io/package/develop/<spack-pkg-name>/specs/" | \
+  python3 -c "import sys,re; data=sys.stdin.read(); vers=re.findall(r'\"versions\":\[\"([^\"]+)\"', data); [print(v) for v in sorted(set(vers))]"
+```
+
```
cmelone, line 262: this is great!

.github/skills/package-update.md
```diff
@@ -0,0 +1,518 @@
...
+Updated the following packages to their latest upstream releases:
+
+| Package | Old version | New version |
+|---------|-------------|-------------|
+| `<name>` | `<old>` | `<new>` |
+
+## Dependency changes
+
+<List each change: package, what changed, which version introduced the change.
+If none, write "No dependency changes.">
```
cmelone: flagging this re: the thread in #packaging .. probably want to make this prompt more detailed

cmelone (commented):
i think this is good! one general comment I have is whether we should be splitting up a package update skill that would work for a single package with another skill that details what's needed to do bulk updates

.github/skills/pr-review.md
```diff
@@ -0,0 +1,126 @@
...
+
+### 3. Version Currency
+
+For every package version being added or updated:
+
+- Query PyPI (`https://pypi.org/pypi/<package-name>/json`) to retrieve the current latest stable
+  release. Compare it to the version being added. If the PR adds version X but PyPI already shows
+  a newer stable release Y, flag this.
+- If the package uses a non-PyPI source (e.g., a GitHub `url` or `git` attribute), check the
+  upstream repository's tags or releases page to verify the submitted version is the latest.
```
cmelone:
```suggestion
- If the package uses a non-PyPI source (e.g., a GitHub `url` or `git` attribute), check the upstream repository's tags or releases page to verify the submitted version is the latest.
- Query PyPI (`https://pypi.org/pypi/<package-name>/json`) to retrieve the current latest stable release. Compare it to the version being added. If the PR adds version X but PyPI already shows a newer stable release Y, flag this.
```

.github/skills/pr-review.md
```diff
@@ -0,0 +1,126 @@
...
+- Do **not** flag pre-releases (alpha, beta, rc) as missing unless the PR explicitly targets
+  pre-releases.
+
+### 4. Dependency Completeness
+
+For each changed package, examine the upstream build configuration to verify that all dependency
+changes are reflected in `package.py`.
```
cmelone: https://github.com/spack/spack-packages/pull/3741#discussion_r2913081860

.github/skills/package-update.md
```diff
@@ -0,0 +1,517 @@
...
```
wdconinc:
```suggestion
<List and justify each change: package, what changed, which version introduced the change. Always include links to the commit that introduced the dependency change to allow reviewers to validate, or justify by quoting the release notes if the package has no git attribute.
If none, write "No dependency changes.">
```

.github/skills/pr-review.md
```diff
@@ -0,0 +1,123 @@
...
+
+- If the package uses a non-PyPI source (e.g., a GitHub `url` or `git` attribute), check the upstream repository's tags or releases page to verify the submitted version is the latest.
+- Query PyPI (`https://pypi.org/pypi/<package-name>/json`) to retrieve the current latest stable release. Compare it to the version being added. If the PR adds version X but PyPI already shows a newer stable release Y, flag this.
+- Do **not** flag pre-releases (alpha, beta, rc) as missing unless the PR explicitly targets
+  pre-releases.
+
+### 4. Dependency Completeness
+
+For each changed package, examine the upstream build configuration to verify that all dependency
+changes are reflected in `package.py`.
```
wdconinc:
```suggestion
For each changed package, examine the upstream build configuration to verify that all dependency
changes are reflected in `package.py`. Use links to the commit that introduced the dependency change to allow reviewers to validate the changes, or justify by quoting the release notes if the package has no git attribute.
```

.github/skills/package-update.md
wdconinc: Applied in 512345a

.github/skills/pr-review.md
wdconinc: Applied in 512345a

.github/skills/package-update.md
wdconinc: There's probably a better way to get this info than scraping a url. Presumably there's a json somewhere that the page is built from... but this gets the job done.

wdconinc:
> i think this is good! one general comment I have is whether we should be splitting up a package update skill that would work for a single package with another skill that details what's needed to do bulk updates

Yeah, this went more in the direction of bulk update than I wanted. I think in some sense this is something we should start using and then tweak based on experience since this is all so new that I at least don't really know yet how it will work out in practice when used by someone who isn't using the same LLM or has the same way of prompting that I use.

cmelone (approved):
> I think in some sense this is something we should start using

agreed!

bernhardkaindl (approved):

PR #1572, opened 2025-09-18
Resolves a python@12 ^py-numpy@1.26.2:1.26:3 conflict; adds depends_on('cxx') to lib tool

markyoder:
Resolves compile issues originally encountered building py-torch. Our application requires python@12.x, which I believe is what triggers py-torch to prefer py-numpy@1.26.x. The That combination of python@3.12 ^py-torch@1.26.2:1.26.3 breaks with a "cannot generate metadata" error.

It may be possible to fix this by specifying a different version of py-setuptools, but I fixed it with
py_numpy/package.py:
conflicts('@:1.26.3', when='^python@3.12', msg="Known bug with @1.26.2, 1.26.3 when ^python@3.12")

In the course of building that, I also added:
libtool/package.py
depends_on("cxx", type="build")

PR #3742, opened 2026-03-08
py-numpy: fix setuptools constraint for numpy 1.26

### LINKED ISSUE #3695: Missing constraint on py-setuptools version for py-numpy@1.26
## Description

`py-numpy@1.26.*` use deprecated `distutils` features in `py-setuptools`, similar to `py-nump@1.25`. But there is no constraint on the versions of `py-setuptools` for `py-numpy@1.26`:

https://github.com/spack/spack-packages/blob/71b95aa5d0a215a6564e8327676b4e9c51a8832c/repos/spack_repo/builtin/packages/py_numpy/package.py#L135-L137

I don't know the exact versions, but in the environments we've been using widely, `py-numpy@1.26.4` (and presumably `.5`) work fine with `py-setuptools@69`. Since we are talking about deprecated versions of either of these packages, would it be reasonable to be pragmatic and simply add
```
        depends_on("py-setuptools@:69", when="@:1.26")
```
? Otherwise I'll have to dig deeper to figure exactly which versions of `py-setuptools` work with `py-numpy@1.26`.

adamjstewart:
Closes #3695 @climbfuji 
Closes #1572 @markyoder

### DIFF
--- a/repos/spack_repo/builtin/packages/py_numpy/package.py
+++ b/repos/spack_repo/builtin/packages/py_numpy/package.py
@@ -127,7 +127,7 @@ class PyNumpy(PythonPackage):
     with default_args(type="build"):
         depends_on("py-pyproject-metadata@0.7.1:", when="@1.26.0:1.26.3")
         depends_on("py-tomli@1:", when="@1.26.0:1.26.3 ^python@:3.10")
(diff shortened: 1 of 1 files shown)


rgommers (approved):
LGTM, thanks @adamjstewart

haampie (approved):

PR #3746, opened 2026-03-09
py-openai: add new version 2.26.0

yizeyi18:
Add the newer version of `py-openai@2.26.0` with related updates to its dependencies.


yizeyi18:
An unchanged package, `py-maturin`, fails to be built during ci. It seems that rust linker do not recognize some intel's private symbol which are in static library not referred by cmdline.
As previous ci pipelines for the PRs adding new version of `py-maturin` (#2666, #2098) did not try to build the package with oneapi toolchain (see https://gitlab.spack.io/spack/spack-packages/-/pipelines/1336270, https://gitlab.spack.io/spack/spack-packages/-/pipelines/1297801), it is reasonable to guess that `py-maturin` may be incompatible with the toolchain.

yizeyi18:
@eugeneswalker The `py-maturin` case for this PR is similar to some others in `stacks/e4s-oneapi/spack.yaml`. 
As `py-openai` have a `py-openai->py-jiter->py-maturin` dependency chain since its version 1.40.0[1], may I remove it from oneapi test list?

[1] `py-jiter` dependent added by https://github.com/openai/openai-python/commit/bf1ca86cf392eb0ffed1e146937c5d73d8a568f0

yizeyi18:
`py-openai` has been removed from the failing ci pipeline. 

@spackbot re-run pipeline.

yizeyi18:
Still tested by e4s-oneapi and fails. That's strange. Do spack ci pipeline run according to the source branch, instead of develop + change? It may be better to update the branch to eliminate the errorneous oneapi test for the source branch.

yizeyi18:
@spackbot request review

yizeyi18:
No reply from spack bot. Did I mistype the command?
@spackbot help

yizeyi18:
@spackbot maintainers

yizeyi18:
Hi @bernhardkaindl : 
  The requested changes were applied. May you have a re-review of this PR?

repos/spack_repo/builtin/packages/py_openai/package.py
```diff
@@ -21,28 +21,50 @@ class PyOpenai(PythonPackage):
...
+    depends_on("py-typing-extensions@4.11:4", type=("build", "run"), when="@1.40:")
+    depends_on("py-anyio@3.5.0:4", type=("build", "run"), when="@1.3.8:")
+    depends_on("py-distro@1.7.0:1", type=("build", "run"), when="@1:")
+    depends_on("py-sniffio", type=("build", "run"), when="@1:")
+    depends_on("py-jiter@0.10:0", type=("build", "run"), when="@2.3.0:")
+    depends_on("py-aiohttp", type=("build", "run"), when="@:0")
 
     with when("+datalib"):
         depends_on("py-numpy", type=("build", "run"))
+        depends_on("py-numpy@1:", when="@1:")
```
bernhardkaindl: ## Missing `type=` on `py-numpy@1:` inside `+datalib`
Without `type=`, Spack defaults to ("build","link"), so the `@1:` version constraint won't apply at run time.

Suggested fix:

```suggestion
        depends_on("py-numpy@1:", type=("build", "run"), when="@1:")
```
@yizeyi18, I think this should be applied.

repos/spack_repo/builtin/packages/py_openai/package.py
```diff
@@ -21,28 +21,50 @@ class PyOpenai(PythonPackage):
...
+        when="@1.89.0:",
+    )
+    variant("realtime", default=False, description="support for Realtime API", when="@1.58:")
 
     depends_on("python@3.7.1:", type=("build", "run"))
-    depends_on("py-setuptools", type="build")
-    depends_on("py-requests@2.20:", type=("build", "run"))
+    depends_on("python@3.9:", type=("build", "run"), when="@2.7.2:")
+    depends_on("py-setuptools", type="build", when="@:0")
+    depends_on("py-hatchling@1.26.3", type="build", when="@1:2")
```
bernhardkaindl, line 47: ## Exact version pin py-hatchling@1.26.3 in py_openai/package.py

The hatchling dependency is pinned to an exact version rather than using a minimum version constraint, which goes against Spack conventions and could cause issues if a newer version is available and wanted by the user or another package.

Suggestion: Use a minimum version instead of an exact version:

```suggestion
    depends_on("py-hatchling@1.26.3:", type="build", when="@1:2")
```
@yizeyi18, I think this should be applied unless you think that only this single version works for py-openai.

repos/spack_repo/builtin/packages/py_openai/package.py
```diff
@@ -51,6 +73,17 @@ class PyOpenai(PythonPackage):
...
+    with when("+aiohttp"):
+        depends_on("py-aiohttp", type=("build", "run"))
+        depends_on("py-httpx-aiohttp@0.1.9:", type=("build", "run"))
+
+    with when("+realtime"):
+        depends_on("py-websockets@13:15", type=("build", "run"))
+
+    with when("+voice"):
+        depends_on("py-sounddevice@0.5.1:", type=("build", "run"))
+        depends_on("py-numpy@2.0.2", type=("build", "run"))
```
bernhardkaindl: ## Exact version pin py-numpy@2.0.2 in +voice block in py_openai/package.py

In spack, `@2.0.2` is an exact pin, not a minimum. This will force every user of the `+voice` variant to use exactly numpy 2.0.2, which I assume is not needed and instead, I assume 2.0.2 or newer is the real dependency for `+voice`:

```suggestion
        depends_on("py-numpy@2.0.2:", type=("build", "run"))
```
@yizeyi18, unless you suggest otherwise, I think this should be applied.

repos/spack_repo/builtin/packages/py_jiter/package.py
```diff
@@ -15,8 +15,10 @@ class PyJiter(PythonPackage):
 
     license("MIT", checked_by="qwertos")
 
+    version("0.13.0", sha256="f2839f9c2c7e2dffc1bc5929a510e14ce0a946be9365fd1219e7ef342dae14f4")
     version("0.5.0", sha256="1d916ba875bcab5c5f7d927df998c4cb694d27dceddf3392e58beaf10563368a")
 
     depends_on("python@3.8:", type=("build", "run"))
     depends_on("py-maturin@1", type="build")
     depends_on("rust@1.73:", type=("build", "run"))
+    depends_on("rust@1.83:", type=("build", "run"), when="@0.13:")
```
bernhardkaindl: ## Advisory (pre-existing, not introduced by this PR):

In `py_jiter`, `rust` is listed as `type=("build", "run")`, and as far as I am aware, Rust is a compile-time-only toolchain, and should not be a runtime dependency for Python/maturin packages. The new @0.13: line repeats the same pattern.

Possible change, if rust is just used as build time dependency and not used as a rust compiler at runtime:

```suggestion
    depends_on("rust@1.73:", type="build")
    depends_on("rust@1.83:", type="build", when="@0.13:")
```
@yizeyi18, If you agree, I'd like to suggest to fix this pre-existing issue as well.

bernhardkaindl (approved):
@yizeyi18, after the initial review by Copilot on the Web, a deeper review found further items to address in this PR:

bernhardkaindl (changes requested):
Oh, I didn't mean to approve with these issues open, so I now have to set my state to request changes to revert that.

repos/spack_repo/builtin/packages/py_openai/package.py
yizeyi18: I think the version-free line above should have constrained the dependent type of py-numpy, and the version-specific line is a narrowing of the version-free so unnecessary to repeat the `type`.

repos/spack_repo/builtin/packages/py_openai/package.py
yizeyi18: `pyproject.toml` of `py-openai` explicitly required an exact version. See:
https://github.com/openai/openai-python/blob/6d9262d5c666a1e4d47f63178db907ba3087ac5d/pyproject.toml#L112

bernhardkaindl:
Status review: The other two review suggestions are not resolved yet.

repos/spack_repo/builtin/packages/py_openai/package.py
bernhardkaindl: But you aren't sure? What about making the safe choice?

When I look here, if type is not give, the default argument is `("build", "link")`, so not giving it is like passing `("build", "link")`, so AFAICS, this lines effectively passes  `("build", "link")`:

https://github.com/spack/spack/blob/develop/lib/spack/spack/directives.py#L302
https://github.com/spack/spack/blob/develop/lib/spack/spack/deptypes.py#L32

bernhardkaindl: Follow-up from https://github.com/spack/spack-packages/pull/3746#discussion_r3548202509 to make this remaining visible at the bottom of the PR:

When I look here, if type is not given, the default argument is `("build", "link")`.

Not passing it is like passing `("build", "link")`, so AFAICS, this lines effectively passes  `("build", "link")`:

https://github.com/spack/spack/blob/develop/lib/spack/spack/directives.py#L302
https://github.com/spack/spack/blob/develop/lib/spack/spack/deptypes.py#L32

bernhardkaindl (changes requested):

repos/spack_repo/builtin/packages/py_openai/package.py
yizeyi18: You are right, I will apply the suggested change.

bernhardkaindl (approved):

PR #3748, opened 2026-03-09
root: Set ROOT_INCLUDE_PATH in build as well

paulgessinger:
Without this, starting from (I think) ROOT 3.38.02, the build fails when `+cxxmodules`, which is the default.

@jmcarcell pointed out that the LCG build sets `ROOT_INCLUDE_PATH` for the build: https://gitlab.cern.ch/sft/stacks/lcgcmake/-/blob/master/projects/CMakeLists.txt#L93

The spack package currently only does this for the run and dependent build. Adding it to the build seems to solve the issue.

### DIFF
--- a/repos/spack_repo/builtin/packages/root/package.py
+++ b/repos/spack_repo/builtin/packages/root/package.py
@@ -906,6 +906,10 @@ class Root(CMakePackage):
             # warnings when building against ROOT
             env.unset("MACOSX_DEPLOYMENT_TARGET")
 
+        # https://github.com/root-project/root/issues/18949
+        if "+cxxmodules" in self.spec and "+vc" in self.spec:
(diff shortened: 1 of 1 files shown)


jmcarcell (approved):
I have been able to build with `cxxmodules` on without this 6.38.00 so I guess something has changed.

paulgessinger:
@jmcarcell yeah it changed from 6.38.0 to 6.38.2 for me.

wdconinc:
See https://github.com/spack/spack-packages/pull/3750 for a fix to the inadvertent restriction that has likely prevented us from seeing this issue in spack CI.

wdconinc:
@paulgessinger Does this: https://gitlab.spack.io/spack/spack-packages/-/jobs/20914086#L294 look like the error you got?

paulgessinger:
@wdconinc yes that looks like what I got

wdconinc (approved):

PR #3750, opened 2026-03-09
stacks/hep: don't require `root +mysql +postgres` (restricts to @:6.36)

wdconinc:
This PR removes the `mysql` and `postgres` variants in the `root` package as required by the HEP stack. I was surprised the issue in https://github.com/spack/spack-packages/pull/3748 wasn't apparent in CI, so this PR should now cause us to rebuild with ROOT v6.38.02, and it will likely fail at first due to that issue.

### DIFF
--- a/repos/spack_repo/builtin/packages/root/package.py
+++ b/repos/spack_repo/builtin/packages/root/package.py
@@ -825,11 +825,8 @@ class Root(CMakePackage):
             env.unset("MACOSX_DEPLOYMENT_TARGET")
 
         # https://github.com/root-project/root/issues/18949
-        if "+cxxmodules" in self.spec:
-            if "+vc" in self.spec:
-                env.prepend_path("ROOT_INCLUDE_PATH", self.spec["vc"].prefix.include)
-            if "+veccore" in self.spec:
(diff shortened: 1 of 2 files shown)


wdconinc:
This now indeed builds ROOT v6.38.02 at https://gitlab.spack.io/spack/spack-packages/-/pipelines/1463747 (and may fail there).

jmcarcell (approved):

wdconinc:
@jmcarcell It seems veccore would need the same treatment at vc, see https://gitlab.spack.io/spack/spack-packages/-/jobs/20922008#L327, but the issue is a bit deeper too: where the failure happens, the variable is overwritten with DEFAULT_ROOT_INCLUDE_PATH, which does not include veccore. Since it seems this is going to be deprecated anyway (per https://github.com/root-project/root/issues/18949), I'm going to just disable veccore here in CI.

jmcarcell:
It's possible but that one I don't know since I've never built ROOT with `veccore` and the LCG stacks don't enable it either.

wdconinc:
The remaining CI error (py-tensorflow in ml-darwin-aarch64-mps-build) is in the develop branch as well.

kwryankrattiger (approved):
LGTM

PR #3760, opened 2026-03-09
freeimage: deprecate 3.18.0 (CVE-2024-31570)

wdconinc:
This PR marks `freeimage`, v3.18.0, as deprecated due to CVE-2024-31570 (critical). This is the only version (package is in maintenance mode, no releases since 2018). This package is only used by `opencascade`, where the variant that enables it defaults to off.


wdconinc:
@spackbot fix style

tldahlgren (approved):

PR #3761, opened 2026-03-09
py-pysam: fixed installation by adding py-setuptools version constraint

V-Karch:
Add dependency on `py-setuptools` version constraints for compatibility.

The package won't install with the current package recipe due to the removal of the `pkg-resources` package from `py-setuptools@82.0.0:` I just required all the versions we currently have to use `py-setuptools@:81` so it'll stay at an older `py-setuptools`


repos/spack_repo/builtin/packages/py_pysam/package.py
```diff
@@ -27,6 +27,9 @@ class PyPysam(PythonPackage):
 
     depends_on("c", type="build")  # generated
 
+    # pysam requries pkg-resources which is removed from setuptools in
```
tldahlgren:
```suggestion
    # pysam requires pkg-resources which is removed from setuptools in
```

repos/spack_repo/builtin/packages/py_pysam/package.py
```diff
@@ -27,6 +27,9 @@ class PyPysam(PythonPackage):
...
+    # version 82.0.0
+    depends_on("py-setuptools@:81", when="@:0.21.0", type="build")
     depends_on("py-setuptools@59.0:", when="@0.21:", type="build")
```
tldahlgren: Since the `when` condition is the same, they can be combined:


```suggestion
    depends_on("py-setuptools@59:80", when="@:0.21", type="build")
```

tldahlgren (changes requested):

V-Karch:
@tldahlgren Changes applied! Thanks for the review!

tldahlgren (dismissed):

wdconinc:
@V-Karch Note that this has a merge conflict. Can you take a look?

V-Karch:
@tldahlgren @wdconinc Should be fixed now

repos/spack_repo/builtin/packages/py_pysam/package.py
wdconinc: The when condition was actually not the same here, so these should have been kept separate.

repos/spack_repo/builtin/packages/py_pysam/package.py
```diff
@@ -27,10 +27,10 @@ class PyPysam(PythonPackage):
     version("0.7.7", sha256="c9f3018482eec99ee199dda3fdef2aa7424dde6574672a4c0d209a10985755cc")
 
     depends_on("c", type="build")  # generated
-
     depends_on("python@3.8:", type=("build", "run"))
-
-    depends_on("py-setuptools@59.0:", when="@0.21:", type="build")
+    # pysam requires pkg-resources which is removed from setuptools in
+    # version 82.0.0
+    depends_on("py-setuptools@59:80", when="@:0.23", type="build")
```
wdconinc:
```suggestion
    depends_on("py-setuptools@:80", when="@:0.23", type="build")
    depends_on("py-setuptools@59:", when="@0.21:", type="build")
```

V-Karch:
@wdconinc suggestion applied, thanks for the review

wdconinc (approved):

wdconinc:
@tldahlgren Are you able to give this another look (since you have your change request lock on this)?

V-Karch:
@tldahlgren any updates?

PR #3766, opened 2026-03-09
orca: support arm install

green-br:
ORCA has supported Arm builds on their website.  Provided a mechanism to dynamically work out which versions are available depending on architecture.

Maintainer: @snehring

### DIFF
--- a/repos/spack_repo/builtin/packages/orca/package.py
+++ b/repos/spack_repo/builtin/packages/orca/package.py
@@ -4,27 +4,26 @@
 
 import os
 
+import platform
+
 from spack_repo.builtin.build_systems.generic import Package
 
 from spack.package import *
 
 _versions = {
-    "6.0.1": {
-        "Linux-aarch64": "0699cbccb6dbee66e14e69c4bb1300b35820b4222afdd7371e50aa23fe28be48",
-        "Linux-x86_64": "5e9b49588375e0ce5bc32767127cc725f5425917804042cdecdfd5c6b965ef61",
+    "avx2-6.1.1": {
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/orca/package.py
```diff
@@ -70,12 +85,12 @@ def url_for_version(self, version):
...
-        if self.spec.satisfies("@=avx2-6.0.0"):
-            url = f"file://{os.getcwd()}/orca_{ver_underscored}_linux_x86-64_avx2_shared_openmpi{openmpi_version}.tar.xz"
-
-        return url
+        features = ver_parts[:-1] + ["shared"]
+        feature_text = "_".join(features)
+        orca_arch = "linux_x86-64"
+        if platform.system() == "Linux" and platform.machine() == "aarch64":
+            orca_arch = "linux_arm64"
+        return f"file://{os.getcwd()}/orca_{ver_underscored}_{orca_arch}_{feature_text}_openmpi{openmpi_version}.tar.xz"
```
snehring: This doesn't seem to account for the change in filename in x86_64 avx2 versions after 6.0.0.

green-br: Ah good point - only tested vanilla versions on x86_64 and aarch64 not the avx variation.  I assume we can add the check for avx2 as before.  Let me revise.

green-br: I have now tested using avx2 downloads as well.  Seems to work now after modifying the code slightly.

snehring (approved):
lgtm

alalazo (approved):
Maintainers approved

PR #3774, opened 2026-03-10
acts: add v45.4.0, v45.5.0, v46.0.0

stephenswat:
This commit adds versions 45.4.0, 45.5.0, and 46.0.0 of the ACTS package and a new version with support for a Millipede plugin.

### DIFF
--- a/repos/spack_repo/builtin/packages/acts/package.py
+++ b/repos/spack_repo/builtin/packages/acts/package.py
@@ -50,6 +50,9 @@ class Acts(CMakePackage, CudaPackage):
 
     # Supported Acts versions
     version("main", branch="main")
+    version("46.0.0", commit="a28f778b767fb8fcf9eba5675aecc71e28e30d50")
+    version("45.5.0", commit="1fe226214174c4b18b8d3a77c44a7d09de9c2e47")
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/acts/package.py
```diff
@@ -146,6 +147,9 @@ class Acts(CMakePackage, CudaPackage):
     requires("+svg", when="+traccc")
     requires("+json", when="+traccc")
 
+    variant("mille", default=False, description="Build the Mille plugin")
+    requires("millepede@01-00-00:", when="+mille")
```
wdconinc: Any reason to place this here and not with the other dependencies?
```suggestion
    depends_on("millepede@01-00-00:", when="+mille")
```

stephenswat: Wow I shouldn't maintain Spack specs before my morning coffee it seems. :laughing:

stephenswat:
Sorry for the delay @wdconinc; updated according to your feedback and added v45.5.0.

stephenswat:
Added version 46.0.0.

wdconinc (approved):

PR #3776, opened 2026-03-10
ascent: add catalyst support

albestro:
This PR enables build of Ascent Catalyst implementation via the new `catalyst` variant. I had a quick look and there's no clear consensus on `catalyst` vs `libcatalyst` for variant name, but I hope that package maintainers will eventually end up sharing a single name for that.

I took also the chance to make `fortran` language dependency conditional.


cyrush (approved):
Looks good to me, thanks.

For the variant name, I think catalyst is a good choice.

tldahlgren (approved):

PR #3779, opened 2026-03-10
py-throttler: add py-setuptools version constraint

V-Karch:
Added because the install fails after `py-setuptools@82:` due to the removal of `pkg-resources`


V-Karch:
Opened too soon

PR #3781, opened 2026-03-10
brotli: explcitly set required CMake versions

berquist:
This solves a problem where I need to use an older version of CMake as an env root.


haampie (approved):
The second lowerbound isn't really necessary (10+ years)

PR #3782, opened 2026-03-10
gcc: restrict patchset from #2596 

Chrismarsh:
- restrict the patchset from https://github.com/spack/spack-packages/pull/2596 to only apply to linux as it conflicts with the darwin patches

@haampie @msimberg


haampie (approved):

msimberg (approved):
Thanks @Chrismarsh for noticing this!

PR #3789, opened 2026-03-11
Git: Minor clean-ups

RemiLacroix-IDRIS:
Remove some forgotten checks that were needed for old Git versions which depended on PCRE instead of PCRE2.

### DIFF
--- a/repos/spack_repo/builtin/packages/git/package.py
+++ b/repos/spack_repo/builtin/packages/git/package.py
@@ -193,6 +193,7 @@ class Git(AutotoolsPackage):
             f"--with-curl={spec['curl'].prefix}",
             f"--with-expat={spec['expat'].prefix}",
             f"--with-openssl={spec['openssl'].prefix}",
+            f"--with-libpcre2={spec['pcre2'].prefix}",
             f"--with-zlib={spec['zlib-api'].prefix}",
         ]
 
@@ -202,10 +203,6 @@ class Git(AutotoolsPackage):
         if self.spec.satisfies("+perl"):
             configure_args.append(f"--with-perl={spec['perl'].command.path}")
 
-        if self.spec.satisfies("^pcre"):
-            configure_args.append(f"--with-libpcre={spec['pcre'].prefix}")
-        if self.spec.satisfies("^pcre2"):
-            configure_args.append(f"--with-libpcre2={spec['pcre2'].prefix}")
         if self.spec.satisfies("+tcltk"):
             configure_args.append(f"--with-tcltk={self.spec['tk'].prefix.bin.wish}")
         else:


repos/spack_repo/builtin/packages/git/package.py
jennfshr, line 1: Are you dropping the support for PCRE variant altogether with this change - i.e., forcing the configure option `--with-libpcre={spec['pcre'].prefix}`  without testing whether the spec satisfies it?

If so, I'd first argue that there's no strong justification I can find for that decision, and even so, it's better placed in the initialization of `configure_args`. What's the motivation? If not, I think we have to use a conditional to check the spec before appending `configure_args`.  

I agree, it makes sense to cleanup support for PCRE v1, since that was deprecated almost a decade ago, but am concerned with implementation and motivation for this change. I may just be misunderstanding what you're doing. Can you explain?

jennfshr (commented):
Please review my comment and let's evaluate this PR more carefully.

repos/spack_repo/builtin/packages/git/package.py
RemiLacroix-IDRIS: The conditional `depend_on` for PCRE v1 was removed from the Git package in c379b95 so the conditional check for `if self.spec.satisfies("^pcre"):` will always return false.

I will reword my commit message to make that clearer.

RemiLacroix-IDRIS:
@jennfshr PR updated to clarify the commit message. I hope this clears things up.

repos/spack_repo/builtin/packages/git/package.py
jennfshr: I see, appreciate the clarification. That said, since pcre2 is a requirement with no variant to toggle it on/off, then I wonder whether it would be cleaner to add it to this initialization of the `configure_args` here? It seems out of place leaving it in with the spec contains conditional checks:
```python
        configure_args = [
            f"--with-curl={spec['curl'].prefix}",
            f"--with-expat={spec['expat'].prefix}",
            f"--with-openssl={spec['openssl'].prefix}",
            f"--with-libpcre2={spec['pcre2'].prefix}", << include it here?
            f"--with-zlib={spec['zlib-api'].prefix}",
        ]
```

RemiLacroix-IDRIS: Indeed, fixed!

jennfshr:
Cool, thank you! I’m testing it locally now.

From: Rémi Lacroix ***@***.***>
Date: Thursday, March 12, 2026 at 8:19 AM
To: spack/spack-packages ***@***.***>
Cc: Green, Jennifer Kathleen ***@***.***>, Mention ***@***.***>
Subject: [EXTERNAL] Re: [spack/spack-packages] Git: Minor clean-ups (PR #3789)


@RemiLacroix-IDRIS commented on this pull request.

On repos/spack_repo/builtin/packages/git/package.py<https://github.com/spack/spack-packages/pull/3789#discussion_r2925002339>:

Indeed, fixed!

jennfshr (approved):
This tested fine on my linux cluster against the system gcc compiler. Approved.

tldahlgren (approved):

PR #3790, opened 2026-03-11
Axom: update to v0.13.0

gberg617:
- Update Axom to v0.13.0
- Add profiling support for release versions
- Add "longid" variant

### DIFF
--- a/repos/spack_repo/builtin/packages/axom/package.py
+++ b/repos/spack_repo/builtin/packages/axom/package.py
@@ -53,6 +53,7 @@ class Axom(CachedCMakePackage, CudaPackage, ROCmPackage):
 
     version("main", branch="main")
     version("develop", branch="develop")
+    version("0.13.0", tag="v0.13.0", commit="d00f6c66ef390ad746ae840f1074d982513611ac")    
     version("0.12.0", tag="v0.12.0", commit="297544010a3dfb98145a1a85f09f9c648c00a18c")
     version("0.11.0", tag="v0.11.0", commit="685960486aa55d3a74a821ee02f6d9d9a3e67ab1")
     version("0.10.1", tag="v0.10.1", commit="6626ee1c5668176fb64dd9a52dec3e8596b3ba6b")
@@ -101,7 +102,6 @@ class Axom(CachedCMakePackage, CudaPackage, ROCmPackage):
     variant(
         "profiling",
         default=False,
-        when="@develop",
         description="Build with hooks for Adiak/Caliper performance analysis",
     )
 
@@ -117,6 +117,8 @@ class Axom(CachedCMakePackage, CudaPackage, ROCmPackage):
 
     varmsg = "Build development tools (such as Sphinx, Doxygen, etc...)"
     variant("devtools", default=False, description=varmsg)
+
+    variant("longid", default=True, when="@0.13.0:", description="Build with 64-bit index type")
 
     # -----------------------------------------------------------------------
     # Dependencies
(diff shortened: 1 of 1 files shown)


gberg617:
@white238 @bmhan12

repos/spack_repo/builtin/packages/axom/package.py
```diff
@@ -101,7 +102,6 @@ class Axom(CachedCMakePackage, CudaPackage, ROCmPackage):
     variant(
         "profiling",
         default=False,
-        when="@develop",
```
becker33, line 104: Should this be `when="@0.13:"` now?

bmhan12: I think we're deprecating `+profiling` variant in favor of `+adiak` and `+caliper` (`when="@:0.12.0"`): https://github.com/llnl/axom/blob/30777634710b772dab3b40637dbf8be15c97466e/scripts/spack/packages/axom/package.py#L121-L127

@kennyweiss, can you double-check I got that right?

becker33: That's fine, but neither of those variants are in the package yet. If `profiling` was only ever supported on the develop branch and not in a release, then this definitely shouldn't be made available on all versions the way it is here.

kennyweiss: I agree with @becker33 -- Starting in version 0.13, we want to use the `adiak` and `caliper` variants, and stop using the `profiling` variant.  I'd love a clean way to mark a variant as deprecated with a `when=` clause  (https://github.com/spack/spack/issues/38829).

I only marked it as @develop in  https://github.com/llnl/axom/pull/1740 since we had not yet release axom@0.13

repos/spack_repo/builtin/packages/axom/package.py
```diff
@@ -118,6 +118,8 @@ class Axom(CachedCMakePackage, CudaPackage, ROCmPackage):
     varmsg = "Build development tools (such as Sphinx, Doxygen, etc...)"
     variant("devtools", default=False, description=varmsg)
 
+    variant("longid", default=True, when="@0.13.0:", description="Build with 64-bit index type")
```
kennyweiss, line 121: @gberg617  -- we already have an `int64` variant in Axom's local spack recipe -- https://github.com/llnl/axom/blob/30777634710b772dab3b40637dbf8be15c97466e/scripts/spack/packages/axom/package.py#L153 -- 

Let's use that instead of a new `longid` variant since codes are already using that.

`int64` appears to be the conventional way to do this, e.g.
https://github.com/search?q=repo%3Aspack%2Fspack-packages%20int64&type=code

and I think we modeled ours off of `hypre`: https://github.com/spack/spack-packages/blob/107d3dbda28f475b2f0f241a407fc92787d821c7/repos/spack_repo/builtin/packages/hypre/package.py#L89

bmhan12: `int64` makes sense to me, best to keep the variant conventions consistent across packages if we can.

kennyweiss (changes requested):
I don't think we should add an axom@0.13 version to spack-packages without the `components` variant (which allows selectively enabling/disabling individual axom components)

See: 
* https://github.com/llnl/axom/blob/30777634710b772dab3b40637dbf8be15c97466e/scripts/spack/packages/axom/package.py#L129-L138

@gberg617 if you need this for local development, and don't want to take responsibility for all of the updates, would a local axom package that derives from the one in spack-packages and adds the 0.13 version work for you in the meantime?

gberg617:
> I don't think we should add an axom@0.13 version to spack-packages without the `components` variant (which allows selectively enabling/disabling individual axom components)
> 
> See:
> 
> * https://github.com/llnl/axom/blob/30777634710b772dab3b40637dbf8be15c97466e/scripts/spack/packages/axom/package.py#L129-L138
> 
> @gberg617 if you need this for local development, and don't want to take responsibility for all of the updates, would a local axom package that derives from the one in spack-packages and adds the 0.13 version work for you in the meantime?

@kennyweiss I only need these changes to build another package that depends on Axom. I can apply the changes locally rather than push them here if you and @bmhan12 prefer this approach.

gberg617:
Closing this PR in favor of putting changes in a local package file.

kennyweiss:
Thanks @gberg617 -- we'll update axom's spack recipe to reflect recent changes to our local package (https://github.com/llnl/axom/blob/develop/scripts/spack/packages/axom/package.py)

PR #3792, opened 2026-03-11
kokkos: allow other LLVM-based compilers for ROCm

rbberger:
Instead of always forcing hipcc, allow kokkos to be compiled directly with either llvm or llvm-amdgpu (rocmcc).

The main motivation for this is to allow building directly with amdclang++ instead of hipcc to avoid different inlining defaults between hipcc and amdclang++. We've gotten the recommendation to move toward using amdclang++ directly. This enables this without also forcing CMake language support.

### DIFF
--- a/repos/spack_repo/builtin/packages/kokkos/package.py
+++ b/repos/spack_repo/builtin/packages/kokkos/package.py
@@ -464,10 +464,7 @@ class Kokkos(CMakePackage, CudaPackage, ROCmPackage):
             options.append(self.define("CMAKE_CXX_COMPILER", self.kokkos_cxx))
         elif "+rocm" in self.spec:
             if "+cmake_lang" in self.spec:
-                if (
-                    self.spec.satisfies("%cxx=clang")
-                    or self.spec.satisfies("%cxx=rocmcc")
-                ):
+                if self.spec.satisfies("%cxx=clang") or self.spec.satisfies("%cxx=rocmcc"):
                     options.append(self.define("CMAKE_HIP_COMPILER", self.compiler.cxx))
                 else:
                     options.append(
@@ -483,10 +480,7 @@ class Kokkos(CMakePackage, CudaPackage, ROCmPackage):
                     )
                 )
                 options.append(self.define("CMAKE_HIP_EXTENSIONS", False))
-            elif not (
-                self.spec.satisfies("%cxx=clang")
-                or self.spec.satisfies("%cxx=rocmcc")
-            ):
+            elif not (self.spec.satisfies("%cxx=clang") or self.spec.satisfies("%cxx=rocmcc")):
                 options.append(self.define("CMAKE_CXX_COMPILER", self.spec["hip"].hipcc))
             options.append(self.define("Kokkos_ENABLE_ROCTHRUST", True))
         elif "+cuda" in self.spec and "+cmake_lang" in self.spec:
--- a/repos/spack_repo/builtin/packages/kokkos/package.py
+++ b/repos/spack_repo/builtin/packages/kokkos/package.py
@@ -464,7 +464,10 @@ class Kokkos(CMakePackage, CudaPackage, ROCmPackage):
             options.append(self.define("CMAKE_CXX_COMPILER", self.kokkos_cxx))
         elif "+rocm" in self.spec:
             if "+cmake_lang" in self.spec:
-                if self.spec.satisfies("%cxx=clang"):
+                if (
+                    self.spec.satisfies("%cxx=clang")
+                    or self.spec.satisfies("%cxx=rocmcc")
+                ):
                     options.append(self.define("CMAKE_HIP_COMPILER", self.compiler.cxx))
                 else:
                     options.append(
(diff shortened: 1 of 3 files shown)


rbberger:
@spackbot run pipeline.

haampie:
Is this still all based on CMAKE_CXX_COMPILER or should it be CMAKE_HIP_COMPILER?

rbberger:
@haampie this is still based on `CMAKE_CXX_COMPILER`. `CMAKE_HIP_COMPILER` is only set when `+cmake_lang`. But I also added the necessary logic for that case as well.

JBludau (approved):
AFAIU this is fine.
For our support status of things we test, see our CDash https://my.cdash.org/index.php?project=Kokkos do we know if an where we would want to increase coverage?
amdclang++ is tested by us, see https://github.com/kokkos/kokkos/blob/9c5b400e6fb69712ce627e701cce0827d6c74c5f/.gitlab/olcf-gitlab-ci.yml#L26
~Crayclang does fail to compile our tests and we have very limited experience with it in user codes, see https://github.com/kokkos/kokkos/blob/9c5b400e6fb69712ce627e701cce0827d6c74c5f/.gitlab/olcf-gitlab-ci.yml#L45. Should we print a warning about this?~

tldahlgren:
@yasahi-hpc

Added missing maintainer

rbberger:
@spackbot fix style

repos/spack_repo/builtin/packages/kokkos_fft/package.py
```diff
@@ -75,7 +75,11 @@ def cmake_args(self):
             self.define("KokkosFFT_ENABLE_ONEMKL", self.spec.satisfies("device_backend=onemkl")),
         ]
 
-        if self.spec.satisfies("^kokkos+rocm"):
+        if self.spec.satisfies("^kokkos+rocm") and not (
```
tpadioleau, line 78: Is Kokkos also turning `cce`/`clang`/`rocmcc` into HIP language mode for dependent projects ? If I remember correctly, by default, when these compilers compile a file `.cpp` they use the C++ language mode.

repos/spack_repo/builtin/packages/kokkos/package.py
```diff
@@ -464,7 +464,7 @@ def cmake_args(self):
             options.append(self.define("CMAKE_CXX_COMPILER", self.kokkos_cxx))
         elif "+rocm" in self.spec:
             if "+cmake_lang" in self.spec:
-                if self.spec.satisfies("%cxx=clang"):
+                if self.spec.satisfies("%cxx=clang") or self.spec.satisfies("%cxx=rocmcc"):
```
tpadioleau, line 467: Should we also add `cce` ?

rbberger: I originally added cce. But there is no tests for it and I'm not expecting anyone to use it in the near future.

repos/spack_repo/builtin/packages/kokkos_fft/package.py
rbberger: It's explicitly adding the right flags in CXX mode. That's the point, since HIP language mode is already covered with +cmake_lang. This allows projects that don't use HIP language to use non-hipcc LLVMs.

repos/spack_repo/builtin/packages/kokkos/package.py
tpadioleau: I see, ok for me.

For a future PR, should we remove the logic about `cce` elsewhere then ?

repos/spack_repo/builtin/packages/kokkos_fft/package.py
```diff
@@ -75,7 +75,11 @@ def cmake_args(self):
...
+            self.spec.satisfies("^kokkos %cxx=clang")
+            or self.spec.satisfies("^kokkos %cxx=cce")
```
tpadioleau:
```suggestion
```

If the Kokkos package does not handle it, let's remove it.

repos/spack_repo/builtin/packages/kokkos_kernels/package.py
```diff
@@ -223,7 +223,11 @@ def cmake_args(self):
         ]
 
         options.append(self.define("Kokkos_ROOT", spec["kokkos"].prefix))
-        if spec.satisfies("^kokkos+rocm"):
+        if spec.satisfies("^kokkos+rocm") and not (
+            spec.satisfies("^kokkos %cxx=clang")
+            or spec.satisfies("^kokkos %cxx=cce")
```
tpadioleau:
```suggestion
```

repos/spack_repo/builtin/packages/kokkos_kernels/package.py
tpadioleau, line 1: I think the changes are logically good but shouldn't we move it inside `kokkos_cxx` ? Do we have a good reason to separate the case of `nvcc_wrapper` from `hipcc`.

rbberger: I would like to get this one in to enable my work at LANL. Further cleaning up or even bigger structural changes should go into another PR.

rbberger:
@spackbot fix style.

tpadioleau (approved):
Ok for me, I think we should later discuss:
- how to ease passing compilers to downstream projects
- the support status of `cce` in the packages

lucbv (approved):

tldahlgren (approved):

PR #3795, opened 2026-03-12
Unrestrict raja and umpire versions when using newer hiop versions

chapman39:
When using a hiop version newer than 1.1.1, we should not be limited to using only raja/ umpire version 2024.07.0. I have tested using raja/ umpire version 2025.09.0 with hiop@1.1.1 for a few months now and have not experienced any issues.

@tepperly


pelesh:
@chapman39, you are absolutely right. As far as I remember, the restriction was because of some issue with spack packaging. I don't remember having a problem with building HiOp with newer RAJA/Umpire versions by hand.

pelesh (approved):

RikkiButler20 (approved):
Looks good to me.

PR #3796, opened 2026-03-12
go-sh: add v3.13.0

mcmehrtens:
- Added `sh` v3.13.0
- Updated Go dependency constraints for each version
- Starting with v3.13.0, `shfmt` no longer uses an ldflags-settable `main.version` variable (see https://github.com/mvdan/sh/issues/1284). Instead, the version is set by Go at build time using VCS (Git) metadata. This is why v3.13.0 references the Git repo rather than the release tarball.
- Updated Go build targets to use module paths instead of file paths


tldahlgren (approved):
Confirmed commit+tag.

PR #3799, opened 2026-03-12
raxml-ng: add 2.0.0 and previous versions

stelzch:


tldahlgren (approved):
Confirmed commits and  use of new variant.

PR #3800, opened 2026-03-12
abinit: Fix recipe and bump versions

marcsrt:
Hi,

This MR brings 2 changes to `abinit` recipe.

First, it fixes the recipe by removing the requirement of `mvapich +cuda` when `+gpu_aware_mpi` is enabled. The `cuda` variant has been removed from `mvapich` by #1864, hence this recipe has likely been broken ever since.

Then, it adds two new versions of ABINIT, 10.6.3 and 10.6.5.


downloadico:
These changes look good to me.

tldahlgren (approved):
Confirmed the version sha256s.

PR #3801, opened 2026-03-12
Update paraview patch conditions for VTK-m version 5.13.2

bissantz:
Also required for newer gcc compilers. Since, it is not a compiler relevant patch and it is fixed in the upstream vtkm project, it should always be patched.

And still required for paraview version 5.13.3.


vicentebolea (approved):
thanks for the contrib

PR #3812, opened 2026-03-12
r: add java variant

drkrynstrng:
R does not depend on Java, but some R packages do. This change adds a java variant, disabled by default, so that Java support can be enabled only if needed. This allows removing the Java dependency, which is not needed in most cases.

R docs for reference: https://cran.r-project.org/doc/manuals/r-release/R-admin.html#Java-support


drkrynstrng:
@spackbot fix style

wdconinc:
Are there any R packages in spack which may have assumed Java support in the R install and now need to be adapted as well?

wdconinc:
> Are there any R packages in spack which may have assumed Java support in the R install and now need to be adapted as well?

Never mind, I should learn to scroll past the first file, clearly :-)

repos/spack_repo/builtin/packages/r/package.py
```diff
@@ -185,6 +186,11 @@ def configure_args(self):
             config_args.append("--without-tcltk")
             config_args.append("--without-x")
 
+        if "+java" in spec:
```
wdconinc: This could be written with `self.enable_or_disable("java")` too, I think.

repos/spack_repo/builtin/packages/r_xlconnect/package.py
```diff
@@ -25,15 +25,15 @@ class RXlconnect(RPackage):
...
     depends_on("r-rjava@1.0-1:", type=("build", "run"), when="@1.0.6:")
-    depends_on("java@6:")
-    depends_on("java@8:11", when="@1.0.1:")
-    depends_on("java@8:17", when="@1.0.5:")
-
     depends_on("r-xlconnectjars@0.2-9", type=("build", "run"), when="@0.2-11")
     depends_on("r-xlconnectjars@0.2-12", type=("build", "run"), when="@0.2-12")
     depends_on("r-xlconnectjars@0.2-14", type=("build", "run"), when="@0.2-14")
     depends_on("r-xlconnectjars@0.2-15", type=("build", "run"), when="@0.2-15")
+    depends_on("r+java", type=("build", "run"))
```
wdconinc, line 28: Generally the R packages have the R dependencies listed before  the other package dependencies. Can we keep that order here too?

repos/spack_repo/builtin/packages/r/package.py
drkrynstrng: Yes, updated.

repos/spack_repo/builtin/packages/r_xlconnect/package.py
drkrynstrng: Okay, updated to list R and Java dependencies first, followed by R package dependencies.

drkrynstrng:
I missed the Java dependency for R package dismo. I think I've updated all the R packages with Java dependencies now.

wdconinc (approved):
Looks good to me! Thanks!

PR #3818, opened 2026-03-12
kokkos: disable IMPL_HIP_MALLOC_ASYNC with ROCm 7

rbberger:
This is enforced starting in Kokkos 5.0.1. Earlier versions defaulted to enable it when ROCm was enabled.

### DIFF
--- a/repos/spack_repo/builtin/packages/kokkos/package.py
+++ b/repos/spack_repo/builtin/packages/kokkos/package.py
@@ -483,6 +483,11 @@ class Kokkos(CMakePackage, CudaPackage, ROCmPackage):
             else:
                 options.append(self.define("CMAKE_CXX_COMPILER", self.spec["hip"].hipcc))
             options.append(self.define("Kokkos_ENABLE_ROCTHRUST", True))
+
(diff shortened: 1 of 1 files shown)


cedricchevalier19 (approved):
We also could have used the `alloc_async` variant like for cuda, but as the support in Kokkos is very broken since the new semantic in Rocm, I think this solution to always disable is better.

rbberger:
@cedricchevalier19 I was considering `alloc_async`, but in the end since we haven't exposed it so far for HIP, this seemed like the cleaner solution for now.

JBludau (dismissed):
Let me first check with core, if we want to fix it in a 4.7.03 and not just in spack

cedricchevalier19:
> Let me first check with core, if we want to fix it in a 4.7.03 and not just in spack

I do not see any reason to wait from core for the Spack recipe. The goal is to prescribe the config, and not always rely on Kokkos default (especially for older versions).

JBludau:
I don't necessarily like that our CMake and spack are more disjunct with this, but if you think this is the best way forward, my review can be dismissed

kwryankrattiger (approved):
Seems to be consensus from reviewers now.

rbberger:
@kwryankrattiger I would merge.

PR #3819, opened 2026-03-12
nlopt: set python cmake option based on variant

chapman39:
if `~python` the cmake option needs to be set, otherwise it doesn't do anything.

https://github.com/stevengj/nlopt/blob/master/CMakeLists.txt#L41


cessenat (approved):
Good catch, thanks.

chapman39:
@spackbot style

repos/spack_repo/builtin/packages/nlopt/package.py
```diff
@@ -60,6 +60,8 @@ def cmake_args(self):
             self.define_from_variant("BUILD_SHARED_LIBS", "shared"),
             self.define_from_variant("NLOPT_OCTAVE", "octave"),
             self.define_from_variant("NLOPT_CXX", "cxx"),
+            self.define_from_variant("NLOPT_PYTHON", "python"),
+            self.define_from_variant("NLOPT_GUILE", "guile")
```
chapman39:
```suggestion
            self.define_from_variant("NLOPT_GUILE", "guile"),
```

cessenat (approved):

alalazo (approved):

PR #3821, opened 2026-03-13
py-pyproj ver3.7.2 needs newer setuptools

Sinan81:

### DIFF
--- a/repos/spack_repo/builtin/packages/py_pyproj/package.py
+++ b/repos/spack_repo/builtin/packages/py_pyproj/package.py
@@ -40,6 +40,7 @@ class PyPyproj(PythonPackage):
 
     # In pyproject.toml
     with default_args(type="build"):
+        depends_on("py-setuptools@77.0.1:", when="@3.7.2:")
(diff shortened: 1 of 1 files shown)


Sinan81:
https://github.com/pyproj4/pyproj/blob/main/pyproject.toml

repos/spack_repo/builtin/packages/py_pyproj/package.py
```diff
@@ -40,6 +40,7 @@ class PyPyproj(PythonPackage):
 
     # In pyproject.toml
     with default_args(type="build"):
+        depends_on("py-setuptools@77.0.1:", when="@3.7.2:")
```
adamjstewart, line 44:
```suggestion
        # https://github.com/pyproj4/pyproj/pull/1575
        depends_on("py-setuptools@77.0.1:", when="@3.7.2:")
```
It isn't documented in 3.7.2, but is needed in 3.7.2+

adamjstewart (approved):

PR #3833, opened 2026-03-13
py-html5lib: restrict py-setuptools to @:81

V-Karch:
Doing this because this package tries to import pkg-resources which is removed in py-setuptools@82:

And of course it won't install without using this older setuptools or adding new versions


V-Karch:
Accidentally deleted the branch temporarily

PR #3836, opened 2026-03-13
chapel: add v2.8

arifthpe:
Add Chapel version 2.8, which was released March 12th.

Changes:
- Add 2.8
- Deprecate 2.5
- Add a new version of an existing patch for 2.8
- Refactor expression of ROCm and CUDA requirements for clarity
- Update LLVM and ROCm dependencies for 2.8
- Drop @e-kayrakli as a maintainer (with his permission given offline)

Replaces and is based off of https://github.com/spack/spack-packages/pull/3692, due to it taking longer for us to see through than @e-kayrakli has available time for at the moment.

### DIFF
--- a/repos/spack_repo/builtin/packages/chapel/package.py
+++ b/repos/spack_repo/builtin/packages/chapel/package.py
@@ -699,20 +699,10 @@ class Chapel(AutotoolsPackage, CudaPackage, ROCmPackage):
 
         # Undo spack compiler wrappers:
         # the C/C++ compilers must work post-install
-        if self.spec.satisfies("+rocm llvm=spack"):
-            env.set(
-                "CHPL_LLVM_CONFIG",
-                join_path(self.spec["llvm-amdgpu"].prefix, "bin", "llvm-config"),
-            )
-            real_cc = join_path(self.spec["llvm-amdgpu"].prefix, "bin", "clang")
-            real_cxx = join_path(self.spec["llvm-amdgpu"].prefix, "bin", "clang++")
-
-            # +rocm appears to also require a matching LLVM host compiler to guarantee linkage
-            env.set("CHPL_HOST_COMPILER", "llvm")
-            env.set("CHPL_HOST_CC", real_cc)
-            env.set("CHPL_HOST_CXX", real_cxx)
-
-        elif self.spec.satisfies("llvm=spack"):
+        # If we ever switch back to using llvm-amdgpu for some ROCm versions,
+        # we will need a separate case here to use the binaries it provides
+        # rather than those of vanilla LLVM.
+        if self.spec.satisfies("llvm=spack"):
             env.set("CHPL_LLVM_CONFIG", join_path(self.spec["llvm"].prefix, "bin", "llvm-config"))
             real_cc = join_path(self.spec["llvm"].prefix, "bin", "clang")
             real_cxx = join_path(self.spec["llvm"].prefix, "bin", "clang++")
--- a/repos/spack_repo/builtin/packages/chapel/package.py
+++ b/repos/spack_repo/builtin/packages/chapel/package.py
@@ -63,11 +63,17 @@ class Chapel(AutotoolsPackage, CudaPackage, ROCmPackage):
 
     version("main", branch="main")
 
+    version(
+        "2.8.0",
+        url="https://chapel-lang.org/tmp/chapel-2.8.0.tar.gz",
+        sha256="3e5e8da6c56"
+    )
+
     version("2.7.0", sha256="5e3269babdae334c80fc3f25114698fdfe53e84ea06626af22d2b54eeb75bee6")
     version("2.6.0", sha256="e469c35be601cf1f59af542ab885e8a14aa2b087b79af0d5372a4421976c74b6")
-    version("2.5.0", sha256="020220ca9bf52b9f416e9a029bdc465bb1f635c1e274c6ca3c18d1f83e41fce1")
 
     with default_args(deprecated=True):
+        version("2.5.0", sha256="020220ca9bf52b9f416e9a029bdc465bb1f635c1e274c6ca3c18d1f83e41fce1")
         version("2.4.0", sha256="a51a472488290df12d1657db2e7118ab519743094f33650f910d92b54c56f315")
         version("2.3.0", sha256="0185970388aef1f1fae2a031edf060d5eac4eb6e6b1089e7e3b15a130edd8a31")
         version("2.2.0", sha256="bb16952a87127028031fd2b56781bea01ab4de7c3466f7b6a378c4d8895754b6")
--- a/repos/spack_repo/builtin/packages/chapel/package.py
+++ b/repos/spack_repo/builtin/packages/chapel/package.py
@@ -519,6 +519,10 @@ class Chapel(AutotoolsPackage, CudaPackage, ROCmPackage):
             depends_on("hsa-rocr-dev@6.0:6.2")
             depends_on("hip@6.0:6.2")
         with when("@2.8:"):
+            # ROCm 6.4 is specifically prohibited. Although Chapel allows it
+            # (see https://github.com/chapel-lang/chapel/pull/28220), the
+            # support is untested; being stricter here to reduce the amount of
+            # variables in getting the Spack package to work.
             depends_on("hsa-rocr-dev@6.0:6.3,7")
             depends_on("hip@6.0:6.3,7")
 
--- a/repos/spack_repo/builtin/packages/chapel/package.py
+++ b/repos/spack_repo/builtin/packages/chapel/package.py
@@ -512,17 +512,33 @@ class Chapel(AutotoolsPackage, CudaPackage, ROCmPackage):
     with when("+rocm"):
         conflicts("llvm=none", msg="ROCm support requires building with LLVM")
 
-        # Chapel restricts the allowable ROCm versions
-        depends_on("hsa-rocr-dev@6.0:6.2")
-        depends_on("hip@6.0:6.2")
-
         conflicts("@:2.1", msg="ROCm support in spack requires Chapel 2.2.0 or later")
 
-        # This is the case because the package only supports ROCm 6, and Chapel
-        # requires bundled LLVM for that version.
-        # TODO: Modify this constrant and message if/when Chapel supports an
-        # additional ROCm version without tht requirement.
-        requires("llvm=bundled", msg="Chapel ROCm support requires llvm=bundled")
+        # Chapel restricts the allowable ROCm versions
+        with when("@:2.7"):
+            depends_on("hsa-rocr-dev@6.0:6.2")
+            depends_on("hip@6.0:6.2")
+        with when("@2.8:"):
+            depends_on("hsa-rocr-dev@6.0:6.3,7")
+            depends_on("hip@6.0:6.3,7")
+
+        # Chapel requires using the (patched) bundled LLVM for some versions
+        # of ROCm.
+        # Switching off of hsa-rocr-dev version rather than both it and hip,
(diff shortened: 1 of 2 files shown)


repos/spack_repo/builtin/packages/chapel/package.py
```diff
@@ -484,36 +486,67 @@ class Chapel(AutotoolsPackage, CudaPackage, ROCmPackage):
...
+            "^llvm targets=all",
+            msg="llvm=spack +cuda requires LLVM support the nvptx target",
+            when="llvm=spack",
+        )
+        conflicts(
+            "^llvm@20",
+            when="@:2.5",
+            msg="Chapel through 2.5 does not support Nvidia GPUs with LLVM 20, see "
+            "https://github.com/chapel-lang/chapel/issues/27273",
+        )
```
bonachea, line 508: At the time the referenced issue was open (and later closed) LLVM 20 was the latest. However LLVM now has releases 21 and 22. 

Until/unless we have testing showing otherwise, we should probably conservatively assume that Chapel 2.5 +cuda is also broken with newer versions of LLVM.



```suggestion
        conflicts(
            "^llvm@20:",
            when="@:2.5",
            msg="Chapel through 2.5 does not support Nvidia GPUs with LLVM 20+, see "
            "https://github.com/chapel-lang/chapel/issues/27273",
        )

```

bonachea (changes requested):
Thanks for refactoring the GPU support logic, it looks much more readable and maintainable now!

I realize this is still a draft PR, but I added an initial nitpick based on reading the diffs.

repos/spack_repo/builtin/packages/chapel/package.py
arifthpe: I think this won't affect concretization as Chapel 2.5 doesn't (and won't) support LLVM newer than 20 (https://github.com/spack/spack-packages/blob/78750ec2f60d9c4b2a4d50b722379f713d4479c7/repos/spack_repo/builtin/packages/chapel/package.py#L581), however I think tightening this restriction is more correct anyways so will change it.

repos/spack_repo/builtin/packages/chapel/package.py
```diff
@@ -484,36 +486,67 @@ class Chapel(AutotoolsPackage, CudaPackage, ROCmPackage):
...
+
+        conflicts("@:2.1", msg="ROCm support in spack requires Chapel 2.2.0 or later")
+
+        # Chapel restricts the allowable ROCm versions
+        with when("@:2.7"):
+            depends_on("hsa-rocr-dev@6.0:6.2")
+            depends_on("hip@6.0:6.2")
+        with when("@2.8:"):
+            depends_on("hsa-rocr-dev@6.0:6.3,7")
+            depends_on("hip@6.0:6.3,7")
```
bonachea, line 527: It appears ROCm 6.4 excluded here, which is _odd_ but deliberate?

<https://chapel-lang.org/docs/technotes/gpu.html#requirements> at least seems to imply it should be permitted:
> Only with ROCm 6.3 and above, CHPL_LLVM=system can be used as long as it is at least LLVM 21+.

However when I loosen this constraint I get an error from configure:
```
Warning: Chapel requires ROCm versions 5.0 to 5.4.x, 6.0 to 6.3.x, or 7.0 to 7.x, detected version 6.4.3 on system.

Error: Cannot target AMD GPUs with ROCm 6.3 without CHPL_LLVM=bundled or LLVM 21+
```

So maybe the problem is "with ROCm 6.3 and above" in the documentation should instead read "with ROCm 6.3.x"?

I suggest a comment to clarify/document this deliberate omission:

```suggestion
        with when("@2.8:"):
            # ROCm 6.4 is specifically prohibited
            depends_on("hsa-rocr-dev@6.0:6.3,7")
            depends_on("hip@6.0:6.3,7")
```

repos/spack_repo/builtin/packages/chapel/package.py
bonachea, line 1: It seems Chapel's `configure` script has recently(?) started enforcing requirements on the `CHPL_*` envvars. As a result, many of my `spack install` failed attempts are now dropping  dead during Spack's `configure` step instead of the `build` step where such failures used to occur.

The spackage `build` method generates output showing the `CHPL_*` variable settings, which is critically important to understanding many failures, but we may no longer reach that output if `configure` is enforcing the requirements before `build`.

I would like to see the following method added to improve this situation:
```python
    @run_before("configure") 
    def print_configuration(self):
        # print all the explicitly set CHPL_* config variables to assist in debugging
        for var in self.chpl_env_vars:
            tty.info(var + "=" + os.getenv(var, "<unset>"))
```

I've tested this [in my local copy](https://github.com/bonachea/spack-packages/commit/6d09ad3cc9ff0c6c568a22134814909c89b8c334) and it resolves this spackage debugging concern.

bonachea (changes requested):

repos/spack_repo/builtin/packages/chapel/package.py
bonachea: Based on comments in Slack from @jabraham17, we need to rip out this dependency on `llvm-amdgpu` and instead ensure we are passing the vanilla LLVM paths here.

Unfortunately I believe the `ROCmPackage` superclass we're using will [still force a dependency on the `llvm-amdgpu` package for `+rocm`](https://github.com/spack/spack-packages/blob/develop/repos/spack_repo/builtin/build_systems/rocm.py#L156), which is now completely unused by Chapel. It would be nice if there was a way to disable that dependency, because it will add unnecessary time/space churn for the end user in finding/installing dead software. However I don't know of any way to do that short of removing/replacing (or somehow patching) the `ROCmPackage` superclass.

bonachea (changes requested):
`+rocm llvm=spack` support is currently unusable. 

For `llvm=spack` the spackage is forcing Chapel to use `llvm-amdgpu` as the compiler (`CHPL_*_{CC,CXX}` and `CHPL_LLVM_CONFIG`), but the `llvm-amdgpu` packages corresponding to supported ROCm versions do not advertise an LLVM version satisfying Chapel's corresponding LLVM version constraints.

repos/spack_repo/builtin/packages/chapel/package.py
arifthpe: > It appears ROCm 6.4 excluded here, which is odd but deliberate?

It is intentional, see https://github.com/chapel-lang/chapel/pull/28220/changes#diff-4c6fe7e6f01dadbf3eba9ac00e66c1c6546abaed9ed14bf5da46f278c7892bb4R553 . We don't test ROCm 6.4 though it _might_ work. As of that PR Chapel will let you try a version we don't claim support for, with a warning instead of error. However for the Spack package I think it makes more sense to force use of tested and supported versions.

> So maybe the problem is "with ROCm 6.3 and above" in the documentation should instead read "with ROCm 6.3.x"?

To my understanding that statement is accurate about the LLVM version requirements -- though maybe you're saying it's confusing because it implies support for _all_ ROCm versions 6.3 and above? I don't think changing it to "with ROCm 6.3.x" would be adequate, because this requirement also applies to ROCm 7.x.

IMO the error message you got stating "with ROCm 6.3" is more problematic and should say "with ROCm 6.3 and above", because that error message doesn't make sense with the version you were using (6.4).

repos/spack_repo/builtin/packages/chapel/package.py
arifthpe: Cherry-picked

repos/spack_repo/builtin/packages/chapel/package.py
bonachea: I think I agree with your position on the Chapel documentation and error messages, but neither are things we'd potentially change in this repo.

As far as the Spackage goes, I still think it's worth adding the comment suggested above to record the fact that the omission of ROCm 6.4 was deliberate (possibly with a reference to Chapel PR 28220)

repos/spack_repo/builtin/packages/chapel/package.py
arifthpe: In theory we could literally delete it from the [dependencies `dict`](https://spack.readthedocs.io/en/latest/package_api.html#spack.package.Package.dependencies), however that would certainly void the warranty and I doubt Spack maintainers would accept it. I would rather leave the unnecessary dep than do that.

Another option is to remove it as a dependence, and just duplicate or reference [the things it provides](https://github.com/spack/spack/blob/develop/lib/spack/docs/build_systems/rocmpackage.rst). The only one of those that sounds like an ongoing burden is `amdgpu_target`/the `amdgpu_targets` list, but we could just create our own `amdgpu_targets = ROCmPackage.amdgpu_targets`, I think. However, there is also a dependency on `llvm-amdgpu` from each of [`hip`](https://github.com/spack/spack-packages/blob/develop/repos/spack_repo/builtin/packages/hip/package.py#L123) and [`hsa-rocr-dev`](https://github.com/spack/spack-packages/blob/develop/repos/spack_repo/builtin/packages/hsa_rocr_dev/package.py#L114), so I don't know if we can get away from it even without `ROCmPackage`.

I will just adjust to not use the `llvm-amdgpu` binaries and leave the wasted dependency.

repos/spack_repo/builtin/packages/chapel/package.py
arifthpe: > I think I agree with your position on the Chapel documentation and error messages, but neither are things we'd potentially change in this repo.

I will PR it on the Chapel repo.

> As far as the Spackage goes, I still think it's worth adding the comment suggested above to record the fact that the omission of ROCm 6.4 was deliberate (possibly with a reference to Chapel PR 28220)

Will do

bonachea (approved):
LGTM!

Thanks @arifthpe, @e-kayrakli and the whole Chapel team for all the work on this!

I've successfully built this Chapel spackage in all of the following configurations:

NERSC Perlmutter (HPE Cray EX, NVIDIA GPUs):
* `chapel@2.8+cuda+pshm hwloc=spack comm=gasnet comm_substrate=ofi %c,cxx=gcc@14.3.0`
* `chapel@2.8+cuda hwloc=spack comm=ofi %c,cxx=gcc@14.3.0`
* `chapel@2.8+cuda+pshm hwloc=spack comm=gasnet comm_substrate=ofi ^llvm@21`

JLSE amdgpu: (Linux cluster with AMD GPUs)
* `chapel@2.8 +rocm amdgpu_target=gfx90a comm=none gmp=none hwloc=spack llvm=bundled ^hip@6.3 ^hsa-rocr-dev@6.3 ^llvm-amdgpu@6.3`
* `chapel@2.8 +rocm amdgpu_target=gfx90a comm=none gmp=none hwloc=spack llvm=spack ^hip@6.3.2 ^hsa-rocr-dev@6.3.2 ^llvm-amdgpu@6.3.2 ^llvm@21`
* `chapel@2.8 +rocm amdgpu_target=gfx90a comm=none gmp=none hwloc=spack llvm=spack ^hip@7 ^hsa-rocr-dev@7 ^llvm-amdgpu@7 ^llvm@21`
* `chapel@2.8 +rocm +pshm amdgpu_target=gfx90a comm=gasnet comm_substrate=ibv gmp=spack hwloc=spack llvm=spack ^hip@6.3.2 ^hsa-rocr-dev@6.3.2 ^llvm-amdgpu@6.3.2 ^llvm@21`

cgpu: (Linux cluster with NVIDIA GPUs)
* `chapel@2.8+pshm comm=gasnet comm_substrate=ibv ^llvm@22`
* `chapel@2.8+pshm+cuda comm=gasnet comm_substrate=ibv %cuda@:12.6+allow-unsupported-compilers ^llvm@22`

arezaii (approved):

arezaii (approved):

bonachea (approved):

tgamblin (approved):

PR #3838, opened 2026-03-13
meshoptimizer: new package

cmelone:
https://meshoptimizer.org

Toggled each variant, each spec installed and passed a functionality test.

I used an LLM to help write this package.


tldahlgren (approved):
Confirmed the homepage and version sha256. `spack spec` also succeeds.

PR #3840, opened 2026-03-13
parallelio: add patch for LibFind.cmake

danrosen25:
This patch fixes discovery for dependencies, such as netcdf-c, installed outside of `<prefix>/lib`. When installing with apt libraries will be installed into `/usr/lib/<architecture>` where `<prefix>=/usr`. This patch also adds a conflict for when newer versions of netcdf-c are used as a dependency for older versions of parallelio.

### DIFF
--- a/repos/spack_repo/builtin/packages/parallelio/package.py
+++ b/repos/spack_repo/builtin/packages/parallelio/package.py
@@ -53,6 +53,7 @@ class Parallelio(CMakePackage):
     # This patch addresses an issue when compiling pio2.6.0 with a serial netcdf library.
     # netcdf4 filters are only available with the parallel build of netcdf.
     patch("pio_260.patch", when="@2.6.0")
+    patch("cmake-libfind-arch.patch", when="@2.6.0:2.6.6")
 
     depends_on("c", type="build")  # generated
     depends_on("cxx", type="build")  # generated
@@ -66,6 +67,12 @@ class Parallelio(CMakePackage):
     depends_on("netcdf-fortran", type="link", when="+fortran")
     depends_on("parallel-netcdf", type="link", when="+pnetcdf")
     depends_on("netcdf-c ~parallel-netcdf", type="link", when="~pnetcdf")
+
+    conflicts(
+        "^netcdf-c@4.9.3:",
+        when="@:2.6.4",
+        msg="netcdf-c@4.9.3+ is incompatible with parallelio versions prior to 2.6.5",
+    )
 
     resource(
         name="genf90",
--- a/repos/spack_repo/builtin/packages/parallelio/package.py
+++ b/repos/spack_repo/builtin/packages/parallelio/package.py
@@ -20,7 +20,10 @@ class Parallelio(CMakePackage):
 
     license("Apache-2.0")
 
+    version("2.6.10", sha256="a65bd3c75f5ee6e8c4c5b922b9c6fa2598af9586468d8dbf51caf010261b4128")
+    version("2.6.9", sha256="22f34258f0e8b9e8271383bc7c6c657f4bbee560d33f87e3c0db53e0da8c8f3b")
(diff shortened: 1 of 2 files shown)


repos/spack_repo/builtin/packages/parallelio/package.py
```diff
@@ -53,6 +53,7 @@ class Parallelio(CMakePackage):
     # This patch addresses an issue when compiling pio2.6.0 with a serial netcdf library.
     # netcdf4 filters are only available with the parallel build of netcdf.
     patch("pio_260.patch", when="@2.6.0")
+    patch("cmake-libfind-arch.patch", when="@2.6.0:2.6.6")
```
alalazo: Can you add a brief comment above the `patch` directive explaining:
2. Why it is needed
3. Where it was taken from

?

alalazo (changes requested):

repos/spack_repo/builtin/packages/parallelio/package.py
danrosen25: @alalazo 
1. This patch adds a suffix to the search paths for libraries when calling find_library, find_shared_library, and find_static_library functions within the CMake build system. The patch also changes the result of `${PKGCOMP}_FOUND` to only set the value to `${PKGCOMPUP}_FOUND`, which is the uppercase modified package name, when needed.
2. The new search path supports the structure used when installing libraries via `apt`, which installs libraries into `usr/lib/${CMAKE_LIBRARY_ARCHITECTURE}`. This allows users to install dependencies external from spack, via apt, then link the libraries into the parallelio build system. The patch also fixes a bug in `${PKGCOMP}_FOUND`, which was always setting the result to `${PKGCOMPUP}_FOUND`, even when the system was not using the uppercase version of the package.
3. The patch is taken from the following revision: https://github.com/NCAR/ParallelIO/commit/6db2078e45ff77ff4f64e75c3a46502aba4a05aa.

danrosen25: @alalazo 
I'm seeing that the maintainer of `paralleio` tagged the `LibFind.cmake` revisions in `v2.6.10+` so I think I should change the application of this patch to `@2.6.0:2.6.9`. Is it okay if I make that change?

alalazo: @danrosen25 Thank for the explanation. Can you add that as a brief comment above the patch directive? It would help the next reader understanding what the patch does without having to open it.

danrosen25:
@alalazo @jedwards4b 
- I've added new versions as well from the tar.gz files here: https://github.com/NCAR/ParallelIO/tags
- I added a comment for the new patch and extended the versions it covers through 2.6.9

bernhardkaindl:
Hi @alalazo, @danrosen25 applied the comment you asked for, can you re-review the PR?

bernhardkaindl:
@danrosen25, automated review raised a question regarding LibFind for 2.6.10 and newer.

danrosen25:
@bernhardkaindl
Yes, parallelio v 2.6.10 includes the LibFind fix, which was merged in with PR 2023 https://github.com/NCAR/ParallelIO/commit/6db2078e45ff77ff4f64e75c3a46502aba4a05aa

As for the comment to fix the comment about usr/lib/ to /usr/lib/ that's totally fine. Oddly enough copilot is suggesting another comment with usr/lib that should be /usr/lib. Should I make these changes, which are sort of cosmetic.

bernhardkaindl (approved):
@danrosen25, agreed, only the cosmetic changes do not warrant the work an cost to update the PR agaim, so I'm approving.

Because @alalazo has set the requested changes review state for his review, he must also approve.

bernhardkaindl:
@alalazo: LGTM, can reset your changes requested state to approve?

alalazo (approved):

PR #3849, opened 2026-03-14
zziplib: new version 0.13.80

cessenat:
https://github.com/gdraheim/zziplib/releases/tag/v0.13.80
Mainly fixes to v0.13.78


rbberger:
@spackbot run pipeline.

rbberger (approved):
LGTM. I only had trouble building `zip` on MacOS.

cessenat:
> LGTM. I only had trouble building `zip` on MacOS.

I too had some problem with `zip`, it seems not to be maintained. I had hoped that `zziplib` would no longer require it in this latest release, false hope.

rbberger:
I found a patch that solves the issues on MacOS, that will be its own PR.

PR #3857, opened 2026-03-16
sepp: restrict py-setuptools version because of pkg-resources again

V-Karch:
This is again needed because of dependency on `pkg-resources` which is removed in `py-setuptools@82:`


tldahlgren:
@snehring 

Including missing package maintainer.

snehring (approved):
lgtm

tldahlgren (approved):
Thanks.

PR #3863, opened 2026-03-16
zip: add patch for newer apple-clang

rbberger:


becker33 (approved):

cessenat:
LGTM but I'm no longer using Apple. Sorry for being so long.

PR #3807, opened 2026-03-12
e4s oneapi ci: upgrade to use oneapi 2025.3

eugeneswalker:
E4S OneAPI CI Stack: Upgrade to use use OneAPI 2025.3

PR #3868, opened 2026-03-17
py-openai: remove from e4s-oneapi test list

yizeyi18:
`py-openai` depends on `py-jiter` in recent versions, which introduces a `py-maturin` dependent. For `e4s-oneapi` ci array fails to build `py-maturin`, this PR removes `py-openai` from this array.

### DIFF
--- a/stacks/e4s-oneapi/spack.yaml
+++ b/stacks/e4s-oneapi/spack.yaml
@@ -204,7 +204,7 @@ spack:
   - opencv +python3
   - py-mpi4py
   - py-numpy
-  - py-openai
+  # - py-openai                     # py-maturin: rust-lld: error: undefined symbol: _intel_fast_memcpy
   - py-pooch
   - py-pytest
   - py-scikit-learn


yizeyi18:
As the failed ci pipeline `windows-vis-generate` is non-related to this PR, some earlier PRs also fail on it and some later not, I call for a re-run pipeline here.

@spackbot re-run pipeline

yizeyi18:
@spackbot help

eugeneswalker:
I see you disabled the spec and added a comment with error annotation. But isn't this building just fine in CI right now? Otherwise develop pipelines would be failing, which they are not... Why disable it?

yizeyi18:
@eugeneswalker Because the version of `py-openai` currently in `develop` is 0.27.8 which does not depend on `py-maturin`. #3746 is blocked when trying to bump it to 2.26.0 .

eugeneswalker (approved):
I see now. Thank you!

yizeyi18:
As `py-openai` has already been deleted in #3807, this PR would be closed. Thanks for Eugenes!

PR #3871, opened 2026-03-17
PythonPackage: speed up import tests

adamjstewart:
### Before

```console
> time spack test run py-keras
==> Spack test 4y2u6jhhhgcv7pxrayesd54fzltdl7dl
==> Testing package py-keras-3.13.2-jrk3c5j
============================== 1 passed of 1 spec ==============================

________________________________________________________
Executed in  344.56 secs    fish           external
   usr time  290.48 secs    1.45 millis  290.48 secs
   sys time   51.96 secs    3.82 millis   51.95 secs
```

### After

```console
> time spack test run py-keras
==> Spack test 4y2u6jhhhgcv7pxrayesd54fzltdl7dl
==> Testing package py-keras-3.13.2-jrk3c5j
============================== 1 passed of 1 spec ==============================

________________________________________________________
Executed in    2.98 secs    fish           external
   usr time    2.58 secs    1.49 millis    2.58 secs
   sys time    0.58 secs    3.20 millis    0.58 secs
```
So a 10x speedup if you only invoke the Python executable a single time. Thanks for reporting this @haampie! Only downside is logging is a little less clear about which import actually failed and you only see a single failure at a time. An alternative would be to somehow parametrize this, but I think it would still be slower than a for-loop.

### DIFF
--- a/repos/spack_repo/builtin/build_systems/python.py
+++ b/repos/spack_repo/builtin/build_systems/python.py
@@ -187,7 +187,7 @@ class PythonExtension(PackageBase):
         # Make sure we are importing the installed modules,
         # not the ones in the source directory
         python = self.module.python
-        with test_part(self, f"test_imports", purpose=f"checking imports", work_dir="spack-test"):
+        with test_part(self, "test_imports", purpose="checking imports", work_dir="spack-test"):
             python(
                 "-c",
                 f"""
--- a/repos/spack_repo/builtin/build_systems/python.py
+++ b/repos/spack_repo/builtin/build_systems/python.py
@@ -190,11 +190,15 @@ class PythonExtension(PackageBase):
         with test_part(self, "test_imports", purpose="checking imports", work_dir="spack-test"):
             python(
                 "-c",
-                f"""
+                """
 import importlib
-for module in {self.import_modules}:
+import sys
+
+for module in sys.argv[1:]:
+    print(module)
     importlib.import_module(module)
 """,
+                *self.import_modules
             )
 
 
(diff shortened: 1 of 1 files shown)


haampie (dismissed):
lgtm, maybe a small improvement would be to read from sys.argv, or split lines from sys.stdin (probably best in case of many modules).

adamjstewart:
@spackbot fix style

repos/spack_repo/builtin/build_systems/python.py
```diff
@@ -187,14 +187,15 @@ def test_imports(self) -> None:
...
         python = self.module.python
-        for module in self.import_modules:
-            with test_part(
-                self,
-                f"test_imports_{module}",
-                purpose=f"checking import of {module}",
-                work_dir="spack-test",
-            ):
-                python("-c", f"import {module}")
+        with test_part(self, f"test_imports", purpose=f"checking imports", work_dir="spack-test"):
```
alalazo:
```suggestion
        with test_part(self, "test_imports", purpose="checking imports", work_dir="spack-test"):
```

haampie:
I would just go for

```python
python("-c", f"""
import sys, importlib
for module in sys.argv[1:]:
    importlib.import_module(module)
""", *self.import_modules)
```

cause I don't think argv length is a concern and this is safer

adamjstewart:
> I would just go for
> 
> ```python
> python("-c", f"""
> import sys, importlib
> for module in sys.argv[1:]:
>     importlib.import_module(module)
> """, *self.import_modules)
> ```
> 
> cause I don't think argv length is a concern and this is safer

Can you elaborate on what makes this safer? I think your concern is basically with shell injection?

haampie:
Yeah, it's unlikely that it would end up running `shutil.rmtree("/")`, but after your change it's ruled out ;p

haampie (approved):

adamjstewart:
Wouldn't be the first code injection vulnerability I introduced into a Python project 😅 
https://www.cvedetails.com/cve/CVE-2024-49048/

PR #3874, opened 2026-03-17
gpi_space: new version 26.3

mrahn:
- Support new version 26.3 of GPI-Space.
- Fix dependency versions for older versions.

### DIFF
--- a/repos/spack_repo/builtin/packages/gpi_space/package.py
+++ b/repos/spack_repo/builtin/packages/gpi_space/package.py
@@ -75,35 +75,14 @@ class GpiSpace(CMakePackage):
     depends_on("chrpath@0.13:", type=("build", "run"))
     depends_on("pkgconfig", type="build")
     depends_on(
-        "boost@1.62.0:1.63.0"
-        "+atomic +chrono +coroutine +context +date_time +iostreams"
-        " +program_options +random +regex +serialization +test +timer cxxstd=17",
-        when="@26.3:",
+        "boost+atomic +chrono +date_time +iostreams +program_options +random +regex +serialization"
+        "+test +timer"
     )
-    depends_on(
-        "boost@1.62.0:1.63.0"
-        "+atomic +chrono +date_time +filesystem +iostreams"
-        " +program_options +random +regex +serialization +test +timer cxxstd=17",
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/gpi_space/package.py
```diff
@@ -62,39 +63,81 @@ class GpiSpace(CMakePackage):
...
     depends_on("pkgconfig", type="build")
     depends_on(
         "boost@1.62.0:1.63.0"
-        "+atomic +chrono +coroutine +context +date_time +filesystem +iostreams"
-        " +program_options +random +regex +serialization +test +timer cxxstd=14",
-        when="@:23.06",
+        "+atomic +chrono +coroutine +context +date_time +iostreams"
+        " +program_options +random +regex +serialization +test +timer cxxstd=17",
+        when="@26.3:",
     )
```
alalazo: Can we say:
```python
depends_on(
    "boost+atomic +chrono +date_time +filesystem +iostreams"
    " +program_options +random +regex +serialization +test +timer"
)
```
once, and later just deal with version / C++ standard constraints? I think that would clean up the recipe quite a bit.

mrahn: Happy to learn how that works. It is not only about the c++ version but also about the filesystem component that is no longer required by 26.3. Would you sketch a solution that doesn't repeat the common components?

alalazo: Sure, you can say:
```python
depends_on(
    "boost+atomic +chrono +date_time +iostreams +program_options +random +regex +serialization +test +timer"
)
```
and that is unconditional. Then you can add below the diff wrt the base case:
```python
depends_on("boost@1.62.0:1.63.0 cxxstd=17", when="@26.3:")
depends_on("boost@1.61.0:1.63.0 +filesystem cxxstd=17", when="@24.12 ^openssl@:1.0)
```
etc.

mrahn: Did the change to factor the common boost components in 75dfdc6c452627c73d41d7080fe3e25d9617e33f

alalazo (approved):

Issue #3817, opened 2026-03-10
LLVM does not inject gcc-runtime


pearzt:
### Steps to reproduce

My system compiler is GCC 11.4.1, but I installed GCC 15.2.0 using Spack, before building LLVM 21.1.8 with this new GCC. Then, I want to compile Score-P with LLVM.

Simple Spack environment:
```yaml
spack:
  specs:
  - 'llvm@21 +flang %gcc@15'
  - scorep %c,cxx,fortran=llvm
  view: true
  concretizer:
    unify: true
```

Concretization and installation works as expected.

<details>

<summary>Output of `spack spec` in the environment</summary>

```
$ spack spec
[+]  llvm@21.1.8+clang~cuda+flang~gold~ipo+libomptarget~libomptarget_debug~link_llvm_dylib+lld+lldb+llvm_dylib+lua~mlir+offload+polly~python~split_dwarf+utils~z3~zstd build_system=cmake build_type=Release compiler-rt=runtime generator=ninja libcxx=runtime libunwind=runtime openmp=runtime shlib_symbol_version=none targets:=aarch64,amdgpu,nvptx,x86 version_suffix=none platform=linux os=rhel9 target=cascadelake %c,cxx=gcc@15.2.0
[+]      ^cmake@3.31.11~doc+ncurses+ownlibs~qtgui build_system=generic build_type=Release platform=linux os=rhel9 target=cascadelake %c,cxx=gcc@15.2.0
[+]          ^curl@8.18.0~gssapi~ldap~libidn2~librtmp~libssh~libssh2+nghttp2 build_system=autotools libs:=shared,static tls:=openssl platform=linux os=rhel9 target=cascadelake %c,cxx=gcc@15.2.0
[+]              ^nghttp2@1.67.1 build_system=autotools platform=linux os=rhel9 target=cascadelake %c,cxx=gcc@15.2.0
[+]      ^compiler-wrapper@1.0 build_system=generic platform=linux os=rhel9 target=cascadelake 
[+]      ^gcc@15.2.0+binutils+bootstrap~graphite+libsanitizer~mold~nvptx~piclibs~profiled~strip build_system=autotools build_type=RelWithDebInfo languages:='c,c++,fortran' patches:=0999dbf platform=linux os=rhel9 target=cascadelake %c,cxx=gcc@11.4.1
[...]

PR #3600, opened 2026-03-02
pcluster: bootstrap OneAPI in the public pipeline

alalazo:

PR #3880, opened 2026-03-18
llvm: add missing binutils and gcc-runtime dependency

rbberger:
Previously +gold pulled in binutils as dependency. That default was recently changed since gold is deprecated. See #3454.

However, building a custom GCC will typically pull in a newer binutils than what is on the system. Using such a GCC for building LLVM then fails if the system binutils is too old. We therefore need to make sure that LLVM also uses the newer binutils.

Also adds an explicit `-B<binutils-prefix>/bin` to the LLVM configuration files to it prefers using those binaries over defaulting to a location such as `/usr/bin`. Just having the binutils prefix in PATH is not good enough.

It further injects a gcc-runtime dependency for packages building with a llvm that is built with gcc.

### DIFF
--- a/repos/spack_repo/builtin/packages/llvm/package.py
+++ b/repos/spack_repo/builtin/packages/llvm/package.py
@@ -354,7 +354,9 @@ class Llvm(CMakePackage, CudaPackage, LlvmDetection, CompilerPackage):
 
     # if gcc was built with newer binutils than the system default, we need the
     # same for our own build
-    depends_on("binutils+gas+ld+plugins~libiberty", type=("build", "link", "run"), when="%gcc+binutils")
+    depends_on(
+        "binutils+gas+ld+plugins~libiberty", type=("build", "link", "run"), when="%gcc+binutils"
+    )
 
     # Older LLVM do not build with newer compilers, and vice versa
     with when("@16:"):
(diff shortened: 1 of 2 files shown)


alalazo:
Seems similar to what I had to do in #3600 for the OneAPI toolchain.

rbberger:
@alalazo also seems related to #3817. are we missing a `llvm-runtime`?

alalazo:
Yeah, we don't have `llvm-runtime` atm

rbberger:
@alalazo @haampie please take a closer look. I've needed to add `-B` to the LLVM build so it makes sure to prefer the newer binutils prefix over defaulting to `/usr/bin/ld`. Apparently, even if `binutils` is a dependency, that isn't good enough since the LLVM lookup logic considers `PATH` too late. I'm using `-B` over `--ld-path`, since the latter would prevent `-fuse-ld=lld`.

pearzt:
I just tested this with a fresh Spack installation and can confirm that it fixes the problem I reported in #3817 \o/

haampie:
Looks good to me, I'm gonna leave this to @alalazo to look at since he's doing very similar work right now.

rbberger:
@spackbot fix style

srekolam (approved):

bernhardkaindl (approved):
@rbberger, Approving as I saw no negative comments and AFAICS, everyone who I expect as reviewer did take a look at it.

I assume it can be merged, and think you can merge if you wish. Otherwise, I guess I might merge sometime later.

PR #3881, opened 2026-03-18
llvm-amdgpu: add missing gcc-install-dir logic

rbberger:
Previous versions of llvm-amdgpu set `GCC_INSTALL_DIR` via CMake, but that option was deprecated some time ago. Instead llvm builds need to write config files as part of post install. Also, during runtime builds the just-built clang needs to know where to find the GCC installation. All this logic was backported from the llvm recipe.


rbberger:
@spackbot run pipeline.

rbberger:
@spackbot run pipeline.

khuck-AMD:
Thanks @rbberger ! I am testing this in RHEL8 with gcc 14 right now...

khuck-AMD:
> Thanks @rbberger ! I am testing this in RHEL8 with gcc 14 right now...

It worked great in RHEL8 with GCC 14!

alalazo (approved):

PR #3885, opened 2026-03-18
nccl, aws-ofi-nccl: add new versions and add rocm support to aws-ofi-nccl

msimberg:
Since 1.18.0 aws-ofi-nccl has support for rocm (https://github.com/aws/aws-ofi-nccl/pull/1014). This effectively replaces aws-ofi-rccl. Note that `--with-rccl` is not needed. `--with-nccl` could possibly also be removed, but I have not tested that change so leaving it out from here.


repos/spack_repo/builtin/packages/aws_ofi_nccl/package.py
```diff
@@ -47,15 +49,22 @@ class AwsOfiNccl(AutotoolsPackage):
     version("1.7.1", sha256="d50a160c7aba76445e5c895fba0f3dbfdec51f702d218168a5e5017806cf0fb0")
     version("1.6.0", sha256="19a6fc91afe9a317fd3154c897fa219eab48fcdddefa66d881f1843c1165f7ee")
 
+    variant("cuda", default=True, description="Enable CUDA support")
```
albestro, line 52: is there any reason for not making this a `CudaPackage`/`ROCmPackage`?

msimberg: No, I'm glad you asked. It just didn't cross my mind... Let me try to make that change.

repos/spack_repo/builtin/packages/aws_ofi_nccl/package.py
```diff
@@ -47,15 +49,22 @@ class AwsOfiNccl(AutotoolsPackage):
...
     depends_on("c", type="build")
     depends_on("cxx", type="build", when="@1.15:")
 
     depends_on("libfabric")
-    depends_on("cuda")
-    depends_on("nccl fabrics=auto")
+    depends_on("cuda", when="+cuda")
+    depends_on("nccl fabrics=auto", when="+cuda")
+    depends_on("hip", when="+rocm")
+    depends_on("rccl", when="+rocm")
```
msimberg: I should actually check if this is needed since `--with-rccl` isn't needed...

repos/spack_repo/builtin/packages/aws_ofi_nccl/package.py
msimberg: It works, of course. The downside is that the package gets the cuda_arch/amdgpu_target variants as well, which are not necessary since the plugin doesn't compile any device code. That might've been the original intent behind not using CudaPackage/ROCmPackage. I think I prefer the explicit cuda/rocm variants to avoid the extra unnecessary variants, but don't feel too strongly. Do you have an opinion? If not, I'd revert the latest change.

repos/spack_repo/builtin/packages/aws_ofi_nccl/package.py
msimberg: Removed.

albestro (approved):
LGTM let's give some more time to maintainers of the package to review/comment.

PR #3843, opened 2026-03-14
windows ci: evaluate capability of uo-frank runners, large and small

eugeneswalker:
!!! *DO NOT MERGE THIS* !!!

This PR is just for evaluating capacity of the new Windows runners, large and small, at University of Oregon Frank.

!!! *DO NOT MERGE THIS* !!!

PR #3887, opened 2026-03-18
Windows Gitlab CI: Explicitly override image entrypoint

johnwparent:
By default, all docker image entrypoints are overridden for CI runs on Gitlab as configured by the `default:image:override` setting in the base gitlab-ci yaml. When overriding part of an image definition in a job definition, gitlab will throw away the non overriden component as well. Windows specifies a specific docker image, and in doing so, removes the entrypoint override. This is fine for our cloud hosted runners as kube overrides the entrypoint itself. Docker+machine runners however will respect the non overridden entrypoint.

Explicitly override the entrypoint in our Windows CI config to bring Windows stacks in line with the rest of Spack's stacks.

Adds an explicit entrypoint override to the CI config image override for Windows generate and build jobs.

Fix spun out from https://github.com/spack/spack-packages/pull/3843

### DIFF
--- a/.ci/gitlab/.gitlab-ci.yml
+++ b/.ci/gitlab/.gitlab-ci.yml
@@ -216,7 +216,10 @@ default:
     - $ErrorActionPreference=$ErrorActionOld
 
   tags: ["spack", "public", "medium", "x86_64-win"]
-  image: "ghcr.io/johnwparent/windows-server21h2:sha-1c12b61"
+  image: {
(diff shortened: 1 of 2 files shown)


johnwparent:
@zackgalbreath @eugeneswalker I spun out the fix from the UO test branch and added the fix to the build jobs as well.

eugeneswalker (approved):
Thank you!

zackgalbreath (approved):

kwryankrattiger (approved):

PR #3904, opened 2026-03-19
Add versions and fix url zip/tar.gz function for py-pathos and py-pox

KiSchnelle:


wdconinc:
Is there a particular reason to add the older versions? They are unlikely to be picked by the solver but add maintenance. It also makes the diff much larger than necessary and that's likely why the review has not happened yet (combined with restructuring the organization of the file, which also just increases the diff for reviewers to try to understand). If there are downstream packages that need a specific intermediate version and can't work with newer versions, then that's a good reason of course.

bernhardkaindl (changes requested):
@KiSchnelle, see https://github.com/spack/spack-packages/pull/3904#issuecomment-4232959781

PR #3913, opened 2026-03-19
Update ExaGO package with ORNL Github.

nkoukpaizan:
This updates the ExaGO GitHub repository from PNNL to ORNL and cleans up the available versions.

### DIFF
--- a/repos/spack_repo/builtin/packages/exago/package.py
+++ b/repos/spack_repo/builtin/packages/exago/package.py
@@ -29,51 +29,15 @@ class Exago(CMakePackage, CudaPackage, ROCmPackage):
     git = "https://github.com/ornl/ExaGO.git"
     maintainers("ryandanehy", "cameronrutherford", "pelesh")
 
-    version(
-        "1.6.0",
-        commit="159cd173572280ac0f6f094a71dcc3ebeeb34076",
-        submodules=submodules,
-    )
-    version(
-        "1.5.1",
-        commit="84e9faf9d9dad8d851075eba26038338d90e6d3a",
-        submodules=submodules,
-    )
-    version(
-        "1.5.0",
-        commit="227f49573a28bdd234be5500b3733be78a958f15",
-        submodules=submodules,
-    )
-    version(
-        "1.4.1",
-        commit="ea607c685444b5f345bfdc9a59c345f0f30adde2",
-        submodules=submodules,
-    )
-    version(
-        "1.4.0",
-        commit="4f4c3fdb40b52ace2d6ba000e7f24b340ec8e886",
-        submodules=submodules,
-    )
-    version(
-        "1.3.0",
-        commit="58b039d746a6eac8e84b0afc01354cd58caec485",
-        submodules=submodules,
-    )
-    version(
-        "1.1.2",
-        commit="db3bb16e19c09e01402071623258dae4d13e5133",
-        submodules=submodules,
-    )
-    version(
-        "1.1.1",
-        commit="0e0a3f27604876749d47c06ec71daaca4b270df9",
-        submodules=submodules,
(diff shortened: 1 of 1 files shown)


nkoukpaizan:
@spackbot fix style

pelesh (approved):
The changes are appropriate.

repos/spack_repo/builtin/packages/exago/package.py
```diff
@@ -25,93 +25,37 @@ class Exago(CMakePackage, CudaPackage, ROCmPackage):
...
-    )
+    version("kpp2", commit="1da764d80a2db793f4c43ca50e50981f7ed3880a", submodules=submodules)
 
-    # Progrmming model options
+    # Programming model options
     variant("mpi", default=True, description="Enable/Disable MPI")
     variant("raja", default=False, description="Enable/Disable RAJA")
     variant("python", default=True, when="@1.4:", description="Enable/Disable Python bindings")
     variant("logging", default=True, description="Enable/Disable spdlog based logging")
+    variant("testing", default=True, description="Enable/Disable testing")
```
cameronrutherford, line 58: I am vaguely remembering that it isn't preferred to have a variant for tests like this, but am fuzzy on the details. From the [palace `package.py`](https://github.com/spack/spack-packages/blob/develop/repos/spack_repo/builtin/packages/palace/package.py), I think we use `self.run_tests` since that's more spack generic...

repos/spack_repo/builtin/packages/exago/package.py
```diff
@@ -25,93 +25,37 @@ class Exago(CMakePackage, CudaPackage, ROCmPackage):
...
+    version("1.5.0", commit="227f49573a28bdd234be5500b3733be78a958f15", submodules=submodules)
+    version("1.4.1", commit="ea607c685444b5f345bfdc9a59c345f0f30adde2", submodules=submodules)
+    version("1.4.0", commit="4f4c3fdb40b52ace2d6ba000e7f24b340ec8e886", submodules=submodules)
+    version("1.3.0", commit="58b039d746a6eac8e84b0afc01354cd58caec485", submodules=submodules)
+    version("1.1.2", commit="db3bb16e19c09e01402071623258dae4d13e5133", submodules=submodules)
+    version("1.1.1", commit="0e0a3f27604876749d47c06ec71daaca4b270df9", submodules=submodules)
+    version("1.1.0", commit="dc8dd85544ff1b55a64a3cbbbdf12b8a0c6fdaf6", submodules=submodules)
+    version("1.0.0", commit="230d7df2f384f68b952a1ea03aad41431eaad283")
+    version("0.99.2", commit="56961641f50827b3aa4c14524f2f978dc48b9ce5")
+    version("0.99.1", commit="0ae426c76651ba5a9dbcaeb95f18d1b8ba961690")
```
cameronrutherford, line 43: I guess no more tags is a change that doesn't impact functionality, but I think the tags are nice...

cameronrutherford (approved):
Just NIT comments that don't block merging. LGTM! :shipit:

repos/spack_repo/builtin/packages/exago/package.py
nkoukpaizan: The reason for removing the tags is that we did not move these tags over to ORNL. I checked that the commits are available, however.

repos/spack_repo/builtin/packages/exago/package.py
nkoukpaizan: The idea is to make the running of tests independent from building them. We still need to clarify the CMake options on ExaGO's side, but this will do for what we're planning for CI testing.

wdconinc (approved):
Thanks for this update.

PR #3918, opened 2026-03-20
ants: add master

manuelakuhn:
Add master branch

Although there is a newer version of ants available (2.6.5) this does not build with gcc@15 because of [this](https://github.com/ANTsX/ANTsPy/issues/849) issue.


RikkiButler20 (approved):
LGTM

PR #3920, opened 2026-03-20
octopus: fix typos when using `~mpi`

alalazo:
Variant propagation is ignored in `package.py` because it's unsupported.


RemiLacroix-IDRIS (approved):
LGTM

haampie (approved):

PR #3925, opened 2026-03-20
megahit: add version 1.2.9

V-Karch:
Swapped over to cmake build system because that's what the package does upstream, removed older versions and added a newer one. I'm not sure how to do a mixed package that could be both a CMakePackage and a MakefilePackage otherwise I would try to make both and merge them together


repos/spack_repo/builtin/packages/megahit/package.py
```diff
@@ -1,31 +1,26 @@
-# Copyright Spack Project Developers. See COPYRIGHT file for details.
+# Copyright 2013-2024 Lawrence Livermore National Security, LLC and other
+# Spack Project Developers. See the top-level COPYRIGHT file for details.
```
alalazo: Can you revert these two lines?

repos/spack_repo/builtin/packages/megahit/package.py
```diff
@@ -1,31 +1,26 @@
...
-    patch("amd.patch", when="target=aarch64:")
+    version("1.2.9", sha256="09026eb07cc4e2d24f58b0a13f7a826ae8bb73da735a47cb1cbe6e4693118852")
 
-    def install(self, spec, prefix):
-        mkdirp(prefix.bin)
-        install("megahit", prefix.bin)
-        install("megahit_asm_core", prefix.bin)
-        install("megahit_sdbg_build", prefix.bin)
-        install("megahit_toolkit", prefix.bin)
+    depends_on("zlib", type="build")
```
alalazo:
```suggestion
    depends_on("zlib-api")
```

repos/spack_repo/builtin/packages/megahit/package.py
```diff
@@ -1,31 +1,26 @@
...
+    depends_on("cmake@2.8:", type="build")
+    depends_on("gcc@4.8.4:", type="build")
```
alalazo: You must specify the dependency on the language

alalazo (changes requested):

V-Karch:
@alalazo I believe I've done what your review has asked for

wdconinc:
> I'm not sure how to do a mixed package that could be both a CMakePackage and a MakefilePackage

See https://spack.readthedocs.io/en/latest/packaging_guide_advanced.html#multiple-build-systems

The concern about removing the older versions is explained in https://spack.readthedocs.io/en/latest/packaging_guide_creation.html#deprecate
> Version deprecations should always last at least one release cycle of the builtin package repository before the version is completely removed. No version should be removed without such a deprecation process.

I'll leave it up to @alalazo to decide here but I am sure the review will raise fewer concerns if it keeps the Makefile build for those who are using the older versions, deprecates those versions, and adds the new CMake build for the new version :-)

V-Karch:
@alalazo What do you think?

alalazo:
@V-Karch Apologies for the delay. I agree with @wdconinc that we should deprecate versions first. Here you can find the guide on how to add multiple build systems: https://spack.readthedocs.io/en/latest/packaging_guide_advanced.html#multiple-build-systems

V-Karch:
@alalazo is this fine?

V-Karch:
@alalazo @wdconinc Thank you both, I'm not sure that this is the right method but it wasn't working with any other way I could get to work, this seems to be functional. Please let me know your thoughts whenever you are available

repos/spack_repo/builtin/packages/megahit/package.py
```diff
@@ -2,28 +2,55 @@
...
-from spack_repo.builtin.build_systems.makefile import MakefilePackage
+from spack_repo.builtin.build_systems import makefile
 
 from spack.package import *
 
 
-class Megahit(MakefilePackage):
-    """MEGAHIT: An ultra-fast single-node solution for
-    large and complex metagenomics assembly via succinct de Bruijn graph"""
+class Megahit(Package):
```
wdconinc:
```suggestion
class Megahit(CMakePackage, MakefilePackage):
```

repos/spack_repo/builtin/packages/megahit/package.py
```diff
@@ -2,28 +2,55 @@
 #
 # SPDX-License-Identifier: (Apache-2.0 OR MIT)
 
-from spack_repo.builtin.build_systems.makefile import MakefilePackage
+from spack_repo.builtin.build_systems import makefile
```
wdconinc:
```suggestion
from spack_repo.builtin.build_systems.cmake import CMakePackage
from spack_repo.builtin.build_systems.makefile import MakefilePackage
```

repos/spack_repo/builtin/packages/megahit/package.py
```diff
@@ -2,28 +2,55 @@
...
     version("1.1.3", sha256="b6eefdee075aaf7a8f9090e2e8b08b770caff90aa43a255e0e220d82ce71c492")
 
+    variant(
+        "generator",
+        default="make",
+        values=("make", "ninja"),
+        when="build_system=cmake",
+        description="CMake generator",
+    )
+
```
wdconinc: Unneeded, by inheritance.
```suggestion
```

repos/spack_repo/builtin/packages/megahit/package.py
```diff
@@ -2,28 +2,55 @@
...
+class Megahit(Package):
+    """Ultra-fast and memory-efficient (meta-)genome assembler"""
 
     homepage = "https://github.com/voutcn/megahit"
-    url = "https://github.com/voutcn/megahit/archive/v1.1.3.tar.gz"
+    url = "https://github.com/voutcn/megahit/archive/refs/tags/v1.2.9.tar.gz"
 
     license("GPL-3.0-only")
 
+    build_system("cmake", "makefile", default="cmake")
```
wdconinc:
```suggestion
    build_system(conditional("cmake", when="@1.2:"), conditional("makefile", when="@:1.1", default="cmake")
```
Will need style reflow.

repos/spack_repo/builtin/packages/megahit/package.py
```diff
@@ -2,28 +2,55 @@
...
     depends_on("zlib-api")
 
-    patch("amd.patch", when="target=aarch64:")
+    # CMake path
+    depends_on("cmake@2.8:", type="build", when="@1.2.9: build_system=cmake")
+    depends_on("gcc@4.8.4:", type="build", when="@1.2.9: build_system=cmake")
+    depends_on("c", type="build", when="@1.2.9: build_system=cmake")
+    depends_on("cxx", type="build", when="@1.2.9: build_system=cmake")
+    depends_on("gmake", type="build", when="generator=make")
+    depends_on("ninja", type="build", when="generator=ninja")
```
wdconinc: Inherited.
```suggestion
```

repos/spack_repo/builtin/packages/megahit/package.py
```diff
@@ -2,28 +2,55 @@
...
+        when="build_system=cmake",
+        description="CMake generator",
+    )
+
     depends_on("zlib-api")
 
-    patch("amd.patch", when="target=aarch64:")
+    # CMake path
+    depends_on("cmake@2.8:", type="build", when="@1.2.9: build_system=cmake")
+    depends_on("gcc@4.8.4:", type="build", when="@1.2.9: build_system=cmake")
```
wdconinc, line 31: Is this an explicit gcc dependency? Or just a way to satisfy the C/C++ language dependency below?

repos/spack_repo/builtin/packages/megahit/package.py
```diff
@@ -2,28 +2,55 @@
...
+    depends_on("c", type="build", when="@1.2.9: build_system=cmake")
+    depends_on("cxx", type="build", when="@1.2.9: build_system=cmake")
```
wdconinc: Since also needed for `makefile` build:
```suggestion
    depends_on("c", type="build")
    depends_on("cxx", type="build")
```
and likely more to top of dependencies.

wdconinc (commented):
Thanks! Some comments.

repos/spack_repo/builtin/packages/megahit/package.py
V-Karch: @wdconinc that's from the official repository, it says it wants `gcc@4.8.4:` listed [here](https://github.com/voutcn/MEGAHIT) Under the "Prerequisites" Section

V-Karch:
@wdconinc your suggestion for the `conditional()` things is giving me a syntax error, is it perhaps a newer addition?

`==> Error: cannot load package 'megahit' from the 'builtin' repository: conditional() got an unexpected keyword argument 'default'`

repos/spack_repo/builtin/packages/megahit/package.py
wdconinc: Sorry, the `default` should be outside the `conditional`.

V-Karch:
@wdconinc should be all good now!

repos/spack_repo/builtin/packages/megahit/package.py
```diff
@@ -2,28 +2,48 @@
...
+    depends_on("c", type="build")
+    depends_on("cxx", type="build")
+
+    depends_on("gzip", type="run", when="@1.2.9: build_system=cmake")
+    depends_on("bzip2", type="run", when="@1.2.9: build_system=cmake")
+
+    # Makefile path
+    depends_on("c", type="build", when="build_system=makefile")
+    depends_on("cxx", type="build", when="build_system=makefile")
+
```
wdconinc: This is duplicative with L32-L33 above.

wdconinc (commented):
One more comment.

repos/spack_repo/builtin/packages/megahit/package.py
V-Karch: @wdconinc should be removed now

V-Karch:
Also ran a rebase

wdconinc (approved):
LGTM. Needs approval from @alalazo to clear "requested changes".

V-Karch:
@alalazo This has been approved by @wdconinc, what do you think?

alalazo (approved):

PR #3926, opened 2026-03-20
paraview: add version 6.0.1

cquammen:
Simple addition of ParaView 6.0.1 patch release.

### DIFF
--- a/repos/spack_repo/builtin/packages/paraview/package.py
+++ b/repos/spack_repo/builtin/packages/paraview/package.py
@@ -64,11 +64,8 @@ class Paraview(CMakePackage, CudaPackage, ROCmPackage):
     license("Apache-2.0")
 
     version("master", branch="master", submodules=True)
-    version(
-        "6.0.0",
-        sha256="0ee07ae6377e5e97766aebf858eb9758668a52df041f319e7c975037a63bf189",
-        preferred=True,
-    )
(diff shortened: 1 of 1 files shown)


vicentebolea (approved):

repos/spack_repo/builtin/packages/paraview/package.py
```diff
@@ -65,10 +65,11 @@ class Paraview(CMakePackage, CudaPackage, ROCmPackage):
 
     version("master", branch="master", submodules=True)
     version(
-        "6.0.0",
-        sha256="0ee07ae6377e5e97766aebf858eb9758668a52df041f319e7c975037a63bf189",
+        "6.0.1",
+        sha256="5e56ac7af5e925b3cfd3fab82470933cbabc7e8fda87e14af64f995d6064eb06",
         preferred=True,
     )
```
alecbcs: Since we are now preferring the latest version of paraview, we can drop the `preferred=True` parameter on the version directive since this is Spack's default behavior. This'll simplify future additions by allowing folks to just add in the new directive via `spack checksum -a`

cquammen: Roger that. Updated.

alecbcs (approved):
Looks good to me. Thanks!

PR #3927, opened 2026-03-20
py-black: deprecate versions due to CVE

adamjstewart:
@tgamblin another reason to move to ruff. Note that we will no longer be able to use EOL Python versions to install black if we remove these deprecated versions.

May require running black on all files to ensure they are up-to-date.


adamjstewart:
@spackbot fix style

tgamblin:
Thanks @adamjstewart. I'm hoping we can move to ruff this week.  @johnwparent and @kwryankrattiger how is that going?

Sinan81:
I've checked out modifications to qt, qscintilla, & bazel packages. looks good to me.

mwkrentel (approved):

adamjstewart:
Removed formatting changes, let's merge this and then switch to ruff in a separate PR.

alecbcs (dismissed):
Modifications to `py-black`, `py-pathspec`, and the addition of `py-pytokens` look good to me. Does anyone here want to hold off on merging these changes now that this isn't otherwise modifying the formatter?

tgamblin (approved):

PR #3931, opened 2026-03-20
py-pylibjpeg-openjpeg: add new package

manuelakuhn:
https://github.com/pydicom/pylibjpeg-openjpeg/releases/tag/v2.5.0

### DIFF
--- a/repos/spack_repo/builtin/packages/openjpeg/package.py
+++ b/repos/spack_repo/builtin/packages/openjpeg/package.py
@@ -22,6 +22,7 @@ class Openjpeg(CMakePackage):
 
     license("BSD-2-Clause-NetBSD")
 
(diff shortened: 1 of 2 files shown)


alecbcs:
Should this package also depend on `openjpeg` itself?

manuelakuhn:
Yes you're right, I completely overlooked that because I was too focused on the python stuff. Thunk you very much. I have added `openjpeg` as dependency now.
I was not sure if `openjpeg@2.5.3:` was a hard requirement for `py-pylibjpeg-openjpeg` but from the release noted it looked like it for me, so I also updated `openjpeg`. 

https://github.com/uclouvain/openjpeg/releases/tag/v2.5.4

alecbcs (approved):
No worries :) I figured it was a quick fix, but just wanted to check that I wasn't missing something or had misunderstood the package's relationship to openjpeg.

manuelakuhn:
Thank you very much!

PR #3932, opened 2026-03-21
ior: fixes for IOR 4.0.0 and add +aio suport

yuchenzzhang-G:
Support IOR and Lustre for IOR 4.0.0 version.

1. IOR 4.0.0 fixes the lustre issue and does not the patch.
2. Add AIO support for IOR, which is added on 4.0.0 
3. compiling of IOR 4.0.0 also requires pkgconf

### DIFF
--- a/repos/spack_repo/builtin/packages/ior/package.py
+++ b/repos/spack_repo/builtin/packages/ior/package.py
@@ -29,8 +29,7 @@ class Ior(AutotoolsPackage):
     variant("hdf5", default=False, description="support IO with HDF5 backend")
     variant("ncmpi", default=False, description="support IO with NCMPI backend")
     variant("lustre", default=False, description="support configurable Lustre striping values")
-    variant("aio", default=False, description="support AIO backend API")
-    conflicts("+aio", when="@:3.9.9", msg="AIO support requires IOR version 4.0.0 or newer")
+    variant("aio", default=False, description="support AIO backend API", when="@4:")
 
     depends_on("c", type="build")  # generated
 
@@ -58,7 +57,7 @@ class Ior(AutotoolsPackage):
     patch(
         "https://github.com/glennklockwood/ior/commit/e49476be64d4100c2da662ea415f327348b3d11d.patch?full_index=1",
         sha256="ee3527023ef70ea9aee2e6208f8be7126d5a48f26c587deed3d6238b4f848a06",
-        when="+lustre @:3.9.9",
+        when="+lustre @:3",
     )
 
     @run_before("autoreconf")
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/ior/package.py
```diff
@@ -29,17 +29,21 @@ class Ior(AutotoolsPackage):
     variant("hdf5", default=False, description="support IO with HDF5 backend")
     variant("ncmpi", default=False, description="support IO with NCMPI backend")
     variant("lustre", default=False, description="support configurable Lustre striping values")
+    variant("aio", default=False, description="support AIO backend API")
```
wdconinc:
```suggestion
    variant("aio", default=False, description="support AIO backend API", when="@4:")
```

repos/spack_repo/builtin/packages/ior/package.py
```diff
@@ -29,17 +29,21 @@ class Ior(AutotoolsPackage):
...
+    conflicts("+aio", when="@:3.9.9", msg="AIO support requires IOR version 4.0.0 or newer")
 
     depends_on("c", type="build")  # generated
 
     depends_on("autoconf", type="build")
     depends_on("automake", type="build")
     depends_on("libtool", type="build")
     depends_on("m4", type="build")
+    depends_on("pkgconf", type="build", when="@4.0.0:")
```
wdconinc, line 40: If this is logically connected to aio detection, then maybe:
```suggestion
    depends_on("pkgconf", type="build", when="+aio")
```

repos/spack_repo/builtin/packages/ior/package.py
```diff
@@ -54,7 +58,7 @@ class Ior(AutotoolsPackage):
     patch(
         "https://github.com/glennklockwood/ior/commit/e49476be64d4100c2da662ea415f327348b3d11d.patch?full_index=1",
         sha256="ee3527023ef70ea9aee2e6208f8be7126d5a48f26c587deed3d6238b4f848a06",
-        when="+lustre",
+        when="+lustre @:3.9.9",
```
wdconinc:
```suggestion
        when="+lustre @:3",
```

wdconinc (commented):
Thanks for the contribution. A few comments.

repos/spack_repo/builtin/packages/ior/package.py
yuchenzzhang-G: Thanks for the comments. I think it is actually bind to the IOR version, for IOR 4.0.0 pkgconf is required.

wdconinc (approved):

PR #3939, opened 2026-03-21
opam: new version 2.5.0 and dependency to cpp

cessenat:
New version 2.5.0 release notes:
https://opam.ocaml.org/blog/opam-2-5-0/
This compiles on Ubuntu 20 with gcc 9.4.0, at least.
Cpp dependency required for both previous 2.2.1 and latest 2.5.0.

### DIFF
--- a/repos/spack_repo/builtin/packages/opam/package.py
+++ b/repos/spack_repo/builtin/packages/opam/package.py
@@ -37,7 +37,7 @@ class Opam(AutotoolsPackage):
     version("1.2.1", sha256="f210ece7a2def34b486c9ccfb75de8febd64487b2ea4a14a7fa0358f37eacc3b")
 
     depends_on("c", type="build")
-    depends_on("cxx", type="build", when="@2.2.1:")
+    depends_on("cxx", type="build", when="@2.2.0:")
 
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/opam/package.py
```diff
@@ -35,7 +36,8 @@ class Opam(AutotoolsPackage):
     version("1.2.2", sha256="15e617179251041f4bf3910257bbb8398db987d863dd3cfc288bdd958de58f00")
     version("1.2.1", sha256="f210ece7a2def34b486c9ccfb75de8febd64487b2ea4a14a7fa0358f37eacc3b")
 
-    depends_on("c", type="build")  # generated
+    depends_on("c", type="build")
+    depends_on("cxx", type="build")
```
alecbcs:
```suggestion
    depends_on("cxx", type="build", when="@2.2.1:")
```

@cessenat based on the requirements you state in your PR description, shouldn't we just apply the `c++` dependency on versions newer than or equal to v2.2.1?

scemama (approved):

alecbcs (approved):
Looks good to me. Thanks!

PR #3940, opened 2026-03-22
fortuno: new package

SeanBryan51:
Adds a spack package recipe for the Fortuno unit test framework: https://github.com/fortuno-repos/fortuno

I was able to confirm all variants were installed correctly on my darwin m1 laptop.

Please let me know if you need any further details. Thanks 🙂

### DIFF
--- a/repos/spack_repo/builtin/packages/fortuno/package.py
+++ b/repos/spack_repo/builtin/packages/fortuno/package.py
@@ -1,0 +1,47 @@
+# Copyright Spack Project Developers. See COPYRIGHT file for details.
+#
+# SPDX-License-Identifier: (Apache-2.0 OR MIT)
+
+from spack_repo.builtin.build_systems.cmake import CMakePackage
+
+from spack.package import *
+
+
+class Fortuno(CMakePackage):
+    """Flexible & extensible object oriented Fortran unit testing framework for
+    serial, MPI-parallel and coarray-parallel projects"""
+
+    homepage = "https://fortuno.readthedocs.io/en/latest/"
+    url = "https://github.com/fortuno-repos/fortuno/archive/refs/tags/v0.1.0.tar.gz"
+    git = "https://github.com/fortuno-repos/fortuno.git"
+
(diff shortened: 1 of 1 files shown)


SeanBryan51:
Hi @aradi, I thought I'd loop you in on this. I'd be more than happy to have your feedback on any elements regarding the Fortuno build process. Would you mind if I add you to the list of [maintainers](https://spack.readthedocs.io/en/latest/roles_and_responsibilities.html#maintainers-package-owners) for this package?

repos/spack_repo/builtin/packages/fortuno/package.py
```diff
@@ -0,0 +1,47 @@
...
+
+    version("0.1.0", sha256="9639be1159e03d5e31032af6aa9d4f8483d1f3df0b84a3b43a564273a68ff5e1")
+
+    variant("serial", default=True, description="Build serial interface")
+    variant("mpi", default=False, description="Build MPI interface")
+    variant("coarray", default=False, description="Build coarray interface")
+    variant("shared", default=True, description="Build as shared library")
+
+    depends_on("fortran", type="build")
+    depends_on("cmake@3.22:3.31", type="build")
```
wdconinc: The CMake policy maximum in the version range means something different than this strict compatibility range. We typically leave it open on the upper end unless you know 3.32 doesn't work from some reason.

```suggestion
    depends_on("cmake@3.22:", type="build")
```

wdconinc (commented):
LGTM. One comment.

repos/spack_repo/builtin/packages/fortuno/package.py
SeanBryan51: Thanks, I wasn't able to find anything that would hint at things breaking at 3.32 - so it should be okay to remove the upper bound.

SeanBryan51:
@wdconinc Thank you for reviewing this, I'm happy for this to go in when you are.

wdconinc (approved):

PR #3942, opened 2026-03-22
prrte: external hwloc with multiarch pkgconfig compatibility

cessenat:
When hwloc is set as external on a multiarch configuration, such as on Ubuntu, include is on standard path `/usr/include/hwloc.h` while lib is on `/usr/lib/x86_64-linux-gnu/libhwloc.so`.
The prrte configure script first looks for a pkgconfig setup on "`${check_package_libdir}/pkgconfig/hwloc.pc`".

So, both setting in packages.yaml:
```
  hwloc:
    externals:
    - spec: hwloc@2.10.0
      prefix: /usr/lib/x86_64-linux-gnu
```
and this PR, makes the configure script work.


cessenat:
Another fix for this issue would be to copy what is done in the pmix recipe.

cessenat:
> Another fix for this issue would be to copy what is done in the pmix recipe.

Finally the `pmix` logic is much better as it does not alter the "prefix" in the external packages.yaml file.
It also needs to be consistant, as explained in https://github.com/openpmix/prrte/issues/2419 by @rhc54.

cessenat:
ping @alalazo ? This PR does the same trick as `pmix`.

repos/spack_repo/builtin/packages/prrte/package.py
```diff
@@ -95,6 +96,18 @@ def autoreconf(self, spec, prefix):
...
+        spec = self.spec
+        tgt_libpath = ""
+        dir_list = spec[pkg_name].libs
+        for entry in dir_list:
+            if path_match_str == "" or (path_match_str != "" and path_match_str in entry):
+                tgt_libpath = entry
+                break
+        path_list = tgt_libpath.split(os.sep)
+        del path_list[-1]
+        return (os.sep).join(path_list)
```
alalazo, line 130: Can we avoid duplication with `pmix`? This seems a verbatim copy of the method used in that package.

alalazo (changes requested):

repos/spack_repo/builtin/packages/prrte/package.py
cessenat: Yes @alalazo it is an exact verbatim copy. I do not have any idea how to do it in another way.

bernhardkaindl: @cessenat, @alalazo, I also don't know what to do about the duplication, maybe add the function to `repos/spack_repo/builtin/build_systems/autotools.py`?

PR #3943, opened 2026-03-22
py-uv: deprecate versions due to CVE

adamjstewart:
Only medium severity, but I don't know of any reason anyone would want to install an older version of uv other than to use an older rust version.


adamjstewart:
@spackbot fix style

rbberger (approved):

PR #3950, opened 2026-03-23
libfabric: add 2.5.0 and 2.5.1

msimberg:
- https://github.com/ofiwg/libfabric/releases/tag/v2.5.0
- https://github.com/ofiwg/libfabric/blob/v2.5.0/NEWS.md#v250-fri-march-20-2026
- https://github.com/ofiwg/libfabric/releases/tag/v2.5.1
- https://github.com/ofiwg/libfabric/blob/v2.5.1/NEWS.md#v251-mon-april-13-2026

Adds a version constraint with 2.5.0 on libcxi when using `fabrics=cxi` because of https://github.com/ofiwg/libfabric/issues/12036.


albestro (dismissed):

msimberg:
Ping @darrylabbate (I think you're not in maintainers so I can't explicitly request a review from you).

msimberg:
@ax3l, @vicentebolea, @williamfgc, @soumagne the adios2 and mercury builds failed with libfabric-related linker errors which I assume really comes from this new version:

- adios2: https://gitlab.spack.io/spack/spack-packages/-/jobs/21128168
- mercury: https://gitlab.spack.io/spack/spack-packages/-/jobs/21128170

In the latest commit I tried to add generic conflicts with the latest libfabric version (still waiting for CI to see if that's sufficient (edit: the conflicts seem to have helped in the sense that nothing is rebuilt now)) but perhaps you want to have a look if more specific constraints are required?

soumagne:
hmm that's odd. I think what this indicates is that somehow an older libfabric version is picked up at link time, ie. it is correctly compiling against libfabric 2.5.0 but then maybe another library is pulling an older libfabric ? this is usually what would cause the errors of the type `undefined reference to fi_freeinfo@FABRIC_1.9'`  where `FABRIC_1.9` symbols are only defined in the newer version of libfabric.

msimberg:
I've added a conflict for this issue with the cxi provider: https://github.com/ofiwg/libfabric/issues/12036.

> hmm that's odd. I think what this indicates is that somehow an older libfabric version is picked up at link time, ie. it is correctly compiling against libfabric 2.5.0 but then maybe another library is pulling an older libfabric ? this is usually what would cause the errors of the type `undefined reference to fi_freeinfo@FABRIC_1.9'` where `FABRIC_1.9` symbols are only defined in the newer version of libfabric.

Yeah, it is strange indeed. I unfortunately don't have any insight into the runners if there might actually be another libfabric installed somewhere on that runner/in that environment. Would you be ok with keeping the conflict for now or would that block you in reasonable use cases? @alalazo or @haampie would you have more information on the runner tha failed (or know who else to ask)?

darrylabbate (approved):

soumagne:
> I've added a conflict for this issue with the cxi provider: [ofiwg/libfabric#12036](https://github.com/ofiwg/libfabric/issues/12036).
> 
> > hmm that's odd. I think what this indicates is that somehow an older libfabric version is picked up at link time, ie. it is correctly compiling against libfabric 2.5.0 but then maybe another library is pulling an older libfabric ? this is usually what would cause the errors of the type `undefined reference to fi_freeinfo@FABRIC_1.9'` where `FABRIC_1.9` symbols are only defined in the newer version of libfabric.
> 
> Yeah, it is strange indeed. I unfortunately don't have any insight into the runners if there might actually be another libfabric installed somewhere on that runner/in that environment. Would you be ok with keeping the conflict for now or would that block you in reasonable use cases? @alalazo or @haampie would you have more information on the runner tha failed (or know who else to ask)?

Yeah I don't think you should add that conflict and we'll need to remove it anyway but that's a good experiment.

repos/spack_repo/builtin/packages/adios2/package.py
```diff
@@ -181,6 +181,7 @@ class Adios2(CMakePackage, CudaPackage, ROCmPackage):
         depends_on(
             "libfabric@1.6.0:", when=f"+sst platform={_platform}"
         )  # optional in EVPath and SST
+        conflicts("libfabric@2.5:")
```
msimberg:
```suggestion
```

repos/spack_repo/builtin/packages/mercury/package.py
```diff
@@ -70,6 +70,7 @@ class Mercury(CMakePackage):
         depends_on("libfabric@1.5:", when="@:2.0.1")
         depends_on("libfabric@1.7:", when="@2.1.0:")
         conflicts("libfabric@2:", when="@:2.4.0")
+        conflicts("libfabric@2.5:")
```
msimberg:
```suggestion
```

soumagne:
@msimberg I think I know why the CI is failing, it's because it's pulling intel-oneapi-mpi which also has its own copy of the libfabric library (`intel-oneapi-mpi@2021.17.2`)  and it's compiling mercury with mpi enabled (`mercury@2.4.1~bmi+boostsys+checksum~debug~hwloc~ipo+mpi+ofi+perf~psm~psm2+shared+sm~ucx~udreg`). So the issue then becomes that mercury is compiled against libfabric 2.5.0 from this package but then also linked to the libfabric from Intel MPI that has an older libfabric, hence creating the missing symbol errors. I'm not sure what is adding the +mpi to mercury (looks like that variant is added by ` stacks/e4s-oneapi/spack.yaml` ?) by default it would be off, I imagine it's added to run parallel tests etc.
I think the real solution for that should be that spack disables the use of the internal Intel MPI libfabric libfabric (see https://www.intel.com/content/www/us/en/docs/mpi-library/developer-guide-linux/2021-16/libfabric-support.html), I am not certain how to do that with spack and it does not seem to be supported at the moment but it seems that when sourcing the oneapi environment, one should then add something to that file: `repos/spack_repo/builtin/packages/intel_oneapi_compilers/package.py` to make sure that `-i_mpi_ofi_internal=0` is set and that the internal intel mpi libfabric library does not get used.
I would expect this also solves the similar issue with adios.

msimberg:
@soumagne that makes a lot of sense, thanks so much for digging into it! I was only able to reproduce the issue with intel-oneapi-mpi (but not with mpich or openmpi) but didn't understand what difference that would make. I'll do some more tests locally, but I see that `intel-oneapi-mpi` actually has an `external-libfabric` variant which is off in the failing pipelines (and which can be used to switch the `_mpi_ofi_internal` option), so this all fits your hypothesis. We may be able to get by with a more specific conflict on intel-oneapi-mpi or worst case still in adios2 and mercury, but constrained to `intel-oneapi-mpi`.

msimberg:
In https://github.com/spack/spack-packages/pull/3950/commits/5a1b66a2916bfae7713f6656a39bd88e9b8b8ba7 I added a `conflicts("libfabric", when="~external-libfabric")` to `intel-oneapi-mpi` which seems to have worked in CI at least. Some envs were reconcretized with `~external-libfabric` and `libfabric` was gone, and some of the others changed to `+external-libfabric` and they have an explicit libfabric spec.

@rscohn2 would mind having a look if this is ok for you? Perhaps @alalazo can comment as well if this seems safe to you? Is this overkill? As far as I understand (something like) this is one of the few valid use cases for a conflicts without `%` or `^`, but I'm not sure if this is safe to do here.

msimberg:
I think the patch that I added breaks compilation with `fabrics=cxi` because it touches `prov/cxi/configure.m4`. I'm doing some tests, but I'm considering change libfabric to always do the `autoreconf` stage instead of only `@main`. This is in principle more correct as well, since `@git.xxxxxx=1.2.3` versions currently don't run autoreconf as it should.

soumagne:
> I think the patch that I added breaks compilation with `fabrics=cxi` because it touches `prov/cxi/configure.m4`. I'm doing some tests, but I'm considering change libfabric to always do the `autoreconf` stage instead of only `@main`. This is in principle more correct as well, since `@git.xxxxxx=1.2.3` versions currently don't run autoreconf as it should.

you could also just add the same dependencies with when="@2.5.0 fabrics=cxi" (i.e. when this patch gets applied) while still limiting autoreconf to main ?

msimberg:
@soumagne yeah, I agree I could limit the dependencies conditionally, but figured it doesn't hurt much to have them always there, especially as it would also correctly allow building custom branches other than main (which as far as I know one can't currently filter for in `when`, i.e. something like `when="@<git-version>`). In a bigger environment I'd expect those dependencies to be pulled in anyway. I may too naive in thinking that this is fine though, so please let me know if you think it's important to limit the dependencies to the minimum possible.

jscook2345:
2.5.1 has dropped, if that helps anything here

msimberg:
> 2.5.1 has dropped, if that helps anything here

From my perspective this is just waiting for an approval (or concrete suggestions for what needs to be changed). Thanks for the note on 2.5.1, I'll add that here. I already included the cxi patch for 2.5.0 so that in itself has not been a blocker.

soumagne:
Since 2.5.1 has now dropped, I think it would be fair to say that you can simply drop the patch that you are applying to 2.5.0 since it's included in 2.5.1 and also drop the autoreconf changes, which I think should be constrained to the main branch to potentially avoid any autoreconf issues on past releases.

msimberg:
> Since 2.5.1 has now dropped, I think it would be fair to say that you can simply drop the patch that you are applying to 2.5.0 since it's included in 2.5.1 and also drop the autoreconf changes, which I think should be constrained to the main branch to potentially avoid any autoreconf issues on past releases.

I've removed the patch and added a version constraint on libcxi when `fabrics=cxi` and libfabric is at 2.5.0, exactly. Note that I made the version constraint for libcxi `14:` even though spack-packages doesn't yet have version 14. 14 has been tagged upstream though and it includes the [Added netns id in rgroup AC list to support RDMA over k8 pods](https://github.com/HewlettPackard/shs-libcxi/commit/4775e844e6ce8731db2a1b98d834fee62470bd25) commit that was assumed to be present previously.

albestro (approved):
LGTM

the only remaining doubt is about `conflicts` without any specifier about direct/indirect dependency. but since it is a support use-case, and it sounds reasonable in this case, I'd go for it and in case it can be improved later.

PR #3954, opened 2026-03-23
libsegfault: new package

etiennemlb:
This used to be in glibc. This is a very useful, simple and lightweight tool that hooks signals and dump a stack trace on crash.

It can be linked at build time or at runtime with LD_PRELOAD (the provided `catchsegv` script).

Then using addr2line the executable and an address from the stack trace, you can deduce the line at which the program failed.

@spackbot fix style

### DIFF
--- a/repos/spack_repo/builtin/packages/libsegfault/package.py
+++ b/repos/spack_repo/builtin/packages/libsegfault/package.py
@@ -19,7 +19,7 @@ class Libsegfault(AutotoolsPackage):
 
     license("GPL-2.0")
 
-    version("main", commit="ff16adff4a6af738eb4deabfb0eb107f6fa6e048")
+    version("2023-07-25", commit="ff16adff4a6af738eb4deabfb0eb107f6fa6e048")
 
     depends_on("c", type="build")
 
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/libsegfault/package.py
```diff
@@ -0,0 +1,26 @@
...
+    LibSegFault."""
+
+    homepage = "https://github.com/zatrazz/glibc-tools"
+    git = "https://github.com/zatrazz/glibc-tools.git"
+
+    maintainers("etiennemlb")
+
+    license("GPL-2.0")
+
+    version("main", commit="ff16adff4a6af738eb4deabfb0eb107f6fa6e048")
```
tldahlgren: Interesting. This does not reflect "standard" practice. People almost always specify the branch that matches the version name.

Providing a `commit` is more secure but could lead to confusion should the repository ever get updated.

Typically, in cases like this where there is no `tag`, people will name the version with the date of the commit, for example. See the example under the `Commits` subsection of https://spack.readthedocs.io/en/latest/packaging_guide_creation.html#git.

tldahlgren (changes requested):
There only a few exceptions to using a commit for a named branch out of the thousands of built-in packages.

Let me know how you want to proceed.

repos/spack_repo/builtin/packages/libsegfault/package.py
etiennemlb: Agreed I fixed that version using the commit date?

tldahlgren (approved):
Thanks.

PR #3959, opened 2026-03-23
py-pysam: adding v0.23.3

snehring:
version bump

### DIFF
--- a/repos/spack_repo/builtin/packages/py_pysam/package.py
+++ b/repos/spack_repo/builtin/packages/py_pysam/package.py
@@ -32,7 +32,7 @@ class PyPysam(PythonPackage):
 
     depends_on("py-setuptools@59.0:", when="@0.21:", type="build")
     depends_on("py-setuptools", type="build")
-    depends_on("py-cython@0.29.30:3", when="@0.23.3:", type="build")
+    depends_on("py-cython@0.29.12:3", when="@0.23.3:", type="build")
     depends_on("py-cython@0.29.30:2", when="@0.21", type="build")
     depends_on("py-cython@0.29.12:2", when="@0.18:0.19", type="build")
     depends_on("py-cython@0.21:2", when="@0.14:0.15", type="build")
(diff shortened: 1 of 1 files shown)


tldahlgren (commented):
Confirmed the version sha256,.

repos/spack_repo/builtin/packages/py_pysam/package.py
```diff
@@ -27,14 +28,18 @@ class PyPysam(PythonPackage):
...
 
+    depends_on("python@3.8:", type=("build", "run"))
+
     depends_on("py-setuptools@59.0:", when="@0.21:", type="build")
     depends_on("py-setuptools", type="build")
-    depends_on("py-cython@0.29.30:2", when="@0.21:", type="build")
-    depends_on("py-cython@0.29.12:2", when="@0.18:", type="build")
-    depends_on("py-cython@0.21:2", when="@0.14:", type="build")
-    depends_on("py-cython@0.17:2", type="build")
+    depends_on("py-cython@0.29.30:3", when="@0.23.3:", type="build")
```
tldahlgren: Technically the lower bound for this version appears, per `pyproject.toml`, to be `0.29.12` (see https://github.com/pysam-developers/pysam/blob/cefbaa9079b0b2ad65dd8d60a532bc3eb31389ee/pyproject.toml#L21).

tldahlgren (commented):
Let me know if this should just be merged as is without the lower bound in the comment.

repos/spack_repo/builtin/packages/py_pysam/package.py
snehring: Yeah, it felt weird to decrease the lower bound compared to the previous version. I can change it though.

tldahlgren: You're right. It does seem weird. I would've merged as is but thanks for the fix.

tldahlgren (approved):

PR #3960, opened 2026-03-23
gfan: new package

d-torrance:
Gfan computes Gröbner fans and tropical varieties of polynomial ideals, with tools for universal Gröbner bases, tropical curves, hypersurfaces, and related structures in tropical geometry.


tldahlgren (approved):
Confirmed the homepage and version sha256.

PR #3961, opened 2026-03-23
prrte: add patch to fix segfault on macOS

mcmehrtens:
- Adds a patch based on https://github.com/openpmix/prrte/pull/2417 to fix a bug where PRRTE would segfault on some Apple ARM architectures


tldahlgren (approved):
LGTM

PR #3962, opened 2026-03-23
Ruff & Ty & Uv: Make It Cargo

johnwparent:
Ruff and Ty have been PythonPackages up to this point. This bring in an unnecessary python dep and makes the dep graph deeper than need be for both. 

Move ruff and ty to build purely from rust sources.
Builds were based on the arguments being used to drive `cargo build` from maturin. 

Deprecates all python based versions of ruff and ty.

Note: ty builds from the ruff repo, but due to differing version cadences, is most easily packaged as a disparate package in Spack. 

Added UV on request


trws:
I'm generally supportive of the change, but is there a mechanism we can use to make this easier on users? If the `ruff` and `ty` builds work the same as the `py-` versions for all practical purposes, can we keep the `py-` names as aliases for example?

adamjstewart:
Should uv be treated similarly? We don't have to change the names at all if we don't want to.

johnwparent:
> Should uv be treated similarly? 

Sure, why not. Added.

> We don't have to change the names at all if we don't want to.

A package named `py-` that is not a python package seems misleading and isn't really the name of any of these things.

I left the old packages for this reason, to allow users who rely on py-ty and py-ruff at the moment to use them, but with depcrecations to give us an off ramp to names that better represent the packages behavior. 

Further, using `py-ty` and `py-ruff` for these new packages may cause issues as all of the version hashes have changed.

I'm not aware of a naming alias function, but if there is one that doesn't impact concretization and wouldn't cause issues with the changing version hashes, then I'm fine with that.

johnwparent:
@spackbot fix style

adamjstewart:
For naming, the official guidance I wrote in https://spack.readthedocs.io/en/latest/build_systems/pythonpackage.html#choosing-a-package-name is:

> When in doubt, think about how this package will be used. Is it primarily a Python library that will be imported in other Python scripts? Or is it a command-line tool, or C/C++/Fortran program with optional Python modules? The former should be prepended with `py-`, while the latter should not.

These tools clearly fall into the latter, as they are command-line tools that should never be imported. People have also argued that we should rename `py-pip` to `pip` for the same reason. I don't want to blow this PR up into a bigger discussion, but I think we should consider that rename in a separate PR.

I'm unaware of a package name alias mechanism, but I'm fine with this PR without such an alias.

One request: can we still set `PYTHONPATH` appropriately? This will ensure that `python -m uv` still works correctly.

repos/spack_repo/builtin/packages/ruff/package.py
```diff
@@ -0,0 +1,48 @@
...
+        depends_on("rust@1.85:", when="@0.11.11:")
+        depends_on("rust@1.84:", when="@0.11.4:")
+        depends_on("rust@1.83:", when="@0.9.8:")
+        depends_on("rust@1.80:", when="@0.7.1:")
+        depends_on("rust@1.76:", when="@0.5.6:")
+        depends_on("rust@1.71:")
+
+    def url_for_version(self, version):
+        v = ""
+        if version < Version("0.5.0"):
```
adamjstewart: We don't have any versions older than this so this whole function could be removed

repos/spack_repo/builtin/packages/uv/package.py
```diff
@@ -0,0 +1,61 @@
...
+
+class Uv(CargoPackage):
+    """An extremely fast Python package and project manager, written in Rust."""
+
+    homepage = "https://github.com/astral-sh/uv"
+    git = "https://github.com/astral-sh/uv.git"
+    url = "https://github.com/astral-sh/uv/archive/refs/tags/0.11.0.tar.gz"
+
+    license("APACHE 2.0 or MIT")
+    maintainers("johnwparent")
```
adamjstewart: You can also add me as maintainer here, not that I know anything about rust, but I at least use the package and can help update it.

repos/spack_repo/builtin/packages/uv/package.py
```diff
@@ -0,0 +1,61 @@
...
+
+
+class Uv(CargoPackage):
+    """An extremely fast Python package and project manager, written in Rust."""
+
+    homepage = "https://github.com/astral-sh/uv"
+    git = "https://github.com/astral-sh/uv.git"
+    url = "https://github.com/astral-sh/uv/archive/refs/tags/0.11.0.tar.gz"
+
+    license("APACHE 2.0 or MIT")
```
adamjstewart:
```suggestion
    license("MIT OR Apache-2.0")
```
This is actually the correct SPDX specification.

repos/spack_repo/builtin/packages/uv/package.py
```diff
@@ -0,0 +1,61 @@
...
+
+    build_directory = "crates/uv"
+
+    executables = ["^uv$"]
+
+    version("0.11.0", sha256="d61c168d4a3b0dad3c2c2a73c591fe7a2bbad96ea5a5e15f857d03308e15ca50")
+    version("0.10.12", sha256="73c256e80ac1b6e030aeb0643a80d84c32f327fbe09450ff39069103503f46c9")
+    version("0.9.30", sha256="c89f7d66c6aa8324e14f6b07b35af64e37368d6cb4e2b93d8bb7fdd2159f78f6")
+    version("0.8.24", sha256="33f4ad5c8fe980e35c2eb79d9d3bfbada3d1a1ef63f0aa7c8ce0d3affce79617")
+    with default_args(deprecated=True):
```
adamjstewart: You can remove all of these. Honestly, I would personally only keep the latest version of all packages since they are new, I don't know anyone who requires older versions and it adds to the package complexity.

adamjstewart:
> One request: can we still set `PYTHONPATH` appropriately? This will ensure that `python -m uv` still works correctly.

Oh, another big difference between PythonPackage and CargoPackage is that we no longer have anything installed to site-packages. This might be a deal breaker. Still completely fine with the rename, but we may want to keep these as PythonPackage to ensure that `python -m uv` still works.

johnwparent:
> > One request: can we still set `PYTHONPATH` appropriately? This will ensure that `python -m uv` still works correctly.

Yeah that seems like a good change, thanks for suggesting it, it wouldn't have occurred to me that folks would want to invoke ruff/ty like that (clearly). 

> Oh, another big difference between PythonPackage and CargoPackage is that we no longer have anything installed to site-packages. This might be a deal breaker. Still completely fine with the rename, but we may want to keep these as PythonPackage to ensure that `python -m uv` still works.

The core intent here is to removes python from the ruff/ty/uv packages as it's not required to build/link/run them. We could maybe add a `+py` variant that bridges some of the gap of this behavior.

repos/spack_repo/builtin/packages/uv/package.py
johnwparent: Thanks

repos/spack_repo/builtin/packages/uv/package.py
johnwparent: Ok, I'll update it in the "py" uv package as well since this was copied.

repos/spack_repo/builtin/packages/uv/package.py
johnwparent: Was trying to mirror more or less the versions the old ruff/ty/uv packages had so users weren't surprised, but sounds good to me.

adamjstewart:
Is there a problem with Python being in the DAG? They aren't really useful without a Python installation to use them with. Yes, Python isn't required to use them, but no one is using ruff without Python files to format.

johnwparent:
> Is there a problem with Python being in the DAG? They aren't really useful without a Python installation to use them with. Yes, Python isn't required to use them, but no one is using ruff without Python files to format.

Yes
a. it's extra , unnecessary cruft in the DAG, and as you comment above, it's better to reduce package complexity. 
b. It adds an entire python install to the build of ty/ruff/uv, which is adding a non trivial build on top of an already involved build
c. the run dep on python means pulling ruff/ty/uv from the buildcache will require a python, either to be pulled or present (also adds a wrinkle when bootstrapping ruff/ty as it means we have to splice python in).

> but no one is using ruff without Python files to format

That is correct, but those files aren't coming from Spack (spack develop non withstanding) so I don't see why we need to be bringing an entire python ecosystem on the spack side outside of `python -m ruff` which I think we can accomplish without installing an entire venv


Edit: deleted example as it was too contrived to be useful

johnwparent:
To be clear, I would like for this package to continue working for users as it has been

Perhaps something like
```
variant("python")

depends_on("python-venv", when="+python")

def install():
 super().install()
 install ruff into venv
```

adamjstewart:
In my experience, it's really hard to tell what the correct site-packages directory is without a dependency on Python itself. 

We could probably keep things as is and wait for someone to complain? I always see conflicting advice about using `pytest` vs. `python -m pytest` and same for most other Python tools. The same is now true for `pytest` vs. `uv run pytest` of course.

johnwparent:
> In my experience, it's really hard to tell what the correct site-packages directory is without a dependency on Python itself.

That's what the optional python dep keyed on the variant would be for.

> We could probably keep things as is and wait for someone to complain? 

As in merge as is? I'm fine with that. 

> I always see conflicting advice about using `pytest` vs. `python -m pytest` and same for most other Python tools. The same is now true for `pytest` vs. `uv run pytest` of course.

Yeah, I've seen similar discussions as well... maybe we should let more folks weigh in?

adamjstewart (dismissed):
* Name change: ✅ 
* Base class change: not convinced it's useful, but I'm fine with it if others find it useful and no one complains

johnwparent:
Alright, it seems like ruff actually goes out if its way to make `python -m ruff` work as expected. I'd prefer to avoid breaking that workflow for folks in the next packages release. Going to try to stand-up that behavior here.

adamjstewart:
Could we make the name change first, merge that, and then consider switching to CargoPackage in a separate PR? I want the name change sooner rather than later, and I feel like we'll eventually give up on CargoPackage.

johnwparent:
@adamjstewart ... I over-complicated this. We just need to inherit from both `CargoPackage` and `PythonPackage` and then switch the build system to pip (maturin backend) as needed.

Still working out a hitch where there's no python in the dag somehow, but I pushed up a reformatted ruff for your perusal in terms of recipe layout.

adamjstewart:
I like that idea much more, glad we went back and forth so many times!

johnwparent:
> I like that idea much more, glad we went back and forth so many times!

yeah, a bit of a 🤦 for not seeing this immediately but so it goes. 

Where's the best place to document this? I feel like this is best implemented by example and doc

repos/spack_repo/builtin/packages/ruff/package.py
```diff
@@ -0,0 +1,48 @@
...
+#
+# SPDX-License-Identifier: (Apache-2.0 OR MIT)
+
+from spack_repo.builtin.build_systems.cargo import CargoPackage
+from spack_repo.builtin.build_systems.python import PythonExtension
+
+from spack.package import *
+
+
+class Ruff(CargoPackage, PythonExtension):
```
adamjstewart:
```suggestion
class Ruff(CargoPackage, PythonPackage):
```
I think this is what you mean? So it inherits from both build systems and dynamically chooses which deps to add.

adamjstewart:
@tgamblin is suggesting we move all Python-related packages to a separate `py` namespace such that `py.ruff` and `ruff` both work. Maybe we should hold off on the rename for now?

adamjstewart:
> Where's the best place to document this? I feel like this is best implemented by example and doc

We already have too many docs on multi-build system packages, I wouldn't add any more docs. 99.9% of users will just `spack install ruff` and continue on with their lives. It's not like this is a widely requested feature.

tgamblin:
> @tgamblin is suggesting we move all Python-related packages to a separate py namespace such that py.ruff and ruff both work. Maybe we should hold off on the rename for now?

If it's `ruff` in the PR right now I think it's fine.

johnwparent:
> We already have too many docs on multi-build system packages, I wouldn't add any more docs. 99.9% of users will just `spack install ruff` and continue on with their lives. It's not like this is a widely requested feature.

Agreed not at all relevant to users, but for packagers, I can see this case coming up semi-frequently. This idiom could extend to some C++ packages as well that have optional python interfaces. I was basically just thinking "if you want an optional wheel install inherit from both python and whatever else and set your build system variant as appropriate" but if its obvious enough then I'm fine not documenting it.

johnwparent:
@spackbot fix style

johnwparent:
@spackbot fix style

johnwparent:
@spackbot fix style

johnwparent:
@spackbot fix style

adamjstewart:
> > @tgamblin is suggesting we move all Python-related packages to a separate py namespace such that py.ruff and ruff both work. Maybe we should hold off on the rename for now?
> 
> If it's `ruff` in the PR right now I think it's fine.

May be better to rename all Python packages in one go instead of one at a time. If it's all packages, users only need to be surprised once.

johnwparent:
> > > @tgamblin is suggesting we move all Python-related packages to a separate py namespace such that py.ruff and ruff both work. Maybe we should hold off on the rename for now?
> > 
> > 
> > If it's `ruff` in the PR right now I think it's fine.
> 
> May be better to rename all Python packages in one go instead of one at a time. If it's all packages, users only need to be surprised once.

I think since this is also a build system/overall refactor, we can get this in sooner (although after we land style ruff would be ideal :) )

johnwparent:
@spackbot fix style

adamjstewart:
@spackbot fix style

adamjstewart:
Apologies for the accidental ping, testing to see whether or not spackbot is working with ruff now. The answer is yes, although we should probably first sync the branch on which ruff runs. Anyways, ignore the request for review!

hppritcha (approved):
not sure why the ws change but its fine with me

adamjstewart:
Note that uv was converted in #5145, would be happy to see the same treatment for the others as well.

hyoklee (approved):

PR #3970, opened 2026-03-24
root: recipe cleanup

jmcarcell:
Modify ranges and checks that now are always true because of the minimum
version. Delete patches that will never apply.


marcmengel (approved):
Looks good; pretty much removing references to older versions, and related patches.

gartung (approved):

PR #3977, opened 2026-03-24
nccl package: add git attribute

eugeneswalker:
Add `git` attribute to nccl package so we can `spack install nccl@git.<commit>`


RikkiButler20 (approved):
LGTM

PR #3982, opened 2026-03-24
netcdf-fortran: Fixes for the NAG Fortran compiler on darwin

billsacks:
# Brief summary:

* netcdf-fortran: Fixes for the NAG Fortran compiler on darwin,  Fix NAG PIC flag
* netcdf-fortran: Apply fixes to gfortran and nag just for older versions
  See point (4) in https://github.com/spack/spack-packages/pull/3982#discussion_r2990092964 for rationale.
* Patch libtool for NAG on Darwin. see https://github.com/spack/spack-packages/pull/3982#discussion_r2990671326 for details.

# Details

These fixes were needed to build netcdf-fortran with the NAG Fortran compiler on a Mac.

(1) A patch for libtool's ltmain.sh to have it correctly identify NAG when invoked by spack: spack uses the full compiler path, which doesn't work with the out-of-the-box logic in ltmain.sh. This issue has nothing to do with netcdf-fortran itself. (Note a similar fix needed for openmpi with flang [here](https://github.com/open-mpi/ompi/pull/13772). The OpenMPI patch is in the OpenMPI repo itself, not the spack package; I wasn't sure if there was a mechanism to put this in the actual netcdf-fortran repo.)

(2) A fix for NAG's use of -PIC instead of the more common -fPIC. This one feels like something I might be doing wrong in my spack setup... e.g., maybe there's some way for me to specify that in my compiler configuration, but I couldn't figure out how. (Note that I'm using NAG installed outside of spack, not installed through spack.)

(3) Explicitly telling NAG to use the preprocessor on all Fortran files. My sense is that this is needed on a Mac because the file system is case insensitive, so NAG doesn't distinguish between F90 and f90 files the way it does elsewhere. I'm not sure if my fix here is the best / right fix for this issue: I have limited experience with spack so did something that worked but I'm not sure if it's spack best practice.

(https://github.com/esmf-org/esmf/wiki/cgdm-fairy#installing-netcdf-2 gives some more context.)


billsacks:
@spackbot fix style

skosukhin:
@billsacks, Thank you. I would expect at least some of the issues to be covered in [`_do_patch_libtool`](https://github.com/spack/spack-packages/blob/85566405b40de909029b0af15452fffe2c8249c9/repos/spack_repo/builtin/build_systems/autotools.py#L375). Could you check, please?

@alalazo do you know if we have a function that checks whether the filesystem is case-insensitive, or if there is an idiomatic solution for that?

billsacks:
> I would expect at least some of the issues to be covered in [`_do_patch_libtool`](https://github.com/spack/spack-packages/blob/85566405b40de909029b0af15452fffe2c8249c9/repos/spack_repo/builtin/build_systems/autotools.py#L375). Could you check, please?

Sorry, can you explain what / how you want me to check? I did confirm that all of the changes here were needed for the netcdf-fortran build to succeed. In particular, without the patch to ltmain.sh, the build failed with a message like "unknown argument: compatibility_version" (or something like that). Is there something more that you'd suggest I check?

I should acknowledge that my spack experience is fairly limited, so it's very possible that these fixes don't follow spack best practices.

repos/spack_repo/builtin/packages/netcdf_fortran/package.py
```diff
@@ -79,7 +85,13 @@ def flag_handler(self, name, flags):
                 flags.append(self.compiler.cc_pic_flag)
         elif name == "fflags":
             if "+pic" in self.spec:
-                flags.append(self.compiler.f77_pic_flag)
+                if self.spec.satisfies("%fortran=nag"):
+                    # NAG uses -PIC, but the f77_pic_flag may be incorrectly set to the
+                    # more common -fPIC. This may be user error, but we work around this
+                    # issue here.
+                    flags.append("-PIC")
```
skosukhin: @alalazo, this looks like an attempt to address a bug introduced https://github.com/spack/spack/pull/45189, which dropped [`-PIC`](https://github.com/spack/spack/blob/32bf55f1fb5180c01021c404d8e58c51ffc7e297/lib/spack/spack/compilers/nag.py#L86).

repos/spack_repo/builtin/packages/netcdf_fortran/package.py
```diff
@@ -79,7 +85,13 @@ def flag_handler(self, name, flags):
...
+                else:
+                    flags.append(self.compiler.f77_pic_flag)
```
skosukhin: If I understand it correctly, the "new" way to get the flag is `self["fortran"].pic_flag`. It makes sense to do that in other places as well.

repos/spack_repo/builtin/packages/netcdf_fortran/package.py
```diff
@@ -91,6 +103,13 @@ def flag_handler(self, name, flags):
                 # The following flag forces the compiler to produce module
                 # files with lowercase names.
                 flags.append("-ef")
+            elif self.spec.satisfies("%fortran=nag"):
+                if sys.platform == "darwin":
+                    # The MacOS file system is case-insensitive. NAG therefore treats .F90
+                    # files as .f90 files, and so doesn't run them through its
+                    # preprocessor. So add -fpp to force NAG to run the preprocessor on
+                    # all Fortran files.
+                    flags.append("-fpp")
```
skosukhin: I guess this is relevant for all compilers on case-insensitive filesystems, so it might deserve `self["fortran"].pp_flag` or something like that.

Ideally, the flag is added for a case-insensitive filesystem, not just for `darwin` (I personally find it strange to build anything on a case-insensitive filesystem, so my `darwin` is case-sensitive and I don't think I need this).

repos/spack_repo/builtin/packages/netcdf_fortran/nag_linking.patch
```diff
@@ -0,0 +1,12 @@
...
++++ b/ltmain.sh	2026-03-22 18:17:08
+@@ -9360,7 +9360,8 @@
+ 	  xlcverstring="$wl-compatibility_version $wl$minor_current $wl-current_version $wl$minor_current.$revision"
+ 	  verstring="-compatibility_version $minor_current -current_version $minor_current.$revision"
+           # On Darwin other compilers
+-          case $CC in
++          func_cc_basename "$CC"
++          case $func_cc_basename_result in
+               nagfor*)
+                   verstring="$wl-compatibility_version $wl$minor_current $wl-current_version $wl$minor_current.$revision"
```
skosukhin: We replace the value for `wl` [here](https://github.com/spack/spack-packages/blob/85566405b40de909029b0af15452fffe2c8249c9/repos/spack_repo/builtin/build_systems/autotools.py#L415), and I am unsure whether the `verstring` is set to the originally intended value (although I would not be surprised if it still works).

repos/spack_repo/builtin/packages/netcdf_fortran/package.py
```diff
@@ -53,6 +54,11 @@ class NetcdfFortran(AutotoolsPackage):
         depends_on("mpi", when="^netcdf-c+parallel-netcdf")
         depends_on("mpi", when="^hdf5+mpi~shared")
 
+    # This patch to ltmain.sh is needed at link time to properly detect NAG when the
+    # compiler is specified using a full path (as Spack does). (It is only needed for NAG,
+    # but it is safe to always apply it.)
+    patch("nag_linking.patch")
```
skosukhin: The patch is relevant for all Libtool-based packages on macOS with NAG. We just don't have that many users with that set-up. We should avoid adding this patch to every package and address the problem in the aforementioned function that runs for all packages. We already add [`with_nag`](https://github.com/spack/spack-packages/blob/85566405b40de909029b0af15452fffe2c8249c9/repos/spack_repo/builtin/build_systems/autotools.py#L485) to the libtool script. I don't know if that helps, though. Unfortunately, in this case, it might be tricky to come up with a robust regexp for the replacement, but we have to.

billsacks:
Thanks a lot for your review, @skosukhin ! I agree that most / all of these issues feel like band-aids for upstream issues, and I don't mind if this PR gets closed in favor of getting these fixed upstream.

Some replies:

(1) For the ltmain.sh patch, I'm not entirely following your comments, but in case it helps:

The current code in ltmain.sh looks like this:

```bash
          # On Darwin other compilers
          case $CC in
              nagfor*)
                  verstring="$wl-compatibility_version $wl$minor_current $wl-current_version $wl$minor_current.$revision"
                  ;;
              *)
                  verstring="-compatibility_version $minor_current -current_version $minor_current.$revision"
                  ;;
          esac
```

Notably, elsewhere in ltmain.sh, there is code using func_cc_basename for a similar check:

```bash
        solaris*)
          func_cc_basename "$CC"
          case $func_cc_basename_result in
          CC* | sunCC*)
            func_suncc_cstd_abi

            if test no != "$suncc_use_cstd_abi"; then
              func_append postdeps ' -library=Cstd -library=Crun'
            fi
            ;;
          esac
```

I completely agree that it would make sense to fix this in one place rather than every package. At least with the current ltmain.sh, it seems like a robust fix would be changing any occurrence of `case $CC in` to:

```bash
          func_cc_basename "$CC"
          case $func_cc_basename_result in
```

(There is currently only the one occurrence of `case $CC in`, but I think any new occurrences of that would also warrant that replacement.)

Ideally this would actually be fixed in libtool, but I'm not sure what the process is to get it fixed there.

(2) For the PIC flag, what you say makes sense. I did just try getting `self["fortran"].pic_flag` with NAG on my system, and it still returns "-fPIC", but maybe that's expected based on your other comment.

(3) For "-fpp": Your thoughts about making this more general make sense to me.

repos/spack_repo/builtin/packages/netcdf_fortran/package.py
skosukhin: Ok. Let's do it step by step. Revert this, add `pic_flag = "-PIC"` right before [this line](https://github.com/spack/spack-packages/blob/cda12facbee13e53048fd006328978dbb0387832/repos/spack_repo/builtin/packages/nag/package.py#L108) and check if it does the job.

billsacks: > Ok. Let's do it step by step. Revert this, add `pic_flag = "-PIC"` right before [this line](https://github.com/spack/spack-packages/blob/cda12facbee13e53048fd006328978dbb0387832/repos/spack_repo/builtin/packages/nag/package.py#L108) and check if it does the job.

That doesn't work for me. But maybe that's because I'm not installing NAG via spack: I tried to install NAG via spack but ran into trouble, maybe because it currently only supports x86 and I'm trying to install this on an M1 (ARM) Mac:

https://github.com/spack/spack-packages/blob/cda12facbee13e53048fd006328978dbb0387832/repos/spack_repo/builtin/packages/nag/package.py#L44-L48

So I'm installing NAG myself, rather than via a spack install. Does your suggested fix assume that I'm installing NAG via spack?

I was hoping that there would be a way to specify this in my compiler definition in my `.spack/packages.yaml` (after running `spack compiler find`). I couldn't figure out whether that's possible - and, if so, how to do it - but if there's a way to do it that way, I'd be happy to try it out.

skosukhin: 1. [There is a way to specify compiler flags you like](https://spack.readthedocs.io/en/latest/configuring_compilers.html#configuring-compilers).
2. If you run `spack config blame packages`, you'll see something like
    ```
    nag:
      externals:
      - spec: nag@7.2.7200
        prefix: /path/to/nag
        extra_attributes:
          compilers:
            fortran: /path/to/nag/bin/nagfor
    ```
    That `spec: nag` is the link between your manual installation of NAG and `nag/package.py`.
3. I have just checked, and adding `pic_flag = "-PIC"` as I described above worked as expected.
4. There is some legacy that needs to be addressed: we inject `-fallow-argument-mismatch` when `self.spec.satisfies("%gcc@10:")`, which is incorrect after the big MR that I mentioned above. Now, `gcc` is a direct dependency of `netcdf-fortran` when building with `nag`, and we pass the flag originally intended for `gfortran 10+` to `nagfor`, which breaks the building. The mismatch issue was addressed in `netcdf-fortran` upstream repo long ago, and we don't need that for newer versions of the package. Could you, please, apply the following patch:
    ```patch
    --- a/repos/spack_repo/builtin/packages/netcdf_fortran/package.py
    +++ b/repos/spack_repo/builtin/packages/netcdf_fortran/package.py
    @@ -79,10 +79,14 @@ class NetcdfFortran(AutotoolsPackage):
             elif name == "fflags":
                 if "+pic" in self.spec:
                     flags.append(self.compiler.f77_pic_flag)
    -            if self.spec.satisfies("%gcc@10:"):
    +            if self.spec.satisfies("@:4.5.2"):
    +              if self.spec.satisfies("%fortran=gcc@10:"):
                     # https://github.com/Unidata/netcdf-fortran/issues/212
                     flags.append("-fallow-argument-mismatch")
    -            elif self.compiler.name == "cce":
    +              elif self.spec.satisfies("%fortran=nag"):
    +                # https://github.com/Unidata/netcdf-fortran/issues/218
    +                flags.appen("-mismatch_all")
    +            if self.spec.satisfies("%fortran=cce"):
                     # Cray compiler generates module files with uppercase names by
                     # default, which is not handled by the makefiles of
                     # NetCDF-Fortran:
    ```

I'm not sure why `pic_flag = "-PIC"` doesn't work for you, but adding the flag via the `flag_handler` does. Another problem is that your changes address `netcdf-fortran+pic` only. Without the suggested solution, I expect `netcdf-fortran~pic+shared` to fail (at the linking stage). Could you apply the patch from `4.` and try again? If it still fails, I would like to take a look at the relevant logs and error messages. Thanks!

billsacks: Sorry! I realized that I made a mistake in my original testing of adding `pic_flag = "-PIC"`: I made this change in the wrong place (I had cloned a separate copy of the spack-packages repo for poking around in, and mistakenly made the change there). When I make this change in the right place, and revert my PIC-related changes in the netcdf_fortran package, this indeed solves that problem. Thank you, and sorry about that mistake on my part initially! But I did appreciate your explanations / clarifications.

Since that works, is it still worth my trying the change in (4)? I wasn't clear on whether that's still useful for me to try given that my correct attempt at your first suggestion now works.

skosukhin: The changes in `4.` are relevant in any case, I just have no idea when I get the time to create a PR for that. Hence, it would be nice to get them in this one. After all, it's relevant for NAG. Additionally, please replace `self.compiler.cc_pic_flag` with `self["c"].pic_flag` and `self.compiler.f77_pic_flag` with `self["fortran"].pic_flag`. Thank you.

repos/spack_repo/builtin/packages/netcdf_fortran/package.py
skosukhin: Adding `pp_flag` might be too much for this PR. Let's do the following:
1. When on `darwin` with NAG, add "-fpp".
2. When on `darwin` with any other compiler, add "-cpp".
It would be great if you could check whether `2.` makes it possible to build with `gcc` on your machine. I don't think anybody is going to build on `darwin` with anything, but `nagfor`, `gfortran` or `flang`.

repos/spack_repo/builtin/packages/netcdf_fortran/package.py
skosukhin: > For the ltmain.sh patch, I'm not entirely following your comments

There is a number of Libtool issues that we address for all `Autotools` packages by patching their `libtool` scripts right after the `configure` stage. One of them is [fixing the `wl` variable](https://github.com/spack/spack-packages/blob/85566405b40de909029b0af15452fffe2c8249c9/repos/spack_repo/builtin/build_systems/autotools.py#L415). If I remember correctly, if NAG is hidden behind an MPI compiler wrapper, Libtool does not recognize it (because `FC` is `mpif90`) and leaves the variable empty. This is similar to what you are trying to solve here, but I now see that it is not relevant for `netcdf-fortran`, which never uses the MPI Fortran compiler.

Unfortunately, it's not so easy to write a regular expression that captures the code in the `libtool` script (`ltmain.sh` becomes part of the `libtool` script generated by the `configure` script) that we need to modify (what your patch does).

You can find an example of another tricky replacement [here](https://github.com/spack/spack-packages/blob/85566405b40de909029b0af15452fffe2c8249c9/repos/spack_repo/builtin/build_systems/autotools.py#L504).

Basically, we need to add [here](https://github.com/spack/spack-packages/blob/85566405b40de909029b0af15452fffe2c8249c9/repos/spack_repo/builtin/build_systems/autotools.py#L530) something like:
```patch
--- a/repos/spack_repo/builtin/build_systems/autotools.py
+++ b/repos/spack_repo/builtin/build_systems/autotools.py
@@ -530,6 +530,15 @@ To resolve this problem, please try the following:
                 stop_at=stop_at,
             )
 
+            if sys.platform == "darwin":
+                # Libtool does not recognize NAG behind the compiler wrappers:
+                x.filter(
+                    regex=r"(\s*case\s+)(\$CC)(\s+in\s*)$",
+                    repl="\\1`test \"x$with_nag\" = xyes && echo nagfor || echo \\2`\\3",
+                    start_at="# On Darwin other compilers",
+                    stop_at="esac",
+                )
+
     @property
     def configure_directory(self) -> str:
         """Return the directory where 'configure' resides."""
```

This reminded me of [my favourite line in `ltmain.sh`](https://cgit.git.savannah.gnu.org/cgit/libtool.git/tree/build-aux/ltmain.in?h=v2.6.0&id=e40fdc22cf727fb885f9df7d9affa827e3253d1c#n609).

Anyway, could you revert your patching and try the patch above? Thanks!

repos/spack_repo/builtin/packages/netcdf_fortran/package.py
billsacks: For whatever reason, gfortran doesn't have a problem knowing it should run the preprocessor on F90 files on my machine; only NAG has that problem. So adding "-cpp" is unnecessary. (I built netcdf_fortran with gfortran successfully a few months ago; I haven't retried recently, but I did just try a small test program that required preprocessing, in a file with extension F90, and gfortran worked fine on that file despite NAG requiring an explicit addition of `-fpp`.) (I could still add "-cpp" for other compilers if you want me to, but I have no way of testing to see if that solves any problems, since I'm not having problems without it.)

I also see that [NAG mentions this being an issue for Windows](https://support.nag.com/nagware/np/r71_doc/manual/compiler_2_3.html), too. Would you want this generalized, or is just handling Mac/darwin enough for now?

skosukhin: I don't remember what is the prepocessing needed in `netcdf-fortran` for but I would check the gcc build log. There might be signs of the preprocessor directives being simply ignored.

skosukhin: Sorry, replied too early. If you absolutely sure this is not needed for gfortran, let's add the flag only for NAG.

billsacks: > If you absolutely sure this is not needed for gfortran, let's add the flag only for NAG.

I'm sure on my system. (I compiled a test program that would have very clearly failed if it wasn't run through the preprocessor.) I can't speak for other systems.

billsacks:
@skosukhin - thank you very much for all of your guidance here! I think that my just-pushed commits address all of your suggestions, but it would be great if you could double-check. I am much happier with these fixes now that they feel more like they're fixing the true issues rather than just band-aids for this package!

repos/spack_repo/builtin/packages/netcdf_fortran/package.py
skosukhin: Good enough for me. Thanks. Let's just handle NAG only.

billsacks:
@spackbot fix style

skosukhin (approved):

mathomp4:
Closes #4453

I'm not sure if Github will see that in a comment, but if this goes in first, then my PR at #4453 is moot.

repos/spack_repo/builtin/build_systems/autotools.py
```diff
@@ -528,6 +529,16 @@ def _do_patch_libtool(self) -> None:
                 stop_at=stop_at,
             )
 
+            if sys.platform == "darwin":
```
alalazo: Does this work:
```suggestion
            if self.spec.satisfies("platform=darwin"):
```
?

repos/spack_repo/builtin/packages/netcdf_fortran/package.py
```diff
@@ -76,21 +77,32 @@ class NetcdfFortran(AutotoolsPackage):
...
+            if self.spec.satisfies("%fortran=cce"):
                 # Cray compiler generates module files with uppercase names by
                 # default, which is not handled by the makefiles of
                 # NetCDF-Fortran:
                 # https://github.com/Unidata/netcdf-fortran/pull/221.
                 # The following flag forces the compiler to produce module
                 # files with lowercase names.
                 flags.append("-ef")
+            elif self.spec.satisfies("%fortran=nag"):
+                if sys.platform == "darwin":
```
alalazo:
```suggestion
            elif self.spec.satisfies("%fortran=nag platform=darwin"):
```

skosukhin: Isn't `platform` in the spec the target one, and `sys.platform` is the one we build on? I understand that the chance they don't match in this case is zero. I'm only asking about the semantics.

alalazo: Right now we just have the target of the node - `sys.platform` is just a workaround for cases at the class level where we don't support conditional values, i.e. in most cases versions of binary packages .

billsacks: @alalazo and @skosukhin - Since I don't have as much experience/knowledge of this as you do, I'm happy to go with any changes you feel are best here.

bernhardkaindl:
@billsacks Review by GitHub Copilot:

The earlier feedback from skosukhin appears resolved, and I see no clearly unresolved blocking discussion threads from the data returned. I do have one new review concern on the current diff.

Discussion status

- Resolved / addressed
  - Move PIC handling to the NAG compiler package via pic_flag = "-PIC" instead of keeping a package-local workaround in netcdf_fortran.
  - Use self["c"].pic_flag / self["fortran"].pic_flag instead of self.compiler.*_pic_flag.
  - Limit the argument-mismatch workarounds to older netcdf-fortran versions and add the NAG-specific mismatch flag there.
  - Move the libtool/NAG Darwin handling into the shared autotools libtool patching path instead of patching only this package.
  - Keep the preprocessor workaround NAG-only after discussion about whether other compilers need it.

- Probably resolved but not explicitly thread-closed in the data I fetched
  - The sys.platform vs spec-based platform check discussion with alalazo. The current PR still uses sys.platform, and the thread reads like a semantics discussion rather than an unresolved request to change it.
  - The “is this better generalized for case-insensitive filesystems?” discussion. Reviewer consensus appears to have landed on handling NAG only for now.

# PR Summary

This PR improves netcdf-fortran builds with the NAG compiler on macOS by:

- adding a shared libtool patch for Darwin/NAG detection in the autotools build system,
- fixing NAG PIC flag support in the NAG compiler package,
- refining netcdf_fortran compiler flag handling for older versions and NAG-specific preprocessing behavior.

# Review

History shows substantial discussion followed by an approval, so overall risk is medium-low: the changes are targeted, but they affect shared build-system/compiler behavior.

## Core Changes

### Shared autotools fix
autotools.py now patches libtool on Darwin to recognize NAG when hidden behind wrappers.
This is better than a package-local patch because it addresses the behavior where it belongs.

### Compiler definition fix
- nag/package.py adds pic_flag = "-PIC".
- This centralizes NAG PIC behavior in the compiler package.

### Package-specific compatibility logic
netcdf_fortran/package.py now:
- uses preferred PIC flag accessors,
- limits mismatch workarounds to @:4.5.2,
- adds -mismatch_all for older NAG builds,
- adds -fpp for NAG on Darwin.

## Other Changes
The main risk is concentrated in the shared autotools patch:
spack / spack-packages / files
```py
-            )
+            )
+
+            if sys.platform == "darwin":
+                # Libtool has a check for nagfor on darwin systems, but it does not
+                # recognize NAG behind the compiler wrappers, so fix that check:
```
This is clever and likely matches the reviewed intent, but it edits a generated shell case statement via a regex anchored to a specific comment block. That makes the fix somewhat sensitive to upstream libtool script shape changes.
## Merge Readiness and Risk Assessment

### Assessment:

likely ready to merge from a review-thread perspective, but I would want one more look at the shared autotools patch behavior because it affects all Darwin autotools packages using that path.

- Reviews: prior reviewer concerns appear addressed; one approval is present.
- Discussion: no clearly outstanding blocker.
- Risk: medium-low overall, but the autotools patch is the highest-risk part because it changes shared infrastructure.

# Possible Improvements

I found one substantive issue to raise:

Potential over-broad shared libtool patch condition
- The new Darwin libtool patch in autotools.py is gated only by sys.platform == "darwin", but the replacement itself depends on with_nag having been injected and on the targeted libtool block matching the expected structure.
- Because this is shared infrastructure, it would be safer if the code path were explicitly tied to the NAG case as much as possible in the surrounding logic, or at least documented as relying on the prior with_nag injection in this same patching flow.
- Why this matters: if future libtool variants or patch ordering differ, this may silently no-op or modify an unintended case block on Darwin packages unrelated to NAG.

# Bottom line
-  My review: I would leave one non-nit comment on the shared Darwin libtool patch robustness; otherwise the PR looks solid.

billsacks:
@bernhardkaindl - I'd like to defer to @skosukhin on your Copilot comments, since my changes just implemented @skosukhin 's suggestions.

skosukhin:
To be honest, I don't understand how these Copilot's comments deserve any answer :) It clearly does not get the context well enough. And it's such a wall of text. Anyway:

1. The way we patch libtool is no different from the way we apply other patches in `autotools.py` (right above that we add). Libtool is rather conservative, at least in this part of the script. If this or any other patch doesn't apply, the respective workaround will stop working, which will be reported, and we will introduce version-specific branching.
2. As for @alalazo's comment, he knows better, and I would apply his suggestion.

**UPD:**
Clarification on
> will stop working, which will be reported, and we will introduce

This is not a general remark; it's my opinion on this specific case. Libtool is supposed to provide portability out of the box, yet it fails in certain situations. Things may have changed since I last checked, but its release cycle is rather long. I genuinely wish anyone working to fix these issues upstream good luck, and I promise my support.

repos/spack_repo/builtin/packages/netcdf_fortran/package.py
bernhardkaindl: marking as resolved as per the last comments on 26 June.

repos/spack_repo/builtin/packages/netcdf_fortran/package.py
bernhardkaindl: Resolving a per the discussion on 26 June

repos/spack_repo/builtin/packages/netcdf_fortran/package.py
bernhardkaindl: Resolving a per the discussion on 26 June

repos/spack_repo/builtin/packages/netcdf_fortran/nag_linking.patch
bernhardkaindl: Resolving a per the discussion on 26 June

bernhardkaindl (approved):
@gartung / @alecbcs, see the last comment by @skosukhin:

- All suggested changes are applied. This now needs an approval by someone else than the last pusher.

billsacks:
@bernhardkaindl - thank you for applying the recent changes and helping to move this towards finalization!

bernhardkaindl:
@alecbcs, as I did the last push, this now needs a review from another maintainer, thanks!

alalazo (approved):

PR #3992, opened 2026-03-25
mpi: add versions to the mpi virtual package

alalazo:
This is just metadata that was missing. Might help with:
- https://github.com/spack/spack/pull/52138 
 
to get a unique optimal solution from ASP (currently there are internal states that are not uniquely determined in the optimal solution, but they don't affect the concrete spec).


alalazo:
@spackbot fix style

tldahlgren (approved):
LGTM

PR #3997, opened 2026-03-25
samtools: rework ncurses link flags

snehring:
I was having issues compiling samtools with an external ncurses (distro provided ncurses lib and headers, redhat specifically). rhel (at least in 9) does not distribute their ncurses with a libtinfow, only a libtinfo, despite ncurses being built with wide support. So the hardcoded ```-ltinfow``` was causing problems.

After looking into how ncurses figures out its includes and libs, I think the most "correct" way would actually be ```spec["ncurses:wide"].libs.link_flags```, but that won't grab the tinfo flag on el9 as there is no libtinfow.so. If you do ```spec["ncurses"].libs.link_flags``` you get link flags for both the w and non w versions of the libs (if present). This works, and results in samtools binary that is linked against both ncursesw and ncurses, along with tinfo and tinfow (if present). This doesn't appear to be a problem though, tview seemed to work when I did a quick test with both the external ncurses and a spack compiled ncurses.

my preference would be to use this as it wouldn't require any modifications for a user to use an external ncurses (on rhel 9 at the least)

maintainer: @jbeal-work

### DIFF
--- a/repos/spack_repo/builtin/packages/samtools/package.py
+++ b/repos/spack_repo/builtin/packages/samtools/package.py
@@ -78,26 +78,23 @@ class Samtools(Package):
     depends_on("htslib@1.3.1", when="@1.3.1")
 
     def install(self, spec, prefix):
-        if "+termlib" in spec["ncurses"]:
-            curses_lib = "-lncursesw -ltinfow"
-        else:
-            curses_lib = "-lncursesw"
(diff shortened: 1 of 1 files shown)


snehring:
@tldahlgren I don't think we're going to hear back from the maintainer on this one

tldahlgren (approved):

PR #4000, opened 2026-03-25
metis/parmetis: add petsc patches

Sbozzolo:
The METIS and ParMETIS packages are in need of improvements.

The original source codes are no longer available at the url in the package files and can only be retrieved from the Spack mirrors.

The packages had seen very little development over time, so PETSc forked them and added a few patches starting from the offiical version 5.1.0 and 4.0.3 (released in 2013):

https://bitbucket.org/petsc/pkg-metis
https://bitbucket.org/petsc/pkg-parmetis

The patches fix various bugs and result in a better mesh partitioning.

Recently (in the grand scheme of things), an official GitHub repo was added:
https://github.com/KarypisLab/METIS
https://github.com/KarypisLab/ParMETIS

The repos have seen very little development and patches, but they have already partially diverged from the PETSc fork. In particular, they now require GKLib to be installed separately.

I don't know what is the best way to handle these packages in Spack, but I think the PETSc patches are very valuable. So, I added them as a variant.


Sbozzolo:
@spackbot lint

Sbozzolo:
ping @mthcrts

mthcrts (approved):
Using the PETSc patches by default may break some other projects that depend on it.
But this should probably be recommended.

I do not currently use spack intensively and I was added as maintainer just because I fixed some things few years ago.
You should set yourself as maintainer in the metis packages.

Mathieu

hainest:
It's a little strange to have a variant just for determining when to apply a patch, particularly when that variant is pinned to a specific version. Is there a strong motivation for doing it this way? If not, I'd recommend using [`@run_before("patch")`](https://spack.readthedocs.io/en/latest/build_systems/custompackage.html#run-before-run-after) to conditionally add `patch` directives.

Sbozzolo:
> It's a little strange to have a variant just for determining when to apply a patch, particularly when that variant is pinned to a specific version. Is there a strong motivation for doing it this way? If not, I'd recommend using [`@run_before("patch")`](https://spack.readthedocs.io/en/latest/build_systems/custompackage.html#run-before-run-after) to conditionally add `patch` directives.

The PETCs patches are essentially an incompatible fork of the original packages. I don't know how that should be handled. 
Until recently, they were de facto the only maintained version of metis/parmetis, but the packages have started seeing some development from the original author. I don't think the patches should be applied without opt-in because they change the behavior of existing code and because it is debatable what is the "correct" version of metis/parmetis.

Could you clarify how I would use `@run_before` in this context?

tgamblin (changes requested):
Same question here as on #4447 -- can you host the patches somewhere or pull them from a commit .patch URL?

Sbozzolo:
> Same question here as on #4447 -- can you host the patches somewhere or pull them from a commit .patch URL?

What is the preferred place where to host patches? The patches cannot be derived from a commit .patch URL

Sbozzolo:
ping @tgamblin

Sbozzolo:
I found a way to do this directly using the commits from bitbucket.

Sbozzolo:
friendly ping @tgamblin

tgamblin (approved):

tgamblin:
@Sbozzolo thanks for adding this. Can you submit a separate PR to add yourself as a maintainer if you're interested?  Just need to change the `maintainers()` line, and you can probably cite @mthcrts's comment above.

I think adding these as `default=False` is good for now. I don't have a good understanding of when dependents might break.

Sbozzolo:
> Can you submit a separate PR to add yourself as a maintainer if you're interested?

Thank you, I don't really know much about these packages. They are dependencies of the code I am developing, but I don't I should be maintaining them

PR #3856, opened 2026-03-16
blast-plus: Fix typo in sqlite3 configure option

nboelte:
Checked with source tarballs downloaded from https://ftp.ncbi.nlm.nih.gov/blast/executables/blast+/
```
~/Downloads/ncbi-blast-2.15.0+-src/c++
13:38 $ ./configure --help|grep sqlite
 --with-sqlite3=DIR      use SQLite 3.x installation in DIR
 --without-sqlite3       do not use SQLite 3.x
~/Downloads/ncbi-blast-2.16.0+-src/c++
13:38 $ ./configure --help|grep sqlite
 --with-sqlite3=DIR      use SQLite 3.x installation in DIR
 --without-sqlite3       do not use SQLite 3.x
~/Downloads/ncbi-blast-2.17.0+-src/c++
13:31 $ ./configure --help|grep sqlite           
 --with-sqlite3=DIR      use SQLite 3.x installation in DIR
 --without-sqlite3       do not use SQLite 3.x
```

PR #4007, opened 2026-03-26
blast-plus: fix sqlite configure, rework zstd/sqlite reqs

snehring:
#2722 added zstd and sqlite deps, the added sqlite configure option was missing a character and failing to compile.

I also think it makes more sense to add these in as variants (they were previously optional functionality) and just make them required at the relevant versions.

maintainer: @weijianwen


snehring:
Ah, probably should have looked first. Slight overlap with #3856 here.

becker33 (approved):

PR #4009, opened 2026-03-26
suite-sparse: add v7.12.2, fix +cuda on newer versions and minor blas cmake bug

snehring:
New version.

+cuda was broken for the past couple of versions, they removed the explicit gpu targets in 7.4, they just get made now if cuda is enabled.

There seems to be a minor bug in the 7.12 releases with blas detection with cmake, setting BLA_VENDOR=' ' works around it.


snehring:
Bit of a divergence between my repo contents and what I was test building before walking out the door last night... Should be good now

snehring:
@spackbot re-run pipeline

becker33 (approved):

PR #4010, opened 2026-03-27
libsvm: sanitize version notation

w8jcik:
Version notation in `libsvm` package is unconventional and confusing. It probably comes from the fact that the author is using unseparated version tags, for example `v323` for version 3.23. Nevertheless, even the author of `libsvm` is using normal version notation in other places. There is no need to use versions like `323` here.

Two packages that depend on `libsvm` do not specify version constraint, so there is no issue in changing it here.


RikkiButler20 (approved):
LGTM

PR #4011, opened 2026-03-27
py-heat: add version 1.7.0 and 1.8.0

ClaudiaComito:
This PR updates the py-heat recipe to include the latest releases (1.7.0 and 1.8.0) .

Key Changes:

- New Versions: Added 1.7.0 and 1.8.0 with updated SHA256 hashes.

- Dependency Updates:  Version 1.8.0 now requires Python 3.11+ and supports PyTorch up to 2.11.0.

- New Variants:  +zarr: Added support for Zarr IO.


JuanPedroGHM (approved):

repos/spack_repo/builtin/packages/py_heat/package.py
```diff
@@ -33,6 +37,7 @@ class PyHeat(PythonPackage):
     variant(
         "netcdf", default=False, description="Use the py-netcdf4 package needed for NetCDF support"
     )
+    variant("zarr", default=False, description="Use the py-zarr package for Zarr support")
```
wdconinc:
```suggestion
    variant("zarr", default=False, description="Use the py-zarr package for Zarr support", when="@1.6:")
```
https://github.com/helmholtz-analytics/heat/releases/tag/v1.6.0

wdconinc (commented):
There's a minimum version where zarr support was added. Otherwise looks good and maintainer approved.

wdconinc (approved):

PR #4015, opened 2026-03-27
Dev stacks: pin mypy to bootstrap requirement

johnwparent:
Pin mypy to `py-mypy@0.900: ^py-mypy-extensions@:1.0` in dev stacks


kwryankrattiger (approved):
LGTM

PR #4016, opened 2026-03-27
boost: add version restriction to variants

snehring:
The contract library was added in 1.67.

### DIFF
--- a/repos/spack_repo/builtin/packages/boost/package.py
+++ b/repos/spack_repo/builtin/packages/boost/package.py
@@ -176,7 +176,7 @@ class Boost(Package):
         "fiber": {"when": "@1.62:"},
         "type_erasure": {"when": "@1.60:"},
         "atomic": {"when": "@1.53:"},
-        "coroutine": {"when": "@1.53:"},
+        "coroutine": {"when": "@1.54:1.66"},
         "context": {"when": "@1.51:"},
         "container": {"when": "@1.56:"},
         "locale": {"when": "@1.48:"},
--- a/repos/spack_repo/builtin/packages/boost/package.py
+++ b/repos/spack_repo/builtin/packages/boost/package.py
@@ -168,6 +168,7 @@ class Boost(Package):
         "charconv": {"when": "@1.85.0:"},
         "cobalt": {"when": "@1.84.0:"},
         "signals": {"when": "@:1.68"},
+        "contract": {"when": "@1.67:"},
         "signals2": {"when": "@1.4:"},
     }
 
--- a/repos/spack_repo/builtin/packages/boost/package.py
+++ b/repos/spack_repo/builtin/packages/boost/package.py
@@ -174,14 +174,14 @@ class Boost(Package):
         "contract": {"when": "@1.67:"},
         "stacktrace": {"when": "@1.65:"},
         "fiber": {"when": "@1.62:"},
-        "type_erasure": {"when": "@1.54:"},
+        "type_erasure": {"when": "@1.60:"},
         "atomic": {"when": "@1.53:"},
         "coroutine": {"when": "@1.53:"},
         "context": {"when": "@1.51:"},
-        "container": {"when": "@1.48:"},
+        "container": {"when": "@1.56:"},
         "locale": {"when": "@1.48:"},
         "chrono": {"when": "@1.47:"},
-        "signals2": {"when": "@1.39:"},
+        "signals2": {"when": "@1.87:"},
     }
 
     for lib in all_libs:
--- a/repos/spack_repo/builtin/packages/boost/package.py
+++ b/repos/spack_repo/builtin/packages/boost/package.py
@@ -163,13 +163,25 @@ class Boost(Package):
     # signals library was removed from boost in 1.69
     # https://www.boost.org/releases/1.69.0/#:~:text=Discontinued
     all_libs_opts = {
-        "openmethod": {"when": "@1.90.0:"},
-        "conversion": {"when": "@1.87.0:"},
-        "charconv": {"when": "@1.85.0:"},
-        "cobalt": {"when": "@1.84.0:"},
+        "openmethod": {"when": "@1.90:"},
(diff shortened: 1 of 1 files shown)


snehring:
maintainer: @hainest

hainest (approved):

snehring:
I actually hit another one so I decided to just audit the whole list if you're comfortable with that being tossed in this PR. I should have it in a bit.

hainest:
> I actually hit another one so I decided to just audit the whole list if you're comfortable with that being tossed in this PR. I should have it in a bit.

Surely. Let's update the title before merging, though.

snehring:
I didn't add anything for libraries that have been present before 1.39. Versions were pulled by going to ```https://www.boost.org/library/1.90.0/<name>``` and getting the version it says it was added.

snehring:
I did change signals2 to 1.39 and removed conversion (added in 1.16.1) per docs.

hainest:
> I didn't add anything for libraries that have been present before 1.39.

That's ok. The only versions earlier than 1.50 that I could get to build were 1.43 and 1.44 (https://github.com/hainest/spack-packages/commit/dbb0cd1f082a9b6f34641e4796a26b865150e278). Definitely let me know if you have been successful at that.

> Versions were pulled by going to `https://www.boost.org/library/1.90.0/<name>` and getting the version it says it was added.

Oh, if only that were the ground truth... If you need additional ones, I've gone through the massive headache of finding them all (https://github.com/hainest/spack-packages/blob/boost_redesign/repos/spack_repo/builtin/packages/boost/boostorg/variants.py). As soon as I figure out how some updates to the concretizer broke my variants, I'll get that PR posted. Then we won't ever have to worry about version constraints for old libraries ever again.

>I did change signals2 to 1.39 

It's only been buildable since 1.87, but that's a nit we don't need to worry about here.

>removed conversion (added in 1.16.1) per docs.

Indeed. `conversion` is actually header-only. I messed up when I approved it as a variant (https://github.com/hainest/spack-packages/commit/4c21129bec549099096bfec17685d476b151b9df).

repos/spack_repo/builtin/packages/boost/package.py
```diff
@@ -163,12 +163,25 @@ class Boost(Package):
...
+        "cobalt": {"when": "@1.84"},
+        "url": {"when": "@1.81:"},
+        "json": {"when": "@1.75:"},
+        "nowide": {"when": "@1.73:"},
         "signals": {"when": "@:1.68"},
-        "signals2": {"when": "@1.4:"},
+        "contract": {"when": "@1.67:"},
+        "stacktrace": {"when": "@1.65:"},
+        "fiber": {"when": "@1.62:"},
+        "type_erasure": {"when": "@1.54:"},
```
hainest: type_erasure was introduced in 1.54, but was header-only until 1.60. Will `boost@1.54+type_erasure` concretize here?

repos/spack_repo/builtin/packages/boost/package.py
```diff
@@ -163,12 +163,25 @@ class Boost(Package):
...
+        "atomic": {"when": "@1.53:"},
+        "coroutine": {"when": "@1.53:"},
```
hainest: coroutine is deprecated. I'd recommend leaving it off this list. coroutine2 was introduced in 1.59, but it's always been header-only.

repos/spack_repo/builtin/packages/boost/package.py
```diff
@@ -163,12 +163,25 @@ class Boost(Package):
...
+        "context": {"when": "@1.51:"},
+        "container": {"when": "@1.48:"},
```
hainest: container is only buildable since 1.56.

repos/spack_repo/builtin/packages/boost/package.py
```diff
@@ -163,12 +163,25 @@ class Boost(Package):
...
+        "locale": {"when": "@1.48:"},
+        "chrono": {"when": "@1.47:"},
+        "signals2": {"when": "@1.39:"},
```
hainest: signals2 is only buildable since 1.87.

hainest (commented):
The current model of what gets passed to `--with-libraries` isn't quite right. Currently, everything in `all_libs` gets passed to `--with-libraries` (less the filtering done in `install`), but that's only valid for versions of Boost where those libraries are buildable. The versions in `all_libs_opts` should reflect when the library is buildable, not when it is available. However, those versions are used to constrain the variant which _should_ reflect when its available. Nothing is quite right here. I'm trying to get a fix for all of this, but I keep breaking the concretizer...

For now, it's probably safe to add things to `all_libs_opts` using the version when they were added. I've left a note for each one where that version is different from the versions where it's valid to pass it to `--with-libraries` (i.e., it's "buildable").

repos/spack_repo/builtin/packages/boost/package.py
snehring: It does concretize, but does not install correctly, complaining about type_erasure being a wrong library name. I've changed this to 1.60.

repos/spack_repo/builtin/packages/boost/package.py
snehring: Changed.

repos/spack_repo/builtin/packages/boost/package.py
snehring: Changed.

repos/spack_repo/builtin/packages/boost/package.py
snehring: I think it'd be good to bound it. It doesn't seem to build until 1.54, and was deprecated in 1.67. Would you be okay with those as bounds?

hainest: That works for me.

hainest (approved):
LGTM. Thanks so much for taking the time to do this while I toil away on getting the conditional variants working!

wdconinc (approved):
Maintainer approved. Thanks!

PR #4018, opened 2026-03-27
py-pip: add v26.0.1

manuelakuhn:

### DIFF
--- a/repos/spack_repo/builtin/packages/py_arcgis/package.py
+++ b/repos/spack_repo/builtin/packages/py_arcgis/package.py
@@ -33,5 +33,10 @@ class PyArcgis(PythonPackage):
     depends_on("py-requests-toolbelt", type=("build", "run"))
     depends_on("py-requests-ntlm", type=("build", "run"))
 
+    @when("^py-pip@23.1:")
+    def config_settings(self, spec, prefix):
+        return {"--global-option": "--conda-install-mode"}
+
+    @when("^py-pip@:23.0")
     def global_options(self, spec, prefix):
         return ["--conda-install-mode"]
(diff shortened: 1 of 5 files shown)


adamjstewart (dismissed):
FYI, I might rename this package to `pip`. There is already a PR to rename ruff/ty/uv without the `py-`.

adamjstewart:
@spackbot run pipeline

manuelakuhn:
The pipeline fails because `--global-options` was removed now (it was deprecated since 23.1, see https://github.com/pypa/pip/issues/11859). It's an easy fix by switching over to `--config-settings`.
The packages I identified using `--global-options` are
- py_arcgis
- py_mumps4py
- py_pyomo
- py_pyyaml
- (py_torch_nvidia_apex is already fixed)

Since I need a `when(^py-pip@...)` statement for them: Do you want me to wait for your renaming, so that we use the new name here, or should I just fix them and you change the name later?

adamjstewart:
I don't think you can so easily switch. Feel free to do the fix or conflicts now.

manuelakuhn:
I have added the changes so that these packages build with `py-pip@26` now.

mathomp4 (approved):

adamjstewart:
@spackbot run pipeline

adamjstewart (approved):
Glad `--global-option` worked! I think I tried this with `--install-option` in the past and it didn't work.

manuelakuhn:
@spackbot run pipeline

manuelakuhn:
@spackbot run pipeline

manuelakuhn:
Thanks

PR #4023, opened 2026-03-28
Switch from black/flake8/isort to ruff

adamjstewart:
One-line change to `pyproject.toml` and `ruff check` now passes for the entire repo. No need for any fancy post-processing scripts @johnwparent. Long URLs/checksums are automatically ignored by ruff, and we already ignore all other offending rules. I was actually able to remove several unused ignores in the process.


adamjstewart:
Cross-linking to https://github.com/spack/spack/pull/52156 and https://github.com/spack/spack/pull/52161.

johnwparent:
Great! 
I could not get
`version("1.0", url="https://example.com/this-is-a-really-long-url/that-goes-over-99-characters/that-ruff-will-not-ignore-by-default",)` to pass ruff check, so I'm not sure what I was doing wrong ...

adamjstewart:
Ruff will reformat that line onto multiple lines, but it won't complain about the resulting long line.

johnwparent:
Huh... maybe I was testing check and format in isolation too much... glad I only spent  < an hour on that useless filter.

Thanks

repos/spack_repo/builtin/packages/g4pii/package.py
```diff
@@ -34,7 +34,7 @@ def setup_dependent_run_environment(
 
     def url_for_version(self, version):
         """Handle version string."""
-        return "https://geant4-data.web.cern.ch/geant4-data/datasets/G4PII.1.3.tar.gz" % version
+        return f"https://geant4-data.web.cern.ch/geant4-data/datasets/G4PII.{version}.tar.gz"
```
adamjstewart, line 37: This appears to be a real bug discovered by ruff and missed by black/flake8. The version string wasn't actually used in the URL.

johnwparent:
Is this compatible with the style check spackbot runs on the repo? If not, we either need to wait for my `--repo` PR to land, or re-write the script spackbot uses to do the repo style checks.

pyproject.toml
```diff
@@ -264,7 +265,7 @@ skip-magic-trailing-comma = false
 
 [tool.ruff.lint]
 extend-select = ["I"]
-ignore = ["E731", "E203", "F403", "F811"]
```
johnwparent, line 267: Why not ignore E203? IIRC it's labeled as unstable by the ruff docs
https://docs.astral.sh/ruff/rules/whitespace-before-punctuation/

pyproject.toml
```diff
@@ -280,10 +281,6 @@ section-order = [
 [tool.ruff.lint.isort.sections]
 spack = ["spack"]
 
-[tool.ruff.lint.per-file-ignores]
-"var/spack/*/package.py" = ["F403", "F405", "F811", "F821"]
```
johnwparent, line 284: Looks like we ignore `f405` from our flake8 config. I think removing here is the right call, just want to point that out.

pyproject.toml
```diff
@@ -264,7 +265,7 @@ skip-magic-trailing-comma = false
 
 [tool.ruff.lint]
 extend-select = ["I"]
```
johnwparent: Do we want to expand this more to include more of what core spack includes? 
Core spack is E F W I, and I think that makes sense here as well? I think at least we want F and E

pyproject.toml
johnwparent, line 20:
```suggestion
target-version = "py37"
```

pyproject.toml
```diff
@@ -264,7 +265,7 @@ skip-magic-trailing-comma = false
```
johnwparent, line 264: Let's enforce LF line endings, no crlf on windows. 

```suggestion
line-ending = "lf"
```

johnwparent (commented):
Only reviewing ruff config since I assume everything else is just the results of `ruff format`

Looks good overall, mostly just some suggestions for things to add.

pyproject.toml
adamjstewart: Preview rules are not enabled anyway, so we can't ignore this because there is nothing to ignore.

pyproject.toml
adamjstewart: select defaults to `["E4", "E7", "E9", "F"]`, so we are adding `"I"` to this list.

pyproject.toml
adamjstewart: According to https://docs.astral.sh/ruff/settings/#target-version:

> If you're already using a `pyproject.toml` file, we recommend `project.requires-python` instead, as it's based on Python packaging standards, and will be respected by other tools.

I realize that 3.6 is not supported by ruff, and ideally I would also like to not support it in Spack. Did you notice the tool make any mistakes because 3.6 is unknown?

pyproject.toml
adamjstewart: You're the expert here, but I'm curious why we don't want native?

pyproject.toml
adamjstewart: I just removed everything that isn't currently used, no need to ignore something that isn't an issue.

pyproject.toml
johnwparent: I think this is technically redundant since Git will enforce this for us, but avoiding crlf is something we want because a. there are almost no pieces of software/code that do not understand lf, there are many such cases will choke on `crlf` so in general its nicer and more robust to use just lf, and b. consistency and expectations are both nice to have, and very important for the  spec hashes. 
In general there's almost never a good reason to use `crlf` unless you're dealing with notepad or WIX or some other legacy DOS holdover. 

There have been a few cases where crlf snuck into the spack repo and broke things in hard to find ways. We want to be aware if that's going to happen (and have our linter auto-fix).

pyproject.toml
johnwparent: I came across this for the core repo, it doesn't recognize 3.6 and fell back to 3.10 (or whatever the default is, I recall 3.10)

pyproject.toml
johnwparent: We don't want the rest of the E category or pep-8 warnings?

repos/spack_repo/builtin/packages/g4pii/package.py
johnwparent: Nice!

pyproject.toml
adamjstewart: Added.

pyproject.toml
adamjstewart: I don't see this behavior:
```console
> ruff format --verbose
[2026-04-22][11:49:02][ruff::resolve][DEBUG] Using configuration file (via parent) at: /Users/Adam/spack-packages/pyproject.toml
[2026-04-22][11:49:02][ruff_workspace::pyproject][DEBUG] Detected minimum supported `requires-python` version: 3.6
```

mabraham (approved):

mtaillefumier (commented):
dbcsr is cleared

tpadioleau (approved):

hyoklee (approved):

adamjstewart:
> Is this compatible with the style check spackbot runs on the repo? If not, we either need to wait for my `--repo` PR to land, or re-write the script spackbot uses to do the repo style checks.

This I don't know. We need to update both the check in CI and the auto-formatting by spackbot (both should hopefully just use ruff directly)

MakisH (approved):

kwryankrattiger:
This also should update the `.ci/sytle_check.sh` command. This will get the CI to pass as well as update spackbot to use Ruff going forward.

adamjstewart:
Only last remaining place that still refers to black/flake8/isort is the bootstrap test which uses `spack style -b -t black`. It seems we are not yet running on the latest spack core so can't use `spack style -b -t ruff-format`. Not sure how/when to sync these up.

junghans (dismissed):

hppritcha (approved):

hyoklee (approved):

johnwparent (approved):
LGTM, we can sync the bootstrap job when we sync core and packages.

.ci/style_check.sh
```diff
@@ -24,11 +28,9 @@ if ! python_files > /dev/null; then
 fi
 [ -d "spack-core" ] ||  die "no 'spack-core' dir found: should be a clone of 'spack/spack'"
```
kwryankrattiger, line 29: This is no longer needed without flake8. Please remove and the drop the requirement in CI jobs.

kwryankrattiger (changes requested):

.ci/style_check.sh
```diff
@@ -24,11 +28,9 @@ if ! python_files > /dev/null; then
...
 python_files | xargs -0 printf "%s\n"
-info "running flake8"
-python_files | xargs -0 -n 100 flake8 || error=1
-info "running isort"
-python_files | xargs -0 -n 100 isort $flags || error=1
-info "running black"
-python_files | xargs -0 -n 100 black --color $flags || error=1
+info "running ruff format"
```
kwryankrattiger, line 31: These need to filter by changed files only

adamjstewart: Why?

kwryankrattiger: packages has a higher velocity and it is _possible_ that merging will create style issues unrelated to what other people are working on. This will lead to contributors potentially having to, across multiple PRs, deal with style errors that have nothing to do with their changes. This is unnecessary friction imo. For the cases where this kind of issue will arise, there is likely enough active development that will trigger a style fix for those cases shortly after.

kwryankrattiger: For the other case, where up update the ruff version and that changes things across the project, there will likely be a corresponding batch change.

For the spackbot, there is going to be a short lag (for now) where the spackbot version of ruff will not match the packages/core version. Without this running `spackbot fix style` could reformat the entire repo the wrong way.

lrknox:
Hi John,

I'm sending this to you because I have your email address, and I assume the "johnwparent" Vtk maintainer is you.  I have a spack PR #4822 for the hdf5 package that adds versions 2.0.0 and 2.1.1 (someone else added version 2.1.0) and also makes 2.1.1 the preferred version.  The vtk package build in spack ci testing fails with hdf5 2.1.1 because it uses the value of HDF5_ENABLE_PARALLEL in line 509 of VTK-9.5.2/CMake/patches/99/FindHDF5.cmake, but the variable name was changed from HDF5_ENABLE_PARALLEL to HDF5_PROVIDES_PARALLEL, consequently HDF5_ENABLE_PARALLEL and HDF5_IS_PARALLEL in turn are undefined, causing this error when configuring vtk:

-- Found HDF5: hdf5-shared (found version "2.1.1") found components: C HL
CMake Error at CMake/vtkModule.cmake:6153 (message):
  The variable `HDF5_IS_PARALLEL` was expected to have been available, but
  was not defined.

I'm not sure whether the solution, I assume to use ${HDF5_PROVIDES_PARALLEL} if the hdf5 version is >= 2.0.0, else use ${HDF5_ENABLE_PARALLEL} belongs in the VTK-9.5.2/CMake/patches/99/FindHDF5.cmake file, or in either the spack vtk package directory or in the spack hdf5 package directory.

I also am not sure if the change is in a spack package.py file or a patch file not in the spack hdf5 package, whether that change should be added to PR #4822 to make the vtk test (and I think silo as well) pass, or elsewhere.  Please advise, or forward to the appropriate person.

Thanks! and have a good weekend!
Larry

PR #4026, opened 2026-03-29
py-installer: add v1.0.0

adamjstewart:
https://github.com/pypa/installer/releases/tag/1.0.0

### DIFF
--- a/repos/spack_repo/builtin/packages/py_installer/package.py
+++ b/repos/spack_repo/builtin/packages/py_installer/package.py
@@ -19,10 +19,14 @@ class PyInstaller(Package, PythonExtension):
     )
     list_url = "https://pypi.org/simple/installer/"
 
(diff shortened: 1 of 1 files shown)


wdconinc (approved):
LGTM.

Curious if you think that there is any benefit to a py-installer-bootstrap and then a non-wheel py-installer package. Similar to go-bootstrap and go.

adamjstewart:
Nah, it's all pure Python anyway, no speed boost from bootstrapping and compiling ourselves. Only benefit would be if we wanted to patch or develop the source code. But no one has asked for that feature yet, so let's not complicate things.

wdconinc:
> Nah, it's all pure Python anyway, no speed boost from bootstrapping and compiling ourselves. Only benefit would be if we wanted to patch or develop the source code. But no one has asked for that feature yet, so let's not complicate things.

Thanks. That confirms my thinking.

PR #4031, opened 2026-03-30
py_pyspod: New package

LydDeb:

### DIFF
--- a/repos/spack_repo/builtin/packages/py_pyspod/package.py
+++ b/repos/spack_repo/builtin/packages/py_pyspod/package.py
@@ -24,9 +24,6 @@ class PyPyspod(PythonPackage):
 
     with default_args(type="build"):
         depends_on("py-setuptools@42:")
-        depends_on("py-hatchling")
-        depends_on("py-flit-core")
-        depends_on("py-poetry-core")
 
     with default_args(type=("build", "run")):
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/py_pyspod/package.py
```diff
@@ -0,0 +1,41 @@
...
+
+    version("2.0.0", sha256="b27b23e35b7fb3a2672d7707951e1756ae7d60aa196aa03aa095225bec9df5a4")
+
+    depends_on("python@3.7:", type=("build", "run"))
+
+    with default_args(type="build"):
+        depends_on("py-setuptools@42:")
+        depends_on("py-hatchling")
+        depends_on("py-flit-core")
+        depends_on("py-poetry-core")
```
manuelakuhn: `py-hatchling`, `py-flit-core` and `py-poetry-core` are not needed to build the package. `py-pyspod` only has `py-setuptools` as a build dependency.

LydDeb: My apologies, I committed too quickly. I've removed the unnecessary dependencies.
Thank you for your review.

manuelakuhn (approved):
LGTM and builds fine on rocky 8 with gcc@15

PR #4034, opened 2026-03-30
pyrefly: add new package

adamjstewart:
@johnwparent this is in the exact same boat as #3962, a Python type checking tool written in Rust. The developers suggest using `pip install` for installation, and maturin adds the right plumbing to enable `python -m pyrefly`. I kept the name change but haven't yet switched to CargoPackage. I'll wait for a final decision in #3962 before switching to cargo.

### DIFF
--- a/repos/spack_repo/builtin/packages/pyrefly/package.py
+++ b/repos/spack_repo/builtin/packages/pyrefly/package.py
@@ -1,0 +1,20 @@
+# Copyright Spack Project Developers. See COPYRIGHT file for details.
+#
+# SPDX-License-Identifier: (Apache-2.0 OR MIT)
+
+from spack_repo.builtin.build_systems.python import PythonPackage
+
+from spack.package import *
+
+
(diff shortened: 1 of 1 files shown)


johnwparent:
Previously I wanted this rust+python modeling resolved before pushing the ruff switch over, but since we're still relying on pytest and mypy, there's no real gain from pushing the CargoPackage switch so immediately (although I do want something in for this for 1.2), which is to say, we have time to think about this. 

If you're ok with an eventual refactor, I don't think this should be help up on  https://github.com/spack/spack-packages/pull/3962 

My Current Plan is to make a new package base for this, that is less tightly coupled to Python, should have a rough pass tonight or tomorrow.

johnwparent (approved):
I don't want to hold up new packages folks need on a potential update to how we do rust+python modeling, approving for now, but noting this package as a candidate for the initial pass of refactors along with ruff, ty, uv, etc.

PR #4035, opened 2026-03-30
py_torch: set PYTHON_SIX_SOURCE_DIR to avoid NNPACK to download 'six'

LydDeb:
NNPACK thirdparties downloads "six" during the build if `PYTHON_SIX_SOURCE_DIR` is not set.
However, I haven't been able to set this environment variable from the Spack recipe.
The  `nnpack_six_path.patch` will add a line to the NNPACK `CMakeLists.txt`  to explicitly define `PYTHON_SIX_SOURCE_DIR` from the `PYTHON_SIX_SOURCE_DIR` environment variable.

These modifications are intended to achieve an air gapped installation using the 'py-six' dependencies of the Spack recipe.

### DIFF
--- a/repos/spack_repo/builtin/packages/py_torch/package.py
+++ b/repos/spack_repo/builtin/packages/py_torch/package.py
@@ -583,6 +583,9 @@ class PyTorch(PythonPackage, CudaPackage, ROCmPackage):
         working_dir="third_party/fbgemm",
     )
 
+    # Force to use PYTHON_SIX_SOURCE_DIR to avoid NNPACK thirdparties downloads "six"
+    patch("nnpack_six_path.patch", when="@2.10.0+nnpack")
+
     def patch(self):
         # https://github.com/pytorch/pytorch/issues/52208
         filter_file(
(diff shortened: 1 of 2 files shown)


repos/spack_repo/builtin/packages/py_torch/package.py
```diff
@@ -821,6 +824,14 @@ def enable_or_disable(variant, keyword="USE", var=None):
         else:
             env.set("BUILD_CUSTOM_PROTOBUF", "OFF")
 
+        # NNPACK thirdparties downloads "six" during the build if PYTHON_SIX_SOURCE_DIR is not set
+        if self.spec.satisfies("+nnpack"):
+            python_dir = "python{0}".format(self.spec["python"].version.up_to(2))
+            six_site_packages = join_path(
+                self.spec["py-six"].prefix.lib, python_dir, "site-packages"
+            )
```
adamjstewart, line 832: This is heavily dependent on OS (lib vs. lib64 vs. Lib, site-packages vs. dist-packages), build (python vs. python3 vs. python3.14 vs. python3.14t), and a number of other factors. See [here](https://github.com/spack/spack-packages/blob/v2026.03.0/repos/spack_repo/builtin/packages/python/package.py#L1189) and [here](https://github.com/spack/spack-packages/blob/v2026.03.0/repos/spack_repo/builtin/packages/python/package.py#L1210) if you're curious.

I would suggest instead using:
```python
purelib = self.spec["python"].package.purelib
six_site_packages = os.path.join(self.spec["py-six"].prefix, purelib)
```

repos/spack_repo/builtin/packages/py_torch/package.py
```diff
@@ -583,6 +583,9 @@ class PyTorch(PythonPackage, CudaPackage, ROCmPackage):
         working_dir="third_party/fbgemm",
     )
 
+    # Force to use PYTHON_SIX_SOURCE_DIR to avoid NNPACK thirdparties downloads "six"
+    patch("nnpack_six_path.patch", when="@2.10.0+nnpack")
```
adamjstewart, line 587: Why only 2.10.0? It isn't needed in older or newer versions? Can we contribute this patch upstream if this is the right way to do things? I don't want to maintain this patch forever.

adamjstewart (changes requested):

PR #4047, opened 2026-03-30
print spec before doing rebuild

becker33:
Multiple users have requested this to make debugging failed builds easier.

PR spack/spack#49021 makes this default for `spack ci`, but we override defaults so we re-enable it here.


kwryankrattiger (approved):
Seems ok to me.

PR #4048, opened 2026-03-30
freecad: add version 1.0

jeffreyt:
Added versions 1.0.1 and 1.0.2 following format for `https://github.com/FreeCAD/FreeCAD/archive/refs/tags/x.x.x.tar.gz`

### DIFF
--- a/repos/spack_repo/builtin/packages/freecad/package.py
+++ b/repos/spack_repo/builtin/packages/freecad/package.py
@@ -15,13 +15,14 @@ class Freecad(CMakePackage):
 
     homepage = "https://www.freecad.org/"
     url = "https://github.com/FreeCAD/FreeCAD/archive/refs/tags/0.20.2.tar.gz"
+    git = "https://github.com/FreeCAD/FreeCAD"
 
     maintainers("aweits")
 
     license("LGPL-2.0-or-later")
 
-    version("1.0.2", sha256="228ee52f00627c7d8fa61998179deb01865ece69390829feb1300228d24f7e9e")
-    version("1.0.1", sha256="f62bc07c477544eff62b6ab0fc3bb63fa7f1e6f94763c51b0049507842d444f3")
+    version("1.0.2", tag="1.0.2", submodules=True)
+    version("1.0.1", tag="1.0.1", submodules=True)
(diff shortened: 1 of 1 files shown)


tldahlgren (dismissed):
Confirmed the version checksums.

@aweits Any issues with merging this PR as-is?

jeffreyt:
I'm sorry, I don't think this is quite right yet.  Some of the tar builds from the repo don't have submodules included.  I can to update to use the git repo directly for newer versions.  Hopefully that's okay?

tldahlgren:
> I'm sorry, I don't think this is quite right yet. Some of the tar builds from the repo don't have submodules included. I can to update to use the git repo directly for newer versions. Hopefully that's okay?

Sure.

jeffreyt:
@tldahlgren @alecbcs  Hi, I was wondering is there anyway to push this through?

repos/spack_repo/builtin/packages/freecad/package.py
```diff
@@ -15,11 +15,14 @@ class Freecad(CMakePackage):
...
     homepage = "https://www.freecad.org/"
     url = "https://github.com/FreeCAD/FreeCAD/archive/refs/tags/0.20.2.tar.gz"
+    git = "https://github.com/FreeCAD/FreeCAD"
 
     maintainers("aweits")
 
     license("LGPL-2.0-or-later")
 
+    version("1.0.2", tag="1.0.2", submodules=True)
+    version("1.0.1", tag="1.0.1", submodules=True)
```
alecbcs:
```suggestion
    version("1.0.2", commit="256fc7eff3379911ab5daf88e10182c509aa8052", submodules=True)
    version("1.0.1", commit="878f0b8c9c72c6f215833a99f2762bc3a3cf2abd", submodules=True)
```

@jeffreyt instead of pinning the versions to a git tag which can move, what if we instead pin them to the commit that is currently the head of each tag for reproducibility?

jeffreyt: Yes I think that should be fine as long as the `submodules` label still works.  (Which looks like it does according to the docs) 👍

alecbcs (approved):
Looks good to me. Thanks @jeffreyt!

jeffreyt:
@tldahlgren @alecbcs Thank you so much for your help!

PR #4055, opened 2026-03-31
py-bidskit: add v2025.11.7

manuelakuhn:
https://github.com/jmtyszka/bidskit/releases/tag/v2025.11.7

### DIFF
--- a/repos/spack_repo/builtin/packages/py_bidskit/package.py
+++ b/repos/spack_repo/builtin/packages/py_bidskit/package.py
@@ -15,6 +15,7 @@ class PyBidskit(PythonPackage):
 
     license("MIT")
 
+    version("2025.11.7", sha256="fc478bc5eb5b4808aa4ff2df2d5def54dbf8411313e16a2a2f639adbf0965a56")
     version("2025.1.30", sha256="91daa1041b07b029da2e720b91270be9ab39c42ccbb6e8d8bb302441ea1a1bf5")
     version("2023.9.7", sha256="029d9aecbbcb2df733858ceb3e6d5dd5013c36e431e40fb522a580adc7b667a5")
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/py_bidskit/package.py
```diff
@@ -25,18 +26,23 @@ class PyBidskit(PythonPackage):
...
+    with default_args(type="build"):
+        depends_on("py-poetry-core@2", when="@2025.11.7:")
 
-    depends_on("py-pydicom@2.3.1:", type=("build", "run"), when="@2025.1.30:")
-    depends_on("py-pydicom@2.2:", type=("build", "run"))
-    depends_on("py-numpy@1.26.3:", type=("build", "run"), when="@2025.1.30:")
-    depends_on("py-numpy@1.21:", type=("build", "run"))
+    with default_args(type=("build", "run")):
+        # ignore upper version limits, because it works with newer versions as well
+        depends_on("py-pydicom@2.3.1:", when="@2025.1.30:")
```
RikkiButler20, line 34:
```suggestion
        depends_on("py-pydicom@2.3.1:2.3", when="@2025.1.30:")
```
Does `pydicom` need an upper bound version constraint?

RikkiButler20 (changes requested):

repos/spack_repo/builtin/packages/py_bidskit/package.py
manuelakuhn: As stated in the comment, I could not find any reason why this upper bound version constraint is needed. It works fine without it (so far).

RikkiButler20: Alright then, thanks!

RikkiButler20 (approved):
Looks good.

PR #3443, opened 2026-02-16
r-ggplot2: add 3.5.2

Chrismarsh:
- add the last of the 3.x ggplot2 versions
- no req updates seemed to be required (CRAN no longer lists the 3.x so I based off of the github)

PR #4056, opened 2026-03-31
R ggplot2 4.0.2

snehring:
version bump with some deps and misc fixes.


repos/spack_repo/builtin/packages/r_gtable/package.py
Chrismarsh: I belive this should also have `depends_on("r@4:", type=("build", "run"), when="@0.3.6:")`

Chrismarsh:
I have this PR https://github.com/spack/spack-packages/pull/3444, which add gtables (same as here) needed for r-patchwork. However, patchwork needs `ggplot2@3.5.2:`  (which is the last of the stable 3.5.x line) in https://github.com/spack/spack-packages/pull/3443. However, https://github.com/spack/spack-packages/pull/3443 won't pass CI because the CI can't download the non-current CRAN package (because of how it handles URLs).

I wonder if we can sneak https://github.com/spack/spack-packages/pull/3443 in here as it the last of the 3.x line. There were no dep updates, just the package so it is the simplest review case. @RikkiButler20 would that work for you?

repos/spack_repo/builtin/packages/r_gtable/package.py
snehring: Yep, missed that.

snehring:
> I have this PR #3444, which add gtables (same as here) needed for r-patchwork. However, patchwork needs `ggplot2@3.5.2:` (which is the last of the stable 3.5.x line) in #3443. However, #3443 won't pass CI because the CI can't download the non-current CRAN package (because of how it handles URLs).
> 
> I wonder if we can sneak #3443 in here as it the last of the 3.x line. There were no dep updates, just the package so it is the simplest review case. @RikkiButler20 would that work for you?

I did throw ggplot2@3.5.2 in here when I saw that.

I'm going to rebase this to the current develop and see if that picks up the ci changes.

snehring:
Ah, #4060 hasn't been merged yet, my bad.

Chrismarsh:
perfect thanks, I'll close https://github.com/spack/spack-packages/pull/4056 in favour of this.

becker33 (approved):

PR #4058, opened 2026-03-31
prrte: update patch location

Chrismarsh:
The patch added in https://github.com/spack/spack-packages/pull/3961 uses the  `patch-diff.githubusercontent.com` subdomain. This subdomain doesn't seem to be consistently reachable from all machines (my HPC can't reach it). I've updated it to be consistent with the `commit.patch` style used elsewhere in spack


Chrismarsh:
cc @mcmehrtens

mcmehrtens:
@Chrismarsh LGTM. Sorry for breaking things... I wasn't sure the best way to encode the patch. I'll do it in this manner in the future! Thank you!

becker33 (approved):

PR #4060, opened 2026-03-31
RPackage: include archive URL for old versions in urls attribute

alecbcs:
This PR fixes a bug where we had previously been silently falling back to spidering urls for old versions of `r-*` packages since we weren't correctly modeling the archive URL. This PR swaps to setting the url**s** attribute on the class with both the archive url path and the latest version path.

### DIFF
--- a/repos/spack_repo/builtin/build_systems/r.py
+++ b/repos/spack_repo/builtin/build_systems/r.py
@@ -109,3 +109,6 @@ class RPackage(Package):
     url: ClassProperty[Optional[str]] = classproperty(_url)
     list_url: ClassProperty[Optional[str]] = classproperty(_list_url)
     git: ClassProperty[Optional[str]] = classproperty(_git)
+
+    def url_for_version(self, version):
+        return f"https://cran.r-project.org/src/contrib/Archive/{self.cran}/{self.cran}_{version}.tar.gz"
--- a/repos/spack_repo/builtin/build_systems/r.py
+++ b/repos/spack_repo/builtin/build_systems/r.py
@@ -113,3 +113,5 @@ class RPackage(Package):
     def url_for_version(self, version):
         if self.cran:
             return f"https://cran.r-project.org/src/contrib/Archive/{self.cran}/{self.cran}_{version}.tar.gz"
+
+        return super().url_for_version(version)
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/build_systems/r.py
```diff
@@ -109,3 +109,6 @@ class RPackage(Package):
     url: ClassProperty[Optional[str]] = classproperty(_url)
     list_url: ClassProperty[Optional[str]] = classproperty(_list_url)
     git: ClassProperty[Optional[str]] = classproperty(_git)
+
+    def url_for_version(self, version):
+        return f"https://cran.r-project.org/src/contrib/Archive/{self.cran}/{self.cran}_{version}.tar.gz"
```
wdconinc: Gate on hasattr cran?

alecbcs: Decided to gate on `if self.cran` since the attribute is set to `None` in the base class

becker33 (dismissed):

repos/spack_repo/builtin/build_systems/r.py
```diff
@@ -109,3 +109,7 @@ class RPackage(Package):
     url: ClassProperty[Optional[str]] = classproperty(_url)
     list_url: ClassProperty[Optional[str]] = classproperty(_list_url)
     git: ClassProperty[Optional[str]] = classproperty(_git)
+
+    def url_for_version(self, version):
+        if self.cran:
```
haampie: super in the else case?

haampie (changes requested):

repos/spack_repo/builtin/build_systems/r.py
alecbcs: Yep, good idea! Implemented in the latest commit

haampie (dismissed):

haampie:
Don't we support urls (plural) too? I don't get this part, but also haven't look at the implementation

> In the case of the latest release, we fail to find the url via this method and fall back to the package class URL which will be correct.

It sounds suboptimal even if it works

alecbcs:
I originally stayed away from using `urls` because a package cannot have both `url` and `urls` properties and there are some `r-*` packages that set a `url`, however since we gate populating the urls on the cran class attribute I think a package setting both a cran identifier and a custom url is reasonably expected to fail.

becker33 (approved):

PR #4066, opened 2026-04-01
crux-toolkit: add new package

w8jcik:
> suite of analysis tools for interpreting protein mass spectrometry data

https://crux.ms

### DIFF
--- a/repos/spack_repo/builtin/packages/crux_toolkit/package.py
+++ b/repos/spack_repo/builtin/packages/crux_toolkit/package.py
@@ -1,0 +1,48 @@
+# Copyright Spack Project Developers. See COPYRIGHT file for details.
+#
+# SPDX-License-Identifier: (Apache-2.0 OR MIT)
+
+from spack_repo.builtin.build_systems.cmake import CMakePackage
+
+from spack.package import *
+
+
+class CruxToolkit(CMakePackage):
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/crux_toolkit/package.py
```diff
@@ -0,0 +1,48 @@
...
+    homepage = "https://crux.ms"
+    url = "https://github.com/crux-toolkit/crux-toolkit/archive/refs/tags/crux-4.3.2.tar.gz"
+
+    license("Apache-2.0")
+
+    version("4.3.2", sha256="6911c1cb45e5a188036edcbfbaf9ab6d34679321988a88de1cd192f1e4a2630b")
+
+    depends_on("cmake@3.15:", type="build")
+
+    # Additional dependencies are downloaded during build (the logic cannot be disabled)
```
wdconinc, line 22: The risk in this context (as you likely know well) is that the vendored versions and the spack installed versions in the same spack environment may lead to runtime library inconsistencies. It seems the project already attempts to avoid this with static linking.

wdconinc (approved):
Looks good. Too bad about the vendoring but unlikely to change (https://github.com/crux-toolkit/crux-toolkit/issues/425).

PR #4069, opened 2026-04-01
openmpi: add restriction for libfabric when @:4.0

snehring:
Was running into compilation issues with openmpi@3 and libfabric with libfabric versions later than 1.x. Tested a couple other versions and it seems that got resolved in 4.1 and later.


hppritcha (approved):

rbberger (approved):

snehring:
@spackbot re-run pipeline

PR #4078, opened 2026-04-02
py-aiohttp: deprecate versions with CVEs

adamjstewart:
It's amazing that we didn't deprecate these sooner, many of these (especially the older ones) have known active exploits in the wild.


adamjstewart:
@spackbot fix style

RikkiButler20 (approved):
LGTM

PR #4079, opened 2026-04-02
py-requests: deprecate versions with CVEs

adamjstewart:
Many of these older CVEs have known active exploits in the wild.


adamjstewart:
@spackbot fix style

manuelakuhn (approved):
LGTM and builds fine on rocky 8 with gcc@15.
Thanks that are a lot of exploits...

PR #4090, opened 2026-04-02
`[will be resumed]` sundials: turn sundials into a CachedCMakePackage

balos1:
Synchronizing our repository internal Spack package, which is a CachedCMakePackage, and the upstream one here.

### DIFF
--- a/repos/spack_repo/builtin/packages/sundials/package.py
+++ b/repos/spack_repo/builtin/packages/sundials/package.py
@@ -3,7 +3,12 @@
 import os
 import sys
 
-from spack_repo.builtin.build_systems.cmake import CMakePackage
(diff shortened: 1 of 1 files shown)


bernhardkaindl:
Hi @balos1, as far as I can see, likely at least some of the review comments have merit.

Could you apply those and reply to those which don't fit your understanding?

bernhardkaindl:
@balos1, at least the first few finding of the review are valid blockers or bugs for a merge.

Do you have plans to review those review comments and fix them, where applicable?

balos1:
> @balos1, at least the first few finding of the review are valid blockers or bugs for a merge.
> 
> Do you have plans to review those review comments and fix them, where applicable?

Yes, I will return to this draft PR soon. I can close if that is preferred.

PR #4093, opened 2026-04-02
salmon: add v1.11.4

snehring:
Version bump with some not insignificant changes. 

v1.11 uses a newer backend for indexing and the README suggests that older index files would need to be regenerated, since I'm not really sure how exactly trinity uses salmon I've locked the version to the last 1.10 release out of an abundance of caution .

### DIFF
--- a/repos/spack_repo/builtin/packages/salmon/package.py
+++ b/repos/spack_repo/builtin/packages/salmon/package.py
@@ -174,7 +174,8 @@ class Salmon(CMakePackage):
             filter_file("PACKAGE_VERSION", "IOLIB_VERSION", findstadenio_module, string=True)
             filter_file("io_lib_config.h", "version.h", findstadenio_module, string=True)
 
-        # filter file to remove boost_system dep if boost @1.69: (or 1.89: when they removed the stub)
+        # filter file to remove boost_system dep if boost @1.69:
+        # (or 1.89: when they removed the stub)
         if self.spec.satisfies("@1.11.4:"):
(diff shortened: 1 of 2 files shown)


snehring:
@spackbot re-run pipeline

repos/spack_repo/builtin/packages/salmon/package.py
```diff
@@ -41,36 +42,62 @@ class Salmon(CMakePackage):
...
+    depends_on("catch2", type="test", when="@1.11:")
+    # probably just for tests, but cmake looks anyway
+    depends_on("python@3:", type="build", when="@1.11:")
 
     patch("fix_hts.patch", when="@1.10.2")
 
     conflicts("%gcc@:5.1", when="@0.14.1:")
 
+    # SalmonDependencies.camke
+    resource(
```
becker33, line 85: Is there a reason that this is modeled as a resource instead of being modeled as a depencency?

Same for fqfeeder

snehring: It's a weird set up.

In the case of pufferfish the install process directly refers to the source files, not the install artifacts. It also refers to the source of an external dependency of pufferfish that pufferfish fetches at build time.

FQfeeder doesn't have an install target and it's output is a static library, but like with pufferfish salmon uses the source files directly and not resulting libs.

becker33 (approved):
Thanks for explaining, I hate all of that but I don't see a better option than treating it as a resource for now.

PR #588, opened 2025-07-14
rocksdb: add v10.4.2

michaelkuhn:

PR #4094, opened 2026-04-02
papi: Add variants for disable cpu, disable perf_event, disable perf_event_uncore, and with sysdetect

Treece-Burgess:
This pull requests adds four new variants to the PAPI package:
```
disable_cpu
disable_perf_event
disable_perf_event_uncore
with_sysdetect
```

I sanity tested these changes on a machine with an OS of RHEL 8.10 and a CPU of AMD EPYC 7402 24-core
- `spack install papi +disable_cpu`: ✅ 
- `spack install papi +disable_perf_event`: ✅ 
- `spack install papi +disable_perf_event_uncore`: ✅ 
- `spack install papi with_sysdetect=no`: ✅ 
- `spack install papi with_sysdetect=yes`: ✅ 
- `spack install papi +infiniband with_sysdetect=no`: ✅

### DIFF
--- a/repos/spack_repo/builtin/packages/papi/package.py
+++ b/repos/spack_repo/builtin/packages/papi/package.py
@@ -71,6 +71,11 @@ class Papi(AutotoolsPackage, ROCmPackage):
     # The PAPI configure option "--with-shlib-tools" is deprecated
     # and therefore not implemented here
 
+    variant("disable_cpu", default=False, when="@7.2:", description="Disable the CPU components")
+    variant("disable_perf_event", default=False, when="@7.2:", description="Disable the perf_event component")
+    variant("disable_perf_event_uncore", default=False, when="@7.2:", description="Disable the perf_event_uncore component")
+    variant("with_sysdetect", default="none", when="@7.2:", values=("no", "yes"), multi=False, description="Build the sysdetect component, the default is yes")
+
     depends_on("c", type="build")
     depends_on("cxx", type="build")
     depends_on("fortran", type="build")
@@ -187,6 +192,18 @@ class Papi(AutotoolsPackage, ROCmPackage):
 
         if "+debug" in spec:
             options.append("--with-debug=yes")
+
+        if "+disable_cpu" in spec:
+            options.append("--disable-cpu")
+
+        if "+disable_perf_event" in spec:
+            options.append("--disable-perf-event")
+
+        if "+disable_perf_event_uncore" in spec:
+            options.append("--disable-perf-event-uncore")
+
+        if spec.variants["with_sysdetect"].value != "none":
+            options.append("--with-sysdetect={0}".format(spec.variants["with_sysdetect"].value))
 
         if self.run_tests:
             options.append("--with-tests=ctests")


repos/spack_repo/builtin/packages/papi/package.py
```diff
@@ -71,6 +71,11 @@ class Papi(AutotoolsPackage, ROCmPackage):
     # The PAPI configure option "--with-shlib-tools" is deprecated
     # and therefore not implemented here
 
+    variant("disable_cpu", default=False, when="@7.2:", description="Disable the CPU components")
+    variant("disable_perf_event", default=False, when="@7.2:", description="Disable the perf_event component")
+    variant("disable_perf_event_uncore", default=False, when="@7.2:", description="Disable the perf_event_uncore component")
+    variant("with_sysdetect", default="none", when="@7.2:", values=("no", "yes"), multi=False, description="Build the sysdetect component, the default is yes")
```
rbberger, line 77: If the default is `yes`, why have 3 states? Why not just have:
```
variant("sysdetect", default=True, when="@7.2:", ...)
```

repos/spack_repo/builtin/packages/papi/package.py
```diff
@@ -71,6 +71,11 @@ class Papi(AutotoolsPackage, ROCmPackage):
     # The PAPI configure option "--with-shlib-tools" is deprecated
     # and therefore not implemented here
 
+    variant("disable_cpu", default=False, when="@7.2:", description="Disable the CPU components")
+    variant("disable_perf_event", default=False, when="@7.2:", description="Disable the perf_event component")
+    variant("disable_perf_event_uncore", default=False, when="@7.2:", description="Disable the perf_event_uncore component")
```
rbberger, line 76: Instead of long negative variants, how about:
```python
variant("cpu", default=True, when="@7.2:", description="Enable the CPU components")
variant("perf_event", default=True, when="@7.2:", description="Enable the perf_event component")
variant("perf_event_uncore", default=True, when="@7.2:", description="Enable the perf_event_uncore component")
```
and then do this during configure:
```python
options.extend(self.enable_or_disable("cpu"))
options.extend(self.enable_or_disable("perf-event", variant="perf_event"))
options.extend(self.enable_or_disable("perf-event-uncore", variant="perf_event_uncore"))
```

rbberger (commented):
Added suggestions for improving the variant names.

G-Ragghianti:
The pattern that we have used in the papi spack package is mostly that +component means to enable the component, and that is implemented in the papi spack class in the list of components found in the "configure_args" method.  Also, variant names like "disable_*" invert the normal variant name logic where +variant means to enable something.  Instead of a variant with `name=yes`, I suggest `+name` if possible. The variant names suggested by rbberger look good.

G-Ragghianti:
I see that a complication with some papi components is that they default to "on" such that leaving them out of the `--with-components' list has no effect.  In that case (as with e.g. sysdetect), they can be handled like this:

```
options.extend(self.with_or_without("sysdetect"))
```

G-Ragghianti:
Some CI failures are due to code style.

bernhardkaindl (changes requested):
@Treece-Burgess See https://github.com/spack/spack-packages/pull/4094#pullrequestreview-4056751670 and code style fixes are needed (might be just ruff format to fix)

bernhardkaindl:
@Treece-Burgess, if you want to continue this PR, would be able to work on the review comments?
> See https://github.com/spack/spack-packages/pull/4094#pullrequestreview-4056751670 and code style fixes are needed (might be just ruff format to fix)

Treece-Burgess:
@bernhardkaindl I am going to close this PR for now. 

I have [PR #588](https://github.com/icl-utk-edu/papi/pull/588) in the PAPI repository which will change `--with-sysdetect` to follow `--disable` and `--enable` logic. Once the PR is merged, I will reopen and make the appropriate updates to this PR.

PR #4101, opened 2026-04-03
Package rename: py-pip -> pip

adamjstewart:
There has long been discussion of how to name Python tools. According to our [docs](https://spack.readthedocs.io/en/latest/build_systems/pythonpackage.html#choosing-a-package-name):

> When in doubt, think about how this package will be used. Is it primarily a Python library that will be imported in other Python scripts? Or is it a command-line tool, or C/C++/Fortran program with optional Python modules? The former should be prepended with `py-`, while the latter should not.

Pip is not intended to be imported, it is a command-line tool, and therefore should be called `pip`, not `py-pip` following these docs.

This has come up a few times already, first with pipx in https://github.com/spack/spack/pull/38658 (@blue42u @eli-schwartz) and again with ruff/ty/uv in #3962 (@johnwparent). @manuelakuhn may also have opinions on this discussion.

@eli-schwartz makes a good point in https://github.com/spack/spack/pull/38658#issuecomment-1616862234 that tools like pip are only useful within the Python ecosystem, and therefore a `py-` prefix makes sense. I would suggest we either follow the above documentation (I wrote it, it's not gospel) or change the above documentation and close this PR. But let's settle this once and for all.

If this is approved, I'll make a similar PR in https://github.com/spack/spack, there are a couple things (e.g., `spack create`) that refer to `py-pip`. If this is rejected, we should consider renaming tools like pipx, conda, pyrefly, etc. to add a `py-` prefix.

### DIFF
--- a/repos/spack_repo/builtin/build_systems/python.py
+++ b/repos/spack_repo/builtin/build_systems/python.py
@@ -242,7 +242,7 @@ class PythonPackage(PythonExtension):
 
     with when("build_system=python_pip"):
         extends("python")
-        depends_on("py-pip", type="build")
+        depends_on("pip", type="build")
         # FIXME: technically wheel is only needed when building from source, not when
         # installing a downloaded wheel, but I don't want to add wheel as a dep to every
         # package manually
@@ -353,7 +353,7 @@ class PythonPipBuilder(BuilderWithDefaults):
         """Extra global options to be supplied to the setup.py call before the install
         or bdist_wheel command.
 
-        Deprecated in pip 23.1.
+        Requires pip 25.2 or older.
 
         Args:
             spec: Build spec.
@@ -373,7 +373,7 @@ class PythonPipBuilder(BuilderWithDefaults):
         config_settings = self.config_settings(spec, prefix)
 
         # Pass -jN for compile-args if supported and needed
-        if spec.satisfies("%py-pip@22.1: %py-meson-python@0.11:"):
+        if spec.satisfies("%pip@22.1: %py-meson-python@0.11:"):
             # get_effective_jobs returns None when a jobserver is active, then we don't pass -j.
             jobs = get_effective_jobs(
                 jobs=determine_number_of_jobs(parallel=pkg.parallel), supports_jobserver=True
--- a/repos/spack_repo/builtin/packages/pip/package.py
+++ b/repos/spack_repo/builtin/packages/pip/package.py
@@ -10,11 +10,8 @@ from spack_repo.builtin.build_systems.python import PythonExtension, PythonPipBu
 from spack.package import *
 
 
-class PyPip(Package, PythonExtension):
-    """The PyPA recommended tool for installing Python packages.
-
-    Deprecated in favor of pip package.
-    """
+class Pip(Package, PythonExtension):
+    """The PyPA recommended tool for installing Python packages."""
 
     homepage = "https://pip.pypa.io/"
     url = "https://files.pythonhosted.org/packages/py3/p/pip/pip-20.2-py3-none-any.whl"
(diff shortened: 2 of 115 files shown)


repos/spack_repo/builtin/build_systems/python.py
```diff
@@ -353,7 +353,7 @@ def global_options(self, spec: Spec, prefix: Prefix) -> Iterable[str]:
         """Extra global options to be supplied to the setup.py call before the install
         or bdist_wheel command.
 
-        Deprecated in pip 23.1.
+        Requires pip 25.2 or older.
```
adamjstewart, line 356: @manuelakuhn updated these docs here to be more specific now that this feature has been removed.

repos/spack_repo/builtin/packages/pip/package.py
adamjstewart, line 1: The new package is just a copy-n-paste and rename. There are several older versions with CVEs that I'll deprecate in a follow-up PR. Didn't want to directly remove them here so as to avoid complicating the PR.

repos/spack_repo/builtin/packages/py_pip/package.py
adamjstewart, line 1: Only other file that isn't a simple sed replacement.

adamjstewart:
@spackbot fix style

johnwparent:
> that tools like pip are only useful within the Python ecosystem, and therefore a py- prefix makes sense.

The same is true of ruff/ty/uv. 

I'm of the overall opinion we should go with @tgamblin's longer term game plan, factor out all of these packages into a `py` namespace and let that resolve this vs deliberating about the current scheme and then going with the namespace change eventually anyway.

blue42u:
I generally agree with naming this package `pip`. The argument for `py-pip` mostly revolves around the idea that Pip is so tied to the Python ecosystem that it deserves to have `py-` in the name. I find that too subjective an argument, especially when I'm installing something to satisfy some command I saw on the web and don't really know what "ecosystem" the package is part of[^1]. IMHO the package name should match what it "provides," e.g. `ruff` should provide an executable called `ruff`, thus `pip` should provide an executable called `pip`. It just saves time and user headaches.

[^1]: To give a motivating example: the [install docs for Prettier](https://prettier.io/docs/install) say `npm install prettier`. I don't have `npm` installed, so I `spack install npm` and move on. I don't need to know that NPM is part of the "Javascript ecosystem," I don't need to say `spack install js-npm`, it *just works* and I can Prettier my YAML files just fine.

The only opinion I have stronger than `pip` is outright removing the package. IME running Pip in a Spack environment is riddled with caveats: tainted/corrupted Python installs, loss of installs on view regeneration, etc. etc. Rather than document/fix those caveats, the UX would be much better if `pip` was not even a package, and instead users were pushed towards the [official means to install Pip, `ensurepip`](https://pip.pypa.io/en/stable/installation/)[^2]. Even better if the Python install was marked as an external install ([PEP 668](https://peps.python.org/pep-0668/)) so users can never taint their Python by accident.

[^2]: Yes, including `PythonPackage`s, they should make a temporary `venv` + `ensurepip` and use Pip from there. Obviously there are edge cases if a package requires a particular Pip, but IMO those can be solved without globally exposing `pip`.

I have no strong opinions on a `py.*` namespace, IMHO `py.pip` is just unnecessary syntax sugar that no one is going to use in practice. But that decision need not block this change, presumably if the name is just `pip` Spack will magically resolve the namespace (https://github.com/spack/spack-packages/pull/3962#issuecomment-4187595466).

eli-schwartz:
@blue42u,

The idea that:
- every project name corresponds one-to-one to an executable name
- every project only installs a single executable

is two terrible(y myopic) ideas that not only don't I believe are representative of all software, I'm not even convinced it's representative of more than a superminority of software. And frankly, I wouldn't want to live in a world where it was true.

I believe I can reasonably be convinced that it's representative of Python software distributed on PyPI, although I think you're mixing up cause and effect. Python packaging standards are ill designed for many use cases, although it would be better described as "well designed for a select handful of other use cases" which doesn't include shipping programs at all. Even "cli entrypoints to a library" (which are not entirely the same thing) is pushing the outer limits of it.

None of this should be new here, because this gap between "mainstream PyPI users" and "the wider world of software development using python" is one of the core reasons why spack (and conda) exist.

...

Anyways it's all a totally moot point. You've got the tail wagging the dog here. You want to be able to ask your package manager to get you a program so you think that means the package needs to be named after the program. This is an understandable confusion if you're *solely* used to library managers like pip (and npm too), but the wider world of package management isn't limited to such crude hacks. You want to design for success. That means being able to support for example `Provides: pip = %{version}` times whatever or however many programs a package may install. Or `Provides: /usr/bin/git`. You can then ask the package manager to install a particular program file for you *and the package manager performs a reverse lookup and locates the correct package to install for you*. There is no reason a package manager cannot support this if it wishes to (pip can't but that is because it's not a package manager, it's a recursive *library* installer). Outside of that, every package manager with a global database of built packages is generally capable of providing tools like `apt-file` (Debian) `e-file` (Gentoo), `dnf whatprovides` (fedora), `pacman --files` (archlinux) etc, that provide rich search capabilities for the contents of uninstalled packages.

There may still be reasons to prefer particular naming scenes such as "pip" without a "py-" prefix. It is simply unimaginative and an evolutionary dead end to start making decisions like that "because we need to, so that people can install a program name and get the right thing".

(My preferred package manager uses namespaces already so I don't have this problem. The "is it a python tool or a program written in python" question is answered by calling pip "dev-python/pip" and calling say Meson -- written in python -- "dev-build/meson". I will note that ruff here is "dev-util/ruff", not "dev-python/ruff", which makes sense in a way because ruff just analyzes python files, whereas pip has to operate on a specific python3 interpreter.)

adamjstewart:
@blue42u @eli-schwartz thanks for the discussion. I talked to @tgamblin and we decided it would be best to rename all Python packages in a single PR instead of one package at a time. Basically, instead of prepending `py-` to all PyPI package names, we would create a new `py` namespace such that both `pip` and `py.pip` would refer to the same package. The idea is that for most Python packages with unique names, you wouldn't need to be specific, but for packages that have both Python and non-Python implementations, they would still have unique names.

P.S. @eli-schwartz while I agree with you, I think there are nicer and less antagonistic ways to phrase this. The point is that most packages don't install executables at all, and for the ones that do, there is no guarantee that the name is unique. Similarly, the import name is also not guaranteed to be unique (PIL, pillow, and pillow-simd are all `import PIL`). This is why Spack has always used and will always use the PyPI name (guaranteed to be unique), which matches pip/uv so there are the least surprises for users.

I'm going to close this PR, let's create a new `py` namespace and rename all `py-` packages in a separate PR.

eli-schwartz:
> P.S. @eli-schwartz while I agree with you, I think there are nicer and less antagonistic ways to phrase this.

Maybe. I did try and I think reasonably succeed at critiquing the idea without making things personal. I think ideas are fair game to be criticized and I have a lot of experience with various package managers and feel I can directly blame the popularity of this *idea* as an important reason why pip's design doesn't serve the needs of many of its users, which I think is an unfortunate outcome. And that's one of the reasons why I think it's important to emphasize the reasoning behind why I think something is a really bad idea. Nobody can design better systems if they don't understand the flaws in the current systems (or know how other systems differ). And I also believe that's why users of pip and npm often suggest to make other package managers more like pip and npm. ;)


> The point is that most packages don't install executables at all, and for the ones that do, there is no guarantee that the name is unique. Similarly, the import name is also not guaranteed to be unique (PIL, pillow, and pillow-simd are all `import PIL`)

Yes, all three of these are good distinctions. There are others though, such as the fact that you cannot rename https://packages.spack.io/package.html?name=coreutils into any of these exclusive alternative names

```
spack install sha256sum
spack install cp
spack install tr
spack install uname 
```

(But "provides" technology can support this, should it be seen as interesting.)

> This is why Spack has always used and will always use the PyPI name (guaranteed to be unique), which matches pip/uv so there are the least surprises for users.

Yup, in combination with namespacing this is quite powerful. Gentoo has the same convention (`dev-python/${pypiname}`)

That means for example:
- "dev-python/sphinx" or simply "sphinx" provides `sphinx-build`
- "dev-python/pygments" or simply "pygments" provides `pygmentize`
- "dev-python/mako" or simply "mako"... actually no, gui-apps/mako exists, https://github.com/emersion/mako is a notification daemon for wayland. So you need "dev-python/mako" explicitly. But it does provide `mako-render`
- "dev-python/build" provides `pyproject-build`

PR #4117, opened 2026-04-05
py-numpy: update license

adamjstewart:
License was clarified in https://github.com/numpy/numpy/pull/29535

@rgommers do you think we should apply this retroactively to all versions? Like you, I didn't include sdist licenses, only bdist licenses assuming no vendoring.


rgommers:
Good catch. Yes, I think we should apply this to all versions. Almost all that code has been in `numpy` for a very long time, `numpy` was never BSD3-only.

rgommers (approved):
LGTM

RikkiButler20 (approved):

PR #4123, opened 2026-04-06
pcre2: add v10.47

jmcarcell:
Edit: it seems there are unrelated CI failures

### DIFF
--- a/repos/spack_repo/builtin/packages/pcre2/package.py
+++ b/repos/spack_repo/builtin/packages/pcre2/package.py
@@ -19,7 +19,11 @@ class Pcre2(AutotoolsPackage, CMakePackage):
 
     license("BSD-3-Clause AND PCRE2-exception", when="@10:", checked_by="wdconinc")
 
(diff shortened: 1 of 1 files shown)


wdconinc:
@spackbot run pipeline

wdconinc (changes requested):
https://github.com/PCRE2Project/pcre2/compare/pcre2-10.44...pcre2-10.47

CMake version required is now 3.15

repos/spack_repo/builtin/packages/pcre2/package.py
wdconinc, line 39: The jit tree seems to have been moved into a submodule in 10.45. Please verify that the source tarball linked still contains the necessary source files.

jmcarcell: Added `submodules=true` following the README.

jmcarcell:
Dependency on cmake added.

repos/spack_repo/builtin/packages/pcre2/package.py
wdconinc: I guess I was incorrect: they did switch to submodules but we are not using the automatic git archives and the uploaded artifacts include the jit submodule. So we don't need the submodules after all.

PR #4124, opened 2026-04-06
kokkos: add version 5.1 and remove examples option

JBludau:
Update recipe for Kokkos 5.1.0 release (including nvcc_wrapper and Kokkos-kernels).
Also removes option to build the examples. There is little reason for that in the first place and no one should be using it so straight up removing it should be ok


repos/spack_repo/builtin/packages/kokkos/package.py
```diff
@@ -144,7 +145,6 @@ class Kokkos(CMakePackage, CudaPackage, ROCmPackage):
         "debug_bounds_check": [False, None, "Use bounds checking - will increase runtime"],
         "debug_dualview_modify_check": [False, "@:4", "Debug check on dual views"],
         "deprecated_code": [False, "@:4", "Whether to enable deprecated code"],
-        "examples": [False, "@:4", "Whether to build examples"],
```
JBludau, line 148: I don't think it is reasonable to have this in here.

repos/spack_repo/builtin/packages/kokkos/package.py
```diff
@@ -161,7 +161,6 @@ class Kokkos(CMakePackage, CudaPackage, ROCmPackage):
         "zen5": "ZEN5",
         "steamroller": "KAVERI",
         "excavator": "CARIZO",
-        "power7": "POWER7",
```
JBludau: Power 7 is not even listed in our wiki anymore. We also have no CI for it

repos/spack_repo/builtin/packages/kokkos_kernels/package.py
JBludau, line 181: @lucbv could you take a look at this (I have not looked at the tpls)? I guess that is way faster if you add new ones if there are some. Feel free to push to this branch

repos/spack_repo/builtin/packages/kokkos/package.py
```diff
@@ -233,6 +232,7 @@ class Kokkos(CMakePackage, CudaPackage, ROCmPackage):
         "gfx90a": "vega90A",
         "gfx940": "amd_gfx940",
         "gfx942": "amd_gfx942",
+        "gfx950": "amd_gfx950",
```
rbberger: Also missing, AMD_GFX1103 and AMD_GFX1201.

repos/spack_repo/builtin/packages/kokkos/package.py
rbberger: I would be good to know which versions were the last to support it.

repos/spack_repo/builtin/packages/kokkos/package.py
rbberger: I agree.

repos/spack_repo/builtin/packages/kokkos/package.py
JBludau: good catch. Added in https://github.com/spack/spack-packages/pull/4124/commits/728c654e58f32eb11d4c7d516496586dc4da1a4e

repos/spack_repo/builtin/packages/kokkos/package.py
JBludau: according to https://github.com/kokkos/kokkos/blob/301c37189a7fef46e68768ad9df160113f7ea052/CHANGELOG.md?plain=1#L652 it was 4.2.01.

rbberger: In that case we need to keep it until those versions are gone, no?

repos/spack_repo/builtin/packages/kokkos/package.py
tpadioleau: Do we have a policy to remove variants ? We have also other variants that should be removed at some point (like `cuda_uvm`).

repos/spack_repo/builtin/packages/kokkos/package.py
tpadioleau: I agree with @rbberger we should keep it if we still provide 4.2 in the package.

Should we later rework the architecture mapping to introduce a `when` conditional ?

repos/spack_repo/builtin/packages/kokkos/package.py
tpadioleau, line 1: Looking at the changelog https://github.com/kokkos/kokkos/blob/5.1.0/CHANGELOG.md, I only see:
- `Added Kokkos_ARCH_BLACKWELL103 configure option for NVIDIA B300 GPUs`
- `Add support for AMD MI355 and MI350 (AMD_GFX950)`

Did we miss the other architectures in previous PR ?

tpadioleau:
```suggestion
        "100": "blackwell100",
        "103": "blackwell103",
```

tpadioleau (changes requested):

repos/spack_repo/builtin/packages/kokkos/package.py
tpadioleau, line 383: I think we need something like the following patch to fix the neoverse build

```suggestion
    # adds missing `include(CheckIncludeFileCXX)` in the CMakeLists.txt
    patch(
      "https://github.com/kokkos/kokkos/commit/9e138d0f351693700a3cb0a403c754d6b7ba7da6.patch?full_index=1",
        sha256="c17f90ecf4704398e254a0f23d9c68e726906d31a2936aa185a5af459687e9b4",
        when="@4.7",
    )
```

cedricchevalier19: Not related to 5.1?

repos/spack_repo/builtin/packages/kokkos/package.py
cedricchevalier19: It is still not clear how to deprecate variants in Spack.

@JBludau can you update the PR description to list what "old options" you are removing?

repos/spack_repo/builtin/packages/kokkos/package.py
JBludau: fine with me. Although I doubt there is anyone still owning such a system

cedricchevalier19: And by the way we do not have a Kokkos CI for most of the architectures :(.

But it is true that we should condition the architecture to know when we can get rid of them.

repos/spack_repo/builtin/packages/kokkos/package.py
tpadioleau: I don't think so, the CI shows a problem with Kokkos 4.7, I wonder why this only comes up now

tpadioleau: https://gitlab.spack.io/spack/spack-packages/-/pipelines/1511258

repos/spack_repo/builtin/packages/kokkos/package.py
JBludau: done

repos/spack_repo/builtin/packages/kokkos/package.py
JBludau: This is a different problem about checking which architectures are actually supported by which kokkos version and is addressed in https://github.com/spack/spack-packages/pull/4161.
We decided in the dev meeting to put the patch in the queue for 4.7.04

repos/spack_repo/builtin/packages/kokkos/package.py
JBludau: The architectures are now done in https://github.com/spack/spack-packages/pull/4161

cedricchevalier19 (approved):

cedricchevalier19 (approved):
For Kokkos and nvcc_wrapper.

lucbv (approved):
Looks good to me, I will review Kokkos Kernels TPLs and update accordingly in another PR if that's fine

repos/spack_repo/builtin/packages/kokkos/package.py
dalg24: Power 7 was EOL'd by IBM in 2019 and I don't know any site that still deploy them.

repos/spack_repo/builtin/packages/kokkos/package.py
Adrien-Tab, line 119: This backend is removed in 5.1

Adrien-Tab: Why `OpenACC` isn't listed here ?

JBludau: good catch. I added a conflict for openmptarget. And according to our slack no one asked about the OpenACC so far

JBludau:
@spackbot fix style

rbberger:
@eugeneswalker There is an issue with the e4s-oneapi stack. You recently upgraded oneapi to a newer version, however, because of the variants for Trilinos (some of which are only supported in older versions), you are still building version 15.0.0. I suggest the variants in e4s-oneapi get updated to pull in the latest version @cgcgcg FYI:
https://github.com/spack/spack-packages/blob/dfff797e1a694372fa132208876e6b020ed5910f/stacks/e4s-oneapi/spack.yaml#L39-L43

cgcgcg:
The most recent version for Trilinos through spack is 16.2 at the moment. We are hoping to expose 17.0 very soon.
Trilinos 16.2 cannot handle the variant `rythmos`. Trilinos 17.0 will drop `amesos aztec epetra epetraext ifpack intrepid isoroppia ml`.

repos/spack_repo/builtin/packages/kokkos/package.py
cedricchevalier19, line 119: This variant should be conditional instead of a conflict, but we can improve the recipe later.

cedricchevalier19 (commented):
Ok for `kokkos` and `kokkos-nvcc-wrapper`.

cedricchevalier19 (approved):

repos/spack_repo/builtin/packages/kokkos/package.py
JBludau: According to our slack channel there never was anyone asking for this backend on spack thus we did not include it

cgcgcg (approved):

tpadioleau (approved):
Ok for me, thx!

repos/spack_repo/builtin/packages/kokkos/package.py
rbberger: Fixed in the other arch PR.

rbberger (approved):

PR #4137, opened 2026-04-07
davix: patch to avoid gsoap reading from stdin

wdconinc:
This PR fixes the installation of `davix +thirdparty` under the new installer. Due to calling `soapcpp2 -v` to get the version, it tries to read from stdin which doesn't get hangup right away as it does in the old installer (or something like that). The correct call is likely `soapcpp2 -V` which returns the version right away. This worked in the old installer since, accidentally, the failing output of `soapcpp2 -v` also contains a matching version number...


repos/spack_repo/builtin/packages/davix/package.py
```diff
@@ -62,6 +62,10 @@ class Davix(CMakePackage):
     variant("thirdparty", default=False, description="Build vendored libraries")
     depends_on("gsoap", when="+thirdparty")
 
+    # soapcpp2 -v (lowercase) reads from stdin instead of exiting; use -V (uppercase)
+    # which prints the version string and exits immediately.
+    patch("fix-gsoap-version-detection.patch", when="+thirdparty")
```
alalazo, line 67: Does the patch apply unconditionally to all versions?

wdconinc: Yes, the patch is applied to a file that has been changed only once since 0.6.7 (before the oldest spack version), in https://github.com/cern-fts/davix/commit/ba210a01ec3a86b32314c0ba726d19f5a60d955a, and that affected line 88 outside of the patch context window.

alalazo (approved):

PR #4142, opened 2026-04-07
libsvm: build library part

w8jcik:
`libsvm` has very simple Makefile https://github.com/cjlin1/libsvm/blob/master/Makefile which is not fully usable downstream.

Downstream projects like OpenMS build the package by themselves https://github.com/OpenMS/contrib/tree/186aacfbc1157719735b497cd6de1b3e83289eea/patches/libsvm, although only when their CMake does package management.

With this changes `libsvm` is built and installed without need for any further workarounds downstream.

This change is needed for OpenMS.


wdconinc (approved):
This looks sensible. Thanks for the addition!

PR #4143, opened 2026-04-07
dia-umpire-se: add new package

w8jcik:


repos/spack_repo/builtin/packages/dia_umpire_se/package.py
```diff
@@ -0,0 +1,74 @@
...
+    homepage = "https://diaumpire.nesvilab.org/"
+    url = "https://github.com/Nesvilab/DIA-Umpire/archive/refs/tags/v2.2.8.tar.gz"
+    git = "https://github.com/Nesvilab/DIA-Umpire.git"
+
+    maintainers("w8jcik")
+
+    license("GPL-3.0")
+
+    version("2.3.4", commit="68b73feeec6b4ef4812b5b3a7c410270e609b1ad")
+    version("2.3.3", commit="f4f806d517655aa06fd651ac7894f7da07636873")
```
wdconinc: Do we need to keep the older patch versions?

wdconinc (commented):
Looks good. Thanks for adding. One suggestion to reduce support burden.

repos/spack_repo/builtin/packages/dia_umpire_se/package.py
w8jcik: Removed patch versions as suggested

wdconinc (approved):

PR #4146, opened 2026-04-07
py-curryreader: add new package

manuelakuhn:
https://github.com/mne-tools/curry-python-reader/releases/tag/0.1.2


wdconinc (approved):
Thanks. LGTM.

PR #4149, opened 2026-04-07
g4hepem: new package

wdconinc:
This PR adds `g4hepem`, v20251114. This is a Geant4 EM physics simulation R&D project.

Test build (unable to try with CUDA):
```
-- linux-ubuntu26.04-skylake / %c,cxx=clang@23.0.0 --------------
h6yr537 g4hepem@20251114~cuda~ipo build_system=cmake build_type=Release dev_path=/home/wdconinc/git/g4hepem generator=make
==> 1 installed package
```

### DIFF
--- a/repos/spack_repo/builtin/packages/g4hepem/package.py
+++ b/repos/spack_repo/builtin/packages/g4hepem/package.py
@@ -1,0 +1,41 @@
+# Copyright Spack Project Developers. See COPYRIGHT file for details.
+#
+# SPDX-License-Identifier: (Apache-2.0 OR MIT)
+
+from spack_repo.builtin.build_systems.cmake import CMakePackage
+from spack_repo.builtin.build_systems.cuda import CudaPackage
+
+from spack.package import *
+
+
+class G4hepem(CMakePackage, CudaPackage):
+    """Geant4 EM physics simulation R&D project looking for solutions
+    to reduce the computing performance bottleneck experienced by
+    HEP detector simulation applications."""
+
+    homepage = "https://github.com/mnovak42/g4hepem"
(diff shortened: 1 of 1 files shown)


wdconinc:
@spackbot fix style

repos/spack_repo/builtin/packages/g4hepem/package.py
```diff
@@ -0,0 +1,43 @@
...
+    depends_on("cxx", type="build")
+    depends_on("cmake@3.17:", type="build")
+
+    depends_on("geant4@10.6:")
+
+    def cmake_args(self):
+        args = [
+            self.define("BUILD_TESTING", self.run_tests),
+            self.define("CMAKE_CXX_STANDARD", self.spec["geant4"].variants["cxxstd"].value),
+            self.define("G4HepEm_GEANT4_BUILD", True),
```
sethrj, line 40: If the purpose is to use this for adept, you also will need the flag `G4HepEm_EARLY_TRACKING_EXIT`: see https://github.com/apt-sim/AdePT#build-requirements .

sethrj (dismissed):

repos/spack_repo/builtin/packages/g4hepem/package.py
sethrj: It looks like turning this flag off is simply an optimization that skips a virtual function call in a tight loop.

wdconinc:
@spackbot fix style

repos/spack_repo/builtin/packages/g4hepem/package.py
wdconinc: I just added it as a variant :shrug:

sethrj (approved):
Thanks!

PR #4151, opened 2026-04-07
veccore: add cuda support (ahem)

wdconinc:
This PR adds to `veccore` a dependency on CudaPackage. This doesn't actually change what gets compiled (nothing) or installed in this header-only library, but it adds compile definitions to VecCoreConfig.cmake which are then available to downstream users when doing `find_package(VecCore COMPONENTS CUDA)` (e.g. https://github.com/apt-sim/AdePT/blob/master/CMakeLists.txt#L188-L189).

### DIFF
--- a/repos/spack_repo/builtin/packages/veccore/package.py
+++ b/repos/spack_repo/builtin/packages/veccore/package.py
@@ -3,11 +3,12 @@
 # SPDX-License-Identifier: (Apache-2.0 OR MIT)
 
 from spack_repo.builtin.build_systems.cmake import CMakePackage
+from spack_repo.builtin.build_systems.cuda import CudaPackage
 
 from spack.package import *
 
 
-class Veccore(CMakePackage):
+class Veccore(CMakePackage, CudaPackage):
     """SIMD Vectorization Library for VecGeom and GeantV.
 
     VecCore is a header-only (interface) library so no cmake arguments are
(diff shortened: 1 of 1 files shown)


wdconinc:
@spackbot fix style

sethrj (changes requested):
@wdconinc I'm pretty sure this won't affect anything about the install: the [config file](https://github.com/root-project/veccore/blob/master/cmake/VecCoreConfig.cmake.in) is quite primitive (constructing its own targets rather than using a Targets.cmake file) and it literally creates those targets based on what requested COMPONENTS are passed by the dependent. Unless I'm missing something it has nothing to do with what's done at compile time. (IIRC I'm also the one who removed the cuda inheritance from the spack package: the cuda_arch variable becomes a pain otherwise, especially since it's unused.)

wdconinc:
> @wdconinc I'm pretty sure this won't affect anything about the install: the [config file](https://github.com/root-project/veccore/blob/master/cmake/VecCoreConfig.cmake.in) is quite primitive (constructing its own targets rather than using a Targets.cmake file) and it literally creates those targets based on what requested COMPONENTS are passed by the dependent. Unless I'm missing something it has nothing to do with what's done at compile time. (IIRC I'm also the one who removed the cuda inheritance from the spack package: the cuda_arch variable becomes a pain otherwise, especially since it's unused.)

Yeah, you're right, with our without CUDA, the installed `lib/cmake/VecCore` files are all completely identical. I saw https://github.com/root-project/veccore/blob/master/CMakeLists.txt#L56-L60 and I figured it would end up in those targets.

PR #4152, opened 2026-04-07
pprof: add package

mcmehrtens:
- Add the `pprof` package
- `pprof` does not have a release schedule, release tags, or release branches so I only added one version (main)

### DIFF
--- a/repos/spack_repo/builtin/packages/pprof/package.py
+++ b/repos/spack_repo/builtin/packages/pprof/package.py
@@ -1,0 +1,36 @@
+# Copyright Spack Project Developers. See COPYRIGHT file for details.
+#
+# SPDX-License-Identifier: (Apache-2.0 OR MIT)
+
+from spack_repo.builtin.build_systems.go import GoPackage
+
+from spack.package import *
+
+
+class Pprof(GoPackage):
+    """pprof is a tool for visualization and analysis of profiling data."""
+
+    homepage = "https://github.com/google/pprof"
+    git = "https://github.com/google/pprof.git"
+
+    maintainers("mcmehrtens")
+    license("Apache-2.0", checked_by="mcmehrtens")
+
+    # pprof doesn't have tagged releases
+    version("main", branch="main", get_full_repo=True)
+
+    # pprof's go.mod requires go 1.24; Go supports the two most recent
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/pprof/package.py
```diff
@@ -0,0 +1,58 @@
...
+    """pprof is a tool for visualization and analysis of profiling data."""
+
+    homepage = "https://github.com/google/pprof"
+    git = "https://github.com/google/pprof.git"
+
+    maintainers("mcmehrtens")
+    license("Apache-2.0", checked_by="mcmehrtens")
+
+    # pprof doesn't have tagged releases
+    version("latest", branch="main", get_full_repo=True)
```
wdconinc: We try to keep names if branches in sync with their name in spack:

```suggestion
    version("main", branch="main", get_full_repo=True)
```

repos/spack_repo/builtin/packages/pprof/package.py
```diff
@@ -0,0 +1,58 @@
...
+        description="Enable graphic visualization of profiles (SVG, PDF, etc.)",
+    )
+
+    depends_on("graphviz", type="run", when="+graphviz")
+
+    @property
+    def sanity_check_is_file(self):
+        return [join_path("bin", "pprof")]
+
+    def build(self, spec: Spec, prefix: Prefix) -> None:
```
wdconinc: This seems like it reimplements more of GoPackage.build than may be needed: https://github.com/spack/spack-packages/blob/ad93f8d2faa86d3626f229cc0bbee1f0beb2f661/repos/spack_repo/builtin/build_systems/go.py#L85-L97

Tagging @alecbcs

wdconinc (commented):
Thanks for adding this package. One comment on potentially reducing boiler plate.

repos/spack_repo/builtin/packages/pprof/package.py
mcmehrtens: @wdconinc Yes, you're absolutely right. Good catch. I was copying from a different Go package I'd implemented where I did have to change these functions from their defaults.

This should be addressed now after rebasing.

wdconinc (approved):

PR #4154, opened 2026-04-07
pytest: add v9.0.3, deprecate older versions due to CVE

adamjstewart:
Note that deprecating/removing these older versions may make `spack unit-test` tricky on EOL vulnerable Python versions which no developer in their right mind should be using. It looks like we no longer vendor pytest, so one less thing to fix.

It's only a moderate CVE, so we could decide if we are okay with this while waiting on https://github.com/spack/spack/issues/52205


stacks/developer-tools-aarch64-linux-gnu/spack.yaml
```diff
@@ -83,7 +83,7 @@ spack:
   # spack developer tools
   # style
   - "py-mypy@0.900: ^py-mypy-extensions@:1.0"
-  - py-pytest@9.0.2
+  - py-pytest
```
johnwparent: Lets pin this since we use it for dev tool bootstrapping

johnwparent (commented):
Just one comment about leaving pytest pinned

stacks/developer-tools-aarch64-linux-gnu/spack.yaml
adamjstewart: Does pytest need to be pinned? mypy isn't. I've never seen a new pytest release make backwards-incompatible changes (compared to ruff and ty where this is frequent).

johnwparent: Mypy is pinned. `"py-mypy@0.900: ^py-mypy-extensions@:1.0"`
Its not about backwards compatibility, its about ensuring we have a stable cache hit when bootstrapping from binary only. 
Maybe I should just factor out the stuff that needs pinning into its own stack...

adamjstewart: Should we not always build the latest version such that the cache contains multiple versions? Or do old versions get purged from the cache? I would prefer not to pin the version used in the spack command such that even old versions of spack-core use the latest version in spack-packages. Basically, I don't like pinning anything that dependabot can't toggle.

johnwparent: I know we do purge things from the cache but I'm not sure of the policy. Let me checkup on that. 

The spack commands will use whatever version of ruff/pytest/mypy/etc that they're able to find in your environment, but for bootstrapping we use a specific version to allow for a stable bootstrap and baseline behavior guarantee.

I'm fine with building multiple py-tests (the more the better) and bootstrap currently points at a version of the cache that has the versions it needs, so this won't cause issues now, but in general I do think we need to provide some guarantees about the available versions of things we bootstrap so we can make that process as reliable as possible.

adamjstewart: Okay, I'll pin it for now so we can get this merged and we can continue discussing things when it comes up. Both your desire for stability and cache hits and my desire for latest-and-greatest are valid.

becker33 (changes requested):
This change would mean that Spack's unit tests cannot be run against the version of python on macos Tahoe, I think that is too restrictive.

The CVE in question is listed as moderate and the attack surface for it is fairly limited as well.

adamjstewart:
@becker33 how about the following compromise (just pushed): we deprecate everything, but we only remove older versions that aren't useful anyway and keep versions needed for running `spack test`. This is the same strategy we use in the Python package. This could be made more granular once we make progress on https://github.com/spack/spack/issues/52205. Thoughts?

becker33 (approved):
I still don't like this, but I think I agree that it's the best compromise we have available at the moment.

For a separate PR, we'll need to make it easier to use deprecated versions in bootstrapping to take advantage of this.

adamjstewart:
@alalazo is working on all deprecations becoming warnings instead of errors, so this should get better.

adamjstewart:
I would also like to continue to push on Spack to start dropping support for EOL Python versions. 3.10 will no longer receive security fixes in October, while we still support 3.6: https://devguide.python.org/versions/.

PR #4160, opened 2026-04-07
legion: propagate sysomp and max_dims to realm for recent legion versions

mpokorny:
A couple of overlooked issues with the last Legion package update regarding propagation of variants to Realm:
1. propagation of `max_dims` should be guarded with the minimum Legion version number to avoid adding an extraneous dependency of old Legion packages on the Realm package
2. missing propagation of `sysomp` variant entirely


elliottslaughter (approved):

rbberger (approved):

rbberger:
@mpokorny thanks for catching those!

PR #4163, opened 2026-04-07
openmpi: Add check for 'cc' in spack_compiler

AlexanderRichert-NOAA:
This PR fixes finding external openmpi. Currently, get_spack_compiler_spec() (sometimes? always?) fails with:
```console
  File "/dev/shm/spack-20770/user_cache/package_repos/fncqgg4/repos/spack_repo/builtin/packages/openmpi/package.py", line 1498, in get_spack_compiler_spec
    if spack_compiler.cc and spack_compiler.cc == compiler:
       ^^^^^^^^^^^^^^^^^
AttributeError: 'Spec' object has no attribute 'cc'
```

### DIFF
--- a/repos/spack_repo/builtin/packages/openmpi/package.py
+++ b/repos/spack_repo/builtin/packages/openmpi/package.py
@@ -1495,9 +1495,10 @@ def get_spack_compiler_spec(compiler):
     actual_compiler = None
     # check if the compiler actually matches the one we want
     for spack_compiler in spack_compilers:
-        if spack_compiler.cc and spack_compiler.cc == compiler:
-            actual_compiler = spack_compiler
-            break
+        if "cc" in spack_compiler:
+            if spack_compiler.cc and spack_compiler.cc == compiler:
+                actual_compiler = spack_compiler
+                break
     return actual_compiler.spec if actual_compiler else None
 
 


repos/spack_repo/builtin/packages/openmpi/package.py
```diff
@@ -1495,9 +1495,10 @@ def get_spack_compiler_spec(compiler):
     actual_compiler = None
     # check if the compiler actually matches the one we want
     for spack_compiler in spack_compilers:
-        if spack_compiler.cc and spack_compiler.cc == compiler:
-            actual_compiler = spack_compiler
-            break
+        if "cc" in spack_compiler:
```
johnwparent, line 1498: The check is for `"c" in <spec>`. 
`cc` is the attribute that provides the c compiler, but the virtual is named `c`. 

This check will also fall over as `find_compilers` returns `Spec` objects which, in the case of specs for compilers returned from `find_compilers` will be for external packages that themselves do not have a virtual dependency on `c` (generally). 
Instead we can check whether the underlying package has a valid c compiler via `spec.package.cc` (None if no detected compiler), or `spec.satisfies("c")`

AlexanderRichert-NOAA: Right, that should have been "c". I'm not sure what you mean about the underlying package? If you mean `spack_compiler` (i.e., `spack_compiler.package.cc` or `spack_compiler.satisfies("c")`), I don't think that works.
I could do `hasattr(spack_compiler, 'cc')` which would be narrower but feels a little ugly...

johnwparent: In this context, `spack_compiler` is a `Spec` for the compilers returned by `find_compiler`, so maybe "underlying" was a poor phrasing, I should have said "spec's package". 
 
For a compiler package (or at least those inheriting from `CompilerPackage`), they all have the attribute `cc`, but `cc` may be `None`, so the `hasattr` check won't be too informational. 

> i.e., spack_compiler.package.cc or spack_compiler.satisfies("c")),

This is exactly what I meant.

johnwparent: Actually I can't remember if the specs from `find_compiler` are concrete or not, if they're not the `spec.package` approach will not work, but the `spec.satisfies` approach will.

AlexanderRichert-NOAA: Indeed they're not concrete, so no .package attribute. But even with `if spack_compiler.satisfies("c")` I get the same error, so the satisfies check returns true but still no cc attribute. Here's the abstract compiler spec it's failing on:
{'name': 'llvm', 'versions': [17.0.6], 'architecture': None, '_dependencies': {}, 'namespace': None, 'abstract_hash': None, '_hash': None, '_package_hash': None, '_full_hash': None, '_build_hash': None, '_prefix': None, '_dunder_hash': None, '_concrete': False, '_external_path': '/usr', 'external_modules': None, 'extra_attributes': {'compilers': {'c': '/usr/bin/clang-17', 'cxx': '/usr/bin/clang++-17'}}, '_build_spec': None, 'annotations': SpecAnnotations().with_spec_format(5), '_package': None, 'variants': <spack.spec.VariantMap object at 0x7a56a6799610>, 'compiler_flags': <spack.spec.FlagMap object at 0x7a56a6799550>, '_dependents': {}}

hppritcha:
please address others' comments and I'll review.

repos/spack_repo/builtin/packages/openmpi/package.py
johnwparent: Sorry, I should have been clearer.

> but still no cc attribute.

`cc` is an attribute on the package, not the spec and `spack_compiler` is a spec. If you can't access the package, we can't query the `cc` attribute of said package.

However, because we're processing from `find_compiler` we know we're using the results of an external detection, and as such, we can use the `extra_attributes` spec property. `extra_attributes` is what is used to populate the section of the packages config entry that actually specifies the compilers, so pretty much exactly what we're looking for here (it's also what `CompilerPackage.cc` does under the hood). 

So a check could be something like:

```
if spack_compiler.satisfies("c") and spack_compiler.external:
  potential_compiler = spack_compiler.extra_attributes["compilers"]["c"]
  if potential_compiler == compiler: 
    actual_compiler = potential_compiler
    break
```
Potentially would want to add more checking for the `compilers` and `c` dict key presences.

AlexanderRichert-NOAA: Okay I see what's going on. Your solution didn't work unless I separately detect the external compilers first, but it's exactly the same with the current approach as far as I can tell. Maybe I'll create an issue for this to see if there's a way to detect compilers and openmpi in one `spack external find` command, but in any case, I can work around it by doing the compiler detection first. Closing this PR.

PR #4164, opened 2026-04-08
millepede: add v05-01-01; require gcc

wdconinc:
This PR adds to `millepede` the bugfix version v05-01-01 ([diff](https://gitlab.desy.de/millepede/millepede-ii/-/compare/V05-01-00...V05-01-01)). No changes needed to the package in spack.

Also, require gcc since the package hardcoded gcc, g++, gfortran use in https://gitlab.desy.de/millepede/millepede-ii/-/blob/main/CMakeLists.txt?ref_type=heads#L5-7. Revisit at next release in light of https://gitlab.desy.de/millepede/millepede-ii/-/merge_requests/54.


haampie (approved):
more correct is %c,cxx,fortran=gcc, but it probably does the job in practice

PR #4185, opened 2026-04-09
GoPackage: remove cgo_enabled attribute in favor of `%c`

alecbcs:
A while back I'd added a `cgo_enabled` property to GoPackages to disable CGO functionality by default. 

I realized we could better model this by simply checking if the package directly depends on a C compiler to set the environment variable without adding yet another package attribute to maintain. This change should be backwards compatible since the previous attribute didn't imply a C dependency and all packages here correctly specified both.


haampie (approved):

haampie:
lgtm, we'll see if there are exceptions to the rule later on.

PR #4191, opened 2026-04-09
cmake builder: add helper to force require/disable a package

sethrj:
This defines `CMAKE_{REQUIRE|DISABLE}_FIND_PACKAGE_ {Y}` (supported by CMake 3.22 and higher) based on a package variant (default `Y.lower()`). This will be used for packages whose CMake logic depends on `find_package({Y})` resulting in `{Y}_FOUND` rather than an explicit `{X}_USE_{Y}` variable.

### DIFF
--- a/repos/spack_repo/builtin/build_systems/cmake.py
+++ b/repos/spack_repo/builtin/build_systems/cmake.py
@@ -419,6 +419,11 @@ class CMakeBuilder(BuilderWithDefaults):
     def define(cmake_var: str, value: Any) -> str:
         return define(cmake_var, value)
 
+    @staticmethod
+    def define_package_from_variant(pkg: PackageBase, cmake_pkg: str, variant:
+                                    Optional[str] = None) -> str:
+        return define_package_from_variant(pkg, cmake_pkg, variant)
+
     def define_from_variant(self, cmake_var: str, variant: Optional[str] = None) -> str:
         return define_from_variant(self.pkg, cmake_var, variant)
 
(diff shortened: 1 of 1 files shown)


sethrj:
@spackbot fix style

sethrj:
Hi @haampie  , it looks like the builder/package/cmake changes a couple years ago (#47576) were largely yours. Could you please take a quick look at this to make sure I'm following the intended pattern? Thanks!

repos/spack_repo/builtin/build_systems/cmake.py
```diff
@@ -599,6 +609,59 @@ def define_from_variant(pkg: PackageBase, cmake_var: str, variant: Optional[str]
...
+    if variant is None:
+        variant = cmake_pkg.lower()
+
+    if not pkg.has_variant(variant):
+        raise KeyError(f'"{variant}" is not a variant of "{pkg.name}"')
+
+    if variant not in pkg.spec.variants:
+        return ""
+
+    behavior = "REQUIRE" if pkg.spec.variants[variant].value else "DISABLE"
```
alalazo, line 661: This seems to work only for boolean variant

sethrj: Thanks for taking a look at this! I think a REQUIRE/DISABLE macro like this only makes sense for a boolean variant: supporting "threads=openmp" would require an additional function argument (`value` rather than implicitly `True`) . Is that worth adding or should I document that it's intended only for boolean package variants (which I think are the predominant type in spack)?

PR #4194, opened 2026-04-09
intel-tbb: modify gcc conflict

afzpatel:
The GCC restriction was meant for versions 13 and newer. This update ensures the restriction now applies correctly to GCC 13 and above for Intel TBB versions up to 2021.3, fixing a build error I was getting with GCC 14.


afzpatel:
@rscohn2 please review

rscohn2 (approved):
Thanks!

PR #4201, opened 2026-04-09
root: add builtin_llvm variant to allow external LLVM

wdconinc:
This PR adds to `root` a user-settable `builtin_llvm` variant (default: true for backward compatibility).  When set to false, ROOT links against an external `llvm@20.1.0:20.1` package instead of building its own bundled copy.

ROOT's patches to llvm-project only touch `clang/` (65 files, zero in `llvm/` core), so vanilla LLVM is sufficient for this case.  ROOT still builds its own patched Clang from the bundled interpreter/llvm-project/clang sources (and `builtin_clang=ON` remains explicitly enforced), which is required because vanilla Clang cannot be used with ROOT.

External LLVM is gated to ROOT 6.36 to 6.38 where the required LLVM major version (20.1.x) is known.  A conflict is added to prevent `~builtin_llvm` on older and newer ROOT versions.

### DIFF
--- a/repos/spack_repo/builtin/packages/root/package.py
+++ b/repos/spack_repo/builtin/packages/root/package.py
@@ -299,6 +299,11 @@ class Root(CMakePackage):
     variant("x", default=(not _is_macos), description="Enable set of graphical options")
     variant("xml", default=True, description="Enable XML parser interface")
     variant("xrootd", default=False, description="Build xrootd file server and its client")
+    variant(
+        "builtin_llvm",
+        default=True,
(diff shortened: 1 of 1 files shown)


gartung (approved):

alecbcs (approved):
Looks good to me. Waiting on final approval from @vvolkl, @drbenmorgan, @marcmengel, @greenc-FNAL, @vitodb.

alecbcs:
Oops missed that @gartung had approved. I'll go ahead and merge.

PR #4206, opened 2026-04-10
libtool: Update package.py

nearsu:
Update Fortran compiler check logic to cope when 'f77' and/or 'fc' are missing as attributes in 'self.compiler'
Fixes compilation of libtool on my system when the compiler is autodetected


wdconinc:
I am not sure this is the right solution here.

repos/spack_repo/builtin/packages/libtool/package.py
```diff
@@ -117,9 +117,9 @@ def setup_build_environment(self, env: EnvironmentModifications) -> None:
 
     def setup_test_environment(self, env: EnvironmentModifications):
         """When Fortran is not provided, a few tests need to be skipped"""
-        if self.compiler.f77 is None:
+        if (not hasattr(self.compiler, "f77")) or (self.compiler.f77 is None):
             env.set("F77", "no")
-        if self.compiler.fc is None:
+        if (not hasattr(self.compiler, "fc")) or (self.compiler.fc is None):
```
alalazo, line 125: This is not the right way to check if a spec needs Fortran. This method needs more updates. A trivial one would be to set the environment variables unconditionally. Wondering if we can delete the entire method though.

alalazo (changes requested):

repos/spack_repo/builtin/packages/libtool/package.py
nearsu: I will give that a try on my machine and see if it is working or not, else will try to update my pull request some more

nearsu: I have been reviewing this issue and think that the "env.set("F77", "no")" and "env.set("FC", "no")" need to stay due to the behavior of the libtool test suites (and spack).

When they are removed, there are many more fails in the test suite.
The libtool changelog documents the use of these environment flags to skip certain tests when there is no fortran compiler.
It seems that if this isn't set, then libtool does not detect that there is no fortran compiler and tries to run them anyway.
(Perhaps the libtool test suites needs a patch? Spack certainly doesn't use the latest version of libtool)
https://fossies.org/linux/libtool/ChangeLog
17776 2005-07-30  Peter Ekberg  <peda@lysator.liu.se>
17778 	* m4/libtool.m4 (_LT_PROG_F77): Set it up so that saying F77=no
17779 	to configure disables the fortran tests in the testsuite.

I think I read last week in the documentation the use of F77=no and FC=no to indicate that there is no fortran compiler is supported by autotools (!?) but can't find the reference again...
But you can always read it in the configure script:
25072 if test -z "$FC" || test no = "$FC"; then
25073   _lt_disable_FC=yes
25074 fi

Spack doesn't configure these environment variable to 'no' by default if there is no fortran compiler, so this probably has to stay for this package

I also tried to add fortran/cxx as dependencies (with depends_on) in the libtool package for testing purposes.
But Spack doesn't seem to identify these dependencies for some reason, and it isn't picked up by the 'compiler-wrapper' package so no FC or CXX flags are set, only 'CC'.
I have been trying to debug this but no luck yet as I am new to Spack...

Perhaps alalazo's trivial solution of "set(ting) the environment variables unconditionally" in the test suite is OK.
When fortran is not available at compile time then FC=no, F77=no will be set.
But if fortran is available then hopefully Spack will overwrite these environment values automatically
However, I can't get fortran added to either the test or build environment so can't test this at the moment.

nearsu: It seems that the concretizer doesn't like to update compiler specs for packages that are already installed, and the fix is to force uninstall the existing package before trying to re-concrete the specs and re-install the same package.
Using the spack install --overwrite option doesn't update the spec properly by itself, seems like a bug to me.

I have updated the PR for this package to set 'FC/F77=no' for both build and test environments so that libtool 'configure' knows whether or not fortran is around by removing the 'setup_test_environment' method. 
It seems like the build and test environments should have been the same in the first place.

The option of just always setting 'FC/F77=no' works fine for gcc as far as I can tell
The more cautious tests to see if the fortran compilers are defined in 'self.compiler' are still in this PR though. 
But these tests could be removed if it can be shown that both 'FC' and 'F77' are guaranteed to be overwritten by the spack compiler wrappers for all fortran compilers when fortran is present.

I also added some more 'depends_on' rules about dependencies to auto-load cxx, fortran and java if they are in the environment or not. 
However, the 'java' rule doesn't seem to be working at the moment
What do the reviewers think of the following type of optional dependency statement:

depends_on("cxx", type=("build", "test"), when="^cxx")

If you think this style OK for optional dependencies that should be pulled in when in the environment, I will try to diagnose and fix the java issue
Otherwise, using variants to describe optional test dependencies could be more ugly and possibly non-consistent with the rest of the package repo.

nearsu: Or I could just remove all the unnecessary dependencies that are not 'c' to keep things simple...

PR #4221, opened 2026-04-10
py-gensim: added version 4.4.0

qwertos:


repos/spack_repo/builtin/packages/py_gensim/package.py
```diff
@@ -19,6 +19,11 @@ class PyGensim(PythonPackage):
 
     license("LGPL-2.1-only")
 
+    version(
+        "4.4.0",
+        url="https://github.com/piskvorky/gensim/archive/refs/tags/4.4.0.tar.gz",
```
adamjstewart, line 27: Why GitHub and not PyPI?

adamjstewart (commented):
Everything else looks correct, just one question

repos/spack_repo/builtin/packages/py_gensim/package.py
qwertos: > Using the tarball from github as the one from pypi contains c and cpp
> files generated using cython version 0.29.32. These will not compile
> with python@3.12: or py-numpy@1.26:

adamjstewart: Can you add this comment to the file?

adamjstewart (approved):

PR #4224, opened 2026-04-10
py-rachis: new package

qwertos:


alecbcs (approved):
Looks good to me. Thanks @qwertos!

PR #4227, opened 2026-04-11
openmpi: --disable-dependency-tracking

haampie:
With dependency tracking enabled, make parses about a million
lines of dependency info generated by configure. The overhead
of this for the build is about 30% on my macbook. (edit: on my
desktop with tmpfs it's just 10%)

Enabling dependency tracking is only relevant if you're developing
openmpi. In case people do that, they can conditionally add the flag,
like `if not spec.satisfies("dev_path=*"): args.append("...")`, but I
don't think that's used right now.


alecbcs (approved):
Looks good to me. Will wait a couple days to hear back from @hppritcha @naughtont3 before merging. Good find @haampie!

hppritcha (approved):

PR #4230, opened 2026-04-12
awscli-v2: relax dependencies to support newer cryptography

adamjstewart:
Starting to feel like a broken record. Perhaps we could remove upper bounds on all deps, they don't seem useful.


teaguesterling:
I agree with that. They were originally added in probably because I didn't know what I was doing.

teaguesterling (approved):

wdconinc (approved):
LGTM and approved by maintainer.

PR #4233, opened 2026-04-12
uv: add v0.11.6

adamjstewart:
@johnwparent sorry for the PR conflict, let me know if you want me to try hacking on the multi-build system thing so it doesn't end up abandoned.


wdconinc (approved):
LGTM. (MSRV 1.90 also folded in 0.9.27 so gap is intended.)

PR #4238, opened 2026-04-13
py-sphinxcontrib-bibtex: add v2.6.5

jagot:
I bumped the compat bounds where applicable.

Actually, maybe I have should have bumped the compat bound of py-pybtex-docutils to `1.0.2:`, per https://github.com/mcmtroffaes/pybtex-docutils/issues/26 but I opted not too, since in the 2.6.5 release, `pyproject.toml` lists 1.0.0 as the lower bound, and I felt it could be confusing to have differing bounds. As soon as [2.6.6 is released](https://github.com/mcmtroffaes/sphinxcontrib-bibtex/blob/603d5ecbafa439eb4ee75d630b9f9094f2bd7221/CHANGELOG.rst), this will of course change.

PR #4239, opened 2026-04-13
Updated packages: furo (2025.12.19) and py-accessible-pygments (0.0.5)

jagot:
I opted to update py-accessible-pygments as well, since the new version of furo depends on it. Please let me know if you prefer the commits to be split over two PRs instead.

### DIFF
--- a/repos/spack_repo/builtin/packages/py_accessible_pygments/package.py
+++ b/repos/spack_repo/builtin/packages/py_accessible_pygments/package.py
@@ -16,7 +16,19 @@ class PyAccessiblePygments(PythonPackage):
 
     license("BSD-3-Clause")
 
+    version("0.0.5", sha256="40918d3e6a2b619ad424cb91e556bd3bd8865443d9f22f1dcdf79e33c8046872")
     version("0.0.4", sha256="e7b57a9b15958e9601c7e9eb07a440c813283545a20973f2574a5f453d0e953e")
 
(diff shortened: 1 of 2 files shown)


repos/spack_repo/builtin/packages/py_accessible_pygments/package.py
alecbcs, line 24:
```suggestion
depends_on("python@3.9:", type=("build", "run"), when="@0.0.5:")

depends_on("py-pygments@1.5:", type=("build", "run"))
depends_on("py-setuptools", type=("build"))
```

Quick suggestion to add the newly added Python lower bound that was added in v0.0.5.

alecbcs (changes requested):
One quick suggestion to add the new lower bound on Python v3.9. Otherwise looks good to me. Thanks!

jagot:
I can't seem to get the `verify-checksums` check to work, although I have copied the checksum from here: https://pypi.org/project/accessible-pygments/#accessible_pygments-0.0.5.tar.gz (I'm seeing the same problem in https://github.com/spack/spack-packages/pull/4238).

alecbcs:
PyPI began disallowing hyphens in package names last year and replacing them with underscores in URLs. We've had to update [other packages](https://github.com/spack/spack-packages/blob/develop/repos/spack_repo/builtin/packages/py_scikit_optimize/package.py) by modifying their `url_for_version()` methods.

alecbcs (approved):
Looks good to me. Thanks @jagot!

PR #4243, opened 2026-04-13
py-virtualenv: add v21.2.0

manuelakuhn:
https://github.com/pypa/virtualenv/releases/tag/21.2.0


wdconinc (approved):
LGTM. Verified dependency changes.

PR #4250, opened 2026-04-13
py-tzdata: add v2026.1

RobertMaaskant:
Release notes: https://github.com/python/tzdata/blob/2026.1/NEWS.md#version-20261
Diff: https://github.com/python/tzdata/compare/2025.3..2026.1


manuelakuhn (approved):
LGTM and builds fine on rockylinux 8 with gcc@15

PR #4256, opened 2026-04-13
gperf: skip broken gnulib conftest when building with oneapi

teabagk7:
**Change:** Add `configure_args()` that passes `gl_cv_func_working_error=yes` for `%oneapi`.

 **What happens:** Configure hangs forever, eating RAM until OOM kill.

 **Why:** icx has a bug where it unconditionally marks glibc's `error()` as `noreturn`. In reality, `error(status, ...)` only exits when `status != 0`. The gnulib conftest calls `error(0, 0, "foo")` expecting it to return. icx optimizes away the return path at `-O2`, so the program falls off the end of `main()` into an infinite loop.

 **Fix rationale:** Setting the cache variable skips the runtime test. The answer `yes` is correct — glibc's `error()` works fine, it's purely an icx misoptimization. Same pattern already used in libiconv for an NVHPC workaround (`gl_cv_cc_wallow=none`).

### DIFF
--- a/repos/spack_repo/builtin/packages/gperf/package.py
+++ b/repos/spack_repo/builtin/packages/gperf/package.py
@@ -40,7 +40,9 @@ class Gperf(AutotoolsPackage, GNUMirrorPackage):
         # Intel oneAPI icx incorrectly marks glibc's error() as noreturn,
         # causing the gnulib gl_cv_func_working_error configure test to
         # infinite-loop and consume unbounded memory.
-        if self.spec.satisfies("%oneapi"):
+        # Fix available in icx 2026, but this workaround is needed for all versions of icx up to 2025.
+        # https://community.intel.com/t5/Intel-oneAPI-DPC-C-Compiler/All-versions-of-icx-miscompile-error-0-resulting-in-segfaults/m-p/1744208
+        if self.spec.satisfies("%oneapi@:2025"):
             args.append("gl_cv_func_working_error=yes")
 
         return args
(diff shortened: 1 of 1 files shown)


haampie:
Hm... does this mean you have to update almost every gnu package.py? I think most of them bundle gnulib.

Is this fixed upstream either in gnulib or the compiler?

haampie:
Sigh, indeed: https://godbolt.org/z/jME7G3nGb it is a bug in `icx`.

And gnulib has this check since 2023 with no fix since: https://gitweb.git.savannah.gnu.org/gitweb/?p=gnulib.git;a=blob;f=m4/error_h.m4;h=73e969430d992ead6a7c063d74d885b97689ab5e;hb=HEAD

Should really be fixed upstream, primarily in oneapi compilers, but it should have a workaround in gnulib too.

I would not merge this until there are references to those issues, otherwise it's hard to keep track of.

haampie:
Can you add a link to 

https://community.intel.com/t5/Intel-oneAPI-DPC-C-Compiler/All-versions-of-icx-miscompile-error-0-resulting-in-segfaults/m-p/1744208

I've also raised the issue in the gnulib mailing list, will update this comment once it's published in the archives.

haampie:
Upstream they say it's fixed in the upcoming release, can you make it conditional on `intel-oneapi-compilers@:2025`?

haampie (approved):

teabagk7:
Thank you for putting in the effort to research all of that!

PR #4259, opened 2026-04-13
pbzip2: include oneapi in reserved-user-defined-literal workaround

teabagk7:
**Change:** Add `or self.spec.satisfies("%cxx=oneapi")` to the existing `flag_handler` condition.

**What happens:** 6 compile errors from `"%"PRIuMAX` patterns (26 occurrences) — C++11 requires a space between the string
literal and the macro.

The package already suppresses this for `%cxx=clang` and `%cxx=apple-clang` with `-Wno-reserved-user-defined-literal`. oneAPI icpx is clang-based and needs the same flag, but wasn't included.


teabagk7:
@Markus92 Could you please take a look?

For some reason you're not tagged automatically...

Markus92:
@teabagk7 LGTM!

haampie (approved):

PR #4260, opened 2026-04-13
relion: make py-relion and ghostscript optional via variants

teabagk7:
### `python` variant (default=true, `@5:`)

 **Problem:** `py-relion` is unconditionally pulled in for relion 5.x. It depends on py-torch, py-napari, model-angelo, py-pyqt5, and dozens more. This makes concretization extremely slow or impossible with non-default compilers — the solver has to reconcile hundreds of extra packages.

 **Fix:** Gate `py-relion` behind `+python`. HPC users who only need the core C++ binaries (`relion_refine`, `relion_autopick`, etc.) can build with `~python`. When `~python`, `cmake_args` sets `-DPYTHON_EXE_PATH=` to disable Python integration at the CMake level.

 ### `ghostscript` variant (default=true, `@4:`)

 **Problem:** ghostscript is unconditional for `@4:` and drags in GTK+3, librsvg, cairo, pango, fontconfig, gperf, libiconv, and many X11 libs — a heavy chain that significantly increases build time and surface area for compiler compatibility issues.

 **What ghostscript actually does:** Two functions in `src/CPlot2D.cpp` call `gs` via `system()` to merge EPS diagnostic plots into PDF logfiles. When `gs` is missing, relion catches the failure, prints a warning, and creates an empty placeholder PDF. Individual EPS plots are still produced. Core scientific computation is completely unaffected.

 **Fix:** Gate behind `+ghostscript`. Default=true preserves backward compatibility for both variants.

### DIFF
--- a/repos/spack_repo/builtin/packages/relion/package.py
+++ b/repos/spack_repo/builtin/packages/relion/package.py
@@ -44,6 +44,12 @@ class Relion(CMakePackage, CudaPackage):
     variant("cuda", default=True, description="enable compute on gpu")
     variant("double", default=True, description="double precision (cpu) code")
     variant("double-gpu", default=False, description="double precision gpu")
+    variant(
+        "python",
+        default=True,
+        description="Include py-relion Python tools (torch, napari, etc.)",
+        when="@5:",
+    )
 
     # if built with purpose=cluster then relion will link to gpfs libraries
     # if that's not desirable then use purpose=desktop
@@ -79,6 +85,12 @@ class Relion(CMakePackage, CudaPackage):
         default=False,
         description="Have external motioncor2 available in addition to Relion builtin",
     )
+    variant(
+        "ghostscript",
+        default=True,
+        description="Use ghostscript for merging EPS diagnostic plots into PDF logfiles",
+        when="@4:",
+    )
 
     depends_on("c", type="build")
     depends_on("cxx", type="build")
@@ -101,18 +113,22 @@ class Relion(CMakePackage, CudaPackage):
     depends_on("ctffind@:4", type="run")
     depends_on("motioncor2", type="run", when="+external_motioncor2")
 
-    depends_on("ghostscript", type="run", when="@4:")
+    # ghostscript is only needed at runtime for merging EPS diagnostic plots
+    # into PDF logfiles. Relion gracefully degrades when gs is missing
+    # (creates empty placeholder PDFs, individual EPS files are still produced).
+    depends_on("ghostscript", type="run", when="+ghostscript")
     depends_on("pbzip2", type="run", when="@5:")
     depends_on("xz", type="run", when="@5:")
     depends_on("zstd", type="run", when="@5:")
 
+    # py-relion is now gated behind +python variant
     for arch in CudaPackage.cuda_arch_values:
         depends_on(
             f"py-relion@5.0.1 +cuda cuda_arch={arch}",
             type=("build", "run"),
-            when=f"@5.0.1 +cuda cuda_arch={arch}",
+            when=f"@5.0.1 +python +cuda cuda_arch={arch}",
         )
-    depends_on("py-relion@5.0.1 ~cuda", type=("build", "run"), when="@5.0.1 ~cuda")
+    depends_on("py-relion@5.0.1 ~cuda", type=("build", "run"), when="@5.0.1 +python ~cuda")
 
     patch("0002-Simple-patch-to-fix-intel-mkl-linking.patch", when="@:3.1.1 os=ubuntu18.04")
     patch(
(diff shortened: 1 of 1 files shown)


teabagk7:
+@Markus92

Markus92:
The `py-relion` is indeed an absolute beast of a package. I think in our system it loads like 289 modules on runtime and concretizing took an hour or so. It's mostly pure Python which works fine across most compilers but I get your reasoning.

One problem is the RELION 5 GUI does not degrade too gracefully without the Python environment. It will keep showing all the options but half the commands just wont work. This led to some very "interesting" (read: angry) user feedback that half the RELION commands didn't work with vague Python errors and that led me to the rabbit hole of creating the `py-relion` package.

I'm not sure what to best way to avoid this is. Maybe remove all the `relion_python_*` files in `prefix/bin` if `~python`, so the errors get a lot more clear? Or replace them with a different stub that shows a "not built with Python" message? It seems to be around a dozen commands that use Python.

teabagk7:
Yeah, stubs with a clear message sounds like the right call. We need to loop over the `relion_python_*` scripts in post-install and replace them with a one-liner that tells users to rebuild with python support.

I can put that together. Or if you'd rather handle it yourself since you know the Python entry points best — happy either way, just let me know.

Markus92:
Something like this should do, right? Makes it clear for the user

```bash
#!/usr/bin/env bash

echo "This build of RELION does not support Python commands."
exit 1
```

Not sure if I have permissions to add it to your PR so if you can add something like that it's probably easiest

teabagk7:
> Something like this should do, right? Makes it clear for the user
> 
> ```shell
> #!/usr/bin/env bash
> 
> echo "This build of RELION does not support Python commands."
> exit 1
> ```
> 
> Not sure if I have permissions to add it to your PR so if you can add something like that it's probably easiest

No problem, I've done it 👍🏻

teabagk7:
@Markus92 @dacolombo
Can we merge the changes?

Markus92:
@teabagk7 I've tested your build out with the new options and it works as expected! LGTM

teabagk7:
> @teabagk7 I've tested your build out with the new options and it works as expected! LGTM

I thought you can approve, you're listed as maintainer

Markus92:
Don't see an approval button anywhere nor can I add myself as a reviewer it seems. I'll ask on Slack.

teabagk7:
@alalazo something doesn't work properly.
Markus is listed as a maintainer, but it wasn’t automatically tagged and lacks an approval button.
The same for https://github.com/spack/spack-packages/pull/4259

teabagk7:
@haampie can you please help merging this?

repos/spack_repo/builtin/packages/relion/package.py
```diff
@@ -101,18 +115,22 @@ class Relion(CMakePackage, CudaPackage):
     depends_on("ctffind@:4", type="run")
     depends_on("motioncor2", type="run", when="+external_motioncor2")
 
-    depends_on("ghostscript", type="run", when="@4:")
+    # ghostscript is only needed at runtime for merging EPS diagnostic plots
+    # into PDF logfiles. Relion gracefully degrades when gs is missing
+    # (creates empty placeholder PDFs, individual EPS files are still produced).
+    depends_on("ghostscript", type="run", when="+ghostscript")
```
haampie, line 121: Are you using ~ghostscript? Otherwise I would not introduce the variant and instead make it an unconditional dependency.

teabagk7: Yes, we use ~ghostscript. We're generally trying to minimize the dependency footprint to make build easier and faster.

repos/spack_repo/builtin/packages/relion/package.py
```diff
@@ -155,7 +173,10 @@ def cmake_args(self):
             args.append("-DCUDA=OFF")
 
         if self.spec.satisfies("@5:"):
-            args.append(f"-DPYTHON_EXE_PATH={self.spec['python'].command.path}")
+            if self.spec.satisfies("+python"):
+                args.append(f"-DPYTHON_EXE_PATH={self.spec['python'].command.path}")
+            else:
+                args.append("-DPYTHON_EXE_PATH=")
```
haampie: Maybe do something like `/does-not-exist`? https://github.com/3dem/relion/blob/f2c1a384400aec37dc6805856a5ba645650a44f1/cmake/FindCondaPython.cmake#L36

Or submit a PR upstream to make it possible to disable. They should just use `find_package(Python`

teabagk7: Good point, empty string triggers the Conda auto-discovery path...
I've changed to empty string to proposed `/does-not-exist` and added comment. Thanks for that suggestion!

@Markus92, what are your thoughts about Relion build options changes?

haampie (approved):

alecbcs:
@haampie are you ready to merge this PR?

repos/spack_repo/builtin/packages/relion/package.py
Markus92: @teabagk7 sorry was out on vacation, yeah looks good to me, should disable the Python stuff just fine!

teabagk7:
@haampie @alalazo please help merging it. PR is already approved.

PR #4265, opened 2026-04-13
musica: add c compiler

K20shores:
I noticed we couldn't install one of our options because we needed the c compiler to be set

### DIFF
--- a/repos/spack_repo/builtin/packages/musica/package.py
+++ b/repos/spack_repo/builtin/packages/musica/package.py
@@ -47,6 +47,7 @@ class Musica(CMakePackage):
 
     # Dependencies
     depends_on("cmake@3.21:", type="build")
+    depends_on("c", type="build")
     depends_on("cxx", type="build")
     depends_on("fortran", type="build")
     depends_on("mpi", when="+mpi")


boulderdaze (approved):

wdconinc (approved):
LGTM. CMake's `project` defaults to C and CXX as LANGUAGES when unspecified.

K20shores:
@wdconinc yes it does, but for some reason spack, at least on my local computer, wouldn't build my package with the tuvx option unless this was specified. I don't really understand why

wdconinc:
> @wdconinc yes it does, but for some reason spack, at least on my local computer, wouldn't build my package with the tuvx option unless this was specified. I don't really understand why

As of a year or so ago, spack requires specifying all compilers that are needed by a project through the specification of these `c` and `cxx` dependencies. A common issue for C++ projects is that they only contain C++ code, but have a simple `project` definition in the CMakeLists.txt that omits the `LANGUAGES CXX`. In that case, CMake will still try to compile a C program, and without `c` dependency it won't have access to a C compiler. So that's why that fails (unrelated to the variant; this was likely broken since last year).

PR #4282, opened 2026-04-13
openblas: add 0.3.33, variant for max_num_threads

### LINKED ISSUE #4178: Considerations on OpenBLAS NUM_THREADS limit
Hello, 

We are starting to get machine with a realy large amount of physical threads. For instance, the AMD EPYC 9965 ships 192 cores with 2way SMT. In a bisocket configuration, that's a total of 768 hardware threads.

In some recipes, such as OpenBLAS, there is a hardcoded limit to 512 threads : 
https://github.com/spack/spack-packages/blob/adaa2a170dbf8c5a98c71446a2e9cb68721b7d39/repos/spack_repo/builtin/packages/openblas/package.py#L613

I understand that going for 768 threads might not be the best solution for performance; but sometime you need to be able to benchmark to have numbers to help you conclude and convince people. That's actually not possible with the current hardcoded limit.

I am wondering if this NUM_THREADS value should be exposed as a variant, or if the hardcoded limit should simply be increased (regularly ?). Or if there are some other (more elegant?) solutions... 

Best regards.

mathomp4:
Closes #4178 

This PR adds `openblas` 0.3.33. It also a new variant for `openblas`, `max_num_threads`, to allow users to change the `NUM_THREADS` used at build time.

We default to the old value of 512, as that is good for most machines without much cost (see https://github.com/spack/spack-packages/issues/4178#issuecomment-4239472982)

cc: @antoine-morvan @martin-frbg


alalazo:
Can you separate adding the new version from adding the new variant?

alalazo:
@spackbot run pipeline

repos/spack_repo/builtin/packages/openblas/package.py
```diff
@@ -131,6 +132,16 @@ class Openblas(CMakePackage, MakefilePackage):
...
 
+    # We add a variant to allow setting of NUM_THREADS as on some machines, 512 might be
+    # too small. But we default to 512 as per OpenBLAS maintainer higher numbers
+    # will only lead to unnecessary memory usage and potential bottlenecks
+    # see https://github.com/spack/spack-packages/issues/4178#issuecomment-4239472982
+    variant(
+        "max_num_threads",
+        default="512",
+        description="Set the default number of threads for OpenBLAS",
+    )
```
alalazo: Can we make this variant conditional? Right now it enters the hash also for `threads=none` but in that case it's unused

alalazo (changes requested):

repos/spack_repo/builtin/packages/openblas/package.py
mathomp4: @alalazo Oh, makes sense. That said, I think it's learning time for me. How would I best do this? I could duplicate that block twice, one for `when='threads=openmp'` and one for `when='threads=pthreads'` but that seems inelegant.

And I can't seem to figure out how to do an "or" in a `when`.

alalazo: We don't have logical connectors for specs, but you can use a `for` loop:
```python
for _when_condition in ("threads=openmp", "threads=pthreads"):
    variant(..., when=_when_condition)
```

mathomp4: Ah. Okay. So you get two "entries" in `spack info openblas` no matter what:

```
    max_num_threads [512]           none
      when  threads=openmp
        Set the default number of threads for OpenBLAS

    max_num_threads [512]           none
      when  threads=pthreads
        Set the default number of threads for OpenBLAS
```

I was hoping to avoid, but at least the code is a bit cleaner now 🙂

mathomp4: Okay. That seems to work:
```
❯ spack spec -L openblas threads=none
 -   gl5vj5qrzr6bkukp2eeurqaosozn6c7h  openblas@0.3.33~bignuma~consistent_fpcsr~dynamic_dispatch+fortran~ilp64~ipo+locking+pic+shared~static build_system=cmake build_type=Release generator=make max_num_threads=512 symbol_suffix=none threads=none platform=darwin os=tahoe target=m2 %c,cxx=apple-clang@21.0.0 %fortran=gcc@15.2.0
...
❯ spack spec -L openblas threads=none max_num_threads=512
 -   gl5vj5qrzr6bkukp2eeurqaosozn6c7h  openblas@0.3.33~bignuma~consistent_fpcsr~dynamic_dispatch+fortran~ilp64~ipo+locking+pic+shared~static build_system=cmake build_type=Release generator=make max_num_threads=512 symbol_suffix=none threads=none platform=darwin os=tahoe target=m2 %c,cxx=apple-clang@21.0.0 %fortran=gcc@15.2.0
```

mathomp4: @alalazo Change pushed. Please take a look.

alalazo (approved):

alalazo:
@spackbot run pipeline

PR #4284, opened 2026-04-13
topaz: adding v0.3.18

snehring:
new version, supposedly good to unbound the torch dep.

also bumps py-trove-classifiers as I was running into build errors with some dependencies with the older one.


RikkiButler20 (approved):
LGTM

PR #4286, opened 2026-04-13
Add version labels to checkout action version

kwryankrattiger:
Add labels to match core


alecbcs:
I don't think we should add these. Dependabot is already automatically updating these actions and unfortunately it won't (or didn't last time I looked) keep these labels up to date. So they'll either be quickly out of date, or will require human intervention by us to manually keep up-to-date.

kwryankrattiger:
Fair enough, it would be nice if dependabot did better on that. It is useful to be able to glance at the action and know what it is.

PR #4291, opened 2026-04-14
ucx: add external package detection

dhairyashilRG:
## What
Add external package detection for ucx (`executables`, `determine_version`,`determine_variants`), so `spack external find ucx` picks up system/vendor UCX installs (distro packages, NVIDIA OFED/HPCX, etc.).

## Why
Addresses [openucx/ucx#8497](https://github.com/openucx/ucx/issues/8497) - users want Spack to auto-detect externally installed UCX.

## How
Detection reads the build configuration reported by `ucx_info -b` (the installed `config.h`), not the `# Configured with:` line from `ucx_info -v`. The configure line is unreliable as a source because most UCX components are either enabled by default (`rc`, `ud`, `dc`, `dm`, `ib_hw_tm`, `mlx5`) or auto-detected (`rdmacm`, `knem`, `xpmem`, `bfd`, `fuse3`, `cuda`, `rocm`), so they never appear there unless a packager explicitly overrode them — which badly under-detects vendor and distro builds.

From `ucx_info -b`:
- transports / build options come from their `#define`s (`HAVE_TL_RC`, `HAVE_IBV_DM`, `IBV_HW_TM`, `ENABLE_MT`, ...)
- components come from the `<component>_MODULES` lists (`uct_MODULES`, `uct_cuda_MODULES`, `uct_ib_MODULES`, `ucm_MODULES`, `ucs_MODULES`)

Module-backed components key off the module lists rather than a `HAVE_*` define because a `HAVE_*` define only records that headers were found, not that the module was built (e.g. a build with ROCm headers but configured `--without-rocm reports `HAVE_HIP 1` while `uct_rocm_MODULES` stays empty). `java` is the only variant with no build-config signal, so it still comes from the configure line.

## Testing
Verified on Ubuntu 24.04 (x86_64) with real `spack external find ucx`, against a distro UCX **1.16.0** and a source-built **1.18.0** configured with explicit negatives. Detected specs pass Spack's variant validation. Reading the build config instead of the configure line corrected 7 variants on the stock distro

build:

before: ucx@1.16.0 +assertions ~backtrace_detail ~cma ~debug +logging +mlx5_dv
                      +parameter_checking +thread_multiple +verbs
                      
after:  ucx@1.16.0 +assertions ~backtrace_detail +cma ~cuda +dc ~debug +dm
                   ~gdrcopy +ib_hw_tm ~java ~knem +logging +mlx5_dv
                   +parameter_checking +rc +rdmacm ~rocm +thread_multiple +ud
                   +verbs ~vfs ~xpmem

The CUDA/ROCm/gdrcopy/knem/xpmem/vfs positive paths are covered by synthetic `ucx_info -b` inputs (module token names taken from UCX's own `configure.m4`); they haven't been confirmed on a real GPU/IB build or on aarch64 / Grace Hopper

### DIFF
--- a/repos/spack_repo/builtin/packages/ucx/package.py
+++ b/repos/spack_repo/builtin/packages/ucx/package.py
@@ -231,19 +231,22 @@ class Ucx(AutotoolsPackage, CudaPackage):
 
             # cm variant (only for UCX @:1.10)
             if version in ver(":1.10"):
-                if re.search(r"--with-cm\b", cfg) and "--without-cm" not in cfg:
+                if (re.search(r"--with-cm\b", cfg) and
+                        "--without-cm" not in cfg):
                     variant_list.append("+cm")
                 elif "--without-cm" in cfg:
                     variant_list.append("~cm")
 
             # MLX5 (flag name changed in v1.18: --with-mlx5-dv → --with-mlx5)
             if version in ver(":1.17"):
-                if re.search(r"--with-mlx5-dv(?!=no)", cfg) and "--without-mlx5-dv" not in cfg:
+                if (re.search(r"--with-mlx5-dv(?!=no)", cfg) and
+                        "--without-mlx5-dv" not in cfg):
                     variant_list.append("+mlx5_dv")
                 else:
                     variant_list.append("~mlx5_dv")
             else:
-                if re.search(r"--with-mlx5(?!=no)", cfg) and "--without-mlx5" not in cfg:
+                if (re.search(r"--with-mlx5(?!=no)", cfg) and
+                        "--without-mlx5" not in cfg):
                     variant_list.append("+mlx5_dv")
                 else:
                     variant_list.append("~mlx5_dv")
@@ -255,7 +258,8 @@ class Ucx(AutotoolsPackage, CudaPackage):
                 elif "--without-fuse3" in cfg:
                     variant_list.append("~vfs")
 
-            # backtrace_detail (version-specific: --enable-backtrace-detail → --with-bfd)
+            # backtrace_detail (version-specific: --enable-backtrace-detail
+            # → --with-bfd)
             if version in ver(":1.11"):
                 if re.search(r"--enable-backtrace-detail\b", cfg):
                     variant_list.append("+backtrace_detail")
@@ -273,7 +277,8 @@ class Ucx(AutotoolsPackage, CudaPackage):
             elif "--without-ib-hw-tm" in cfg:
                 variant_list.append("~ib_hw_tm")
 
-            # logging (detect actual build config; UCX default=enabled, Spack variant default=False)
+            # logging: detect actual build config
+            # (UCX default=enabled, Spack variant default=False)
             if "--disable-logging" in cfg:
                 variant_list.append("~logging")
             else:
--- a/repos/spack_repo/builtin/packages/ucx/package.py
+++ b/repos/spack_repo/builtin/packages/ucx/package.py
@@ -1,6 +1,7 @@
 # Copyright Spack Project Developers. See COPYRIGHT file for details.
 #
 # SPDX-License-Identifier: (Apache-2.0 OR MIT)
+import re
 import shutil
 
 from spack_repo.builtin.build_systems.autotools import AutotoolsPackage
@@ -20,6 +21,8 @@ class Ucx(AutotoolsPackage, CudaPackage):
     maintainers("hppritcha")
 
     license("BSD-3-Clause")
+
+    executables = ["^ucx_info$"]
 
     version("master", branch="master", submodules=True)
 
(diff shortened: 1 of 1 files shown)


hppritcha (approved):
I am assuming interested parties have verified this feature.  I don't see any obvious issues so approving.

mathomp4:
@dhairyashilRG I just saw this. Is your update controlled by a switch/variant? That is, what if I am on a machine with a UCX install (so `ucx_info` is around), but I don't want it used. Will this PR make it so that it'll always just use the one on the system? 

Or does this just use the internal one to determine variants? 

Just not sure

mathomp4:
Update. I had claude analyze your pr and give me a "standalone" script I could use to test the detection. I tested the detection approach on a system UCX 1.20.0 install on an NVIDIA GraceHopper system and noticed a few cases where relying only on `ucx_info -v` configure flags and `ucx_info -b` defines appears to under-detect enabled capabilities.

For example, this UCX was configured with:

```console
--enable-cma --with-cuda --with-gdrcopy --with-verbs --with-mlx5 --with-rdmacm --with-xpmem
```

and `ucx_info -d` reports available transports/devices including:

```console
cma
rc_verbs
ud_verbs
dc_mlx5
rc_mlx5
ud_mlx5
cuda_copy
cuda_ipc
rdmacm
xpmem
```

However, detection based only on the current configure/define logic would not necessarily mark `+cma`, `+rc`, `+ud`, or `+dc`, because this UCX configure line does not include explicit `--with-rc`, `--with-ud`, or `--with-dc`, and `ucx_info -b` did not expose `HAVE_CMA=1`.

Would it make sense to also parse `ucx_info -d` as a fallback/confirmation source? For example:

- `Transport: rc_verbs` or `Transport: rc_mlx5` => `+rc`
- `Transport: ud_verbs` or `Transport: ud_mlx5` => `+ud`
- `Transport: dc_mlx5` => `+dc`
- `Memory domain: cma` or `Transport: cma` => `+cma`
- `Memory domain: cuda_cpy` / `Transport: cuda_copy` or `Memory domain: cuda_ipc` / `Transport: cuda_ipc` => `+cuda`
- `Memory domain: mlx5_*` or `Transport: *_mlx5` => evidence for `+mlx5_dv`
- `Connection manager: rdmacm` => evidence for `+rdmacm`
- `Memory domain: xpmem` or `Transport: xpmem` => `+xpmem`

This may make external detection more robust for vendor/system UCX builds where configure flags do not explicitly list every enabled transport, but the installed UCX clearly exposes them through `ucx_info -d`.

I realize `ucx_info -d` can be node-dependent, so maybe it should be treated as a fallback or corroborating signal rather than the sole source of truth. But for external detection, it seems useful to capture capabilities that are actually present in the installed UCX even when the configure line is not granular enough.

bernhardkaindl:
@dhairyashilRG, could you review the review comments?

Setting to draft due to the reivew comments and as there was no response so far.

hppritcha (approved):

dhairyashilRG:
@mathomp4 Thanks for both comments, and sorry for the slow reply.

On your first question: no, this never takes over automatically.  `executables` / `determine_version` / `determine_variants` only run when you explicitly ask for detection, i.e. `spack external find ucx` (or `spack external find --all`). Detection writes entries into `packages.yaml`; nothing consumes them unless they are there. On a machine with `ucx_info` in `PATH` where you don't want the system UCX, just don't run `spack external find ucx` or delete the entry afterwards, or set `buildable: false` under `packages: ucx:` to force Spack to keep building its own. There is no variant gating this because there is nothing to gate: it's opt-in by construction.

On your second comment: you were right, and the cause is worse than it looked. I dug into UCX's configure and the reason vendor builds under-detect is that the configure line is structurally the wrong source:

- `rc`, `ud`, `dc`, `dm`, `ib_hw_tm`, `mlx5` are `enabled by default`
- `rdmacm`, `knem`, `xpmem`, `bfd`, `fuse3`, `cuda`, `rocm` default to guess (auto-detect)

So none of them appear in `# Configured with:` unless a packager explicitly overrode them. My original code could only ever see explicit overrides. On a stock Ubuntu 24.04 UCX 1.16.0 that meant 7 wrong variants (`cma`, `rc`, `ud`, `dc`, `dm`, `ib_hw_tm`, `rdmacm`), which I reproduced before changing anything.

I went with `ucx_info -b` rather than the `ucx_info -d` you suggested. `-d` enumerates *runtime-available devices*, so it depends on the hardware present when detection runs  a UCX built with IB support on a node without an HCA would show nothing, which would trade your false negatives for a new set. `-b` is the installed `config.h`: build-time truth, hardware-independent. It carries everything needed:

- transports and build options as `#define`s (`HAVE_TL_RC`, `HAVE_IBV_DM`, `IBV_HW_TM`,`ENABLE_MT`, ...)
- components as `<component>_MODULES` lists (`uct_MODULES`, `uct_cuda_MODULES`,`uct_ib_MODULES`, `ucm_MODULES`, `ucs_MODULES`)

One trap that shows why the module lists matter: the Ubuntu build reports `HAVE_HIP 1` and `HAVE_HSA_H 1` while `uct_rocm_MODULES` is empty the ROCm headers were found, but the module was not built. Keying `rocm` off the define would produce a false `+rocm`, so module-backed components key off the module lists and not the `HAVE_*` defines.

Also fixed while in here: `cma` was keyed off a `HAVE_CMA` define that UCX never emits (so it was always `~cma`); `rc`/`ud`/`dc`/`dm` used a `\b` match that also matched `--with-rc=no`; `backtrace_detail` now uses `HAVE_DETAILED_BACKTRACE`, which covers both the old `--enable-backtrace-detail` and the newer `--with-bfd` spelling; and `logging` now reads `UCS_MAX_LOG_LEVEL` instead of inferring from the absence of
`--disable-logging`. The version-dependent flag-name branches are gone entirely, since the defines are stable across versions the detection code is now shorter than before.

Verified with real `spack external find ucx` against a distro UCX 1.16.0 and a source-built 1.18.0 configured with explicit negatives:

before: ucx@1.16.0 +assertions ~backtrace_detail ~cma ~debug +logging +mlx5_dv
                       +parameter_checking +thread_multiple +verbs
after:  ucx@1.16.0 +assertions ~backtrace_detail +cma ~cuda +dc ~debug +dm ~gdrcopy
                       +ib_hw_tm ~java ~knem +logging +mlx5_dv +parameter_checking +rc
                       +rdmacm ~rocm +thread_multiple +ud +verbs ~vfs ~xpmem

One ask: I don't have access to a Grace Hopper box, so the CUDA/ROCm/gdrcopy/knem/ xpmem/vfs paths are covered only by synthetic `ucx_info -b` inputs, and I could not test a real 1.20.0. If you still have that system, could you re-run your standalone script against this version? Your 1.20.0 install is exactly the case I can't reproduce, and `ucx_info -b` output from it would be the best confirmation.

dhairyashilRG:
@bernhardkaindl Apologies for the silence. I've now addressed both of @mathomp4's comments. The first was a question about whether detection is opt-in (it is), and the second was a real defect, which I've reworked: detection now reads the build configuration from `ucx_info -b` instead of parsing the configure line, since most UCX components are enabled-by-default or auto-detected and so never appear there. Details and before/after output in my reply above.

The branch is rebased onto current develop and squashed to a single commit. Taking it out of draft now so CI can run.

bernhardkaindl:
Hi @mathomp4, because of your comment and @dhairyashilRG, detailed answer to it, I invited you for review.

I saw an ask for you in the comment:

> One ask: I don't have access to a Grace Hopper box, so the CUDA/ROCm/gdrcopy/knem/ xpmem/vfs paths are covered only by synthetic ucx_info -b inputs, and I could not test a real 1.20.0. If you still have that system, could you re-run your standalone script against this version? Your 1.20.0 install is exactly the case I can't reproduce, and ucx_info -b output from it would be the best confirmation.

mathomp4:
@dhairyashilRG Sorry. I missed your ask above. I'm attaching the `ucx_info -b` and `ucx_info -d` output from my grace hopper.

Hopefully they can give final confirmation.

[ucx_info_b.txt](https://github.com/user-attachments/files/30555916/ucx_info_b.txt)
[ucx_info_d.txt](https://github.com/user-attachments/files/30555917/ucx_info_d.txt)

dhairyashilRG:
@mathomp4 Perfect, thank you — this is the exact confirmation I was after. I ran the reworked detection against your real `ucx_info -b` and it produces:

    ~assertions +backtrace_detail +cma +cuda +dc ~debug +dm +gdrcopy +ib_hw_tm
    ~java ~knem ~logging +mlx5_dv ~parameter_checking +rc +rdmacm ~rocm
    +thread_multiple +ud +verbs ~vfs +xpmem

All eight variants you originally flagged as under-detected are now correct:
`+cma +rc +ud +dc +cuda +gdrcopy +mlx5_dv +rdmacm +xpmem`, even though this build's configure line has no explicit `--with-rc/--with-ud/--with-dc` and emits no `HAVE_CMA`. They're picked up from the `*_MODULES` lists in `-b` (`uct_MODULES = :cuda:ib:rdmacm:cma:xpmem`, `uct_cuda_MODULES = :gdrcopy`, `uct_ib_MODULES = :mlx5`, `ucm_MODULES = :cuda`), and they line up one-to-one with the transports/MDs in your `ucx_info -d` (`rc_verbs`/`rc_mlx5`, `ud_*`, `dc_mlx5`, `cuda_copy`/`cuda_ipc`, `gdr_copy`, `xpmem`, `rdmacm`).

`~rocm` is also correct despite the auto-detect: `uct_rocm_MODULES` is empty, so no false `+rocm`. This was exactly the aarch64 / 1.20.0 case I couldn't reproduce locally, so your output is the final validation. Thanks for digging it out.

Nothing further needed on the code from this, it's confirmed on real hardware for the platform that started the whole thread.

mathomp4 (approved):

mathomp4:
@dhairyashilRG Great job! Thanks for doing this. I've approved as this seems good to me.

bernhardkaindl:
Hi, from what I can see, the latest improvement appears to be confirmed to work.

@hppritcha, as you're the spack maintainer of ucx,
and I'm not sure if you already reviewed the latest changes,
I think should ping you in case you want to review it again.

As far as I understood from when I heard last about it, we should provide for about 5 days for your review (normally).

Just as an idea, not a review comment, and of course only if @hppritcha approves it:
@mathomp4, would you be interested to be added as a 2nd maintainer of `ucx`? (as a spack member, you'd be invited for review of ucx changes by @spackbot)
And @dhairyashilRG, as you added this new capability to the recipe, would you be interested in the list of reviewers? (as you're not a Spack member so far, you'd just be listed in the maintainers property, so when reviewing ucx PRs, maintainers would currently only consult you if needed)

hppritcha:
well i keep approving then someone does some other changes then i have to review again.

hppritcha (approved):

bernhardkaindl:
> well i keep approving then someone does some other changes then i have to review again.

Hi @hppritcha, I'm sorry, I wasn't meant this way. I just remembered that I was asked to provide time for maintainers to review.

A hppritcha has approved, and the PR has already passed all GitLab CI pipelines, I think there is nothing standing in the way to merge this PR.

@mathomp4 and @dhairyashilRG: If you want, you can submit a new PR regarding my proposals to add you as maintianers.

bernhardkaindl (approved):

dhairyashilRG:
Thanks @bernhardkaindl, and thanks for the review throughout. Yes, I'd like to be listed in the `ucx` maintainers property. Appreciate the help getting it over the line. 

@mathomp4 can open that PR for both of us or I can do it if he is fine with it.

bernhardkaindl:
@dhairyashilRG, merged!

PR #4294, opened 2026-04-14
adios2: add min version numbers to yaml-cpp and pugixml dependencies

mpokorny:
Minimum version numbers on some dependencies are missing. I encountered a build problem when I added `adios2` to an environment with a pre-existing `yaml-cpp` with version 0.6.3. The minimum version number on `yaml-cpp` appears to have been added with `adios2` v2.9. While looking at the `CMakeLists.txt` file, I noticed also that `pugixml` was missing a minimum version number in the Spack package.


vicentebolea (approved):
checked the versions. looks good; Thanks!

alalazo (approved):

PR #4299, opened 2026-04-14
grpc: add patch for missing includes

wspear:
Add patches for GCC 13+ and Clang 15+ compatibility.

Targets top level port_platform.h header which gets most of the missing includes where they need to go. The only outlier is glob.cc which gets its own filter call.


wspear:
@spackbot fix style

alecbcs (approved):
Looks good to me. Thanks @wspear!

PR #4301, opened 2026-04-15
gflags: improve spack build isolation

etiennemlb:
Gflags will try to write to the home directory of the user. 
https://github.com/gflags/gflags/issues/386

This breaks spack containment and is not needed due to spack handles dependencies.

### DIFF
--- a/repos/spack_repo/builtin/packages/gflags/package.py
+++ b/repos/spack_repo/builtin/packages/gflags/package.py
@@ -28,6 +28,6 @@ class Gflags(CMakePackage):
 
     def cmake_args(self):
         return [
-            self.define("BUILD_SHARED_LIBS", True), 
+            self.define("BUILD_SHARED_LIBS", True),
             self.define("REGISTER_INSTALL_PREFIX", False),
         ]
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/gflags/package.py
```diff
@@ -27,4 +27,4 @@ class Gflags(CMakePackage):
     depends_on("cmake@2.8.12:", type="build")
 
     def cmake_args(self):
-        return ["-DBUILD_SHARED_LIBS=ON"]
+        return ["-DBUILD_SHARED_LIBS=ON", "-DREGISTER_INSTALL_PREFIX=OFF"]
```
alecbcs:
```suggestion
        return [
            self.define("BUILD_SHARED_LIBS", True), 
            self.define("REGISTER_INSTALL_PREFIX", False),
        ]
```

Quick suggestion to use `self.define()` here to clean up the syntax a bit

alecbcs (dismissed):
Quick optional suggestion here. Otherwise looks good to me. Thanks @etiennemlb!

repos/spack_repo/builtin/packages/gflags/package.py
etiennemlb: Aggred, while we are at it

alecbcs:
@spackbot fix style

alecbcs (approved):
Looks good to me. Thanks!

PR #4302, opened 2026-04-15
proteowizard: new package

w8jcik:


repos/spack_repo/builtin/packages/proteowizard/package.py
```diff
@@ -0,0 +1,59 @@
...
+
+
+class Proteowizard(Package):
+    """ProteoWizard: A set of software libraries and tools for rapid development of mass
+    spectrometry and proteomic data analysis software."""
+
+    homepage = "https://proteowizard.sourceforge.io/"
+    git = "https://github.com/ProteoWizard/pwiz.git"
+
+    version("3.0.26102-0783ec5", commit="0783ec56810626af7888a80d7b32fb3e47d02d52")
```
hainest: Please add a SHA256.

hainest (changes requested):

repos/spack_repo/builtin/packages/proteowizard/package.py
w8jcik: As far as I know, it is not possible to specify checksum for git commit-based versions in Spack. Did something change recently in this regard?

Authors of proteowizard release commits, they don't mark versions in any other way, see attached screenshot  

image

w8jcik: Ok, I see that one can get a tarball for a commit in GitHub, like in  
http://github.com/ProteoWizard/pwiz/tarball/0783ec56810626af7888a80d7b32fb3e47d02d52  

I will change it.

w8jcik: Added checksum as suggested.

hainest: > As far as I know, it is not possible to specify checksum for git commit-based versions in Spack. Did something change recently in this regard?

No, I forgot about that limitation. Thanks for updating it to use a tarball.

hainest (approved):
Confirmed SHAs.

w8jcik:
Thank you for merging.

PR #4305, opened 2026-04-15
py-id: add v1.6.1

manuelakuhn:
https://github.com/di/id/releases/tag/v1.6.1

### DIFF
--- a/repos/spack_repo/builtin/packages/py_id/package.py
+++ b/repos/spack_repo/builtin/packages/py_id/package.py
@@ -12,11 +12,20 @@ class PyId(PythonPackage):
 
     homepage = "https://pypi.org/project/id/"
     pypi = "id/id-1.5.0.tar.gz"
+    git = "https://github.com/di/id.git"
 
     license("Apache-2.0", checked_by="RobertMaaskant")
 
+    version("1.6.1", sha256="d0732d624fb46fd4e7bc4e5152f00214450953b9e772c182c1c22964def1a069")
     version("1.5.0", sha256="292cb8a49eacbbdbce97244f47a97b4c62540169c976552e497fd57df0734c1d")
 
     depends_on("py-flit-core@3.2:3", type="build")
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/py_id/package.py
```diff
@@ -12,11 +12,20 @@ class PyId(PythonPackage):
 
     homepage = "https://pypi.org/project/id/"
     pypi = "id/id-1.5.0.tar.gz"
+    git = "https://github.com/di/id.git"
 
     license("Apache-2.0", checked_by="RobertMaaskant")
 
+    version("1.6.1", sha256="d0732d624fb46fd4e7bc4e5152f00214450953b9e772c182c1c22964def1a069")
     version("1.5.0", sha256="292cb8a49eacbbdbce97244f47a97b4c62540169c976552e497fd57df0734c1d")
```
alecbcs, line 20:
```suggestion
    version("main", branch="main")
    version("1.6.1", sha256="d0732d624fb46fd4e7bc4e5152f00214450953b9e772c182c1c22964def1a069")
version("1.5.0", sha256="292cb8a49eacbbdbce97244f47a97b4c62540169c976552e497fd57df0734c1d")
```

Do we want to add a main version here to track the git repository since that was added too?

manuelakuhn: I did not plan to. I only added the git repository  to have this information there in the future because it makes updating the package easier.
But if you think it is needed, we can add it.

alecbcs: Let's skip it for now. I typically add a main version when adding a git attribute, but I think it's fairly unlikely that someone will want to develop this package via `spack develop`. Just wanted to make sure you hadn't accidentally left it out when you meant to add it.

alecbcs (approved):

PR #4317, opened 2026-04-15
pcl: add v1.15.1, bound eigen dep for now

snehring:
version bump along with an upper bound on eigen version, at least for now.


RikkiButler20 (approved):
LGTM

PR #4319, opened 2026-04-15
beast2: correct bound on template dir name

snehring:
The directory is called "templates" up to 2.6.7.


alecbcs (approved):
Looks good to me. Thanks for finding this @snehring!

PR #4320, opened 2026-04-15
kraken2: adding v2.17.1, cmake implementation

snehring:
Adding 2.17.1

They seem to be putting effort into cmake, but they don't actually have an install target defined yet, so we've still got a bit of a custom install process.

### DIFF
--- a/repos/spack_repo/builtin/packages/kraken2/package.py
+++ b/repos/spack_repo/builtin/packages/kraken2/package.py
@@ -5,12 +5,13 @@
 import glob
 import os
 
+from spack_repo.builtin.build_systems.cmake import CMakeBuilder, CMakePackage
 from spack_repo.builtin.build_systems.generic import Package
 
 from spack.package import *
 
 
-class Kraken2(Package):
+class Kraken2(Package, CMakePackage):
     """Kraken2 is a system for assigning taxonomic labels to short DNA
     sequences, usually obtained through metagenomic studies."""
 
@@ -21,6 +22,7 @@ class Kraken2(Package):
 
     license("MIT")
 
+    version("2.17.1", sha256="4dc64ead045b5ae9180731c260046aa37b6642244be085a9ba9b15db78ab442d")
     version("2.1.2", sha256="e5f431e8bc3d5493a79e1d8125f4aacbad24f9ea2cc9657b66da06a32bef6ff3")
     version("2.1.1", sha256="8f3e928cdb32b9e8e6f55b44703d1557b2a5fc3f30f63e8d16e465e19a81dee4")
     version(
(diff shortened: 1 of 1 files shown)


snehring:
@spackbot re-run pipeline

repos/spack_repo/builtin/packages/kraken2/package.py
```diff
@@ -47,3 +56,21 @@ def install(self, spec, prefix):
...
+        with working_dir(join_path(self.build_directory, "src")):
+            files = [
+                "estimate_capacity",
+                "build_db",
+                "classify",
+                "dump_table",
+                "lookup_accession_numbers",
+                "k2mask",
+                "blast_to_fasta",
+                "libtax.so",
```
alecbcs: Are we meaning to install this shared object file into the prefix bin? Or should this be going in lib?

alecbcs (commented):
One quick question, otherwise looks good to me.

repos/spack_repo/builtin/packages/kraken2/package.py
snehring: Since they've not defined an actual install target this basically copies the result of the original recipe, everything in bin. I do think it'd be better to have the lib in a lib directory, but that's not what the upstream package itself is doing.

snehring:
@alecbcs do you need any changes made with this?

repos/spack_repo/builtin/packages/kraken2/package.py
alecbcs: I'm alright merging as it since this is how the package has done it in the past. I think it'd be cool if we could install everything into `prefix.lib` instead and then create symlinks into `prefix.bin` for the couple of things that are actually user facing. If you don't have time for that I can look at doing it in a follow up PR.

For reference this lib + symlink to bin method appears to be how Debian structure's their implementation of the package. (https://packages.debian.org/trixie/amd64/kraken2/filelist)

alecbcs (dismissed):
Looks good to me. Quick suggestion to swap this to `prefix.lib` + symlinks to `prefix.bin` but I can handle that in a follow up PR if you'd prefer to merge this as is.

snehring:
Let me poke this a bit more

snehring:
Realized I forgot the entire scripts directory.

I'm also making cmake for new versions only. A few of the artifacts aren't actually present in the older versions.

Also the thing that made me look at this package in the first place (gmake is needed or the new installer fails) is fixed.

snehring:
I did move the lib into the prefix.lib dir. @alecbcs let me know if you'd like further changes.

alecbcs (approved):
Thanks @snehring!

PR #4324, opened 2026-04-16
cyrus-sasl: don't install to /Library on macOS by default

cmelone:
This package will attempt to install to a privileged location by default. I can make this a variant but I'm not sure if people are really installing Spack packages with sudo.

### DIFF
--- a/repos/spack_repo/builtin/packages/cyrus_sasl/package.py
+++ b/repos/spack_repo/builtin/packages/cyrus_sasl/package.py
@@ -69,9 +69,7 @@ class CyrusSasl(AutotoolsPackage):
     depends_on("libxcrypt", type="link")
     depends_on("krb5", type="link")
 
+    @when("platform=darwin")
     def configure_args(self):
-        args = []
         # avoid installing to /Library/Frameworks/SASL2.framework
-        if self.spec.satisfies("platform=darwin"):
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/cyrus_sasl/package.py
```diff
@@ -68,3 +68,10 @@ class CyrusSasl(AutotoolsPackage):
     depends_on("openssl", type="link")
     depends_on("libxcrypt", type="link")
     depends_on("krb5", type="link")
+
+    def configure_args(self):
+        args = []
+        # avoid installing to /Library/Frameworks/SASL2.framework
+        if self.spec.satisfies("platform=darwin"):
+            args.append("--disable-macos-framework")
+        return args
```
alecbcs:
```suggestion
    @when("platform=darwin")
    def configure_args(self):
        # avoid installing to /Library/Frameworks/SASL2.framework
        return ["--disable-macos-framework"]
```

Potentially a slightly cleaner way to write the same thing. Up to you if you'd like to swap it or merge it as is.

alecbcs (dismissed):
One optional suggestion, otherwise looks good to me.

alecbcs (approved):
Looks good to me. Thanks!

PR #4326, opened 2026-04-16
py-python-xmp-toolkit: add v2.1.0

manuelakuhn:
https://github.com/python-xmp-toolkit/python-xmp-toolkit/releases/tag/v2.1.0

This fixes the changed url on pypi and also that exempi is found. Without the latter, the build was fine but running the examples from the homepage resulted in `libxmp.ExempiLoadError: Exempi library not found.`


RikkiButler20 (approved):
LGTM

PR #4334, opened 2026-04-16
costa@2.3.1

simonpintarelli:
new costa version

### DIFF
--- a/repos/spack_repo/builtin/packages/costa/package.py
+++ b/repos/spack_repo/builtin/packages/costa/package.py
@@ -50,7 +50,7 @@ class Costa(CMakePackage):
     depends_on("cxxopts", when="+apps")
     depends_on("cxxopts", when="+tests")
     depends_on("semiprof", when="+profiling")
-    depends_on("googletest", when="+tests")
+    depends_on("googletest", when="@2.3.1: +tests")
 
     def url_for_version(self, version):
         if version == Version("2.0"):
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/costa/package.py
```diff
@@ -49,6 +50,7 @@ class Costa(CMakePackage):
     depends_on("cxxopts", when="+apps")
     depends_on("cxxopts", when="+tests")
     depends_on("semiprof", when="+profiling")
+    depends_on("googletest", when="@2.3.1: +tests")
```
RMeli, line 53: Is this actually a dependency? Looks like COSTA is unconditionally fetching `52eb8108c5bdec04579160ae17225d66034bd723` using CMake.

I think the best solution would be to have something like `COSTA_USE_EXTERNAL_GTEST` in CMake, but that would require another release etc, so maybe just removing this dependency is fine (since it is handled by COSTA internally)?

simonpintarelli: Yes, before gtest.h was contained in the repo. Now it uses `FetchContent`. But I thought the clean way is to explicitly list the dependency in spack if it's required. Gtest will be fetched automatically if it's built manually and gtest isn't present.

mtaillefumier:
we can safely include the google test dependency explicitly and I would support this rather than depending on `fetch_content`. Either method will work though.

mtaillefumier (approved):
LGTM

repos/spack_repo/builtin/packages/costa/package.py
RMeli: Yes, sorry. I thought it was fetching unconditionally given the `GIT_TAG` (I've been bitten by something like that before with `FetchContent`), but looks like it is indeed using `find_package` beforehand so adding an explicit dependency should be the right thing to do.

RMeli (approved):

PR #4336, opened 2026-04-16
sage: add new package

w8jcik:


alecbcs:
@w8jcik I would recommend swapping this to a `CargoPackage` and that'll handle the build logic for you. If you'd like an example package to pull from I'd recommend looking at [`typos`](https://github.com/spack/spack-packages/blob/develop/repos/spack_repo/builtin/packages/typos/package.py).

w8jcik:
Changed as you suggested.

alecbcs (approved):
Looks good to me. Thanks!

PR #4337, opened 2026-04-16
panda: adding c dep

snehring:
cmake checks for c


alecbcs (approved):

PR #4339, opened 2026-04-16
scallop: adding c dep

snehring:
needs a c compiler


alecbcs (approved):
Confirmed that the configure script checks for a C compiler. Thanks for the fix @snehring!

PR #4340, opened 2026-04-16
OSU MicroBenchmark: Fix compilation without the NVIDIA driver

RemiLacroix-IDRIS:
Using the libcuda stub for execution doesn't make sense and won't work. We want it at link time.


RemiLacroix-IDRIS:
Any thoughts on this?

alecbcs (approved):
Looks good to me. Thanks @RemiLacroix-IDRIS!

Merging without hearing back from maintainers in > 1 month.

PR #4346, opened 2026-04-16
kokkos: add version 4.7.04

JBludau:
- [x] add Kokkos-Kernels @lucbv


lucbv (approved):

repos/spack_repo/builtin/packages/kokkos/package.py
rbberger: This should now add the necessary mapping for `ARMV9_GRACE`
```
"armv9.0a":  ("ARMV9_GRACE", "@4.7.04:,@5.1:")
```

JBludau: added in https://github.com/spack/spack-packages/pull/4346/commits/f1404ad49a7d0421a8a4d507f6a44d4438e07765

tpadioleau: The syntax does not seem valid

repos/spack_repo/builtin/packages/kokkos/package.py
```diff
@@ -162,6 +163,7 @@ class Kokkos(CMakePackage, CudaPackage, ROCmPackage):
     spack_micro_arch_map = {
         "armv8.1a": ("ARMV81", None),
         "armv8.4a": ("ARMV84", "@4.7.00:"),
+        "neoverse_v2": ("ARMV:w9_GRACE", "@4.7.04:,%5.1:"),
```
tpadioleau: What about

```suggestion
        "neoverse_v2": ("ARMV:w9_GRACE", "@4.7.04:4,5.1:"),
```
I think we want to exclude 5.0.x ?

tpadioleau (approved):
Looks ok to me

repos/spack_repo/builtin/packages/kokkos/package.py
rbberger: The `%` is wrong. In both cases it's a `@`.

rbberger: Looks like the actual syntax is no second `@`. So `@4.7.04:,5.1:`. Assuming it's also not supported in `5.0.0`, `5.0.1`, `5.0.2`, then it has to be `@4.7.04:4,5.1:`.

rbberger (approved):

PR #4348, opened 2026-04-16
kmergenie: adding c dep

snehring:
needs a c compiler


alecbcs (approved):

PR #4350, opened 2026-04-16
rocm: improve external detection

afzpatel:
This PR adds and fixes external package detection in the ROCm recipes.  If the ROCm version cannot be determined from the library name or executable, the process will attempt to locate a path in the format `/opt/rocm-<version>/` to identify the version. As a secondary option, it will also check `rocm_version.h` for the ROCm version.


afzpatel:
@spackbot fix style

repos/spack_repo/builtin/packages/amdsmi/package.py
```diff
@@ -90,3 +79,9 @@ def cmake_args(self):
         args.append(self.define("BUILD_TESTS", "ON"))
         args.append("-DCMAKE_INSTALL_LIBDIR=lib")
         return args
+
+    @classmethod
+    def determine_version(cls, exe):
+        output = Executable(exe)("version", output=str, error=str)
```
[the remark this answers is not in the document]
afzpatel: no failure detection required since it will just fail to detect the spec.

afzpatel:
@spackbot rerun pipeline

renjithravindrankannath (approved):

repos/spack_repo/builtin/packages/hipify_clang/package.py
```diff
@@ -91,9 +92,30 @@ class HipifyClang(CMakePackage):
...
-        output = Executable(exe)("--version", output=str, error=str)
-        match = re.search(r"HIP version (\S+)", output)
-        return match.group(1) if match else "None"
+        match = re.search(r"rocm-(\d+\.\d+\.\d+)", exe)
+        if match:
+            return match.group(1)
+        return cls.version_from_rocm_version_h(exe)
+
+    @classmethod
+    def version_from_rocm_version_h(cls, lib):
```
scheibelp: This is repeated fairly frequently: it would be good to create a `ROCmLibrary` build_system class that packages inherit (note: I am not suggesting merging this into the existing `ROCmPackage` class, but you could but it in a class inside of the `rocm.py` module within `build_systems`).

repos/spack_repo/builtin/packages/hsa_amd_aqlprofile/package.py
```diff
@@ -44,3 +46,17 @@ def root_cmakelists_dir(self):
...
+    @classmethod
+    def determine_version(cls, lib):
+        match = re.search(r"lib\S*\.so\.\d+\.\d+\.(\d)(\d\d)(\d\d)", lib)
+        if match:
+            ver = "{0}.{1}.{2}".format(
+                int(match.group(1)), int(match.group(2)), int(match.group(3))
+            )
+            major = int(ver.split(".")[0])
+            if major < 7:
+                ver = None
```
scheibelp, line 59: Question: I see this happen in several packages in this PR. Generally returning `None` is to indicate that a version wasn't detected, and I think the effect in this case would be that `hsa-amd-aqlprofile@7:` would never be detected as external - is that what you wanted to happen?

If so, I'm curious why you want to avoid detecting externals for versions 7 and later.

repos/spack_repo/builtin/packages/hipfort/package.py
```diff
@@ -18,6 +21,8 @@ class Hipfort(CMakePackage):
     license("MIT")
 
     maintainers("cgmb", "srekolam", "renjithravindrankannath", "afzpatel")
+    libraries = ["libhipfort-amdgcn.a", "libhipfort-nvptx.a"]
```
scheibelp, line 22: Question: Are there never shared versions? This should be fine to do, but FYI would never detect an instance of `libhipfort-amdgcn.so` as an external instance of `hipfort`.

scheibelp (changes requested):
I have a request and a couple questions

Also I looked into it and we don't have a good mechanism for detecting header-only packages at the moment.

repos/spack_repo/builtin/packages/hsa_amd_aqlprofile/package.py
afzpatel: It's the opposite, we don't want `hsa-amd-aqlprofile@:6` to be detected. We changed `aqlprofile` to `hsa-amd-aqlprofile` in ROCm 7. So if the version is ROCm 6 or below we want it to be `aqlprofile` not  `hsa-amd-aqlprofile`

repos/spack_repo/builtin/packages/hipfort/package.py
afzpatel: no there aren't any shared libraries for `hipfort`.  I can see `spack external find hipfort` can still detect the static libraries

afzpatel:
@spackbot fix style

afzpatel:
closing this PR since spackbot edited files outside of my PR and pinged a bunch of maintainers. I'll open a new PR for this.

PR #4351, opened 2026-04-16
adios2: add new release

vicentebolea:


repos/spack_repo/builtin/packages/adios2/package.py
albestro:
```python
conflicts("%cuda@13:", when="@:2.11 +cuda")
```

?

albestro (changes requested):

repos/spack_repo/builtin/packages/adios2/package.py
vicentebolea: yep thanks we need that, can you add it as a suggestion?

albestro: I tried really hard before commenting...but the new github interface was not cooperating (it was showing it as removing a blank line and add the new suggested one) 😅

vicentebolea: done!

albestro (approved):

repos/spack_repo/builtin/packages/adios2/package.py
```diff
@@ -272,6 +266,15 @@ class Adios2(CMakePackage, CudaPackage, ROCmPackage):
...
     )
 
+    # https://github.com/ornladios/ADIOS2/pull/5006
+    # Using a diff rather than patch since the commit is a git subtree commit which does not play
+    # well with the github .patch URL param
+    patch(
+        "https://github.com/ornladios/ADIOS2/compare/98c51cc2207fd178d2f84f493d19710cf21f84c1^1...98c51cc2207fd178d2f84f493d19710cf21f84c1.diff?full_index=1",
+        sha256="0fe8ecf75eabf975caf5de447ac34084b574f78e73cf83cf158a9e58b692f2e3",
+        when="@2.12.0",
+    )
```
albestro, line 276: Just one comment, because I don't understand the problem.

Why can't we do something like this?

```suggestion
    patch(
        "https://github.com/ornladios/ADIOS2/compare/98c51cc2207fd178d2f84f493d19710cf21f84c1.patch?full_index=1",
        sha256="0fe8ecf75eabf975caf5de447ac34084b574f78e73cf83cf158a9e58b692f2e3",
        when="@2.12.0",
    )
```

vicentebolea: The commit patch since it is a git-subtree commit describes paths withing the git-subtree, this in turn results in failing to apply the path in the root of the repo (since it is missing the prefix to the dir)

PR #4352, opened 2026-04-16
truchas: add v25.11-v26.04

zjibben:
Adds Truchas 25.11, 25.12, 26.02, and 26.04, with the new dependency on fvtkhdf as of 26.04.


RikkiButler20 (approved):
LGTM

PR #4353, opened 2026-04-16
prrte: add patch for hetero nodes, fix autoreconf, fix runtime env

Chrismarsh:
- This adds a patch that resolves a segfault on nodes with substantial NUMA heterogeneity https://github.com/openpmix/prrte/pull/2430
- Ensures the `prrte` prefix is added to the env `PRTE_PREFIX` (the single R is correct). This ensures that a FQP to mpirun isn't required when run outside of a scheduler env
- The autoreconf wasn't always triggering, and would result in trying to use a system automake. This change ensures it is always recreated to match whatever version of `automake` spack has resolved

### DIFF
--- a/repos/spack_repo/builtin/packages/prrte/package.py
+++ b/repos/spack_repo/builtin/packages/prrte/package.py
@@ -2,13 +2,9 @@
 #
 # SPDX-License-Identifier: (Apache-2.0 OR MIT)
 
-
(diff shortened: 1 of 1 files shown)


Chrismarsh:
@spackbot fix style

becker33 (approved):

PR #4357, opened 2026-04-17
parmgridgen: GCC 14: build issues

etiennemlb:
Since GCC 14, some flags are needed to make this software compiler. Indeed, it is not standard conformant.

### DIFF
--- a/repos/spack_repo/builtin/packages/parmgridgen/package.py
+++ b/repos/spack_repo/builtin/packages/parmgridgen/package.py
@@ -29,7 +29,7 @@ class Parmgridgen(Package):
     depends_on("gmake", type="build")
 
     def flag_handler(self, name, flags):
-        if name in ("cflags",):
+        if name == "cflags":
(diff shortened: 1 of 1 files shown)


etiennemlb:
@spackbot fix style

repos/spack_repo/builtin/packages/parmgridgen/package.py
```diff
@@ -28,16 +28,16 @@ class Parmgridgen(Package):
     depends_on("mpi", when="+mpi")
     depends_on("gmake", type="build")
 
+    def flag_handler(self, name, flags):
+        if name in ("cflags",):
```
alecbcs:
```suggestion
        if name == "cflags":
```

Since we've only got the one value right now I'd recommend to swap this to a simple string comparison.

alecbcs (dismissed):
One quick nitpick, otherwise looks good to me. Thanks @etiennemlb

alecbcs (approved):
Looks good to me. Thanks!

PR #4359, opened 2026-04-17
py-numexpr: add v2.14.1

manuelakuhn:
https://github.com/pydata/numexpr/releases/tag/v2.14.1

Also switched sources to PyPI and checked that all checksums are still correct (all versions build fine for me).


RikkiButler20 (approved):
LGTM

PR #4368, opened 2026-04-17
precice: add 3.4.1 with kokkos-kernel mapping support

fsimonis:
This PR adds the version 3.4.1 with support for kokkos-kernel-based mappings to preCICE.

It also adds variants for `openmp` and `sycl` backends.

FYI @davidscn

### DIFF
--- a/repos/spack_repo/builtin/packages/precice/package.py
+++ b/repos/spack_repo/builtin/packages/precice/package.py
@@ -3,11 +3,13 @@
 # SPDX-License-Identifier: (Apache-2.0 OR MIT)
 
 from spack_repo.builtin.build_systems.cmake import CMakePackage
+from spack_repo.builtin.build_systems.cuda import CudaPackage
+from spack_repo.builtin.build_systems.rocm import ROCmPackage
 
 from spack.package import *
 
 
-class Precice(CMakePackage):
+class Precice(CMakePackage, CudaPackage, ROCmPackage):
     """preCICE (Precise Code Interaction Coupling Environment) is a
     coupling library for partitioned multi-physics simulations.
     Partitioned means that preCICE couples existing programs (solvers)
@@ -59,6 +61,24 @@ class Precice(CMakePackage):
     variant("petsc", default=True, description="Enable PETSc support")
     variant("python", default=False, description="Enable Python support", when="@2:")
     variant("shared", default=True, description="Build shared libraries")
+    variant(
+        "kokkos-kernels",
+        default=False,
+        description="Enable performance-portable mapping support via Kokkos Kernels",
+        when="@3.4.1:",
+    )
+    variant(
+        "sycl",
+        default=False,
+        description="Enable SYCL backend preCICE Mappings.",
+        when="+kokkos-kernels",
+    )
+    variant(
+        "openmp",
+        default=False,
+        description="Enable OpenMP backend preCICE Mappings.",
+        when="+kokkos-kernels",
+    )
 
     depends_on("c", type="build")  # generated
     depends_on("cxx", type="build")  # generated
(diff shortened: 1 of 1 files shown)


fsimonis:
@spackbot fix style

repos/spack_repo/builtin/packages/precice/package.py
```diff
@@ -141,6 +149,18 @@ class Precice(CMakePackage):
...
+    with when("+kokkos-kernels"):
+        depends_on("kokkos")
+        depends_on("kokkos-kernels")
+        depends_on("kokkos+sycl", when="+sycl")
+        depends_on("kokkos+openmp", when="+openmp")
+        depends_on("kokkos+rocm", when="+rocm")
+        conflicts("amdgpu_target=none", when="+rocm", msg="must select an AMD GPU architecture")
+
+        depends_on("kokkos+cuda+wrapper+cuda_lambda+cuda_constexpr", when="+cuda")
+        depends_on("kokkos-kernels+cuda", when="+cuda")
```
davidscn: I think the last line here can go, because kokkos-kernels is already declared above and should inherit from Kokkos?

fsimonis: Kokkos-kernels depends on Kokkos, and cuda isn't a sticky variant, so we need to propagate this.

davidscn: Maybe that's true from a spack perspective, but Kokkos-kernels cannot simply "ignore" the cuda dependency in CMake.  Also, when adding it for cuda, why not adding it for all other backends as well, i.e., kokkos-kernels+openmp https://github.com/spack/spack-packages/blob/69e7c860e841310076969639534238d1f1bc3ab8/repos/spack_repo/builtin/packages/kokkos_kernels/package.py#L168-L170 ?

What would be quite useful here is to declare that the Kokkos-kernels version should match the Kokkos version.

MakisH (approved):

MakisH (approved):

davidscn:
```
spack install precice@develop +kokkos-kernels+cuda cuda_arch=80 ~petsc~python ^kokkos cuda_arch=80 ^cuda@12.8
```
succeeded for me with GCC 13. It concretizes Kokkos(-Kernels) 4.6 for some reason. I cant enforce the latest kokkos 5 though.

repos/spack_repo/builtin/packages/precice/package.py
```diff
@@ -141,6 +162,17 @@ class Precice(CMakePackage):
         sha256="6a38783eec984a59991f0895d411212e0ba1ebd2ec2c8f53f962df8facbc0344",
     )
 
+    with when("+kokkos-kernels"):
+        depends_on("kokkos")
+        depends_on("kokkos+openmp", when="+openmp")
+        depends_on("kokkos+rocm", when="+rocm")
+        depends_on("kokkos+sycl", when="+sycl")
+        depends_on("kokkos+cuda+wrapper+cuda_lambda+cuda_constexpr", when="+cuda")
```
davidscn:
```suggestion
        depends_on("kokkos+cuda+wrapper+cuda_constexpr", when="+cuda")
```
We have to remove the lambdas
https://github.com/spack/spack-packages/blob/d52b2732787884eb6311e1d75d9352a59e856e87/repos/spack_repo/builtin/packages/kokkos/package.py#L145

not sure this has an impact anhow, according to the changelog this options is always enforced since v4.1 

https://github.com/kokkos/kokkos/blob/develop/CHANGELOG.md#4100-2023-06-16

davidscn (approved):
The CUDA recipe now works as expected

fsimonis:
Sorry for the noise, this took some more back and forth as we ran into issues with specifying the cuda arch using `cuda_arch==N` and various cross interactions between precice kokkos and kokkos-kernels. This is now ready for review.

fsimonis:
I added the newly released preCICE version 3.4.1. The package now refers to an existing version for the newly added variants.

MakisH (approved):

PR #4381, opened 2026-04-19
osmosis: new package

mtpham99:
`Osmosis` is a command line Java application for processing openstreetmap (OSM) data.

No license file in repo, but "Public Domain" is specified in [build.gradle](https://github.com/openstreetmap/osmosis/blob/0.49.2/build.gradle#L99) file. Also listed as "public" in [maven repository](https://mvnrepository.com/artifact/org.openstreetmap.osmosis).

homepage: [https://wiki.openstreetmap.org/wiki/Osmosis](https://wiki.openstreetmap.org/wiki/Osmosis)
github: [https://github.com/openstreetmap/osmosis](https://github.com/openstreetmap/osmosis)

### DIFF
--- a/repos/spack_repo/builtin/packages/osmosis/package.py
+++ b/repos/spack_repo/builtin/packages/osmosis/package.py
@@ -1,0 +1,31 @@
+# Copyright Spack Project Developers. See COPYRIGHT file for details.
+#
+# SPDX-License-Identifier: (Apache-2.0 OR MIT)
+
+
+from spack_repo.builtin.build_systems.generic import Package
+
+from spack.package import *
+
+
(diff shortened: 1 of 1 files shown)


hainest:
SHA confirmed. Do you want to add yourself as the maintainer (you don't have to)?

mtpham99:
> SHA confirmed. Do you want to add yourself as the maintainer (you don't have to)?

I would prefer not to. I'm not actually too familiar with Java/gradle or OSM tools/data. I'm adding for convenience on behalf of an end user.

The package isn't receiving active development though. Primarily only light/minor maintenance.

hainest (approved):

PR #4389, opened 2026-04-19
dftd4 and dependencies: build shared libs

simonpintarelli:
add variant shared when using cmake as build system.

The static library lack a dependency on gfortran in their cmake config.

### DIFF
--- a/repos/spack_repo/builtin/packages/dftd4/package.py
+++ b/repos/spack_repo/builtin/packages/dftd4/package.py
@@ -42,6 +42,8 @@ class Dftd4(MesonPackage, CMakePackage):
         when="build_system=meson",
         description="Build Python extension module",
     )
+    with when("build_system=cmake"):
+        variant("shared", default=True)
 
     depends_on("c", type="build")
     depends_on("fortran", type="build")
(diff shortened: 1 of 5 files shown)


mtaillefumier (approved):
LGTM

repos/spack_repo/builtin/packages/dftd4/package.py
```diff
@@ -42,6 +42,8 @@ class Dftd4(MesonPackage, CMakePackage):
         when="build_system=meson",
         description="Build Python extension module",
     )
+    with when("build_system=cmake"):
+        variant("shared", default=True, description="Build shared libraries")
```
RMeli, line 46: Since this is turned on unconditionally in the CMake arguments, maybe it would be good to add a conflict with ~shared?

simonpintarelli: Thanks! I forgot to add the `self.define_from_variant` after adding the `variant('shared')`. I've changed it now.

mtaillefumier (approved):

repos/spack_repo/builtin/packages/dftd4/package.py
RMeli: I see. I thought it was done on purpose because of
> The static library lack a dependency on gfortran in their cmake config.

RMeli (approved):

PR #4392, opened 2026-04-20
py-torch: disable SME and KleidiAI on A64FX

otsukay:
### Summary

This PR fixes `py-torch` builds on Fujitsu A64FX by disabling
unsupported AArch64 SME/SME2 and KleidiAI code paths for
`target=a64fx`.

The change does the following:

- disables `XNNPACK_ENABLE_ARM_SME`
- disables `XNNPACK_ENABLE_ARM_SME2`
- disables `USE_KLEIDIAI`
- patches PyTorch's CMake helper so that the XNNPACK SME options are
  forwarded from the environment to CMake

### Motivation

On Fugaku / A64FX, the PyTorch build failed when newer AArch64
XNNPACK / KleidiAI code paths were enabled.

A64FX supports Armv8.2-A SVE, but it does not support Arm SME or
SME2. Therefore, SME/SME2-specific kernels should not be enabled for
`target=a64fx`.

`USE_KLEIDIAI=OFF` is passed through automatically because PyTorch
forwards `USE_*` environment variables to CMake. However,
`XNNPACK_ENABLE_ARM_SME` and `XNNPACK_ENABLE_ARM_SME2` do not use a
`USE_`, `BUILD_`, or `CMAKE_` prefix, so they need to be added to
PyTorch's explicit environment allow list.

### Tested on

Fugaku / Fujitsu A64FX:

### DIFF
--- a/repos/spack_repo/builtin/packages/py_torch/package.py
+++ b/repos/spack_repo/builtin/packages/py_torch/package.py
@@ -467,6 +467,9 @@ class PyTorch(PythonPackage, CudaPackage, ROCmPackage):
     patch("fj-ssl2_1.8.patch", when="@1.8^fujitsu-ssl2")
     patch("fj-ssl2_1.6-1.7.patch", when="@1.6:1.7^fujitsu-ssl2")
 
+    # Disable XNNPACK SME on A64FX
+    patch("disable-arm-sme-kernels-on-a64fx.patch", when="target=a64fx")
+
     # Fix compilation of +distributed~tensorpipe
     # https://github.com/pytorch/pytorch/issues/68002
     patch(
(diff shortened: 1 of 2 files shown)


repos/spack_repo/builtin/packages/py_torch/package.py
```diff
@@ -467,6 +467,9 @@ class PyTorch(PythonPackage, CudaPackage, ROCmPackage):
     patch("fj-ssl2_1.8.patch", when="@1.8^fujitsu-ssl2")
     patch("fj-ssl2_1.6-1.7.patch", when="@1.6:1.7^fujitsu-ssl2")
 
+    # Disable XNNPACK SME on A64FX
+    patch("disable-arm-sme-kernels-on-a64fx.patch", when="target=a64fx")
```
adamjstewart, line 471: Have any patches like this been contributed upstream to PyTorch? I want to try to avoid maintaining our own private Spack patches forever, and generally prefer to fix bugs in the underlying software directly. I also need to test that this patch works for all versions of PyTorch.

otsukay: Sorry for the delayed response.

I understand your point. I will withdraw this Spack PR for now, and if I pursue this further, I will submit it upstream to PyTorch instead.

In addition, since this patch is specific to A64FX, which is mostly used only on Fugaku in practice, I am not sure whether it would be very useful to pursue this upstream.

Thank you.

adamjstewart: I'm fine with adding patches to Spack for older versions, as long as it's on the developers' radar and they can review them. I imagine they would be happy to support A64FX, just don't have the hardware to test it themselves.

PR #4393, opened 2026-04-20
Update py-pandas Cython dependency to 3.1.0 for version for v3.0.2

KiSchnelle:
3.0.2 seems to make use of the `@cython.critical_section` decorator which was added in Cython 3.1.0 beta 1 (2025-04-03) so the Cython dependency needs to be updated.

### DIFF
--- a/repos/spack_repo/builtin/packages/py_pandas/package.py
+++ b/repos/spack_repo/builtin/packages/py_pandas/package.py
@@ -80,6 +80,7 @@ class PyPandas(PythonPackage):
         depends_on("py-meson-python@0.13.1:", when="@2.1:")
         depends_on("meson@1.2.1:", when="@2.1.1:")
         depends_on("meson@1.0.1:", when="@2.1.0")
+        depends_on("py-cython@3.1.0:", when="@3.0.2:")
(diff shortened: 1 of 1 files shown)


rgommers (approved):
Looks right, thanks!

https://github.com/pandas-dev/pandas/blob/9552b12786ba1ca1285c63fe533b8ef8e2f6b09f/pyproject.toml#L8 shows that Cython 3.1.0 is the minimum required indeed.

alecbcs (approved):
Looks good to me. Thanks!

repos/spack_repo/builtin/packages/py_pandas/package.py
```diff
@@ -80,6 +80,7 @@ class PyPandas(PythonPackage):
         depends_on("py-meson-python@0.13.1:", when="@2.1:")
         depends_on("meson@1.2.1:", when="@2.1.1:")
         depends_on("meson@1.0.1:", when="@2.1.0")
+        depends_on("py-cython@3.1.0:", when="@3.0.2:")
```
adamjstewart, line 83:
```suggestion
        depends_on("py-cython@3.1.0:", when="@3.0.1:")
```
based on https://github.com/pandas-dev/pandas/pull/64078

PR #4399, opened 2026-04-20
py-cfgv: add v3.5.0

RobertMaaskant:
Diff: https://github.com/asottile/cfgv/compare/v3.4.0...v3.5.0


manuelakuhn (approved):

PR #4401, opened 2026-04-20
py-identify: add v2.6.19

RobertMaaskant:
Diff: https://github.com/pre-commit/identify/compare/v2.6.15...v2.6.19


manuelakuhn (approved):

PR #4402, opened 2026-04-20
py-zipp: add v3.23.1

RobertMaaskant:
Release notes: https://github.com/jaraco/zipp/blob/v3.23.1/NEWS.rst
Diff: https://github.com/jaraco/zipp/compare/v3.23.0...v3.23.1


manuelakuhn (approved):

PR #4405, opened 2026-04-20
py-packaging: add v26.1

RobertMaaskant:
Release notes: https://github.com/pypa/packaging/releases/tag/26.1
Diff: https://github.com/pypa/packaging/compare/26.0...26.1

### DIFF
--- a/repos/spack_repo/builtin/packages/py_packaging/package.py
+++ b/repos/spack_repo/builtin/packages/py_packaging/package.py
@@ -32,7 +32,7 @@ class PyPackaging(PythonPackage):
     version("17.1", sha256="f019b770dd64e585a99714f1fd5e01c7a8f11b45635aa953fd41c689a657375b")
     version("16.8", sha256="5d50835fdf0a7edf0b55e311b7c887786504efea1177abd7e69329a8e5ea619e")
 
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/py_packaging/package.py
manuelakuhn:
```suggestion
    depends_on("py-flit-core@3.12:", when="@26:", type="build")
```
see https://github.com/pypa/packaging/blob/26.1/pyproject.toml#L2

manuelakuhn (changes requested):

RobertMaaskant:
Thanks for spotting the mistake @manuelakuhn. Missed it, because it did not show up in the diff with 26.0.

manuelakuhn:
Yes, the wrong dependency came from when the last release was added.

Sorry, for some reason my suggestion now overwrote the dependency instead of adding the line (no idea why that happened). Can you add  `depends_on("py-flit-core@3.3:", when="@22:", type="build")` again, please?

manuelakuhn (approved):
Thank you very much. LGTM and builds fine on rocky linux 8 with gcc@15

PR #4407, opened 2026-04-20
kokkos-kernels: route spec cxxflags through CMAKE_CXX_FLAGS

pbrady:
`kokkos_launch_compiler` is installed as the launcher on every target that links `Kokkos::kokkoscore` and strips the Spack compiler wrapper on TUs tagged `-DKOKKOS_DEPENDENCE`, which drops `SPACK_CXXFLAGS`.

We hit this building `kokkos-kernels %clang@19 +cuda` (clang as the CUDA device compiler) against a clang toolchain whose default libstdc++ ships a `<format>` header incompatible with clang-CUDA (llvm/llvm-project#69288). Setting `cxxflags='--gcc-install-dir=...'` on the spec picks a working libstdc++, but because the wrapper is stripped, the flag never reaches the real compiler.

Routing `cxxflags` through `CMAKE_CXX_FLAGS` keeps them in the command CMake assembles, so the launcher forwards them unchanged.

### DIFF
--- a/repos/spack_repo/builtin/packages/kokkos_kernels/package.py
+++ b/repos/spack_repo/builtin/packages/kokkos_kernels/package.py
@@ -255,4 +255,12 @@ class KokkosKernels(CMakePackage, CudaPackage):
         layout_value = spec.variants["layouts"].value
         options.append(self.define(f"KokkosKernels_INST_LAYOUT{layout_value.upper()}", True))
 
+        # kokkos_launch_compiler strips the Spack compiler wrapper on TUs
+        # tagged -DKOKKOS_DEPENDENCE, which loses SPACK_CXXFLAGS injected by
+        # the wrapper. Route the spec's cxxflags through CMAKE_CXX_FLAGS so
+        # they live in the command CMake assembles and survive forwarding.
+        cxxflags = spec.compiler_flags["cxxflags"]
+        if cxxflags:
+            options.append(self.define("CMAKE_CXX_FLAGS", " ".join(cxxflags)))
+
         return options


rbberger:
@pbrady please share more details about your concretization, Spack version, and your Spack compiler setup. I understand the kind of problem you're facing, but based on my experience it usually can be fixed through other means. This is very much a hack.

E.g., on my workstation my Ubuntu clang 17 picks up the wrong GCC 14 headers. Fixing the compiler external definition resolves this for `kokkos` and `kokkos-kernels`.

pbrady:
Here's an excerpt from my `packages.yaml` for llvm:

```yaml
  llvm:
    externals:
    - spec: llvm@19.1.0+clang+flang+lld+lldb
      prefix: /vast/projects/darwin_contrib/spack-builds/2024-10-16/rhel8/x86_64/llvm/19.1.0-gcc_13.2.0
      extra_attributes:
        compilers:
          c: /vast/projects/darwin_contrib/spack-builds/2024-10-16/rhel8/x86_64/llvm/19.1.0-gcc_13.2.0/bin/clang
          cxx: /vast/projects/darwin_contrib/spack-builds/2024-10-16/rhel8/x86_64/llvm/19.1.0-gcc_13.2.0/bin/clang++
          fortran: /vast/projects/darwin_contrib/spack-builds/2024-10-16/rhel8/x86_64/llvm/19.1.0-gcc_13.2.0/bin/flang-new
        flags:
          cxxflags: --gcc-install-dir=/vast/projects/opt/centos8/x86_64/gcc/11.2.0/bin/../lib/gcc/x86_64-pc-linux-gnu/11.2.0
```
Is something missing?

JBludau:
@rbberger is this the same problem as the one you mentioned?

rbberger:
@JBludau I've verified the failure and fix. The real fix would be that the launch compiler uses the Spack compiler wrapper and not llvm directly. But in order for that to work, the `kokkos_launch_compiler` would have to be made "Spack-aware` or Spack would have to "wrap" the wrapper. I'll approve this in the meantime as a feasible workaround. Since cxxflags is empty most of the time, this will not affect most builds.

rbberger (approved):
@pbrady sorry this took so long. we will want a better solution in the long run, but for now this will do.

lucbv (approved):

PR #4410, opened 2026-04-20
ceres-solver: limit eigen to eigen3

snehring:
Needs eigen 3.


alecbcs (approved):

PR #4411, opened 2026-04-20
[WIP] hip-lang support

scheibelp:
Apply changes from https://github.com/spack/spack/pull/49673/ that apply to spack-packages.

Needs 
- [x] https://github.com/spack/spack/pull/52145
- [x] https://github.com/spack/compiler-wrapper/pull/20

Interacts with:

- [ ] https://github.com/spack/spack/pull/52826
- [ ] https://github.com/spack/spack/pull/52706

Notes:

- https://github.com/spack/spack/pull/49673/  changed `SPACK_TEST_COMMAND=dump-env` to `dump-var`, but that would be harder to change (the tests that use this are in `spack-core` ) and it didn't look essential
- I edited a few cases of `requires("%[virtuals=c,cxx] llvm-amdgpu")` to include `when='%c'` to accommodate externals that do not specify a compiler (and in a couple cases I swapped `requires` for `depends_on` for the same reason)

Other than that this should be a copy of https://github.com/spack/spack/pull/49673/


scheibelp:
This will have to update the Spack commit used in `.ci/env` to use at least [a5b4056](https://github.com/spack/spack/commit/a5b4056c9019d27d7d38885d37f3be4446b48dd8) (from March 27) to pass the package audit (Currently pinned to a commit from March 20)

haampie:
You have to bump the package api

repos/spack_repo/builtin/build_systems/rocm.py
```diff
@@ -196,14 +195,14 @@ def asan_on(self, env: EnvironmentModifications):
...
     # processor is included in llvm/lib/Support/TargetParser.cpp
-    depends_on("llvm-amdgpu@5.2.0:", when="amdgpu_target=gfx940")
-    depends_on("llvm-amdgpu@5.7.0:", when="amdgpu_target=gfx941")
-    depends_on("llvm-amdgpu@5.7.0:", when="amdgpu_target=gfx942")
-    depends_on("llvm-amdgpu@5.2.0:", when="amdgpu_target=gfx1036")
-    depends_on("llvm-amdgpu@5.3.0:", when="amdgpu_target=gfx1100")
-    depends_on("llvm-amdgpu@5.3.0:", when="amdgpu_target=gfx1101")
-    depends_on("llvm-amdgpu@5.3.0:", when="amdgpu_target=gfx1102")
-    depends_on("llvm-amdgpu@5.3.0:", when="amdgpu_target=gfx1103")
+    requires("%[virtuals=hip-lang] llvm-amdgpu@5.2.0:", when="amdgpu_target=gfx940")
```
haampie:
```suggestion
    requires("%hip-lang=llvm-amdgpu@5.2.0:", when="amdgpu_target=gfx940")
```

is a bit shorter

repos/spack_repo/builtin/packages/comgr/package.py
```diff
@@ -63,7 +63,7 @@ def url_for_version(self, version):
     depends_on("zlib-api", type="link")
     depends_on("z3", type="link")
     depends_on("ncurses", type="link")
-    requires("%[virtuals=c,cxx] llvm-amdgpu", when="%c")
```
haampie: The `when=%c` can be dropped here and elsewhere, right?

scheibelp: I want to be able to create a `comgr` external without having to specify `%llvm-amdgpu` on the spec

repos/spack_repo/builtin/repo.yaml
```diff
@@ -1,3 +1,3 @@
 repo:
   namespace: builtin
-  api: v2.2
+  api: v2.4
```
alalazo: This is wrong, unfortunately. In https://github.com/spack/spack/pull/52145 we didn't bump the API version (which is not a big deal to fix), so this needs to be v2.5 - which still needs to be bumped in Spack core.

I'll update  https://github.com/spack/spack/pull/52112 to aggregate the two changes in a single bump.

alalazo (changes requested):
Besides the comment, we may want to delay merging this, so that the next release of the repository is still readable by Spack v1.1

eugeneswalker (dismissed):

scheibelp:
Current failures in pipelines:

- `e4s-rocm-external`: https://github.com/spack/spack-packages/blob/develop/stacks/e4s-rocm-external/spack.yaml /  https://gitlab.spack.io/spack/spack-packages/-/pipelines/1573082
  - `heffte`: https://gitlab.spack.io/spack/spack-packages/-/jobs/21969492
    - Not sure why, but even though the concretized spec is `heffte+rocm`, it is not dependent on `hip-lang` (instead is compiled with `gcc`, which leads to a failure
    - Contrast this for example with `camp+rocm` (which succeeds): https://gitlab.spack.io/spack/spack-packages/-/jobs/21969510
    - Locally I can get `heffte` to use `rocmcc` (even when `%gcc` is present), and it could also be set as a preference in the pipeline spack.yaml, but that should not be required (`heffte` should *need* `llvm-amdgpu` when `+rocm` is set, with the changes in this PR)
    - Package definition: https://github.com/spack/spack-packages/blob/develop/repos/spack_repo/builtin/packages/heffte/package.py
  - `tasmanian`: failure appears similar to `heffte`: https://gitlab.spack.io/spack/spack-packages/-/jobs/21969505
  - `mfem`: concretizes properly, does not set `ROCM_PATH`: https://gitlab.spack.io/spack/spack-packages/-/jobs/21969516
  - `kokkos-kernels`: also chooses `%gcc` (like first two list items here), but does not inherit `ROCmPackage` or have a `+rocm` variant (so might be correctible with that)
    - https://github.com/spack/spack-packages/blob/develop/repos/spack_repo/builtin/packages/kokkos_kernels/package.py
    - https://gitlab.spack.io/spack/spack-packages/-/jobs/21969513

scheibelp:
A couple updates on https://github.com/spack/spack-packages/pull/4411#issuecomment-4426062340 (comment immediately above this)

I replicated concretizations locally and they *are* using `llvm-amdgpu` for `hip-lang` (as needed), but not displaying it: the following change would be required to Spack to do so

```
diff --git a/lib/spack/spack/spec.py b/lib/spack/spack/spec.py
index a0a464854b..16e57b1372 100644
--- a/lib/spack/spack/spec.py
+++ b/lib/spack/spack/spec.py
@@ -4469,7 +4469,7 @@ def compilers(self):
         # TODO: get rid of the space here and make formatting smarter
         return " " + self._format_dependencies(
             "{name}{@version}",
-            include=lambda dep: any(lang in dep.virtuals for lang in ("c", "cxx", "fortran")),
+            include=lambda dep: any(lang in dep.virtuals for lang in ("c", "cxx", "fortran", "hip-lang")),
             deptypes=False,
             _force_direct=True,
         )
```

that's a cosmetic change though, and not the cause of the "real" problems mentioned above. I think the real issue is that right now, I cannot set `hip: /opt/rocm/lib/llvm/bin/amdclang++` as I'm doing:

```
    llvm-amdgpu:
      externals:
      - spec: llvm-amdgpu@7.2.0
        prefix: /opt/rocm/lib/llvm
        extra_attributes:
          compilers:
            c: /opt/rocm/lib/llvm/bin/amdclang
            cxx: /opt/rocm/lib/llvm/bin/amdclang++
            fortran: /opt/rocm/lib/llvm/bin/amdflang
            hip: /opt/rocm/lib/llvm/bin/amdclang++
```

this "overloads" `amdclang++`: compiler invocations that don't specify `-x` currently decide what underlying compiler to invoke (SPACK_CC, SPACK_CXX, SPACK_FC, and here SPACK_HIPCXX) based on the command name invoked. Right now `amdclang++` maps to SPACK_CXX. Possibilities on how to address:

* force user to point `hip:` at `hipcc` (and presume hipcc are hip compilations): this would create a rift between what Spack thinks hipcc is for and what rocm thinks hipcc is for (a driver that can dispatch to c++ or hip compiler as needed)
    * (this probably doesn't matter when spack successfully redirects a build to use it's chosen hip wrapper as the hip compiler, but potentially does matter when it fails at this)
* make https://github.com/spack/compiler-wrapper/blob/main/cc.sh look at extensions of input files

repos/spack_repo/builtin/packages/kokkos_kernels/package.py
```diff
@@ -3,11 +3,12 @@
 # SPDX-License-Identifier: (Apache-2.0 OR MIT)
 from spack_repo.builtin.build_systems.cmake import CMakePackage
 from spack_repo.builtin.build_systems.cuda import CudaPackage
+from spack_repo.builtin.build_systems.rocm import ROCmPackage
 
 from spack.package import *
 
 
-class KokkosKernels(CMakePackage, CudaPackage):
+class KokkosKernels(CMakePackage, CudaPackage, ROCmPackage):
```
cedricchevalier19, line 12: Ideally `KokkosKernels` should only be a regular `CMakePackage`, as the hip/rocm part is done in  `Kokkos`

repos/spack_repo/builtin/packages/kokkos/package.py
```diff
@@ -484,6 +484,7 @@ def cmake_args(self):
             spack_microarches.append(kokkos_microarch_name)
 
         if spec.satisfies("+rocm"):
+            options.append(f"-DROCM_PATH={spec['hip'].prefix}")
```
cedricchevalier19:
```suggestion
            options.append(self.define("ROCM_PATH", spec['hip'].prefix))
```
Why using`ROCM_PATH` instead of `ROCM_ROOT`?
I thought dependencies path where already injected to CMake, do we need this?

repos/spack_repo/builtin/packages/kokkos_kernels/package.py
tpadioleau: Not sure of the reason of this change, but it is not obvious to me whether packages depending on Kokkos should not also depend on these new device languages ? They need a device compiler, so they should be able to choose the compiler as we do for host compilers ?

scheibelp: The `kokkos-kernels` build seems to build at least some artifacts with `HIPCXX` when that is set (maybe a test?), so I inherited `ROCmPackage` to set up the new spack hip wrapper properly, that might be overdoing it but I'm experimenting with this for now.

I've added `[WIP]` to the title in the meantime (I might be changing things around significantly in the short term)

lucbv: If anything, with more time on my hands to work on the spack package I would have attempted to remove the CudaPackage inheritance as I want for Kokkos Kernels to inherit Cuda from the Kokkos settings. I can understand that people might want to try more complex things with different compilers for Kokkos and Kokkos Kernels but that's not supported by the developers team so it should really not be done in our official package imho

repos/spack_repo/builtin/packages/legion/package.py
```diff
@@ -495,7 +495,17 @@ def cmake_args(self):
...
+            with working_dir(self.build_directory):
+                if self.generator == "Unix Makefiles":
+                    make(*self.build_targets, parallel=False)
+                elif self.generator == "Ninja":
+                    self.build_targets.append("-v")
+                    ninja(*self.build_targets, parallel=False)
+                else:
+                    raise ValueError(f"Unexpected generator for +rocm: {self.generator}")
+        else:
+            super().build(spec, prefix)
```
rbberger: What's the story here?

repos/spack_repo/builtin/packages/legion/package.py
```diff
@@ -495,7 +495,17 @@ def cmake_args(self):
         return options
 
     def build(self, spec, prefix):
-        super().build(spec, prefix)
+        if spec.satisfies("+rocm"):
+            with working_dir(self.build_directory):
+                if self.generator == "Unix Makefiles":
+                    make(*self.build_targets, parallel=False)
```
elliottslaughter: This is just for debugging, right?

scheibelp: No: I was getting random failures when building `legion+rocm` that go away when I disabled a parallel build (i.e. without this change, builds of `legion+rocm` occasionally fail).

See https://gitlab.spack.io/spack/spack-packages/-/jobs/23068306 for an example of the failure

(I was able to reproduce the issue - and fix - locally).

repos/spack_repo/builtin/packages/legion/package.py
scheibelp: https://github.com/spack/spack-packages/pull/4411#discussion_r3538424039

repos/spack_repo/builtin/packages/legion/package.py
elliottslaughter: @scheibelp I'm not sure I understand the nature of the issue, or whether the fix you mentioned is what is committed into this repo, or something you have in your local clone but have not pushed.

We build Legion with ROCm all the time in parallel. I cannot recall a user report that this didn't work. If the Spack CI machines are especially memory starved, then perhaps it might be necessary to reduce the degree of parallelism, but it shouldn't need to be reduced to zero in any case.

If I'm missing something about the Spack CI environment maybe @rbberger would know how to address it properly.

rbberger: @elliottslaughter my guess is that the ancient HIP CMake macros are fragile and should be replaced with HIP CMake language equivalents.

rbberger: @scheibelp what is weird though is why this is happening now. We've been building this for many years without issues. This seems runner specific.

stacks/e4s/spack.yaml
```diff
@@ -378,42 +378,42 @@ spack:
...
     - gasnet
     - ginkgo
     - heffte
     - hpx
-    - hypre
-    - kokkos
-    - lammps
+    - hypre +rocm
+    - kokkos +rocm
+    - "lammps@20260211: +rocm fft_kokkos=hipfft"
```
rbberger, line 395: You will want to do `+kokkos` here too. The solver should figure it out because of the dependent variant, but just be explicit.

rbberger: beware, you are changing pkg selection here. the current e4s stack uses the GPU package, not Kokkos. Be sure the change is intentional.

repos/spack_repo/builtin/packages/legion/package.py
scheibelp: I agree it is strange but I reproduced the issue myself locally, so I assumed it was a universal problem. This problem occurred in a build of the `e4s-rocm-external` stack. This PR makes some significant changes to how rocm builds are modeled in Spack, though I couldn't think of why that would be inherently wrong for legion, so I was thinking of this as a legion build issue (not to say I am right about that, but just how I was approaching it).

stacks/e4s/spack.yaml
scheibelp: Sorry I'm not sure what you mean, are you talking about

- In terms of what things concretize to, given the constraints
- Or do you think my *inputs* are fundamentally different

?

wdconinc (dismissed):
Approving hep stack changes.

stacks/e4s/spack.yaml
rbberger: I'm talking about that given the new constraints, you are changing which GPU accelerator package is selected in LAMMPS.

LAMMPS has multiple **accelerator packages** that are mutually exclusive and each implement CUDA/ROCm support:
-  **GPU package** is used by default and encoded in Spack as (`~kokkos+cuda`, `~kokkos+rocm`, `~kokkos+opencl`)
-  **Kokkos package** is only activated with `+kokkos` and then uses the respective Kokkos backends for CUDA, ROCm, etc.

If you explicitly add `fft_kokkos=hipfft` this forces `+kokkos`, which changes the default. I can't speak for what E4S prefers. Each GPU backend has its pro and cons. I would expect them to most likely to want the more modern Kokkos package, which would make your change ok. All I'm pointing out is that this should be an intentional choice, since the two accelerator packages fundamentally operate differently.

repos/spack_repo/builtin/packages/legion/package.py
rbberger: One thing I found is our Spack builds are always `legion+shared`. Maybe it only shows up with `~shared`? Wouldn't be the first time that the static library build is fragile.

rbberger: While trying to reproduce, I noticed that a block like
```yaml
packages:
  c:
    prefer: [gcc]
  cxx:
    prefer: [gcc]
  fortran:
    prefer: [gcc]
```
seems to "disable" the hip-lang compiler in a package like `realm+rocm`, even though it is required to have one. At least the concretization result won't show it anymore.
```
 -   vendyjj      ^realm@26.7.1~cuda~hdf5~hwloc~ipo~kokkos+libdl~openmp~papi~python+rocm~shared~sysomp amdgpu_target:=gfx90a build_system=cmake build_type=Release commit=76631cdb7380033619958ad37ec3c94e7129095b conduit=mpi cxxstd=20 generator=make log_level=warning max_dims=3 network=gasnet platform=linux os=rhel8 target=x86_64_v3 %c,cxx,hip-lang=rocmcc@7.2.2
```
vs
```
 -   sth5uyg      ^realm@26.7.1~cuda~hdf5~hwloc~ipo~kokkos+libdl~openmp~papi~python+rocm~shared~sysomp amdgpu_target:=gfx90a build_system=cmake build_type=Release commit=76631cdb7380033619958ad37ec3c94e7129095b conduit=mpi cxxstd=20 generator=make log_level=warning max_dims=3 network=gasnet platform=linux os=rhel8 target=x86_64_v3 %c,cxx=gcc@13.2.0
```

scheibelp: I think that is a display bug: `%hip-lang` can be omitted from the spec printout without the following change in Spack (not PR'ed yet):

```
diff --git a/lib/spack/spack/spec.py b/lib/spack/spack/spec.py
index a0a464854b..16e57b1372 100644
--- a/lib/spack/spack/spec.py
+++ b/lib/spack/spack/spec.py
@@ -4469,7 +4469,7 @@ def compilers(self):
         # TODO: get rid of the space here and make formatting smarter
         return " " + self._format_dependencies(
             "{name}{@version}",
-            include=lambda dep: any(lang in dep.virtuals for lang in ("c", "cxx", "fortran")),
+            include=lambda dep: any(lang in dep.virtuals for lang in ("c", "cxx", "fortran", "hip-lang")),
             deptypes=False,
             _force_direct=True,
         )
```

scheibelp: Now at https://github.com/spack/spack/pull/52706 (I don't consider that PR a prerequisite for this one though, because the effect is cosmetic)

scheibelp:
I ran benchmarks on this for a modified version of the e4s-rocm-external environment with Spack 1.2.1. The `kokkos-kernels `requirement was relaxed and one spec was removed from the speclist because otherwise the same list doesn't run across develop and hip-lang branches of spack-packages repo). There was a minor regression in solve time (attached results).

I also ran this for the `radiuss.txt` example from the spack-benchmark git repository and there was no observed change in solve times.

I would guess that at least for now, because there is only one hip-lang provider available (`llvm-amdgpu`), there should not generally be a major impact on concretization times.

I could run this for a non-external case, although I think that would take a bit of additional work (to make sure an llvm-amdgpu spec is available, because I don't see how spec groups integrate "comfortably" into the benchmarking workflow).

[base-external-rocm-env-example1.yaml](https://github.com/user-attachments/files/30068463/base-external-rocm-env-example1.yaml)
[minor-solve-regression.txt](https://github.com/user-attachments/files/30068465/minor-solve-regression.txt)

stacks/e4s/spack.yaml
```diff
@@ -378,42 +378,42 @@ spack:
 
 
     # ROCM 90a
-    - adios2 +kokkos
-    - amrex
-    - arborx
+    - adios2 +kokkos +rocm
+    - amrex +rocm
+    - arborx +rocm
```
haampie, line 383: Why is +rocm added if it's also under all:prefer?

repos/spack_repo/builtin/packages/kokkos/package.py
```diff
@@ -440,6 +440,33 @@ def kokkos_cxx(self) -> str:
         # Assumes build-time globals have been set already
         return spack_cxx
 
+    def get_clang_rt_library(self):
+        """Find the clang runtime library for ROCm with relocatable device code.
+
+        Returns the path to libclang_rt.builtins-x86_64.a by running
+        amdclang++ --print-resource-dir and searching for the library.
+        """
+        import llnl.util.filesystem as fs
```
haampie: You should not import anything besides `spack.package`. This modules doesn't even exist. Why did that not break for you?

repos/spack_repo/builtin/packages/comgr/package.py
```diff
@@ -101,9 +103,9 @@ def url_for_version(self, version):
         "7.2.3",
         "7.13.0",
     ]:
-        # llvm libs are linked statically, so this *could* be a build dep
+        # in reality this should be a link type dep, but libLLVM is linked statically,
```
haampie: llvm-amdgpu is a link dep here, the rephrased comment suggests it's not. Better to leave the comment as it was.

repos/spack_repo/builtin/packages/legion/package.py
```diff
@@ -495,7 +495,17 @@ def cmake_args(self):
         return options
 
     def build(self, spec, prefix):
-        super().build(spec, prefix)
+        if spec.satisfies("+rocm"):
```
haampie: Please extract unrelated changes, s.t. reverting just hip-lang changes is easier.

Besides that, don't inline `super()` with modifications but define `def parallel` to avoid verbosity and drift between the base/derived class.

repos/spack_repo/builtin/packages/llvm_amdgpu/package.py
```diff
@@ -471,10 +473,11 @@ def cmake_args(self):
         args.append(self.define("RUNTIMES_CMAKE_ARGS", runtime_cmake_args))
         return args
 
-    compiler_languages = ["c", "cxx", "fortran"]
+    compiler_languages = ["c", "cxx", "fortran", "hip-lang"]
     c_names = ["amdclang"]
     cxx_names = ["amdclang++"]
     fortran_names = ["amdflang"]
+    hip_lang_names = ["amdclang++"]
```
haampie: This shadows cxx_names, so you end up with compilers detected for the languages `c`, `fortran`, `hip-lang`, but `cxx` will be missing.

repos/spack_repo/builtin/build_systems/compiler.py
```diff
@@ -210,6 +210,17 @@ def _cxx_path(self) -> Optional[str]:
         """Returns the path to the C++ compiler, if the package was installed by Spack"""
         return None
 
+    @property
+    def hip(self) -> Optional[str]:
+        assert self.spec.concrete, "cannot retrieve HIP compiler, spec is not concrete"
+        if self.spec.external:
+            return self.spec.extra_attributes["compilers"].get("hip", None)
```
haampie: this is `"hip"` but in detection it's `"hip-lang"`.

haampie: Also, `extra_attributes["compilers"]` is used here which is not consistent with the defensive `extra_attributes.get("compiler", {})` elsewhere. I think the defensive version is better.

repos/spack_repo/builtin/packages/kokkos/package.py
haampie: Why do you hard-code an architecture? Looks wrong/unnecessary.

repos/spack_repo/builtin/packages/rocfft/package.py
```diff
@@ -143,7 +146,6 @@ def root_cmakelists_dir(self):
             return "."
 
     def setup_build_environment(self, env: EnvironmentModifications) -> None:
-        env.set("CXX", self.spec["hip"].hipcc)
```
haampie, line 146: Inconsistent with rocblas/rocsolver, why remove this only here?

repos/spack_repo/builtin/build_systems/compiler.py
```diff
@@ -210,6 +210,17 @@ def _cxx_path(self) -> Optional[str]:
         """Returns the path to the C++ compiler, if the package was installed by Spack"""
         return None
 
+    @property
+    def hip(self) -> Optional[str]:
```
haampie, line 214: Is it useful to add this and produce `str | None`? It looks brittle. In particular I think you now run in a potential TypeError in compiler-wrapper when it is `None` in the `_var_list.append(("hip-lang", "hip", "HIPCXX", "SPACK_HIPCXX"))` bit.

In the GCC package we have the invariant that if `languages=c` then `c` is provided. Here you have something like `hip` is maybe provided. Would be good to make the None state impossible.

haampie:
Can you extract changes that are not related to `hip-lang` as a virtual? That makes reverts easier.

repos/spack_repo/builtin/packages/legion/package.py
rbberger: @scheibelp just fyi, Legion/Realm are moving [forward with changes ](https://gitlab.com/StanfordLegion/legion/-/merge_requests/2086) to use HIP lang instead of the older HIP macros. I'm expecting this get into the September Legion release. The hope is that this helps with the parallel build.

repos/spack_repo/builtin/packages/legion/package.py
scheibelp: https://github.com/spack/spack-packages/pull/6121 (that PR will be necessary to merge this PR)

stacks/ml-linux-x86_64/spack.yaml
```diff
@@ -142,6 +142,10 @@ spack:
         hwloc:
           require:
             - ~rocm
+    needs: [hip-compiler]
+  - group: hip-compiler
+    specs:
+    - llvm-amdgpu
```
adamjstewart, line 157: What is this?

adrienbernede:
@scheibelp it looks like CachedCMakePackages should be updated too.

I also created this branch in radiuss-spack-configs to trigger tests of the present PR:
https://github.com/llnl/radiuss-spack-configs/pull/200

It currently fails building camp in our setup. 
```console
-- BLT HIP support is ON
-- Creating BLT HIP targets...
-- The HIP compiler identification is unknown
/usr/workspace/radiuss/rsc-ci/shared-ci/corona/zen/ref-woptim-test-spack-packages-pr-4411/install/[padded-to-128-chars]/linux-rhel8-zen/none-none/compiler-wrapper-1.2.0-u3zmgx5lweuq63gqi5macchrxwqy5ibp/libexec/spack/rocmcc/spackhip: line 213: exec: None: not found
CMake Error at /opt/rocm-6.4.3/lib/cmake/hip-lang/hip-lang-config.cmake:139 (message):
  hip-lang Error:127 - clangrt builtins lib could not be found.
Call Stack (most recent call first):
  /usr/tce/backend/installations/linux-rhel8-x86_64/gcc-10.3.1/cmake-3.25.2-2p7loskhisrpohhvpwdmse2nadqgzqcb/share/cmake-3.25/Modules/CMakeHIPInformation.cmake:146 (find_package)
  /usr/workspace/radiuss/rsc-ci/shared-ci/corona/zen/ref-woptim-test-spack-packages-pr-4411/install/[padded-to-128-chars]/linux-rhel8-zen/none-none/blt-0.7.2-kl55lhkpwul6ozj6bovps4ccjug4iogl/cmake/thirdparty/BLTSetupHIP.cmake:14 (enable_language)
  /usr/workspace/radiuss/rsc-ci/shared-ci/corona/zen/ref-woptim-test-spack-packages-pr-4411/install/[padded-to-128-chars]/linux-rhel8-zen/none-none/blt-0.7.2-kl55lhkpwul6ozj6bovps4ccjug4iogl/cmake/BLTSetupTargets.cmake:118 (include)
  /usr/workspace/radiuss/rsc-ci/shared-ci/corona/zen/ref-woptim-test-spack-packages-pr-4411/install/[padded-to-128-chars]/linux-rhel8-zen/none-none/blt-0.7.2-kl55lhkpwul6ozj6bovps4ccjug4iogl/cmake/SetupThirdParty.cmake:6 (include)
  /usr/workspace/radiuss/rsc-ci/shared-ci/corona/zen/ref-woptim-test-spack-packages-pr-4411/install/[padded-to-128-chars]/linux-rhel8-zen/none-none/blt-0.7.2-kl55lhkpwul6ozj6bovps4ccjug4iogl/SetupBLT.cmake:129 (include)
  cmake/load_blt.cmake:27 (include)
  CMakeLists.txt:48 (include)
```

I may need to update the hip config, or does this suppose an update of Spack itself ?

stacks/ml-linux-x86_64/spack.yaml
scheibelp: Language dependents (i.e. things that need compilers), need those providers to be available "before" they concretize, either as an installed package, an external, or as a spec group.

scheibelp:
@adrienbernede thanks for trying that. I'm wondering if you haven't added `hip:` to the `compilers:` extra attribute for `llvm-amdgpu`

```
    llvm-amdgpu:
      buildable: false
      externals:
      - spec: llvm-amdgpu@7.2.1
        prefix: /opt/rocm...
        extra_attributes:
          compilers:
            c: ...
            cxx: ...
            fortran: ...
            hip: (same thing as cxx, usually) <-------
```

I see in your pipeline concretization that it's choosing an external `llvm-amdgpu`, but didn't see a change to `packages.yaml` in your PR

 I'm able to build `camp+rocm` with this PR (example config for llvm-amdgpu is in `stacks/e4s-rocm-external/spack.yaml`.

stacks/ml-linux-x86_64/spack.yaml
adamjstewart: Interesting, so this is still a holdover from when compilers were not dependencies? Hopefully we can bootstrap compilers someday as well, I know Harmen is working on that. Is this something that should also be done for things like rust and cuda? Or are they different because they can be bootstrapped (kinda)?

I'm fine with these changes. Wonder if we should use require instead of prefer, but I defer to you on that.

PR #4415, opened 2026-04-21
fastjet: support CMake build system for @3.5.0:

wdconinc:
FastJet 3.5.0 introduced an official CMake build system. This PR converts the spack package to support both `AutotoolsPackage` and `CMakePackage`, using the builder class pattern.

- CMake is the default build system for `@3.5.0:`; autotools is used for older versions
- The `shared` variant is kept for both build systems; `~shared` conflicts with `build_system=cmake` since CMake always builds shared
- The `cxxstd` variant is restricted to `build_system=autotools`; CMake handles the C++ standard via CMakeLists.txt (defaults to C++14)
- `AutotoolsBuilder` carries the existing `configure_args` logic, with an added guard for the `@3.4.0:`-only thread-safety variant
- `CMakeBuilder` maps plugins, auto-ptr, and thread-safety variants to the corresponding `FASTJET_*` CMake options

Test build:
```
-- linux-ubuntu26.04-skylake / %c,cxx=clang@23.0.0 --------------
cbezku4 fastjet@3.5.0~atlas~auto-ptr~ipo+shared build_system=cmake build_type=Release generator=make plugins:=cxx thread-safety=limited
==> 1 installed package
```


wdconinc:
@spackbot fix style

wdconinc:
FYI @m-fila

m-fila (approved):
Nice, thank you. I build locally some binaries against these versions and everything worked fine

repos/spack_repo/builtin/packages/fastjet/package.py
```diff
@@ -143,14 +167,40 @@ def configure_args(self):
...
+                    args += [
+                        self.define(f"FASTJET_ENABLE_PLUGIN_{cmake_name}", v.value in plugins)
+                    ]
+        args += [self.define_from_variant("FASTJET_HAVE_AUTO_PTR_INTERFACE", "auto-ptr")]
+        thread_safety = self.spec.variants["thread-safety"].value
+        args += [
+            self.define("FASTJET_HAVE_THREAD_SAFETY", thread_safety == "full"),
+            self.define(
+                "FASTJET_HAVE_LIMITED_THREAD_SAFETY", thread_safety in ("limited", "full")
+            ),
```
alalazo, line 204: Is it ok to have both options on in the `thread_safety=full` case?

alalazo (approved):

repos/spack_repo/builtin/packages/fastjet/package.py
wdconinc: Yes, they can both be `ON` (and this is actually the default; https://gitlab.com/fastjet/fastjet/-/blob/master/CMakeLists.txt#L132; though for probably spack-historical reasons we default to `limited` here).

wdconinc:
Took a few months, but bumped this up to recent, and it's ready for merge (once the github actions meltdown is over).

bernhardkaindl (approved):

PR #4416, opened 2026-04-21
DRAFT: kokkos: rewire supported GPU arch checks

rbberger:
Gives an error during concretization instead of during build

@alalazo here is an attempt of moving the errors from build time to concretization time. There is two concerns:
- the errors during concretization can be long and unhelpful. I have yet to find a way to make them more concise.
- I was asked whether the concretizer will because of this take longer in these error case, creating an unwanted penalty.


JBludau:
Alternative approach in #5873

PR #4417, opened 2026-04-21
metabat: limiting boost versions, fixing vendored htslib build

snehring:
Fixing a couple installation issues.

Firstly it needed a c compiler, then that expanded into incompatibilities with std=c++11 with newer boost, and then the vendored htslib needed autoreconfigured, and then (at least in my env) I needed to persuade htslib's configure of the host target.

Fun stuff.

### DIFF
--- a/repos/spack_repo/builtin/packages/metabat/package.py
+++ b/repos/spack_repo/builtin/packages/metabat/package.py
@@ -3,7 +3,6 @@
 # SPDX-License-Identifier: (Apache-2.0 OR MIT)
 
 from spack_repo.builtin.build_systems.cmake import CMakePackage
-from spack_repo.builtin.packages.boost.package import Boost
 
 from spack.package import *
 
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/metabat/package.py
```diff
@@ -21,20 +20,29 @@ class Metabat(CMakePackage):
...
     depends_on("autoconf", type="build")
     depends_on("cmake", type="build")
-    depends_on("boost@1.55.0:", type=("build", "run"))
 
-    # TODO: replace this with an explicit list of components of Boost,
-    # for instance depends_on('boost +filesystem')
-    # See https://github.com/spack/spack/pull/22303 for reference
-    depends_on(Boost.with_default_variants, type=("build", "run"))
+    depends_on(
+        "boost@1.55:1.84+program_options+filesystem+system+graph+serialization+iostreams\
```
hainest:
```suggestion
        "boost@1.55:1.82+program_options+filesystem+system+graph+serialization+iostreams\
```

Boost.Math requires C++14 after 1.82. It's not really enforced in the build system, but it's good to limit it here. Otherwise, you can use bump up to `cxxstd=14`.

repos/spack_repo/builtin/packages/metabat/package.py
```diff
@@ -21,20 +20,29 @@ class Metabat(CMakePackage):
...
+    )
     depends_on("perl", type="run")
     depends_on("zlib-api", type="link")
     depends_on("ncurses", type="link")
 
+    def patch(self):
+        filter_file(r"(autoconf)", r"autoreconf -i && \1", join_path("cmake", "htslib.cmake"))
+        filter_file(
+            "./configure",
+            f"./configure --host={self.spec.build_spec.target.family.name}-linux-gnu",
```
hainest:
```suggestion
            f"./configure --host={self.spec.build_spec.target.family.name}-linux-gnu --without-libdeflate ",
```

Although not directly related to the changes here, this is necessary to get htslib to build in a sterile environment because it dynamically links against libdeflate but doesn't set the rpath so there's a link error when building metabat. This issue was solved by https://bitbucket.org/berkeleylab/metabat/src/2fd1332d62b671b0564124eeb42c1d95f5c8e54e/cmake/htslib.cmake?at=master in 2024, but isn't available until the 2.18 release in 2025 (which isn't available here).

This is one of the dangers of using ExternalProject with spack. My recommendation would be to create a find module for htlib in metabat that will try CMake's `find_library` and then fall back to ExternalProject. That also has the benefit of automatically using a system-provided htslib and libz or lets metabat users specify one to use. For the spack package, you could then conditionally add `depends_on(htslib)` to let spack do the work for you.

I am more than happy to contribute that upstream to metabat, if that would be of interest.

hainest (changes requested):

repos/spack_repo/builtin/packages/metabat/package.py
snehring: I think it sounds like a good idea, hopefully they'd appreciate it.

hainest (approved):
Looks good. Thanks for the updates!

PR #4418, opened 2026-04-21
gcc-runtime: fixing samefiles error

snehring:
similar issue to #4313 

I only seem to hit this with compilers I've installed with spack, so maybe it's a common cause that can be better resolved elsewhere.


haampie:
thanks, likely related to `config:shared_linking:bind:true` in your config

haampie (approved):

PR #4420, opened 2026-04-21
openmolcas: adding v26.02

snehring:
version bump.

their mpi implementation detection is also incompatible with openmpi5 on older versions.


snehring:
@spackbot re-run pipeline

RikkiButler20 (approved):
LGTM

PR #4429, opened 2026-04-21
velero: new package

RobertMaaskant:

### DIFF
--- a/repos/spack_repo/builtin/packages/velero/package.py
+++ b/repos/spack_repo/builtin/packages/velero/package.py
@@ -1,0 +1,45 @@
+# Copyright Spack Project Developers. See COPYRIGHT file for details.
+#
+# SPDX-License-Identifier: (Apache-2.0 OR MIT)
+
+from spack_repo.builtin.build_systems.go import GoPackage
+
+from spack.package import *
+
+
+class Velero(GoPackage):
+    """
+    Velero is an open source tool to safely backup and restore, perform disaster recovery,
+    and migrate Kubernetes cluster resources and persistent volumes.
+    """
+
+    homepage = "https://velero.io/"
+    url = "https://github.com/velero-io/velero/archive/refs/tags/v1.18.0.tar.gz"
+
+    maintainers("RobertMaaskant")
+
+    license("Apache-2.0", checked_by="RobertMaaskant")
+
+    version("1.18.0", sha256="e94e51437b7cc54b633fcd253d5003494382e4de69c34992463588f248ba409c")
+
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/velero/package.py
```diff
@@ -0,0 +1,45 @@
...
+    def build_args(self):
+        ldflags = [f"-X 'github.com/vmware-tanzu/velero/pkg/buildinfo.Version=v{self.version}'"]
+
+        args = super().build_args
+
+        if "-ldflags" in args:
+            ldflags_index = args.index("-ldflags") + 1
+            args[ldflags_index] = args[ldflags_index] + " " + " ".join(ldflags)
+        else:
+            args.extend(["-ldflags", " ".join(ldflags)])
```
becker33:
```suggestion
    def build_args(self):
        ldflags = f"-X 'github.com/vmware-tanzu/velero/pkg/buildinfo.Version=v{self.version}'"

        args = super().build_args

        if "-ldflags" in args:
            ldflags_index = args.index("-ldflags") + 1
            args[ldflags_index] += " " + ldflags
        else:
            args.extend(["-ldflags", ldflags])
```
Currently we're using `" ".join` on a list with only one element, there's no need to complicate it like that.

becker33 (changes requested):

repos/spack_repo/builtin/packages/velero/package.py
RobertMaaskant: Thanks for the review! I don't mind changing it, but kept it like this so it is consistent with other `go` packages such as `kubectl`. Probably better to add the construction to the `GoPackage`, but haven't got around to that.

Let me know what you prefer and I'll do that.

alecbcs:
@RobertMaaskant it might be worth holding off on merging this PR until after https://github.com/spack/spack-packages/pull/4795 and then refactor to make the ldflags handling easier.

repos/spack_repo/builtin/packages/velero/package.py
RobertMaaskant: Updated the PR to use the ldflags property. A lot cleaner now, thanks @alecbcs for implementing that feature!

alecbcs (approved):
Looks good to me. Thanks @RobertMaaskant!

repos/spack_repo/builtin/packages/velero/package.py
RobertMaaskant: @becker33 friendly ping, would like to move this forward if possible. Let me know if you would like more changes!

becker33 (approved):

PR #4430, opened 2026-04-21
healpix-cxx: restrict cfitsio@3

snehring:
Needs cfitsio@3.

The libsharp dep is actually vendored in this version and isn't necessary as a dependency. 

I tried for a while to update this to the latest version but was hitting compilation issues that seem to stem from compatibility issues with libsharp, which they do not vendor in later versions.

The patch method doesn't actually seem to be needed with this version, but would in subsequent versions.


snehring:
explicit c,cxx deps also

alecbcs (approved):

PR #4432, opened 2026-04-22
SAMRAI: fix the finding of the CMakeBuilder class

white238:
The build system finding does not look for a `CachedCMakeBuilder`. This works around that and a missing import.


alecbcs (approved):
Looks good to me. Thanks @white238

becker33 (commented):
I'm pretty certain this is incorrect, and that the correct solution is to change all of the references to `build_system=cmake` to `build_system=cached_cmake` and change the `build_system` directive to

```
    build_system(
        conditional("cached_cmake", when="@4:"), conditional("autotools", when="@:3"), default="cached_cmake"
    )
```

@alecbcs @white238 I'll put up a PR for this

PR #4441, opened 2026-04-22
ispc: add version 1.30.0

albestro:
I just took the chance to add at least the latest release.

Updated LLVM dependency according to this (left out `llvm@23` because considered a development version)
https://github.com/ispc/ispc/blob/3fc6d50cf24dc8b4a0cdb346553445ff28962051/src/ispc_version.h#L25-L26

### DIFF
--- a/repos/spack_repo/builtin/packages/ispc/package.py
+++ b/repos/spack_repo/builtin/packages/ispc/package.py
@@ -124,7 +124,12 @@ class Ispc(CMakePackage):
         try:
             Executable(self.compiler.cc)("-m32", "-shared", "check-m32.c", error=str)
         except ProcessError:
-            filter_file("bit 32 64", "bit 64", "cmake/GenerateBuiltins.cmake")
+            # https://github.com/ispc/ispc/commit/3e03dffa8b58e77ea628d614f65763a8fdd90c18
+            if self.spec.satisfies("@1.25:"):
+                cmake_target_file = "cmake/CommonStdlibBuiltins.cmake"
+            else:
+                cmake_target_file = "cmake/GenerateBuiltins.cmake"
+            filter_file("bit 32 64", "bit 64", cmake_target_file)
 
     def cmake_args(self):
         spec = self.spec
--- a/repos/spack_repo/builtin/packages/ispc/package.py
+++ b/repos/spack_repo/builtin/packages/ispc/package.py
@@ -28,6 +28,7 @@ class Ispc(CMakePackage):
     license("BSD-3-Clause")
 
     version("main", branch="main")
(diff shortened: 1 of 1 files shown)


aumuell:
The added version limits on LLVM seem to be fine, as far as I could see from the root `CMakeLists.txt`.

The CI error `https://gitlab.spack.io/spack/spack-packages/-/jobs/21614650` because of the missing `<gnu/stubs-32.h>` header was taken care of by `filter_file("bit 32 64", "bit 64", "cmake/GenerateBuiltins.cmake")` in `patch`, but the build process seems to have changed. I also tried to reproduce that locally by removing all `lib32*` packages from an Arch system, but the build went fine. So I do not know what is going on here.

albestro:
Thanks for the pointer @aumuell! As you, I tried replicating the problem without luck, despite starting from the same docker image used in CI...not sure what I'm missing.

In any case, I investigated a bit more the `patch` you mentioned and it seems that file targeted went through a redesign in `ispc@1.25` with https://github.com/ispc/ispc/pull/2743. I then blindly updated the logic of `patch` in https://github.com/spack/spack-packages/pull/4441/commits/952fea6afabfe19f86112c656a1f7544cb28d8dc, let's see if it works.

repos/spack_repo/builtin/packages/ispc/package.py
```diff
@@ -58,6 +59,7 @@ class Ispc(CMakePackage):
     depends_on("llvm libcxx=none", when="platform=darwin", type="build")
     depends_on("llvm targets=arm,aarch64", when="target=arm:", type="build")
     depends_on("llvm targets=arm,aarch64", when="target=aarch64:", type="build")
+    depends_on("llvm@18:22", when="@1.30", type="build")
```
RMeli, line 62:
```suggestion
    depends_on("llvm@18:23", when="@1.30", type="build")
```
From the release notes LLVM 23.0 is also supported. Is the `when` range correct (i.e. restricted to only `1.30.z`)?

RMeli (approved):

repos/spack_repo/builtin/packages/ispc/package.py
albestro: I wanted to stay on the safe side, as other versions did (e.g. [1.23 accepts up to 18](https://github.com/ispc/ispc/blob/releases/1.23.x/src/ispc_version.h), but in [spack package accepts till 17](https://github.com/spack/spack-packages/blob/952fea6afabfe19f86112c656a1f7544cb28d8dc/repos/spack_repo/builtin/packages/ispc/package.py#L64) probably because of the development status at that time)


In principle we can switch to

```suggestion
    depends_on("llvm@18:23.0", when="@1.30", type="build")
```

RMeli: I think staying on the same size is OK, I was just checking.

PR #4442, opened 2026-04-22
fides: add 1.3.0

albestro:
Fides 1.3.0 has been recently released https://gitlab.kitware.com/vtk/fides/-/releases/v1.3.0 

Which is the first that uses `viskores` in place of `vtk-m` (~I added a `conflict` in `viskores` to avoid having both in the same concretization~ we opted for constraining ascent pre-viskores migration to use fides pre-viskores in order to match dependencies, see [CI error](https://gitlab.spack.io/spack/spack-packages/-/jobs/21614672#L485)).

This dependency was already correctly set up in a previous PR, so no further changes should be needed for it.

https://github.com/spack/spack-packages/blob/6845f1ce71052a4e8228a9b9b80c6615f1e54b81/repos/spack_repo/builtin/packages/fides/package.py#L39-L42

### DIFF
--- a/repos/spack_repo/builtin/packages/ascent/package.py
+++ b/repos/spack_repo/builtin/packages/ascent/package.py
@@ -310,7 +310,7 @@ class Ascent(CMakePackage, CudaPackage, ROCmPackage):
 
     # fides
     with when("+fides"):
-        depends_on("fides", when="+fides")
+        depends_on("fides")
         depends_on("fides@:1.2", when="@:0.9.5")
 
     # Adios2
--- a/repos/spack_repo/builtin/packages/ascent/package.py
+++ b/repos/spack_repo/builtin/packages/ascent/package.py
@@ -309,7 +309,9 @@ class Ascent(CMakePackage, CudaPackage, ROCmPackage):
     depends_on("occa", when="+occa")
 
     # fides
-    depends_on("fides", when="+fides")
+    with when("+fides"):
+        depends_on("fides", when="+fides")
+        depends_on("fides@:1.2", when="@:0.9.5")
 
     # Adios2
     depends_on("adios2", when="+adios2")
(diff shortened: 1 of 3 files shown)


kmorel:
I'm not sure making viskores and vtk-m conflict is the right thing to do. The two should be able to co-exist.

I think the real problem here is that the ascent package depends on both fides and vtk-m. I'm not sure why we are seeing the error where we are, but ascent cannot use fides 1.3 to read and write files. I think the proper solution is to change the ascent package to use the proper version of fides.

Replace

https://github.com/spack/spack-packages/blob/c10a65a479651eb49e74162ea8fbef7984000afa/repos/spack_repo/builtin/packages/ascent/package.py#L311-L312

with

```python
    # fides
    with when("+fides"):
      depends_on("fides")
      # Must match use of viskores/vtk-m with fides
      depends_on("fides@:1.2", when="@:0.9.5")
      # NOTE: Add dependency on "fides@1.3:" when moved to viskores
```

repos/spack_repo/builtin/packages/ascent/package.py
```diff
@@ -309,7 +309,9 @@ class Ascent(CMakePackage, CudaPackage, ROCmPackage):
     depends_on("occa", when="+occa")
 
     # fides
-    depends_on("fides", when="+fides")
+    with when("+fides"):
+        depends_on("fides", when="+fides")
```
kmorel: The `when` clause is redundant. It is already specified from the `with` condition.

```suggestion
        depends_on("fides")
```

That is not necessarily bad, but it makes the following line a bit confusing.

kmorel (approved):

repos/spack_repo/builtin/packages/ascent/package.py
albestro: good catch, it was an overlook. thanks!

RMeli (approved):

PR #4444, opened 2026-04-22
survey: add v1.1.4-v2.0.1

jgalarowicz:
Updates for new survey versions and additional dependencies


repos/spack_repo/builtin/packages/survey/package.py
```diff
@@ -153,7 +161,7 @@ def python_lib_dir(self):
     def site_packages_dir(self):
         return join_path(self.python_lib_dir, "site-packages")
 
-    def setup_run_environment(self, env: EnvironmentModifications) -> None:
+    def setup_run_environment(self, env):
```
RikkiButler20: Why remove the type hint?

jgalarowicz:
I guess I didn't understand if that was something old in our package and 
the newer spack didn't need it.

I will put it back.

Thanks for clarifying it is needed.

Jim G

RikkiButler20 (approved):
Looks good. Thanks.

RikkiButler20 (approved):

PR #4450, opened 2026-04-23
install shell completions for Taskwarrior and Timewarrior

taliaferro:
Following along with the way that other packages install their shell completions, this adds a `run_after(install)` hook called `install_completions` that copies the shell completion scripts for `task` and `timew` into the right place.


repos/spack_repo/builtin/packages/timew/package.py
```diff
@@ -23,3 +23,15 @@ class Timew(CMakePackage):
...
     depends_on("cxx", type="build")  # generated
+
+    @run_after("install")
+    def install_completions(self):
+
+        mkdirp(bash_completion_path(self.prefix))
+        copy("completion/timew-completion.bash", bash_completion_path(self.prefix))
+
+        mkdirp(zsh_completion_path(self.prefix))
+        copy("completion/timew.zsh", zsh_completion_path(self.prefix))
```
alecbcs, line 34: Do we need to copy this one as `_timew` instead of `timew.zsh` ?

bernhardkaindl:
@taliaferro, could you reply in the conversation of the review comment?

PR #4451, opened 2026-04-23
adios2: add version 2.12.0 + upper bound for cuda@13 support + patch aarch64

albestro:
Changelog:
- I dropped `2.12.0-rc1` in favour of the new `2.12.0` official release. Not sure if it was a good move, in case I can add it back (probably deprecated?)
- Since we have a newer official release, I also dropped the preference for `2.11.0` version.
- Added an upper bound for `cuda@13` conflict, now that 2.12.0 should support it
- Due to https://github.com/ornladios/ADIOS2/issues/5005#issuecomment-4302092413 I had to add also 4d2f27e8286be002a31b6438abc3b0098e4e41db with a conflict for target `aarch64`

On the latter point, in case a 2.12.1 patch version will be release, we can add it later.
For the moment on `aarch64` we can use
- `adios2@git.release_212=2.12.1`
- `adios2@git.98c51cc2207fd178d2f84f493d19710cf21f84c1=2.12.1`
- or simply `adios2@master`.

### DIFF
--- a/repos/spack_repo/builtin/packages/adios2/package.py
+++ b/repos/spack_repo/builtin/packages/adios2/package.py
@@ -30,14 +30,8 @@ class Adios2(CMakePackage, CudaPackage, ROCmPackage):
     license("Apache-2.0")
 
     version("master", branch="master")
-    version(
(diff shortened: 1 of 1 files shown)


vicentebolea:
#4351

PR #4469, opened 2026-04-23
dock: adding v6.13.1, reworking recipe

snehring:
New version, it's also available without a manual download now!

Reworking the recipe to bypass their configure script. I suspect this will work fine for about any compiler. The package limitations just seem to be 'these are the config templates we've made'. With this rework we make our own template from the gcc template as a base.

I don't have the time, but I am dangerously tempted to submit a PR upstream to switch this to cmake.

### DIFF
--- a/repos/spack_repo/builtin/packages/dock/package.py
+++ b/repos/spack_repo/builtin/packages/dock/package.py
@@ -18,13 +18,16 @@ class Dock(Package):
 
     homepage = "http://dock.compbio.ucsf.edu/DOCK_6/index.htm"
     url = "https://github.com/docking-org/dock6/archive/refs/tags/v6.13.1.tar.gz"
-    license("BSD-3-Clause", checked_by="snehring")
+
     maintainers("snehring")
+
+    license("BSD-3-Clause", checked_by="snehring")
 
     with when("@=6.9"):
         manual_download = True
 
     version("6.13.1", sha256="1bfe7ff777e60af6bff785dc74f9f7d1b0403c518c34ce4d1e6db98ac0865db1")
+
     # I don't think you can actually get 6.9 anymore since they moved to github
     version(
         "6.9",
--- a/repos/spack_repo/builtin/packages/dock/package.py
+++ b/repos/spack_repo/builtin/packages/dock/package.py
@@ -18,6 +18,7 @@ class Dock(Package):
 
     homepage = "http://dock.compbio.ucsf.edu/DOCK_6/index.htm"
     url = "https://github.com/docking-org/dock6/archive/refs/tags/v6.13.1.tar.gz"
+    license("BSD-3-Clause", checked_by="snehring")
     maintainers("snehring")
 
     with when("@=6.9"):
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/dock/package.py
```diff
@@ -17,48 +17,61 @@ class Dock(Package):
     site, and the associated binding energy."""
 
     homepage = "http://dock.compbio.ucsf.edu/DOCK_6/index.htm"
-    url = "file://{0}/dock.6.9_source.tar.gz".format(os.getcwd())
+    url = "https://github.com/docking-org/dock6/archive/refs/tags/v6.13.1.tar.gz"
+    license("BSD-3-Clause", checked_by="snehring")
     maintainers("snehring")
```
alecbcs:
```suggestion

    maintainers("snehring")
    
    license("BSD-3-Clause", checked_by="snehring")
```

Maintainers before license separated by a blank line.

repos/spack_repo/builtin/packages/dock/package.py
```diff
@@ -17,48 +17,61 @@ class Dock(Package):
...
-    manual_download = True
 
-    version("6.9", sha256="c2caef9b4bb47bb0cb437f6dc21f4c605fd3d0d9cc817fa13748c050dc87a5a8")
+    with when("@=6.9"):
+        manual_download = True
+
+    version("6.13.1", sha256="1bfe7ff777e60af6bff785dc74f9f7d1b0403c518c34ce4d1e6db98ac0865db1")
+    # I don't think you can actually get 6.9 anymore since they moved to github
```
alecbcs, line 31:
```suggestion

    # I don't think you can actually get 6.9 anymore since they moved to github
```

Add an extra blank line to separate versions.

alecbcs (changes requested):
A couple of quick suggestions, otherwise looks good to me.

repos/spack_repo/builtin/packages/dock/package.py
```diff
@@ -17,48 +17,64 @@ class Dock(Package):
...
 
-    version("6.9", sha256="c2caef9b4bb47bb0cb437f6dc21f4c605fd3d0d9cc817fa13748c050dc87a5a8")
+    license("BSD-3-Clause", checked_by="snehring")
+
+    with when("@=6.9"):
+        manual_download = True
+
+    version("6.13.1", sha256="1bfe7ff777e60af6bff785dc74f9f7d1b0403c518c34ce4d1e6db98ac0865db1")
+
+    # I don't think you can actually get 6.9 anymore since they moved to github
```
alecbcs, line 31: I'm tempted to say if v6.9 isn't available for download anymore we should drop it and any of the backwards compatibility fixes from the recipe here to simplify the package moving forward.

snehring: I'm inclined to agree, but people may have the older version sitting around (as we do) from when it was manual download only, and they may want that version for whatever reason. I think we can probably leave it deprecated for a while and then remove it.

alecbcs (approved):

PR #4476, opened 2026-04-23
libarchive: fix configure flags issues from issue 39381

jscook2345:
There are no with or without flags for libxml2 in libarchive. That package refers to them as xml2. This results in warnings that spack ignores during configure time and ultimately an incorrect build.

I've put in a fix below, but I also suggest a different patch in the issue.

See #39381


jscook2345:
I suggested another type of fix in the issue that you should consider because they fix the issue in two different ways. Not sure which is best, happy to do it based on your thoughts.

haampie (approved):

PR #4477, opened 2026-04-23
adding get_full_repo=True to all bioc packages that use git

snehring:
Changes to the way git fetches work have made bioc packages fail to fetch if they're not the absolute latest version in the bioc git repos. Adding `get_full_repo=True` to the versions works around this.

There may be a better way to solve this than adding `get_full_repo=True` to each version, but I'm not familiar enough with the inner workings to suggest one right now.

Going to open this as a draft, perhaps we can come up with a better solution.


snehring:
This is the result of a grep/sed one-liner so it's possible there are errors or omissions.

psakievich:
You can use `with default_args(get_full_repo=True):` as context manager to reduce the size of this diff.

snehring:
switched to with default_args

psakievich (approved):

PR #4492, opened 2026-04-24
vep: add perl_bio_ensembl* dependencies for v113

GKalmus:
As the title implies, added perl_bio_ensembl dependencies for vep/113.
Vep/113 will not work without those and will result in error.


teaguesterling (approved):
LGTM! Thank you for the update!

teaguesterling (approved):

hainest (changes requested):
None of the SHAs for the new versions is correct.

GKalmus:
> None of the SHAs for the new versions is correct.

Sorry for a long reply. I added the shasums with `curl -L <link> | sha256sum` and assumed these were correct sums. Now I pulled them from spack itself. Should be fine now!

I am new to pull requesting in general, please have mercy on me!

hainest:
> > None of the SHAs for the new versions is correct.
> 
> Sorry for a long reply. I added the shasums with `curl -L <link> | sha256sum` and assumed these were correct sums. 

I would have assumed that would work, as well. :shrug:

> Now I pulled them from spack itself. Should be fine now!

That's usually the safest way since some packages have special URL handling.

> I am new to pull requesting in general, please have mercy on me!

No worries! Thanks for doing the work to keep these packages updated.

hainest (approved):

hainest:
Thanks for the updates!

teaguesterling (approved):

PR #4493, opened 2026-04-24
CI: spack-core no longer needed for style checks

adamjstewart:
Follow-up to https://github.com/spack/spack-packages/pull/4023#discussion_r3138021903

PR #4496, opened 2026-04-24
Cleanup post ruff

kwryankrattiger:
Post-ruff merge cleanup.

Fixes bug where linting is applied to untouched files.
Removes now unnecessary clone of spack core from style check.

### DIFF
--- a/.ci/style_check.sh
+++ b/.ci/style_check.sh
@@ -26,11 +26,10 @@ if ! python_files > /dev/null; then
   info "skipping style checks: no Python files changed"
   exit 0
 fi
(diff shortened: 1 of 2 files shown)


adamjstewart:
Partial duplicate of #4493, that should be safe to merge.

You never responded to my comment in https://github.com/spack/spack-packages/pull/4023#discussion_r3138130449

adamjstewart (approved):
Perhaps a future merge queue would allow ruff to reliably run on all files, but this is definitely safer and slightly faster (1 s -> 200 ms) when running on 1 file instead of 10k files.

kwryankrattiger:
Pipeline is passing now, spurious failure: 
https://gitlab.spack.io/spack/spack-packages/-/pipelines/1547511

PR #4501, opened 2026-04-25
pplacer: add v1.1.alpha22, rework package

snehring:
Version bump and transition to source build vs downloading prebuilt binaries.


hainest (approved):
Confirmed SHA for new version.

PR #4502, opened 2026-04-26
cuDSS: Add v0.7.1

RemiLacroix-IDRIS:


RikkiButler20 (approved):
LGTM

RemiLacroix-IDRIS:
@ddement : I was thinking to add support for the CUDA 13 versions of cuDSS. What do you think of doing something similar to cuDDN (i.e. `<cuDSS_version>-<cuda_version>`)?

Unfortunately I don't think there is a way to avoid adding the CUDA version in cuDSS version currently.

PR #4503, opened 2026-04-26
nvpl: new package

rbberger:
New attempt to add `nvpl` package.

@RMeli @fspiga @albestro @rasolca FYI

Related to https://github.com/spack/spack/pull/48732

### DIFF
--- a/repos/spack_repo/builtin/packages/nvpl/package.py
+++ b/repos/spack_repo/builtin/packages/nvpl/package.py
@@ -1,0 +1,136 @@
+# Copyright Spack Project Developers. See COPYRIGHT file for details.
+#
+# SPDX-License-Identifier: (Apache-2.0 OR MIT)
+
+from spack_repo.builtin.build_systems.generic import Package
+
+from spack.package import *
+
+
+class Nvpl(Package):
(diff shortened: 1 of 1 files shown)


rbberger:
@spackbot fix style

rbberger:
@eugeneswalker does it make sense to change the E4S neoverse v2 stack here and see how far we get, or should that be its own PR?

hainest (dismissed):
Confirmed SHA.

@eugeneswalker Any thoughts on Richard's [question](https://github.com/spack/spack-packages/pull/4503#issuecomment-4328627606) about E4S?

eugeneswalker:
> @eugeneswalker does it make sense to change the E4S neoverse v2 stack here and see how far we get, or should that be its own PR?

Sorry, what change is being proposed? I think either way is fine, although, depending on how many issues there are perhaps it makes sense to get the package merged first so it is available and follow up with separate PR to update E4S neoverse stack.

eugeneswalker (dismissed):

rbberger:
Moving this back to draft. We prefer individual Spack recipes for the libraries. I'll update the other packages.

PR #4509, opened 2026-04-27
arbor: add verions 0.12.2 to 0.10.0

thorstenhater:
Add versions 10, 11, 12 of Arbor.

### DIFF
--- a/repos/spack_repo/builtin/packages/arbor/package.py
+++ b/repos/spack_repo/builtin/packages/arbor/package.py
@@ -19,31 +19,27 @@ class Arbor(CMakePackage, CudaPackage):
     maintainers = ("thorstenhater",)
     submodules = True
 
-    version("master", branch="master", submodules=True)
-    version("develop", branch="master", submodules=True)
+    version("master", branch="master")
+    version("develop", branch="master")
     version(
         "0.12.1",
         sha256="6584ccfaef1c8f9eab6c8794d0f4f06745cbfb7c2d701815c74ed9e6a101a3dc",
         url="https://github.com/arbor-sim/arbor/archive/refs/tags/v0.12.1.tar.gz",
-        submodules=True,
     )
     version(
         "0.11.0",
         sha256="6df68b308dd629df993eda40319676cd43407ae211d0846100b0cf42e8c9ad22",
         url="https://github.com/arbor-sim/arbor/archive/refs/tags/v0.11.0.tar.gz",
-        submodules=True,
     )
     version(
         "0.10.0",
-        sha256="6b6cc900b85fbf833fae94817b9406a0d690dc28",
+        sha256="72966b7a2f45ce259b8ba167ca3e4f5ab9f212136a300267aaac0c04ed3fe3fc",
         url="https://github.com/arbor-sim/arbor/releases/download/v0.10.1/arbor-v0.10.0-full.tar.gz",
-        submodules=True,
     )
     version(
         "0.9.0",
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/arbor/package.py
```diff
@@ -13,13 +14,28 @@ class Arbor(CMakePackage, CudaPackage):
 
     homepage = "https://arbor-sim.org"
     git = "https://github.com/arbor-sim/arbor.git"
-    url = "https://github.com/arbor-sim/arbor/releases/download/v0.8.1/arbor-v0.8.1-full.tar.gz"
-
-    maintainers("thorstenhater", "brenthuisman")
+    url = "https://github.com/arbor-sim/arbor/archive/refs/tags/v0.12.1.tar.gz"
```
brenthuisman: The Spack CI was removed. I'm sure you checked it builds successfully.

I can't think if a reason why it'd be a problem, but I'm looking for a double check that CPM and Spack play nice.

repos/spack_repo/builtin/packages/arbor/package.py
```diff
@@ -13,13 +14,28 @@ class Arbor(CMakePackage, CudaPackage):
...
     license("BSD-3-Clause")
+    maintainers = ("thorstenhater",)
+    submodules = True
 
     version("master", branch="master")
-    version("develop")
+    version("develop", branch="master")
```
brenthuisman, line 22: I guess this is what you want, but for reference: the CI job would build from a local copy of arbor's source, and therefore no branch would need to be named, because that would override the local copy (which you'd want to test before merging at master).

brenthuisman (approved):

repos/spack_repo/builtin/packages/arbor/package.py
```diff
@@ -1,4 +1,5 @@
-# Copyright Spack Project Developers. See COPYRIGHT file for details.
+# Copyright 2013-2023 Lawrence Livermore National Security, LLC and other
+# Spack Project Developers. See the top-level COPYRIGHT file for details.
```
alecbcs: We shouldn't be reverting the license here to the old version.

repos/spack_repo/builtin/packages/arbor/package.py
thorstenhater: This PR is running through Spack's CI which I understand builds packages

repos/spack_repo/builtin/packages/arbor/package.py
thorstenhater: Yes, true. I'll fix it.

repos/spack_repo/builtin/packages/arbor/package.py
brenthuisman: AFAIK it only tests if Spack builds. Can't find Arbor being built looking under 'Checks'.

thorstenhater: Ok, then I'll double check that. Thanks.

bernhardkaindl:
@thorstenhater, this looks like there's a typo, I guess the 2nd line should CUDA 13? Otherwise it is narrowing the first line and you would remove the 1st line as dead code:
```py
        depends_on("cuda@12:", when="@0.9.1:")
        depends_on("cuda@12:", when="@0.10.0:")
```

bernhardkaindl (changes requested):
Updating the status of the PR to changes requested for visibility.

PR #4520, opened 2026-04-27
r: add v4.6.0

afuetterer:
This PR adds 4.6.0 of R. I tested the build locally.

Ref: https://cran.r-project.org/src/base/R-4

Sub tasks:

- [x] rebase branch
- [x] add 4.6.1
- [x] set 4.5.3 to preferred
- [x] texinfo dependency

### DIFF
--- a/repos/spack_repo/builtin/packages/r/package.py
+++ b/repos/spack_repo/builtin/packages/r/package.py
@@ -27,6 +27,7 @@ class R(AutotoolsPackage):
     license("GPL-2.0-or-later")
 
     version("trunk", svn="https://svn.r-project.org/R/trunk")
+    version("4.6.0", sha256="b8dc9b4543660c7b596b87938df532394350360976527d344228ee0ed12e45ec")
     version("4.5.3", sha256="aa5c1ed4293c7271ac513d654670356ac0e8a6ad5e42be014365d11150b5b8f2")
     version("4.5.2", sha256="0d71ff7106ec69cd7c67e1e95ed1a3cee355880931f2eb78c530014a9e379f20")
     version("4.5.1", sha256="b42a7921400386645b10105b91c68728787db5c4c83c9f6c30acdce632e1bb70")
--- a/repos/spack_repo/builtin/packages/r/package.py
+++ b/repos/spack_repo/builtin/packages/r/package.py
@@ -256,6 +256,10 @@ class R(AutotoolsPackage):
         env.set("MAKEFLAGS", "-j{0}".format(make_jobs))
         env.set("R_HOME", join_path(self.prefix, "rlib", "R"))
 
+    def setup_build_environment(self, env: EnvironmentModifications) -> None:
+        # Standardize the timezone for the builds
+        env.set("TZ", "UTC")
+
     def setup_dependent_run_environment(
         self, env: EnvironmentModifications, dependent_spec: Spec
     ) -> None:
--- a/repos/spack_repo/builtin/packages/r/package.py
+++ b/repos/spack_repo/builtin/packages/r/package.py
@@ -29,7 +29,11 @@ class R(AutotoolsPackage):
     version("trunk", svn="https://svn.r-project.org/R/trunk")
     version("4.6.1", sha256="4da6e61d2c0aac5f14a2e7e432cb5fcc269efe83da4293050ba7f03dff4e2cf4")
     version("4.6.0", sha256="b8dc9b4543660c7b596b87938df532394350360976527d344228ee0ed12e45ec")
-    version("4.5.3", sha256="aa5c1ed4293c7271ac513d654670356ac0e8a6ad5e42be014365d11150b5b8f2")
+    version(
+        "4.5.3",
+        sha256="aa5c1ed4293c7271ac513d654670356ac0e8a6ad5e42be014365d11150b5b8f2",
+        preferred=True
+    )
     version("4.5.2", sha256="0d71ff7106ec69cd7c67e1e95ed1a3cee355880931f2eb78c530014a9e379f20")
     version("4.5.1", sha256="b42a7921400386645b10105b91c68728787db5c4c83c9f6c30acdce632e1bb70")
     version("4.5.0", sha256="3b33ea113e0d1ddc9793874d5949cec2c7386f66e4abfb1cef9aec22846c3ce1")
--- a/repos/spack_repo/builtin/packages/r/package.py
+++ b/repos/spack_repo/builtin/packages/r/package.py
@@ -55,6 +55,7 @@ class R(AutotoolsPackage):
     depends_on("cxx", type="build")
     depends_on("fortran", type="build")
     depends_on("findutils", type="build")
+    depends_on("texinfo@6.8:", when="@4.6:")
     depends_on("texinfo", type="build")
 
     depends_on("blas")
--- a/repos/spack_repo/builtin/packages/r/package.py
+++ b/repos/spack_repo/builtin/packages/r/package.py
@@ -27,6 +27,7 @@ class R(AutotoolsPackage):
     license("GPL-2.0-or-later")
 
     version("trunk", svn="https://svn.r-project.org/R/trunk")
+    version("4.6.1", sha256="4da6e61d2c0aac5f14a2e7e432cb5fcc269efe83da4293050ba7f03dff4e2cf4")
     version("4.6.0", sha256="b8dc9b4543660c7b596b87938df532394350360976527d344228ee0ed12e45ec")
     version("4.5.3", sha256="aa5c1ed4293c7271ac513d654670356ac0e8a6ad5e42be014365d11150b5b8f2")
     version("4.5.2", sha256="0d71ff7106ec69cd7c67e1e95ed1a3cee355880931f2eb78c530014a9e379f20")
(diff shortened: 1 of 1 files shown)


afuetterer:
Not sure why gitlab-ci failed.

kshea21 (dismissed):
Thanks!

afuetterer:
Still not sure what is going on exactly.

https://gitlab.spack.io/spack/spack-packages/-/jobs/22043073#L2274

Is this causing the build error on GitLab CI? Missing TZ env?

What to do next @kshea21?

kshea21:
@spackbot re-run pipeline

emwjacobson:
Sorry, just wanted to bump this if possible. Would like to get R/4.6.0 installed w/o needing to make our own fork :)

kshea21:
Hi! I'm so sorry this fell off of my radar! We've been having some problems with the CI pipelines lately, hoping that they're resolved and this new kickoff goes through. Will check back in tomorrow morning!!

afuetterer:
Still pending gitlab CI job. Do you need to trigger that somehow?

alecbcs:
@afuetterer it looks like the build is failing in CI because some of our runners are missing the `TZ` environment variable. Can you try adding this method to the package to try standardizing the timezone for the builds.

```python
    env.set("TZ", "UTC")
```

afuetterer:
Yes, will do. Thanks for finding this one out.

afuetterer:
Rebased to develop and added the `setup_build_environment` as suggested by @alecbcs. Fingers crossed.

kshea21:
@spackbot run pipeline

kshea21:
@spackbot re-run pipeline

alecbcs:
@spackbot run pipeline

emwjacobson:
In case it's of any help getting this pushed through, r-rcpp seems to be failing due to them using some now-removed features in r@4.6.0.

https://github.com/RcppCore/Rcpp/issues/1473

Within r-rcpp's, limiting R's version fixes it.
```diff
-    depends_on("r@3.5.0:", when="@1.1.1:", type=("build", "run"))
+    depends_on("r@3.5.0:4.5", when="@1.1.1:", type=("build", "run"))
```

Version 1.1.1-1 also supposedly fixes this, but I haven't done any testing on it.

afuetterer:
What else can be done here to get gitlab ci to pass?

Not sure what @emwjacobson is suggesting. That does affect other packages downstream, not this one itself, right?

bernhardkaindl (changes requested):
@afuetterer:

> Not sure what @emwjacobson is suggesting. That does affect other packages downstream, not this one itself, right?

Correct. The impact is on downstream packages rather than this package itself.

When adding a new package version (or opening any PR), it's important to look into any failing GitLab CI pipelines before requesting review. With the current review backlog, reviewers and maintainers unfortunately can't investigate CI failures for every open PR, so having those addressed first makes the review process much smoother.

@emwjacobson has already outlined the steps needed if you'd like to make this new version the default.

The main challenge is that, when no specific R version has been built yet, the newest version (currently 4.6.0) will be selected if possible. Since R 4.x releases can introduce backwards-incompatible changes, we need to make sure that downstream R packages continue to build correctly.

In practice, this means checking not only the R packages covered by CI, but ideally every R package recipe in the repository. To fully validate making R 4.6 the default, it's also important to consider older versions of those packages. Otherwise, users building an older package version that isn't compatible with R 4.6 could encounter build failures because the solver selects the newer R version.

As a result, the R version constraints for existing packages (and their maintained versions) need to be reviewed and updated where appropriate.

This isn't unique to R—similar considerations apply when introducing new versions of tools like CMake. A common approach is to initially avoid making the new version the preferred one. In Spack, this can be done by leaving the existing preferred version in place and not setting preferred=True on the new version() entry until compatibility has been validated. There are many examples of this pattern throughout the repository.

It might also be helpful to collaborate with @w8jcik, who has done a lot of work on R packages and would likely have valuable input on the best approach. I suggest discussing the overall direction with them before proceeding.

In the meantime, it would probably make sense to mark this PR as a draft until the compatibility work is complete and local package builds of the downstream R packages are in a good state. That way, you'll get faster results as well.

emwjacobson:
Just a heads up, I pushed an update to `r-rcpp` in #5790 with newer versions that would theoretically support r@4.6.0.

I technically approached the r-rcpp change by adding a `conflicts` instead of modifying the `depends_on` - not sure what the most favorable approach really is.

w8jcik:
@bernhardkaindl Thank you for the recommendation. I will try to test Spack R packages against R 4.6 in a systematic way.

w8jcik:
This should give all the packages that are using R

```sh
git clone --depth 1 https://github.com/spack/spack-packages.git

cat \
    <( cd spack-packages/repos/spack_repo/builtin/packages && grep -r "RPackage" | cut -d'/' -f 1 | sed 's|_|-|g' | sort -u ) \
    <( cd spack-packages/repos/spack_repo/builtin/packages && { cat <(grep -r "depends_on(\"r\"" | cut -d'/' -f 1) <(grep -r "depends_on(\"r-" | cut -d'/' -f 1) ; } | sed 's|_|-|g' | sort -u ) \
    | sort -u > r-packages.txt
```

* [r-packages.txt](https://github.com/user-attachments/files/30600750/r-packages.txt)

A brute-force approach is to install all of them with `--test=root` for both versions of R (4.5 and 4.6) and analyze failing installations.

repos/spack_repo/builtin/packages/r/package.py
w8jcik, line 59:
```diff
+ depends_on("texinfo@6.8:", when="@4.6:")
  depends_on("texinfo", type="build")
```

Image

w8jcik (changes requested):
> What else can be done here to get gitlab ci to pass?
> Not sure what @emwjacobson is suggesting.

* @afuetterer could you press update button (through rebase) in GitHub, then the branch will get fix for `r-rcpp` from @emwjacobson and the pipeline should pass.
* Dependency on `texinfo` needs to be corrected.
* There is one more version available
  ```diff
  + version("4.6.1", sha256="4da6e61d2c0aac5f14a2e7e432cb5fcc269efe83da4293050ba7f03dff4e2cf4")
  ```
* Multiple packages are failing with the update so I would suggest to add preferred next to version 4.5. I will follow with more pull requests with constraints before selecting 4.6 as the default.

  ```diff
  - version("4.5.3", sha256="aa5c1ed4293c7271ac513d654670356ac0e8a6ad5e42be014365d11150b5b8f2")
  + version(
  +     "4.5.3",
  +     sha256="aa5c1ed4293c7271ac513d654670356ac0e8a6ad5e42be014365d11150b5b8f2",
  +     preferred=True
  + )
  ```

image

(still running)

afuetterer:
> @afuetterer:
> 
> > Not sure what @emwjacobson is suggesting. That does affect other packages downstream, not this one itself, right?
> 
> Correct. The impact is on downstream packages rather than this package itself.
> 
> When adding a new package version (or opening any PR), it's important to look into any failing GitLab CI pipelines before requesting review. With the current review backlog, reviewers and maintainers unfortunately can't investigate CI failures for every open PR, so having those addressed first makes the review process much smoother.
> 
> @emwjacobson has already outlined the steps needed if you'd like to make this new version the default.
> 
> The main challenge is that, when no specific R version has been built yet, the newest version (currently 4.6.0) will be selected if possible. Since R 4.x releases can introduce backwards-incompatible changes, we need to make sure that downstream R packages continue to build correctly.
> 
> In practice, this means checking not only the R packages covered by CI, but ideally every R package recipe in the repository. To fully validate making R 4.6 the default, it's also important to consider older versions of those packages. Otherwise, users building an older package version that isn't compatible with R 4.6 could encounter build failures because the solver selects the newer R version.
> 
> As a result, the R version constraints for existing packages (and their maintained versions) need to be reviewed and updated where appropriate.
> 
> This isn't unique to R—similar considerations apply when introducing new versions of tools like CMake. A common approach is to initially avoid making the new version the preferred one. In Spack, this can be done by leaving the existing preferred version in place and not setting preferred=True on the new version() entry until compatibility has been validated. There are many examples of this pattern throughout the repository.
> 
> It might also be helpful to collaborate with @w8jcik, who has done a lot of work on R packages and would likely have valuable input on the best approach. I suggest discussing the overall direction with them before proceeding.
> 
> In the meantime, it would probably make sense to mark this PR as a draft until the compatibility work is complete and local package builds of the downstream R packages are in a good state. That way, you'll get faster results as well.

Thanks @bernhardkaindl. As I said I tested the build locally succesfully and could not understand why the GitLab CI was failing on my own, that is why I asked for help.

afuetterer:
Thanks a lot @w8jcik. I added all of your suggestions to the PR and set it to draft as @bernhardkaindl suggested.

w8jcik:
> and set it to draft

With preferred set to 4.5.3 I think there is no reason to make it a draft. If the pipeline passes it could be merged.

I will follow up with more pull requests adding version constraints.

afuetterer:
Thank you @w8jcik!

bernhardkaindl (approved):
Approving with the approval of @w8jcik.

It enables building r- packages with ^r@4.6 when required by packages,
or requested by the spec and the default stays with the latest r@4.5 release to not break the build of packages with missing version contraints for r@4.6.

It also allows testing R packages with the r@4.6 to add version constraints for packages which need them.

PR #4506, opened 2026-04-26
 expat: 2.8.0 + deprecate vulnerable 2.7.5 (for CVE-2026-41080)

hartwork:
Similar to #4113

Related blog post: https://blog.hartwork.org/posts/expat-2-8-0-released/

CVE-2026-41080

PR #4522, opened 2026-04-27
uriparser: 1.0.1 + deprecate vulnerable 0.9.6 and 0.9.7 (for CVE-2026-42371)

hartwork:
Similar to #4506

CVE-2026-42371

### DIFF
--- a/repos/spack_repo/builtin/packages/uriparser/package.py
+++ b/repos/spack_repo/builtin/packages/uriparser/package.py
@@ -9,13 +9,17 @@ from spack.package import *
 
 class Uriparser(CMakePackage):
     """uriparser is a strictly RFC 3986 compliant URI parsing and handling
-    library written in C89 ("ANSI C")."""
+    library written in C99."""
 
     homepage = "https://uriparser.github.io/"
     url = "https://github.com/uriparser/uriparser/releases/download/uriparser-0.9.3/uriparser-0.9.3.tar.gz"
 
     license("BSD-3-Clause")
 
+    version(
+        "1.0.1",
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/uriparser/package.py
RikkiButler20, line 39:
```suggestion

    depends_on("cmake@3.15:", type="build", when="@1.0.1:")
```
Update the cmake minimum version requirement.
https://github.com/uriparser/uriparser/blob/2027ca1cf44f20ae43570351599251f633138d51/CMakeLists.txt#L37

RikkiButler20 (dismissed):

repos/spack_repo/builtin/packages/uriparser/package.py
hartwork: @RikkiButler20 good point :+1: — done, pushed

RikkiButler20: Should we keep the `cmake@3.3` dependency for versions `0.9.7` and `0.9.6`?

hartwork: @RikkiButler20 the other two are vulnerable and deprecated so I'd personally vote to not increase complexity for these two zombie ones and just use CMake >=3.15 (released already in 2019) for everything. I'd also be good with dropping the two older ones altogether, personally. But I can make it do other things if you prefer.

RikkiButler20: Ok. We can keep the deprecated versions and `cmake@3.15:`

RikkiButler20 (approved):
LGTM. Thanks

PR #4528, opened 2026-04-27
py_pyfive: New package

LydDeb:

### DIFF
--- a/repos/spack_repo/builtin/packages/py_pyfive/package.py
+++ b/repos/spack_repo/builtin/packages/py_pyfive/package.py
@@ -27,4 +27,6 @@ class PyPyfive(PythonPackage):
         depends_on("py-setuptools-scm@8:")
     with default_args(type=("build", "run")):
         depends_on("py-numpy@2:")
+        # Not in pyproject.toml but necessary to import the 'pyfive' module.
+        # This dependence appears in high_level.py on line 13
         depends_on("py-typing-extensions")
--- a/repos/spack_repo/builtin/packages/py_pyfive/package.py
+++ b/repos/spack_repo/builtin/packages/py_pyfive/package.py
@@ -1,0 +1,30 @@
+# Copyright Spack Project Developers. See COPYRIGHT file for details.
+#
+# SPDX-License-Identifier: (Apache-2.0 OR MIT)
+
+from spack_repo.builtin.build_systems.python import PythonPackage
+
+from spack.package import *
+
+
+class PyPyfive(PythonPackage):
+    """A pure python HDF5 reader"""
+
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/py_pyfive/package.py
```diff
@@ -0,0 +1,30 @@
...
+    version("1.1.1", sha256="a43e852947b9bc4703bc174d4e9cec77d61c2324d35961bbae9914aa69821ce5")
+
+    depends_on("python@3.10:", type=("build", "run"))
+
+    with default_args(type="build"):
+        depends_on("py-setuptools@80:")
+        depends_on("py-setuptools-scm@8:")
+    with default_args(type=("build", "run")):
+        depends_on("py-numpy@2:")
+        depends_on("py-typing-extensions")
```
manuelakuhn, line 32: This dependency is not needed for version `1.1.1` (https://github.com/NCAS-CMS/pyfive/blob/v1.1.1/pyproject.toml#L38) it is only there in `main`.

LydDeb: You are correct, however this dependency is necessary to import the `pyfive` module:

https://github.com/NCAS-CMS/pyfive/blob/026e388f33a6b56d23eea7aba327cae9ea7bf785/pyfive/high_level.py#L13

Without this dependency, we get the following error, for example:

```
>>> import pyfive
  File "<python-input-0>", line 1, in <module>
    import pyfive
  File "/spack/var/spack/environments/pyfive/.spack-env/view/lib/python3.14/site-packages/pyfive/__init__.py", line 7, in <module>
    from pyfive.high_level import File, Group, Dataset
  File "/spack/var/spack/environments/pyfive/.spack-env/view/lib/python3.14/site-packages/pyfive/high_level.py", line 13, in <module>
    from typing_extensions import Self  # Python 3.10-compat
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
ModuleNotFoundError: No module named 'typing_extensions'
```

manuelakuhn: OK, thanks. Let me know in case you want to add a comment (to prevent that mix up in the future).

manuelakuhn (dismissed):

repos/spack_repo/builtin/packages/py_pyfive/package.py
LydDeb: Good idea, I just added a comment.

manuelakuhn: Thank you

manuelakuhn (approved):

PR #4531, opened 2026-04-27
nvtx: support libc without char8_t

cwsmith:
This PR addresses a compilation issue with the nvtx package with the python bindings enabled (as a dep of py-torch) when built with GCC 15.2.0 and glibc 2.34 on a RHEL9 system.  IIUC, glibc 2.36 or newer is needed to use char8_t from the c20 standard (https://lists.gnu.org/archive/html/info-gnu/2022-08/msg00000.html).

A PR to the nvtx repo is here: https://github.com/NVIDIA/NVTX/pull/163

On our system, the following code reproduces the bug:

```
$ cat foo.c 
#include <stdio.h> 
//---- uncomment the following block to compile ----
//#if defined(__STDC_VERSION__) && __STDC_VERSION__ >= 202311L
//#ifndef __cpp_char8_t
//typedef unsigned char char8_t;
//#endif
//#endif
int main() { char8_t a; printf("%ld\n", (long)__STDC_VERSION__); return 0; }

$ gcc --version
gcc (Spack GCC) 15.2.0
Copyright (C) 2025 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.

##  show glibc version
$ ldd --version   
ldd (GNU libc) 2.34
Copyright (C) 2021 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
Written by Roland McGrath and Ulrich Drepper.

$ gcc foo.c -o foo
foo.c: In function 'main':
foo.c:9:14: error: unknown type name 'char8_t'; did you mean 'char'?
    9 | int main() { char8_t a; printf("%ld\n", (long)__STDC_VERSION__); return 0; }
      |              ^~~~~~~
      |              char
```

### DIFF
--- a/repos/spack_repo/builtin/packages/nvtx/package.py
+++ b/repos/spack_repo/builtin/packages/nvtx/package.py
@@ -50,9 +50,7 @@ class Nvtx(Package, PythonExtension):
         setup.filter("include_dirs=include_dirs", f"include_dirs=['{include_dir}']", string=True)
         # see comment in patch below
         if self.spec.satisfies("@3.3.0:"):
-            header = FileFilter(
-                "c/include/nvtx3/nvtxDetail/nvtxExtPayloadTypeInfo.h"
-            )
+            header = FileFilter("c/include/nvtx3/nvtxDetail/nvtxExtPayloadTypeInfo.h")
             header.filter(
                 "#define NVTX_HAVE_CHAR8 1",
                 "/* Fallback for systems where glibc < 2.36 does not define char8_t\n"
@@ -74,4 +72,3 @@ class Nvtx(Package, PythonExtension):
 
         with working_dir(self.build_directory):
             pip(*PythonPipBuilder.std_args(self), f"--prefix={self.prefix}", ".")
-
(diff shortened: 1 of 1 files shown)


thomas-bouvier (approved):
LGTM, thanks for your contribution.

becker33 (approved):

becker33:
It looks like this is trying to build python components even when `~python`.

I think it's missing a conditional around the `pip` call for whether python is enabled.

sethrj (changes requested):
@becker33 Shouldn't this be using the `.patch` file from the GitHub PR rather than a file filter?

cwsmith:
This is fixed upstream with [ba4ec76](https://github.com/NVIDIA/NVTX/commit/ba4ec76719bcbdc6225a7edd2dd65a2a360a96b2) in the v3 release of NVTX: https://github.com/NVIDIA/NVTX/commits/release-v3/ .

PR #4533, opened 2026-04-27
libvdwxc bugfix

alikhamze:
`libvdwxc` currently does not correctly get the path of the `fftw-api` provider. This small PR fixes that.
My guess is that when this line was added in https://github.com/spack/spack-packages/commit/5f376d73e480ac6d71f789938a950083820c3635 that the committer had `fftw` installed, so this didn't fail.

Tested by successfully building `libvdwxc ^amdfftw` (which was failing before, which was how I found the issue).

### DIFF
--- a/repos/spack_repo/builtin/packages/libvdwxc/package.py
+++ b/repos/spack_repo/builtin/packages/libvdwxc/package.py
@@ -45,7 +45,9 @@ class Libvdwxc(AutotoolsPackage):
         spec = self.spec
 
         args = [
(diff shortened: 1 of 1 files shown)


alikhamze:
@spackbot fix style

mtaillefumier (approved):
I agree with you. I mostly targeted fftw not fftw-api. Thanks for correcting this

RikkiButler20 (approved):

PR #4534, opened 2026-04-27
dcm2niix: add jpeg2000 variant

Markus92:
One of my users asked for JPEG2000 support, so I took the time to overhaul dcm2niix.

The superbuild pulls in a lot of third party stuff but Spack is better at managing that, so I only compile the console app (the previous version resulted in the same output, too, so nothing really lost).

### DIFF
--- a/repos/spack_repo/builtin/packages/dcm2niix/package.py
+++ b/repos/spack_repo/builtin/packages/dcm2niix/package.py
@@ -14,7 +14,7 @@ class Dcm2niix(CMakePackage):
     homepage = "https://github.com/rordenlab/dcm2niix"
     url = "https://github.com/rordenlab/dcm2niix/archive/refs/tags/v1.0.20220720.tar.gz"
 
-    license("BSD-3-Clause", checked_by="Markus92")
+    license("BSD-3-Clause AND MIT", checked_by="Markus92")
 
     version(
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/dcm2niix/package.py
```diff
@@ -14,7 +14,7 @@ class Dcm2niix(CMakePackage):
     homepage = "https://github.com/rordenlab/dcm2niix"
     url = "https://github.com/rordenlab/dcm2niix/archive/refs/tags/v1.0.20220720.tar.gz"
 
-    license("Zlib")
+    license("BSD-3-Clause", checked_by="Markus92")
```
becker33: I don't think this license statement is quite correct -- it looks like they've vendored in files (`Console/ujpeg.{cpp,h}`) that are under the MIT license as well.


```suggestion
    license("BSD-3-Clause AND MIT", checked_by="Markus92")
```

becker33 (changes requested):

repos/spack_repo/builtin/packages/dcm2niix/package.py
Markus92: Fair enough! I was like, wait a minute, this doesn't look like a zlib license at all, but forgot to check the rest.

becker33 (approved):

PR #4538, opened 2026-04-28
llvm: bump update as `libllvm` provider

albestro:
I might be missing something, but I'd say that it should be updated also as `libllvm` provider.


repos/spack_repo/builtin/packages/llvm/package.py
```diff
@@ -268,6 +268,8 @@ class Llvm(CMakePackage, CudaPackage, LlvmDetection, CompilerPackage):
 
     variant("utils", default=False, description="Install utility binaries (FileCheck, etc.)")
 
+    provides("libllvm@22", when="@22.0.0:22")
+    provides("libllvm@21", when="@21.0.0:21")
```
albestro: I'm wondering if we could just say like this to express the same thing...

```suggestion
    provides("libllvm@22", when="@22")
    provides("libllvm@21", when="@21")
```

rbberger: @albestro Good catch. Yes. Make it so!

rbberger: And then just do a `for v in range(3, 23)`.

albestro: just did it in https://github.com/spack/spack-packages/pull/4538/commits/f63658dd84006014c3f76f28a1cef5cde1086c0f but for range `[5,23)`, considering that:
- version 5 is the oldest version listed and
- 23 is not listed yet.

do you think is ok, or would you prefer to have it starting from 3 to exactly map what was there before?

rbberger: LGTM

rbberger (dismissed):

albestro:
For some reason the `for ... provides` in https://github.com/spack/spack-packages/pull/4538/commits/4c115bd653e7a146675b7433c605ecaf355c95ed triggers a fetch error. Not sure why, but I reverted the iteration while keeping the range simplification (i.e. `@22` instead of `@22.0.0:22`) and at least locally it worked. Let's see if CI works too.

RMeli (approved):

PR #4539, opened 2026-04-28
n2p2: C++ standard to C++14 and add py-setuptools dep

teabagk7:
1. Change cxx11_flag to cxx14_flag for Eigen 5.x compatibility. Eigen 5.x uses std::enable_if_t which requires C++14.
2. Add py-setuptools build dependency for Python 3.12+ where distutils was removed from stdlib.

### DIFF
--- a/repos/spack_repo/builtin/packages/n2p2/package.py
+++ b/repos/spack_repo/builtin/packages/n2p2/package.py
@@ -42,6 +42,7 @@ class N2p2(MakefilePackage):
     depends_on("doxygen", type="build", when="+doc")
     depends_on("texlive", type="build", when="+doc")
     depends_on("py-cython", type=("build", "run"))
+    depends_on("py-setuptools", type="build")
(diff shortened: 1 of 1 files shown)


hainest:
>Change cxx11_flag to cxx14_flag for Eigen 5.x compatibility. Eigen 5.x uses std::enable_if_t which requires C++14.

Since there's no version constraint on the Eigen dependency, should the change in the Makefile be done conditionally? Or is it safe to build older versions of eigen with c++14?

teabagk7:
> > Change cxx11_flag to cxx14_flag for Eigen 5.x compatibility. Eigen 5.x uses std::enable_if_t which requires C++14.
> 
> Since there's no version constraint on the Eigen dependency, should the change in the Makefile be done conditionally? Or is it safe to build older versions of eigen with c++14?

In my understanding it's safe. C++14 is fully backward compatible with C++11. Older Eigen versions are unaffected by the higher standard.

hainest (approved):

PR #4540, opened 2026-04-28
conquest: fix oneAPI compiler support

teabagk7:
Two fixes for building with Intel oneAPI compilers (ifx):
1. Make -fallow-argument-mismatch conditional on %gcc@10: only, since it's a GCC-specific flag not recognized by ifx.
2. Remove hardcoded -lscalapack from LIBS line, since MKL provides scalapack as libmkl_scalapack_lp64.so (already linked via scalapack.libs.ld_flags in LINKFLAGS).


davidbowler (approved):

RikkiButler20 (approved):
LGTM

PR #4548, opened 2026-04-28
hdf-eos2: fix macOS build for version 3.0

mathomp4:
Four issues fixed, all found when building on macOS (darwin/arm64):

1. GCTP not linked (`@3.0` only): the source tarball ships `Makefile.in` with an empty `libhdfeos_la_LIBADD`, so the bundled `gctp/src/libGctp.la` is never linked into `libhdfeos`. Patch both `Makefile.am` and `Makefile.in` before configure to add `libhdfeos_la_LIBADD=$(LIBGCTP)`. The bug is masked on Linux where the dynamic linker permits unresolved symbols by default.
2. `h4cc` breaks configure (all versions): `CC=h4cc` causes autoconf's cross-compile detection to fail because `h4cc` has a hardcoded `CCBASE` that bypasses Spack wrappers. Override `CC` with the Spack compiler wrapper during configure, then let `h4cc` take over for build. Also set `HDF4_CC`/`HDF4_CLINKER` so `h4cc` itself uses the Spack wrapper.
3. `rpl_malloc`/`rpl_realloc` on macOS (all versions): Spack compiler wrappers prevent autoconf from running test programs, causing `malloc`/`realloc`/`memcmp` checks to default to `no` and generate replacement stubs that are never defined. Pre-seed the autoconf cache with correct values on `platform=darwin`.
4. `--with-hdf4` passed lib dir instead of prefix (all versions): configure appends `/lib` and `/include` to the `--with-hdf4` value, so passing the lib directory produces `lib/lib` and `lib/include`. Pass `spec["hdf"].prefix` instead.

---

I've confirmed this builds on macOS. I'm keeping draft until I can test on Linux (which is less available for me). It's odd because some of these fixes seem like "Linux shouldn't have built" and yet...it did.

cc: @danrosen25 @climbfuji


mathomp4:
Well, it builds on my Linux testbed (with the fancy new installer by @haampie too!).

I've undrafted.

climbfuji (approved):
Wow, this is quite involved. Glad that you were able to figure this out!

mathomp4:
> Wow, this is quite involved. Glad that you were able to figure this out!

I freely admit some was me, some was various ML models. Pasting a lot of build log failures to figure this out! 

Meanwhile, Linux: whee! Simple! All happy! :D

repos/spack_repo/builtin/packages/hdf_eos2/package.py
```diff
@@ -119,6 +121,25 @@ def fix_configure(self):
         filter_file("\\$CC -show &> /dev/null", "true", "configure")
         filter_file("CC=./\\$SZIP_CC", "", "configure")
 
+    @run_before("autoreconf")
+    @when("@3.0")
+    def fix_gctp_linking(self):
```
alalazo:
```suggestion
    @run_before("autoreconf", when=@3.0)
    def fix_gctp_linking(self):
```

alalazo (commented):
Didn't check the rest, but you should use the `when=` argument of `run_before`

repos/spack_repo/builtin/packages/hdf_eos2/package.py
mathomp4: Ah. My guess is I saw someone else in the code doing that and I copied it.

repos/spack_repo/builtin/packages/hdf_eos2/package.py
```diff
@@ -158,6 +183,22 @@ def setup_build_environment(self, env: EnvironmentModifications) -> None:
                     "CFLAGS", "-Wno-error=implicit-function-declaration -Wno-error=implicit-int"
                 )
 
+    @run_before("configure")
```
climbfuji: I guess this doesn't need a `when` condition?

climbfuji (approved):
Happy with those changes, just one question left.

repos/spack_repo/builtin/packages/hdf_eos2/package.py
mathomp4: Huh. yeah, maybe it does for symmetry's sake if nothing else. Updated

danrosen25:
Hi @mathomp4 
I was testing this with gcc@15 and I had to add `-std=gnu17` to the CFLAGS in `flag_handler` and `setup_build_environment`. Otherwise `gctp.c` errors due to `incompatible pointer type` and `too many arguments to function`.

What compiler are you using? Ubuntu 25.10 installs gcc@15.2.0 as the default and it's the compiler for most -dev libraries. gcc@15 uses std=gnu23.

mathomp4:
> Hi @mathomp4 I was testing this with gcc@15 and I had to add `-std=gnu17` to the CFLAGS in `flag_handler` and `setup_build_environment`. Otherwise `gctp.c` errors due to `incompatible pointer type` and `too many arguments to function`.
> 
> What compiler are you using? Ubuntu 25.10 installs gcc@15.2.0 as the default and it's the compiler for most -dev libraries. gcc@15 uses std=gnu23.

Huh. I guess I probably am not building this (yes) in my gcc 15 testing. I'm starting with a smaller set of libraries and hdf-eos2 is not one I usually do.

Feel free to suggest a fix.

repos/spack_repo/builtin/packages/hdf_eos2/package.py
```diff
@@ -158,6 +183,22 @@ def setup_build_environment(self, env: EnvironmentModifications) -> None:
                     "CFLAGS", "-Wno-error=implicit-function-declaration -Wno-error=implicit-int"
```
danrosen25:
```suggestion
                    "CFLAGS", "-Wno-error=implicit-function-declaration -Wno-error=implicit-int -std=gnu17"
```

danrosen25 (changes requested):
Force std=gnu17 to avoid build errors in gctp.c

repos/spack_repo/builtin/packages/hdf_eos2/package.py
mathomp4, line 154: Hmm. this one got confused?

danrosen25: @mathomp4 Yes, it's supposed to insert code without modifying any lines but I can't find a way to suggest without modifying a line.

danrosen25: @mathomp4 And I don't know if you saw it but the style checker didn't like the line length for my other suggestion and it suggests the following change.
```
                 env.set(
-                    "CFLAGS", "-Wno-error=implicit-function-declaration -Wno-error=implicit-int -std=gnu17"
+                    "CFLAGS",
+                    "-Wno-error=implicit-function-declaration -Wno-error=implicit-int -std=gnu17",
                 )
```

mathomp4: @danrosen25 Okay. I think I just pushed up your other changes. Can you take a look?

bernhardkaindl:
@danrosen25 can you re-review this PR? I think there were some questions for you.

danrosen25 (approved):
I did some rudimentary testing in an ubuntu container and the hdf-eos2 package built and linked fine. Looks good to me and thanks for the comments.

repos/spack_repo/builtin/packages/hdf_eos2/package.py
bernhardkaindl: As danrosen25 approved now, I assume this is resolved.

bernhardkaindl (approved):
@srekolam, @mathomp4: As @danrosen25 approved now, I think the review is completed: I merged the PR.
I hope this was not too quick. If it was, I apologize in advance.

PR #4549, opened 2026-04-28
opentofu: new package

taliaferro:
This PR adds a package recipe for the `tofu` command-line tool for [OpenTofu](https://github.com/opentofu/opentofu).


taliaferro:
@spackbot fix style, if you'd be so kind

repos/spack_repo/builtin/packages/opentofu/package.py
```diff
@@ -0,0 +1,44 @@
...
+        return [
+            "-p",
+            str(make_jobs),
+            "-modcacherw",
+            "-ldflags",
+            f"-X main.version={self.version} -X github.com/opentofu/opentofu/version.dev=no ",
+            "-o",
+            "tofu",
+            "./cmd/tofu",
+        ]
```
alecbcs, line 44:
```suggestion
    build_directory = "cmd/tofu"
```

I think we *should* be able to use the builtin GoPackage methods here if we set the build_directory instead.

alecbcs (changes requested):
One quick suggestion, otherwise looks good to me. Thanks @taliaferro!

repos/spack_repo/builtin/packages/opentofu/package.py
taliaferro: That does look much cleaner, but it also means that the final executable [will think it's a dev build](https://github.com/opentofu/opentofu/blob/main/contributing/FAQ.md#when-i-run-tofu-version-it-contains-a--dev-suffix-how-do-i-get-rid-of-it) because of the lack of that `version.dev=no` linker flag.

taliaferro: Is there a good way to specify just the `ldflags` for Go packages?

alecbcs: No, maybe this is the kick in the butt I need to actually do it though.

alecbcs: @taliaferro I went and modernized the GoPackage builder class a little while back. If you want to rebase this to define only the custom `ldflags` this should make the PR much smaller.

https://github.com/spack/spack-packages/pull/4795

PR #4563, opened 2026-04-29
cuda: remove bogus libxml2 dep for 12+

sethrj:
At some point cuda apparently needed libxml2, but now it does not. Verified installation of this updated recipe with `cuda@13.1.1` on linux-ubuntu24.04-x86_64_v3 . I also used `strings`/`grep` with the binary installer for 12.9 to check for any references to XML and `libtree`/`ldd` on the installed libraries and binaries. I can't find any use of it nor reference to the requirement in the CUDA installation guide.

### DIFF
--- a/repos/spack_repo/builtin/packages/cuda/package.py
+++ b/repos/spack_repo/builtin/packages/cuda/package.py
@@ -785,12 +785,12 @@ class Cuda(Package):
         description="Allow unsupported host compiler and CUDA version combinations",
     )
 
-    # depends on libxml2.so.2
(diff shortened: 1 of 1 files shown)


alecbcs (approved):
Looks good to me. Thanks for digging into this @sethrj 

Will give it a couple days for @ax3l @pauleonix @Rombur to have a chance to review

alecbcs:
Merging without hearing back from @ax3l @pauleonix @Rombur after > 1 week.

PR #4568, opened 2026-04-29
detray: add v1.0.0, v1.0.1

stephenswat:
This commit adds versions 1.0.0 and 1.0.1 of the detray package.

### DIFF
--- a/repos/spack_repo/builtin/packages/detray/package.py
+++ b/repos/spack_repo/builtin/packages/detray/package.py
@@ -21,6 +21,8 @@ class Detray(CMakePackage):
 
     license("MPL-2.0", checked_by="stephenswat")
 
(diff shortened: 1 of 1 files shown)


stephenswat:
@spackbot re-run pipeline

wdconinc (dismissed):
This looks good to me, but does it now need a `python` runtime dependency to deal with the installed python package?

stephenswat:
@niermann999 can you weigh in on this? Is Python now a dependency?

niermann999:
I would say yes, since the metadata generation uses python to generate the corresponding header files. Unless you do not generate your own detector type, then it is not needed.

stephenswat:
Added a build dependency on Python. Thank you both!

wdconinc (approved):

PR #4570, opened 2026-04-29
plumed: Fix Spack compilers wrappers injection

RMeli:
The Spack compiler wrappers are injected into `plumed` at build time, resulting in the following error with `plumed mklib`:

```
[spack cc]: Error: Spack compiler must be run from Spack! Input 'SPACK_COMPILER_WRAPPER_PATH' is missing.
```

I'm not sure if this is the correct way forward, but this PR uses `filter_compiler_wrappers` to filter out the Spack compiler wrappers so that `plumed mklib` works correctly.

### DIFF
--- a/repos/spack_repo/builtin/packages/plumed/package.py
+++ b/repos/spack_repo/builtin/packages/plumed/package.py
@@ -181,6 +181,14 @@ class Plumed(AutotoolsPackage):
 
     parallel = True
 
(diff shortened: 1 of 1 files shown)


RMeli:
@spackbot fix style

albestro (approved):
LGTM

albestro:
@spackbot re-run pipeline

PR #4573, opened 2026-04-29
arborx: add v2.1

aprokop:

### DIFF
--- a/repos/spack_repo/builtin/packages/arborx/package.py
+++ b/repos/spack_repo/builtin/packages/arborx/package.py
@@ -14,7 +14,7 @@ class Arborx(CMakePackage, CudaPackage, ROCmPackage):
     """ArborX is a performance-portable library for geometric search"""
 
     homepage = "https://github.com/arborx/arborx"
-    url = "https://github.com/arborx/arborx/archive/v1.1.tar.gz"
+    url = "https://github.com/arborx/arborx/archive/v2.1.tar.gz"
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/arborx/package.py
```diff
@@ -14,7 +14,7 @@ class Arborx(CMakePackage, CudaPackage, ROCmPackage):
     """ArborX is a performance-portable library for geometric search"""
 
     homepage = "https://github.com/arborx/arborx"
-    url = "https://github.com/arborx/arborx/archive/v1.1.tar.gz"
+    url = "https://github.com/arborx/arborx/archive/v2.1.tar.gz"
```
alecbcs, line 17: Technically we don't need to update the URL here since Spack will automatically find and replace the version in the url string for each version, but it also doesn't hurt.

alecbcs (approved):
Looks good to me. Thanks @aprokop

repos/spack_repo/builtin/packages/arborx/package.py
aprokop: Yeah, I knew that. I saw some people updating those, thought that I haven't updated that in a while, so just did the change.

aprokop:
Some CI tests failed with `Could not resolve host: gitlab.spack.io`

alecbcs:
@spackbot run pipeline

PR #4575, opened 2026-04-29
kokkos: add v5.1.1

JBludau:

### DIFF
--- a/repos/spack_repo/builtin/packages/kokkos/package.py
+++ b/repos/spack_repo/builtin/packages/kokkos/package.py
@@ -28,6 +28,7 @@ class Kokkos(CMakePackage, CudaPackage, ROCmPackage):
 
     version("develop", branch="develop")
 
+    version("5.1.1", sha256="8bdbee0f0ac383436743ad8a9e3e928705b34b31a25a92dc5179c52a3aa98519")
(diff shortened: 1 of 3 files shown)


lucbv (approved):
Thanks @JBludau

tpadioleau (approved):
Looks good to me!

alecbcs (dismissed):
Looks good to me. Thanks @JBludau

nmm0 (approved):

JBludau:
@alecbcs Thanks for taking a look. The CI failures look like related to gihub problems to me, maybe rerunning is enough?

alecbcs:
@spackbot run pipeline

rbberger:
Looks like e4s-oneapi fails because it asks for `cxxstd=17` and therefore falls back to an older version of Kokkos.

JBludau:
> Looks like e4s-oneapi fails because it asks for `cxxstd=17` and therefore falls back to an older version of Kokkos.

Does the fact, that all the other ci jobs passed indicate that the spack CI only catches these problems spuriously? I added a version restriction to kokkos_kernels (requiring kokkos 5.1.1 for kernels 5.1.1) ... but this should have affected all other pipelines as well, right?

tpadioleau (approved):
Good catch, I should have read our checklist on the first review https://github.com/kokkos/development/blob/main/meeting_notes/wg-build-packaging/checklist/spack_release.md :)

kshea21 (approved):

PR #4588, opened 2026-04-29
ParaView: Make OSMesa a link dep when +osmesa_fallback

kwryankrattiger:
This is not really true, but it solves the explicit fallback not being found in LD_LIBRARY_PATH problem by putting it in the rpath.

### DIFF
--- a/repos/spack_repo/builtin/packages/paraview/package.py
+++ b/repos/spack_repo/builtin/packages/paraview/package.py
@@ -322,7 +322,7 @@ class Paraview(CMakePackage, CudaPackage, ROCmPackage):
         # * the system rendering default (WGL/AGL/GLX)
         # * EGL
         # * OSMesa (guarenteed to exist and work on all systems)
-        depends_on("osmesa", type=("run"), when="+osmesa_fallback")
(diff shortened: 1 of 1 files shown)


kmorel:
I have not tried this yet. Is this to fix the issue I am having on Frontier? This does not look like it would modify `LD_LIBRARY_PATH`. Will this change the build to bypass the `dlopen` and use osmesa directly?

vicentebolea (approved):

kwryankrattiger:
> Will this change the build to bypass the dlopen and use osmesa directly?

This should just put it into the rpath for searching, but it will still only be linked with dlopen. We load GL via GLAD only now.

kmorel (approved):
Well, I'll be a monkey's uncle. That worked. I can render on Frontier now.

I didn't know paths were captured for libraries not linked. Thanks.

johnwparent:
@spackbot run pipeline

johnwparent:
Not sure what happened to gitlab but I couldn't manually rerun the failing pipeline, retriggering the whole thing hopefully helps

kwryankrattiger:
rebased on a somewhat newer develop, let's see if that helps.

johnwparent (approved):
LGTM

PR #4591, opened 2026-04-30
gcc: add v16.1.0

alalazo:
Let's see if anything breaks in pipelines :slightly_smiling_face:


haampie:
We don't build it apparently. Was the goal to build it too?

alalazo:
Goal was to check if anything required to be changed in pipelines. It seems there is no pipeline where we build just `gcc`. I tried a default build on my platform and it works, but obviously that's not an indication that all variants etc. work with `gcc@16`. 

Only question is if we want the preferred version to be the latest or not for the time being.

tgamblin (approved):

PR #4594, opened 2026-04-30
py-pydantic-settings: add v2.14

cmelone:
https://github.com/pydantic/pydantic-settings/blob/8916beeecc6d0510e3d0532a0ed839937400ddc3/pyproject.toml#L41-L46


repos/spack_repo/builtin/packages/py_pydantic_settings/package.py
kshea21:
```suggestion
   depends_on("py-hatchling", type="build")
   
   with default_args(type=("build", "run")):
        depends_on("python@3.10:", when="@2.14:")
        depends_on("py-typing-inspection@0.4:")
        depends_on("python@3.8:")
        depends_on("py-pydantic@2.7.0:")
        depends_on("py-python-dotenv@0.21:")
```
Not necessary, but you could add `default_args` to cleanup/organize the deps a little more.

kshea21 (approved):
Great, thanks!

PR #4606, opened 2026-04-30
prrte: fixes for develop

Chrismarsh:
- Fixes for being able to build git@develop
- This includes the new flags to support the PBS changes from https://github.com/openpmix/prrte/pull/2434

### DIFF
--- a/repos/spack_repo/builtin/packages/prrte/package.py
+++ b/repos/spack_repo/builtin/packages/prrte/package.py
@@ -45,7 +45,12 @@ class Prrte(AutotoolsPackage):
     version("2.0.0", sha256="9f4abc0b1410e0fa74ed7b00cfea496aa06172e12433c6f2864d11b534becc25")
     version("1.0.0", sha256="a9b3715e059c10ed091bd6e3a0d8896f7752e43ee731abcc95fb962e67132a2d")
 
-    depends_on("c", type="build")  # generated
+    depends_on("c", type="build")
+
+    # Python is used to generate the docs and the show-help array,
+    # both of which are included pre-built in tarballs.
+    # https://github.com/openpmix/prrte/issues/2438
+    depends_on("python@3.7:", type="build", when="@develop")
 
     depends_on("pmix")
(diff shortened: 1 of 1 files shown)


Chrismarsh:
@wdconinc the addition of python for the develop branch doesn't seem to be a requirement anymore. [As you added this dependency](https://github.com/spack/spack-packages/blame/develop/repos/spack_repo/builtin/packages/prrte/package.py#L66), I'm wondering if there was some other reason I have missed?

wdconinc:
> @wdconinc the addition of python for the develop branch doesn't seem to be a requirement anymore. [As you added this dependency](https://github.com/spack/spack-packages/blame/develop/repos/spack_repo/builtin/packages/prrte/package.py#L66), I'm wondering if there was some other reason I have missed?

https://github.com/openpmix/prrte/blob/master/configure.ac#L709 was the reason.

wdconinc:
> > @wdconinc the addition of python for the develop branch doesn't seem to be a requirement anymore. [As you added this dependency](https://github.com/spack/spack-packages/blame/develop/repos/spack_repo/builtin/packages/prrte/package.py#L66), I'm wondering if there was some other reason I have missed?
> 
> https://github.com/openpmix/prrte/blob/master/configure.ac#L709 was the reason.

Maybe this doesn't apply anymore due to changes to the git fetching that doesn't put the `.git/` directory there anymore? That could have changed since I tried this.

Chrismarsh:
image
Indeed there is no `.git/`, so presumably why I didn't hit issues. I'll ask for further clarification.

Chrismarsh:
Ah vscode just doesn't show the `.git/` folder? It is there in the checkout. Anyway, the answer is:
```
Python is used to generate the docs and the show-help array, both of which are included pre-built in tarballs. You'll still be able to execute, but won't get any show-help output for errors as configure will assume it is not in a git clone.
```
I've added it back. Thanks @wdconinc

wdconinc (approved):
LGTM!

PR #4607, opened 2026-04-30
Move contributing guidance higher for spack-packages

tgamblin:
Probably the most important detail in `README.md` is how to set your dev environment up for contribution. Move that to the top of `README.md` so that people can more easily find it.

The rest of `README.md`, I think, is stuff people already expect to find there, so they'll look for it anyway.


alecbcs (approved):
Agreed. Yep let's move this to the top of our README.

PR #4615, opened 2026-05-01
Only run Gitlab CI Stacks if there are changes that may require it

kwryankrattiger:
Try to save some resources for the odd job that does not change anything that affects CI.

The current unaffected pruning rules will attempt to rebuild all missing shas (ie. if there are develop failures, they will be forced to run) on changes only to `README.md`. This results in unnecessary friction and CI cost  on PRs that should otherwise be very quick to turn around.

CC: @tgamblin

### DIFF
--- a/.ci/gitlab/.gitlab-ci.yml
+++ b/.ci/gitlab/.gitlab-ci.yml
@@ -83,6 +83,11 @@ default:
       variables:
         <<: *common_pr_vars
         SPACK_PRUNE_UNTOUCHED: "True"
+      changes:
+        - .ci/env
+        - .ci/gitlab/*
+        - repos/spack_repo/builtin/packages/*
+        - stacks/$SPACK_CI_STACK_NAME/*
 
     - if: $CI_COMMIT_REF_NAME =~ /^pr[\d]+_.*$/ && $CI_PROJECT_NAME == "spack"
       # Pipelines on PR branches rebuild only what's missing, and do extra pruning
@@ -247,6 +252,19 @@ dotenv:
   - echo "$WRITE_DOTENV_SCRIPT_SHA256  write_dotenv.py" | sha256sum -c
   - python3 write_dotenv.py .ci/env
   - cat env
+
+force_pipeline_status:
+  stage: generate
+  tags: ["spack", "service_noop"]
+  image: busybox@sha256:37f7b378a29ceb4c551b1b5582e27747b855bbfaa73fa11914fe0df028dc581f
+  variables:
+    CI_OIDC_REQUIRED: 0
+    GIT_STRATEGY: "none"
+    CI_JOB_SIZE: "small"
+    KUBERNETES_CPU_REQUEST: "100m"
+    KUBERNETES_MEMORY_REQUEST: "5M"
+  script:
+  - echo "This job is a placeholder job to ensure a pipeline status is posted even if no jobs are run"
 
 ########################################
 # TEMPLATE FOR ADDING ANOTHER PIPELINE
--- a/.ci/gitlab/.gitlab-ci.yml
+++ b/.ci/gitlab/.gitlab-ci.yml
@@ -89,7 +89,7 @@ default:
           - .ci/gitlab/*
           - repos/spack_repo/*
           - stacks/$SPACK_CI_STACK_NAME/*
-        compare_to: develop
+        compare_to: HEAD~1
 
     - if: $CI_COMMIT_REF_NAME =~ /^pr[\d]+_.*$/ && $CI_PROJECT_NAME == "spack"
       # Pipelines on PR branches rebuild only what's missing, and do extra pruning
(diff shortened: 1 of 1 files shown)


johnwparent:
Failing gitlab pipeline unrelated. Re-ran

johnwparent (dismissed):
Looks good. Approved.

Probably a dumb thought, but would we want to guard this commit to commit? that way if I'm modifying a package and then need to make some non package change, we don't need to spin gitlab?

Something like `compare_to: HEAD~1`

.ci/gitlab/.gitlab-ci.yml
```diff
@@ -83,6 +83,13 @@ default:
       variables:
         <<: *common_pr_vars
         SPACK_PRUNE_UNTOUCHED: "True"
+      changes:
+        paths:
+          - .ci/env
+          - .ci/gitlab/*
+          - repos/spack_repo/builtin/packages/*
```
becker33: I think this needs to include `repos/spack_repo/builtin/build_systems/*` as well

becker33 (dismissed):

.ci/gitlab/.gitlab-ci.yml
tgamblin: `spack_repo` is the root python package -- why not include all of it?  Anything there could be imported by a package.

kwryankrattiger: since, afaik, that doesn't affect hashes I left it out. I can include it, but my guess is unless one of those other places has changed there will be no hash changes and it will just noop everything.

alecbcs:
It would be nice if we could limit the `force_pipeline_status` job to only be rendered out when no other generate jobs are run so that we're not creating yet another noop job that always runs in CI.

We could potentially do something like the following to not have to redefine the list of files twice.

```yaml
.generate-paths: &generate-paths
  - .ci/env
  - .ci/gitlab/*
  - repos/spack_repo/*
  - stacks/$SPACK_CI_STACK_NAME/*

.base-job:
   # ... existing conditionals, vars, tags, image ...
  changes:
    paths: *generate-paths

force_pipeline_status:
  stage: generate
  # ... existing vars, tags, image ...
  rules:
    - changes:
        paths: *generate-paths
      when: never
    - when: always
```

kwryankrattiger:
CI failure from develop
https://gitlab.spack.io/spack/spack-packages/-/pipelines/1581279

kwryankrattiger:
> We could potentially do something like the following to not have to redefine the list of files twice.

I am not worried about the list of files so much as the entire rules block. If we only want to run the dummy job when nothing else is running we would need to invert the entire rules block not just copy this list once. This is also more likely to get out of sync and increases the maintenance surface for CI. We will more likely break things doing this than save time.

alecbcs:
> I am not worried about the list of files so much as the entire rules block. If we only want to run the dummy job when nothing else is running we would need to invert the entire rules block not just copy this list once.

The `when: never` is doing the negation for us so that we can reuse the positive list of files.

> This is also more likely to get out of sync and increases the maintenance surface for CI. We will more likely break things doing this than save time.

If we make use of a yaml anchor we're only defining the list once. I don't see how this would get out of sync and increase the maintenance surface for CI. Same list, referenced two places, saves us a bunch of noop jobs when they aren't needed.

kwryankrattiger:
> If we make use of a yaml anchor we're only defining the list once. I don't see how this would get out of sync and increase the maintenance surface for CI. Same list, referenced two places, saves us a bunch of noop jobs when they aren't needed.

We run this for both core and packages, so that is not the full list of cases. You need even more from the main rules block, not just the list of packages. It would be something like this

```yaml
.packages_on_changes: &packages_on_changes
  paths:
    - .ci/env
    - .ci/gitlab/*
    - repos/spack_repo/*
    - stacks/$SPACK_CI_STACK_NAME/* # Evaluates to stacks//* when callled from force_pipeline_status
   compare_to: HEAD~1  # Assuming this is a merge commit created by sync script, will need to change for hubcast

.base-job:
    rules:
       - if: $CI_COMMIT_REF_NAME =~ /^pr[\d]+_.*$/ && $CI_PROJECT_NAME == "spack-packages"
         changes: *package_on_changes
         when: on_success
       - ....

force_pipeline_status:
  stage: generate
  # ... existing vars, tags, image ...
  rules:
    - if: $CI_COMMIT_REF_NAME =~ /^pr[\d]+_.*$/ && $CI_PROJECT_NAME == "spack-packages" # Don't run for Package PRs with changes
      changes: *packages_on_changes
      when: never
    - if: $CI_PROJECT_NAME == "spack": # Never run for core pipelines
      when: never
    - if: $CI_COMMIT_REF_NAME =~ /^pr[\d]+_.*$/: # Only run for PRs that are not core or excluded by packages
      when: always
    - when: never # Don't run for any other case
```

This is just too much imo for little gain. It is more to wrangle if there ever need to be changes.

alecbcs (approved):
Looks good to me. Thanks @kwryankrattiger!

PR #4622, opened 2026-05-01
r-fs: add v2.1.0

manuelakuhn:
https://cran.r-project.org/package=fs

Exchanged `gmake` build dependency with `cmake` because I got the following error otherwise:
```
  /bin/sh: line 1: cmake: command not found
```


RikkiButler20 (approved):
LGTM

PR #4635, opened 2026-05-02
r-hms: add v1.1.4

manuelakuhn:
https://cran.r-project.org/package=hms

### DIFF
--- a/repos/spack_repo/builtin/packages/r_hms/package.py
+++ b/repos/spack_repo/builtin/packages/r_hms/package.py
@@ -17,6 +17,7 @@ class RHms(RPackage):
 
     license("MIT")
 
+    version("1.1.4", sha256="efc525f797b62b3740d06c6fa8202593ab5aa4fc1edeefb76b6eb9be89e87b94")
     version("1.1.3", sha256="e626f4c60af46efd53ea631b316a103e089470d8fd63c0e0c0efb99364990282")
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/r_hms/package.py
```diff
@@ -17,20 +17,25 @@ class RHms(RPackage):
...
+        depends_on("r-lifecycle", when="@1.0.0:")
+        depends_on("r-pkgconfig", when="@0.5.0:")
+        depends_on("r-rlang@1.0.2:", when="@1.1.3:")
+        depends_on("r-rlang", when="@0.5.0:")
+        depends_on("r-vctrs@0.3.8:", when="@1.1:")
+        depends_on("r-vctrs@0.2.1:", when="@1.0.0:")
+        depends_on("r-vctrs@0.2.0:", when="@0.5.0:")
+
+        # Historical dependencies
+        depends_on("r-ellipsis@0.3.2:", when="@1.1.2:1.1.2")
```
RikkiButler20:
```suggestion
        depends_on("r-ellipsis@0.3.2:", when="@1.1.2")
```
I think that this would be easier to read

RikkiButler20 (changes requested):

repos/spack_repo/builtin/packages/r_hms/package.py
manuelakuhn: Thanks. That makes sense. I did not see it, because I just took what was already there :-)

RikkiButler20 (approved):
Looks good. Thanks

PR #4650, opened 2026-05-04
ecp-data-vis-sdk: drop old ascent conflict

albestro:
I've just ran across the comment and I'm trying to do what it suggests.

Let's see if it works.


kwryankrattiger:
This packages specifically does not want to build the `develop` version of ascent

haampie:
@kwryankrattiger this PR looks sensible to me? The comment is a "delete me after ..." which is now, right?

kwryankrattiger:
> @kwryankrattiger this PR looks sensible to me? The comment is a "delete me after ..." which is now, right?

That comment was not well written. Specifically the issue was the version was forced to always be `develop` because of that case. It was always the intention to not build development branches with this package and as time has gone on that has been an even stricter requirement.

I closed this because it is not what we wanted to do for the SDK regardless of the comment that was there. Feel free to delete just the comment, but eventually this package should probably just be removed as the specification for the SDK has moved to configs/environments to solve issues like this but still allow overrides at deployment.

albestro:
thanks for the details @kwryankrattiger!

PR #4656, opened 2026-05-04
openblas: Add patch for gfortran library

paulgessinger:
Add https://github.com/OpenMathLib/OpenBLAS/pull/5796 in openblas `0.3.32` (current version). If merged, future versions would not need the patch.


mathomp4 (approved):

mathomp4:
@paulgessinger Interesting. I've been building mixed Apple Clang and GNU/NAG stacks and OpenBLAS seemed happy. But maybe that's because I'm building with CMake? Hmm.

paulgessinger:
That's possible. I didn't consider switching to the cmake build.

paulgessinger:
I can't switch to the cmake build, because another package hard requires `+ilp64`, which `openblas` records as a conflict because it's not available in the cmake build.

mathomp4 (approved):

haampie (approved):

PR #4663, opened 2026-05-04
`py-ranger-fm`: dependency update and new version

taliaferro:
`py-ranger-fm` is a terminal file manager written in Python. One feature it provides is dynamic previews of files -- it attempts to determine the type of a file, then shows it in the file preview window if the system has the right tools installed to display the file in a terminal-friendly way.

In the versions that have so far been released, part of this process relied on the `imghdr` module from the standard library, which was removed in Python 3.13 as part of [PEP 0594](https://peps.python.org/pep-0594/) which cleaned out some of the stdlib's "dead batteries." However, this means that right now, `py-ranger-fm` crashes due to an ImportError when run with Pythons 3.13 or newer. This PR adds a dependency constraint to reflect this.

(For good measure, I also checksummed the latest patch release of `py-ranger-fm` and added it to the package as well; it doesn't fix the dependency on `imghdr`, though.)

### DIFF
--- a/repos/spack_repo/builtin/packages/py_ranger_fm/package.py
+++ b/repos/spack_repo/builtin/packages/py_ranger_fm/package.py
@@ -15,5 +15,11 @@ class PyRangerFm(PythonPackage):
 
     version("1.9.2", sha256="0ec62031185ad1f40b9faebd5a2d517c8597019c2eee919e3f1c60ce466d8625")
 
+    # Ranger uses the `imghdr` module from the standard library to determine the type of an image
+    # file, as the first step in determining whether and how to display a preview of the image.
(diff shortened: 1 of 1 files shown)


manuelakuhn (commented):
I can reproduce the problem and confirm that this fixes the issue on rocky linux 8 with gcc@15.

It seems that this problem was fixed in `1.9.4` (see https://ranger.fm/). Is there a reason why you only added `1.9.3` and not the newest version?

bernhardkaindl (approved):
@taliaferro, approving, but please see https://github.com/spack/spack-packages/pull/4663#pullrequestreview-4246700815

PR #4670, opened 2026-05-04
Fix VEP dependency chain build failures with gcc@13+

teaguesterling:
## Summary
- **perl-dbd-mysql**: Add missing `mysql-client` dependency (links against libmysqlclient but wasn't declaring it)
- **kentutils**: Add patch for missing `#include <cstdint>` in `straw.cpp` that fails with GCC 13+; narrow existing mysql pointer patch to `gcc@13.0:13.2` where it actually applies
- **perl-bio-bigfile**: Fix htslib linker path — Perl's build system calls gcc directly (bypassing spack's compiler wrapper), so `-lhts` needs an explicit library path

## Context
These three packages form part of the VEP (Variant Effect Predictor) dependency chain. With GCC 13+, the build fails at multiple points due to stricter header requirements and missing library paths in Perl's native build system.

## Test plan
- [x] Full `spack install` of VEP environment (167 packages) succeeds with gcc@13.3
- [x] kentutils@478 builds with the cstdint patch applied
- [x] perl-bio-bigfile links successfully against external htslib
- [x] perl-dbd-mysql finds mysql headers/libs

🤖 Generated with [Claude Code](https://claude.com/claude-code)


hainest:
@teaguesterling Have you been able to successfully build these changes?

teaguesterling:
@hainest: Yes. These changes were added as a result of trying to reinstall and use the package after an OS upgrade. That said, I haven't tested it on any other systems aside from the CI. 

FWIW: I know these packages well and only had Claude collect the changes into the PR, not actually determine what needed fixing.

hainest (approved):
LGTM.

PR #4671, opened 2026-05-04
py-stevedore: fix dependency type

sethrj:
`pbr` is listed in `requirements.txt` but is only used by `setup.py`.


alalazo (approved):

sethrj:
Thanks @alalazo !

PR #4672, opened 2026-05-05
omega-h: Use seacas for exodus (instead of trilinos)

Fuad-HH:
Trilinos was needed for exodus mesh files which only needs seacas. Now secas is directly used by Omega_h instead of through trilinos.


repos/spack_repo/builtin/packages/omega_h/package.py
```diff
@@ -67,7 +67,7 @@ class OmegaH(CMakePackage, CudaPackage):
     variant("shared", default=True, description="Build shared libraries")
     variant("mpi", default=True, description="Activates MPI support")
     variant("zlib", default=True, description="Activates ZLib support")
-    variant("trilinos", default=True, description="Use Teuchos and Kokkos")
+    variant("trilinos", default=False, description="Use Teuchos and Kokkos")
```
Fuad-HH: I have removed trilinos from default configuration since seacas is preferred. Should I update the description string?

repos/spack_repo/builtin/packages/omega_h/package.py
```diff
@@ -90,9 +90,10 @@ class OmegaH(CMakePackage, CudaPackage):
     depends_on("gmsh@4.4.1:", when="+gmsh")
     depends_on("mpi", when="+mpi")
     depends_on("trilinos +kokkos", when="+trilinos")
-    depends_on("trilinos +exodus", when="+exodus")
     depends_on("kokkos", when="+kokkos")
+    depends_on("kokkos@4.3.00:", when="@:11.0.0-scorec+kokkos")
```
Fuad-HH: Kokkos minimum version needed is 4.3.00 for kokkos sort (ref [here](https://github.com/SCOREC/omega_h/blob/991302239e7c5aa2f2052eb2f76e7959f332441e/CMakeLists.txt#L92-L94)). Should I do it for previous older versions of omega-h too?

repos/spack_repo/builtin/packages/omega_h/package.py
```diff
@@ -90,9 +90,10 @@ class OmegaH(CMakePackage, CudaPackage):
...
     depends_on("zlib-api", when="+zlib")
+    depends_on("seacas@:2024-8-15~x11~tests~fortran", when="+exodus")
```
Fuad-HH: I am using seacas version less than 2024-8-15 because it includes `netcdf-c` which is needed for some function calls in omega-h exodus related header. Although it is resolved in a [commit later](https://github.com/SCOREC/omega_h/commit/4dd682ef16ebf2502239ad06883e9f10c611f1c4), it is not included in omega-h version 11.0.0.

repos/spack_repo/builtin/packages/omega_h/package.py
cwsmith: Please do. I'm not aware of a Teuchos dependency and grepping for `teuchos` didn't return any results.

repos/spack_repo/builtin/packages/omega_h/package.py
cwsmith: This should be fine.  If we can figure out when 4.3 became the requirement for tagged versions we should try to do so... but it isn't critical.

repos/spack_repo/builtin/packages/omega_h/package.py
cwsmith: Yeah... I'll create a new minor revision to avoid this added dependency.

repos/spack_repo/builtin/packages/omega_h/package.py
Fuad-HH: It was added in this pull request: https://github.com/SCOREC/omega_h/pull/146/changes#diff-1e7de1ae2d059d21e1dd75d5812d5a34b0222cef273b7c3a2af62eb747f9d20aL95

repos/spack_repo/builtin/packages/omega_h/package.py
Fuad-HH: I also didn't find anything related to Teuchos in the SCOREC fork. But it was in sandialabs. Sandialabs removed it in this commit: https://github.com/sandialabs/omega_h/commit/be5679ade8f108915aaee12ef9bba980d7b2246c

On another note: we can deprecate the trilinos variant when spack adds the deprecate feature for variants (https://github.com/spack/spack/pull/52112).

repos/spack_repo/builtin/packages/omega_h/package.py
cwsmith: @Fuad-HH the new v11.1.0 tag is here: https://github.com/SCOREC/omega_h/releases/tag/scorec-v11.1.0

cwsmith:
notes from offline discussion:
- add conflict for +trilinos +kokkos and +trilinos +seacas
- restrict +seacas dep for latest omegah tag scorec-v11.1.0

repos/spack_repo/builtin/packages/omega_h/package.py
```diff
@@ -89,10 +94,15 @@ class OmegaH(CMakePackage, CudaPackage):
     depends_on("gmsh", when="+examples")
     depends_on("gmsh@4.4.1:", when="+gmsh")
     depends_on("mpi", when="+mpi")
-    depends_on("trilinos +kokkos", when="+trilinos")
-    depends_on("trilinos +exodus", when="+exodus")
+    depends_on("trilinos +kokkos+exodus", when="+trilinos")
+    depends_on("trilinos +kokkos+exodus", when="@:11.0.0-scorec+exodus")
```
Fuad-HH, line 98: Making a new seacas variant doesn't sound right to me to resolve this conflict for v11.1.0. Should I just keep it as it is or add a conflict that if we are using seacas, we don't need trilinos. It would be easy if we could use `^` in when in conflicts, but it is not allowed.

repos/spack_repo/builtin/packages/omega_h/package.py
Fuad-HH, line 125: It says that cuda can work without kokkos after version 10.8.5 (for example 11.1.0). Is this right? Or does it have to be any version after 10.5?

cwsmith: I'm reading this as a conflict with +cuda~kokkos for the  omegah version range from 10.5 - 10.8.5 (`@10.5:10.8.5`).

repos/spack_repo/builtin/packages/omega_h/package.py
cwsmith: This works for me. I mistakenly wrote `+seacas` in my comment; sorry for the confusion.

cwsmith (approved):

repos/spack_repo/builtin/packages/omega_h/package.py
Fuad-HH: If it is correct, then good. I just wanted to mention it because @jacobmerson was saying that the current omega-h might also have this behavior: cuda only works with kokkos.

cwsmith: > cuda only works with kokkos.

Yeah, we dropped support for the cuda-only (no kokkos) backend.

Fuad-HH: How does this conflict then enforce that kokkos is enabled when +cuda for later versions, let's say 11.1.0-scorec?

cwsmith: It doesn't.  Those versions of omegah (10.5 to 10.8.5) were broken when the cuda-only backend was enabled.

Fuad-HH: I see, got it. Thanks!

cwsmith (approved):

cwsmith (approved):

bernhardkaindl:
Hi @cwsmith, @Fuad-HH and @Sichao25:

- I see that the `git diff` context of #5792 has some lines this PR modifies. So if 5792 would be merged, this PR would need to be reworked.
- As @Sichao25's PR 5792 is new, when we merge this PR first, he can then easiliy rebase his PR on top of the merge in develop.

Is that fine with you?

If there are no objections, I'd approve and merge this PR tomorrow and add a note for @Sichao25 to rebase to #5792 on top of these changes (they just change context, so it should be relatively straight-forward form him to do it) after the merge.

The other option exists of course too, but someone, (@Sichao25) would have to do the rebase with conflict resoltion and add it to their PR or open a new PR.

This PR can't be easily rebased, at least GitHub says:
> Rebase conflict between head and base.

When I try to use the Rebase option of the Update button of this PR. Update with merge was successful.

Later or tomorrow, we should see how this PR did in CI.

cwsmith:
@bernhardkaindl Thanks for checking on this.  Yes, lets merge this first then rebase https://github.com/spack/spack-packages/pull/5792.

Sichao25:
I'm fine with merging this first. Thanks for letting me know.

bernhardkaindl (approved):
@cwsmith, approving with your approval to make progress for the next PR!

@Fuad-HH:

When enabling +exodus, there where builds which failed with:
> /usr/bin/ld: libomega_h.so: undefined reference to `nc_set_log_level'

However, other builds were successful.

I wasn't able to pinpoint the case yet, it must be a missing `-lnetcdf-c` or so when linking libomega_h.so.

Of course, it didn't show with `~shared`, and the trigger seems to be:
`spack install omega-h +exodus +mpi +shared`

I didn't work this out completely, but with +mpi, I see `parallel-netcdf-c` is used and so far, I saw nc_set_log_level seems to be inly in the non-parallel netcdf:
```py
grep -r nc_set_log_level ../spack/opt/spack/*/*netcdf*
../spack/opt/spack/linux-skylake/netcdf-c-4.10.0-oacmggfhgct2hgkvsf2el7os63aj7mok/include/netcdf.h:nc_set_log_level(int new_level);
grep: ../spack/opt/spack/linux-skylake/netcdf-c-4.10.0-oacmggfhgct2hgkvsf2el7os63aj7mok/bin/ncgen: binary file matches
```

@Fuad-HH, @cwsmith, as approved to base #5792 on top of this PR, so I'd say this the undefined undefined reference to `nc_set_log_level' can be solved that PR or another follow-up PR.

Cc: @Sichao25

cwsmith:
Thank you @bernhardkaindl .  We appreciate the help.

It seems this built against an older version of omegah as https://github.com/SCOREC/omega_h/pull/214/changes/4dd682ef16ebf2502239ad06883e9f10c611f1c4 removed the call to `nc_set_log_level`.

repos/spack_repo/builtin/packages/omega_h/package.py
```diff
@@ -113,8 +127,6 @@ class OmegaH(CMakePackage, CudaPackage):
     # https://gcc.gnu.org/bugzilla/show_bug.cgi?id=86610
     conflicts("%gcc@8:8.2", when="@:9.22.1")
 
-    conflicts("+exodus", when="~trilinos", msg="exodus requires trilinos to be enabled")
-
     def patch(self):
         if "@:9.34.8" in self.spec:
```
bernhardkaindl, line 131: @cwsmith, this `filter_file()` removes (tested to build both affected versions) remaining debug call for the +exodus versions that still have the debug call:
```suggestion
    def patch(self):
        # https://github.com/SCOREC/omega_h/commit/4dd682ef16ebf2502239ad06883e9f10c611f1c4
        if self.spec.satisfies("@10.8.6-scorec:11.0.0-scorec"):
            filter_file(r"nc_set_log_level\(5\);", "", "src/Omega_h_exodus.cpp")
        if self.spec.satisfies("@:9.34.8"):
```
I guess a would be worth a suggestion to #5792 or alternatively a new PR.

Suggestion submitted: https://github.com/spack/spack-packages/pull/5792#discussion_r3735581919

BTW, off-topic: Is deprecating some of the older versions of omega-h a consideration for spack?

PR #4674, opened 2026-05-05
Earlier versions require setuptools-scm too

vucoda:
setuptools-scm is required to build earlier versions of python-dateutil, looks like from 2.7 onwards


alecbcs (approved):
Looks good to me. Thanks @vucoda!

Chrismarsh:
Not really the fault of this change per se, but this change breaks concretizing `spack spec py-xarray@2026: +viz` because xarray needs
```
depends_on("py-setuptools-scm@8:", when="@2025.8:")
```
and this change enforces
```
depends_on("py-setuptools-scm@:7", when="@2.7:")
```
in `py-python-dateutils`, a req that comes in via matplotlib.

```
spack spec py-xarray@2026: +viz
==> Error: failed to concretize `py-xarray@2026:+viz` for the following reasons:
     1. Cannot satisfy 'py-setuptools-scm@:7'
     2. Cannot satisfy 'py-setuptools-scm@:7' (version 9.2.2 does not match)
        required because py-python-dateutil depends on py-setuptools-scm@:7 when @2.7:
          required because py-matplotlib depends on py-python-dateutil@2.1:
            required because py-cartopy depends on py-matplotlib@3.4: when @0.22:
              required because py-xarray depends on py-cartopy@0.24: when @2026.4:+viz
                required because py-xarray@2026:+viz requested explicitly
              required because py-xarray depends on py-cartopy@0.23: when @2025.7:+viz

```

I'm not sure how to resolve this. cc/ @adamjstewart \

In the past this has worked, so presumably dateutils was using `py-setuptools-scm@8:` just fine. Is 7 a breaking API change?

adamjstewart:
If it's a build-only dep, it should be possible to install multiple versions in the same DAG. Problems may arise if it is a run-time dep of at least one package. @alalazo may be able to investigate.

Chrismarsh:
Looks like `py-setuptools-scm` is a `   tags = ["build-tools"]` so I'm unsure how to tell if there is a non-build usage of it

Chrismarsh:
Ok looks like the `@:7` pin is only for python 2.x so I have a PR that loosens this https://github.com/spack/spack-packages/pull/4935

PR #4678, opened 2026-05-05
Bug fixes for scalasca and scorep to work with Intel oneAPI in Spack v1

climbfuji:
## Description

Scalasca and Scorep have not been updated to work with Spack v1. This PR fixes it. I tested this with Intel oneAPI compilers+MPI on my laptop. Other updates for the MPI virtuals in the Scalasca recipe are copied over from the Scorep recipe.


climbfuji:
@spackbot fix style

climbfuji:
@spackbot fix style

repos/spack_repo/builtin/packages/scorep/package.py
```diff
@@ -138,6 +139,10 @@ def configure_args(self):
         ]
 
         cname = self.clean_compiler(spec.compiler.name)
+        # cname must be the legacy name of the compiler:
+        #  --with-nocross-compiler-suite=(gcc|ibm|intel|oneapi|nvhpc|pgi|clang|aocc|amdclang|cray)
+        if cname in BUILTIN_TO_LEGACY_COMPILER.keys():
+            cname = BUILTIN_TO_LEGACY_COMPILER[cname]
```
wrwilliams: This seems backwards to me. `clean_compiler` converts a (currently legacy) name to a valid argument for scorep's `configure`. So either `clean_compiler` gets the incoming cname preconverted to a legacy cname, which it converts to a scorep arg, or `clean_compiler` gets an incoming cname and does all necessary twiddling (including `BUILTIN_TO_LEGACY`) internally. I favor the second (and a rename to something better...maybe `to_scorep_compiler_arg` or similar?).

wrwilliams (commented):
Also looking into whether there's a better way than code duplication to handle the MPI package => MPI configure arg conversion.

repos/spack_repo/builtin/packages/scorep/package.py
climbfuji: Thanks for reviewing. Let me take a look at how to consolidate this best.

climbfuji: @wrwilliams I think commit https://github.com/spack/spack-packages/pull/4678/changes/7e063e4a26b778f5bb75a0792b86fe05fce75643 does what we want. The `amdclang` was confusing at first, because BUILTIN_TO_LEGACY would convert llvm-amdgpu to rocmcc, and then rocmcc would be converted to amdclang. Now it's in one go.

climbfuji:
@spackbot fix style

wrwilliams (approved):
LGTM now.

becker33 (approved):

PR #4679, opened 2026-05-05
Update vcftools

ebagrenrut:
- Add `vcftools` 0.1.17
- Fix issue with 0.1.16 and newer in which the path prepended to `PERL5LIB` was too shallow. Because these newer versions don't use `--with-pmdir` configure option, the environment setup needs to prepend self.prefix.lib/site_perl instead of just self.prefix.lib in so that `Vcf.pm` can be found by the interpreter.


mdorier:
Did you mean to push such a large PR? I don't know about the other packages but the modifications to mochi-thallium were already merged yesterday, which makes me think you did something incorrect with git...

teaguesterling:
> Did you mean to push such a large PR? I don't know about the other packages but the modifications to mochi-thallium were already merged yesterday, which makes me think you did something incorrect with git...

I most certainly did not! Thanks for catching. It was only small changes in 3 files at https://github.com/spack/spack-packages/pull/4679/changes/ea5f1229cb65e1c95406e4b5cdafbaeec367c566 and https://github.com/spack/spack-packages/pull/4679/changes/b73ab1aa5410849fdb6a904592c1e8469c31fd81. 

I must've pulled from develop and brought in a lot more than intended.

ebagrenrut:
Sorry; my fault. I think it happened because I rebased develop into my branch.

teaguesterling:
No worries. I'll re-create the PR.

PR #4681, opened 2026-05-05
py-xonsh: add v0.23.4

taliaferro:
Just a new version with checksum for py-xonsh.


repos/spack_repo/builtin/packages/py_xonsh/package.py
RikkiButler20, line 24:
```suggestion
    depends_on("py-setuptools@61:", type="build", when="@0.23.4:")
    depends_on("py-setuptools", type="build")
```
I think you need to add a version constraint on `py-setuptools`.
https://github.com/xonsh/xonsh/blob/4645eecf7c44d8478c7860c5299b31da0d21815c/pyproject.toml#L3

RikkiButler20 (changes requested):

mdorier:
The checksum is correct.

mdorier (approved):

RikkiButler20 (approved):
Looks good. Thanks

PR #4683, opened 2026-05-05
pfunit: add v4.18.1

mathomp4:
This adds pFUnit v4.18.1 which has a workaround found needed in testing with GCC 16.1.0 which had a bug now [fixed upstream in GCC](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=125117).

Note: The bug presented only in the pFUnit unit tests. Runs with GEOS had no issues.


RikkiButler20 (approved):
LGTM

PR #4684, opened 2026-05-05
vcftools: add v0.1.17

ebagrenrut:
- Add vcftools 0.1.17
- Fix issue with 0.1.16 and newer in which the path prepended to PERL5LIB was too shallow. Because these newer versions don't use --with-pmdir configure option, the environment setup needs to prepend self.prefix.lib/site_perl instead of just self.prefix.lib in so that Vcf.pm can be found by the interpreter.

### DIFF
--- a/repos/spack_repo/builtin/packages/vcftools/package.py
+++ b/repos/spack_repo/builtin/packages/vcftools/package.py
@@ -19,6 +19,7 @@ class Vcftools(AutotoolsPackage):
 
     license("LGPL-3.0-only")
 
+    version("0.1.17", sha256="b9e0e1c3e86533178edb35e02c6c4de9764324ea0973bebfbb747018c2d2a42c")
     version("0.1.16", sha256="dbfc774383c106b85043daa2c42568816aa6a7b4e6abc965eeea6c47dde914e3")
     # this is "a pre-release"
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/vcftools/package.py
```diff
@@ -85,4 +86,7 @@ def filter_sbang(self):
             filter_file(match, substitute, *files, **kwargs)
 
     def setup_run_environment(self, env: EnvironmentModifications) -> None:
-        env.prepend_path("PERL5LIB", self.prefix.lib)
+        if self.spec.satisfies("@0.1.16:"):
```
RikkiButler20, line 89:
```suggestion
        if self.spec.satisfies("@0.1.17:"):
```
Should this new path be for versions 0.1.16 or 0.1.17?

RikkiButler20 (changes requested):

repos/spack_repo/builtin/packages/vcftools/package.py
ebagrenrut: @RikkiButler20 No, it should be `@0.1.16:` because 0.1.16 does not currently work without it. This change _fixes_ 0.1.16.

RikkiButler20: Thanks for explaining.

RikkiButler20 (approved):

PR #4698, opened 2026-05-06
dftd4: add v4.1.0, v4.1.1 and v4.2.0

mkrack:

### DIFF
--- a/repos/spack_repo/builtin/packages/r_withr/package.py
+++ b/repos/spack_repo/builtin/packages/r_withr/package.py
@@ -19,6 +19,7 @@ class RWithr(RPackage):
 
     license("MIT")
 
+    version("3.0.2", sha256="0a3a05f493d275cca4bf13c8c1b95a1a4eed7f83b2493f41fde02ce3fc92c1a3")
     version("3.0.1", sha256="d573f1ac2f733c7dd89669feb495d68f1f15d7a9774473c2e46a9848945a3841")
     version("2.5.0", sha256="37317b3ed790a08407072993a05ab255f6305f95a12a16e0e28aa6aa80fc8bc0")
     version("2.4.3", sha256="9bdac7459ccc6c2d599ecfd132a7f0aa68d958942d9fe7dbb0442c9eda129d4c")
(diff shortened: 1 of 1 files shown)


awvwgk:
Can you add 4.1.1 and 4.2.0 as well?

mkrack:
> Can you add 4.1.1 and 4.2.0 as well?

Sure, but for the versions 4.1.1 and 4.2.0 no archive files with the pattern `dftd4-4.x.y-source.tar.xz` are provided for download. A [function](https://github.com/spack/spack-packages/blob/develop/repos/spack_repo/builtin/packages/dftd4/package.py#L68-L71) has already been introduced to account for the file name pattern change. I am not sure what the naming policy for dftd4 archive files is.

awvwgk (approved):

mkrack:
Please, could someone review this PR? Thanks!

repos/spack_repo/builtin/packages/dftd4/package.py
```diff
@@ -66,7 +69,9 @@ class Dftd4(MesonPackage, CMakePackage):
     extends("python", when="+python")
 
     def url_for_version(self, version):
-        if version <= Version("3.6.0") or version >= Version("4.0.0"):
+        if version <= Version("3.6.0") or (
+            version >= Version("4.0.0") and version <= Version("4.1.0")
+        ):
             return f"https://github.com/dftd4/dftd4/releases/download/v{version}/dftd4-{version}-source.tar.xz"
         return super().url_for_version(version)
```
alalazo, line 71: It seems the only outlier is v3.7.0. Yet, that's the only version treated in the usual way, while exceptions are made for everything else. Can we:
1. Change the `url` at class level to match the pattern of most of the versions
2. Add a `url=` argument to the v3.7.0 version declaration
3. Delete this function

?

alalazo (changes requested):

repos/spack_repo/builtin/packages/dftd4/package.py
mkrack: @alalazo Thanks for reviewing and your suggestions.
I find 3 outliers which have no `source.tar.xz` file for download: 3.7.0, 4.1.1 and 4.2.0. I will add an url arg to version for these and change the pattern at class level as you suggest.

PR #4699, opened 2026-05-06
py-typer: fix dependency on py-click

mdorier:
This PR forces py-typer to depend on a py-click dependency that's less than 8.2.0. py-click introduced a breaking change to its API in 8.2.0 and py-typer didn't seem to have adapted until their 0.24.0 version. In the meantime, using py-click with a version of py-typer greater or equal to 8.2.0 causes this error:

```
TypeError: Parameter.make_metavar() missing 1 required positional argument: 'ctx'
```

There are many newer versions of py-typer since 0.15.2 but I'll leave adding them to another PR as it will require tracking how each dependencies evolved. I just wanted this PR to fix the compatibility issue.


alecbcs (approved):
Looks good to me. Thanks @mdorier!

PR #4700, opened 2026-05-06
sratoolkit: add v3.4.1

snehring:
version bump

### DIFF
--- a/repos/spack_repo/builtin/packages/sratoolkit/package.py
+++ b/repos/spack_repo/builtin/packages/sratoolkit/package.py
@@ -17,11 +17,7 @@ class Sratoolkit(Package):
 
     maintainers("robqiao")
 
-    version(
-        "3.4.1",
-        sha256="f4850a32975596ca0755d928b166ecb0095918284ea992fba10f1acb3fcff64d",
-        url="https://ftp-trace.ncbi.nlm.nih.gov/sra/sdk/3.4.1/sratoolkit.3.4.1-alma_linux64.tar.gz",
-    )
+    version("3.4.1", sha256="f4850a32975596ca0755d928b166ecb0095918284ea992fba10f1acb3fcff64d")
     version("3.0.0", sha256="89d11fc284e32d38b6639c551df6d658075cd268913f8ba2af65026e4a161dfe")
     version("2.10.9", sha256="2c849b4b9865737ff17732e3befa70718616ce31cac98e8a61b1c5ed5a6514c5")
     version("2.10.7", sha256="b3f319974f0c7a318554d6383a13dd30f7d447533d92b6fd3bd057d3524e0140")
--- a/repos/spack_repo/builtin/packages/sratoolkit/package.py
+++ b/repos/spack_repo/builtin/packages/sratoolkit/package.py
@@ -17,6 +17,11 @@ class Sratoolkit(Package):
 
     maintainers("robqiao")
 
+    version(
+        "3.4.1",
+        sha256="f4850a32975596ca0755d928b166ecb0095918284ea992fba10f1acb3fcff64d",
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/sratoolkit/package.py
```diff
@@ -17,6 +17,11 @@ class Sratoolkit(Package):
 
     maintainers("robqiao")
 
+    version(
+        "3.4.1",
+        sha256="f4850a32975596ca0755d928b166ecb0095918284ea992fba10f1acb3fcff64d",
+        url="https://ftp-trace.ncbi.nlm.nih.gov/sra/sdk/3.4.1/sratoolkit.3.4.1-alma_linux64.tar.gz",
```
alecbcs: Is the URL required here? It looks like it's still following the format of the previous urls.

alecbcs (commented):
Quick question on the explicit URL. Otherwise looks good to me.

repos/spack_repo/builtin/packages/sratoolkit/package.py
snehring: the 'centos' changed to 'alma', wasn't sure if the fetcher would be clever enough to figure that out.

alecbcs: Ah! It probably wouldn't be that smart. Would you be willing to write a `url_for_version()` method that swaps between the two based on the version?

snehring: sounds good

repos/spack_repo/builtin/packages/sratoolkit/package.py
```diff
@@ -17,13 +17,22 @@ class Sratoolkit(Package):
...
     version("3.0.0", sha256="89d11fc284e32d38b6639c551df6d658075cd268913f8ba2af65026e4a161dfe")
     version("2.10.9", sha256="2c849b4b9865737ff17732e3befa70718616ce31cac98e8a61b1c5ed5a6514c5")
     version("2.10.7", sha256="b3f319974f0c7a318554d6383a13dd30f7d447533d92b6fd3bd057d3524e0140")
     version("2.9.6", sha256="faab687c822d0c02956f73f35e04875dde420ce9f602b88bbf3f2e8d79a17155")
     version("2.9.2", sha256="17dbe13aa1ed7955d31e1e76e8b62786e80a77e9ed9d396631162dc3ad8b716d")
     version("2.8.2-1", sha256="b053061aae7c6d00162fe0f514be4128a60365b4b2b5b36e7f4798b348b55cf5")
 
+    def url_for_version(self, version):
+        url = "https://ftp-trace.ncbi.nlm.nih.gov/sra/sdk/{0}/sratoolkit.{0}-{1}_linux64.tar.gz"
+        if self.spec.satisfies("@:3.0.0"):
```
alecbcs:
```suggestion
        if version <= Version("3.0.0"):
```

Unfortunately I think the right way to do this right now is to create a new version object and compare it against the version given to the method. (Until we fix `version.satisfies()` which doesn't currently work with strings.)

snehring: What broke with satisfies?

alecbcs: The issue is that we're not guaranteed to have a spec here. (For example creating a mirror with all versions of a package.) So we can't run `self.spec.satisfies` there's another separate `version.satisfies` method, but it doesn't act quite like `spec.satisfies` and that's something we're hoping to fix.

For now I'd stick with creating the version object and comparing against the version that's given to the method.

alecbcs (approved):

PR #4709, opened 2026-05-06
py-torch: patch for GCC-14.2 on aarch64 (abolish ICE)

wdconinc:
This PR adds a patch to `py-torch` to get past an [ICE](https://github.com/eic/containers/actions/runs/25415186561/job/74616049315?pr=266#step:15:9442) when using GCC-14.2 on aarch64. The issue was [introduced](https://github.com/pytorch/pytorch/commit/d6237721c074484ea5e72fc05614587886e57fd6) just before 2.9.0, and [fixed](https://github.com/pytorch/pytorch/commit/8fd509399e25cb4b265dff663d3f777406001f2e) after 2.10.0. A single line in a single file is patched. No when clause is added since this patch is not harmful on other compilers, on other versions of GCC, or on other architectures. Verified that the patch applies cleanly to the git tags of v2.9.0 and v2.10.0.


adamjstewart:
> abolish ICE

Thought this was a political statement at first 🤣

adamjstewart (approved):

PR #4714, opened 2026-05-06
r-rstantools: add v2.6.0

manuelakuhn:
https://cran.r-project.org/package=rstantools

### DIFF
--- a/repos/spack_repo/builtin/packages/r_rstantools/package.py
+++ b/repos/spack_repo/builtin/packages/r_rstantools/package.py
@@ -20,6 +20,7 @@ class RRstantools(RPackage):
 
     license("GPL-3.0-or-later")
 
+    version("2.6.0", sha256="6fe4d8f48e42bb48a0d65aedf017b4e1a919a29e3431301cc0295f7777a7f837")
     version("2.4.0", sha256="bff72ca2f0352c6c5d2868823e286fdb73a6ead74508a4124cbcb222c83b4faa")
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/r_rstantools/package.py
```diff
@@ -28,8 +29,10 @@ class RRstantools(RPackage):
...
-    depends_on("r-rcpp@0.12.16:", type=("build", "run"), when="@2.1.1:")
-    depends_on("r-rcppparallel@5.0.1:", type=("build", "run"), when="@2.1.1:")
-    depends_on("pandoc", type="build")
+    with default_args(type=("build", "run")):
+        depends_on("r+X")
+        depends_on("r-desc", when="@2.1.1:")
+        depends_on("r-rcpp@0.12.16:", when="@2.1.1:")
+        depends_on("r-rcppparallel@5.0.1:", when="@2.1.1:")
+
+    depends_on("pandoc")
```
kshea21, line 38: Why did you remove the `type="build"` from this dep?

manuelakuhn: I was following this guideline: https://spack.readthedocs.io/en/latest/build_systems/rpackage.html#non-r-dependencies
`pandoc` is listed as "SystemRequirements" and I don't see any reason why it should only be a build dependency. If you have different information, please let me know, then I can change it again.

kshea21: Okay, that makes sense and seems correct to me. Thanks for clarifying!

kshea21 (approved):

PR #4715, opened 2026-05-06
remove codeowners who do not have repo write access

becker33:
As discussed in the TSC meeting today. Github does not play nicely with a codeowners file with users who do not have repo write access.


alecbcs (approved):
Looks good to me. Thanks for catching this.

PR #4718, opened 2026-05-07
compiler-wrapper: set `-frandom-seed="$*"` for gcc

haampie:
To make builds deterministic with GCC, use [`-frandom-seed`][1] with the
value set to all command line arguments. GCC uses the CRC checksum of
this string to initialize its random number generator.

LLVM is deterministic by default.

One concern could be "Argument list too long", but I think this one extra
flag is typically smaller than all the `-I`, `-L`, `-Wl,` flags we inject.

[1]: https://gcc.gnu.org/onlinedocs/gcc/Developer-Options.html#index-frandom-seed

### DIFF
--- a/repos/spack_repo/builtin/packages/compiler_wrapper/cc.sh
+++ b/repos/spack_repo/builtin/packages/compiler_wrapper/cc.sh
@@ -365,20 +365,17 @@ if [ -z "$mode" ] || [ "$mode" = ld ]; then
     done
 fi
 
-# Finish setting up the mode.
+# Finish setting up the mode, and check whether -frandom-seed needs to be set.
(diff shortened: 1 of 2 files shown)


haampie:
Regarding binutils, we can assume it's configured with `--enable-deterministic-archives` (Spack default, and a default in various distros)

The only non-determinism can come from `ar -U`, which should then be fixed with `ARFLAGS=...`, a `@run_after` hook, or a patch.

haampie:
Converted to draft cause I've submitted this to `spack/compiler-wrapper`, and we need #4829 first.

haampie:
Combined with #4829 to avoid double full rebuild.

PR #4724, opened 2026-05-07
acts: add v46.4.0 through v46.8.1

stephenswat:
This commit adds versions 46.4.0 through 46.8.1 of the ACTS package. Comes with new dependency versions for vecmem and detray. Also updates missing dependency updates from version 44.1.0.

Depends on #4568.

### DIFF
--- a/repos/spack_repo/builtin/packages/acts/package.py
+++ b/repos/spack_repo/builtin/packages/acts/package.py
@@ -50,6 +50,12 @@ class Acts(CMakePackage, CudaPackage):
 
     # Supported Acts versions
     version("main", branch="main")
+    version("46.8.1", commit="3ded307e142c6ce9e6c1fa205604582ff6aef920")
+    version("46.8.0", commit="59d95bd56db22034219a526e7dff563622cc268d")
+    version("46.7.0", commit="4e95978f8e23f863f9e7c6b219b5100531f4ba41")
+    version("46.6.0", commit="7791df44925fadd1de375c42c41d36a6f82602d4")
+    version("46.5.0", commit="34edd48852f766e1b9d94d3dc996e27476339f1b")
+    version("46.4.0", commit="4f0e5d8ce158728ff275a733b720d087065563bd")
     version("46.3.0", commit="4f0e5d8ce158728ff275a733b720d087065563bd")
     version("46.2.0", commit="fcecf616be7ddd240700f65a1508f21811ee4167")
     version("46.1.0", commit="d6d17c9c2b28a54141477438d63176a8af2d3c76")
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/acts/package.py
```diff
@@ -198,6 +199,7 @@ class Acts(CMakePackage, CudaPackage):
     depends_on("dd4hep +ddg4", when="+dd4hep +geant4 +examples")
     depends_on("detray @0.75.3:", when="+traccc")
     depends_on("detray @0.101.0:", when="@42.1: +traccc")
```
wdconinc, line 208: Looking at the release diff, this was min-version 0.103.0 before. Did we forget to add a bump for detray somewhere?

repos/spack_repo/builtin/packages/acts/package.py
```diff
@@ -249,6 +251,7 @@ class Acts(CMakePackage, CudaPackage):
         depends_on("root @6.28:", when="@42:")
 
     depends_on("vecmem@1.17.0:", when="@42: +traccc")
```
wdconinc, line 261: Same thing here, vecmem min-version was 1.21 before.

repos/spack_repo/builtin/packages/acts/package.py
wdconinc, line 273: Does the `+traccc` dependency work? We are not depending on `traccc` at all (as there is no `traccc` package in spack). And our `ACTS_USE_SYSTEM_LIBS=ON` should cause the Acts CMakeLists.txt to call `find_package(traccc)`.

wdconinc (commented):
Thanks. A few comments, but they are all probably about previous releases where we maybe missed some dependency requirements.

repos/spack_repo/builtin/packages/acts/package.py
stephenswat: Correct, I forgot to update the bump in 44.1.0. I'm adding it now.

stephenswat: Updated now.

repos/spack_repo/builtin/packages/acts/package.py
stephenswat: Also bumped in 44.1.0, now added to this PR.

repos/spack_repo/builtin/packages/acts/package.py
stephenswat: Admittedly this does not work right now, which is a problem. I think people have been using this only to set up development environments but I don't think there is a good way to make this work with ACTS right now. In the coming weeks, we will finally be able to properly build traccc with ACTS though, so as much as it is technical debt I'd keep this variant for now and then clean up the dependencies asap. What do you think?

stephenswat:
@wdconinc updated to address your comments, and also added versions 46.5.0 and 46.6.0.

stephenswat:
Updated to add new versions up to and including 46.8.1.

wdconinc (approved):
LGTM.

PR #4742, opened 2026-05-07
flecsi: update package requirements

rbberger:
- allow building with `amdclang++` directly
- allow building with Kokkos 5
- avoid `kokkos_cxx` logic in non-CUDA cases
- mark documentation deps as build-only

### DIFF
--- a/repos/spack_repo/builtin/packages/flecsi/package.py
+++ b/repos/spack_repo/builtin/packages/flecsi/package.py
@@ -75,7 +75,8 @@ class Flecsi(CMakePackage, CudaPackage, ROCmPackage):
     depends_on("kokkos", when="+kokkos @2.3:")
     depends_on("kokkos", when="@2.4:")
     depends_on("kokkos +cuda", when="+kokkos +cuda")
-    requires("^kokkos +cuda_constexpr +cuda_lambda", when="^kokkos +cuda")
+    requires("^kokkos +cuda_lambda", when="^kokkos@:4 +cuda")
+    requires("^kokkos +cuda_constexpr", when="^kokkos +cuda")
     depends_on("kokkos +rocm", when="+kokkos +rocm")
     depends_on("kokkos +openmp", when="+kokkos +openmp")
     requires("+openmp", when="@:2.3 ^kokkos +openmp")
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/flecsi/package.py
```diff
@@ -143,13 +144,15 @@ def cmake_args(self):
...
+        if self.spec.satisfies("^kokkos +rocm") and not self.spec.satisfies(
+            "^kokkos %cxx=llvm-amdgpu"
+        ):
             options.append(self.define("CMAKE_CXX_COMPILER", self.spec["hip"].hipcc))
             options.append(self.define("CMAKE_C_COMPILER", self.spec["hip"].hipcc))
             if self.spec.satisfies("backend=legion"):
                 # CMake pulled in via find_package(Legion) won't work without this
                 options.append(self.define("HIP_PATH", "{0}/hip".format(spec["hip"].prefix)))
-        elif self.spec.satisfies("^kokkos"):
+        elif self.spec.satisfies("^kokkos +cuda"):
```
opensdh, line 155: I think I understand the rest of these changes, but what's happening here?  We don't want to use `kokkos_cxx` at all with `%cxx=llvm-amdgpu`, I suppose, but why is it gated on `+cuda` rather than just starting with
```py
if self.spec.satisfies("^kokkos %cxx=llvm-amdgpu"): pass
elif self.spec.satisfies("^kokkos +rocm"):
```
?

rbberger: Updated the description. It's an unrelated change.

repos/spack_repo/builtin/packages/flecsi/package.py
```diff
@@ -95,13 +96,13 @@ class Flecsi(CMakePackage, CudaPackage, ROCmPackage):
...
     # FleCSI documentation dependencies
-    depends_on("py-sphinx", when="+doc")
-    depends_on("py-sphinx-rtd-theme@:2", when="+doc")
-    depends_on("py-recommonmark", when="@:2.2 +doc")
-    depends_on("doxygen", when="+doc")
+    depends_on("py-sphinx", when="+doc", type="build")
+    depends_on("py-sphinx-rtd-theme@:2", when="+doc", type="build")
+    depends_on("py-recommonmark", when="@:2.2 +doc", type="build")
+    depends_on("doxygen", when="+doc", type="build")
     depends_on("graphviz", when="+doc")
```
opensdh: Isn't this one (but _not_ the one based on `+graphviz`) also build-only?

rbberger: You're right. Fixed.

opensdh (approved):

hainest (approved):
LGTM.

PR #4745, opened 2026-05-08
perl: set TZ=UTC for reproducibility

alalazo:


alalazo:
Not sure if this should be done more generally, on the Spack-tool side, when we setup the build environment

haampie:
I think it's fine per-package, otherwise it's just taking env bytes redundantly (albeit not that many).

haampie (approved):

PR #4748, opened 2026-05-08
py-xgboost@1 : fix lib64 support

### LINKED ISSUE #4677: Installation issue: py-xgboost

```console
$ spack install py-xgboost@1.6
...
```

### Error message

@adamjstewart 
patches for xgboost paths don't work in older versions as they only replace the prefix but xgboost library is installed in `<xgboost prefix>/lib64`, whereas setup looks for `<xgboost prefix>/lib`.
Older version of those patches (spack 0.21) where replacing the full library path:
```
    def patch(self):
        # https://github.com/dmlc/xgboost/issues/6706
        # 'setup.py' is hard-coded to search in Python installation prefix
        filter_file(
            "lib_path = os.path.join(sys.prefix, 'lib')",
            "lib_path = '{0}'".format(self.spec["xgboost"].libs.directories[0]),
            "setup.py",
            string=True,
        )

        # Same for run-time search
        filter_file(
            "os.path.join(curr_path, 'lib'),",
            "'{0}',".format(self.spec["xgboost"].libs.directories[0]),
            os.path.join("xgboost", "libpath.py"),
            string=True,
        )
```
whereas now:
```
    def patch(self):
        # Hard-coded to search for system libxgboost in the Python installation prefix
        # https://github.com/dmlc/xgboost/issues/6706
        files = [os.path.join("xgboost", "libpath.py")]
        if self.spec.satisfies("@2:"):
            regex = "sys.base_prefix"
            files.append(os.path.join("packager", "nativelib.py"))
        else:
            regex = "sys.prefix"
            files.append("setup.py")
        filter_file(regex, repr(self.spec["xgboost"].prefix), *files, string=True)
```

* **Spack:** 1.1.0 (https://github.com/spack/spack/commit/0c2be44e4ece21eb091ad5de4c97716b7c6d4c87)
* **Builtin repo:** https://github.com/spack/spack-packages/commit/b46f6bcf508b243d2752710ce5e940e474a45d86
* **Python:** 3.12.12
* **Platform:** linux-rocky8-x86_64

### Additional information

_No response_

### General information

adamjstewart:
Fixes #4677 @fgava90 

We already had a patch for xgboost 2, just needed to add a patch for xgboost 1. Note that xgboost 3 is out now, I may consider deprecating 1 and 2.


repos/spack_repo/builtin/packages/py_xgboost/package.py
```diff
@@ -71,6 +70,9 @@ class PyXgboost(PythonPackage):
             depends_on("py-graphviz")
             depends_on("py-matplotlib")
 
+    patch("add-lib64-1.patch", when="@1")
+    patch("add-lib64-2.patch", when="@2:")
```
alalazo, line 75: Can you add a brief comment above the `patch` directives explaining:
2. Why it is needed
3. Where it was taken from

?

adamjstewart: 1/2: will do. 3: me

alalazo (approved):

PR #4751, opened 2026-05-08
ruff: automatically fix issues

adamjstewart:
Allows `ruff check` to automatically fix things missed by `ruff format`.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -16,6 +16,7 @@ dev = [
 ci = ["pytest-cov", "codecov[toml]"]
 
 [tool.ruff]
+fix = true
 line-length = 99
 
 # This is the Spack package API. We allow these functions to be imported with


johnwparent (approved):
I think this approach makes the most sense, I think most folks are just interested in having their linter issues fixed, vs being made aware of them and having to run ruff a second time to actually fix them (not that that's particularly expensive). 
I can't remember the last time I ran `ruff check` without `--fix`.

Shouldn't cause any issues with the `spack style --repo` feature, I can just have the non fix version run `--no-fix`.

I'll let other folks give their two cents before merging.

One question: Should we do this for core?

adamjstewart:
We should also do this for core, yes. Want to open a PR on that side?

johnwparent:
> We should also do this for core, yes. Want to open a PR on that side?

On it!

~~ My only point of concern is the behavior of `spack style --repo --fix` will change depending on the commit of core and packages that are checked out. I.e. if you have a core that flips `--fix` to set no cli arg and not fix to set  `--not-fix` and a packages without the changes in this PR, `--fix` won't work.  ~~

~~That concern doesn't apply to core.~~

Concern resolved. Just leaving `--fix` in core.

adamjstewart:
Didn't think about that. I don't use spack style but I think we can override the default settings in pyproject.toml?

johnwparent:
> Didn't think about that. I don't use spack style but I think we can override the default settings in pyproject.toml?

Oh whoops, looks like the strike through didn't work. I think not changing the core style behavior and leaving a redundant `--fix` in will prevent this being an issue.

adamjstewart:
`spack style --fix` uses `ruff check --fix` and `spack style` uses `ruff check --no-fix`, so it should override the settings in pyproject.toml regardless. I think we should make the same change in core so that `ruff check` without `spack style` will auto-fix. Whether `spack style --fix` should be the default behavior or not is a different question I don't care about answering.

PR #4755, opened 2026-05-08
gchp: Add released versions

tkoskela:
I have added all tagged versions from the gchp upstream repo to the package recipe. This also required fine-tuning some versions of the dependencies.


repos/spack_repo/builtin/packages/gchp/package.py
```diff
@@ -19,7 +19,31 @@ class Gchp(CMakePackage):
 
     license("MIT")
 
+    version("14.7.1", commit="503642dfe2b02c219e69fed52e027faea1a0063d", submodules=True)
```
kshea21, line 22: Hi! I'm noticing that the checksums you've provided for the added `gchp` versions aren't matching what I'm getting when I run `spack checksum gchp <version>`. I'm looking into it, but for now I'm going to hold off on the PR getting merged. I'll report back what I figure out soon!

kshea21 (changes requested):

repos/spack_repo/builtin/packages/gchp/package.py
kshea21: I'm noticing that there are checksum mismatches for versions of `gchp` prior to this PR as well.

kshea21: Ah, my fault, I was checking for a SHA256 checksum, not commits.

tkoskela: The package was using commit hashes so I stuck to their style. Do you have a preference between commit hashes and checksums?

repos/spack_repo/builtin/packages/gchp/package.py
```diff
@@ -38,8 +62,11 @@ class Gchp(CMakePackage):
     depends_on("fortran", type="build")  # generated
 
     depends_on("esmf@8.0.1", when="@:13")
-    depends_on("esmf@8.8.0:", when="@:14")
-    depends_on("udunits@2", when="@:14")
+    depends_on("esmf@8.0.0:", when="@14.0:14.0")
+    depends_on("esmf@8.1.0:", when="@14.1:14.1")
```
kshea21: These should become just : 
```suggestion
    depends_on("esmf@8.0.0:", when="@14.0")
    depends_on("esmf@8.1.0:", when="@14.1")
```
if there's no range.

kshea21 (changes requested):

repos/spack_repo/builtin/packages/gchp/package.py
kshea21: Commits are perfectly fine, I just got mixed up.

kshea21:
Does this all look good to you, @lizziel and @laestrada ?

lizziel:
Bringing in @yantosca since who is the GCST member supporting GEOS-Chem in spack.

yantosca (approved):

kshea21 (approved):
Great, thanks!

PR #4769, opened 2026-05-08
llvm-amdgpu: add rocm.cfg

afzpatel:
The llvm-amdgpu installation is missing a `rocm.cfg` file that should be generated post-install which contains `-frtlib-add-rpath`. Adding this `rocm.cfg` will eliminate the need for `002-Add-rpath-to-hiprt.patch‎`.

Also adding required python dependency to comgr:
https://github.com/ROCm/llvm-project/blob/rocm-7.2.3/amd/comgr/cmake/DeviceLibs.cmake#L113


srekolam (approved):
LGTM

renjithravindrankannath (approved):

repos/spack_repo/builtin/packages/llvm_amdgpu/package.py
```diff
@@ -470,6 +469,14 @@ def post_install(self):
                 with open(os.path.join(self.prefix.bin, cfg), "w") as f:
                     print(gcc_install_dir_flag, file=f)
 
+        if self.spec.satisfies("@7:"):
+            with open(os.path.join(self.prefix.bin, "rocm.cfg"), "w") as f:
+                print("-Wl,--enable-new-dtags", file=f)
```
becker33: Please do not enable new dtags here. This breaks rpath support, and spack already has mechanisms for determining whether to use old or new dtags.

becker33 (changes requested):
I think we should remove the patch if it's no longer used.

repos/spack_repo/builtin/packages/llvm_amdgpu/package.py
afzpatel: I removed it

afzpatel:
@spackbot rerun pipeline

afzpatel:
@becker33 please review this PR again

becker33 (approved):

PR #4772, opened 2026-05-08
intel-pin: remove compiler dependencies

berquist:
Even though Pin includes source code, nothing requires compilation; the installer just extracts files into a tree.


matthiasdiener (approved):

alecbcs (approved):
Looks good to me. Thanks @berquist!

PR #4776, opened 2026-05-10
kpblas: new package

wjunLu:
## Description

Add a new wrapper package for **KML BLAS (kpblas)** that provides the `blas` virtual package via the `hpckit` dependency on aarch64 Kunpeng platforms.

Kpblas is the BLAS component of Kunpeng Math Library (KML) bundled within HPCKit — this wrapper package resolves the kpblas library and headers from an installed `hpckit` dependency.

- Homepage: https://www.hikunpeng.com/developer/hpc/hpckit-download

## Versions

| Version  | Notes  |
|----------|--------|
| `25.1.0` | Requires hpckit@25.1.0 |

## Build System

Uses `Package` (generic). The `install()` phase creates a minimal `.spack_wrapper` marker file since kpblas binaries are provided by the `hpckit` dependency.

## Dependencies

- `hpckit@25.1.0` — provides the actual KML BLAS libraries

## Variants

- `compiler_variant`: `gcc` (default) or `bisheng` — selects the compiler-specific KML subtree within HPCKit
- `kblas_mode`: `multi` (default), `locking`, or `nolocking` — KML BLAS threading mode

## Features

- Provides `blas` virtual package
- Auto-detects CPU ISA (sve/sme/neon) for selecting correct library path
- Sets `blas_libs` for dependent packages


repos/spack_repo/builtin/packages/kpblas/package.py
```diff
@@ -0,0 +1,138 @@
...
+class Kpblas(Package):
+    """KML BLAS wrapper for HPCKit - provides blas virtual package."""
+
+    homepage = "https://www.hikunpeng.com/developer/hpc/hpckit-download"
+
+    has_code = False
+
+    license("UNKNOWN", checked_by="github_user1")
+
+    version("25.1.0")
```
hainest: Please add a sha256.

repos/spack_repo/builtin/packages/kpblas/package.py
```diff
@@ -0,0 +1,138 @@
...
+
+    # 依赖基础包，版本必须匹配
+    depends_on("hpckit@25.1.0", when="@25.1.0")
```
hainest: If there is a new version of kpblas, say 25.1.1, should it also depend on hpckit@25.1.0? If so, then this should be


```suggestion
    depends_on("hpckit@25.1.0", when="@25.1.0:")
```

hainest (changes requested):

wjunLu (commented):
Fixed

wjunLu (commented):
Fixed

wjunLu (commented):
.

repos/spack_repo/builtin/packages/kpblas/package.py
wjunLu: yes, thanks for your suggestion

PR #4779, opened 2026-05-10
geant4: set optional data variants

sethrj:
Some of the installed datasets aren't exported into the Geant4 config file due to some missing g4 config variables. This makes `spack load geant4-data` consistent with using the `Geant4_DATASET_DESCRIPTIONS` in `Geant4Config.cmake`.


wdconinc (dismissed):
LGTM.

wdconinc:
@sethrj but needs style fix.

sethrj:
@spackbot fix style

sethrj:
Sorry @wdconinc , should've fixed the style before requesting reviews. Needs reapproval. Thanks!

wdconinc (approved):

PR #4781, opened 2026-05-10
mars-earth: new package

edithatogo:
Add an H0-only `mars-earth` package definition for the Spack package repository.

This recipe covers the source-installable Python package with its Rust-backed runtime and scikit-learn compatible dependency stack. It intentionally does not claim accelerator, MPI, or distributed execution support.

Validation:
- Python syntax check on the new package file
- `git diff --check`

### DIFF
--- a/repos/spack_repo/builtin/packages/mars_earth/package.py
+++ b/repos/spack_repo/builtin/packages/mars_earth/package.py
@@ -1,0 +1,30 @@
+# Copyright Spack Project Developers. See COPYRIGHT file for details.
+#
+# SPDX-License-Identifier: (Apache-2.0 OR MIT)
+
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/mars_earth/package.py
```diff
@@ -0,0 +1,30 @@
+# Copyright Spack Project Developers. See COPYRIGHT file for details.
+#
+# SPDX-License-Identifier: (Apache-2.0 OR MIT)
+
+from spack_repo.builtin.build_systems.python import PythonPackage
+
+from spack.package import *
+
+
+class MarsEarth(PythonPackage):
```
tgamblin, line 10: needs to be PyMarsEarth to follow our naming conventions

PR #4784, opened 2026-05-11
expat: 2.8.1 + deprecate vulnerable 2.8.0 (for CVE-2026-45186)

hartwork:
Similar to #4506

Related blog post: https://blog.hartwork.org/posts/expat-2-8-1-released/

CVE-2026-45186

CC @wdconinc

### DIFF
--- a/repos/spack_repo/builtin/packages/expat/package.py
+++ b/repos/spack_repo/builtin/packages/expat/package.py
@@ -18,7 +18,10 @@ class Expat(AutotoolsPackage, CMakePackage):
     url = "https://github.com/libexpat/libexpat/releases/download/R_2_2_9/expat-2.2.9.tar.bz2"
 
     license("MIT")
-    version("2.8.0", sha256="586494499ac3ad46d87f3beda7b1f770c1c8026a9b60e151593f8b29089a52ca")
+    version(
+        "2.8.0",
+        sha256="586494499ac3ad46d87f3beda7b1f770c1c8026a9b60e151593f8b29089a52ca",
+    )
     # deprecate all releases before 2.8.0 because of various security issues
     version(
         "2.7.5",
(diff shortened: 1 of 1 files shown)


hartwork:
@wdconinc PS: Please note that I reported the vulnerability to Mitre as "remote" but someone later mis-classified it to "complex" and "local" when it can be exploited easily, even without the attack payload generator that is public and ready to download from the description of https://github.com/libexpat/libexpat/pull/1216 .

cve-2026-45186_cvss

(as seen at https://github.com/advisories/GHSA-r396-2q2c-pjhr, emphasis mine)

A good example and reminder that CVSS are not a reliable metric to make decisions about vulnerabilities: vulnerabilities just all need to be be fixed, no matter "the score".

Related: https://daniel.haxx.se/blog/2025/01/23/cvss-is-dead-to-us/

CC @hannob

wdconinc (approved):

wdconinc:
> A good example and reminder that CVSS are not a reliable metric to make decisions about vulnerabilities: vulnerabilities just all need to be be fixed, no matter \"the score\".

Very useful input into how we should best handle CVSS. We are always happy to accept contributed deprecations of older versions and adds of newer versions, regardless of the score. We do use CVSS to make some prioritization internally when we actively 'hunt' for packages that may need to be updated, but this is indeed a tricky process (spack being used in HPC contexts means that it's unlikely someone will run a server open to the world, but it's also a more desirable target for LPE than an endpoint PC). We also sometimes have to make judgement calls when an affected version is a required version for a widely used dependency.

Thanks for this input!

hartwork:
For the record and anyone coming here, I have just asked Mitre to fix the classification now via web form https://cveform.mitre.org/ . I hope that helps.

wdconinc:
I just noticed this was blocked by an unrelated CI issue. Can you rebase or update the branch to give it another go?

hartwork:
@wdconinc done :+1:

PR #4785, opened 2026-05-11
theia-ide: add v1.70.200

RobertMaaskant:
Diff: https://github.com/eclipse-theia/theia-ide/compare/v1.59.1...v1.70.200

### DIFF
--- a/repos/spack_repo/builtin/packages/theia_ide/package.py
+++ b/repos/spack_repo/builtin/packages/theia_ide/package.py
@@ -51,6 +51,9 @@ class TheiaIde(Package):
         # https://github.com/nodejs/node-gyp/tree/v9.0.0?tab=readme-ov-file#on-unix
         depends_on("python@3.7:3.10")
 
+    with default_args(type=("build", "link")):
+        depends_on("zlib-api")
+
     with default_args(type="run"):
         depends_on("git@2.11.0:")
 
@@ -59,7 +62,6 @@ class TheiaIde(Package):
         # https://github.com/microsoft/vscode/blob/1.119.0/.nvmrc
         depends_on("node-js@22.22.1:22", when="@1.70.200")
 
-        depends_on("zlib-ng")
 
     def install(self, spec, prefix):
         yarn = which("yarn", required=True)
--- a/repos/spack_repo/builtin/packages/theia_ide/package.py
+++ b/repos/spack_repo/builtin/packages/theia_ide/package.py
@@ -18,7 +18,12 @@ class TheiaIde(Package):
 
     license("MIT", checked_by="RobertMaaskant")
 
-    version("1.59.1", sha256="f3e4fdb76aa0d5a4f034d9fe8889e8c798425d8c0a452688277002b7f09ea7d0")
+    version("1.70.200", sha256="c110f7e13b465fe78cd89e251554729ba6135fa735ff5dbd0dcd7d5e7fb623c5")
+    version(
+        "1.59.1",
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/theia_ide/package.py
```diff
@@ -47,8 +53,13 @@ class TheiaIde(Package):
...
     with default_args(type="run"):
         depends_on("git@2.11.0:")
+
         # https://github.com/microsoft/vscode/blob/1.98.2/.nvmrc
-        depends_on("node-js@20.18.2:20")
+        depends_on("node-js@20.18.2:20", when="@1.59.1")
+        # https://github.com/microsoft/vscode/blob/1.119.0/.nvmrc
+        depends_on("node-js@22.22.1:22", when="@1.70.200")
+
+        depends_on("zlib-ng")
```
alecbcs: Is `zlib-ng` really a run dependency here? (Just double checking it didn't get accidentally placed in here under the default_args() since most times it's a build/link dependency.)

alecbcs (commented):
Quick question. Otherwise looks good to me.

repos/spack_repo/builtin/packages/theia_ide/package.py
RobertMaaskant: I must admit, I am not totally sure. I known what the differences are, but usually only deal with Python and Go.

I found the dependency through:

```
spack verify libraries theia-ide
==> Checking 1 packages for shared library resolution
==> Error: theia-ide@1.59.1/7grhexc3one55irqhgm6elebzhzhaxqi: /home/robert/code/github.com/spack/spack/opt/spack/linux-zen4/theia-ide-1.59.1-7grhexc3one55irqhgm6elebzhzhaxqi:
    bin/theia-ide
            libz.so.1 => not found
==> Error: Cannot resolve shared libraries in 1 package
```

Will look into this and get back to you. Any tips how to determine which type it is, would be welcome.

alecbcs: If it's looking for `libz.so.1` it's likely a build and link dependency (the default type). 

Instead of depending on `zlib-ng` however, I'd recommend depending on `zlib-api` which is the virtual package.

RobertMaaskant: Fixed the zlib dependency. Thanks for the help.

repos/spack_repo/builtin/packages/theia_ide/package.py
```diff
@@ -38,17 +43,25 @@ class TheiaIde(Package):
...
+        depends_on("npm@10.8.2:", when="@1.59.1")
+        depends_on("npm@10.9.7:", when="@1.70.200")
         depends_on("yarn@1.7.0:1")
 
         # required by node-gyp ^9.0.0: https://github.com/eclipse-theia/theia/blob/v1.59.0/package.json#L45
         # https://github.com/nodejs/node-gyp/tree/v9.0.0?tab=readme-ov-file#on-unix
         depends_on("python@3.7:3.10")
 
+    with default_args(type=("build", "link")):
+        depends_on("zlib-api")
```
alecbcs: `type=("build", "link")` is actually the default for dependencies so we can rewrite this as simply

```suggestion
    depends_on("zlib-api")
```

RobertMaaskant: Thanks for the suggestion to simplify. Applied as suggested.

RobertMaaskant:
@alecbcs good to merge, or more changes required?

alecbcs (approved):
Looks good to me. Thanks @RobertMaaskant

PR #4786, opened 2026-05-11
pelican: new package, v7.24.3

wdconinc:
This PR adds `pelican`, https://pelicanplatform.org. It's used with the HTC community for data distribution, in particular for the [Open Science Data Federation](https://osg-htc.org/services/osdf.html). This package include the client by default, and installs the server by variant.

Test build:
```
-- linux-ubuntu26.04-skylake / no compilers ---------------------
zliolvk pelican@7.24.3+server build_system=go

-- linux-ubuntu26.04-x86_64 / no compilers ----------------------
2akzz7u go@1.26.0 build_system=generic
==> 2 installed packages
```

### DIFF
--- a/repos/spack_repo/builtin/packages/pelican/package.py
+++ b/repos/spack_repo/builtin/packages/pelican/package.py
@@ -1,0 +1,66 @@
+# Copyright Spack Project Developers. See COPYRIGHT file for details.
+#
+# SPDX-License-Identifier: (Apache-2.0 OR MIT)
+
+from spack_repo.builtin.build_systems.go import GoPackage
+
+from spack.package import *
+
+
+class Pelican(GoPackage):
+    """The Pelican command line tool allows one to use a Pelican
+    federation as a client and serve datasets through running a
+    Pelican origin service."""
+
+    homepage = "https://pelicanplatform.org/"
+    url = "https://github.com/PelicanPlatform/pelican/archive/refs/tags/v7.24.3.tar.gz"
+    git = "https://github.com/PelicanPlatform/pelican.git"
+
+    maintainers("wdconinc")
+
+    license("Apache-2.0", checked_by="wdconinc")
+
+    version("7.24.3", sha256="fd73d4c9193f3a25d82c696e7269d10f84f4941786e957c8dffa1562d4b605a0")
+
+    variant("server", default=False, description="Also build the pelican-server binary")
+
+    @run_before("build")
+    def create_frontend_placeholder(self):
+        # web_ui/ui.go uses //go:embed frontend/out/*, which requires at least one
+        # file to exist. Releases prior to v7.25 did not commit this placeholder,
+        # so we create it here to satisfy the embed directive without a full frontend
+        # build.
+        out_dir = join_path(self.stage.source_path, "web_ui", "frontend", "out")
+        mkdirp(out_dir)
+        touch(join_path(out_dir, "placeholder"))
+
+    @property
+    def build_directory(self):
+        # The main package entrypoint is in the cmd/ subdirectory
+        return join_path(self.stage.source_path, "cmd")
+
+    @property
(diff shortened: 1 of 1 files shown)


wdconinc:
@alecbcs Is there a better way to reuse GoPackage with multi-executable packages? It seems to reuse pkg.name as executable. I'm not as steeped in the Go ecosystem to know if that's a reasonable thing to support in GoPackage or just something pelican is doing differently in a non-recommended way.

repos/spack_repo/builtin/packages/pelican/package.py
```diff
@@ -0,0 +1,68 @@
...
+        args.extend(["-tags", "forceposix,client"])
+        return args
+
+    @run_after("build", when="+server")
+    def build_server(self):
+        args = list(self.build_args)
+        args[args.index("-tags") + 1] = "forceposix,server"
+        args[args.index("-o") + 1] = "pelican-server"
+        with working_dir(self.build_directory):
+            self.module.go("build", *args)
```
alecbcs:
```suggestion
            go("build", *args)
```

alecbcs (commented):
Optional but modules are automatically added to downstream packages so they can just be called with `module()` to make this a bit cleaner.

alecbcs:
> @alecbcs Is there a better way to reuse GoPackage with multi-executable packages? It seems to reuse pkg.name as executable. I'm not as steeped in the Go ecosystem to know if that's a reasonable thing to support in GoPackage or just something pelican is doing differently in a non-recommended way.

I'm working on a couple fixes to the `GoPackage` builder that should make this easier in the future. For now I'd say this is the right way to go about this even though it's ugly.

repos/spack_repo/builtin/packages/pelican/package.py
```diff
@@ -0,0 +1,68 @@
...
+        # so we create it here to satisfy the embed directive without a full frontend
+        # build.
+        out_dir = join_path(self.stage.source_path, "web_ui", "frontend", "out")
+        mkdirp(out_dir)
+        touch(join_path(out_dir, "placeholder"))
+
+    @property
+    def build_directory(self):
+        # The main package entrypoint is in the cmd/ subdirectory
+        return join_path(self.stage.source_path, "cmd")
```
alecbcs:
```suggestion
    build_directory = "cmd"
```

If the build directory provided is relative it is appended to the stage source path.

repos/spack_repo/builtin/packages/pelican/package.py
```diff
@@ -0,0 +1,68 @@
...
+
+    @run_before("build")
+    def create_frontend_placeholder(self):
+        # web_ui/ui.go uses //go:embed frontend/out/*, which requires at least one
+        # file to exist. Releases prior to v7.25 did not commit this placeholder,
+        # so we create it here to satisfy the embed directive without a full frontend
+        # build.
+        out_dir = join_path(self.stage.source_path, "web_ui", "frontend", "out")
+        mkdirp(out_dir)
+        touch(join_path(out_dir, "placeholder"))
```
alecbcs:
```suggestion
    @run_before("build", when="@:7.24")
    def create_frontend_placeholder(self):
        # web_ui/ui.go uses //go:embed frontend/out/*, which requires at least one
        # file to exist. Releases prior to v7.25 did not commit this placeholder,
        # so we create it here to satisfy the embed directive without a full frontend
        # build.
        out_dir = join_path(self.stage.source_path, "web_ui", "frontend", "out")
        mkdirp(out_dir)
        touch(join_path(out_dir, "placeholder"))
```

Let's add a when clause here to limit this placeholder creation to before when it was committed to the repo.

alecbcs (changes requested):
Two quick suggestions. Otherwise looks good to me!

alecbcs (approved):

PR #4788, opened 2026-05-11
rocm: improve external detection

afzpatel:
same PR as: https://github.com/spack/spack-packages/pull/4350


repos/spack_repo/builtin/packages/hipify_clang/package.py
```diff
@@ -89,12 +88,6 @@ class HipifyClang(CMakePackage):
         depends_on(f"llvm-amdgpu@{ver}", when=f"@{ver}")
         depends_on(f"rocm-core@{ver}", when=f"@{ver}")
 
-    @classmethod
-    def determine_version(cls, exe):
-        output = Executable(exe)("--version", output=str, error=str)
-        match = re.search(r"HIP version (\S+)", output)
-        return match.group(1) if match else "None"
```
scheibelp, line 96: (question) is this method no longer reliable? Generally I would assume that it would be slightly more reliable than the default mechanism you added in `ROCmLibrary` (unless the output from this has changed).

scheibelp (commented):
Thanks for extracting out the common logic into `ROCmLibrary`!

Overall this change is somewhat outside of Spack's idioms for external detection, which are generally not prefix/layout-based (primarily because it means that custom layouts and installation prefixes can "defeat" this type of logic), but I see the convenience of that for this set of components (so this part is fine with me).

I do have a question though about a case where there was already exe-output-based version logic that got removed (was that done for consistency, or because it wasn't working for later versions?).

Also it looks like there's a check (`bootstrap-dev-rhel8`) that needs to be fixed that is not related to these changes. I think this will need to merge/rebase on https://github.com/spack/spack-packages/pull/4796 when that merges to pass CI.

repos/spack_repo/builtin/packages/hipify_clang/package.py
afzpatel: It's outputting the wrong ROCm version when I tested it on ROCm 7.1.1 :confused:
```
# /opt/rocm-7.1.1/bin/hipify-perl --version
HIP version 7.0.0
```

scheibelp (approved):

scheibelp:
Thanks!

PR #4795, opened 2026-05-11
GoPackage: allow packages to define ldflags

alecbcs:
It is common in the go ecosystem for packages to expect an ldflag to set the package version during the build. This PR adds a new optional property to `GoPackage` allowing maintainers to more easily add additional ldflags to the build without having to overwrite the `std_build_args`. Additionally this PR splits out `std_build_args` and `build_args` to follow more closely with other build systems like Cargo and CMake.

### DIFF
--- a/repos/spack_repo/builtin/build_systems/go.py
+++ b/repos/spack_repo/builtin/build_systems/go.py
@@ -85,6 +85,7 @@ class GoBuilder(BuilderWithDefaults):
 
     @property
     def std_ldflags(self) -> List[str]:
+        # Pass ldflags -s = --strip-all and -w = --no-warnings by default
         return ["-s", "-w"]
 
     @property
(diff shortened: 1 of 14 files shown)


alecbcs:
@wdconinc you don't necessarily need to review this PR if you feel unfamiliar with the build system changes. Just wanted to keep you in the loop since you'd asked about this earlier.

repos/spack_repo/builtin/build_systems/go.py
```diff
@@ -79,21 +84,34 @@ def build_directory(self):
...
+        return ["-s", "-w"]
+
+    @property
+    def ldflags(self) -> List[str]:
+        return []
+
+    @property
+    def std_build_args(self):
         """Arguments for ``go build``."""
         # Pass ldflags -s = --strip-all and -w = --no-warnings by default
```
wdconinc: Move this comment up to `std_ldflags` now.

alecbcs: Nice catch! Fixed

kwryankrattiger (approved):
LGTM

PR #4796, opened 2026-05-11
style check is now ruff-format vs. black

scheibelp:
For current spack checkout version (7e864787bd15a516314d86002ce1cbabde7cbbe9) `spack style --help` reports:

```
-s TOOL, --skip TOOL  specify tools to skip (choose from import, ruff-format, ruff-check, mypy)
```

PR #4797, opened 2026-05-11
Use ruff format not black

johnwparent:
Switches the bootstrap check to use `ruff format` instead of `black` as the tool argument


johnwparent:
Closing for #4796

PR #4799, opened 2026-05-11
bootstrap ci: run when ci changes are made

johnwparent:
We should run the bootstrap CI check when we change CI based components of Spack as they may alter the behavior of bootstrap CI runs in addition to python changes that may do the same.


alecbcs (approved):
Looks great! Thanks @johnwparent

PR #4800, opened 2026-05-11
py-onnxruntime: only run tests when self.run_tests

wdconinc:
This PR ensures that `py-onnxruntime` does not build its unit tests, unless we give the `--test` argument. Unit tests are enabled by default, see https://github.com/microsoft/onnxruntime/blob/19f5d413fc6f2c8ac06c4d1c65e4371fd7bdd8da/cmake/CMakeLists.txt#L97.

After the upgrade from spack 1.1.0 to 1.1.1 this started to fail in our internal builds since the FetchContent of googletest has a certificate error which was previously ignored (the joys of ssl rewrites at DOE NLs). After https://github.com/spack/spack/commit/c4ba02b73b this started to fail hard.


alecbcs (approved):

PR #4801, opened 2026-05-11
fix: permanent redirects

RobertMaaskant:
Fix some urls that have permanent redirects based on https://repology.org/repository/spack/problems


junghans (approved):

wdconinc (approved):
This looks fine to me; only http to https conversions.

PR #4802, opened 2026-05-11
py-pipcl: new package

V-Karch:
Surprisingly did not seem to need any other dependencies! Imported just fine when testing!


RikkiButler20 (approved):
LGTM

PR #4803, opened 2026-05-11
Gitlab CI: always upload generate artifacts

johnwparent:
Default behavior is to upload generate artifacts on success, uploading on failure is a useful debugging tool and saves devs having to go add it to the gitlab yaml every time they have a generate failure


johnwparent:
Not sure if there's some CI/Aws reason we don't want to do this, but I've found it useful enough times to want this by default

kwryankrattiger:
We should probably so have this for build jobs if we don't already. Do you mind double checking that.

johnwparent:
> We should probably so have this for build jobs if we don't already. Do you mind double checking that.

Np, on it

johnwparent:
We are already specifying `always` for build job artifacts: 
https://github.com/spack/spack/blob/develop/lib/spack/spack/ci/gitlab.py#L276

kwryankrattiger (approved):

PR #4804, opened 2026-05-11
Patch the fix to support 64 bit integers

yangliu2009:
The fix is neededfor mpi_f08: 
[https://github.com/open-mpi/ompi/pull/13159
](https://github.com/open-mpi/ompi/pull/13159
)


hppritcha (approved):

wdconinc:
The patch was subsequently reverted upstream; https://github.com/open-mpi/ompi/pull/13165. Is this the right solution to patch here?

hppritcha:
okay this brought back some memories.  The "real" fix is the much larger patch from https://github.com/open-mpi/ompi/pull/13168

hppritcha:
I doubt this is going to apply cleanly to currently released branches.  Are you trying to fix only the sessions problem reported in https://www.mail-archive.com/users@lists.open-mpi.org//msg35382.html ?

yangliu2009:
Yes, our user needs to build Dirac with openmpi.

hppritcha:
okay in that event use the patch you selected.  it should probably be applied to all builds out of the 5.0.x release stream.

yangliu2009:
Updated package.py to apply the patch to @5.0:5.0.10

wdconinc (approved):

hppritcha (approved):

alalazo:
@spackbot run pipeline

PR #4807, opened 2026-05-11
ci: switch to node@24 to test compatibility

alecbcs:
GitHub is set to update Actions to use Node v24 by default on June 8th. This PR forces the update to test our compatibility with Node v24 so that we won't be surprised if things start failing next month.

https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/

### DIFF
--- a/.github/workflows/audit.yaml
+++ b/.github/workflows/audit.yaml
@@ -7,6 +7,9 @@ on:
         required: true
         type: string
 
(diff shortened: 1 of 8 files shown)


alecbcs:
Looks like all of our actions are compatible with node@24. So nothing needs to be done here.

PR #4811, opened 2026-05-12
py-urllib3: add v2.7.0

adamjstewart:
All existing versions have high-severity CVEs, all should be deprecated and removed.


adamjstewart:
@alalazo @becker33 can one of you look into this concretization error? I'm able to concretize py-torchgeo locally on this branch, I'm not sure why it doesn't work in CI.

adamjstewart:
2 months of known high-severity CVE being installed by default by all users. Really could use help debugging this concretization issue so we can finally deprecate these vulnerable versions...

alecbcs:
I believe the issue is that `py-python-dateutil` depends on `py-setuptools-scm@:7` when `@2.7:` and with the modifications to this PR `py-urllib3` now only depends on `py-setuptools-scm@8:10`. 

I since both of these are build time constraints we should be able to add the following to `spack/spack`:
**etc/spack/defaults/base/concretizer.yaml**
```patch
 max_dupes:
   default: 1
   # Virtuals
   c: 2
   cxx: 2
   fortran: 1
   # Regular packages
   cmake: 2
   gmake: 2
   python: 2
   python-venv: 2
   py-cython: 2
   py-flit-core: 2
   py-pip: 2
   py-setuptools: 2
+ py-setuptools-scm: 2
   py-versioneer: 2
   py-wheel: 2
   xcb-proto: 2
   # Compilers
   gcc: 2
   llvm: 2
```

alecbcs:
@spackbot run pipeline

adamjstewart:
@spackbot run pipeline

alecbcs (approved):
Looks good to me. Thanks @adamjstewart

cmelone (approved):

PR #4813, opened 2026-05-12
Disabled Compiler Wrapper Staging

AcerP-py:
@omsai as requested :smile: 

For the full explanation behind this see https://spackpm.slack.com/archives/C08Q62S7XEX/p1775163670742649.

In brief it seems that some weird staging of the compiler wrapper keeps the `cc.sh` script from being updated when patched locally.


haampie:
You should use the validated source file, not a path outside of the stage dir.

It's also redundant after #4829.

PR #4820, opened 2026-05-12
ci: rename simple PRs for consistency

alecbcs:
This PR adds a new step in the triage job that will rename simple package version addition PRs to the standard format.

```
{pkg-name}: add v{version}

---

Add version 1.2.3 to py-pkg   ->    py-pkg: add v1.2.3
py-pkg: add new version       ->    py-pkg: add v1.2.3
```

I've been doing this manually for a long time now to help reviewers more easily scan the list of open PRs for those they feel comfortable reviewing. Most contributors over time have started using this common format, but periodically folks will still name their PRs the same as their commit and forget to include the package name. 

If a user renames their PR after the triage bot runs we respect their choice and will skip the rename step on all future runs.

### DIFF
--- a/.github/workflows/bin/spack-pr-renamer.py
+++ b/.github/workflows/bin/spack-pr-renamer.py
@@ -1,0 +1,138 @@
+#!/usr/bin/env python3
+
+# Copyright Spack Project Developers. See COPYRIGHT file for details.
+#
+# SPDX-License-Identifier: (Apache-2.0 OR MIT)
+
+import os
+import re
+
+import requests
+
+IS_PACKAGE_CHANGE = re.compile(r"repos/spack_repo/builtin/packages/([^/]+)/.*$")
+VERSION_LINE = re.compile(r'[\s]+version\([\s+\+]*"([^"]+)"[^\)]*\)')
+
+
+def main():
+    # Validate required environment variables
+    required_vars = ["GH_REPO", "GH_PR_NUMBER"]
+    missing_vars = [var for var in required_vars if var not in os.environ]
+    if missing_vars:
+        raise Exception(f"Missing required environment variables: {', '.join(missing_vars)}")
+
+    repository = os.environ["GH_REPO"]
+    pr_number = os.environ["GH_PR_NUMBER"]
+    token = os.environ.get("GH_TOKEN", "")
+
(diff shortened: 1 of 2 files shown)


.github/workflows/bin/spack-pr-renamer.py
```diff
@@ -0,0 +1,136 @@
...
+            new_title = f"{package_name}: new package"
+
+    # Handle new-version label (but not if also has new-variant)
+    elif "new-version" in labels and "new-variant" not in labels:
+        # Get PR diff to extract version changes
+        diff_resp = session.get(
+            f"{base_url}/pulls/{pr_number}.diff",
+            headers=headers,
+            timeout=30,
+        )
```
cmelone, line 100: can we use the patches returned in `pull_request_files` instead of requesting the diff?

https://docs.github.com/en/rest/pulls/pulls#list-pull-requests-files

.github/workflows/bin/spack-pr-renamer.py
```diff
@@ -0,0 +1,136 @@
...
+    # Only process single-file package changes
+    if len(files) != 1:
+        print("skipping multi-file PR")
+        return
+
+    file = files[0]
+    match = IS_PACKAGE_CHANGE.match(file["filename"])
+    if not match:
+        print("skipping non-package PR")
+        return
```
cmelone, line 80: should we maybe do these checks before the manually changed title check? so we don't end up checking the titles of uneligible PRs

.github/workflows/bin/spack-pr-renamer.py
```diff
@@ -0,0 +1,136 @@
...
+        if diff_resp.status_code != 200:
+            raise Exception(f"Failed to get PR diff [{diff_resp.status_code}]: {diff_resp.text}")
+
+        diff_text = diff_resp.text
+        deleted_versions = set(VERSION_LINE.findall(diff_text.replace("\n+", "\n ")))
+        added_versions = sorted(
+            set(VERSION_LINE.findall(diff_text.replace("\n-", "\n "))) - deleted_versions
+        )
+
+        if added_versions:
```
cmelone, line 110: can you run this script with:

```
GH_REPO="cmelone/testing1" GH_PR_NUMBER="2" python .github/workflows/bin/spack-pr-renamer.py
```

from https://github.com/cmelone/testing1/pull/2

`added_versions` is empty

PR #4822, opened 2026-05-12
hdf5: add v2.1.1

lrknox:


alalazo (changes requested):
There are issues to be fixed with packages. For instance:
- https://gitlab.spack.io/spack/spack-packages/-/jobs/22123983

PR #4829, opened 2026-05-13
compiler-wrapper: unvendor, v1.1.0, make GCC determinstic

haampie:
* Unvendor the `cc.sh` script, so we do proper releases
* Add v1.1.0 with `-frandom-seed` support to make GCC output deterministic.


kwryankrattiger (dismissed):
I agree this is the right call.

haampie:
Sigh,

> `g++: fatal error: cannot execute '/usr/libexec/gcc/aarch64-linux-gnu/13/collect2': execv: Argument list too long`

tgamblin:
I love this -- but please hold off on this so that we can get complete history in the compiler-wrapper repo.  There are a lot of contributors to that script and I don't want them to miss out on credit if it's going standalone.

I think that repo is worth some attention outside Spack -- our compiler wrapper is very much battle tested and the evolution / features, etc. that made it that way are IMO interesting.

haampie:
History of https://github.com/spack/compiler-wrapper/ is restored (from 3 separate repos, using git-filter-repo on spack and spack-packages). I have also deleted the `v1.1.0` and retagged `v1.0` since it wasn't used in Spack yet.

The `v1.1.0` release requires at a minimum https://github.com/spack/compiler-wrapper/pull/8 to fix "Argument list too long".

tgamblin:
Isn't `1.0` the version of the current compiler wrapper in Spack?  I guess if this is the same as the current version it's fine.  Labeling this 1.1 or something would make a clear break and would make it easier to run, e.g., `spack find compiler-wrapper@1.1:` to find only the ones from github.

```console
> spack find compiler-wrapper
-- darwin-sequoia-m1 / no compilers -----------------------------
compiler-wrapper@1.0  compiler-wrapper@1.0

-- darwin-tahoe-m1 / no compilers -------------------------------
compiler-wrapper@1.0  compiler-wrapper@1.0
```

haampie:
In short:

* v1.0 the same script as before
* v1.1.0 faster and makes GCC itself reproducible with `-frandom-seed`

tgamblin (approved):

PR #4831, opened 2026-05-13
plumed: remove libtorch_python

RMeli:
Remove `libtorch_python` from PLUMED. This is a bit of a hack, so I'm not convinced it is good idea.

### DIFF
--- a/repos/spack_repo/builtin/packages/plumed/package.py
+++ b/repos/spack_repo/builtin/packages/plumed/package.py
@@ -296,17 +296,18 @@ class Plumed(AutotoolsPackage):
         # Set flags to help with PyTorch
         if enable_libtorch:
             pytorch_path = Path(spec["py-torch"].package.cmake_prefix_paths[0]).parent.parent
+
             extra_ldflags.append(spec["py-torch"].libs.search_flags)
-            extra_libs.append(spec["py-torch"].libs.link_flags)
-            extra_ldflags.append(spec["python"].libs.search_flags)
-            extra_libs.append(spec["python"].libs.link_flags)
+
+            # Do not link libtorch_python
+            extra_libs.append(spec["py-torch"].libs.link_flags.replace("-ltorch_python ", ""))
+
             # Add include paths manually
             # Spack HeaderList.cpp_flags does not support include paths within include paths
             extra_cppflags.extend(
                 [
                     f"-I{pytorch_path / 'include'}",
                     f"-I{pytorch_path / 'include' / 'torch' / 'csrc' / 'api' / 'include'}",
-                    spec["python"].headers.include_flags,
                 ]
             )
         if enable_libmetatomic:
--- a/repos/spack_repo/builtin/packages/plumed/package.py
+++ b/repos/spack_repo/builtin/packages/plumed/package.py
@@ -303,7 +303,8 @@ class Plumed(AutotoolsPackage):
             # Here we manually remove the torch_python library to avoid linking against it
             # Linking against torch_python requires linking against Python as well,
             # which is not needed and can cause issues with some workflows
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/plumed/package.py
```diff
@@ -296,17 +296,18 @@ def configure_args(self):
         # Set flags to help with PyTorch
         if enable_libtorch:
             pytorch_path = Path(spec["py-torch"].package.cmake_prefix_paths[0]).parent.parent
+
             extra_ldflags.append(spec["py-torch"].libs.search_flags)
-            extra_libs.append(spec["py-torch"].libs.link_flags)
-            extra_ldflags.append(spec["python"].libs.search_flags)
-            extra_libs.append(spec["python"].libs.link_flags)
```
msimberg, line 302: I think I don't understand the motivation. This seems to be removing both `-ltorch_python` specifically, but also removing link flags/search flags from `py-torch` and `python` altogether? If `py-torch` link flags don't get added in the first place, why do they need to get removed?

RMeli: `py-torch` link flags are added below (line 303), but manually excluding `-ltroch_python` (so that Python's link flags are no longer needed).

msimberg: Ah, missed that, thanks (I read it as only removing the flag, not adding all but the `-ltorch_python` flag).

In any case, I don't really have much to comment on the solution itself. If it's needed, it's needed. If you're able to, perhaps add a slightly more descriptive comment or link to an issue about why this is done?

albestro: I second Mikael's position.

Just one additional question: ignoring the cleanliness argument for a moment, how does it complain if no symbol from `libtorch_python` are used? I mean, is it a symbol conflict? because otherwise isn't it that more libraries than needed are currently passed to the linker?

RMeli: Maybe a bit more context is useful to understand the issue. 

PLUMED only requires the C++ distribution of PyTorch, and linking to Python has proven problematic with the appearance of GIL-related issues when mixed with external Python installations (for example PLUMED+LAMMPS provided by Spack, while a ML potential installed in a venv). 

(FYI, this is related to uenv workflows at CSCS: [Using LAMMPS with MLIPs](https://docs.cscs.ch/software/sciapps/lammps/#lammps-ml-iap-using-lammps-with-machine-learning-interatomic-potentials).)

Before this PR, `-ltorch_python` was being used (since it is included in `spec["py-torch"].libs.link_flags`, see https://github.com/spack/spack-packages/pull/1990), which requires to also link against Python (I don't recall the exact error, I can probably reproduce it). With this PR, the idea is to avoid the needless `torch_python`, thus decoupling PLUMED from Python and circumventing the aforementioned GIL-related issues.

I think the correct solution would be to have a `libtorch` package (C++ distribution of PyTorch), but it is something I only explored briefly a while ago and was not confident to push on Spack given how big and widespread PyTorch is.

msimberg: Thanks! So maybe a short comment at least saying that plumed doesn't need `torch_python` and "can cause problems" or something equally vague would still be good to add, so it's not _completely_ without context?

RMeli: Thanks, I added a comment to explain why `torch_python` is manually removed.

msimberg (dismissed):

RMeli:
@spackbot fix style

msimberg (approved):

PR #4834, opened 2026-05-13
github actions: update triage bot to use client id

alecbcs:
GitHub now recommends using an app's client ID instead of the app ID to generate short lived tokens. This PR swaps to using the new client ID method. After merging we should remove the app ID from the internally configured secrets.

https://github.com/actions/create-github-app-token/pull/353


kwryankrattiger (approved):
LGTM

alecbcs:
Force merged as this PR does not affect failing GitLab CI

PR #4839, opened 2026-05-13
libedit: Fix definition of version 3.1-20251016

RemiLacroix-IDRIS:


hainest (approved):
Confirmed `spack install`.

PR #4844, opened 2026-05-14
fann: add git attribute and master version

wdconinc:
This PR adds to `fann` the git attribute and a master version. No changes to dependencies since the last released version ([diff](https://github.com/libfann/fann/compare/2.2.0...master)).


RikkiButler20 (approved):
LGTM

PR #4848, opened 2026-05-14
btop: add v1.4.7, fix `+gpu` build

alalazo:
Modifications:
- Add v1.4.7 and deprecate old versions
- Add a patch to define _GNU_SOURCE for the bundled `intel_gpu_top` C code. See https://github.com/aristocratos/btop/issues/1565
- Bump minimum version of required GCC (`<semaphore>` was added in GCC 11)


alalazo:
@spackbot fix style

rbberger (approved):

PR #4177, opened 2026-04-08
axom: Allow axom ~fortran and conduit +fortran combo

tdrwenski:
Allow axom ~fortran and conduit +fortran combo as this was useful in CI tests and does not seem harmful to allow.

PR #4853, opened 2026-05-14
axom: enable python with conduit dependency

kshea21:
My project needs to build `conduit+python` and `axom`. Currently, `axom` requires `conduit~python` as a dependency, making my conduits incompatible and package unable to build. I'm hoping that we can change the `axom` package to allow `conduit` to build with `python` if it's enabled. 

Also added what we believe to be a fix for `depends_on("conduit+python", when="+devtools")` to be in the `with when("+conduit")` so that `conduit` only need to be built as a dependency when the `conduit` variant is enabled. 

Thanks!


white238 (approved):
Thanks @kshea21 ! @bmhan12 can you look at this too?

repos/spack_repo/builtin/packages/axom/package.py
```diff
@@ -187,10 +181,11 @@ class Axom(CachedCMakePackage, CudaPackage, ROCmPackage):
     # Libraries
     # Forward variants to Conduit
     with when("+conduit"):
-        for _var in ["hdf5", "mpi", "python"]:
+        for _var in ["fortran", "hdf5", "mpi"]:
             depends_on("conduit+{0}".format(_var), when="+{0}".format(_var))
             depends_on("conduit~{0}".format(_var), when="~{0}".format(_var))
-        depends_on("conduit+fortran", when="+fortran")
+        depends_on("conduit+python", when="+devtools")
+        depends_on("conduit+python", when="+python")
```
bmhan12: @kshea21 I think this is close!

There are merge conflicts, so this branch's changes is comparing against an older commit of spack-packages in the diff. Here's what I see on develop:
https://github.com/spack/spack-packages/blob/e2e8f8670ed0fcb2423ca6fa9b84e6e148478fc7/repos/spack_repo/builtin/packages/axom/package.py#L187-L193

@tdrwenski came up with a similar fix to what I think you were looking for related to `axom~fortran` building with `conduit+fortran`: #4177

Assuming I understand correctly, that you need `axom~python` (default variant is ~python) with `conduit+python`, could you try:

```suggestion
        for _var in ["hdf5", "mpi"]:
            depends_on("conduit+{0}".format(_var), when="+{0}".format(_var))
            depends_on("conduit~{0}".format(_var), when="~{0}".format(_var))
        depends_on("conduit+fortran", when="+fortran")
```

Notably, `axom+python` depending on `conduit+python` is already defined elsewhere: https://github.com/spack/spack-packages/blob/e2e8f8670ed0fcb2423ca6fa9b84e6e148478fc7/repos/spack_repo/builtin/packages/axom/package.py#L288)

kshea21: Ah I see, I was pinned to an out of date builtin. Updating and making sure that python is removed from the conduit loop works, thanks!

bmhan12 (approved):
Looks good to me!

becker33 (approved):

PR #4855, opened 2026-05-15
axom: run install tests when +example

keithhealy:
Install tests try to run even if ~examples is specified. The tests then fail because the examples have not been installed.
Modified the testInstall functions to exit if ~examples is specified.


repos/spack_repo/builtin/packages/axom/package.py
```diff
@@ -835,6 +835,8 @@ def build_test(self):
     @run_after("install")
```
white238: You could also do this with:

```
@run_after("install", when="+examples")
```

RikkiButler20 (commented):
LGTM. @white238  any more feedback?

white238 (approved):

RikkiButler20 (approved):

PR #4862, opened 2026-05-15
shamrock: use CMAKE_INSTALL_PYTHONDIR to install python bindings

tdavidcl:
This PR replace the manual installation of the python bindings by one directly handled by Cmake. 
This should yield equivalent results. 
The way to perform the installation was done using the same snippet as ADIOS2 (repos/spack_repo/builtin/packages/adios2/package.py)

### DIFF
--- a/repos/spack_repo/builtin/packages/shamrock/package.py
+++ b/repos/spack_repo/builtin/packages/shamrock/package.py
@@ -2,8 +2,6 @@
 #
 # SPDX-License-Identifier: (Apache-2.0 OR MIT)
 
-import glob
-
 from spack_repo.builtin.build_systems.cmake import CMakePackage, generator
 
 from spack.package import *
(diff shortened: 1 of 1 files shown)


hainest (approved):
LGTM.

I would encourage changing the configuration variable `CMAKE_INSTALL_PYTHONDIR` to `SHAMROCK_INSTALL_PYTHONDIR` since variables startinging with `CMAKE` are technically reserved by CMake for future use. I understand it may not be possible since it might break existing user code.

tdavidcl:
> LGTM.

Thanks! 

> I would encourage changing the configuration variable `CMAKE_INSTALL_PYTHONDIR` to `SHAMROCK_INSTALL_PYTHONDIR` since variables startinging with `CMAKE` are technically reserved by CMake for future use. I understand it may not be possible since it might break existing user code.

Actually we had little choice on this one since it is the variable name expected by scikit-build for python packaging, hence the choice 😅.

PR #4863, opened 2026-05-16
libssh2: add maintainer

CodingYayaToure:
## What this PR does

- Adds `CodingYayaToure` as maintainer for `libssh2`

`libssh2` is a foundational HPC dependency (used by OpenMPI, curl, git)
with no declared maintainer. This PR takes ownership of the package.


kshea21 (commented):
The title of your PR suggests that you'd be adding libssh2 versions 1.10.0 and 1.11.0, but they're not included here. Did you mean to add them?

becker33:
I believe there are substantial CVEs in the older versions, so it may make more sense to update the PR title than to add the older versions.

bernhardkaindl (approved):

PR #4866, opened 2026-05-16
tblite v0.6.0, dftd4 v4.2.0 and simple-dftd3 v1.4.0

Growl1234:


awvwgk (approved):
Thanks, looks good to me.

repos/spack_repo/builtin/packages/dftd4/package.py
```diff
@@ -23,6 +23,9 @@ class Dftd4(MesonPackage, CMakePackage):
     build_system("cmake", "meson", default="meson")
 
     version("main", branch="main")
+    version("4.2.0", sha256="467e024071510ad82b862c66c383c2ebc164fc1140e15dfc79f48d2f999fd184")
+    version("4.1.1", sha256="c8e6388d7d7d748dbcf91117f35aa50108492d4fd2266d60782cf85a16651887")
+    version("4.1.0", sha256="344aafa9e994a08186c95bf4421d70aeb493fd9f8038726fc2782dd3f892c3a9")
```
hainest, line 28: The SHA for 4.1.0 doesn't match what I see. Could you recheck?

hainest (changes requested):
The one SHA for dftd4 doesn't match, but all the others do.

repos/spack_repo/builtin/packages/dftd4/package.py
Growl1234: I have rechecked and the sha256 of `dftd4-4.1.0.tar.xz` (not `dftd4-4.1.0-source.tar.xz`) is right. It also passes the "Verify Checksum" test.

hainest: Now I'm seeing the same one. That was weird.

hainest (approved):

PR #4867, opened 2026-05-16
paraview: Set VTK_PYTHON_OPTIONAL_LINK=OFF by default for OpenFOAM

furstj:
VTK_PYTHON_OPTIONAL_LINK=OFF is required by OpenFOAM Catalyst plugin in order to avoid

undefined symbol: PyExc_ValueError, see e.g.

https://discourse.paraview.org/t/undefined-symbol-pyexc-valueerror/5494


vicentebolea:
Looks good, VTK_PYTHON_OPTIONAL_LINK no needed here. Please only add this option when the python variant is on.

furstj (commented):
VTK_PYTHON_OPTIONAL_LINK=OFF was moved to +python section

vicentebolea (approved):

kwryankrattiger (approved):
This should be ok in Spack

PR #4873, opened 2026-05-16
m4: add maintainer

CodingYayaToure:
## What this PR does

- Adds `CodingYayaToure` as maintainer for `m4`

`m4` is a foundational GNU build tool dependency with no declared
maintainer. Package is already at latest version (1.4.21).


kshea21 (approved):
Thanks for signing up to be a maintainer!

PR #4696, opened 2026-05-06
py-pyrometheus: new package

sbryngelson:
Adds a package recipe for `py-pyrometheus@1.0.7`, a Python code generator for combustion thermochemistry based on Cantera.

Validated locally with:

- `ruff format --check repos/spack_repo/builtin/packages/py_pyrometheus/package.py`
- `ruff check repos/spack_repo/builtin/packages/py_pyrometheus/package.py`
- `spack audit packages py-pyrometheus`
- `spack info py-pyrometheus`
- `spack spec py-pyrometheus`
- import/code-generation smoke test with `cantera@3.1.0+python`

PR #4875, opened 2026-05-17
py-pyrometheus: add main branch version

sbryngelson:
Adds a `main` branch version to `py-pyrometheus` so downstream CI can install pyrometheus from the upstream `main` branch via `spack install py-pyrometheus@main` without pinning to a specific release tag.

Motivation: pyrometheus is adding a Spack CI job that installs the current `HEAD` (via `spack develop`). Without a `@main` spec, the workflow has to pin to whatever the recipe's newest release tag is and gets stale every release. A `version("main", branch="main")` entry tracks `main` automatically.

Validated with:

- `ruff format --check repos/spack_repo/builtin/packages/py_pyrometheus/package.py`
- `ruff check repos/spack_repo/builtin/packages/py_pyrometheus/package.py`
- `spack audit packages py-pyrometheus`
- `spack spec py-pyrometheus@main`

Follow-up to #4696.

### DIFF
--- a/repos/spack_repo/builtin/packages/py_pyrometheus/package.py
+++ b/repos/spack_repo/builtin/packages/py_pyrometheus/package.py
@@ -18,6 +18,7 @@ class PyPyrometheus(PythonPackage):
 
     license("MIT")
 
+    version("main", branch="main")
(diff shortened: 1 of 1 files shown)


RikkiButler20 (approved):
LGTM

PR #4877, opened 2026-05-17
libxml2: add versions 2.15.2 and 2.15.3

CodingYayaToure:
## What this PR does

- Adds version 2.15.3 of `libxml2` (latest in the 2.15.x series)

## Links

- Release notes: https://gitlab.gnome.org/GNOME/libxml2/-/blob/v2.15.3/NEWS?ref_type=tags
- Diff with previous version in Spack (2.15.1 → 2.15.3): https://github.com/GNOME/libxml2/compare/v2.15.1...v2.15.3
- Spack binary cache: https://cache.spack.io/?package=libxml2

### DIFF
--- a/repos/spack_repo/builtin/packages/libxml2/package.py
+++ b/repos/spack_repo/builtin/packages/libxml2/package.py
@@ -32,6 +32,8 @@ class Libxml2(AutotoolsPackage, CMakePackage, NMakePackage):
 
     license("MIT")
 
+    version("2.15.3", sha256="78262a6e7ac170d6528ebfe2efccdf220191a5af6a6cd61ea4a9a9a5042c7a07")
+    version("2.15.2", sha256="c8b9bc81f8b590c33af8cc6c336dbff2f53409973588a351c95f1c621b13d09d")
     version("2.15.1", sha256="c008bac08fd5c7b4a87f7b8a71f283fa581d80d80ff8d2efd3b26224c39bc54c")
     version("2.13.9", sha256="a2c9ae7b770da34860050c309f903221c67830c86e4a7e760692b803df95143a")
     version("2.13.5", sha256="74fc163217a3964257d3be39af943e08861263c4231f9ef5b496b6f6d4c7b2b6")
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/libxml2/package.py
```diff
@@ -32,6 +32,8 @@ def url_for_version(self, version):
 
     license("MIT")
 
+    version("2.15.3", sha256="78262a6e7ac170d6528ebfe2efccdf220191a5af6a6cd61ea4a9a9a5042c7a07")
+    version("2.15.2", sha256="c8b9bc81f8b590c33af8cc6c336dbff2f53409973588a351c95f1c621b13d09d")
```
wdconinc: This is also a case where we may not need to add the .2 version, since the .3 is a bugfix over that version. Often we follow the following approach:
- include only latest patch version for each major.minor sequence,
- include latest minor.patch version for each major (e.g. if v2 is released, then we'd probably make sure v1.23.45 is included if it is not already, but not also add v1.22 and so on).

```suggestion
```

wdconinc (commented):
The sha256 checksum are validated automatically in CI, so no need to mention that they are correct (the assumption is that they are valid).

What helps the reviewer is to include in the PR description is a link to the new version [release notes](https://gitlab.gnome.org/GNOME/libxml2/-/releases/v2.15.3) (if they are useful; here they are not, so a better link is for example [NEWS](https://gitlab.gnome.org/GNOME/libxml2/-/blob/v2.15.3/NEWS?ref_type=tags)), and a link to the [diff](https://github.com/GNOME/libxml2/compare/v2.15.1...v2.15.3) with the last version in spack.

Often it is also useful to mention if the package is built in CI with this latest version. You can tell this by looking at e.g. https://cache.spack.io/?package=libxml2 in this case (or just search on cache.spack.io itself). Sometimes older versions are pinned but often it's the newest version.

repos/spack_repo/builtin/packages/libxml2/package.py
CodingYayaToure: Thanks for the review and the guidelines! I've already applied the suggested 
change in the latest commit — version 2.15.2 has been removed, keeping only 
2.15.3 as the latest patch version in the 2.15.x series.

I've also updated the PR description with links to the NEWS file, the diff, 
and the Spack binary cache.

wdconinc (approved):
Looks good now. Waiting for maintainer review.

AlexanderRichert-NOAA (approved):

PR #4879, opened 2026-05-17
curl: add versions 8.19.0 and 8.20.0

CodingYayaToure:
## What this PR does

- Adds versions 8.19.0 and 8.20.0 of `curl`

Spack was missing the two latest releases of `curl`.
sha256 verified against official tarballs from curl.se.


tgamblin (approved):
Thanks!

PR #4882, opened 2026-05-17
Qt6: add version 6.11.1 (qt-base, qt-tools, qt-...)

melven:
Adds the most recent version of the Qt6 libraries currently installable with spack.


wdconinc (approved):
Thanks! This looks good to me.

PR #4883, opened 2026-05-17
qt-3d: add new package for Qt3D (Qt6)

melven:
Adds a package for the Qt3d in Qt6.

### DIFF
--- a/repos/spack_repo/builtin/packages/qt_3d/package.py
+++ b/repos/spack_repo/builtin/packages/qt_3d/package.py
@@ -29,7 +29,8 @@ class Qt3d(QtPackage):
 
     depends_on("gl", type=("build", "link"))
     depends_on("glu", type=("build", "link"))
-    depends_on("assimp", type=("build", "link"))
+    # unfortunately, the build process does not seem to pick up this external assimp
+    #depends_on("assimp", type=("build", "link"))
 
     for _v in QtBase.versions:
         v = str(_v)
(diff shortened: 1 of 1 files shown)


melven:
Requires https://github.com/spack/spack-packages/pull/4882 because I only tested the newest version.

repos/spack_repo/builtin/packages/qt_3d/package.py
```diff
@@ -0,0 +1,36 @@
...
+    provides a generic framework for near-realtime simulations beyond
+    rendering."""
+
+    url = QtPackage.get_url(__qualname__)
+    git = QtPackage.get_git(__qualname__)
+    list_url = QtPackage.get_list_url(__qualname__)
+
+    license("LGPL-3.0-only OR GPL-2.0-only OR GPL-3.0-only", checked_by="melven")
+
+    version("6.11.1", commit="1f4c3a7548201bcad21a273a49060c96ad9ff3a9", submodules=True)
```
wdconinc, line 23: Does this need to use the commit + submodules even when you use the external assimp? The disadvantage of using the commit-based versioning is that `spack checksum qt3d` won't pick up any new versions for adding.

melven: Ah sorry.

It does not pick up the external assimp, so I removed the assimp dependency instead.
I tried some workarounds by setting CMake flags / qt features but didn't get it to work...

melven:
@spackbot fix style

repos/spack_repo/builtin/packages/qt_3d/package.py
melven: Hi @wdconinc 

this would be ready for merging from my side, so we can continue with the PySide6 PR.

bernhardkaindl (approved):

repos/spack_repo/builtin/packages/qt_3d/package.py
bernhardkaindl: Hi @melven, with this comment answered, I think it is ok to consider this new package ready to merge.

PR #4886, opened 2026-05-17
fix some urls with permanent redirects

RobertMaaskant:
Fix some urls that have permanent redirects based on https://repology.org/repository/spack/problems


RikkiButler20 (approved):
Great catch, thanks for the fix!

PR #4895, opened 2026-05-18
rust bootstrap: Windows support

johnwparent:
Adds support for "building" rust bootstrap on Windows.

Rust bootstrap does not vendor an equivalent install.sh on Windows as it does posix so I wrote a very simple equivalent in powershell that grabs the rust bootstrap manifest and installs the components listed, the same way install.sh does. 

This is a simplified version, it does none of the error checking the general install.sh does simply because we can make assumptions about a lot of the error cases the install.sh script has to handle by virtue of the fact we're operating within Spack

The powershell script can also be copied to boostrap rust install prefix or replaced with an equivalent python method, powershell was a somewhat arbitrary choice on my part (might've been too focused on "need shell script")

Step 1 of (hopefully) 2 for Rust on Windows


johnwparent:
@spackbot fix style

alecbcs:
@johnwparent realizing now that I'd missed this earlier this spring. If you're able to update this with the last couple of releases of rust I think we're likely good to merge.

PR #4896, opened 2026-05-19
Zsh: add build-time dependency on autoconf and add CFLAGS for more recent LLVM

taliaferro:
This PR fixes two issues with zsh.
1) It seems zsh has a build-time dependency on autoconf. (It looks like it's re-running it later in the process and not finding it when it needs it)

2) Add two compiler flags (`-Wno-implicit-int` and `-Wno-implicit-function-declaration`) when building with newer versions of LLVM. 

Without the latter, the build would complete successfully, but the resulting zsh would lock up and become unresponsive to signals the first time it encounters process substitution. (this issue can be demonstrated by running `timeout 3 zsh -c '$(echo echo success)' || echo failure`.)

### DIFF
--- a/repos/spack_repo/builtin/packages/zsh/package.py
+++ b/repos/spack_repo/builtin/packages/zsh/package.py
@@ -49,7 +49,11 @@ class Zsh(AutotoolsPackage):
     patch("pointer-types.patch", when="@5.6.2:")
 
     def flag_handler(self, name, flags):
-        if name=="cflags" and self.spec.satisfies('%llvm@10:') or self.spec.satisfies("%apple-clang@12:"):
+        if (
+            name == "cflags"
+            and self.spec.satisfies("%llvm@10:")
+            or self.spec.satisfies("%apple-clang@12:")
+        ):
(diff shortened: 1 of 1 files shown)


taliaferro:
@spackbot fix style, if you'd be so kind

repos/spack_repo/builtin/packages/zsh/package.py
```diff
@@ -42,11 +42,22 @@ class Zsh(AutotoolsPackage):
...
     conflicts("+lmod", when="~etcdir", msg="local etc required to setup env for lmod")
 
     patch("pointer-types.patch", when="@5.6.2:")
 
+    def flag_handler(self, name, flags):
+        if (
+            name == "cflags"
+            and self.spec.satisfies("%llvm@10:")
+            or self.spec.satisfies("%apple-clang@12:")
+        ):
```
alalazo, line 56: I think there's an issue with operator precedence here:
```
>>> False and False or True
True
```
(`%apple-clang` is always `True` regardless of `name`)

alalazo (changes requested):

PR #4899, opened 2026-05-19
parallel-netcdf: fix building with ~fortran

melven:
The package.py was using the Fortran compiler spec even if Fortran was disabled, leading to an error when compiling without Fortran.

### DIFF
--- a/repos/spack_repo/builtin/packages/parallel_netcdf/package.py
+++ b/repos/spack_repo/builtin/packages/parallel_netcdf/package.py
@@ -129,13 +129,15 @@ class ParallelNetcdf(AutotoolsPackage):
         if self.spec.satisfies("+pic"):
             flags["CFLAGS"].append(self.compiler.cc_pic_flag)
             flags["CXXFLAGS"].append(self.compiler.cxx_pic_flag)
-            flags["FFLAGS"].append(self.compiler.f77_pic_flag)
-            flags["FCFLAGS"].append(self.compiler.fc_pic_flag)
+            if self.spec.satisfies("+fortran"):
+                flags["FFLAGS"].append(self.compiler.f77_pic_flag)
+                flags["FCFLAGS"].append(self.compiler.fc_pic_flag)
 
         # https://github.com/Parallel-NetCDF/PnetCDF/issues/61
         if self.spec.satisfies("@:1.12.1%gcc@10:"):
-            flags["FFLAGS"].append("-fallow-argument-mismatch")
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/parallel_netcdf/package.py
```diff
@@ -129,12 +129,14 @@ def configure_args(self):
...
+            if self.spec.satisfies("+fortran"):
+                flags["FFLAGS"].append(self.compiler.f77_pic_flag)
+                flags["FCFLAGS"].append(self.compiler.fc_pic_flag)
 
         # https://github.com/Parallel-NetCDF/PnetCDF/issues/61
         if self.spec.satisfies("@:1.12.1%gcc@10:"):
-            flags["FFLAGS"].append("-fallow-argument-mismatch")
+            if self.spec.satisfies("+fortran"):
+                flags["FFLAGS"].append("-fallow-argument-mismatch")
             flags["FCFLAGS"].append("-fallow-argument-mismatch")
```
alalazo: Why this is left out of the `+fortran` conditional?

alalazo (changes requested):

repos/spack_repo/builtin/packages/parallel_netcdf/package.py
melven: Sorry, I missed that one while cleaning up my changes.

Nice catch, fixed now.

repos/spack_repo/builtin/packages/parallel_netcdf/package.py
alalazo: Do we unconditionally depend on fortran, even though we have a `fortran` variant?

melven: Compiles without error if I remove the dependency.

I assume that is a sufficient check? The library does also not depend on libgfortran (checked with ldd).

alalazo (approved):

bernhardkaindl (approved):
Rebased onto current develop throuth the Github rebase button on the web to see if re-running GitLab CI again (was failing due to some internal error) works this time.

bernhardkaindl:
Hi @melven, GitLab CI failed to run a job with: Unable to run pipeline: Undefined error.

As I didn't see this problem recently again, I rebased onto current develop using the Github rebase button on the web to update the PR and trigger a re-run of the pipeline.

@melven, if you see another general GitLab CI failure, please don't hestitate to ping me or another maintainer (though I may be away for some weeks).

PR #4907, opened 2026-05-20
`py-accessible-pygments`: Use hatchling for v0.0.5

tgamblin:
The build system for `py-accessible-pygments` changed to hatchling in 0.0.5, so use that instead of setuptools.

This is needed by py-furo which is needed to build the spack docs.


alecbcs (approved):
Looks good to me

alecbcs:
@spackbot run pipeline

PR #4908, opened 2026-05-20
py-idna: deprecate versions with CVEs

adamjstewart:
Only moderate, but no packages require ancient idna, so probably safe to deprecate and remove.


RikkiButler20 (approved):
LGTM

PR #4916, opened 2026-05-20
cupy: add v14.0

tehrengruber:
Adds support for CuPy 14.0.x and backfills several missing 12.x / 13.x point releases.                                                                                                                                             
                                                                                                                                                                                                                             
  - New package py-cuda-pathfinder.  Required runtime dep for CuPy 14.
  - Enable `submodules=True` (CuPy external dependencies in `third_party` via git submodules).                                                                                                                                                              
  - New versions: 14.0.1, 14.0.0, 13.6.0, 13.5.0, 13.4.1, 13.0.0, 12.3.0, 12.2.0. Removed pre-release 14.0.0a1.                                                                                                                    
  - Cython: fix constraints
  - Fix NumPy requirement for Cupy v14 (requires v2)                                                                                                                                                            
  - cuDNN dependency has been removed in v14                                                                                                                                             
  - New dep: py-cuda-pathfinder for @14: +cuda.                                                                                                                                                                                    
  - ROCm 6.3 patch: add upstream cupy#9022 to fix @13.4 ^rocm-core@6.3.                                                                                                                                                            
  - Add rccl to the ROCm include-path mapping so HIP builds find <rccl/...> headers.


tehrengruber:
@spackbot fix style

repos/spack_repo/builtin/packages/py_cuda_pathfinder/package.py
```diff
@@ -0,0 +1,42 @@
...
+    """Pathfinder for CUDA components"""
+
+    homepage = "https://nvidia.github.io/cuda-python/cuda-pathfinder"
+    url = "https://files.pythonhosted.org/packages/py3/c/cuda-pathfinder/cuda_pathfinder-1.3.3-py3-none-any.whl"
+    list_url = "https://pypi.org/project/cuda-pathfinder"
+
+    license("Apache-2.0")
+
+    version("1.5.4", sha256="9563d3175ce1828531acf4b94e1c1c7d67208c347ca002493e2654878b26f4b7")
+    version("1.5.3", sha256="dff021123aedbb4117cc7ec81717bbfe198fb4e8b5f1ee57e0e084fec5c8577d")
```
msimberg, line 20: You might not need to add all of these, but no objections if you think it's useful to have them here.

repos/spack_repo/builtin/packages/py_cupy/package.py
```diff
@@ -21,14 +21,21 @@ class PyCupy(PythonPackage, CudaPackage, ROCmPackage):
     homepage = "https://cupy.dev/"
     pypi = "cupy/cupy-8.0.0.tar.gz"
     git = "https://github.com/cupy/cupy.git"
+    submodules = True
 
-    version("main", branch="main")
```
msimberg, line 25: Why was this removed?

repos/spack_repo/builtin/packages/py_cupy/package.py
```diff
@@ -111,6 +118,12 @@ class PyCupy(PythonPackage, CudaPackage, ROCmPackage):
     conflicts("+cuda +rocm")
     conflicts("+cuda cuda_arch=none")
 
+    patch(
+        "https://patch-diff.githubusercontent.com/raw/cupy/cupy/pull/9022.patch?full_index=1",
+        sha256="bc96ea317748e42f7651d9f3b97f8db224081f627bcfcb6cc48c2ca139143441",
+        when="@13.4 ^rocm-core@6.3",
+    )
```
msimberg, line 129: Can you add a (brief) comment why this is needed?

msimberg (commented):
Looks generally good to me, but no deep knowledge from me on cupy itself.

repos/spack_repo/builtin/packages/py_cupy/package.py
```diff
@@ -21,14 +21,21 @@ class PyCupy(PythonPackage, CudaPackage, ROCmPackage):
     homepage = "https://cupy.dev/"
     pypi = "cupy/cupy-8.0.0.tar.gz"
     git = "https://github.com/cupy/cupy.git"
+    submodules = True
```
msimberg, line 27: It's unclear to me why this is needed. Generally packages should not use bundled dependencies. Is this required?

repos/spack_repo/builtin/packages/py_cupy/package.py
tehrengruber: I noticed this is actually not needed. When `git` is there you can just do `cupy@main`.

repos/spack_repo/builtin/packages/py_cupy/package.py
tehrengruber: Good point, haven't thought about this. The packages are already bundled in the tarballs, `submodules=True` just makes this work the same way when installing a version from git. I've added a TODO to model these dependencies explicitly.

repos/spack_repo/builtin/packages/py_cupy/package.py
tehrengruber: I did not add a comment since it would only repeat what is already in the code here and in the patch: For cupy version 13.4.x and rocm 6.3 this "fixes `/opt/rocm/include/thrust/version.h:34:10: fatal error: cuda/version: No such file or directory", without adding any additional information. Does that make sense?

repos/spack_repo/builtin/packages/py_cuda_pathfinder/package.py
tehrengruber: In the beginning I only had one version, but I like the `spack checksum` workflow and it might be useful, so I just put all here.

repos/spack_repo/builtin/packages/py_cupy/package.py
msimberg: Are you sure about this? `cupy@git.main` does work, but I've never seen `cupy@main` work. Is that a new feature? If it is, I'd probably keep this around for those who are not on the latest version of spack.

tehrengruber:
@spackbot fix style

repos/spack_repo/builtin/packages/py_cupy/package.py
tehrengruber: I messed up my spack repo path. Ignore my previous statement. Reverted.

msimberg (approved):
Thanks, looks good to me!

PR #4919, opened 2026-05-20
esmf: Make FindESMF.cmake findable via CMAKE_MODULE_PATH

### LINKED ISSUE #5045: ESMF: Make FindESMF.cmake findable by dependents?
### Summary

Is there some way that FindESMF.cmake can be exposed to ESMF-using, CMake-based codes without those codes having to set CMAKE_MODULE_PATH on their own? I think in an ideal world, ESMF would provide CMake target information exposed via CMAKE_PREFIX_PATH, but some standardized way of doing this through the Spack recipe would maybe be the next best thing.

### Rationale

_No response_

### Description

_No response_

### Additional information

_No response_

### General information

theurich:
This PR appends to the `CMAKE_MODULE_PATH` to allow dependent packages to find FindESMF.cmake. This follows suggestion by @johnwparent under https://github.com/spack/spack-packages/issues/5045#issuecomment-4479922985

Also cover ESMX and ESMPy in the ESMF package description. Something I noticed was missing.

Resolves https://github.com/spack/spack-packages/issues/5045#issuecomment-4479922985

### DIFF
--- a/repos/spack_repo/builtin/packages/esmf/package.py
+++ b/repos/spack_repo/builtin/packages/esmf/package.py
@@ -178,6 +178,8 @@ class Esmf(MakefilePackage, PythonExtension):
 
     def setup_run_environment(self, env: EnvironmentModifications) -> None:
         env.set("ESMFMKFILE", os.path.join(self.prefix.lib, "esmf.mk"))
+        if self.spec.satisfies("@8:"):
+            env.append_path("CMAKE_MODULE_PATH", self.prefix.cmake)
 
 
 class PythonPipBuilder(python.PythonPipBuilder):
--- a/repos/spack_repo/builtin/packages/esmf/package.py
+++ b/repos/spack_repo/builtin/packages/esmf/package.py
@@ -469,7 +469,7 @@ class MakefileBuilder(makefile.MakefileBuilder):
     ) -> None:
         env.set("ESMFMKFILE", os.path.join(self.prefix.lib, "esmf.mk"))
         if self.spec.satisfies("@8:"):
-            env.append_path("CMAKE_PREFIX_PATH", self.prefix.cmake)
+            env.append_path("CMAKE_MODULE_PATH", self.prefix.cmake)
 
     def install(self, pkg, spec, prefix):
         make("install")
--- a/repos/spack_repo/builtin/packages/esmf/package.py
+++ b/repos/spack_repo/builtin/packages/esmf/package.py
@@ -18,9 +18,24 @@ class Esmf(MakefilePackage, PythonExtension):
     related Earth science applications. The ESMF defines an architecture for
     composing complex, coupled modeling systems and includes data structures
     and utilities for developing individual models.
+
     The National Unified Operational Prediction Capability (NUOPC) Layer
     defines a common model architecture to support interoperable ESMF components.
-    The NUOPC Layer is included with the ESMF package."""
+    The NUOPC Layer is included with the ESMF package.
+
+    ESMX (Earth System Modeling eXecutable) extends this infrastructure by
+    providing a unified, runtime executable layer. It simplifies the deployment
+    of NUOPC-compliant components by allowing users to configure, drive, and
+    execute coupled earth system models dynamically via YAML configuration files,
+    reducing the need to write custom top-level application and driver code.
+    ESMX is included with the ESMF package.
+
+    ESMPy is a Python interface to the ESMF gridding engine. It allows for
+    high-performance, parallel regridding of fields between structured grids,
+    unstructured meshes, and observational data streams directly within Python
+    workflows. This bridges native ESMF capabilities with the broader Python
+    data science ecosystem.
+    ESMPy is included with the ESMF package."""
 
     homepage = "https://earthsystemmodeling.org/"
     url = "https://github.com/esmf-org/esmf/archive/v8.4.1.tar.gz"
@@ -453,6 +468,8 @@ class MakefileBuilder(makefile.MakefileBuilder):
         self, env: EnvironmentModifications, dependent_spec: Spec
     ) -> None:
         env.set("ESMFMKFILE", os.path.join(self.prefix.lib, "esmf.mk"))
+        if self.spec.satisfies("@8:"):
+            env.append_path("CMAKE_PREFIX_PATH", self.prefix.cmake)
 
     def install(self, pkg, spec, prefix):
         make("install")


uturuncoglu (approved):
Looks good to me. Thanks.

theurich:
FYI @danrosen25. For some reason I cannot make you a reviewer. Strange.

repos/spack_repo/builtin/packages/esmf/package.py
```diff
@@ -453,6 +468,8 @@ def setup_dependent_build_environment(
         self, env: EnvironmentModifications, dependent_spec: Spec
     ) -> None:
         env.set("ESMFMKFILE", os.path.join(self.prefix.lib, "esmf.mk"))
+        if self.spec.satisfies("@8:"):
+            env.append_path("CMAKE_PREFIX_PATH", self.prefix.cmake)
```
johnwparent: Is this intended to be a config or find module? Typically if a project is providing its own cmake detection file, its done as a config module and named something like ESMFConfig.cmake. I assumed in the original issue thats what was intended, but if this is indeed a find module, you'll want to use `CMAKE_MODULE_PATH` instead. 
Apologies for any confusion. 

If possible, it may be more idiomatic/easier to generate a config module. Id be happy to provide guidance there as well.

theurich: Yes, right now what we have under `<prefix>/cmake` is a `FindESMF.cmake`. Makes sense to append to `CMAKE_MODULE_PATH`.

theurich: See 35a7317

danrosen25:
@theurich
I'm noticing now that you started this. I had done something like this, which reads `ESMF_CMAKEDIR` from `esmf.mk`. But it might be overkill?

```diff
--- a/repos/spack_repo/builtin/packages/esmf/package.py
+++ b/repos/spack_repo/builtin/packages/esmf/package.py
@@ -453,6 +453,16 @@ class MakefileBuilder(makefile.MakefileBuilder):
         self, env: EnvironmentModifications, dependent_spec: Spec
     ) -> None:
         env.set("ESMFMKFILE", os.path.join(self.prefix.lib, "esmf.mk"))
+        # Append ESMF cmake directory to CMAKE_PREFIX_PATH so that
+        # dependent packages can use FindESMF.cmake
+        esmf_cmakedir = os.path.join(self.prefix, "cmake")
+        if os.path.exists(os.path.join(self.prefix.lib, "esmf.mk")):
+            with open(os.path.join(self.prefix.lib, "esmf.mk")) as f:
+                for line in f:
+                    if line.startswith("ESMF_CMAKEDIR=") :
+                        esmf_cmakedir = line.split("=", 1)[1].strip()
+                        break
+        env.append_path("CMAKE_PREFIX_PATH", esmf_cmakedir)
 
     def install(self, pkg, spec, prefix):
         make("install")
```

theurich:
@danrosen25, my thinking is that we know where the `cmake/FindESMF.cmake` goes during a Spack installation. So that's the location we need to pass on. And `esmf.mk` is by design constructed to be consistent with it, thus should not provide anything different.

danrosen25:
@theurich @johnwparent @AlexanderRichert-NOAA
I'm testing changes but adding CMAKE_MODULE_PATH to the build environment does not pass this variable into the dependencies build.

So I'm testing this instead but it seems a little hacky
```python
    def dependent_cmake_args(self, dependent_spec):
        default_cmakedir = os.path.join(self.prefix, "cmake")
        esmf_cmake_args = f"-DCMAKE_MODULE_PATH:STRING={default_cmakedir};${{CMAKE_MODULE_PATH}}"
        return [esmf_cmake_args]
```

My testing strategy is to add the following lines to a CMake build package that doesn't have a FindESMF.cmake file. This will let me take a look at what CMake is using for the module path, if find_package works, and what the ESMFMKFILE is after find package.
```cmake
message (WARNING "CMAKE_MODULE_PATH is ${CMAKE_MODULE_PATH}")
find_package(ESMF REQUIRED)
message (FATAL_ERROR "ESMFMKFILE is ${ESMFMKFILE}")
```

and it does work
```console
> CMake Warning at CMakeLists.txt:120 (message):
    CMAKE_MODULE_PATH is
    /home/esmfdev/spack/opt/spack/linux-m1/esmf-8.9.1-hc7zxe2vdynotdokkxvukiksa2y5w6lt/cmake;/tmp/esmfdev/spack-stage/spack-stage-pfunit-4.18.1-4xgpoxwmxm2up3tnfg5mryqj3cmcjbiu/spack-src/cmake;/tmp/esmfdev/spack-stage/spack-stage-pfunit-4.18.1-4xgpoxwmxm2up3tnfg5mryqj3cmcjbiu/spack-src/include
  
  
  -- Found ESMF: /home/esmfdev/spack/opt/spack/linux-m1/esmf-8.9.1-hc7zxe2vdynotdokkxvukiksa2y5w6lt/lib/libesmf.so (found version "8.9.1")
> CMake Error at CMakeLists.txt:122 (message):
    ESMFMKFILE is
    /home/esmfdev/spack/opt/spack/linux-m1/esmf-8.9.1-hc7zxe2vdynotdokkxvukiksa2y5w6lt/lib/esmf.mk
```

johnwparent:
> @theurich @johnwparent @AlexanderRichert-NOAA I'm testing changes but adding CMAKE_MODULE_PATH to the build environment does not pass this variable into the dependencies build.
> 
> So I'm testing this instead but it seems a little hacky
> 
> ```python
>     def dependent_cmake_args(self, dependent_spec):
>         default_cmakedir = os.path.join(self.prefix, "cmake")
>         esmf_cmake_args = f"-DCMAKE_MODULE_PATH:STRING={default_cmakedir};${{CMAKE_MODULE_PATH}}"
>         return [esmf_cmake_args]
> ```
> 
> My testing strategy is to add the following lines to a CMake build package that doesn't have a FindESMF.cmake file. This will let me take a look at what CMake is using for the module path, if find_package works, and what the ESMFMKFILE is after find package.
> 
> ```cmake
> message (WARNING "CMAKE_MODULE_PATH is ${CMAKE_MODULE_PATH}")
> find_package(ESMF REQUIRED)
> message (FATAL_ERROR "ESMFMKFILE is ${ESMFMKFILE}")
> ```
> 
> and it does work
> 
> ```
> > CMake Warning at CMakeLists.txt:120 (message):
>     CMAKE_MODULE_PATH is
>     /home/esmfdev/spack/opt/spack/linux-m1/esmf-8.9.1-hc7zxe2vdynotdokkxvukiksa2y5w6lt/cmake;/tmp/esmfdev/spack-stage/spack-stage-pfunit-4.18.1-4xgpoxwmxm2up3tnfg5mryqj3cmcjbiu/spack-src/cmake;/tmp/esmfdev/spack-stage/spack-stage-pfunit-4.18.1-4xgpoxwmxm2up3tnfg5mryqj3cmcjbiu/spack-src/include
>   
>   
>   -- Found ESMF: /home/esmfdev/spack/opt/spack/linux-m1/esmf-8.9.1-hc7zxe2vdynotdokkxvukiksa2y5w6lt/lib/libesmf.so (found version "8.9.1")
> > CMake Error at CMakeLists.txt:122 (message):
>     ESMFMKFILE is
>     /home/esmfdev/spack/opt/spack/linux-m1/esmf-8.9.1-hc7zxe2vdynotdokkxvukiksa2y5w6lt/lib/esmf.mk
> ```

Would this approach not override a dependent package's module path if they were setting one? Or at least create conflicting variables (last one takes precedence IIRC)

Ugh, right, I always forget CMAKE_MODULE_PATH doesn't behave like CMAKE_PREFIX_PATH w.r.t. env variables. 🤦

theurich:
@danrosen25 good point about `setup_dependent_build_environment()` not providing what we need when this is loaded via "spack load esmf". I think it also needs to be added to `setup_run_environment()`. I will push a change...

danrosen25:
@johnwparent
Modifying CMAKE_MODULE_PATH dependent_cmake_args probably does override the CMake arguments set by the dependent package or other package dependencies of the dependent package. That's why I don't like it and think it's "hacky" and I was doing some testing.

@AlexanderRichert-NOAA 
We do provide a way to customize the installation location for the `FindESMF.cmake` module, `ESMF_INSTALL_CMAKEDIR`
https://earthsystemmodeling.org/docs/release/latest/ESMF_usrdoc/node10.html#SECTION000109000000000000000. Would it make sense to expose this to the spack build so that you can choose where to install `FindESMF.cmake`?

AlexanderRichert-NOAA:
Off the top of my head I can't think of a way that would help us. Right now the workaround is something like `-DCMAKE_MODULE_PATH=$(spack location --install-dir esmf)/cmake`, but customizing the path to a different location wouldn't really get around the need for that.

climbfuji:
Sorry for being late to the party. What is the problem with this short code snippet in customers of ESMF?
```
if(DEFINED ENV{esmf_ROOT})
  list(APPEND CMAKE_MODULE_PATH $ENV{esmf_ROOT}/cmake)
  find_package(ESMF 8.9.0 MODULE REQUIRED)
else()
  message(FATAL_ERROR "Environment variable esmf_ROOT not defined")
endif()
```
Maybe trying to solve this here is overkill?

danrosen25:
@climbfuji 
In order for that to work we'd have to add `esmf_ROOT` (or `ESMF_ROOT`) to the `setup_dependent_build_environment` or we can add `[f"-DESMF_ROOT:STRING={self.prefix}"]` to `dependent_cmake_args` and drop the `ENV` requirement. Both provide an option to pass the ESMF_ROOT location directly into a CMake build through spack.

Or the dependent package spack build can compute this using `os.path.join(self.spec['esmf'].prefix, "cmake")` and  set `CMAKE_MODULE_PATH` in `cmake_args`

But I think these are better solutions than having ESMF modify CMAKE_MODULE_PATH for depends, which most likely breaks things.

theurich:
I am still confused what is wrong (or does not work) with 
```
        env.set("ESMFMKFILE", os.path.join(self.prefix.lib, "esmf.mk"))
        if self.spec.satisfies("@8:"):
            env.append_path("CMAKE_MODULE_PATH", self.prefix.cmake)
```
The only thing we have in ESMF's `cmake` directory is the official `FindESMF.cmake` (well and a `README.md`).

I thought that  `setup_run_environment()` was exactly for those cases where you needed to do something to the  runner _environment_ when you do `spack load abc` for package `abc`.

So isn't this doing exactly what @AlexanderRichert-NOAA was asking for, `spack load esmf`, and then your dependent cmake build can do `find_package(ESMF 8.9.0 MODULE REQUIRED)` without any further ado?

AlexanderRichert-NOAA:
With the caveat that I haven't tested it, yes, @theurich what you're describing sounds exactly like what I have in mind. Ideally I think we want to be able to build dependents both in Spack and outside of Spack directly through CMake without further customization. I think in an ideal world something like what @johnwparent described about using ESMFConfig.cmake/CMAKE_PREFIX_PATH would be optimal, but I realize that's a bigger lift, so I think what you're describing makes sense.

danrosen25:
@theurich 
Setting `CMAKE_MODULE_PATH` in `setup_run_environment` will only execute environment changes during `spack load esmf`, which doesn't happen during the spack build chain. It is helpful if you want to run cmake outside of spack **IF** the call to cmake includes `-DCMAKE_MODULE_PATH=${CMAKE_MODULE_PATH}`. Setting `CMAKE_MODULE_PATH` in `setup_dependent_build_environment` will execute environment changes during the spack build chain for dependent software **BUT** spack doesn't automatically pass `CMAKE_MODULE_PATH` (similar to above) from the environment to the cmake build command. Also, the cmake build for the particular software may or may not overwrite `CMAKE_MODULE_PATH` even if it was passed in using `-DCMAKE_MODULE_PATH=${CMAKE_MODULE_PATH}`.

@AlexanderRichert-NOAA wants to replicate the behavior of a cmake ESMF build, which produces a <Package>Config.cmake file(s), which are used for including cmake build targets. These files can be picked up cmake using locations set in CMAKE_PREFIX_PATH to automatically find packages, which is what spack does. Of course it's not the same as FindESMF.cmake, which sets ESMF_ variables.

In the end what's needed is a way for the location of FindESMF.cmake to be picked up by a build but there's no automated way to do so unless FindESMF.cmake is installed into the [CMake included modules](https://github.com/Kitware/CMake/tree/master/Modules). I think setting `CMAKE_MODULE_PATH` in the **dependent build** and **run** environments is a solution but it's not robust and it won't automate a solution. Another package can choose to override it using `env.set` instead of `env.append_path` and it doesn't even pass into a cmake build without `-DCMAKE_MODULE_PATH=${CMAKE_MODULE_PATH}`

tldr; There are caveats using a shared environment variable, and this particular shared environment variable isn't automatically passed from the environment to a cmake build, even in spack. Setting `ESMF_CMAKE_MODULE_PATH` might make more sense.

theurich:
Thank you @danrosen25 for the excellent reminder how `CMAKE_MODULE_PATH` differs from `CMAKE_PREFIX_PATH`.  I've been thinking for a while that we should provide a standard `ESMFConfig.cmake` during the ESMF install step. I think this is also consistent with a [comment](https://github.com/spack/spack-packages/pull/4919#discussion_r3277672805) by @johnwparent earlier. So I finally took some time to work on this and opened https://github.com/esmf-org/esmf/pull/549 on the ESMF side.

My hope is that we get something like that into ESMF v9. If so then on the Spack level there is very little extra work (as far as I can tell). Just a little clean-up, which I have under new PR spack/spack#4931, which essentially supersedes this PR here.

danrosen25:
I tested the changes from https://github.com/esmf-org/esmf/pull/549 within spack by adding the following snippet to a CMakeLists.txt file without adding new MODULE ore PREFIX paths and it found ESMF successfully.
```cmake
message (WARNING "CMAKE_PREFIX_PATH is ${CMAKE_PREFIX_PATH}")
find_package(ESMF REQUIRED)
message (FATAL_ERROR "ESMF_FOUND is ${ESMF_FOUND}")
```

```console
> CMake Warning at CMakeLists.txt:120 (message):
    CMAKE_PREFIX_PATH is
    /home/esmfdev/spack/opt/spack/linux-m1/compiler-wrapper-1.1.0-ojzz7oyy72e76bh7gb3wlf27tosmiofl;/home/esmfdev/spack/opt/spack/linux-m1/esmf-8.9.1-n6acrq5oez2nypcgpespdk5j7xqwtzen;/home/esmfdev/spack/opt/spack/linux-m1/fargparse-1.11.0-vrexpvgack3nxcv7peaaxnpylcuva5zp;/home/esmfdev/spack/opt/spack/linux-m1/parallelio-2.6.8-6jtmjqgptpvpwk4iqnctmltbyadhee46;/home/esmfdev/spack/opt/spack/linux-m1/gftl-shared-1.12.0-6sv25x62uqicirdcgwz5krophkf65f2t;/home/esmfdev/spack/opt/spack/linux-m1/gftl-1.17.0-73fb5f5zc7dyahktahzhvitoxoe2fqjm;/home/esmfdev/spack/opt/spack/linux-m1/gcc-runtime-15.2.0-fyjyijeds2uwlj2ssnu5hxwhyfgsj54g
  
  
  -- Found MPI_C: /usr/lib/aarch64-linux-gnu/mpich/lib/libmpich.so (found version "4.1")
  -- Found MPI_Fortran: /usr/lib/aarch64-linux-gnu/mpich/lib/libmpichfort.so (found version "4.1")
  -- Found MPI: TRUE (found version "4.1") found components: Fortran C
> CMake Error at CMakeLists.txt:122 (message):
    ESMF_FOUND is 1
```

spack automatically adds the CMAKE_PREFIX_PATH for ESMF
``` console
    /usr/bin/cmake -G 'Unix Makefiles' -DCMAKE_INSTALL_PREFIX:STRING=/home/esmfdev/spack/opt/spack/linux-m1/pfunit-4.18.1-smljkqynfgpmiph7ddpczf4znjnlpvdq -DCMAKE_INSTALL_RPATH_USE_LINK_PATH:BOOL=ON '-DCMAKE_INSTALL_RPATH:STRING=/home/esmfdev/spack/opt/spack/linux-m1/pfunit-4.18.1-smljkqynfgpmiph7ddpczf4znjnlpvdq/lib;/home/esmfdev/spack/opt/spack/linux-m1/pfunit-4.18.1-smljkqynfgpmiph7ddpczf4znjnlpvdq/lib64' '-DCMAKE_PREFIX_PATH:STRING=/home/esmfdev/spack/opt/spack/linux-m1/compiler-wrapper-1.1.0-ojzz7oyy72e76bh7gb3wlf27tosmiofl;/home/esmfdev/spack/opt/spack/linux-m1/esmf-8.9.1-n6acrq5oez2nypcgpespdk5j7xqwtzen;/home/esmfdev/spack/opt/spack/linux-m1/fargparse-1.11.0-vrexpvgack3nxcv7peaaxnpylcuva5zp;/home/esmfdev/spack/opt/spack/linux-m1/parallelio-2.6.8-6jtmjqgptpvpwk4iqnctmltbyadhee46;/home/esmfdev/spack/opt/spack/linux-m1/gftl-shared-1.12.0-6sv25x62uqicirdcgwz5krophkf65f2t;/home/esmfdev/spack/opt/spack/linux-m1/gftl-1.17.0-73fb5f5zc7dyahktahzhvitoxoe2fqjm;/home/esmfdev/spack/opt/spack/linux-m1/gcc-runtime-15.2.0-fyjyijeds2uwlj2ssnu5hxwhyfgsj54g' -DCMAKE_BUILD_TYPE:STRING=Release -DCMAKE_VERBOSE_MAKEFILE:BOOL=ON -DCMAKE_INTERPROCEDURAL_OPTIMIZATION:BOOL=OFF -DCMAKE_POLICY_DEFAULT_CMP0090:STRING=NEW -DCMAKE_FIND_USE_PACKAGE_REGISTRY:BOOL=OFF -DCMAKE_EXPORT_COMPILE_COMMANDS:BOOL=ON -DPYTHON_EXECUTABLE:PATH=/usr/bin/python3.13 -DPython_EXECUTABLE:PATH=/usr/bin/python3.13 -DPython3_EXECUTABLE:PATH=/usr/bin/python3.13 -DBUILD_SHARED_LIBS:BOOL=OFF -DCMAKE_Fortran_MODULE_DIRECTORY:STRING=/home/esmfdev/spack/opt/spack/linux-m1/pfunit-4.18.1-smljkqynfgpmiph7ddpczf4znjnlpvdq/include -DENABLE_BUILD_DOXYGEN:BOOL=OFF -DENABLE_TESTS:BOOL=OFF -DSKIP_MPI:BOOL=ON -DSKIP_OPENMP:BOOL=ON -DSKIP_FHAMCREST:BOOL=ON -DSKIP_ESMF:BOOL=OFF -DMAX_ASSERT_RANK:STRING=5 /tmp/esmfdev/spack-stage/spack-stage-pfunit-4.18.1-smljkqynfgpmiph7ddpczf4znjnlpvdq/spack-src
```

And I tested non-spack installations of ESMF dependents. If ESMF_ROOT is set to $(spack location -i esmf) then it also works. So we can take care of this by adding `ESMF_ROOT` to `setup_run_environment`

theurich:
Closing this PR as superseded by spack/spack#4931

PR #4931, opened 2026-05-22
esmf: Clean up two issues in post_install() for ESMF version 9

theurich:
This PR does some house-cleaning of the ESMF `package.py` in anticipation of ESMF version 9 release:
- With generated standard `ESMFConfig.cmake` we no longer need (nor should) install the raw `cmake` directory from the ESMF source tree. Note that this depends on ESMF PR https://github.com/esmf-org/esmf/pull/549
- The issue with MPI C++ link dependency has been fixed inside ESMF.

I think that together with https://github.com/esmf-org/esmf/pull/549 this will address https://github.com/spack/spack-packages/issues/5045

This PR supersedes spack/spack#4919.


repos/spack_repo/builtin/packages/esmf/package.py
```diff
@@ -432,8 +432,11 @@ def setup_build_environment(self, env: EnvironmentModifications) -> None:
 
     @run_after("install")
     def post_install(self):
-        if self.spec.satisfies("@8:"):
+        # Only copy the raw source cmake tree for legacy v8 installations.
+        # The v9.0+ build generates standard ESMFConfig.cmake for lib/cmake/ESMF/
+        if self.spec.satisfies("@8:8.99.99"):
```
climbfuji:
```suggestion
        if self.spec.satisfies("@8"):
```
this should do

theurich: Simpler is better. Thanks!

climbfuji (approved):

theurich:
@climbfuji after some more digging I now see that it is only during the `@8:8.6` version range that requires the explicit `cmake` dir handling during post installation in the Spack recipe for ESMF. For versions since, this is handled by ESMF's install target directly. Sorry for the back and forth on that. This should be it now.

FYI @danrosen25

climbfuji (approved):

alalazo (approved):

PR #4940, opened 2026-05-26
detray: deprecate pre-1.0.0 versions

stephenswat:
This commit deprecates versions of detray before v1.0.0, and also deprecates the dfelibs package; the deprecated detray versions are the only packages that depend on this and thus they can be deprecated together.


wdconinc (approved):
Thanks!

PR #4944, opened 2026-05-26
add python-3.8.16

gberg617:


adamjstewart (changes requested):
Python 3.8 has reached EOL and no longer receives security fixes: https://devguide.python.org/versions/

3.8.16 in particular is riddled with security vulnerabilities: https://www.cvedetails.com/vulnerability-list/vendor_id-10210/product_id-18230/version_id-1371138/Python-Python-3.8.16.html

We already have 3.8.20 in the package (deprecated). Can you explain why you need 3.8.16?

gberg617:
> Python 3.8 has reached EOL and no longer receives security fixes: https://devguide.python.org/versions/
> 
> 3.8.16 in particular is riddled with security vulnerabilities: https://www.cvedetails.com/vulnerability-list/vendor_id-10210/product_id-18230/version_id-1371138/Python-Python-3.8.16.html
> 
> We already have 3.8.20 in the package (deprecated). Can you explain why you need 3.8.16?

Thanks for letting me know.  I will update my code's python version to one that is still supported and close this PR.

PR #4945, opened 2026-05-26
openblas: patch 0.3.33 to fix build on macos

sethrj:
Extends #4569 to the new version from #4282 to fix macos build.


mathomp4 (approved):

alalazo (approved):
@sethrj Feel free to merge if CI goes ✔️

PR #4920, opened 2026-05-21
openbabel: place upper bound on Python dependency

berquist:
Closes https://github.com/spack/spack-packages/issues/2175

This also fixes a few other things I noticed when looking at the CMake output.

PR #4950, opened 2026-05-27
openbabel: add v3.2.0

### LINKED ISSUE #1257: `openbabel` tries to install at `/`
When installing `openbabel@3.1.1 +coordgen ~cairo ~gui` with spack 1.0.1, after the compilation it tries to install something inside the directory `/openbabel` which produces the following error when the user has not sufficient permissions.
```
  >> 8862    CMake Error at scripts/cmake_install.cmake:60 (file):
     8863      file cannot create directory: /openbabel.  Maybe need administrative
     8864      privileges.
     8865    Call Stack (most recent call first):
     8866      cmake_install.cmake:93 (include)
```

@RMeli

berquist:
Closes https://github.com/spack/spack-packages/issues/1257

(This issue was actually closed by https://github.com/spack/spack-packages/pull/4920 but this is a necessary follow-up anyway.)

Building with the GUI requires wxWidgets; I found that my system install was leaking into the build and causing symbol problems.  It wasn't possible to even turn off finding it before https://github.com/openbabel/openbabel/pull/2964.

I think that, because the GUI dependencies are quite heavy, and I don't suspect the GUI is used much, that `gui` should no longer be default.


repos/spack_repo/builtin/packages/openbabel/package.py
```diff
@@ -66,6 +70,9 @@ class Openbabel(CMakePackage):
     # https://github.com/openbabel/openbabel/pull/2493
     patch("cmake-time.patch", when="@3.1.1")
 
+    # https://github.com/openbabel/openbabel/pull/2964
+    patch("2964-optional-wxwidgets.patch", when="@:3.2.0")
```
RMeli: Will be fixed in `3.2.1`. 

Nitpick: since you have a PR with exactly the same diff, I'd patch from the PR instead.

See https://spack.readthedocs.io/en/latest/packaging_guide_creation.html#url-patch-file.

RMeli (dismissed):
LGTM. I agree that defaulting to `~gui` is likely more sensible.

RMeli (approved):
LGTM, thanks!

PR #4956, opened 2026-05-27
Emacs: remove json support @30:

JohnGouwar:
Emacs switched from using `jansson` to native json support in Emacs 30. See: https://www.gnu.org/software/emacs/manual/html_node/efaq/New-in-Emacs-30.html. I also confirmed this with `./configure --help`.


alecbcs (approved):
Looks good to me. Thanks @JohnGouwar!

JohnGouwar:
At some point I'll update to add the pgtk backend, I just haven't had the time. Thanks for the quick review @alecbcs

PR #4962, opened 2026-05-28
rmgdft: enable ROCm builds

alephpiece:
This PR completes the existing `+rocm` variant for RMGDFT.

For `+rocm` builds, it:

- uses `ROCmPackage` and builds and installs the GPU targets;
- enables `RMG_HIP_ENABLED`;
- passes the HIP and ROCm root paths required by RMGDFT's `FindHIP.cmake` logic; and
- declares the ROCm libraries that upstream CMake finds directly.

It also fixes installation of the bundled test inputs. They were previously copied from the CMake build tree, which only contains generated build files. They are now copied from the source tree so that `share/tests/RMG` contains the actual runnable input files.


alephpiece:
@spackbot fix style

repos/spack_repo/builtin/packages/rmgdft/package.py
```diff
@@ -57,6 +58,7 @@ class Rmgdft(CMakePackage, CudaPackage):
     conflicts("%gcc@:4", when="@3.6.0:", msg=compiler_warning14)
     conflicts("%intel@:17", when="@3.6.0:", msg=compiler_warning14)
     conflicts("%llvm@:3.4", when="@3.6.0:", msg=compiler_warning14)
+    conflicts("+cuda", when="+rocm", msg="RMGDFT supports only one GPU backend at a time")
```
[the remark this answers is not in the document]
bernhardkaindl: Do, it doesn't redefine/update the variant, and btw, that would be possible even.

repos/spack_repo/builtin/packages/rmgdft/package.py
```diff
@@ -116,6 +122,11 @@ def cmake_args(self):
             cuda_arch = cuda_arch_list[0]
             if cuda_arch != "none":
                 args.append("-DCUDA_FLAGS=-arch=sm_{0}".format(cuda_arch))
+        if "+rocm" in spec:
+            args.append("-DRMG_HIP_ENABLED=1")
+            args.append(self.define("HIP_PATH", spec["hip"].prefix))
+            args.append(self.define("ROCM_ROOT", spec["rocm-core"].prefix))
```
alalazo, line 128: Doesn't this flag prevent using the package against Spack installed ROCm? Did you try a build of that kind?

repos/spack_repo/builtin/packages/rmgdft/package.py
alalazo, line 44: This can be removed

alalazo (changes requested):

PR #4979, opened 2026-05-28
py-orjson: add v3.11.5

thomas-bouvier:
Also fixes the rust dependency version.


alecbcs (approved):

PR #4985, opened 2026-05-28
py-semver: fix bound on setuptools

thomas-bouvier:


alecbcs (approved):

PR #4988, opened 2026-05-28
py-redis: add v7.1.0

thomas-bouvier:

### DIFF
--- a/repos/spack_repo/builtin/packages/py_redis/package.py
+++ b/repos/spack_repo/builtin/packages/py_redis/package.py
@@ -27,12 +27,18 @@ class PyRedis(PythonPackage):
         description="Support for hiredis which speeds up parsing of multi bulk replies",
     )
 
-    depends_on("py-hatchling", when="@6:", type="build")
-    depends_on("py-setuptools", when="@:5", type="build")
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/py_redis/package.py
```diff
@@ -26,9 +27,11 @@ class PyRedis(PythonPackage):
         description="Support for hiredis which speeds up parsing of multi bulk replies",
     )
 
-    depends_on("py-setuptools", type="build")
+    depends_on("py-hatchling", when="@6:", type="build")
+    depends_on("py-setuptools", when="@:5", type="build")
```
alecbcs:
```suggestion
    depends_on("python@3.10:", type=("build", "run"), when="@7.1.0:")
    
    depends_on("py-hatchling", when="@6:", type="build")
    depends_on("py-setuptools", when="@:5", type="build")
```

According to the `pyproject.toml` in the upstream repo `python@3.10:` is now required.

alecbcs (changes requested):
Missing dependency on `python@3.10:`.

alecbcs (approved):

PR #4990, opened 2026-05-28
py-python-slugify: add v8.0.4

thomas-bouvier:


manuelakuhn (approved):

thomas-bouvier:
Thank you @manuelakuhn

PR #4994, opened 2026-05-28
givaro: add v4.2.2

d-torrance:

### DIFF
--- a/repos/spack_repo/builtin/packages/givaro/package.py
+++ b/repos/spack_repo/builtin/packages/givaro/package.py
@@ -22,6 +22,7 @@ class Givaro(AutotoolsPackage):
 
     license("CECILL-B", checked_by="d-torrance")
 
+    version("4.2.2", sha256="53e9fb290deb0e20799c62d250d65c2226013d60b4cebe6b0b54c73000cb8fff")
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/givaro/package.py
```diff
@@ -23,6 +23,7 @@ class Givaro(AutotoolsPackage):
     license("CECILL-B", checked_by="d-torrance")
 
     version("4.2.1", sha256="feefb7445842ceb756f8bb13900d975b530551e488a2ae174bda7b636251de43")
+    version("4.2.2", sha256="53e9fb290deb0e20799c62d250d65c2226013d60b4cebe6b0b54c73000cb8fff")
```
kshea21, line 25: We like to put the most recent versions at the top of the list of versions, could you switch this with 4.2.1?

kshea21 (changes requested):

repos/spack_repo/builtin/packages/givaro/package.py
d-torrance: Done!

kshea21 (approved):
Thanks!!

PR #4997, opened 2026-05-28
sst: add v16.0.0 for core, elements, and macro

berquist:
Compilation checked on:
- Rocky 8 with GCC 8.5.0
- macOS 26 with Apple Clang 21
all with both Python 3.9 and 3.14


jmlapre (approved):
Everything looks solid to me.

alecbcs (approved):
Looks good to me. Thanks @berquist!

PR #5001, opened 2026-05-28
py-sqlalchemy: add variant +asyncio

thomas-bouvier:
This is a bit artifical as `greenlet` was already modeled as a dependency. Also fixes the upper bound of the conflict.


alecbcs (approved):

PR #5004, opened 2026-05-28
pfunit: add v4.18.2, v4.19.0

mathomp4:
This PR adds pfunit v4.18.2 and v4.19.0

v4.18.2 has a workaround for parameterised test cases when using flang and the second has improved XML output.


kshea21 (approved):
Great, thanks!

PR #5010, opened 2026-05-28
gcc: add `+graphite+bootstrap` conflict and other minor fixes

alalazo:
**gcc:**

- Fix bug in `detect_gdc`: `elif len(candidate_specs) == 0` was dead code and should be `== 1` (single-candidate fast path)
- Replace deprecated `self.compiler` with `self["c"]` (compiler-as-dependency API) in `detect_gdc`
- Remove dead code in `nvptx_install`: options built before the `working_dir` block were overwritten unconditionally (leftover from #40897)
- Convert remaining `.format()` strings to f-strings in `nvptx_install`
- Add `conflicts("+graphite +bootstrap")`: the combination causes reproducible bootstrap comparison failures on GCC 16 (stage 2/3 object files differ) due to the use of hash tables in `isl` keyed by pointer values

**isl:** add v0.27

**mpc:** add v1.4.1

### DIFF
--- a/repos/spack_repo/builtin/packages/gcc/package.py
+++ b/repos/spack_repo/builtin/packages/gcc/package.py
@@ -1026,7 +1026,6 @@ class Gcc(AutotoolsPackage, GNUMirrorPackage, CompilerPackage):
 
         self.link_newlib()
 
-        # self.build_directory = 'spack-build-nvptx'
         with working_dir("spack-build-nvptx", create=True):
             options = [
                 "--prefix={0}".format(prefix),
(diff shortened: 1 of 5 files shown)


haampie (approved):

haampie:
> Add conflicts("+graphite +bootstrap"): the combination causes reproducible bootstrap comparison failures on GCC 16 (stage 2/3 object files differ) due to the use of hash tables in isl keyed by pointer values

Fwiw, we should also fix the bootstrap bits. I think right now it's only PGO and not LTO, and iirc the latter requires a comparison that doesn't check debug symbols. Maybe something similar applies here?

PR #5017, opened 2026-05-29
kokkos: use correct compiler options for our testing

JBludau:
This came up in Kokkos' CI testing the spack recipe for the HIP backend.
The test might be compiled with the default host compiler if we don't catch this and redirect it to `hipcc` when Kokkos identifies as cxx library

### DIFF
--- a/repos/spack_repo/builtin/packages/py_librt/package.py
+++ b/repos/spack_repo/builtin/packages/py_librt/package.py
@@ -15,6 +15,7 @@ class PyLibrt(PythonPackage):
 
     license("MIT AND PSF-2.0")
 
+    version("0.11.0", sha256="075dc3ef4458a278e0195cbf6ac9d38808d9b906c5a6c7f7f79c3888276a3fb1")
     version("0.9.0", sha256="a0951822531e7aee6e0dfb556b30d5ee36bbe234faf60c20a16c01be3530869d")
     version("0.6.3", sha256="c724a884e642aa2bbad52bb0203ea40406ad742368a5f90da1b220e970384aae")
 


JBludau:
@spackbot fix style

repos/spack_repo/builtin/packages/kokkos/package.py
```diff
@@ -619,6 +619,10 @@ def test_run(self):
             cmake_args.append(
                 self.define("CMAKE_CXX_COMPILER", self["kokkos-nvcc-wrapper"].kokkos_cxx)
             )
+        elif self.spec.satisfies("+rocm") and not (
+            self.spec.satisfies("%cxx=clang") or self.spec.satisfies("%cxx=rocmcc")
+        ):
+            cmake_args.append(self.define("CMAKE_CXX_COMPILER", self.spec["hip"].hipcc))
```
tpadioleau: Fine for me in the case of `~cmake_lang`. I guess if we want to handle `+cmake_lang` we will end up repeating the same logic as in lines 524 to 566 ?

tpadioleau (approved):

repos/spack_repo/builtin/packages/kokkos/package.py
rbberger: We could add a utility function to do these kinds of things for both cases.

JBludau:
@spackbot fix style

repos/spack_repo/builtin/packages/kokkos/package.py
JBludau: Hmm I guess we will want to restructure that at some time in the future anyways ... but I hope this makes it work with all our configs for now

JBludau: @rbberger I did not see your comment before pushing, let me update the changes

repos/spack_repo/builtin/packages/kokkos/package.py
```diff
@@ -439,6 +439,50 @@ def kokkos_cxx(self) -> str:
...
+                    )
+                )
+            options.append(
+                self.define("CMAKE_CUDA_ARCHITECTURES", self.spec.variants["cuda_arch"].value)
+            )
+            options.append(self.define_from_variant("CMAKE_CUDA_STANDARD", "cxxstd"))
+            options.append(self.define("CMAKE_CUDA_EXTENSIONS", False))
+
+        else:
+            options.append(self.define("CMAKE_CXX_COMPILER", self["cxx"].cxx))
```
rbberger: This shouldn't be done in the `cmake_args` case since it would avoid the Spack compiler wrapper and therefore changes behavior.

rbberger (changes requested):

repos/spack_repo/builtin/packages/kokkos/package.py
```diff
@@ -439,6 +439,50 @@ def kokkos_cxx(self) -> str:
         # Assumes build-time globals have been set already
         return spack_cxx
 
+    def set_language_compiler_if_required(self, options):
```
cedricchevalier19: `set_language_compiler_if_required`: I think the `if_required` part should be removed and that `CMAKE_CXX_COMPILER` should not be set if not needed (see @rbberger comment below).

repos/spack_repo/builtin/packages/kokkos/package.py
```diff
@@ -615,12 +623,10 @@ def test_run(self):
             prefix_paths = ";".join(get_cmake_prefix_path(self))
             cmake_args.append(self.define("CMAKE_PREFIX_PATH", prefix_paths))
 
-        if self.spec.satisfies("+wrapper"):
-            cmake_args.append(
-                self.define("CMAKE_CXX_COMPILER", self["kokkos-nvcc-wrapper"].kokkos_cxx)
-            )
-        else:
-            cmake_args.append(self.define("CMAKE_CXX_COMPILER", self["cxx"].cxx))
+        self.set_language_compiler_if_required(cmake_args)
```
cedricchevalier19: Here we always need to set `CMAKE_CXX_COMPILER`.

cedricchevalier19 (changes requested):
I am still not convinced that defining compilers for CMake is enough for `test_run`.
What if the GPU stack requires loading some modules?
I do not think that the Spack test architecture is well suited for a library and I am not sure that we should go down the rabbit hole.

tpadioleau:
> I am still not convinced that defining compilers for CMake is enough for `test_run`. What if the GPU stack requires loading some modules? I do not think that the Spack test architecture is well suited for a library and I am not sure that we should go down the rabbit hole.

What about removing this `test_run` from the Spack package and put it directly in the Kokkos CI ?

JBludau:
> > I am still not convinced that defining compilers for CMake is enough for `test_run`. What if the GPU stack requires loading some modules? I do not think that the Spack test architecture is well suited for a library and I am not sure that we should go down the rabbit hole.
> 
> What about removing this `test_run` from the Spack package and put it directly in the Kokkos CI ?

I am rather for removing the test entirely as it seems to be pointless. For our CI I would try to go and see if I can modify Kokkos' smoke test in a way that it can be built against an installed Kokkos.

PR #5021, opened 2026-05-29
scale: needs nc-config that supports --fflags.  Require older NetCDF

GlazerMann:
NetCDF changed to having nc-config and nf-config, instead of just nc-config.  A good idea, but scale expects the old behavior.


t-yamaura (approved):

bernhardkaindl (dismissed):
Indeed, without this PR, the build fails because of --fflags
Other packages depending on netcdf-c might need the same dependency.

bernhardkaindl (approved):

PR #5023, opened 2026-05-29
geant4: be more explicit with cmake variables

sethrj:
This is a small refactor of the geant4 variant usage to better use the `when` and `define_from_variant` to explicitly include, exclude, or omit options based on whether they apply or are enabled/disabled. I verified from the g4 cmake that x11 is not compatible with macOS or Windows.


wdconinc:
Note the macOS audit failure in fairsoft now:
```
1. fairsoft-bundle: wrong variant in 'depends_on' directive
  File "/Users/runner/work/spack-packages/spack-packages/spack-core/lib/spack/spack/main.py", line 1109, in finish_parse_and_run
    variant 'motif' does not exist for 'geant4' when geant4~motif~opengl~qt~vecgeom~x11
```

sethrj:
@wdconinc Are you familiar with that software? It looks like it should be an environment, not a spack package:
```python
    """Bundle package providing default environment for FAIR software"""
    # Pin some variants:
    #...
    depends_on("geant4 ~qt~vecgeom~opengl~x11~motif")

    # ensure that OpenBLAS uses CMake build system (default Makefile causes issues on some Macs)
    depends_on("openblas build_system=cmake ~dynamic_dispatch")

    # Mostly for the experiments:
    depends_on("root +python+tmva+mlp+xrootd+sqlite")
```

wdconinc:
> @wdconinc Are you familiar with that software? It looks like it should be an environment, not a spack package:
> ```python
>     """Bundle package providing default environment for FAIR software"""
>     # Pin some variants:
>     #...
>     depends_on("geant4 ~qt~vecgeom~opengl~x11~motif")
> 
>     # ensure that OpenBLAS uses CMake build system (default Makefile causes issues on some Macs)
>     depends_on("openblas build_system=cmake ~dynamic_dispatch")
> 
>     # Mostly for the experiments:
>     depends_on("root +python+tmva+mlp+xrootd+sqlite")
> ```

I'm not a user but familiar with the software. It's a suite of software, a framework if you wish. I think distributing it as a bundle is fine (and distributing environments is not really supported). I'm not sure FAIRsoft is even supported on macOS. In any case, I think it's probably easy enough to remove the `~motif` since `~x11` is specified. That's what happens sometimes with these conditional variants.

wdconinc (approved):

wdconinc:
Let's give @dennisklein until Monday next week to weigh in on this as well.

dennisklein (approved):
> I verified from the g4 cmake that x11 is not compatible with macOS or Windows.

trusting you

PR #5026, opened 2026-05-29
fmt: libs property needs to check for libfmtd for debug builds

nkgrady:
Debug builds of FMT add a "d" suffix to the library creating `libfmtd` instead of `libfmt`. The default `libs` property does not find the libraries in this case. This PR simply adds a `libs` property that checks for both library names. 

The simplest way I could think of to show the error (I actually hit it with debug builds of an internal package) is to edit `spdlog/package.py` to add the line
```
print(self.spec["fmt"].libs)
```
to `def cmake_args(self):`, then try to install spdlog with a fmt debug build
```
spack install spdlog build_type=Debug ^ fmt build_type=Debug
```
This will fail with
```
Error: NoLibrariesError: Unable to recursively locate fmt libraries in [...]__spack_path_/linux-x86_64_v3/fmt-12.1.0-7imzckjqkc4wcceuldb5sgodw3aafbgg
```
because the library is named `libmftd.a`, not `libfmt.a`.

### DIFF
--- a/repos/spack_repo/builtin/packages/fmt/package.py
+++ b/repos/spack_repo/builtin/packages/fmt/package.py
@@ -157,4 +157,9 @@ class Fmt(CMakePackage):
     @property
     def libs(self):
         # Debug builds provide libfmtd instead of libfmt
-        return find_libraries(["libfmt","libfmtd"], root=self.home, recursive=True, shared=self.spec.satisfies("+shared"))
+        return find_libraries(
+            ["libfmt", "libfmtd"],
+            root=self.home,
+            recursive=True,
+            shared=self.spec.satisfies("+shared"),
+        )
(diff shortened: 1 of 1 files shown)


msimberg (dismissed):
Thanks for fixing this! I think this looks like a reasonable approach.

An alternative could be to change the library suffix with `FMT_DEBUG_POSTFIX` (probably remove the `d`) or find libraries based on `build_type`/`CMAKE_BUILD_TYPE`... I'm not convinced either of these is a better option, so approving, but giving you some time to comment before merging.

nkgrady:
The current approach is my preference: simple and unlikely to result in FMT version dependent logic if something changes in the upstream package. I thought about the two alternatives you suggest and my thinking is:

- `FMT_DEBUG_POSTFIX` is equally good. My thought is if spack can handle the upstream package's default choices without much trouble, I'd prefer to adapt spack to the package than the other way around. Admittedly, I also just like the d postfix :)

- `CMAKE_TOOLCHAIN_FILE` could probably change the build type without the package seeing `build_type or `CMAKE_BUILD_TYPE` in the options. Figure if I can immediately think of one way I could break the check, it is better to just look for the library.

msimberg:
@spackbot fix style

msimberg (approved):

PR #5033, opened 2026-05-29
macaulay2: new package

d-torrance:
[Macaulay2](https://www.macaulay2.com) is a software system devoted to supporting research in algebraic geometry and commutative algebra. It provides a specialized programming language designed for computations with polynomial rings, ideals, modules, and other algebraic structures, with a particular emphasis on Gröbner basis techniques.

### DIFF
--- a/repos/spack_repo/builtin/packages/macaulay2/allow-older-gfan.patch
+++ b/repos/spack_repo/builtin/packages/macaulay2/allow-older-gfan.patch
@@ -1,0 +1,35 @@
+--- a/M2/configure
++++ b/M2/configure
+@@ -14165,7 +14165,7 @@ then gfan_version=`gfan _version | head -2 | tail -1 | sed 's/gfan//'`
+                      -e 's/[^0-9]//g'`
+ 
+ 
+-  ax_compare_version_B=`echo "0.8" | sed -e 's/\([0-9]*\)/Z\1Z/g' \
++  ax_compare_version_B=`echo "0.6" | sed -e 's/\([0-9]*\)/Z\1Z/g' \
+                      -e 's/Z\([0-9]\)Z/Z0\1Z/g' \
+                      -e 's/Z\([0-9][0-9]\)Z/Z0\1Z/g' \
+                      -e 's/Z\([0-9][0-9][0-9]\)Z/Z0\1Z/g' \
+@@ -14181,8 +14181,8 @@ x$ax_compare_version_B" | sed 's/^ *//' | sort -r | sed "s/x${ax_compare_version
+     { printf "%s\n" "$as_me:${as_lineno-$LINENO}: result: yes" >&5
+ printf "%s\n" "yes" >&6; }
+            FILE_PREREQS="$FILE_PREREQS `command -v gfan`"
+-    else { printf "%s\n" "$as_me:${as_lineno-$LINENO}: result: yes, but < 0.8, will build" >&5
+-printf "%s\n" "yes, but < 0.8, will build" >&6; }
++    else { printf "%s\n" "$as_me:${as_lineno-$LINENO}: result: yes, but < 0.6, will build" >&5
++printf "%s\n" "yes, but < 0.6, will build" >&6; }
+            BUILD_gfan=yes
+   fi
+ 
+--- a/M2/Macaulay2/packages/gfanInterface.m2
++++ b/M2/Macaulay2/packages/gfanInterface.m2
+@@ -1013,9 +1013,7 @@
+ 	if gfanProgram === null then
+ 	    gfanProgram = findProgram("gfan", "gfan --help",
+ 		Verbose => gfanVerbose,
+-		-- version 0.8 is required (the 0.6.2 series predates several
+-		-- features used here and is no longer supported upstream)
+-		MinimumVersion => ("0.8",
++		MinimumVersion => ("0.6",
+ 		    "gfan _version | head -2 | tail -1 | sed 's/gfan//'"));
+ 	tmpFile := gfanMakeTemporaryFile data;
+ 
--- a/repos/spack_repo/builtin/packages/macaulay2/package.py
+++ b/repos/spack_repo/builtin/packages/macaulay2/package.py
@@ -1,0 +1,93 @@
+# Copyright Spack Project Developers. See COPYRIGHT file for details.
+#
+# SPDX-License-Identifier: (Apache-2.0 OR MIT)
+
+from spack_repo.builtin.build_systems.autotools import AutotoolsPackage
+
+from spack.package import *
+
+
+class Macaulay2(AutotoolsPackage):
+    """
+    Macaulay2 is a software system devoted to supporting research in
+    algebraic geometry and commutative algebra. It provides a specialized
+    programming language designed for computations with polynomial rings,
+    ideals, modules, and other algebraic structures, with a particular
+    emphasis on Gröbner basis techniques.
+    """
+
+    homepage = "https://macaulay2.com/"
+    url = "https://macaulay2.com/Downloads/SourceCode/Macaulay2-1.25.11.tar.gz"
+
+    maintainers("d-torrance")
+
+    license("GPL-2.0-or-later", checked_by="d-torrance")
+
+    version("1.26.06", sha256="9ba54041626f6364036f84b4168704d74136d1bec2dbe08d9d0fb3fef2a0c854")
+    version("1.26.05", sha256="a76213b4a23542abe5742a4d2b782e26ee5f3f9b1f6a2fc3a992615cc3aa3431")
+    version("1.25.11", sha256="ecee06ea456f87d8182bdaf59f9e2f4a330d861516ab7d00faf9f37699ddeb58")
+
+    # compilers
+    depends_on("c", type="build")
+    depends_on("cxx", type="build")
+    depends_on("fortran", type="build")
+
+    # libraries
+    depends_on("bdw-gc threads=posix +cxx")
+    depends_on("blas")
+    depends_on("boost+regex")
+    depends_on("cddlib")
+    depends_on("eigen")
+    depends_on("singular-factory")
+    depends_on("fflas-ffpack")
+    depends_on("flint")
+    depends_on("fplll")
+    depends_on("frobby")
+    depends_on("gdbm")
+    depends_on("givaro")
+    depends_on("glpk")
+    depends_on("gmp")
+    depends_on("googletest")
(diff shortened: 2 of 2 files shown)


d-torrance:
@spackbot fix style

bernhardkaindl:
@d-torrance, just as an information from me trying to test-build the package on a Ubuntu 24.04 host:
```py -lgmpxx -lgmp -lgc -lgdbm -lgfortran -lm -lquadmath -lglpk: -lmps: -lmpfi: -lgc: -lgdbm:
  checking for dummy main to link with Fortran 77 libraries... unknown
...
/usr/bin/ld: cannot find -lglpk:: No such file or directory
/usr/bin/ld: cannot find -lmps:: No such file or directory
/usr/bin/ld: cannot find -lmpfi:: No such file or directory
/usr/bin/ld: cannot find -lgc:: No such file or directory
/usr/bin/ld: cannot find -lgdbm:: No such file or directory
```

d-torrance:
Thanks for taking a look at this!

This is a known issue: https://github.com/Macaulay2/M2/issues/4074

bernhardkaindl:
@d-torrance 
> Thanks for taking a look at this!
👍 

> This is a known issue: https://github.com/Macaulay2/M2/issues/4074

Sounds like an bug in the configure script.

spack also has support to regenerate the configure script, and may do so automatically.

Alternatively, I see the package also has a CMakeLists.txt file (but it is 2 years old, so a lot older than the configure.ac file, which is the source file for the generated configure script)

BTW, https://macaulay2.com/ says:
> Installation and Repositories
> (the current stable version is 1.26.06, released June 10, 2026)

Could the issue have been fixed meanwhile?
Do you rather like the old version because you already use it, or can you update?
https://github.com/Macaulay2/M2/releases/tag/release-1.26.06

(I think), If you convert the package to use Github as source location, you can also update the package more easily, in the future.

bernhardkaindl:
> Thanks for taking a look at this!
> This is a known issue: https://github.com/Macaulay2/M2/issues/4074

Hi, with spack, the idea is that spack works around known issues so the build just works.

d-torrance:
Found a workaround for https://github.com/Macaulay2/M2/issues/4074 thanks to Claude, and it's building on Ubuntu for me now.

@bernhardkaindl - Ready for another review -- thanks!

bernhardkaindl:
@d-torrance, thanks! I wonder about [allow-older-gfan.patch](https://github.com/spack/spack-packages/pull/5033/changes#diff-afbcdd3a5ad154470302c0e007d9ec4367013fe9b6f8c62561df8e9f9e283bfa):
```py
-		-- version 0.8 is required (the 0.6.2 series predates several
-		-- features used here and is no longer supported upstream)
```
As I read there that 0.6.2 is no longer supported upstream (ok if you use I I guess),
bit what about adding gfan 0.8 or newer as well as an option and requiring `gfan@0.8:`)?

Maybe related to the gfal version:
A depency in the tree of required buils is `mpsolve`.

Using %gcc@14, it failed to compile with `%gcc@14`, seemingly also wanting a newer gfan / gfanlib:
```py
  src/gfanlib_matrix.h: In function 'gfan::Matrix<typ> gfan::operator*(const Matrix<typ>&, const Matrix<typ>&)':
> src/gfanlib_matrix.h:123:18: error: 'const class gfan::Matrix<typ>' has no member named 'vectormultiply' [-Wtemplate-body]
    123 |         ret[i]=a.vectormultiply(b.column(i));
        |                  ^~~~~~~~~~~~~~
```

Note for building:
- Used %gcc@13 for the mpsolve build isse
- `fflas-ffpack` didn't find BLAS, seemingly hip-blas might have  been selected, so it needed ^blas=openblas
- Resulting build command for the deps: `spack install macaulay2 ^blas=openblas %gcc@13`

d-torrance:
Thanks for the review!

> As I read there that 0.6.2 is no longer supported upstream (ok if you use I I guess), bit what about adding gfan 0.8 or newer as well as an option and requiring `gfan@0.8:`)?

Everything still works with 0.6.2 for now -- although that will change at some point soon, hence the comment.  gfan 0.8 needs some work to compile on various sytems, which is why I added the patch. 

I'm not sure if that error message is from mpsolve.  It doesn't depend on gfan.

I'll look into fflas-ffpack

d-torrance:
What system had the failing build?  Is there a log somewhere?

bernhardkaindl:
@d-torrance, the reason for my `fflas-ffpack` failure seems to be `repos/spack_repo/builtin/packages/fflas_ffpack/package.py`:
```py
    depends_on("blas")
```
This matches what the homepage of it says (https://linbox-team.github.io/fflas-ffpack: "any implementation of a numerical BLAS library. We suggest using OpenBLAS on most of nowadays computers."), which is isn't true for 2.5.0.

The configure, INSTALL and the README.md files of 2.5.0 are very specific, README.md says it most precise: https://github.com/linbox-team/fflas-ffpack/tree/v2.5.0#requirements

> A BLAS library conforming to either the C or Fortran BLAS standard: OpenBLAS (recommended), or ATLAS. Make sure to use a single threaded version of the BLAS library.

For ATLAS, the README's instructions are:
- run `./configure --with-blas-libs="-L/usr/lib64/atlas -lsatlas"`.

But as the package.py does not have such support for defining atlas, it only suppors openblas.
Here is the relevant part of the log:
```
grep 'blas\>' spack-stage/spack-stage-fflas-ffpack-2.5.0-../spack-src/config.log 
...
| #include "fflas-ffpack/config-blas.h"
configure:18387: /home/bkaindl/spack/spack/opt/spack/linux-x86_64_v4/compiler-wrapper-1.1.0-gy7tlaoliz6borna6pp3uq7wa7x4uaqt/libexec/spack/gcc/g++ -o conftest    -I. -I.. -I. -I./fflas-ffpack -I/home/bkaindl/spack/spack/opt/spack/linux-x86_64_v4/givaro-4.2.2-xy5urwd3snlzsgkuvfutgqc73olrra43/include   conftest.cpp   -lopenblas -lpthread -lgfortran >&5
/usr/bin/ld: cannot find -lopenblas: No such file or directory
| #include "fflas-ffpack/config-blas.h"
BLAS_LIBS=' -lopenblas -lpthread -lgfortran'
LIBS='  -lopenblas -lpthread -lgfortran'
```

bernhardkaindl:
@d-torrance, So following the requirements as stated in https://github.com/linbox-team/fflas-ffpack/tree/v2.5.0#requirements, not implementing a variant for atlas, I guess this would be the blas dependency to use?
```diff
--- a/repos/spack_repo/builtin/packages/fflas_ffpack/package.py
+++ b/repos/spack_repo/builtin/packages/fflas_ffpack/package.py
@@ -27,6 +27,11 @@ class FflasFfpack(AutotoolsPackage):
     depends_on("cxx", type="build")
     depends_on("pkgconfig", type="build")
 
-    depends_on("blas")
+    # Ref: https://github.com/linbox-team/fflas-ffpack/tree/v2.5.0#requirements
+    # Quote:
+    # > A BLAS library conforming to either the C or Fortran BLAS standard:
+    # > OpenBLAS (recommended), or ATLAS.
+    # > Make sure to use a single threaded version of the BLAS library.
+    depends_on("openblas threads=none")
     depends_on("givaro")
     depends_on("lapack")
```

bernhardkaindl (approved):
Approving as the build works now for me. One furter comment:

@d-torrance, when building with %gcc@14 to %gcc@16, this happens with mpsolve:
```
  /bin/bash ../../libtool  --tag=CC   --mode=compile /home/bkaindl/spack/opt/spack/linux-skylake/compiler-wrapper-1.1.0-zr6dbk4ms64voxqzajisuxatkvcdiwyq/libexec/spack/gcc/gcc -DHAVE_CONFIG_H -I. -I../..  -I../../include -I../../include -D_REENTRANT -Wall -DMPS_USE_BUILTIN_COMPLEX  -DNICE_DEBUG -fcx-limited-range -fno-math-errno -fomit-frame-pointer -D_MPS_PRIVATE      -g -O2 -MT secsolve/secular-iteration.lo -MD -MP -MF $depbase.Tpo -c -o secsolve/secular-iteration.lo secsolve/secular-iteration.c &&\
  mv -f $depbase.Tpo $depbase.Plo
  ../../../src/libmps/monomial/tokenizer.l: In function 'mps_monomial_yacc_parser':
> ../../../src/libmps/monomial/tokenizer.l:51:3: error: implicit declaration of function 'yyparse'; did you mean 'pause'? [-Wimplicit-function-declaration]
```
This fixes it:
```
        if name == "cflags" and self.spec.satisfies("%gcc@14:"):
            flags.append("-Wno-error=implicit-function-declaration")
        return (flags, None, None)
```
These fixes can be done in separate PRs of course.

d-torrance:
Cool, thanks for the info.  The mpsolve issue is https://github.com/robol/MPSolve/issues/48, and the easy fix is to just delete a couple files before the build and let bison regenerate them (see https://github.com/Macaulay2/M2/commit/530dd6ac1cc5baa335a17bae92baa3c029fb9f58).  I'll open an mpsolve PR soon.

I've gotten fflas-ffpack to work with several different BLAS implementations before.  There have been a number of patches merged upstream since the 2.5.0 release that we might be able to backport.  I'll play around with it.

Just to confirm -- it was atlas specifically that was causing trouble?

bernhardkaindl:
> Just to confirm -- it was atlas specifically that was causing trouble?

@d-torrance 
No, I didn't build or install atlas in any way on that host:

It was `intel-oneapi-mkl`: I didn't know how to get the name of the virtual so I added a dummy function to print it to the log, and the log had: `blas: intel-oneapi-mkl`:
```py
    def configure_args(self):
        args = []
        print("blas: %s" % self.spec["blas"].name)
        return args
```

The specfic version and variants, from `spack spec macaulay2 %gcc@13` on that host are:
```py
grep intel macaulay2-fflas-ffpack-BLAS-problem.spec 
[+]      ^intel-oneapi-mkl@2026.1.0~cluster+envmods~gfortran~ilp64+shared build_system=generic mpi_family=none threads=none platform=linux os=ubuntu24.04 target=x86_64_v4 %c=rocmcc@7.2.3
[+]      ^intel-tbb@2023.1.0~ipo+shared+tm build_system=cmake build_type=Release cxxstd=default generator=make platform=linux os=ubuntu24.04 target=x86_64_v4 %c,cxx=gcc@12.4.0
```

d-torrance:
> It was `intel-oneapi-mkl`

Ok cool!  That was one of the several blas/lapack providers I tried out when I tested #6062

PR #5035, opened 2026-05-30
Kokkos: fix spurious cmake option

tpadioleau:
This PR fixes the spurious cmake option `-DKokkos_ENABLE_DEPRECATED_CODE=ON` for `kokkos@5:`.

We also:
- fix the deprecated variant for `kokkos@:4`, this changes its default value
- set the major version to 5 for the develop branch.

The current implementation of `deprecated_code` is: to enable code deprecated in the concretized major version.

### DIFF
--- a/repos/spack_repo/builtin/packages/kokkos/package.py
+++ b/repos/spack_repo/builtin/packages/kokkos/package.py
@@ -153,7 +153,6 @@ class Kokkos(CMakePackage, CudaPackage, ROCmPackage):
         "debug": [False, None, "Activate extra debug features - may increase compiletimes"],
         "debug_bounds_check": [False, None, "Use bounds checking - will increase runtime"],
         "debug_dualview_modify_check": [False, "@:4", "Debug check on dual views"],
-        "deprecated_code": [False, "@:4", "Whether to enable deprecated code"],
         "hpx_async_dispatch": [False, "@:4", "Whether HPX supports asynchronous dispath"],
         "tuning": [False, None, "Create bindings for tuning tools"],
         "tests": [False, None, "Build for tests"],
@@ -342,7 +341,6 @@ class Kokkos(CMakePackage, CudaPackage, ROCmPackage):
     variant(
         "deprecated_code",
         default=True,
-        when="@5:",
         description="Whether to enable deprecated code",
     )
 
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/kokkos/package.py
```diff
@@ -457,16 +455,14 @@ def cmake_args(self):
...
-                highest = max(v for v in self.versions if not v.isdevelop())
-                major_version = int(str(highest.up_to(1)))
-            else:
-                major_version = int(str(spec.version.up_to(1)))
-            options.append(
-                from_variant(f"Kokkos_ENABLE_DEPRECATED_CODE_{major_version}", "deprecated_code")
-            )
+        if spec.version == Version("develop"):
+            highest = max(v for v in self.versions if not v.isdevelop())
+            major_version = int(str(highest.up_to(1)))
```
cedricchevalier19: It is not new, but I wonder if just hardcode the Kokkos version is not as good. The algorithm is kind of broken during the time that a new major release is done upstream but not yet available in the recipe.

cedricchevalier19 (approved):

repos/spack_repo/builtin/packages/kokkos/package.py
tpadioleau: Do you think we can find a way to properly support the branch develop for every commit ?

JBludau: If we want that, we can just always use the highest +1 and tolerate having warnings about unused variables in CMake.

repos/spack_repo/builtin/packages/kokkos/package.py
```diff
@@ -153,7 +153,6 @@ class Kokkos(CMakePackage, CudaPackage, ROCmPackage):
         "debug": [False, None, "Activate extra debug features - may increase compiletimes"],
         "debug_bounds_check": [False, None, "Use bounds checking - will increase runtime"],
         "debug_dualview_modify_check": [False, "@:4", "Debug check on dual views"],
-        "deprecated_code": [False, "@:4", "Whether to enable deprecated code"],
```
rbberger, line 157: One reason we kept this `False` was to not change behavior. Changing the default might surprise people building older versions.

tpadioleau: If I am not mistaken, it was `False` but not working as expected as it were setting `Kokkos_ENABLE_DEPRECATED_CODE=OFF` which was ignored. So I think the behavior was actually closer to `Kokkos_ENABLE_DEPRECATED_CODE_4=ON`.

repos/spack_repo/builtin/packages/kokkos/package.py
tpadioleau: The value is now hardcoded.

rbberger (dismissed):

tpadioleau:
A warning though, I couldn't find any package that requires `+deprecated_code`. When a new Kokkos release removes/reworks some public symbols I am afraid it will break packages in the Spack CI.

tpadioleau:
@spackbot run pipeline

tpadioleau:
@spackbot run pipeline

JBludau (commented):
Hmm, thinking about this some more, I feel like we should make the default `True` to allow a large range of compatibility per default.

And then we should think about how we want to manage the transition to another major version to be prepared for when it actually comes. We could extend it to enable both the current and the last until the stuff from the last is removed

rbberger:
Please update the MR description/title to what it now actually does.

tpadioleau:
> Please update the MR description/title to what it now actually does.

Is it better ?

tpadioleau:
Same as #6051, needs more discussion.

PR #5044, opened 2026-06-01
py-mypy: add v2.1.0 and py-ast-serialize: add new package

manuelakuhn:
https://github.com/python/mypy/releases/tag/v2.1.0

and add dependeny `py-ast-serialize`
https://github.com/mypyc/ast_serialize/releases/tag/v0.5.0

### DIFF
--- a/repos/spack_repo/builtin/packages/py_mypy/package.py
+++ b/repos/spack_repo/builtin/packages/py_mypy/package.py
@@ -18,6 +18,7 @@ class PyMypy(PythonPackage):
 
     license("MIT AND PSF-2.0", checked_by="tgamblin")
 
+    version("2.1.0", sha256="81e76ad12c2d804512e9b13240d1588316531bfba07558286078bfbce9613633")
     version("1.19.0", sha256="f6b874ca77f733222641e5c46e4711648c4037ea13646fd0cdc814c2eaec2528")
     version("1.16.0", sha256="84b94283f817e2aa6350a14b4a8fb2a35a53c286f97c9d30f53b63620e7af8ab")
     version("1.15.0", sha256="404534629d51d3efea5c800ee7c42b72a6554d6c400e6a79eafe15d11341fd43")
@@ -64,6 +65,7 @@ class PyMypy(PythonPackage):
     version("0.740", sha256="48c8bc99380575deb39f5d3400ebb6a8a1cb5cc669bbba4d3bb30f904e0a0e7d")
 
     with default_args(type="build"):
+        depends_on("py-setuptools@77.0.3:", when="@1.20:")
         depends_on("py-setuptools@75.1.0:", when="@1.14:")
         depends_on("py-setuptools@40.6.2:", when="@0.790:")
         depends_on("py-setuptools")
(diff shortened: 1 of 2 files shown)


repos/spack_repo/builtin/packages/py_mypy/package.py
```diff
@@ -74,21 +76,29 @@ class PyMypy(PythonPackage):
...
+        depends_on("python@3.10:", when="@1.20:")
         depends_on("python@3.9:", when="@1.15:")
         depends_on("python@3.8:", when="@1.5:")
         depends_on("python@3.7:", when="@0.981:")
-        depends_on("py-typing-extensions@4.6:", when="@1.11:")
-        depends_on("py-typing-extensions@4.1:", when="@1.5:")
-        depends_on("py-typing-extensions@3.10:", when="@0.930:")
-        depends_on("py-typing-extensions@3.7.4:", when="@0.700:")
+
+        depends_on("py-typing-extensions@4.14:", when="@1.20.2: ^python@3.15:")
```
adamjstewart, line 84: Any idea what the cause of this is? If this is an incompatibility between older typing-extensions and newer Python, we could instead mark this in the typing-extensions package.

adamjstewart (approved):
Mypy changes look good, just one question on motivation

repos/spack_repo/builtin/packages/py_mypy/package.py
manuelakuhn: According to the commit message of the changes it is be because of:
>Update `typing-extensions` to `>=4.14.0` for Python `>=3.15` due to
the removal of `typing.no_type_check_decorator`

see https://github.com/python/mypy/commit/ebaa73aee1b8faf9a4b8b230067a446f32908773

adamjstewart: Hmm, that PR doesn't really clarify if this is a situation where typing-extensions is using a feature that was removed in Python 3.15 (and thus it's typing-extensions that should have a conflict) or if it's something more complex.

manuelakuhn: I think it is because of typing-extensions that does not build with python 3.15, see https://bugzilla.redhat.com/show_bug.cgi?id=2416981

Once python 3.15 is really out and we can properly test it we can add a conflict in `py-typing-extensions`. But for now (without the possibility to test anything) I would just also leave it in `py-mypy`.

adamjstewart: Alright, I'm fine with whatever. Truth be told, I don't even use mypy anymore, fully converted to ty.

repos/spack_repo/builtin/packages/py_mypy/package.py
```diff
@@ -74,21 +76,29 @@ class PyMypy(PythonPackage):
...
+        with when("^python@:3.14"):
+            depends_on("py-typing-extensions@4.6:", when="@1.11:")
+            depends_on("py-typing-extensions@4.1:", when="@1.5:")
+            depends_on("py-typing-extensions@3.10:", when="@0.930:")
+            depends_on("py-typing-extensions@3.7.4:", when="@0.700:")
```
adamjstewart, line 89: Just noticed that this introduces a bug because the py-typing-extensions dependency is missing when `@:1.20.1 ^python@3.15:`. Would this work?
```python
depends_on("py-typing-extensions@4.14:", when="@0.700: ^python@3.15:")
```

manuelakuhn: Sorry for the delay. That makes sense to me, feel free to open a PR and ping me.

adamjstewart: https://github.com/spack/spack-packages/pull/5474

PR #5048, opened 2026-06-01
kokkos: remove Kokkos' test

JBludau:
The relevant discussion is in https://github.com/spack/spack-packages/pull/5017.
TL;DR we do not think the test is actually helping (see discussion), but it would need changes to be useful (we tried to use it in Kokkos' CI). So instead of fixing it, we just remove it and find a better way to test the spack builds in our CI


tpadioleau (approved):
Fine for me, what do other maintainers think about it ?

nmm0 (approved):

JBludau:
@rbberger any objections to just removing the test?

rbberger:
@JBludau I don't use it for anything. This was for Kokkos only. So fine with me.

rbberger (approved):

PR #5057, opened 2026-06-02
amdlibflame: add dual-build support for for shared and static libraries

amd-toolchain-support:
## Description

Enables building both shared and static libraries simultaneously for amdlibflame CMake builds (@4.2:), addressing the need for dual-library installations to address https://github.com/spack/spack-packages/issues/382.

## Changes

**Variant Configuration:**
- Override `static` variant from `LibflameBase`: `default=False` for CMake builds (@4.2:), making shared-only the default
- Explicitly document that this overrides the base class variant definition
- Add conflict `~shared~static` to require at least one library type
- Fix `logging`/`tracing` variants: boolean `False` instead of string `"False"`

**Dual-Build Strategy:**
- Implement two-pass build in `CMakeBuilder.install()` when both `+shared +static` requested
- First pass: shared libraries in standard build directory
- Second pass: static libraries in separate `build-static` directory
- Both install to same prefix (`.so` and `.a` coexist without conflict)

**Library Discovery:**
- Update `libs` and `lapack_libs` properties with clarified behavior documentation
- When only one type is built: returns that type (e.g., `~shared +static` returns static libraries)
- When both types are built: prefers shared libraries
- Maintains deterministic behavior for dependent packages

## Usage

```bash
# Default: shared only
spack install amdlibflame

# Both library types
spack install amdlibflame +shared +static

# Static only
spack install amdlibflame ~shared +static
```

### DIFF
--- a/repos/spack_repo/builtin/packages/amdlibflame/package.py
+++ b/repos/spack_repo/builtin/packages/amdlibflame/package.py
@@ -9,6 +9,7 @@ from spack_repo.builtin.packages.libflame.package import LibflameBase
 
 from spack.package import *
 
+import os
 
 class Amdlibflame(CMakePackage, LibflameBase):
     """libFLAME (AMD Optimized version) is a portable library for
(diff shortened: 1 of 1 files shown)


amd-toolchain-support:
@spackbot fix style

kwryankrattiger (approved):
Lgtm.

PR #5058, opened 2026-06-02
kokkos-kernels: make version of kernels and core always match

JBludau:
The idea is that we would not need to update this copy-pasta dish for every release


JBludau:
ok, here is my two cents ... it looks like this can not be done in spack right now and we should just go and repeat the versions within the Kokkos ecosystem.

PR #5064, opened 2026-06-02
paraview: drop deprecated versions

kwryankrattiger:
Drop all of the deprecated versions of paraview except for the minor versions associated with non-deprecated versions.

Delete all of the unneeded logic used for compatibility with older versions of ParaView.

### DIFF
--- a/repos/spack_repo/builtin/packages/paraview/package.py
+++ b/repos/spack_repo/builtin/packages/paraview/package.py
@@ -100,7 +100,7 @@ class Paraview(CMakePackage, CudaPackage, ROCmPackage):
     variant("fortran", default=False, description="Enable Fortran support")
     variant("mpi", default=True, description="Enable MPI support")
     variant("qt", default=False, description="Enable Qt (gui) support")
-    variant("opengl2", default=True, description="Enable OpenGL2 backend", when="@5:5")
+    variant("opengl2", default=True, description="Enable OpenGL2 backend", when="@5")
     variant("osmesa_fallback", default=False, description="Enable OpenGL2 backend", when="@6:")
     variant("x", default=True, description="Enable X11 support")
     variant("examples", default=False, description="Build examples")
@@ -145,7 +145,7 @@ class Paraview(CMakePackage, CudaPackage, ROCmPackage):
     variant(
         "use_vtkm",
         default="default",
-        when="@5:5",
+        when="@5",
         multi=False,
         values=("default", "on", "off"),
         description="Build VTK-m with ParaView."
@@ -155,8 +155,8 @@ class Paraview(CMakePackage, CudaPackage, ROCmPackage):
 
     conflicts("~hdf5", when="+visitbridge")
     conflicts("+fides", when="~adios2", msg="Fides needs ADIOS2")
-    conflicts("+fides", when="@5:5 use_vtkm=off", msg="Fides needs VTK-m")
-    conflicts("+fides", when="@5:5 use_vtkm=default", msg="Fides needs VTK-m")
+    conflicts("+fides", when="@5 use_vtkm=off", msg="Fides needs VTK-m")
+    conflicts("+fides", when="@5 use_vtkm=default", msg="Fides needs VTK-m")
     conflicts("+openpmd", when="~adios2 ~hdf5", msg="openPMD needs ADIOS2 and/or HDF5")
     conflicts("~shared", when="+cuda")
     conflicts("+cuda", when="use_vtkm=off")
@@ -164,7 +164,7 @@ class Paraview(CMakePackage, CudaPackage, ROCmPackage):
     conflicts("+rocm", when="use_vtkm=off")
     # Legacy rendering dropped in 5.5
     # See commit: https://gitlab.kitware.com/paraview/paraview/-/commit/798d328c
-    conflicts("~opengl2", when="@5:5")
+    conflicts("~opengl2", when="@5")
 
     depends_on("c", type="build")  # generated
     depends_on("cxx", type="build")  # generated
@@ -206,18 +206,18 @@ class Paraview(CMakePackage, CudaPackage, ROCmPackage):
         depends_on("libx11")
         depends_on("libxcursor")
         # When Qt and X are enabled, GLX is required in the runtime
-        requires("^[virtuals=gl] glx", when="@:5")
+        requires("^[virtuals=gl] glx", when="@5")
         depends_on("glx", when="@6:", type=("run"))
 
-    # ParaView@:5 support Qt5 and requires a GL provider to be known at
+    # ParaView@5 support Qt5 and requires a GL provider to be known at
     # build/link time.
-    with when("@:5"):
+    with when("@5"):
         with when("+qt"):
             # https://discourse.paraview.org/t/paraview-5-9-and-minimum-recommended-qt-version/5333
-            depends_on("qt@5.12:5", when="@5:5.13")
+            depends_on("qt@5.12:5", when="@5")
             depends_on("qt+sql")
-            depends_on("qt+opengl", when="@5:5 +opengl2")
-            depends_on("qt~opengl", when="@5:5 ~opengl2")
+            depends_on("qt+opengl", when="@5 +opengl2")
+            depends_on("qt~opengl", when="@5 ~opengl2")
             # Headless rendering not supported with Qt
             conflicts("osmesa")
             conflicts("egl")
@@ -328,7 +328,7 @@ class Paraview(CMakePackage, CudaPackage, ROCmPackage):
     depends_on("netcdf-c+parallel-netcdf", when="+mpi platform=darwin")
     depends_on("netcdf-c+parallel-netcdf", when="+mpi platform=freebsd")
     depends_on("netcdf-c+parallel-netcdf", when="+mpi platform=linux")
-    depends_on("netcdf-c@:4.9.2", when="@5:5.13")
+    depends_on("netcdf-c@:4.9.2", when="@5")
     depends_on("pegtl@2.8.3")
     depends_on("protobuf@3.4:")
     # protobuf requires newer abseil-cpp, which in turn requires C++14,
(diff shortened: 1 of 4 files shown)


repos/spack_repo/builtin/packages/paraview/package.py
```diff
@@ -422,54 +369,24 @@ class Paraview(CMakePackage, CudaPackage, ROCmPackage):
...
-
-    # intel oneapi doesn't compile some code in catalyst
-    patch("catalyst-etc_oneapi_fix.patch", when="@5.10.0:5.10.1%oneapi")
+    patch("vtk-adios2-module-no-kit.patch", when="@:5.11")
 
     # Patch for paraview 5.8: ^hdf5@1.13.2:
     # Even with ~hdf5, hdf5 is part of the dependency tree due to netcdf-c
     # https://gitlab.kitware.com/vtk/vtk/-/merge_requests/9690
-    patch("vtk-xdmf2-hdf51.13.1.patch", when="@5.8:5.10")
     patch("vtk-xdmf2-hdf51.13.2.patch", when="@5.8:5.11.0")
```
johnwparent: These boundaries need cleanup

repos/spack_repo/builtin/packages/paraview/package.py
```diff
@@ -422,54 +369,24 @@ class Paraview(CMakePackage, CudaPackage, ROCmPackage):
...
     # https://gitlab.kitware.com/vtk/vtk/-/merge_requests/8653
-    patch("vtk-adios2-module-no-kit.patch", when="@5.8:5.11")
-
-    # Patch for paraview 5.9.0%xl_r
-    # https://gitlab.kitware.com/vtk/vtk/-/merge_requests/7591
-    patch("xlc-compilation-pv590.patch", when="@5.9.0%xl_r")
-
-    # intel oneapi doesn't compile some code in catalyst
-    patch("catalyst-etc_oneapi_fix.patch", when="@5.10.0:5.10.1%oneapi")
+    patch("vtk-adios2-module-no-kit.patch", when="@:5.11")
```
johnwparent: This can just be @5.11 now right?

repos/spack_repo/builtin/packages/paraview/package.py
```diff
@@ -639,66 +537,35 @@ def variant_bool(feature, on="ON", off="OFF"):
...
-                    "-DVTK_USE_SYSTEM_GL2PS:BOOL=OFF",
-                    "-DVTK_USE_SYSTEM_ICET:BOOL=OFF",
-                    "-DVTK_USE_SYSTEM_LIBHARU:BOOL=OFF",
-                    "-DVTK_USE_SYSTEM_NETCDFCPP:BOOL=OFF",
-                    "-DVTK_USE_SYSTEM_UTF8:BOOL=OFF",
-                    "-DVTK_USE_SYSTEM_XDMF2:BOOL=OFF",
-                    "-DVTK_USE_SYSTEM_XDMF3:BOOL=OFF",
-                ]
-            )
+        cmake_args.extend(
```
johnwparent: This looks like two back to back calls to extend, should just be one

repos/spack_repo/builtin/packages/paraview/package.py
kwryankrattiger: ah yeah, I was going to include 5.10 initially but decided to drop it.

kwryankrattiger: ah, some of this got messed up during rebase, lots of places that were reverted under me.

repos/spack_repo/builtin/packages/paraview/package.py
```diff
@@ -220,7 +178,7 @@ class Paraview(CMakePackage, CudaPackage, ROCmPackage):
     # VTK < 8.2.1 can't handle Python 3.8
     # This affects Paraview <= 5.7 (VTK 8.2.0)
     # https://gitlab.kitware.com/vtk/vtk/-/issues/17670
```
vicentebolea:
```suggestion
```

repos/spack_repo/builtin/packages/paraview/package.py
```diff
@@ -258,12 +216,11 @@ class Paraview(CMakePackage, CudaPackage, ROCmPackage):
...
         with when("+qt"):
-            depends_on("qt@:4", when="@:5.2.0")
             # https://discourse.paraview.org/t/paraview-5-9-and-minimum-recommended-qt-version/5333
-            depends_on("qt@5.12:5", when="@5.9:5.13")
+            depends_on("qt@5.12:5", when="@5:5.13")
             depends_on("qt+sql")
-            depends_on("qt+opengl", when="@5.3.0:5 +opengl2")
-            depends_on("qt~opengl", when="@5.3.0:5 ~opengl2")
+            depends_on("qt+opengl", when="@5:5 +opengl2")
+            depends_on("qt~opengl", when="@5:5 ~opengl2")
```
vicentebolea: would this be equivalent to `@5`

vicentebolea (commented):
I appears that spack-packages does not explicitly test `paraview@5.12`. There could be a chance that these changes might break that spec root.

kwryankrattiger:
> There could be a chance that these changes might break that spec root.

What do you mean by this? The spec hash will change, but that is not an issue. I also have built and installed all of the non-deprecated paraviews in canonical mode this change to make sure that was all still working, so I have pretty high confidence this isn't breaking anything.

repos/spack_repo/builtin/packages/paraview/package.py
kwryankrattiger: This was intentional just to be more clear that it applied to all versions of 5.*. It is a style thing, I prefer it for these short specs.

vicentebolea (approved):
The changes look good paraview-wise. Feels good!

olesenm:
reasonable enough to drop paraview from the openfoam package. We mostly want to be avoiding the paraview plugins route for OpenFOAM in the meantime.  The support for in situ visualization and runtime generation of images does also get removed with this, but that can/should be addressed sometime later.

LGTM

repos/spack_repo/builtin/packages/paraview/package.py
```diff
@@ -377,20 +328,14 @@ class Paraview(CMakePackage, CudaPackage, ROCmPackage):
     depends_on("netcdf-c+parallel-netcdf", when="+mpi platform=darwin")
     depends_on("netcdf-c+parallel-netcdf", when="+mpi platform=freebsd")
     depends_on("netcdf-c+parallel-netcdf", when="+mpi platform=linux")
-    depends_on("netcdf-c@:4.9.2", when="@:5.13")
+    depends_on("netcdf-c@:4.9.2", when="@5:5.13")
```
johnwparent: This can be `@5:5` right? (personally I am not a fan, but I'm not maintaining this package 😜 )  
Seems like the intent was to guard against 6, unless it was also proofing against a .14 patch release?

repos/spack_repo/builtin/packages/paraview/package.py
```diff
@@ -620,6 +511,22 @@ def variant_bool(feature, on="ON", off="OFF"):
...
+            self.define_from_variant("PARAVIEW_BUILD_WITH_KITS", "kits"),
+            self.define_from_variant("PARAVIEW_BUILD_PAGOSA_ADAPTOR", "pagosa"),
+            self.define_from_variant("PARAVIEW_BUILD_PLUGIN_EyeDomeLighting", "eyedomelighting"),
+            self.define_from_variant("PARAVIEW_PLUGIN_ENABLE_pvNVIDIAIndeX", "nvindex"),
+            "-DVTK_MODULE_USE_EXTERNAL_ParaView_vtkcatalyst:BOOL=OFF",
+            "-DVTK_MODULE_USE_EXTERNAL_VTK_ioss:BOOL=OFF",
+            "-DVTK_MODULE_USE_EXTERNAL_VTK_exprtk:BOOL=OFF",
+            "-DVTK_MODULE_USE_EXTERNAL_VTK_fmt:BOOL=OFF",
+            "-DPARAVIEW_BUILD_EDITION:STRING=%s" % spec.variants["build_edition"].value.upper(),
+            "-DPARAVIEW_USE_QT:BOOL=%s" % variant_bool("+qt"),
```
johnwparent: Let's use this opportunity to be rid of printf strings

repos/spack_repo/builtin/packages/paraview/package.py
```diff
@@ -756,8 +598,8 @@ def variant_bool(feature, on="ON", off="OFF"):
 
         cmake_args.append("-DPARAVIEW_BUILD_SHARED_LIBS:BOOL=%s" % variant_bool("+shared"))
 
-        # VTK-m added to ParaView in 5.3.0 and up
-        if spec.satisfies("@5.3.0:5") and spec.variants["use_vtkm"].value != "default":
+        # VTK-m used in ParaView in 5.x
+        if spec.satisfies("@:5") and spec.variants["use_vtkm"].value != "default":
```
johnwparent: Should also be `@5:5` for consistency

repos/spack_repo/builtin/packages/paraview/package.py
```diff
@@ -773,12 +615,7 @@ def variant_bool(feature, on="ON", off="OFF"):
...
-        if spec.satisfies("@5.8:"):
-            cmake_args.append("-DPARAVIEW_USE_CUDA:BOOL=%s" % variant_bool("+cuda"))
-        elif spec.satisfies("@5.7:"):
-            cmake_args.append("-DVTK_USE_CUDA:BOOL=%s" % variant_bool("+cuda"))
-        else:
-            cmake_args.append("-DVTKm_ENABLE_CUDA:BOOL=%s" % variant_bool("+cuda"))
+        cmake_args.append("-DPARAVIEW_USE_CUDA:BOOL=%s" % variant_bool("+cuda"))
 
         # VTK-m expects cuda_arch to be the arch name vs. the arch version.
         if spec.satisfies("@:5 +cuda"):
```
johnwparent: Should also be `@5:5`

repos/spack_repo/builtin/packages/paraview/package.py
```diff
@@ -620,6 +511,22 @@ def variant_bool(feature, on="ON", off="OFF"):
...
+            "-DPARAVIEW_USE_QT:BOOL=%s" % variant_bool("+qt"),
+            "-DPARAVIEW_BUILD_WITH_EXTERNAL=ON",
+            self.define_from_variant("PARAVIEW_ENABLE_EXAMPLES", "examples"),
+            self.define("VTK_MODULE_USE_EXTERNAL_ParaView_cgns", False),
+            self.define("VTK_MODULE_USE_EXTERNAL_VTK_gl2ps", False),
+            self.define("VTK_MODULE_USE_EXTERNAL_VTK_libharu", False),
+            self.define("VTK_MODULE_USE_EXTERNAL_VTK_utf8", False),
         ]
 
         if spec.satisfies("@:5"):
```
johnwparent:
```suggestion
        if spec.satisfies("@5:5"):
```

repos/spack_repo/builtin/packages/paraview/package.py
johnwparent: I think `@5:5` is less clear, if you're familiar with Spack spec syntax, this makes you pause and go "wait why is this a range" and if you're less familiar, it looks like the same as `@5` but then you go to double check the syntax in the docs and come away more confused because the docs say this should be `@5`.

I don't actually have a strong opinion though. Not blocking on this.

johnwparent (dismissed):
There could be more cleanup of some of the versions ranges for consistency (and there are still some printf style strings I'd love to see purged) but none are actually worth blocking over.

repos/spack_repo/builtin/packages/paraview/package.py
kwryankrattiger: ah yeah, there will never be a 5.14 release, ParaView doesn't look back like that.

repos/spack_repo/builtin/packages/paraview/package.py
kwryankrattiger: The biggest offender is variant_bool...which is now gone 🪠

repos/spack_repo/builtin/packages/paraview/package.py
kwryankrattiger: I dropped the 5:5 syntax for just 5. If I am alone in liking that then maybe I am wrong :p

johnwparent (dismissed):

repos/spack_repo/builtin/packages/paraview/package.py
```diff
@@ -420,56 +364,26 @@ class Paraview(CMakePackage, CudaPackage, ROCmPackage):
...
     # https://gitlab.kitware.com/vtk/vtk/-/merge_requests/8653
-    patch("vtk-adios2-module-no-kit.patch", when="@5.8:5.11")
-
-    # Patch for paraview 5.9.0%xl_r
-    # https://gitlab.kitware.com/vtk/vtk/-/merge_requests/7591
-    patch("xlc-compilation-pv590.patch", when="@5.9.0%xl_r")
-
-    # intel oneapi doesn't compile some code in catalyst
-    patch("catalyst-etc_oneapi_fix.patch", when="@5.10.0:5.10.1%oneapi")
+    patch("vtk-adios2-module-no-kit.patch", when="@5:5.11")
```
johnwparent, line 371: The first `5` is redundant here

repos/spack_repo/builtin/packages/paraview/package.py
```diff
@@ -420,56 +364,26 @@ class Paraview(CMakePackage, CudaPackage, ROCmPackage):
...
-
-    # Broken H5Part with external parallel HDF5
-    patch("h5part-parallel.patch", when="@5.7.0:5.7")
-
-    # Broken downstream FindMPI
-    patch("vtkm-findmpi-downstream.patch", when="@5.9.0")
-
-    # Include limits header wherever needed to fix compilation with GCC 11
-    patch("paraview-gcc11-limits.patch", when="@5.8:5.9 %gcc@11.1.0:")
+    depends_on("cuda@:11", when="@5:5.12 +cuda")
```
johnwparent, line 367: First `5` is redundant

johnwparent (approved):
Two very tiny nits for consistency, LGTM

repos/spack_repo/builtin/packages/paraview/package.py
kwryankrattiger: Maybe next time, want this for the release

PR #5066, opened 2026-06-02
chai: add cxxstd variant

adayton1:
Allow specifying the c++ standard to build with.


davidbeckingsale (approved):

repos/spack_repo/builtin/packages/chai/package.py
```diff
@@ -215,6 +224,8 @@ class Chai(CachedCMakePackage, CudaPackage, ROCmPackage):
             )
 
     with when("+raja"):
+        for std in ("11", "14", "17", "20"):
+            depends_on(f"raja cxxstd={std}", when=f"cxxstd={std}")
```
adayton1: Is it a good idea to make raja build with the same standard? I think it makes sense, but I don't know if this is artificially constraining.

kshea21: Thoughts? @davidbeckingsale @kab163

bernhardkaindl:
@adayton1, @davidbeckingsale, kshea21 asked for your input on:
```py
       for std in ("11", "14", "17", "20"):
            depends_on(f"raja cxxstd={std}", when=f"cxxstd={std}")
```
She wrote in https://github.com/spack/spack-packages/pull/5066#discussion_r3364143330
> Is it a good idea to make raja build with the same standard? I think it makes sense, but I don't know if this is artificially constraining.

I guess it is safer to make the C++ standards match, between chai and raja, but the question was asked, so I'd like to send you this reminder to comment on this possible change to be merged.

When this question doesn't get a comment, the PR might be merged at some point.

repos/spack_repo/builtin/packages/chai/package.py
adayton1:
```suggestion
```

repos/spack_repo/builtin/packages/chai/package.py
```diff
@@ -152,6 +152,12 @@ class Chai(CachedCMakePackage, CudaPackage, ROCmPackage):
     variant("examples", default=True, description="Build examples.")
     variant("openmp", default=False, description="Build using OpenMP")
     variant("disable_rm", default=False, description="Make ManagedArray a thin wrapper")
+    variant(
+        "cxxstd",
+        default="17",
+        values=("11", "14", "17", "20"),
```
adayton1:
```suggestion
        values=("11", "14", "17", "20", "23"),
```

repos/spack_repo/builtin/packages/chai/package.py
adayton1: I'm opting for greater flexibility for now.

alecbcs (approved):
Looks good to me. Thanks @adayton1

PR #5078, opened 2026-06-03
kokkos-comm: new package

cedricchevalier19:
Introducing Kokkos Comm, experimental explicit communication interfaces for distributed applications using the Kokkos C++ Performance Portability Programming ecosystem.

### DIFF
--- a/repos/spack_repo/builtin/packages/kokkos_comm/package.py
+++ b/repos/spack_repo/builtin/packages/kokkos_comm/package.py
@@ -12,11 +12,13 @@ class KokkosComm(CMakePackage):
 
     homepage = "https://github.com/kokkos/kokkos-comm"
     url = "https://github.com/kokkos/kokkos-comm/archive/refs/tags/v0.1.0.tar.gz"
+    git = "https://github.com/kokkos/kokkos-comm.git"
 
     maintainers("cedricchevalier19", "dssgabriel", "cwpearson")
 
     license("Apache-2.0 WITH LLVM-exception", checked_by="cedricchevalier19")
 
+    version("develop", branch="develop")
     version("0.1.0", sha256="59f4b953a795adb62f306e0861c7e69ee60c8cd2a6f1bd58eb4623b9ab774d45")
 
     variant("mpi", description="Enable MPI backend", default=True)
--- a/repos/spack_repo/builtin/packages/kokkos_comm/package.py
+++ b/repos/spack_repo/builtin/packages/kokkos_comm/package.py
@@ -41,8 +41,8 @@ class KokkosComm(CMakePackage):
 
     def cmake_args(self):
         args = [
-            self.define("KokkosComm_ENABLE_MPI", "mpi"),
-            self.define("KokkosComm_ENABLE_NCCL", "nccl"),
+            self.define_from_variant("KokkosComm_ENABLE_MPI", "mpi"),
+            self.define_from_variant("KokkosComm_ENABLE_NCCL", "nccl"),
             self.define("KokkosComm_ENABLE_TESTS", True),  # To be removed, for test
         ]
 
--- a/repos/spack_repo/builtin/packages/kokkos_comm/package.py
+++ b/repos/spack_repo/builtin/packages/kokkos_comm/package.py
@@ -20,18 +20,27 @@ class KokkosComm(CMakePackage):
     version("0.1.0", sha256="59f4b953a795adb62f306e0861c7e69ee60c8cd2a6f1bd58eb4623b9ab774d45")
 
     variant("mpi", description="Enable MPI backend", default=True)
+    variant("nccl", description="Enable NCCL backend", default=False)
 
+    # Mandatory dependencies
     depends_on("cxx")
     depends_on("c", when="+mpi")
     depends_on("cmake@3.22:3", type="build")
 
     depends_on("kokkos@4.7:")
 
+    # MPI-backend dependencies
     depends_on("mpi@3:", when="+mpi")
-    
+
+    # NCCL-backend dependencies
+    depends_on("kokkos +cuda", when="+nccl")
+    depends_on("cuda", when="+nccl")
+    depends_on("nccl@2.20:", when="+nccl")
+
     def cmake_args(self):
         args = [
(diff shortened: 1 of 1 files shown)


cedricchevalier19:
@spackbot style --fix

repos/spack_repo/builtin/packages/kokkos_comm/package.py
```diff
@@ -0,0 +1,56 @@
...
+    version("develop", branch="develop")
+    version("0.1.0", sha256="59f4b953a795adb62f306e0861c7e69ee60c8cd2a6f1bd58eb4623b9ab774d45")
+
+    variant("mpi", description="Enable MPI backend", default=True)
+    variant("nccl", description="Enable NCCL backend", default=False)
+
+    # Mandatory dependencies
+    depends_on("cxx")
+    depends_on("c", when="+mpi")
+    depends_on("cmake@3.22:3", type="build")
```
dssgabriel: Kokkos Comm requires CMake 3.25 since [kokkos-comm#212](https://github.com/kokkos/kokkos-comm/pull/212).
Any CMake 4+ works correctly as well, so we can afford not to put a bound on the maximum supported version.
```suggestion
    depends_on("cmake@3.25:", type="build")
```

repos/spack_repo/builtin/packages/kokkos_comm/package.py
```diff
@@ -0,0 +1,56 @@
...
+    depends_on("mpi@3:", when="+mpi")
+
+    # NCCL-backend dependencies
+    depends_on("kokkos +cuda", when="+nccl")
+    depends_on("cuda", when="+nccl")
+    depends_on("nccl@2.20:", when="+nccl")
+
+    def cmake_args(self):
+        args = [
+            self.define("KokkosComm_ENABLE_MPI", "mpi"),
```
rbberger: `define_from_variant`

cedricchevalier19: Yes, indeed.

repos/spack_repo/builtin/packages/kokkos_comm/package.py
```diff
@@ -0,0 +1,56 @@
...
+
+    version("develop", branch="develop")
+    version("0.1.0", sha256="59f4b953a795adb62f306e0861c7e69ee60c8cd2a6f1bd58eb4623b9ab774d45")
+
+    variant("mpi", description="Enable MPI backend", default=True)
+    variant("nccl", description="Enable NCCL backend", default=False)
+
+    # Mandatory dependencies
+    depends_on("cxx")
+    depends_on("c", when="+mpi")
```
tpadioleau, line 29: I am curious to know the reason of adding C language when enabling MPI. Do we enable it in cmake ?

rbberger: I'm not seeing anything in the CMake that would require this. It explicitly only adds `MPI::MPI_CXX`, and `find_package(MPI)` uses the enabled languages (`CXX`) by default.

repos/spack_repo/builtin/packages/kokkos_comm/package.py
```diff
@@ -0,0 +1,56 @@
...
+    license("Apache-2.0 WITH LLVM-exception", checked_by="cedricchevalier19")
+
+    version("develop", branch="develop")
+    version("0.1.0", sha256="59f4b953a795adb62f306e0861c7e69ee60c8cd2a6f1bd58eb4623b9ab774d45")
+
+    variant("mpi", description="Enable MPI backend", default=True)
+    variant("nccl", description="Enable NCCL backend", default=False)
+
+    # Mandatory dependencies
+    depends_on("cxx")
```
rbberger, line 28: `c` and `cxx` should by `type="build"`.

repos/spack_repo/builtin/packages/kokkos_comm/package.py
cedricchevalier19: In fact it was in googletest, and the messages where mixed. If we do not enable unit tests, we do not need the "C" compiler.

repos/spack_repo/builtin/packages/kokkos_comm/package.py
cedricchevalier19: Thanks, I follow your advice, but it is a little weird to not consider the project to have a link dependency towards the lib C++.

cedricchevalier19:
@spackbot fix style

repos/spack_repo/builtin/packages/kokkos_comm/package.py
tpadioleau: Ah I see, actually if you can `find_package(GTest)`, you won't have this issue. GoogleTest does not need a C compiler but they don't want to update the `project` command in their CMakeLists.txt.

cedricchevalier19: In fact there is a comment that the C compiler is usefull for finding Threads. Currently we get GTest by fetch_content but this part will not be shipped/compiled with Spack.

repos/spack_repo/builtin/packages/kokkos_comm/package.py
```diff
@@ -0,0 +1,57 @@
...
+    depends_on("cuda", when="+nccl")
+    depends_on("nccl@2.20:", when="+nccl")
+
+    def cmake_args(self):
+        args = [
+            self.define_from_variant("KokkosComm_ENABLE_MPI", "mpi"),
+            self.define_from_variant("KokkosComm_ENABLE_NCCL", "nccl"),
+        ]
+
+        if self.spec.satisfies("^kokkos+rocm") and not (
```
dssgabriel, line 50: Just a drive-by question: should we forbid `^kokkos+rocm` when requesting `kokkoscomm+nccl`?

cedricchevalier19: I think it is not necessary, `+nccl` implies `kokkos +cuda` and there is a conflict with `+rocm` already. And I do not want having to maintain conflicts between all Kokkos backends.

PR #5082, opened 2026-06-03
scale: set flags newer gfortan compilers with older style code

GlazerMann:
Replace pedantic errors flag with fallow-argument-mismatch


t-yamaura (approved):

GlazerMann:
@t-yamaura can you request a co-pilot review please

t-yamaura (approved):

t-yamaura (approved):

repos/spack_repo/builtin/packages/scale/package.py
```diff
@@ -40,6 +40,7 @@ class Scale(MakefilePackage):
     depends_on("parallel-netcdf")
 
     patch("fj-own_compiler.patch", when="%fj")
+    patch("gcc-own_compiler.patch", when="%gcc@10:")
```
alalazo, line 45: Can you add a brief comment above the `patch` directive explaining:
2. Why it is needed
3. Where it was taken from

?

GlazerMann: done

alalazo (approved):

PR #5083, opened 2026-06-03
CI: Add -y to gpg trust

kwryankrattiger:
Spack now requires acknowledgment of keys as they are trusted. CI is headless and so needs to pass -y to trust keys.

### DIFF
--- a/.ci/gitlab/configs/ci.yaml
+++ b/.ci/gitlab/configs/ci.yaml
@@ -16,10 +16,10 @@ ci:
         - spack compiler list
       - - mkdir -p ${SPACK_ARTIFACTS_ROOT}/user_data
         # AWS runners mount E4S public key (verification), UO runners mount public/private (signing/verification)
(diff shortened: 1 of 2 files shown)


johnwparent (dismissed):
All looks good

alecbcs:
@kwryankrattiger will all users of `spack ci` need to update their CI scripts? If so we need to mark the spack/spack change as breaking and make sure it is documented for the release with an example on how to migrate.

kwryankrattiger:
Yeah, this and the buildcache keys command both require accent by default now.

Note, the CI rebuild command has a mechanism to trust a key encoded in a variable.

alecbcs:
Cool! Let's just make sure we document that variable and alternatively script modifications for the release notes.

alalazo (approved):

alalazo:
Force merging, since there will be failures for broken specs in CI on `develop`. Checked a generate job and a build job and they seem correct.

PR #5089, opened 2026-06-04
tau: add conflict for CUDA 13.2+

rbberger:

### DIFF
--- a/repos/spack_repo/builtin/packages/tau/package.py
+++ b/repos/spack_repo/builtin/packages/tau/package.py
@@ -222,6 +222,9 @@ class Tau(Package):
 
     conflicts("+comm", when="@:2.34 +python", msg="Bug in +comm with +python up to @2.34")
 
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/tau/package.py
```diff
@@ -222,6 +222,9 @@ class Tau(Package):
 
     conflicts("+comm", when="@:2.34 +python", msg="Bug in +comm with +python up to @2.34")
 
+    # header changes require upstream to add explicit includes
+    conflicts("cuda@13.2:", when="@:2.35.1 +cuda")
```
sethrj: I've been corrected before to use the direct dependency, rather than "in the spec somewhere", though for this case I don't think it makes a practical difference:
```suggestion
    conflicts("^cuda@13.2:", when="@:2.35.1 +cuda")
```

sethrj (approved):

PR #5093, opened 2026-06-04
gtkplus : edited patch to fix macos darwin build fail

foglienimatteo:
On macOS with Darwin Tahoe, the istallation of gtkplus fails: 

```bash
$ spack install gtkplus@3.24.50    # other versions fails in the same way
...
> FAILED: [code=1] tests/gdkgears.p/gtkgears.c.o
...
> ../spack-src/tests/gtkgears.c:309:5: error: call to undeclared library function 'sincos' with type 'void (double, double *, double *)'; ISO C99 and later do not support implicit function declarations [-Wimplicit-function-declaration]
    309 |     sincos(i * 2.0 * G_PI / teeth + da * 0, &s[0], &c[0]);
        |     ^
> ../spack-src/tests/gtkgears.c:309:5: note: include the header <math.h> or explicitly provide a declaration for 'sincos'
  1 error generated.
...
> /private/var/folders/bp/hvk0k46s5z948ntq8473mg29ft3f82/T/di75tom/spack-stage/spack-stage-gtkplus-3.24.50-vrquitqdw33ui7fkdkn6xtnkak5bhgvx/spack-src/gtk/gtksearchenginequartz.c:226: unexpected character `@'
> /private/var/folders/bp/hvk0k46s5z948ntq8473mg29ft3f82/T/di75tom/spack-stage/spack-stage-gtkplus-3.24.50-vrquitqdw33ui7fkdkn6xtnkak5bhgvx/spack-src/gtk/gtksearchenginequartz.c:230: unexpected character `@'
> /private/var/folders/bp/hvk0k46s5z948ntq8473mg29ft3f82/T/di75tom/spack-stage/spack-stage-gtkplus-3.24.50-vrquitqdw33ui7fkdkn6xtnkak5bhgvx/spack-src/gtk/gtksearchenginequartz.c:233: unexpected character `@'
...
```

1. `sincos` is not defined without `_DARWIN_C_SOURCE`. Fix: injection of a static inline definition (an alternative would be to suppress test compilation with `-Dtests=false`)

2. Objective-C syntax (@-directives) are not interpreted in `-std=c99`,  and `gtksearchengine.c` referencing `_gtk_search_engine_quartz_new()` make the linker fails with undefined symbol. Fix: replace of the function with a minimal C stub returning NULL (not the best fix I admit, open for improvements)


bernhardkaindl:
@spackbot fix style

bernhardkaindl (approved):

PR #5100, opened 2026-06-05
nco: Add v5.3.5 to v5.3.9, add zstd

mathomp4:
This adds the latest `nco` versions to spack. 

I also added `zstd` as a dependency because, well, I needed it to build. @czender can chime in if he thinks it should be version protected. But I can say at least on my test system 5.3.9 was looking for it (maybe because `netcdf-c` is built with `+zstd` by default?)

### DIFF
--- a/repos/spack_repo/builtin/packages/nco/package.py
+++ b/repos/spack_repo/builtin/packages/nco/package.py
@@ -20,6 +20,11 @@ class Nco(AutotoolsPackage):
 
     license("BSD-3-Clause")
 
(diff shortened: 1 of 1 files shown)


xylar (approved):
Thanks @mathomp4!

czender:
Thanks @mathomp4. It's best to build netcdf-c with `--enable-plugins` and in that case one needs Zstd to complete the build.

kshea21 (approved):
Looks good to me, thanks!

PR #5105, opened 2026-06-05
turbine: add v1.5.1, update use of configure flags

j-woz:
Update version and use of configure-time flags.

### DIFF
--- a/repos/spack_repo/builtin/packages/turbine/package.py
+++ b/repos/spack_repo/builtin/packages/turbine/package.py
@@ -12,10 +12,11 @@ class Turbine(AutotoolsPackage):
     """Turbine: The Swift/T runtime"""
 
     homepage = "http://swift-lang.org/Swift-T"
-    url = "https://swift-lang.github.io/swift-t-downloads/spack/turbine-1.3.0.tar.gz"
+    url = "https://swift-lang.github.io/swift-t-downloads/spack/turbine-1.5.1.tar.gz"
     git = "https://github.com/swift-lang/swift-t.git"
 
     version("master", branch="master")
+    version("1.5.1", sha256="67485207b7d9eef1b942453e0be2bd956001ddf2cec556ab09079613f4e86d9b")
     version("1.3.0", sha256="9709e5dada91a7dce958a7967d6ff2bd39ccc9e7da62d05a875324b5089da393")
     version("1.2.3", sha256="a3156c7e0b39e166da3de8892f55fa5d535b0c99c87a9add067c801098fe51ba")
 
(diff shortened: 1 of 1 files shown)


bernhardkaindl:
@j-woz, quick question, I see this removed these lines but I don't see them moved elsewhere, and I don't see a comment about them in the PR, only `use of configure-time flags.` which doesn't explain it for me. Can you explain this?
```py
        if self.spec.satisfies("^intel-oneapi-mpi"):
            args.append("--with-mpi=" + self["intel-oneapi-mpi"].component_prefix)
        else:
            args.append("--with-mpi=" + self.spec["mpi"].prefix)
```

bernhardkaindl:
@j-woz, see this change in the diff:
```doff
-            "--with-tcl=" + self.spec["tcl"].prefix,
+            "--with-tcl=" + self.spec["tcl"].prefix + "/lib",
```
Using `git grep with.*lib,` I see other packages use `.prefix.lib` instead and when I specifically look for tcl, I see that apparently all packages use another method:
```py
git grep with.*tcl.*lib
repos/spack_repo/builtin/packages/environment_modules/package.py:            "--with-tcl=" + tcl.libs.directories[0],
repos/spack_repo/builtin/packages/expect/package.py:            "--with-tcl={0}".format(spec["tcl"].libs.directories[0]),
repos/spack_repo/builtin/packages/python/package.py:                    "--with-tcltk-libs={0} {1}".format(
repos/spack_repo/builtin/packages/python/tkinter-3.8.patch:         # Check whether --with-tcltk-includes and --with-tcltk-libs were
repos/spack_repo/builtin/packages/tcl_itcl/package.py:            "--with-tcl=" + self.spec["tcl"].libs.directories[0],
repos/spack_repo/builtin/packages/tcl_tclxml/package.py:            "--with-tcl={0}".format(self.spec["tcl"].libs.directories[0]),
repos/spack_repo/builtin/packages/tix/package.py:            "--with-tcl={0}".format(spec["tcl"].libs.directories[0]),
repos/spack_repo/builtin/packages/tk/package.py:            "--with-tcl={0}".format(spec["tcl"].libs.directories[0]),
repos/spack_repo/builtin/packages/wordnet/package.py:        args.append("--with-tcl=%s" % self.spec["tcl"].libs.directories[0])
```

bernhardkaindl (changes requested):
@j-woz, AFAICS, could you check if you can replace appending "lib" with the way this appears to be done by the other packages using the same package? https://github.com/spack/spack-packages/pull/5105#issuecomment-5011224169

j-woz:
Sure- I do not know what the intel-oneapi-mpi is for.  It is not needed for oneAPI systems like Aurora.
I will update the Tcl approach.

j-woz:
Ok, I don't know what the intel-oneapi-mpi is used for, and I updated the use of Tcl.

bernhardkaindl:
> Ok, I don't know what the intel-oneapi-mpi is used for, and I updated the use of Tcl.

@j-woz, the lines your PR branch still removes
```py
-        if self.spec.satisfies("^intel-oneapi-mpi"):
-            args.append("--with-mpi=" + self["intel-oneapi-mpi"].component_prefix)
-        else:
-            args.append("--with-mpi=" + self.spec["mpi"].prefix)
```
Were added / moved last here in this PR: https://github.com/spack/spack/pull/42247/changes

I'd normally suggest to revert this removal and add them back (only change what you know).

To not accidentally remove those lines, I'm keeping my review as with changes requested for now.

PR #5111, opened 2026-06-06
starship: add v1.25.1

mcmehrtens:
- Add v1.25.1
- Drops the `cmake` requirement since it is not needed for any of the checksummed versions
- Added more info to license. License hasn't changed since the first tagged release, v0.2.0.

Confirmed all four versions build successfully.

Maintainers: @ebagrenrut

### DIFF
--- a/repos/spack_repo/builtin/packages/starship/package.py
+++ b/repos/spack_repo/builtin/packages/starship/package.py
@@ -15,14 +15,14 @@ class Starship(CargoPackage):
 
     maintainers("ebagrenrut")
 
-    license("ISC")
+    license("ISC", checked_by="mcmehrtens", when="@0.2.0:")
 
+    version("1.25.1", sha256="521306b14066ee7e332d998ef5b5b6455fdc6085c52e86b6316a7cdc37bae1d8")
     version("1.25.0", sha256="e77f3c23683eb544f6dae7171e3c80676aefc66329225bdcd58e40846bb6445f")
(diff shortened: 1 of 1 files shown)


RikkiButler20:
@ebagrenrut Can you please review?

ebagrenrut:
@mcmehrtens I'm pretty sure that I added the `cmake` dependency because starship relies on the Rust [cmake](https://crates.io/crates/cmake) crate, which has a dependency on the `cmake` command being available. Though it's been a while, I have a vague recollection of running into a build failure via Spack CI as a result of having no `cmake` command in my build image. So though you will find no CMake-related files in the starship source tree, I suspect that the `cmake` crate is included because _some other crate_ needs `cmake`. To confirm that `cmake` really isn't needed, did you confirm that you could build on a system, or in a container, that does not have CMake available?

In fact, the dependency on the `cmake` create is I think not even a direct dependency of starship itself, but a dependency of a starship dependency (or at least _somewhere_ in the tree of dependencies that `cargo` pulls in...).

There are very likely other dependencies missing from the `starship` Spack package that Spack could fulfill, I just didn't encounter them. Because Spack doesn't fully model Rust-based builds, I don't think we currently have a way to ferret out all of the dependencies without brute-force (and possibly laborious) analysis of the full cargo-based build :/

ebagrenrut:
@mcmehrtens At this point, with Spack `develop` and the checksummed versions of `starship`, I'm unable to reproduce any issues with `cmake`, even when in a container with no `cmake` (neither from `spack`, nor from the OS). So I approve of this change.

Thanks!

RikkiButler20 (approved):

PR #5112, opened 2026-06-06
neovim: add v0.11.7, v0.12.0:0.12.2

mcmehrtens:
- Added versions `@0.11.7` and `@0.12:0.12.2`
- Updated tree-sitter dependency ranges
  - Neovim requires tree-sitter `@0.25:` since `@0.11:` (see [v0.11.0](https://github.com/neovim/neovim/blob/v0.11.0/src/nvim/CMakeLists.txt#L34) and [v0.12.2](https://github.com/neovim/neovim/blob/v0.12.2/src/nvim/CMakeLists.txt#L34))
  - tree-sitter `@0.26.0` removed `ts_parser_timeout_micros`/`ts_parser_set_timeout_micros` which is why the existing tree-sitter `@0.26:` conflict was added (see https://github.com/tree-sitter/tree-sitter/pull/4814). Neovim `@0.11.7` still calls the removed functions, so the conflict was expanded. The migration off the removed API landed in `@0.12:` (see https://github.com/neovim/neovim/pull/33141).
- Updated utf8proc dependency range
  - changed `utf8proc@2.10.0` exact pin to `utf8proc@2.10.0:`. Neovim `@0.12.2` bundles `utf8proc@2.11.3` so this range is preferred (see [v0.12.2](https://github.com/neovim/neovim/blob/v0.12.2/cmake.deps/deps.txt#L31-L32))
- Updated gperf dependency range
  - Neovim stopped depending on gperf starting `@0.8:`

I confirmed versions `@0.11.6:0.12.2` and `@stable`/`@master` built successfully.

Maintainers: @albestro and @trws

### DIFF
--- a/repos/spack_repo/builtin/packages/neovim/package.py
+++ b/repos/spack_repo/builtin/packages/neovim/package.py
@@ -16,10 +16,14 @@ class Neovim(CMakePackage):
 
     maintainers("albestro", "trws")
 
-    license("Apache-2.0 AND Vim")
+    license("Apache-2.0 AND Vim", checked_by="mcmehrtens", when="@0.1.0:")
 
     version("master", branch="master")
     version("stable", tag="stable")
+    version("0.12.2", sha256="ef9f58da7d687ed4d1dad9715542bf0dabdeedbfe8089e2ce17fff21b920a268")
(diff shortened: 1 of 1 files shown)


albestro (approved):
LGTM

alecbcs (approved):
Looks good to me. Thank you @mcmehrtens!

PR #5116, opened 2026-06-07
[PACKAGE] Add libxs amd libxstream dependencies to CP2K and DBCSR

mtaillefumier:
- Added `libxs` and `libxstream` dependencies
- Updated `cp2k` and `dbcsr`
- Added gauxc dependency


mtaillefumier:
@spackbot fix style

mtaillefumier:
@spackbot fix style

Growl1234:
Would you like to update the recipe here accordingly?

mtaillefumier:
Sure. I need to update libxstream package as well.

repos/spack_repo/builtin/packages/cp2k/package.py
```diff
@@ -132,21 +164,35 @@ class Cp2k(MakefilePackage, CMakePackage, CudaPackage, ROCmPackage):
...
+        when="@2025.2: +cuda",
     )
     variant(
         "pw_gpu",
         default=False,
         description="Enable FFT calculations on GPU",
         when="@2025.2: +rocm",
     )
+
+    variant("libxs", default=False, description="Enable libxs support", when="@master")
```
RMeli:
```suggestion
    variant("libxs", default=False, description="Enable libxs support", when="@2026.2:")
```
This will include `@master`.

repos/spack_repo/builtin/packages/libxs/package.py
```diff
@@ -0,0 +1,47 @@
...
+    homepage = "hhttps://libxs.readthedocs.io/en/latest"
+    url = (
+        "https://github.com/hfp/libxs/archiv/libxs-ab416130f8c9f7edb8c1bf3d3abaf402f61d0fe0.tar.gz"
+    )
+    git = "https://github.com/hfp/libxs.git"
+
+    license("BSD-3-Clause", checked_by="mkrack")
+
+    version("main", branch="main")
+    version("20260605", commit="ab416130f8c9f7edb8c1bf3d3abaf402f61d0fe0")
```
RMeli: @hfp are there plans for releases, or is it OK to arbitrarily tag commits for the time being?

mtaillefumier:
@spackbot fix style

repos/spack_repo/builtin/packages/cp2k/package.py
mtaillefumier: libxs is now defined with `smm=libxs` instead of a new variant. I added your suggestion as well.

mtaillefumier:
spack style --fix

mtaillefumier:
@spackbot fix style

repos/spack_repo/builtin/packages/libxstream/package.py
```diff
@@ -2,27 +2,54 @@
...
 
 
-class Libxstream(Package):
+class Libxstream(MakefilePackage, CMakePackage):
     """LIBXSTREAM is a library to work with streams, events, and code regions
     that are able to run asynchronous while preserving the usual stream
     conditions."""
 
+    build_system(
+        conditional("cmake", when="@main"),
```
RMeli:
```suggestion
        conditional("cmake", when="@1:"),
```
This includes `@main`. The release notes for `0.9.1` suggest that an `@1` release is incoming?

repos/spack_repo/builtin/packages/libxstream/package.py
```diff
@@ -2,27 +2,54 @@
...
+        conditional("makefile", when="@0.9.0"),
+        default="cmake",
+    )
+
     homepage = "https://github.com/hfp/libxstream"
     url = "https://github.com/hfp/libxstream/archive/0.9.0.tar.gz"
+    git = "https://github.com/hfp/libxstream.git"
+
+    maintainers("mtaillefumier")
```
RMeli, line 28:
```suggestion
    maintainers("mtaillefumier", "hpf")
```
@hfp since you are listed in `libxs`, would you like to be added as maintainer of `libxstream` too?

repos/spack_repo/builtin/packages/libxstream/package.py
```diff
@@ -2,27 +2,54 @@
...
 
-class Libxstream(Package):
+class Libxstream(MakefilePackage, CMakePackage):
     """LIBXSTREAM is a library to work with streams, events, and code regions
     that are able to run asynchronous while preserving the usual stream
     conditions."""
 
+    build_system(
+        conditional("cmake", when="@main"),
+        conditional("makefile", when="@0.9.0"),
```
RMeli: What about `0.9.1`?

repos/spack_repo/builtin/packages/libxstream/package.py
```diff
@@ -2,27 +2,54 @@
...
+
+    generator("ninja")
+
+    variant("shared", default=True, description="Build shared libraries")
 
     depends_on("c", type="build")  # generated
     depends_on("cxx", type="build")  # generated
-    depends_on("gmake", type="build")
+    depends_on("fortran", type="build")  # generated
+    depends_on("gmake", type="build", when="@0.9.0")
```
RMeli: What about `0.9.1`?

repos/spack_repo/builtin/packages/dbcsr/package.py
```diff
@@ -138,8 +174,14 @@ class Dbcsr(CMakePackage, CudaPackage, ROCmPackage):
...
-    # OpenCL backend implementation relies on LIBXSMM
-    requires("smm=libxsmm", when="+opencl")
+    # OpenCL backend implementation relies on LIBXSMM for version up to 2.9.1
+    # libxstream/libxs afterwards. libxstream and libxs are tied and libxs is
+    # hard dependency of libxstream in that configuration. This is reflected in
+    # libxstream package.py.
+
+    depends_on("libxstream@main", when="@2.9.2: +opencl")
+    depends_on("libxs@main", when="@2.9.2: +opencl")
+    depends_on("libxs@main", when="@2.9.2: smm=libxs")
```
RMeli: Better to use specific commits?

awvwgk (approved):
GauXC changes look good to me.

repos/spack_repo/builtin/packages/libxstream/package.py
mtaillefumier: I would prefer to remove the `Makefile` build entirely at that stage because only `dbcsr` and `cp2k` use it and require to build `libxs` and `libxstream` with `cmake`.

repos/spack_repo/builtin/packages/dbcsr/package.py
mtaillefumier: i agree with you. New versions are not tagged yet. To solve this I will tag both libraries to the latest commit and name each version using their respective commit date.

mtaillefumier:
@spackbot fix style

mtaillefumier:
@spackbot fix style

mtaillefumier:
@spackbot re-run pipeline

repos/spack_repo/builtin/packages/dbcsr/package.py
RMeli: @hfp mentioned to me last week that he might be releasing this week (or soon-ish). Worth waiting for a proper release at this point, unless is super urgent?

repos/spack_repo/builtin/packages/libxstream/package.py
RMeli: I think it should be deprecated fist, before we are able to remove it.

repos/spack_repo/builtin/packages/dbcsr/package.py
Growl1234: I'd suggest waiting for a release of libxs, which is also important for the follow-up DBCSR and CP2K releases.

mkrack:
I have separate open requests for updating[ `gauxc` ](https://github.com/spack/spack-packages/pull/5294) as well as for adding the new recipe [`libxs`](https://github.com/spack/spack-packages/pull/5297) open. Sorry for interfering with parts of this PR, but I think it is better to get these pass and reviewed one-by-one separately.

repos/spack_repo/builtin/packages/libxstream/package.py
```diff
@@ -25,7 +25,7 @@ class Libxstream(MakefilePackage, CMakePackage):
     url = "https://github.com/hfp/libxstream/archive/0.9.0.tar.gz"
     git = "https://github.com/hfp/libxstream.git"
 
-    maintainers("mtaillefumier")
+    maintainers("mtaillefumier", "hpf")
```
mkrack: I think you want to add `hfp`?

mtaillefumier:
> I have separate open requests for updating[ `gauxc` ](https://github.com/spack/spack-packages/pull/5294) as well as for adding the new recipe [`libxs`](https://github.com/spack/spack-packages/pull/5297) open. Sorry for interfering with parts of this PR, but I think it is better to get these pass and reviewed one-by-one separately.

Beside duplicating work, I honestly don’t understand why separating them would improve the review, especially since the same people will review both this PR and the others.

With this logic, I would have to open a separate PR for libxs, gauxc, integratexx, dbcsr, etc. every time a new dependency is added.

To me, it is clearer to have a single entry point that justifies why a new package is introduced, rather than creating it in one PR and then modifying the dependent package in another.

repos/spack_repo/builtin/packages/dbcsr/package.py
mtaillefumier: I temporarily "tagged" a pre-release in the spack recipe. pointing to the latest commit.

repos/spack_repo/builtin/packages/libxstream/package.py
mtaillefumier: I agree, the only spack packages using it are cp2k and dbcsr (with this PR) so it could make sense to remove it in practice. I kept the Makefile build for now though.

repos/spack_repo/builtin/packages/libxs/package.py
hfp: There is a release now. Sorry, this needed some time.

repos/spack_repo/builtin/packages/libxstream/package.py
hfp: It's not really necessary to support anything prior to 1.0. It's essentially completely evolved. For people's interest, some of the implementation pre-1 was used to implemented the OpenCL flavor of DBCSR's ACC-interface. The latter was taken out of DBCSR this year (2026) like separated, and the "old" libxstream project was reused to take the separated code. It was completely "rebased" to use LIBXS as foundational library and to leverage for instance the code registry.

hfp (approved):

repos/spack_repo/builtin/packages/cp2k/package.py
```diff
@@ -247,11 +312,16 @@ class Cp2k(MakefilePackage, CMakePackage, CudaPackage, ROCmPackage):
...
+    depends_on("libxs@1:+fortran", when="smm=libxs")
+
     depends_on("fftw-api@3")
     depends_on("greenx", when="+greenx")
     depends_on("hdf5+hl+fortran", when="+hdf5")
     depends_on("trexio", when="+trexio")
-    depends_on("deepmdkit", when="+deepmd")
+    depends_on("pkgconfig", type="build")
+
+    depends_on("gauxc", when="+gauxc")
```
awvwgk: Does CP2K need to enable any settings in GauXC? For example, `+fortran`, `+mpi` and `+openmp`?

repos/spack_repo/builtin/packages/libxstream/package.py
```diff
@@ -2,27 +2,58 @@
...
+    git = "https://github.com/hfp/libxstream.git"
+
+    maintainers("mtaillefumier")
 
     license("BSD-3-Clause")
 
+    version(
+        "0.9.0",
+        sha256="03365f23b337533b8e5a049a24bc5a91c0f1539dd042ca5312abccc8f713b473",
+    )
```
awvwgk: 0.9.0 is defined twice here

repos/spack_repo/builtin/packages/libxs/package.py
```diff
@@ -0,0 +1,49 @@
...
+    homepage = "hhttps://libxs.readthedocs.io/en/latest"
+    url = (
+        "https://github.com/hfp/libxs/archive/refs/tags/1.0.tar.gz"
+    )
+    git = "https://github.com/hfp/libxs.git"
+
+    license("BSD-3-Clause", checked_by="mkrack")
+
+    version("main", branch="main")
+    version("20260605", commit="ab416130f8c9f7edb8c1bf3d3abaf402f61d0fe0")
```
Growl1234: Can this line be removed?

repos/spack_repo/builtin/packages/dbcsr/package.py
Growl1234, line 1: DBCSR reipe can also be updated with 2.10.0 release.

Growl1234:
Hmm, I don't think this should be the blocker for this PR...

The current dashboard setup using local recipes means that merging this change would not automatically add or remove its coverage, but waiting for dashboard integration would not validate this update either. This PR updates the upstream recipes to reflect the dependency paths already supported by CP2K and DBCSR, and it can be validated with targeted Spack builds from this branch.

mtaillefumier:
> I would like to note that the CP2K spack recipe updated here is not yet part of the CP2K regressions tests. All spack builds (CP2K development version) for the CP2K regression tests in the [cp2k dashboard](https://dashboard.cp2k.org/index.html) are based on the `make_cp2k.sh` script which performs spack builds using the local spack recipes in the [CP2K source tree](https://github.com/cp2k/cp2k/tree/master/tools/spack/spack_repo/cp2k_dev/packages) or the upstream spack recipes. As long as this CP2K master recipe for `spack install cp2k ...` here is not integrated into the CP2K regression tests, things are done in my opinion in the wrong order. Therefore, I suggest to close this PR.

We’re mixing responsibilities here. If CP2K wants to test the Spack build in its regression suite, the tests should exercise the official upstream Spack recipe (spack install cp2k), not a modified local fork of those packages. Using a separate internal recipe duplicates work, creates review overhead, and risks divergence between what users get from Spack and what CP2K’s own tests validate.

From my perspective, the correct order is: keep the canonical recipe in Spack up to date, and have CP2K’s regression infrastructure follow that, rather than blocking Spack changes until a parallel internal copy is updated.

I’d like to be very clear about why I’m reluctant to close this PR. The changes here were duplicated into another PR, and I’ve now been asked to review essentially the same work I already implemented. That’s duplication of effort both for contributors and reviewers, and it defeats the purpose of peer review: instead of reviewing the original implementation, we’re reviewing a re‑implementation of the same idea.

If the project prefers a different structure (e.g. smaller, split PRs), that’s fine, but we should agree on that up front and avoid having people redo each other’s work and then ask to close the original contribution.

repos/spack_repo/builtin/packages/dbcsr/package.py
mtaillefumier: fixed to 1.0.0

repos/spack_repo/builtin/packages/cp2k/package.py
mtaillefumier: yes +fortran at least. That's why having this done in a single location simplify things a lot.

mtaillefumier:
@spackbot fix style

mtaillefumier:
@spackbot fix style

mtaillefumier:
> That does match my experience. By hastely updating the spack cp2k recipe which does not include proper testing here, you release additional work to the cp2k developers forcing them to follow the trial and error process here, hampering the cp2k release process at the same time in my opinion.
> 
> I would like to hear the opinion of the other [cp2k admins](https://github.com/orgs/cp2k/teams/cp2k-admins) here.

This PR was opened a few weeks before the smaller follow‑up PRs, specifically to include `libxs` and `libxxstream` at a time when no official releases were available. The goal was to avoid breaking Spack’s CI/CD (for setups using cp2k@master) once `libxs` was enabled by default: without this recipe, users and developers relying on `Spack` would hit build issues. In that sense, what is now called ‘hasty’ was actually a justified and proactive fix.

I’m also one of the maintainers of the `cp2k` and `dbcsr` Spack packages, so I get pinged every time these recipes change. Duplicating the work from this PR into separate smaller PRs doesn’t just increase review noise for me, it adds avoidable maintenance overhead for the project as a whole. Given that, I don’t understand why the work here was duplicated instead of contributing directly to this PR.

mtaillefumier:
@spackbot fix style

mtaillefumier:
@spackbot fix style

Growl1234:
Just a note:

Libxsmm 2.0 has been released, but it no longer provides the legacy `libxsmmext` interface required by older DBCSR and CP2K releases. Their recipes therefore need corresponding version constraints.

This has already been handled in #5354 alongside the libxsmm recipe update, so it should not result in duplicate effort.

repos/spack_repo/builtin/packages/cp2k/package.py
```diff
@@ -69,18 +99,25 @@ class Cp2k(MakefilePackage, CMakePackage, CudaPackage, ROCmPackage):
 
     generator("ninja")
 
+    # both are set to false to avoid breaking spack CI/CD in the eventuality of
+    # static builds
```
RMeli:
```suggestion
```

repos/spack_repo/builtin/packages/cp2k/package.py
```diff
@@ -247,12 +315,19 @@ class Cp2k(MakefilePackage, CMakePackage, CudaPackage, ROCmPackage):
...
+
     depends_on("fftw-api@3")
     depends_on("greenx", when="+greenx")
     depends_on("hdf5+hl+fortran", when="+hdf5")
     depends_on("trexio", when="+trexio")
     depends_on("deepmdkit", when="+deepmd")
 
+    depends_on("gauxc+fortran", when="+gauxc")
+    depends_on("gauxc+fortran+mpi", when="+gauxc+mpi")
+    depends_on("gauxc+pic", when="+pic")
```
RMeli:
```suggestion
    with when("+gauxc"):
        depends_on("gauxc+fortran")
        depends_on("gauxc+fortran+mpi", when="+mpi")
        depends_on("gauxc+pic", when="+pic")
```
The last one should have been `+pic+gauxc`.

repos/spack_repo/builtin/packages/cp2k/package.py
```diff
@@ -272,29 +350,34 @@ class Cp2k(MakefilePackage, CMakePackage, CudaPackage, ROCmPackage):
         # we have to be consistent.
         depends_on("cray-libsci+openmp", when="^[virtuals=blas] cray-libsci")
 
-    with when("smm=libxsmm"):
+    with when("@:2026.1 smm=libxsmm"):
```
RMeli: I think this constraint should be reflected in the `smm` variant; it should make the `libxsmm` option conditionally available (with `conditional()`) on `@:2026.1`.

repos/spack_repo/builtin/packages/cp2k/package.py
```diff
@@ -272,29 +350,34 @@ class Cp2k(MakefilePackage, CMakePackage, CudaPackage, ROCmPackage):
...
-        depends_on("pkgconfig", type="build")
 
     # Several packages provide "opencl" (incl. ICD/loader), e.g., "cuda"
     with when("+opencl"):
         depends_on("opencl", when="+opencl")
         opencl_loader_header_version = "2022.10.24"
         depends_on(f"opencl-c-headers@{opencl_loader_header_version}:")
         requires(f"%opencl=opencl-icd-loader@{opencl_loader_header_version}:")
-        # OpenCL backend implementation relies on LIBXSMM
-        requires("smm=libxsmm")
```
RMeli, line 289: Still required for older versions?

repos/spack_repo/builtin/packages/cp2k/package.py
```diff
@@ -272,29 +350,34 @@ class Cp2k(MakefilePackage, CMakePackage, CudaPackage, ROCmPackage):
...
+        # OpenCL backend implementation relies on LIBXS starting from 2026.2
+        # and master (June 10 2026)
```
RMeli:
```suggestion
```

repos/spack_repo/builtin/packages/cp2k/package.py
```diff
@@ -1204,6 +1317,13 @@ def cmake_args(self):
         else:
             args += ["-DCP2K_USE_LIBXSMM=OFF"]
 
+        # Only master and version starting from 2026.2 support libxs OpenCL
+        # also requires it.
```
RMeli:
```suggestion
```

repos/spack_repo/builtin/packages/dbcsr/package.py
```diff
@@ -44,8 +84,8 @@ class Dbcsr(CMakePackage, CudaPackage, ROCmPackage):
     variant("shared", default=True, description="Build shared library")
     variant(
         "smm",
-        default="libxsmm",
-        values=("libxsmm", "blas"),
+        default="blas",
```
RMeli, line 95: Should the default change conditionally on the version? I think only changing the default going forward would be best.

repos/spack_repo/builtin/packages/dbcsr/package.py
```diff
@@ -138,8 +178,14 @@ class Dbcsr(CMakePackage, CudaPackage, ROCmPackage):
     conflicts("^openblas threads=pthreads", when="+openmp")
     conflicts("^openblas threads=none", when="+openmp")
 
-    # OpenCL backend implementation relies on LIBXSMM
-    requires("smm=libxsmm", when="+opencl")
```
RMeli, line 142: Still needed for old versions?

RMeli:
> As long as this CP2K master recipe for spack install cp2k ... here is not integrated into the CP2K regression tests, things are done in my opinion in the wrong order.

@Growl1234 I suggested a while ago to integrate the CP2K Spack recipe in CP2K itself for co-development and test both `spack install --only=dependencies <cp2k spec>` (to install the required dependencies for a given `<cp2k spec>`) and `spack install cp2k` to make sure the Spack recipe works well. It is of course a bit annoying to keep things in sync

Instead, as far as I can tell (I lost a bit the plot given the fast development in the last few months), Spack is still used only to install dependencies one-by-one and there is no CP2K Spack package co-developed with CP2K. How would you do things differently?

> I have separate open requests for updatinghttps://github.com/spack/spack-packages/pull/5294 as well as for adding the new recipe https://github.com/spack/spack-packages/pull/5297 open. Sorry for interfering with parts of this PR, but I think it is better to get these pass and reviewed one-by-one separately.

@mkrack @mtaillefumier how would you like to progress here?

Growl1234:
I am not fully sure what the best organizational model would be, but I agree that the current separation can slow things down. The CP2K-side dependency setup around `make_cp2k.sh` and the public CP2K recipe in spack-packages can evolve independently, while changes involving shared dependencies often need manual synchronization and validation across both repositories.

Personally, I would be hesitant to move the CP2K recipe itself into the CP2K main repository. It is part of Spack's public package collection and should probably remain maintained there as the single source of truth for `spack install cp2k`; otherwise, we may simply create another source of divergence.

To me, the more important missing piece is a reliable bridge between the two development paths: some way to regularly validate the Spack CP2K recipe against current CP2K development, and to validate CP2K-side dependency changes against the corresponding Spack package state. I do not yet have a concrete proposal for how to implement that cleanly, but the current split clearly makes coordinated changes slower than they should be.

repos/spack_repo/builtin/packages/dbcsr/package.py
mtaillefumier: I agree blas should be the default.

mtaillefumier:
> Personally, I would be hesitant to move the CP2K recipe itself into the CP2K main repository. It is part of Spack's public package collection and should probably remain maintained there as the single source of truth for `spack install cp2k`; otherwise, we may simply create another source of divergence.
> 
> To me, the more important missing piece is a reliable bridge between the two development paths: some way to regularly validate the Spack CP2K recipe against current CP2K development, and to validate CP2K-side dependency changes against the corresponding Spack package state. I do not yet have a concrete proposal for how to implement that cleanly, but the current split clearly makes coordinated changes slower than they should be.

I also agree that there should ideally be one source of truth for the CP2K Spack recipe, namely the official package in Spack. If CP2K wants to maintain a separate copy for internal purposes, that’s of course fine, but it should not result in asking contributors to close open PRs in a repository CP2K does not control.

I am happy to close this PR if there is a clear, technical argument that the changes here are incorrect, obsolete, or otherwise not worth merging. In many cases, however, these PRs are opened reactively because changes on the CP2K side break `spack install cp2k`, which users (myself included) rely on, and keeping the upstream recipe working is an important part of that support.

It is also not difficult to test the official version of the package within the cp2k testing framework. We do this for fedora and other distributions already.

Just to clarify: by ‘technical reason’ I mean things like incorrect dependencies, broken variants, or an approach that we agree is fundamentally wrong for Spack.

Simply splitting the same work into new PRs and then asking me to review that re‑implementation is not such a reason; it effectively means I’m reviewing my own work, which doesn’t make sense from a peer‑review perspective.

If you think this PR should be structured differently, I’d much prefer we iterate here (patches, suggestions, or splitting this branch) instead of duplicating it elsewhere and then requesting closure of the original.

mkrack:
@mtaillefumier I have closed my separate PRs for `libxs` and `gauxc`.

repos/spack_repo/builtin/packages/cp2k/package.py
mtaillefumier: OK. your suggestion is to apply this when we define the variant. It makes sense.

repos/spack_repo/builtin/packages/dbcsr/package.py
RMeli: I understand, but this changes the default for all the versions too, so it might be surprising to some users which rely on the default (and have to change their specs to an explicit `cp2k smm=libxsmm`). What I'm suggesting is to do this change only for upcoming versions with `conditional()`.

repos/spack_repo/builtin/packages/cp2k/package.py
```diff
@@ -69,18 +99,25 @@ class Cp2k(MakefilePackage, CMakePackage, CudaPackage, ROCmPackage):
 
     generator("ninja")
 
+    # both are set to false to avoid breaking spack CI/CD in the eventuality of
+    # static builds
+    variant("shared", default=False, description="Enable shared libraries")
+    variant("pic", default=False, description="Enable position independent code")
```
Growl1234: I don't understand what's the point of setting PIC to false.

mtaillefumier:
@spackbot fix style

Growl1234 (approved):

mkrack (approved):

RMeli (approved):

PR #5121, opened 2026-06-08
Deprecate all the `cbtf` packages and dependents

alalazo:
After an analysis of packages depending on old boost versions it seems these packages are unmaintained, and upstream development stopped around 2018. `openspeedshop` in particular seems to be tied to very old versions of `binutils` and possibly needing Python 2.


rbberger (approved):
No apparent uses from what I can tell.

PR #5124, opened 2026-06-08
Add recent Mille and Millepede-II releases 

goblirsc:
Add four new versions: 
- Mille: 01-00-04 and 01-00-05
- Millepede-II 05-01-04 and 05-01-05. 

These address build issues (configuration and clang-warnings), but do not change runtime behaviour or API.


goblirsc:
checksum test should clear after re-running (succeeded before updating branch)

repos/spack_repo/builtin/packages/mille/package.py
```diff
@@ -12,12 +12,14 @@ class Mille(CMakePackage):
     and supported by the GeneralBrokenLines track fitter."""
 
     homepage = "https://gitlab.desy.de/millepede/mille"
-    url = "https://gitlab.desy.de/millepede/mille/-/archive/V01-00-00/mille-V01-00-00.tar.gz"
+    url = "https://gitlab.desy.de/millepede/mille/-/archive/V01-00-05/mille-V01-00-05.tar.gz"
```
wdconinc, line 15: (FYI) Updating the `url` is not necessary since spack automatically figures out the correct URL for each version.

repos/spack_repo/builtin/packages/mille/package.py
wdconinc, line 43: Should this also get a gcc requirement for older versions?
```
    requires("%gcc", msg="Mille hardcodes gcc/g++/gfortran", when="@:01-00-03")
```

repos/spack_repo/builtin/packages/mille/package.py
goblirsc: Thank you! This was honestly a workaround for a weird behaviour in the CI, where the checksum verification test would fail the first time it is run, and I had to introduce a trivial change that does not upset the formatter to trigger a rerun 😇

repos/spack_repo/builtin/packages/mille/package.py
goblirsc: good idea - fixed!

wdconinc (approved):
Thanks. Looks good to me.

goblirsc:
Seems the CI failed for an unrelated reason (max run time on AWS)? Can it please be retried?

wdconinc:
@spackbot run pipeline

PR #5126, opened 2026-06-08
xolotl: 3.1.0 uses older kokkos 4.4 classes

GlazerMann:


bernhardkaindl (approved):
Confirmed that `xolotl@3.1.0` does not build with `^kokkos@5` but it builds with this PR.

PR #95, opened 2025-06-06
visit: add smoke test

vicentebolea:
> [!NOTE]
> This PR was migrated from **https://github.com/spack/spack/pull/48634**.

Not ready to merge/review

PR #5142, opened 2026-06-09
Disable smoke test of visit in CI pipelines

zackgalbreath:
VisIt CI jobs started timing out after #95 was merged. Disable the smoke test added in that PR to unblock our CI pipelines.


kwryankrattiger (approved):
LGTM.

It would be good to figure out a way to turn this back on. But I think this is the right thing for now.

PR #5145, opened 2026-06-10
py-uv: support Cargo and add latest versions

mcmehrtens:
- add support for both `PythonPackage` and `CargoPackage` with `PythonPackage` as default
- add latest versions


mcmehrtens:
Realizing that this PR will conflict with #3962... Should I close this? @johnwparent @trws @adamjstewart

johnwparent:
> Realizing that this PR will conflict with #3962... Should I close this? @johnwparent @trws @adamjstewart

No, by all means this should go ahead. The other PR is easily rebased.

adamjstewart:
Previous PR was held up for good reason. We eventually decided the right approach is to support multiple build systems in a single package. We should use PythonPackage by default (as there are assumptions about `python -m uv` working) and support CargoPackage as an alternative.

There are two things here: rename and multiple build systems. These don't have to be done in the same PR, and it's probably easier to separate them. But note that we need to deprecate py-uv instead of having two.

mcmehrtens:
@adamjstewart Got it, I can deprecate `py-uv` in this PR.

Do I need to back port the existing versions in `py-uv`?

mcmehrtens:
> We should use PythonPackage by default (as there are assumptions about `python -m uv` working) and support CargoPackage as an alternative.

Ahh, I missed this part of your message, @adamjstewart... In this PR, I can either:

1. Implement both build systems, Python as default (assuming the infrastructure is there to support that).
2. Swap the build system from Cargo to Python.
3. Leave the PR as is.
4. Do something else :)

Let me know how you'd like me to move forward!

adamjstewart:
1

mcmehrtens:
@adamjstewart RFR

adamjstewart:
Just remembered, the other reason the previous PR was held up was that we decided we should have a new `py` namespace so that `uv` and `py.uv` are equivalent. We also wanted to make this backwards compatible so that `py-uv` would map correctly. And we didn't want to move one package at a time, we wanted to move all Python packages at once.

So I would remove the rename from this PR and focus on the multiple build systems only.

mcmehrtens:
@adamjstewart Got it, I should have it implemented as you've advised. I confirmed that ALL versions built (took a while) and tested successfully.

repos/spack_repo/builtin/packages/py_uv/package.py
```diff
@@ -41,11 +66,26 @@ class PyUv(PythonPackage):
...
+        "python_pip",
+        conditional("cargo", when="@0.7.5:"),
+        default="python_pip",
+    )
+
+    # Both build systems compile the Rust sources (maturin shells out to cargo),
+    # and the bundled `-sys` crates need a C compiler
+    depends_on("c", type="build")
+    # the bundled jemalloc allocator builds its C sources with `./configure && make`
+    depends_on("gmake", type="build")
```
adamjstewart: Can we move these under:
```python
```

repos/spack_repo/builtin/packages/py_uv/package.py
```diff
@@ -57,11 +97,49 @@ class PyUv(PythonPackage):
...
         depends_on("rust@1.81:")
 
-        # Historical dependencies
-        depends_on("cmake", when="@:0.6.3")
+    with when("build_system=python_pip"):
+        with default_args(type="build"):
+            depends_on("py-maturin@1")
+            # maturin builds against the Python interpreter; every release sets
+            # requires-python = ">=3.8" in pyproject.toml
+            depends_on("python@3.8:")
```
adamjstewart: Shouldn't this be build/run or build/link/run?

repos/spack_repo/builtin/packages/py_uv/package.py
```diff
@@ -57,11 +97,49 @@ class PyUv(PythonPackage):
...
+            return "crates/uv"
+        return self.stage.source_path
+
+    @property
+    def build_args(self):
+        # uv's `default` feature set bundles `test-defaults` (integration tests
+        # against live crates.io/PyPI/Git/R2); drop it via --no-default-features
+        # and re-enable only the two release features we want
+        features = "uv-distribution/static,performance"
+        return ["--locked", "--no-default-features", "--features", features]
```
adamjstewart: Don't a lot of these things need to be in a separate CargoBuilder class?

repos/spack_repo/builtin/packages/py_uv/package.py
mcmehrtens: Done in 19230c5bd7905a05fdace7c07012d5ff8af85d07.

repos/spack_repo/builtin/packages/py_uv/package.py
mcmehrtens: Yep, build/run. Good catch, fixed it in 61c7174ed81cd23d31cb7407a306fd7bc94654f9.

repos/spack_repo/builtin/packages/py_uv/package.py
mcmehrtens: Yes... this was sloppy on my part. Sorry about that. Addressed this in 87d53df322e56eb2f2f300b0cc5890321ef61ee5.

mcmehrtens:
RFR @johnwparent @adamjstewart

repos/spack_repo/builtin/packages/py_uv/package.py
```diff
@@ -57,11 +101,53 @@ class PyUv(PythonPackage):
...
+    @classmethod
+    def determine_version(cls, exe):
+        output = Executable(exe)("--version", output=str, error=str)
+        match = re.match(r"uv (\S+)", output)
+        return match.group(1) if match else None
+
+    def test_imports(self):
+        """import the uv module"""
+        if self.spec.satisfies("build_system=python_pip"):
+            super().test_imports()
```
adamjstewart, line 125: Can't remember if this is tied to the build system or the package. Shouldn't this be inherited automatically based on base class? Can you try removing this and see if the tests still run?

repos/spack_repo/builtin/packages/py_uv/package.py
```diff
@@ -41,11 +69,27 @@ class PyUv(PythonPackage):
...
+    build_system(
+        "python_pip",
+        conditional("cargo", when="@0.7.5:"),
+        default="python_pip",
+    )
+
     with default_args(type="build"):
+        # Both build systems compile the Rust sources (maturin shells out to cargo),
+        # and the bundled `-sys` crates need a C compiler
+        depends_on("c")
```
adamjstewart, line 86: Below it says:

> uv @:0.6.3 bundles a `-sys` crate

Should this also be version-dependendent?

repos/spack_repo/builtin/packages/py_uv/package.py
mcmehrtens: Please see [py-uv_test_results.txt](https://github.com/user-attachments/files/29562850/py-uv_test_results.txt). I commented out the `test_imports()` function, and the Cargo-built version failed and the Pip-built version passed.

Based on my tracing of the code, it seems like `test_import()` gets inherited from `PythonPackage`, but it doesn't seem to gate itself automatically on just the Python build.

repos/spack_repo/builtin/packages/py_uv/package.py
mcmehrtens: Addressed in 1b98a2cb043b3480f8090604a260ff1b7b271f53. I wrote the comments pretty poorly before. ALL versions need a C compiler for building `-sys` crates. Version :0.6.3 ALSO needed CMake to build one of the `-sys` crates, but this crate was dropped in later versions—hence the version gate for the CMake dep.

repos/spack_repo/builtin/packages/py_uv/package.py
adamjstewart: @alalazo is this the way things are supposed to work?

adamjstewart (approved):

PR #5151, opened 2026-06-10
util-linux: add v2.42.1, v2.41.4 and deprecate old versions

alalazo:
Add new versions that can compiler with glibc@2.42:


becker33 (changes requested):
Ignore this, I can't read

becker33 (approved):

PR #5152, opened 2026-06-10
autotools: relax the nixos requirement to patch configure files

alalazo:
Let `autotools` patch configure file to use `file` from the PATH and not just `/usr/bin/file`. Previously this was done only for `nixos`.

### DIFF
--- a/repos/spack_repo/builtin/build_systems/autotools.py
+++ b/repos/spack_repo/builtin/build_systems/autotools.py
@@ -314,15 +314,11 @@ To resolve this problem, please try the following:
 
     @run_before("configure")
     def _patch_usr_bin_file(self) -> None:
-        """On NixOS file is not available in /usr/bin/file. Patch configure
-        scripts to use file from path."""
-
-        if self.spec.os.startswith("nixos"):
-            x = FileFilter(
-                *filter(is_exe, find(self.build_directory, "configure", recursive=True))
-            )
-            with keep_modification_time(*x.filenames):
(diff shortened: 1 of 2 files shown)


becker33 (dismissed):

alalazo:
Failures seem unrelated:
- https://gitlab.spack.io/spack/spack-packages/-/pipelines/1641926/failures (fails on visit, should be fixed on `develop`)
- https://gitlab.spack.io/spack/spack-packages/-/pipelines/1641896/failures (also visit)

There's another failure to fetch python packages from cache, that I just re-tried.

alalazo:
The failure I saw in https://gitlab.spack.io/spack/spack-packages/-/jobs/22757268 and thought was due to lack of `/usr/bin/file` because of:
```
/tmp/root/spack-stage/spack-stage-openmpi-5.0.10-nzr6pd7wjjwsa7eltwui6dzmv7klrucn/spack-src/configure: line 198640: /usr/bin/file: No such file or directory
```
was a red herring. The real issue, that happened two times in a row, was several lines down in the log file:
```
/tmp/root/spack-stage/spack-stage-openmpi-5.0.10-nzr6pd7wjjwsa7eltwui6dzmv7klrucn/spack-src/configure: fork: Cannot allocate memory
```
This PR is probably still worth considering, but since it's not a blocker for the tutorial, we can do it after we are finished with ISC.

hppritcha (approved):

becker33 (approved):

PR #5159, opened 2026-06-10
py-types-protobuf: new package

V-Karch:


bernhardkaindl (approved):
Built on Ubuntu 24.04. As it only provides typing stubs for protobuf, it is an easy package.

PR #5160, opened 2026-06-10
p3dfft3: add missing releases up to latest

constracktor:
The current package only includes the initial release v3.0.0 and the current master branch. This PR adds the missing releases up to v3.1.2.


bernhardkaindl (approved):
Built all versions on Ubuntu 24.04

PR #5163, opened 2026-06-10
update nektar package

JamesEdgeley:
Adds new versions to the nektar package


repos/spack_repo/builtin/packages/nektar/package.py
```diff
@@ -166,8 +179,13 @@ def hasfeature(feature):
...
         if "+python" in spec:
-            python = which("python", required=True)
-            with working_dir(self.build_directory):
+            python = which("python")
+            if spec.satisfies("@master") or spec.satisfies("@5.6.0:"):
+                python_build_directory = os.path.join(self.build_directory, "python")
+            else:
+                python_build_directory = self.build_directory
+            print("Installing Python bindings using " + python_build_directory + "/setup.py")
+            with filesystem.working_dir(python_build_directory):
```
alecbcs:
```suggestion
            with working_dir(python_build_directory):
```

@JamesEdgeley let's try to drop the `spack.llnl.filesystem` import in this PR. I believe `working_dir` should already be available through `spack.package`

repos/spack_repo/builtin/packages/nektar/package.py
```diff
@@ -97,8 +106,8 @@ class Nektar(CMakePackage):
     depends_on("c", type="build")  # generated
     depends_on("cxx", type="build")  # generated
 
-    depends_on("cmake@2.8.8:", type="build", when="~hdf5")
-    depends_on("cmake@3.2:", type="build", when="+hdf5")
```
alecbcs, line 101: @JamesEdgeley just want to double check that removing these minimum `cmake` version dependencies was intentional?

JamesEdgeley: I've updated them to mirror the CMakeLists.txt for each version (the cmake requirements have been updated so the previous depends_on was redundant)

repos/spack_repo/builtin/packages/nektar/package.py
```diff
@@ -12,19 +12,29 @@
...
+    version("5.3.0", tag="v5.3.0", commit="f286f809cfeb26cb73828c90a689a048898971d2")
 
+    variant("cwipi", default=False, description="Builds with CWIPI support")
     variant("mpi", default=True, description="Builds with mpi support")
     variant("fftw", default=True, description="Builds with fftw support")
     variant("arpack", default=True, description="Builds with arpack support")
     variant("tinyxml", default=True, description="Builds with external tinyxml support")
     variant("hdf5", default=True, description="Builds with hdf5 support")
-    variant("scotch", default=False, description="Builds with scotch partitioning support")
+    variant("scotch", default=True, description="Builds with scotch partitioning support")
```
alecbcs, line 38: We typically try to keep as many variants as possible to `default=False` instead of `True` so that we keep the default build of package minimal and allow users to opt into additional functionality.

Is there a particular reason we'd like to enable `scotch` by default?

bernhardkaindl:
@JamesEdgeley, could you join the conversation in the PR review thread of this PR?

repos/spack_repo/builtin/packages/nektar/package.py
JamesEdgeley: @alecbcs Apologies for the delay - I have confirmed with the devs that this is necessary. We have moved to parallel hdf5 by default and scotch is required for this.  I have added a conflict to express this.

alecbcs:
```diff
+ variant("scotch", default=True, description="Builds with scotch partitioning support", when="@5.6.0:)
  variant("scotch", default=False, description="Builds with scotch partitioning support")
```

No worries :) Is it possible we can modify the variant to only be True by default for newer versions (by adding a when condition like this) so that we aren't changing the default build of nektar for users who are installing the older versions?

JamesEdgeley: Yes, I think there was a mistake in the original (probably my fault).  It should have been scotch that was true by default and hdf5 false by default - up until 5.10.0, when both should be true by default

repos/spack_repo/builtin/packages/nektar/package.py
```diff
@@ -12,19 +12,30 @@
...
 
+    variant("cwipi", default=False, description="Builds with CWIPI support")
     variant("mpi", default=True, description="Builds with mpi support")
     variant("fftw", default=True, description="Builds with fftw support")
     variant("arpack", default=True, description="Builds with arpack support")
     variant("tinyxml", default=True, description="Builds with external tinyxml support")
-    variant("hdf5", default=True, description="Builds with hdf5 support")
-    variant("scotch", default=False, description="Builds with scotch partitioning support")
+    variant("hdf5", default=True, description="Builds with hdf5 support", when="@5.10.0:")
+    variant("hdf5", default=False, description="Builds with hdf5 support")
```
ningli-amd, line 37: Is this a mistake? The second statement will make the first useless.

JamesEdgeley: Intended behaviour is +hdf5  true by default for version>5.10.0 otherwise false by default.  Is this possible using variants?

PR #5166, opened 2026-06-10
update rucio-client to 40.2.0

panta-123:
add rucio-clients 40.2.0

### DIFF
--- a/repos/spack_repo/builtin/packages/py_paramiko/package.py
+++ b/repos/spack_repo/builtin/packages/py_paramiko/package.py
@@ -15,6 +15,7 @@ class PyParamiko(PythonPackage):
 
     license("LGPL-2.1-or-later")
 
+    version("4.0.0", sha256="6a25f07b380cc9c9a88d2b920ad37167ac4667f8d9886ccebd8f90f654b5d69f")
     version("3.5.1", sha256="b2c665bc45b2b215bd7d7f039901b14b067da00f3a11e6640995fd58f2664822")
     version("3.5.0", sha256="ad11e540da4f55cedda52931f1a3f812a8238a7af7f62a60de538cd80bb28124")
     version("3.4.1", sha256="8b15302870af7f6652f2e038975c1d2973f06046cb5d7d65355668b3ecbece0c")
@@ -32,6 +33,7 @@ class PyParamiko(PythonPackage):
     variant("invoke", default=False, description="Enable invoke support")
 
     depends_on("py-setuptools", type="build")
+    depends_on("py-setuptools@77:", when="@4:", type="build")
     depends_on("py-bcrypt@3.1.3:", when="@2.7:", type=("build", "run"))
     depends_on("py-bcrypt@3.2:", when="@3:", type=("build", "run"))
     depends_on("py-cryptography@2.5:", when="@2.7:", type=("build", "run"))
--- a/repos/spack_repo/builtin/packages/py_rucio_clients/package.py
+++ b/repos/spack_repo/builtin/packages/py_rucio_clients/package.py
@@ -17,6 +17,7 @@ class PyRucioClients(PythonPackage):
 
     license("Apache-2.0", checked_by="wdconinc")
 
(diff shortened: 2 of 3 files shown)


panta-123:
failed job: https://gitlab.spack.io/spack/spack-packages/-/jobs/22765496 
```
==> Error: bad_alloc
```

repos/spack_repo/builtin/packages/py_paramiko/package.py
wdconinc, line 35:
```suggestion
    depends_on("py-setuptools", type="build")
    depends_on("py-setuptools@77:", when="@4:", type="build")
```

repos/spack_repo/builtin/packages/py_rucio_clients/package.py
```diff
@@ -41,24 +42,33 @@ class PyRucioClients(PythonPackage):
...
     depends_on("py-setuptools", type="build")
     depends_on("py-requests@2.32.2:", type=("build", "run"), when="@:36")
+    depends_on("py-requests@2.32.3:", type=("build", "run"), when="@37:38")
+    depends_on("py-requests@2.32.5:", type=("build", "run"), when="@40:")
     depends_on("py-urllib3@1.26.18:", type=("build", "run"), when="@:36")
-    depends_on("py-requests@2.32.3:", type=("build", "run"), when="@37:")
-    depends_on("py-urllib3@2.3.0:", type=("build", "run"), when="@37:")
+    depends_on("py-urllib3@2.3.0:", type=("build", "run"), when="@37:38")
+    depends_on("py-urllib3@2.5.0:", type=("build", "run"), when="@40:")
+
```
wdconinc, line 57:
```suggestion
```

repos/spack_repo/builtin/packages/py_rucio_clients/package.py
```diff
@@ -41,24 +42,33 @@ class PyRucioClients(PythonPackage):
     depends_on("python@3.9:", type=("build", "run"))
     depends_on("py-setuptools", type="build")
     depends_on("py-requests@2.32.2:", type=("build", "run"), when="@:36")
+    depends_on("py-requests@2.32.3:", type=("build", "run"), when="@37:38")
```
wdconinc: (FYI) In general we don't need to close these ranges. We can just also have:
```suggestion
    depends_on("py-requests@2.32.3:", type=("build", "run"), when="@37:")
```
and the stricter version for newer versions would also take effect. The benefit of that approach is that you don't need to go close the when-ranges so it removes git diff noise.

repos/spack_repo/builtin/packages/py_rucio_clients/package.py
```diff
@@ -41,24 +42,33 @@ class PyRucioClients(PythonPackage):
...
+    depends_on("py-requests@2.32.5:", type=("build", "run"), when="@40:")
```
wdconinc, line 53: In all these additions you are missing v39 since you jump straight from 37:38 to 40, so someone installing v39 will not get any of the dependencies. That's another thing that isn't able to happen with open-ended ranges.

repos/spack_repo/builtin/packages/py_rucio_clients/package.py
```diff
@@ -41,24 +42,33 @@ class PyRucioClients(PythonPackage):
     depends_on("python@3.9:", type=("build", "run"))
     depends_on("py-setuptools", type="build")
```
wdconinc, line 43:
```suggestion
    depends_on("py-setuptools", type="build")
    depends_on("py-setuptools@78.1:", type="build", when="@39:")
    depends_on("py-wheel@0.45.1:", type="build", when="@39:")
    depends_on("py-packaging", type="build", when="@39:")
```
`py-packaging` is also below, but having the build deps block be complete is helpful.

Ref: https://github.com/rucio/rucio/commit/cadbd8e0f89252e6bc026ebe1c3103cb0434f7cc

wdconinc (approved):

PR #5167, opened 2026-06-10
setuptools: conflicts vendored deps

RobertMaaskant:
Added conflicts for all vendored dependencies, because they are used when present instead of the vendored versions which leads to conflicts.

https://github.com/pypa/setuptools/blob/v71.0.0/NEWS.rst

>   Now setuptools declares its own dependencies in the core extra.
>   Dependencies are still vendored for bootstrapping purposes,
>   but setuptools will prefer installed dependencies if present.

### DIFF
--- a/repos/spack_repo/builtin/packages/py_setuptools/package.py
+++ b/repos/spack_repo/builtin/packages/py_setuptools/package.py
@@ -77,9 +77,89 @@ class PySetuptools(Package, PythonExtension):
 
     depends_on("py-pip", type="build")
 
+    # https://github.com/pypa/setuptools/blob/v71.0.0/NEWS.rst
+    #
+    # Now setuptools declares its own dependencies in the core extra.
+    # Dependencies are still vendored for bootstrapping purposes,
+    # but setuptools will prefer installed dependencies if present.
+
+    # wheel>=0.43.0 v71.0.0 +
+    conflicts("py-wheel@:0.42", when="@71:")
+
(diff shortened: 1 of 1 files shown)


adamjstewart (commented):
Good catch! Maybe we can make the descriptions shorter? I like the URL and first paragraph, but I think all other comments could be removed, you're basically just translating between pyproject.toml and spack so it's duplicated.

RobertMaaskant:
> Good catch! Maybe we can make the descriptions shorter? I like the URL and first paragraph, but I think all other comments could be removed, you're basically just translating between pyproject.toml and spack so it's duplicated.

Agreed. I auto generated the comments from the `pyproject.toml` files and kept them to make reviewing easier. Removed them for the final version as suggested.

adamjstewart:
Are you using LLMs to synthesize pyproject.toml content across multiple versions? Did you carefully validate all versions for correctness?

RobertMaaskant:
I don't use LLM's to code. Wrote a script by hand. Did multiple checks to make sure the script was correct.

adamjstewart (approved):
Would be easier for me to cross-check if these were in the same order as in pyproject.toml.

Note that I'm attempting to deprecate almost all versions of setuptools in another PR. Still a long way off from succeeding, but may make this simpler someday.

RobertMaaskant:
Slight preference to keep them alphabetical as that always works, whereas the sorting in project.tom might change over time. But feel free to change the ordering if that makes your life easier.

PR #5168, opened 2026-06-10
Kokkos: make installation error into a constraint

tpadioleau:
Before, installation error:
```
spack install kokkos +cuda cuda_arch=80 ~wrapper ~cmake_lang target=x86_64 %gcc
...
==> No binary for kokkos-5.0.2-6jmjl5ow6nviaiwzselltkjdwfk5hded found: installing from source
==> Installing kokkos-5.0.2-6jmjl5ow6nviaiwzselltkjdwfk5hded [16/16]
==> Using cached archive: /local/home/tpadiole/spack-1.1.1/var/spack/cache/_source-cache/archive/18/188817bb452ca805ee8701f1c5adbbb4fb83dc8d1c50624566a18a719ba0fa5e.tar.gz
==> No patches needed for kokkos
==> kokkos: Executing phase: 'cmake'
==> Error: InstallError: Kokkos requires +wrapper when using +cuda without %clang, %cce or +cmake_lang

/local/home/tpadiole/.spack/package_repos/fncqgg4/repos/spack_repo/builtin/packages/kokkos/package.py:405, in cmake_args:
        402        if spec.satisfies("~wrapper+cuda") and not (
        403            spec.satisfies("%clang") or spec.satisfies("%cce") or spec.satisfies("+cmake_lang")
        404        ):
  >>    405            raise InstallError(
        406                "Kokkos requires +wrapper when using +cuda without %clang, %cce or +cmake_lang"
        407            )
        408

See build log for details:
  /local/home/tpadiole/tmp_spack/spack-stage/spack-stage-kokkos-5.0.2-6jmjl5ow6nviaiwzselltkjdwfk5hded/spack-build-out.txt
```

After, concretization error:
```
spack install kokkos +cuda cuda_arch=80 ~wrapper ~cmake_lang target=x86_64 %gcc
==> Error: failed to concretize `kokkos~cmake_lang+cuda~wrapper cuda_arch=80 target=x86_64 %gcc` for the following reasons:
     1. 'kokkos' requires conflicting variant values '~cmake_lang' and '+cmake_lang'
     2. 'kokkos' requires conflicting variant values '~cmake_lang' and '+cmake_lang'
        required because kokkos~cmake_lang+cuda~wrapper cuda_arch=80 target=x86_64 %gcc requested explicitly 
        required because +cmake_lang is a requirement for package kokkos when +cuda~wrapper 
          required because kokkos~cmake_lang+cuda~wrapper cuda_arch=80 target=x86_64 %gcc requested explicitly 
```

### DIFF
--- a/repos/spack_repo/builtin/packages/kokkos/package.py
+++ b/repos/spack_repo/builtin/packages/kokkos/package.py
@@ -331,6 +331,7 @@ class Kokkos(CMakePackage, CudaPackage, ROCmPackage):
     depends_on("kokkos-nvcc-wrapper@develop", when="@develop+wrapper")
     conflicts("+wrapper", when="~cuda")
     conflicts("+wrapper", when="+cmake_lang")
+    requires("%clang", "%cce", "+cmake_lang", policy="any_of", when="~wrapper+cuda")
 
     # TODO new major: update c++ std
     with default_args(multi=False, description="C++ standard"):
@@ -442,13 +443,6 @@ class Kokkos(CMakePackage, CudaPackage, ROCmPackage):
     def cmake_args(self):
         spec = self.spec
         from_variant = self.define_from_variant
-
-        if spec.satisfies("~wrapper+cuda") and not (
-            spec.satisfies("%clang") or spec.satisfies("%cce") or spec.satisfies("+cmake_lang")
-        ):
-            raise InstallError(
-                "Kokkos requires +wrapper when using +cuda without %clang, %cce or +cmake_lang"
-            )
 
         options = [
             from_variant("CMAKE_POSITION_INDEPENDENT_CODE", "pic"),


thomas-bouvier:
This looks better to me.

repos/spack_repo/builtin/packages/kokkos/package.py
```diff
@@ -331,6 +331,7 @@ class Kokkos(CMakePackage, CudaPackage, ROCmPackage):
     depends_on("kokkos-nvcc-wrapper@develop", when="@develop+wrapper")
     conflicts("+wrapper", when="~cuda")
     conflicts("+wrapper", when="+cmake_lang")
+    requires("%clang", "%cce", "+cmake_lang", policy="any_of", when="~wrapper+cuda")
```
cedricchevalier19, line 334: Can we check whether clang has a CUDA backend enabled?

lucbv (approved):

cedricchevalier19 (approved):
Looks better and harder to misuse.

repos/spack_repo/builtin/packages/kokkos/package.py
tpadioleau: I don't know, i know the cuda variant of llvm is deprecated

rbberger: You need to check for the nvptx target support in LLVM.

rbberger:
@tpadioleau You can use `~wrapper+cuda` and not pull in `kokkos-nvcc-wrapper`. This still defaults to the launch_compiler being used. The recipe is a bit too strict actually.

repos/spack_repo/builtin/packages/kokkos/package.py
tpadioleau: Yeah indeed, and `all` I guess ? Do we want to try in this PR ?

tpadioleau:
> @tpadioleau You can use `~wrapper+cuda` and not pull in `kokkos-nvcc-wrapper`. This still defaults to the launch_compiler being used. The recipe is a bit too strict actually.

I agree that we don't need to pass the wrapper explicitly but it is still used under the hood. I think it depends how we define this variant.

tpadioleau:
What should we do with this PR ?

bernhardkaindl:
@rbberger, @tpadioleau asked:
> What should we do with this PR ?

Since you looked at it, maybe you can provide further suggestion for @tpadioleau?

repos/spack_repo/builtin/packages/kokkos/package.py
rbberger: I would say the additional constraint can be in another PR.

rbberger:
My concern is still that the restriction would be too strong. +cuda~wrapper%gcc is a valid common case, where the launch compiler does the right thing. We're actually thinking of getting rid of the separate nvcc_wrapper package. this would work against that. I'll put this up for a vote in our Kokkos Build and Packaging WG, since I can see this going either way. My main workflows use clang, not nvcc, so I can be swayed either way.

rbberger:
We could of course decouple +/~wrapper from meaning to pull in the nvcc wrapper package moving forward.

rbberger (approved):
I'm realizing that all my reservations are about the current state and this PR only moves the failure from install to concretization. So I won't block this longer and let the bigger picture issue be solved in another PR.

tpadioleau:
Thanks @rbberger. I must confess that the meaning of the variant `wrapper` is not anymore clear to me, in particular if `~wrapper` would still allow Kokkos to use the wrapper through the launcher. I think it needs to be discussed :)

PR #5174, opened 2026-06-11
Restore py-cylc-uiserver (new versions) + dependency updates

climbfuji:
## Description

This PR restores the previously deleted `py-cylc-uiserver` package and updates its dependencies.

I picked two versions of interest, the last version compatible with Python 3.11 (`py-cylc-uiserver@1.7.1`), and the current version (`py-cylc-uiserver@1.9.1`). For comparison, the latest version in the package that got removed last year was `1.6.1`.

Notes on the dependency updates:
- `py-graphql-core` falsely declared `py-poetry` instead of `py-poetry-core` as a dependency
- `py-graphql-relay` was missing the `py-poetry-core` build dependency
- Deprecated older versions of `py-cylc-{flow,rose}`. Since I don't have any need for and interest in `py-cylc-rose`, I am going to let this package expire (deprecate sequentially, remove once all deprecated)
- Add `py-urwid@2.6.16`, since `py-cylc-flow@8.5:` requires a newer version that the latest one available
- Added myself as maintainer to some of these packages

## Testing

- Built and ran `py-cylc-uiserver@1.7.1` with Python 3.11 on my devbox with `gcc@14.2.1`
- Built and ran `py-cylc-uiserver@1.9.1` with Python 3.12 on my devbox with `gcc@14.2.1`
- Built and ran `py-cylc-uiserver@1.9.1` with Python 3.13 on my devbox with `gcc@14.2.1`


climbfuji:
@adamjstewart and @haampie I would appreciate your eyes on this one. Python dependencies are a tricky business, and I am sure I got something wrong. Thanks!

climbfuji:
@spackbot fix style

repos/spack_repo/builtin/packages/py_cylc_flow/package.py
```diff
@@ -18,34 +18,70 @@ class PyCylcFlow(PythonPackage):
...
-    # canonical by Spack - https://github.com/spack/spack/issues/48479
-    version("8.4.2", commit="5625b0fa594b5a422b37b0a3b85dc12e40fceb05")
-    version("8.3.6", commit="7f63b43164638e27636b992b14b3fa088b692b94")
-    version("8.2.3", sha256="dd5bea9e4b8dad00edd9c3459a38fb778e5a073da58ad2725bc9b84ad718e073")
-    version("8.2.0", sha256="cbe35e0d72d1ca36f28a4cebe9b9040a3445a74253bc94051a3c906cf179ded0")
-    version("8.1.4", sha256="d1835ac18f6f24f3115c56b2bc821185484e834a86b12fd0033ff7e4dc3c1f63")
+    version(
+        "8.6.4",
+        sha256="ba65eee781b26986030ab2ab6bd44bed5c77f4164ae02f5b093b3e452cab35a7",
+        url="https://files.pythonhosted.org/packages/source/c/cylc-flow/cylc_flow-8.6.4.tar.gz",
```
adamjstewart: If the URL format has changed, you should use a `url_for_version` to encode this and update the default `url` (in this case `pypi`).

repos/spack_repo/builtin/packages/py_cylc_flow/package.py
```diff
@@ -18,34 +18,70 @@ class PyCylcFlow(PythonPackage):
...
+    version(
+        "8.4.2",
+        sha256="6d0b3129c9779db764fc6e4b4c397a5b117accdf4e9089cd5edcd5023618aa17",
+        url="https://files.pythonhosted.org/packages/source/c/cylc-flow/cylc_flow-8.4.2.tar.gz",
+    )
+    version(
+        "8.3.6",
+        sha256="43ddda5fa047a6cd19bd1eea6e7ece028f9d535eecce004203a46c7c135d2d8c",
+        url="https://files.pythonhosted.org/packages/source/c/cylc-flow/cylc_flow-8.3.6.tar.gz",
+        deprecated=True,
```
adamjstewart: Could use:
```python
    version(...)
    ...
```
to have less duplication.

adamjstewart (commented):
Let's fix the URLs first, then I'll do a more detailed review.

repos/spack_repo/builtin/packages/py_cylc_flow/package.py
climbfuji: Done, please check

repos/spack_repo/builtin/packages/py_cylc_flow/package.py
climbfuji: Done, please check

repos/spack_repo/builtin/packages/py_cylc_flow/package.py
```diff
@@ -11,41 +11,57 @@ class PyCylcFlow(PythonPackage):
...
+    version("8.6.4", sha256="ba65eee781b26986030ab2ab6bd44bed5c77f4164ae02f5b093b3e452cab35a7")
+    version("8.5.4", sha256="3997e51bb31905027d96374e29d0e61eb84bd7dd816cd6c426eae3467a984bb2")
+    version("8.4.2", sha256="6d0b3129c9779db764fc6e4b4c397a5b117accdf4e9089cd5edcd5023618aa17")
+    with default_args(deprecated=True):
+        version("8.3.6", sha256="43ddda5fa047a6cd19bd1eea6e7ece028f9d535eecce004203a46c7c135d2d8c")
+        version("8.2.3", sha256="dd5bea9e4b8dad00edd9c3459a38fb778e5a073da58ad2725bc9b84ad718e073")
+        version("8.2.0", sha256="cbe35e0d72d1ca36f28a4cebe9b9040a3445a74253bc94051a3c906cf179ded0")
+        version("8.1.4", sha256="d1835ac18f6f24f3115c56b2bc821185484e834a86b12fd0033ff7e4dc3c1f63")
 
+    def url_for_version(self, version):
```
adamjstewart: I usually put functions at the end of the file. Optional style thing

climbfuji: I'll make this change, no problem; I'll also check the other files modified in this PR.

climbfuji:
@adamjstewart I pushed the cosmetic changes you asked for. I also fixed a bug that I uncovered during my testing. cylc-flow@8.5.4 needed a newer version of `py-urwid`, which I had to add first.

adamjstewart:
@spackbot run pipeline

repos/spack_repo/builtin/packages/py_graphql_core/package.py
```diff
@@ -14,17 +14,30 @@ class PyGraphqlCore(PythonPackage):
...
 
     depends_on("python@3.6:3", type=("build", "run"))
-    depends_on("py-poetry@1", when="@3:", type="build")
+    depends_on("py-poetry-core@1:2", when="@3.2:", type="build")
+    depends_on("py-poetry-core@1", when="@3:3.1", type="build")
+    depends_on("py-setuptools@59:80", when="@3", type="build")
     depends_on("py-setuptools", when="@2", type="build")
     depends_on("py-six@1.10.0:", type=("build", "run"), when="@2.3.2")
     depends_on("py-promise@2.3:2", type=("build", "run"), when="@2.3.2")
     depends_on("py-rx@1.6:1", type=("build", "run"), when="@2.3.2")
```
adamjstewart, line 36: Missing a dep on typing-extensions

repos/spack_repo/builtin/packages/py_graphql_relay/package.py
```diff
@@ -13,13 +13,18 @@ class PyGraphqlRelay(PythonPackage):
...
+    maintainers("LydDeb", "climbfuji")
 
     license("MIT")
 
+    version("3.2.0", sha256="1ff1c51298356e481a0be009ccdff249832ce53f30559c1338f22a0e0d17250c")
     version("2.0.1", sha256="870b6b5304123a38a0b215a79eace021acce5a466bf40cd39fa18cb8528afabb")
 
+    depends_on("python@3.6:3", type=("build", "run"))
+    depends_on("py-poetry-core@1", when="@3", type="build")
     depends_on("py-setuptools", type="build")
```
adamjstewart, line 25: More specific in newer versions: setuptools>=59,<70

repos/spack_repo/builtin/packages/py_jaraco_functools/package.py
```diff
@@ -11,13 +11,23 @@ class PyJaracoFunctools(PythonPackage):
...
-    pypi = "jaraco.functools/jaraco.functools-2.0.tar.gz"
+    pypi = "jaraco.functools/jaraco_functools-4.1.0.tar.gz"
 
     license("MIT")
 
+    version("4.1.0", sha256="70f7e0e2ae076498e212562325e805204fc092d7b4c17e0e86c959e249701a9d")
     version("2.0", sha256="35ba944f52b1a7beee8843a5aa6752d1d5b79893eeb7770ea98be6b637bf9345")
 
     depends_on("py-setuptools", type="build")
     depends_on("py-setuptools-scm@1.15.0:", type="build")
```
adamjstewart, line 21: Need to be updated

repos/spack_repo/builtin/packages/py_setuptools/package.py
```diff
@@ -74,8 +74,11 @@ class PySetuptools(Package, PythonExtension):
 
         # https://github.com/pypa/setuptools/issues/3661
         depends_on("python@:3.11", when="@:67")
+        depends_on("python@:3.13", when="@:80")
 
     depends_on("py-pip", type="build")
+    # setuptools@80: ships _distutils which imports jaraco.functools.splat (added in 4.0)
+    depends_on("py-jaraco-functools@4:", when="@80:", type=("build", "run"))
```
adamjstewart: jaraco-functools depends on setuptools, this would introduce a circular dependency.

repos/spack_repo/builtin/packages/py_setuptools/package.py
```diff
@@ -74,8 +74,11 @@ class PySetuptools(Package, PythonExtension):
 
         # https://github.com/pypa/setuptools/issues/3661
         depends_on("python@:3.11", when="@:67")
+        depends_on("python@:3.13", when="@:80")
```
adamjstewart: Where does this constraint come from?

repos/spack_repo/builtin/packages/py_urwid/package.py
```diff
@@ -15,9 +15,16 @@ class PyUrwid(PythonPackage):
...
 
+    version("2.6.16", sha256="93ad239939e44c385e64aa00027878b9e5c486d59e855ec8ab5b1e1adcdb32a2")
     version("2.1.2", sha256="588bee9c1cb208d0906a9f73c613d2bd32c3ed3702012f51efe318a3f2127eae")
     version("1.3.0", sha256="29f04fad3bf0a79c5491f7ebec2d50fa086e9d16359896c9204c6a92bc07aba2")
 
     depends_on("c", type="build")  # generated
 
     depends_on("py-setuptools", type="build")
+    depends_on("py-setuptools@61:", type="build", when="@2.6:")
+    depends_on("py-setuptools-scm@7:", type="build", when="@2.6:")
```
adamjstewart:
```suggestion
    depends_on("py-setuptools-scm@7:+toml", type="build", when="@2.6:")
```

adamjstewart (changes requested):

repos/spack_repo/builtin/packages/py_setuptools/package.py
climbfuji: This was added in https://github.com/pypa/setuptools/pull/4649

See lines 10-11 in https://github.com/pypa/setuptools/blame/v80.0.0/setuptools/_distutils/_modified.py

Are you saying that I need to remove the dependency on setuptools in the py-jaraco-functools package?

I was able to install cylc and its dependencies. Let me redo the installs and check which versions are being picked up.

repos/spack_repo/builtin/packages/py_setuptools/package.py
climbfuji: That's a mistake. I must have added this line in my trial-and-error to get past concretization and build errors, and then forgot about it. Thanks for noticing, will remove.

repos/spack_repo/builtin/packages/py_graphql_relay/package.py
climbfuji: I believe I had to remove this overly conservative constraint to concretize the environments due to otherwise conflicting version ranges. But let me check again. I think it's a similar story in py-graphql-core (existing code).

repos/spack_repo/builtin/packages/py_setuptools/package.py
adamjstewart: It's vendored in the setuptools source code: https://github.com/pypa/setuptools/tree/main/setuptools/_vendor/jaraco/functools

So you can remove this change.

climbfuji: Let me try. I know I hit an error, that's why I added it.

climbfuji: I can confirm that this is not needed. I built three environments of cylc-flow + cylc-uiserver with Python 11, 12, 13 without this change and without `depends_on("python@:3.13", when="@:80")` and they all worked fine. Will remove.

repos/spack_repo/builtin/packages/py_graphql_relay/package.py
climbfuji: >=59 <70 is too restrictive and won't allow you to build py-cylc-uiserver without duplicate p-setuptools. We can't have that, because we need to create ONE module for setuptools so that downstream processes can load the spack environment modules, create virtual environments and install their own packages. We also need to create a view where conflicts aren't allowed (although `type=build` might be ok for that).

I built everything successfully with @59:80, therefore I suggest adding this version constraint and a note why not @59:69.

```
    # 3.2.0 wants @59:69, but works fine with @59:80 (matches py-graphql-core@3)
    depends_on("py-setuptools@59:80", when="@3", type="build")
```

repos/spack_repo/builtin/packages/py_jaraco_functools/package.py
climbfuji: I am actually going to revert the changes in `repos/spack_repo/builtin/packages/py_jaraco_functools/package.py`, because that package (and new version) is no longer required (https://github.com/spack/spack-packages/pull/5174#discussion_r3483798129)

repos/spack_repo/builtin/packages/py_graphql_core/package.py
climbfuji: Added as new line 33

repos/spack_repo/builtin/packages/py_graphql_relay/package.py
climbfuji: see new lines 26-27

repos/spack_repo/builtin/packages/py_setuptools/package.py
climbfuji: Removed in 74932d16ce0d58e3f137d0a18217b0a29f9acebd

repos/spack_repo/builtin/packages/py_setuptools/package.py
climbfuji: Removed in 74932d16ce0d58e3f137d0a18217b0a29f9acebd

repos/spack_repo/builtin/packages/py_graphql_relay/package.py
climbfuji: Should probably say that the environment concretized and built with version 80.9.0.

adamjstewart (approved):

PR #5177, opened 2026-06-11
flux: updates for security, core, & sched

sam-maloney:
Update flux packages for the last 5 months, and clarify some dependencies


grondo (approved):
Nice! Thanks @sam-maloney!

sam-maloney:
Realized I was missing a colon in the when clause of `flux_sched/package.py` at L102, because I had copied the previous line which was also missing it :laughing: and there were 2 lines with `when="@0.19.0"` so I changed the second one to `when="@0.20.0"` as that seems to make sense with the release dates and the commit when it was added.

alecbcs (approved):
Looks good to me. Thanks @sam-maloney

alecbcs:
@spackbot run pipeline

PR #5187, opened 2026-06-12
py-tomopy: fix py-tomopy's dependency on py-setuptools

mdorier:
py-tomopy imports `pkg_resources` at run time, this is a module provided by py-setuptools, and which was removed after version 80.9.0. This PR hence makes py-tomopy depend on py-setuptools@:80.9.0.


alecbcs (approved):
Looks good to me. Thanks for the fix @mdorier!

PR #5190, opened 2026-06-12
py-elecsolver: new package

williampiat3:
Small no-arch python package that I would like to have in spack to avoid having to resort to pypi after installing a spack env

### DIFF
--- a/repos/spack_repo/builtin/packages/py_elecsolver/package.py
+++ b/repos/spack_repo/builtin/packages/py_elecsolver/package.py
@@ -1,0 +1,31 @@
+# Copyright Spack Project Developers. See COPYRIGHT file for details.
+#
+# SPDX-License-Identifier: (Apache-2.0 OR MIT)
+
+from spack_repo.builtin.build_systems.python import PythonPackage
+
+from spack.package import *
+
+
+class PyElecsolver(PythonPackage):
+    """Formalizes electric systems as linear problems for temporal and frequency-domain studies."""
+
+    homepage = "https://github.com/williampiat3/ElecSolver"
+    pypi = "ElecSolver/elecsolver-2.0.1.tar.gz"
+
+    maintainers = ["williampiat3"]
+
+    version("2.0.1", sha256="126b02e90e01405109ddd52255a43015d5e0e634945c46baa13c6d41d0e4e05e")
+
+    with default_args(type="build"):
+        depends_on("py-setuptools")
+    # Python dependencies
+    with default_args(type=("run")):
+        depends_on("python@3.9:")
+        depends_on("py-numpy")
+        depends_on("py-scipy")
+        depends_on("py-networkx")
+
+    # Optional/test deps
+    depends_on("py-pytest", type="test")
+    depends_on("py-python-mumps", type="test")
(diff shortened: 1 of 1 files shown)


williampiat3:
@spackbot fix style

repos/spack_repo/builtin/packages/py_elecsolver/package.py
```diff
@@ -0,0 +1,32 @@
...
+from spack.package import *
+
+
+class PyElecsolver(PythonPackage):
+    """Formalizes electric systems as linear problems for temporal and frequency-domain studies."""
+
+    homepage = "https://github.com/williampiat3/ElecSolver"
+    pypi = "ElecSolver/elecsolver-2.0.1.tar.gz"
+
+    maintainers = ["williampiat3"]
```
alecbcs:
```suggestion
    maintainers("williampiat3")
```

Use the maintainers directive here not a list.

repos/spack_repo/builtin/packages/py_elecsolver/package.py
```diff
@@ -0,0 +1,32 @@
...
+
+    version("2.1.0", sha256="17a33dd4d6f8551904baa55c840d561cff00dc046b4bf7e11bad7e69722c4faa")
+    version("2.0.1", sha256="126b02e90e01405109ddd52255a43015d5e0e634945c46baa13c6d41d0e4e05e")
+
+    with default_args(type="build"):
+        depends_on("py-setuptools@80:")
```
alecbcs:
```suggestion
    depends_on("py-setuptools@80:", type="build")
```

I'd drop the use of `default_args()` here since there's only a single dependency.

repos/spack_repo/builtin/packages/py_elecsolver/package.py
```diff
@@ -0,0 +1,32 @@
...
+    # Python dependencies
+    with default_args(type=("run")):
+        depends_on("python@3.9:")
+        depends_on("py-numpy")
+        depends_on("py-scipy")
+        depends_on("py-networkx")
```
alecbcs:
```suggestion
    with default_args(type=("build", "run")):
        depends_on("python@3.9:")
        depends_on("py-numpy")
        depends_on("py-scipy")
        depends_on("py-networkx")
```

You're going to need these packages to also be build dependencies so that they're accessible at build time.

repos/spack_repo/builtin/packages/py_elecsolver/package.py
```diff
@@ -0,0 +1,32 @@
...
+
+    # Optional/test deps
+    depends_on("py-pytest", type="test")
+    depends_on("py-python-mumps", type="test")
```
alecbcs, line 30:
```suggestion
    # Optional/test deps
    depends_on("py-pytest", type="test")
    depends_on("py-python-mumps", type="test")
    
    @run_after("install")
    @on_package_attributes(run_tests=True)
    def install_test(self):
        with working_dir(self.stage.source_path):
            python("-m", "pytest", "tests")
```

We'll also need to add a method here to run pytest after installation to make use of the test dependencies. Otherwise I'd suggest that we remove them.

alecbcs (changes requested):

repos/spack_repo/builtin/packages/py_elecsolver/package.py
williampiat3: you are right python-mumps was just added to spack, I previously did not have the necessary packages available but now I do

williampiat3:
Thank you for your review, I found all modifications valuable and commited them

alecbcs:
@spackbot run pipeline

alecbcs (approved):
Looks good to me. Thanks for applying the suggestions so quickly!

PR #5192, opened 2026-06-12
py-matplotlib: add v3.11.0

adamjstewart:
https://github.com/matplotlib/matplotlib/releases/tag/v3.11.0

### DIFF
--- a/repos/spack_repo/builtin/packages/py_matplotlib/package.py
+++ b/repos/spack_repo/builtin/packages/py_matplotlib/package.py
@@ -24,9 +24,10 @@ class PyMatplotlib(PythonPackage):
         "mpl_toolkits.mplot3d.tests",
     ]
 
-    license("Apache-2.0")
+    license("PSF-2.0")
     maintainers("adamjstewart", "rgommers")
 
+    version("3.11.0", sha256="68c0c7be01b30dcca3638934f7f591df73401235cbdbf0d1ab1c71e7db7f8b57")
     version("3.10.9", sha256="fd66508e8c6877d98e586654b608a0456db8d7e8a546eb1e2600efd957302358")
     version("3.10.8", sha256="2299372c19d56bcd35cf05a2738308758d32b9eaed2371898d8f5bd33f084aa3")
     version("3.10.7", sha256="a06ba7e2a2ef9131c79c49e63dad355d2d878413a0376c1727c8b9335ff731c7")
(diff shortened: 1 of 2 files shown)


repos/spack_repo/builtin/packages/py_matplotlib/package.py
```diff
@@ -24,9 +24,10 @@ class PyMatplotlib(PythonPackage):
         "mpl_toolkits.mplot3d.tests",
     ]
 
-    license("Apache-2.0")
+    license("PSF-2.0")
```
adamjstewart, line 27: Not sure where Apache came from, has it always been PSF?

In `meson.build`, I see:
```meson
  license: 'PSF-2.0 AND MIT AND CC0-1.0 AND OFL-1.1 AND Bitstream-Vera AND Public-Domain',
```
We use external freetype/libraqm/qhull, are there other licenses we need to add?

rgommers: It's always been PSF, except (IIRC) in the very early days when it was something more custom (up to 2012 or so I believe).

Maintainer comment agrees with your expression, except for specifying `Public-Domain` as `LicenseRef-Fedora-Public-Domain` (not sure if that matters, probably not):

https://github.com/matplotlib/matplotlib/issues/28285#issuecomment-2128586808

repos/spack_repo/builtin/packages/py_matplotlib/package.py
```diff
@@ -247,7 +250,9 @@ class PyMatplotlib(PythonPackage):
     depends_on("pkgconfig", type="build")
 
     # C libraries
+    # extern/meson.build
     depends_on("freetype@2.3: build_system=autotools")
+    depends_on("libraqm@0.10.4:", when="@3.11:")
```
rgommers, line 255: https://github.com/matplotlib/matplotlib/blob/v3.11.x/subprojects/libraqm.wrap says `0.10.5` rather than `0.10.4`, not sure if that bump was needed or not (likely just opportunistic, but might be better to match it either way?)

rgommers (approved):
All changes LGTM, one question about `libraqm` version.

repos/spack_repo/builtin/packages/py_matplotlib/package.py
adamjstewart: I would rather avoid pinning more specifically than necessary, makes it difficult to unify the DAG

alecbcs (approved):

PR #5195, opened 2026-06-12
py-pynumpress: add new package

w8jcik:

### DIFF
--- a/repos/spack_repo/builtin/packages/py_pynumpress/package.py
+++ b/repos/spack_repo/builtin/packages/py_pynumpress/package.py
@@ -1,0 +1,22 @@
+# Copyright Spack Project Developers. See COPYRIGHT file for details.
+#
+# SPDX-License-Identifier: (Apache-2.0 OR MIT)
+
+from spack_repo.builtin.build_systems.python import PythonPackage
+
+from spack.package import *
+
+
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/py_pynumpress/package.py
```diff
@@ -0,0 +1,23 @@
...
+    """A wrapper around the MSNumpress library for mass spectrometry numerical data compression."""
+
+    homepage = "https://github.com/mobiusklein/pynumpress"
+    pypi = "pynumpress/pynumpress-0.0.9.tar.gz"
+
+    license("Apache-2.0")
+
+    version("0.0.9", sha256="c0dafd7837cee64fc59eb5fc0f941a1324e631a2d5b2d6ac7e28becfd29bd86e")
+
+    depends_on("python", type=("build", "run"))
```
alecbcs: If there are no boundaries on the python dependency's version we can remove this from here since it is already specified in base `PythonPackage` class.


```suggestion
```

w8jcik: Thank you for the suggestion. Corrected.

alecbcs (approved):
Looks good to me. Thanks @w8jcik!

PR #5203, opened 2026-06-12
adios2: simplify pic/shared variant logic

vicentebolea:
- Prefer conflicts rather than depends(when=) for particular
  incompatibiltiies.
- CMAKE_POSITION_INDEPENDENT_CODE=ON in windows is no-op.


alecbcs:
@vicentebolea can you explain why you moved from the conditional in the variant to a conflict directive?

vicentebolea:
> @vicentebolea can you explain why you moved from the conditional in the variant to a conflict directive?

It is to simplify the logic and also default in +PIC for static libs which it is needed when when a applications consuming adios2 wants to do an shared build which uses libadios2 static libs. Using a conflict rather than a conditional variant seems to be easier on the concretizer and also reflect more succinctly the purpose.

ax3l (approved):

alecbcs (approved):
Looks good to me. Thanks @vicentebolea

PR #5210, opened 2026-06-15
gcc: add patches for 15.3.0 and 16.1.0 on darwin

rbberger:


michaelkuhn (approved):
LGTM, but it would be good if someone with a Darwin system could test this (CC @alalazo @haampie)

alalazo:
No access to Apple MX here. @rbberger I assume you tried to compile both, right?

rbberger:
@alalazo yes, I tested both.

alalazo (approved):

PR #5211, opened 2026-06-15
mt-metis: additional variants for 64-bit indices and position independent code

jpthiele:
This adds the variants `int64` and `pic` to the multithreaded metis build.
Since both are set to false by default this should not affect existing builds.

### DIFF
--- a/repos/spack_repo/builtin/packages/mt_metis/package.py
+++ b/repos/spack_repo/builtin/packages/mt_metis/package.py
@@ -26,6 +26,7 @@ class MtMetis(CMakePackage):
     variant("shared", default=True, description="Enable build of shared libraries")
 
     variant("int64",default=False,description="Use index type of 64 bit")
+    variant("pic",default=False,description="Build with position-independent-code")
     depends_on("c", type="build")  # generated
     depends_on("cxx", type="build")  # generated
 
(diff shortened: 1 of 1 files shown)


bernhardkaindl:
Hi @jpthiele,
> This adds the variants int64 and pic to the multithreaded metis build.
Since both are set to false by default this should not affect existing builds.

Looks fine, but just one nitpick for the argumentation:

AFAICS, from the perspective of a reviewer, as this adds these added `-D` flags to cmake, this is only true if the CMakeLists.txt of the package versions defaults that these flags are already disabled by default?
```py
+    self.define_from_variant("BIGEDGES","64bit"),
+    self.define_from_variant("BIGVERTICES","64bit"),
+    self.define_from_variant("BIGWEIGHTS","64bit"),
+    self.define_from_variant("BIGPARTITIONS","64bit"),
+    self.define_from_variant("CMAKE_POSITION_INDEPENDENT_CODE", "pic")
```
To verify, can you maybe get a log of these options from the CMakeCache.txt file of a build without these flags passed to CMake?

bernhardkaindl:
Hi @jpthiele, see my https://github.com/spack/spack-packages/pull/5211#issuecomment-4977607698, and the review comments :)

jpthiele:
> Hi @jpthiele, see my [#5211 (comment)](https://github.com/spack/spack-packages/pull/5211#issuecomment-4977607698), and the review comments :)

Sorry, had some other things to do. I checked the CMakeLists.txt of mt_metis and all these 
BIGXYZ are of the form

```
if (DEFINED BIGXYZ)
  add_definitions(...) //64bit 
else()
  //32bit
endif()
```

PR #5213, opened 2026-06-15
fenics-basix: Add 0.11.0

jhale:


chrisrichardson (approved):

jhale:
The `mdspan.patch` fixes our bundled mdspan in the 0.10 build when using CUDA compilers. This patch is already applied in 0.11.

kshea21 (approved):
Looks good to me, thanks!

PR #5221, opened 2026-06-15
kokkos_fft: support rocfft backend

yasahi-hpc:
Kokkos-FFT supports rocFFT as a backend for AMD GPU. 
This can be achieved by `KokkosFFT_ENABLE_ROCFFT=ON`.
We extend the current package to support this, for example, by

`spack install kokkos-fft device_backend=rocfft ^kokkos +rocm amdgpu_target=gfx90a`

### DIFF
--- a/repos/spack_repo/builtin/packages/kokkos_fft/package.py
+++ b/repos/spack_repo/builtin/packages/kokkos_fft/package.py
@@ -32,7 +32,7 @@ class KokkosFft(CMakePackage):
     variant(
         "device_backend",
         default="none",
-        values=("none", "cufft", "hipfft", "onemkl"),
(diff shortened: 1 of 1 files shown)


tpadioleau (approved):
Looks ok to me, but it still surprises me we have two backends for the same constructor, rocfft also being the implementation of hipfft.

yasahi-hpc:
> Looks ok to me, but it still surprises me we have two backends for the same constructor, rocfft also being the implementation of hipfft.

Yeah, we may remove `hipfft` support in the future

cedricchevalier19 (approved):

tpadioleau (approved):

cedricchevalier19 (approved):

rbberger (approved):

PR #5223, opened 2026-06-15
Fix and refactor geant4 data

sethrj:
- Specify exact photon evaporation data for G4 11.3, since otherwise G4's cmake variables point to a nonexistent directory (https://github.com/celeritas-project/celeritas/pull/2424)
- Reduce duplication across data packages and improve clarity by refactoring into a base class `Geant4DataPackage`
- Use `setup_run_environment` instead of `setup_dependent_run_environment` so that you can `spack load g4photonevaporation@=6.1` to set only a single environment variable rather than requiring all of them to be loaded.

I can break these up into separate PRs if needed.


repos/spack_repo/builtin/packages/g4photonevaporation/package.py
```diff
@@ -28,25 +28,14 @@ class G4photonevaporation(Package):
...
-            % version
-        )
-
-    @property
-    def g4datasetname(self):
-        spec = self.spec
-        return "PhotonEvaporation{0}".format(spec.version)
+        datasets_url = Geant4DataPackage.datasets_url
+        prefix = "G4"
+        return f"{datasets_url}/{prefix}{self.g4dirname}.{version}.tar.gz"
```
sethrj, line 40: This is a weird exception where the URL, installation dir, and environment variable are all different 🤪

repos/spack_repo/builtin/packages/g4realsurface/package.py
```diff
@@ -24,26 +24,14 @@ class G4realsurface(Package):
...
-                "G4" if version > Version("1.0") else "", version
-            )
-        )
-
-    @property
-    def g4datasetname(self):
-        spec = self.spec
-        return "RealSurface{0}".format(spec.version)
+        datasets_url = Geant4DataPackage.datasets_url
+        prefix = "G4" if version > Version("1.0") else ""
```
sethrj, line 35: And this retains the even weirder exception that it's version-dependent.

repos/spack_repo/builtin/packages/geant4_data/package.py
```diff
@@ -5,10 +5,48 @@
...
+from spack_repo.builtin.build_systems.generic import Package
 
 from spack.package import *
 
 
+class Geant4DataPackage(Package):
+    """Base class to be used by each dependency in Geant4Data"""
+
+    #: URL to parent directory for dataset downloads
+    datasets_url = "https://geant4-data.web.cern.ch/geant4-data/datasets"
```
sethrj, line 24: Note that now all dataset URLs use https not a mix of http/s.

repos/spack_repo/builtin/packages/g4photonevaporation/package.py
drbenmorgan: 😬 I've never been happy with the naming/versioning of these libraries (not that's any consolation....)....

drbenmorgan (dismissed):
Very nice @sethrj - saves a lot of boilerplate and makes the awkward exceptions more obvious. Approving from my side.

repos/spack_repo/builtin/packages/geant4_data/package.py
```diff
@@ -5,10 +5,48 @@
...
+
+    #: Directory name inside 'share' (e.g., G4EMLOW) before version is appended
+    g4dirname: Optional[str] = None
```
wdconinc, line 28: This isn't really optional since every package has to specify it. Is it possible to use a sensible default (__name__.upper based) and only override where needed? A class hierarchy that requires the inheritor to set base class properties for stuff to work is not always ideal.

sethrj: It seems to be the pattern for mixins that add variables: see `builtin/build_systems/{gnu, sourceforge, xorg}.py`, but I should be raising rather than asserting, and use BasePackage as a mixin rather than a direct Package inheritance.

repos/spack_repo/builtin/packages/g4photonevaporation/package.py
wdconinc: Why is this `url_for_version` actually needed? I can see the need in `g4realsurface` due to the prefix change, but why isn't defining the `url` and relying on automatic versioning sufficient?

sethrj:
@drbenmorgan @wdconinc Ready for final review/merge I think!

wdconinc:
[posted earlier but I noticed it was in a resolved comment]

Why is this `url_for_version` actually needed? I can see the need in `g4realsurface` due to the prefix change, but why isn't defining the `url` and relying on automatic versioning sufficient?

sethrj:
> [posted earlier but I noticed it was in a resolved comment]
> 
> Why is this `url_for_version` actually needed? I can see the need in `g4realsurface` due to the prefix change, but why isn't defining the `url` and relying on automatic versioning sufficient?

Sorry I missed it! It's there because all the previous datasets had `url_for_version` with the same syntax, and I was trying to carefully avoid changing functionality. (I used regex replacements for this PR, not gen AI.) I'd prefer to remove the `url` field from each package: I'll try using something like the `pypi -> url` from PythonPackage to autogenerate the default.

sethrj:
@wdconinc I think the reason for the `url_for_version` is that `G4FOO.1.2.3.tar.gz` in the URL gets parsed as version `4FOO.1.2.3`:
```
Preferred version:
    3.3    https://geant4-data.web.cern.ch/geant4-data/datasets/G3.3.tar.gz
```

wdconinc:
@spackbot run pipeline

wdconinc (approved):
LGTM. Restarted pipeline since seemed transient failure. I think I set the CI pipelines to treat geant4-data as external to avoid the downloads so this won't really get exercised there.

sethrj:
@spackbot run pipeline

sethrj:
Shit I just tried a fresh build and it failed because of an error in my last find-replace: I'll push a hotfix

PR #5220, opened 2026-06-15
llvm: allow llvm@:19 to be built with gcc@15:

haampie:
A matter of missing cstdint includes.

I didn't try older than llvm@18, so the next person to try can add back a more relaxed conflict than the one I removed.

PR #5225, opened 2026-06-15
julia: add resources/dep/certs

haampie:
Some resources were fetched during the builds, this fixes it.

It also ensures that Julia picks up the cert.pem from Spack's
ca-certificates-mozilla if `openssl certs=mozilla`.

Re-enable in CI, which requires #5220 to fix `llvm@18 %gcc@15`

### DIFF
--- a/repos/spack_repo/builtin/packages/julia/package.py
+++ b/repos/spack_repo/builtin/packages/julia/package.py
@@ -328,6 +328,43 @@ class Julia(MakefilePackage):
         when="@:1.11 %curl@8.10:",
     )
 
+    # Prevent that JuliaSyntax and TermInfoDB are fetched by the build system
+    resource(
+        url="https://api.github.com/repos/JuliaLang/JuliaSyntax.jl/tarball/4f1731d6ce7c2465fc21ea245110b7a39f34658a",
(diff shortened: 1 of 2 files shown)


repos/spack_repo/builtin/packages/julia/package.py
```diff
@@ -361,16 +398,18 @@ def edit(self, spec, prefix):
...
-            "JULIA_CPU_TARGET:={0}".format(julia_cpu_target),
+            f"prefix:={prefix}",
+            f"MARCH:={march}",
+            f"JULIA_CPU_TARGET:={julia_cpu_target}",
             "USE_BINARYBUILDER:=0",
             "VERBOSE:=1",
             # Spack managed dependencies
             "USE_SYSTEM_BLAS:=1",
             "USE_SYSTEM_CSL:=1",
+            "USE_SYSTEM_NGHTTP2:=1",
```
giordano, line 412: Oh, so, we were actually already building nghttp2, but not linking it correctly. Oops 😅

giordano (approved):

repos/spack_repo/builtin/packages/julia/package.py
haampie: :sweat_smile:

alalazo (approved):

PR #5240, opened 2026-06-17
texlive: fix dependencies, break x, dvipng, metapost into variants

alecbcs:
Refactored Texlive to break out the x, dvipng, and metapost subcomponents into optional variants. Additionally removed the generated dependency on fortran which wasn't required. This now builds successfully on MacOS and Linux.

### DIFF
--- a/repos/spack_repo/builtin/packages/texlive/package.py
+++ b/repos/spack_repo/builtin/packages/texlive/package.py
@@ -93,26 +93,30 @@ class Texlive(AutotoolsPackage):
             when="@{0}".format(release["version"]),
         )
 
-    depends_on("c", type="build")  # generated
-    depends_on("cxx", type="build")  # generated
-    depends_on("fortran", type="build")  # generated
+    variant("doc", default=False, description="Install the documentation files")
+    variant("src", default=False, description="Install the source files")
+    variant("dvipng", default=False, description="Build the dvipng program")
+    variant("metapost", default=False, description="Build MetaPost programs")
+    variant("x", default=False, description="Build X11 programs such as xdvik and xpdfopen")
+
+    depends_on("c", type="build")
+    depends_on("cxx", type="build")
 
     depends_on("pkgconfig", type="build")
 
-    depends_on("cairo+X")
+    depends_on("cairo~X", when="+metapost")
     depends_on("freetype")
-    depends_on("ghostscript")
-    depends_on("gmp")
+    depends_on("gmp", when="+metapost")
     depends_on("harfbuzz+graphite2")
     depends_on("icu4c")
-    depends_on("libgd")
+    depends_on("libgd", when="+dvipng")
     depends_on("libpaper")
     depends_on("libpng")
-    depends_on("libxaw")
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/texlive/package.py
```diff
@@ -122,39 +126,56 @@ class Texlive(AutotoolsPackage):
...
             "--with-system-zziplib",
         ]
 
+        if self.spec.satisfies("+dvipng"):
+            args.remove("--disable-dvipng")
+            args.append("--with-system-gd")
+
+        if self.spec.satisfies("+metapost"):
+            for flag in ("--disable-mp", "--disable-pmp", "--disable-upmp"):
+                args.remove(flag)
```
becker33: Can you please make those added when they're needed instead?

repos/spack_repo/builtin/packages/texlive/package.py
```diff
@@ -122,39 +126,56 @@ class Texlive(AutotoolsPackage):
...
+            for flag in (
+                "--with-system-cairo",
+                "--with-system-gmp",
+                "--with-system-mpfr",
+                "--with-system-pixman",
+            ):
+                if flag not in args:
+                    args.append(flag)
```
becker33: None of this can be in the args already


```suggestion
            args.extend(
                (
                    "--with-system-cairo",
                    "--with-system-gmp",
                    "--with-system-mpfr",
                    "--with-system-pixman",
                )
            )
```

repos/spack_repo/builtin/packages/texlive/package.py
```diff
@@ -122,39 +126,56 @@ class Texlive(AutotoolsPackage):
...
-            "--with-system-mpfr",
-            "--with-system-pixman",
             "--with-system-poppler",
             "--with-system-teckit",
             "--with-system-zlib",
             "--with-system-zziplib",
         ]
 
+        if self.spec.satisfies("+dvipng"):
+            args.remove("--disable-dvipng")
```
becker33: Just add this when it's needed

repos/spack_repo/builtin/packages/texlive/package.py
```diff
@@ -93,26 +93,30 @@ class Texlive(AutotoolsPackage):
...
+    variant("metapost", default=False, description="Build MetaPost programs")
+    variant("x", default=False, description="Build X11 programs such as xdvik and xpdfopen")
+
+    depends_on("c", type="build")
+    depends_on("cxx", type="build")
 
     depends_on("pkgconfig", type="build")
 
-    depends_on("cairo+X")
+    depends_on("cairo~X", when="+metapost")
```
becker33: Are you sure `~X` is required?

alecbcs: Oops good find! After debugging the build the first time I had a `cairo+X` installed locally and clingo kept snapping to it. Meant this to be a debugging fix and accidentally forgot to remove after.

becker33 (approved):

PR #5250, opened 2026-06-18
recola-sm: Add compiler depends.

Mittagskogel:
The `recola-sm` Spack package seems to require C/CXX compiler dependencies. Not sure why, given that the package is only Fortran files. Perhaps CMake tries to check those compilers by default?

In any case, the install works after adding the `depends_on()` lines.


wdconinc (approved):
Thanks. Indeed, the default in https://gitlab.com/recola/models/sm/-/blob/main/CMakeLists.txt?ref_type=heads#L72 is to enable `C` and `CXX` for `LANGUAGES` in CMake. If upstream changes the project directive to include `LANGUAGES FORTRAN` then we can remove this requirement for newer versions.

PR #5262, opened 2026-06-18
heffte: add variant to build and install benchmarks

tteggelit:
Add `benchmarks` boolean variant to build and install the provided heFFTe benchmarks.
Default is `False` so as to maintain previous behavior if omitted.
Because CMake will compile the benchmarks when the proper flag is provided, but not install them, the python `os` module is also now included to create the destination directory to install them in. The destination is set be alongside the provided heFFTe examples - i.e., `share/heffte/benchmarks/`.
heFFTe versions are checked because <v2.2.0 there were fewer benchmarks provided.


mkstoyanov:
Why do you need to install the binary files associated with the benchmarks?

Applications that use heFFTe would link to the library and not the benchmarks, that's why we don't install them. We look at the those as "development tools", i.e., something for the heFFTe team to use.

tteggelit:
> Why do you need to install the binary files associated with the benchmarks?
> 
> Applications that use heFFTe would link to the library and not the benchmarks, that's why we don't install them. We look at the those as "development tools", i.e., something for the heFFTe team to use.

I'm looking to add heFFTe as a benchmarking application in [Ramble](https://github.com/GoogleCloudPlatform/ramble) and the heFFTe provided benchmarks should do nicely for this, I think.

mkstoyanov:
Wouldn't it be better to add the code directly to heFFTe, right now you're setting CMake flag that isn't recognized.

tteggelit:
Sorry, I don't understand by adding the code directly to heFFTe. Can you clarify?

I completely missed that the benchmarks are built by default, so I'd be happy to remove the erroneous CMake flag and update the description to reflect that.

mkstoyanov:
Spack is only a wrapper around the heFFTe build system https://github.com/icl-utk-edu/heffte

We can make the edits there and have benchmarks always installed without having to mess with the spack package. CMake also does things with linking that may be necessary to have the benchmarks function properly after installation.

tteggelit:
Either way is fine with me. Ramble relies on Spack (largely) for software installs anyway, so at the end of the day the Spack package is still necessary for me. My current proposed method doesn't require you to change how you've been handling the heFFTe install process so far, but I defer to your judgement.

tteggelit:
@mkstoyanov Just in case you prefer to keep this variant, I updated the PR to remove the invalid CMake flag and update the variant description as well.

tteggelit:
@mkstoyanov Just checking to see if you've thought any more about this proposed change.

mkstoyanov:
I will update the main heFFTe repo when I get a chance, but I will not back-port the fix. So you can use this for now.

mkstoyanov (approved):

bernhardkaindl (approved):
Approved by maintainer and built on Ubuntu 24.04

PR #5263, opened 2026-06-18
CI: Don't clone the allocation configs for build

kwryankrattiger:
The allocation configs are only required during the generate job. Exclude them from the configs loaded during build to avoid spurious failures.


kwryankrattiger:
The generate job is showing the git include in the blame artifact.

The build jobs for this pipeline were all noop (another bug), but this at least won't break pipeline allocations.

zackgalbreath (approved):

PR #5267, opened 2026-06-19
go: add 1.27rc1

mcmehrtens:
- add Go 1.27rc1

### DIFF
--- a/repos/spack_repo/builtin/packages/go/package.py
+++ b/repos/spack_repo/builtin/packages/go/package.py
@@ -39,8 +39,9 @@ class Go(Package):
 
     maintainers("alecbcs")
 
-    license("BSD-3-Clause")
+    license("BSD-3-Clause", checked_by="mcmehrtens")
 
+    version("1.27rc1", sha256="338f51f557c6db1235a3eeb89a0936af6f4e0d4116291ef3b8e41186bc655334")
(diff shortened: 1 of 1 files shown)


bernhardkaindl (approved):
@mcmehrtens Ack, I checked that go 1.27rc1 is not built by default with this PR branch checked out, still the latest release 1.26.4 is built, so this looks fine from my viewpoint.

bernhardkaindl:
I tested that the build works on Linux/x86_64, and the RC is not built by default.

@alecbcs, could you respond to this PR within a few days if merging this PR should wait?

alecbcs:
@mcmehrtens do you need the release candidate for another specific package? 

I've so far avoided adding release candidates for go to minimize churn in the package. (Since we'll want to deprecate them once the main release is posted.)

I'm alright adding it, but just want to know what the use case is before we take on additional maintenance work.

mcmehrtens:
> @mcmehrtens do you need the release candidate for another specific package?

@alecbcs Nope.

> I've so far avoided adding release candidates for go to minimize churn in the package. (Since we'll want to deprecate them once the main release is posted.)

Ahh, noted. I didn't know this was the workflow.

> I'm alright adding it, but just want to know what the use case is before we take on additional maintenance work.

That's okay. Unless someone else does have a legitimate use, I'm perfectly fine closing the PR.

alecbcs:
I guess let's close out this PR for now, but I appreciate the packaging work you're doing @mcmehrtens!

PR #5270, opened 2026-06-19
Make py-torch +cuda buildable in air-gapped system

### LINKED ISSUE #5271: Installation issue: fp16 not buildable on air-gapped systems

fp16 is pulling `psim` from the master branch
```
    resource(
        name="psimd",
        git="https://github.com/Maratyszcza/psimd.git",
        branch="master",
        destination="deps",
        placement="psimd",
    )
```
The package has not seen a commit in 6 years perhaps this is time for a release

### Error message

Fetch Error

linux rhel9 zen3

### Additional information

_No response_

### General information

### LINKED ISSUE #5272: Installation issue: pthreadpool not buildable on air-gapped systems

`pthreadpool` is pulling ressources from a master branch of `FXdiv`
```python
    resource(
        name="fxdiv",
        git="https://github.com/Maratyszcza/FXdiv.git",
        branch="master",
        destination="deps",
        placement="fxdiv",
    )
```
This prevents building pthreadpool in air-gaped systems.
The repository has not seen a commit in 6 years, a release can be made to have a proper checksumed file

### Error message

Fetch Error on air gaped systems

linux rhel9 zen3

### Additional information

_No response_

### General information

williampiat3:
A want to make pytorch's recipe buildable on air-gapped systems therefore I have identfied the key problems of the recipe that need to be changed in order to allow the build without internet:

# Mandatory changes for air-gapped  CUDA builds
- [X] NNPACK attempts to pull python-six source from the internet without using spack's python: FIXED by adding lines in the patch function of PyTorch class 
- [X] `fp16` (a dependency) needs to pull `psimd` from github: FIX: add commit in ressource so that a fetched tar source can be reused on a air-gapped system
- [X] `pthreadpool` (a dependency) needs to pull `FXdiv` from github FIX: add commit in ressource so that a fetched tar source can be reused on a air-gapped system

# For ROCm builds
Seing the scale of the modifications for ROCM build (and the number of people involved) I split this PR
See issue #5302 for ROCm tracking

# Non mandatory changes raised in this PR:
- [X] Fixing NNPACK and QNNPACK , the issue was raised in this PR but is not mandatory for the air-gapped build #5304 

closes #5272  
closes #5271


williampiat3:
If issues #5271 and #5272 are solved pytorch will be buildable in an air gaped system with a spack mirror only

repos/spack_repo/builtin/packages/py_torch/package.py
```diff
@@ -373,15 +373,13 @@ class PyTorch(PythonPackage, CudaPackage, ROCmPackage):
 
     # https://github.com/pytorch/pytorch/issues/172630 (GCC-14.2 ICE for aarch64)
     patch(
-        "https://github.com/pytorch/pytorch/commit/8fd509399e25cb4b265dff663d3f777406001f2e.patch?full_index=1",
-        sha256="91d0470cc05f5f0f775f32b70f174af74f5607162852ba1bcdd81381cd735f24",
+        "pytorch-pytorch-8fd509399e25cb4b265dff663d3f777406001f2e.patch",
```
adamjstewart: We typically prefer patches downloaded from official sources instead of adding them to spack, as the license of these patches is ambiguous. Can you explain why patch URLs are not allowed for air-gapped systems? That's likely something we can fix instead of adding 1,700 lines of code to Spack.

repos/spack_repo/builtin/packages/py_torch/package.py
```diff
@@ -600,6 +571,31 @@ def patch(self):
             "torch_global_deps PROPERTIES LINKER_LANGUAGE CXX",
             "caffe2/CMakeLists.txt",
         )
+        # NNPACK needs to be able to find the six package, which is a dependency of peachpy.
+        # The peachpy CMake module sets the variable PYTHON_SIX_SOURCE_DIR,
+        # but NNPACK does not forward that variable to its submodules.
+        # The following two patches make NNPACK use the variable set in the Spack environment.
```
adamjstewart: Can we patch this upstream instead? My goal is that Spack's package is as simple as possible so that future versions will have almost no patches someday.

repos/spack_repo/builtin/packages/py_torch/package.py
```diff
@@ -742,6 +738,16 @@ def enable_or_disable(variant, keyword="USE", var=None):
...
         # cmake/External/nnpack.cmake
         enable_or_disable("nnpack")
+        if "+nnpack" in self.spec and "py-six" in self.spec:
+            # NNPACK/PeachPy wires this path into PYTHONPATH for codegen.
+            # Point it at Spack's installed py-six to avoid network fetches.
+            py_mm = str(self.spec["python"].version.up_to(2))
+            env.set(
+                "PYTHON_SIX_SOURCE_DIR",
+                join_path(
+                    self.spec["py-six"].prefix, "lib", "python{0}".format(py_mm), "site-packages"
```
adamjstewart: Unfortunately this won't work in general, every OS/distro/Python-version has a different style for this directory name. For example:

* `lib/pythonX.Y/site-packages` on most POSIX systems                                     
* `lib64/pythonX.Y/site-packages` on RHEL/CentOS/Fedora with system Python                
* `lib/pythonX/dist-packages` on Debian/Ubuntu with system Python                         
* `lib/python/site-packages` on macOS with framework Python                               
* `Lib/site-packages` on Windows                  

I would instead use `self["py-six"].module.python_purelib` which takes this into account.

adamjstewart:
Glad someone is working on air-gapped PyTorch, this has been on my TODO list for a long time!

repos/spack_repo/builtin/packages/py_torch/package.py
williampiat3: Perhaps this is due to the fact that the patch I used (pytorch-pytorch-8fd509399e25cb4b265dff663d3f777406001f2e.patch) was not in the spack mirror. yet pytorch 2.10 (the one available in release 2026.03.0) was already using it. 
Maybe this is just a question of time for it to be made available

adamjstewart: @scheibelp @becker33 do you know the answer to this question, or do you know who would?

repos/spack_repo/builtin/packages/py_torch/package.py
williampiat3: NNPACK is available through spack but suffer the exact same issue than fp16 and pthreadpoll: needing ressources from github master branches instead of release.
However I do believe your are right and that we should ultimately :
* patch pytorch cmake file to make use of a prebuilt NNPACK
* Make proper release of NNPACK's resources (peachpy, cpuinfo, fp16, fxdiv, psimd and pthreadpool)

repos/spack_repo/builtin/packages/py_torch/package.py
williampiat3: Maybe it is a missing patch in spack-mirror, I am puzzled about the logic of the spack-cache: I though I would find in there every source up to a release but this patch dates back before 2026.03.0... 
I tried to pull this patch: https://github.com/pytorch/pytorch/commit/8fd509399e25cb4b265dff663d3f777406001f2e.patch?full_index=1 with sha256 91d0470cc05f5f0f775f32b70f174af74f5607162852ba1bcdd81381cd735f24 and it wasn't available in the spack cache

williampiat3: I have made the paches changes in 1 commit, it would be easy to revert if need be

repos/spack_repo/builtin/packages/py_torch/package.py
williampiat3: Fixed

williampiat3:
@spackbot fix style

williampiat3:
@adamjstewart:
* I reverted the changes on the patches: they were all checksumed therefore they can be consistently pulled and provided to an air-gapped system
* added commits to the fp16 and pthread package for the fxdiv and psimd package that were forcing the air gapped system to attempt to pull from github: now it can use a locally fetched tar file corresponding to the last commit (dating from 6 years ago)

I think that's it actually!

I just have to tweak the patch on my air-gapped system because the patch I am trying to pull is not in the spack mirror but in the end it will be available in there so I am ok with this! it works on my air-gapped system like this but I have not tested rocm builds

repos/spack_repo/builtin/packages/py_torch/package.py
williampiat3: [NNPACK spack recipe](https://github.com/spack/spack-packages/blob/develop/repos/spack_repo/builtin/packages/nnpack/package.py) points towards https://github.com/Maratyszcza/cpuinfo.git for cpuinfo resource which is a dead link. I kind of understand why pytorch recipe is building NNPACK itself

What should we do about it? Delete it since this is broken and unused?
nnpack

repos/spack_repo/builtin/packages/py_torch/package.py
williampiat3: reverted the changes in the end

repos/spack_repo/builtin/packages/py_torch/package.py
williampiat3: Perhaps I should just make an issue on that problem which we could consider unrelated to this PR

williampiat3:
Currently testing ROCm build (I can't test ROCm running sadly since I don't have an AMD GPU...)

List of changes:
* hipblas-common is trying to pull rocm-cmake from the web: FIX: add `depends_on(rocm-cmake)` in hiplas-commun

TODO:
* Add SPIRV headers to spack https://github.com/KhronosGroup/SPIRV-Headers and add it as a dependency of llvm-amdgpu (llvm-amdgpu is trying to pull it from the web if not available). I will make a separated PR for this!

EDIT: I will not keep this message updated, please check the description for status and ongoing work

afzpatel:
> Currently testing ROCm build (I can't test ROCm running sadly since I don't have an AMD GPU...)
> 
> List of changes:
> 
> * hipblas-common is trying to pull rocm-cmake from the web: FIX: add `depends_on(rocm-cmake)` in hiplas-commun
> 
> TODO:
> 
> * Add SPIRV headers to spack https://github.com/KhronosGroup/SPIRV-Headers and add it as a dependency of llvm-amdgpu (llvm-amdgpu is trying to pull it from the web if not available)

Thanks for doing this, I'm currently working on this as well and just discovered some of the ROCm related failures. I plan to add them to: https://github.com/spack/spack-packages/pull/5244

repos/spack_repo/builtin/packages/py_torch/package.py
adamjstewart: Looks like it was moved to https://github.com/pytorch/cpuinfo? Would also need to update qnnpack. I would make separate PRs for everything if you can, makes it easier to review.

repos/spack_repo/builtin/packages/py_torch/package.py
scheibelp: > Perhaps this is due to the fact that the patch I used (pytorch-pytorch-8fd509399e25cb4b265dff663d3f777406001f2e.patch) was not in the spack mirror. yet pytorch 2.10 (the one available in release 2026.03.0) was already using it.

`spack mirror create` should be able to create a personal mirror with the original patch (`https://github.com/pytorch/pytorch/commit/8fd509399e25cb4b265dff663d3f777406001f2e.patch?full_index=1`) regardless of whether it's in the spack source mirror (it should store that file locally). If it can't, then I think that means either (a)  a bug in `spack mirror` (e.g. the patch is generated based on a concretized spec that doesn't match what would be concretized on the destination system - although I thought the mirror logic accounted for that by just pulling all patches) or (b) you are on a system that accords special permissions to the official spack source mirror.

I'm assuming since this change was reverted, that closing this thread was pending an answer to the question (but I'll leave it to the author to decide if this answer is sufficient).

(also I'll update the source mirror, but I don't think that would actually fix whatever problems led to doing this)

williampiat3:
I'll divide this PR in multiple ones for better reviewing.
Making ROCm Pytorch air-gapped is harder than I thought.

williampiat3:
OK @adamjstewart this PR is now only for CUDA air gapped and it works on my side, I hope it is generic enough but I have good hope that it is: changes are limited

repos/spack_repo/builtin/packages/py_torch/package.py
williampiat3: See #5304 for qnnpack and nnpack

williampiat3:
@adamjstewart, is there anything you want me to change for approving this PR? some more testing?

adamjstewart:
Just need to find time to review. For the six/peachpy stuff, can we contribute this upstream so we don't need to patch it forever?

williampiat3:
So I investigated a little bit further to see if we could do it upstream:
Current situation on Pytorch side:
* Pytorch has his own cmake/External/nnpack.cmake that is a replacement of NNPACK's CmakeList.txt
* cmake/External/nnpack.cmake does not allow for system installed NNPACK, it simply builds it from source


Current situation on NNPACK side:
* CmakeList.txt does not allow for system installed fp16 fxdiv
* The code is not maintained anymore, only patches can be made to ensure air-gapped compilation
* As of today the recipe is broken as far as my testing goes

I see three ways out:
1. Make a NNPACK patch for it to build in a air-gapped environement + make a patch on pytorch to use an installed nnpack
2. Make a patch of pytorch's NNPACK install to allow the air-gapped build (This PR's choice)
3. Make a PR on pytorch to have new versions of pytorch buildable in a air-gapped environement and keep previous versions not buildable in such context

But my point is, if we want latest versions to be buildable in a air-gapped environnement (solution 1 or 2) **there is no espaping patching in the build recipe of pytorch**

Solution 3 keeps some versions broken but it seems only to bother me....

adamjstewart:
I'm fine with patching existing versions as long as future versions no longer need patching. How much work is 1?

williampiat3:
> I'm fine with patching existing versions as long as future versions no longer need patching. How much work is 1?

Solution 1:
50 lines of code in the pytorch recipe for the patch
50 lines of code in the nnpack recipe for the patch

I have only tested the working build on this one, I haven't checked if it work

Solution 2:
35 lines of code in the pytorch recipe for the patch and 

it relies on pytorch's nnpack  cmake file so I am pretty confident it will work (this is my totally unbiased opinion as you have guessed)

Solution 3: 
0 lines of code but will only work for new versions of pytorch I made the PR on pytorch because it is necessary I think: https://github.com/pytorch/pytorch/pull/188263

adamjstewart:
I like solution 3 + 1 line of code in Spack to apply this patch to older versions.

williampiat3:
Of course Pytorch changed his env forwarding process at version 2.12 so I need 2 patches:
* One for @2.5:2.11
* One for @2.12

adamjstewart (approved):
Okay, this looks nice and clean and maintainable now. Thanks for all the hard work, and glad it turned out to be relatively simple!

If you want to keep hacking on this in the future, I would love to get to the point where PyTorch no longer uses any vendored dependencies and only uses Spack-installed dependencies. But that isn't required for air-gapped installs, so it should probably be a separate effort. Again, best way to approach that would be to upstream it to PyTorch first and then bring it to Spack later. That way we have less to maintain and everyone benefits.

I would also be interested in air-gapped and non-vendored support for TF and JAX, but I suspect that's a much larger task.

sethrj:
Maybe a naive question, but if NNPACK is a critical dependency to py-torch, why don't they fork it and patch there? It would make more sense to include as an actual package dependency rather than a built-in vendored one.

adamjstewart:
PyTorch doesn't do that, they like git submodules.

sethrj:
I've noticed the LLM and quantum python ecosystems are particularly bad about that… wish it was something we could help solve 👀

PR #5275, opened 2026-06-19
py-scipy: add v1.18.0

adamjstewart:
https://github.com/scipy/scipy/releases/tag/v1.18.0

@rgommers congrats on another big release! I was able to get the fortran-free build working, but didn't test the ilp64 changes. Would also love to get accelerate working but have no idea where to start, need to learn more about packaging macOS system libs.

### DIFF
--- a/repos/spack_repo/builtin/packages/py_scipy/package.py
+++ b/repos/spack_repo/builtin/packages/py_scipy/package.py
@@ -53,9 +53,6 @@ class PyScipy(PythonPackage):
     version("1.7.1", sha256="6b47d5fa7ea651054362561a28b1ccc8da9368a39514c1bbf6c0977a1c376764")
     version("1.7.0", sha256="998c5e6ea649489302de2c0bc026ed34284f531df89d2bdc8df3a0d44d165739")
 
-    # SciPy 1.18 introduces a new experimental option to build without Fortran compilers
-    variant("fortran", default=True, description="Build with Fortran", when="@1.18:")
-
     # Based on wheel availability on PyPI
     with default_args(type=("build", "link", "run")):
         depends_on("python@3.12:3.14", when="@1.18:")
@@ -73,8 +70,7 @@ class PyScipy(PythonPackage):
     with default_args(type="build"):
         depends_on("c")
         depends_on("cxx")
-        depends_on("fortran", when="@1.18: +fortran")
-        depends_on("fortran", when="@:1.17")
+        depends_on("fortran", when="@:1.18")
 
         # Required to use --config-settings
         depends_on("py-pip@23.1:", when="@1.9:")
@@ -156,7 +152,7 @@ class PyScipy(PythonPackage):
     conflicts("%gcc@:4.7", when="@:1.9", msg="SciPy requires GCC >= 4.8")
     conflicts("%apple-clang@:14", when="@1.16:", msg="SciPy requires clang >= 15.0")
     conflicts("%apple-clang@:11", when="@1.14:", msg="SciPy requires clang >= 12.0")
-    conflicts("%apple-clang@:9", when="@1.10:", msg="SciPy requires Apple Clang >= 10")
+    conflicts("%apple-clang@:9", when="@1.10:", msg="SciPy requires clang >= 10.0")
     conflicts(
         "%msvc@:19.19",
         when="@1.10:",
@@ -176,7 +172,7 @@ class PyScipy(PythonPackage):
     # https://github.com/mesonbuild/meson/pull/10909#issuecomment-1282241479
     # Intel OneAPI ifx claims to support -fvisibility, but this does not work.
     # Meson adds this flag for all Python extensions which include Fortran code.
-    conflicts("%oneapi@:2023.0", when="@1.9:")
+    conflicts("%oneapi@:2023.0", when="@1.9:1.18")
     # Unknown build error, version ranges may be incorrect
     conflicts("%oneapi@2024:", when="@:1.8")
 
@@ -249,8 +245,6 @@ class PyScipy(PythonPackage):
                 fortran_std = "none"
 
             settings["setup-args"]["-Dfortran_std"] = fortran_std
-        else:
-            settings["setup-args"]["-D_without-fortran"] = True
 
         return settings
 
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/py_scipy/package.py
```diff
@@ -52,13 +53,13 @@ class PyScipy(PythonPackage):
     version("1.7.1", sha256="6b47d5fa7ea651054362561a28b1ccc8da9368a39514c1bbf6c0977a1c376764")
     version("1.7.0", sha256="998c5e6ea649489302de2c0bc026ed34284f531df89d2bdc8df3a0d44d165739")
 
-    depends_on("c", type="build")
-    depends_on("cxx", type="build")
-    depends_on("fortran", type="build")
+    # SciPy 1.18 introduces a new experimental option to build without Fortran compilers
+    variant("fortran", default=True, description="Build with Fortran", when="@1.18:")
```
adamjstewart: Although this is still enabled by default, I believe the concretizer is configured to consider additional dependencies (e.g., fortran) to be penalized more than non-default variants. So on macOS, if a Fortran compiler isn't already available, Spack will likely use ~fortran by default. @alalazo can you confirm?

repos/spack_repo/builtin/packages/py_scipy/package.py
```diff
@@ -121,28 +140,22 @@ class PyScipy(PythonPackage):
...
     # meson.build
     # https://docs.scipy.org/doc/scipy/dev/toolchain.html#compilers
-    conflicts("%gcc@:7", when="@1.10:", msg="SciPy 1.10-1.13 requires GCC >= 8.0")
-    conflicts("%gcc@:9.0", when="@1.14:", msg="SciPy 1.14: requires GCC >= 9.1")
+    conflicts("%gcc@:9.0", when="@1.14:", msg="SciPy requires GCC >= 9.1")
+    conflicts("%gcc@:7", when="@1.10:", msg="SciPy requires GCC >= 8.0")
     conflicts("%gcc@:4.7", when="@:1.9", msg="SciPy requires GCC >= 4.8")
+    conflicts("%apple-clang@:14", when="@1.16:", msg="SciPy requires clang >= 15.0")
+    conflicts("%apple-clang@:11", when="@1.14:", msg="SciPy requires clang >= 12.0")
     conflicts("%apple-clang@:9", when="@1.10:", msg="SciPy requires Apple Clang >= 10")
```
adamjstewart: @rgommers is this only for Apple Clang or also for LLVM Clang? I know they used to have very different numbering schemes, not sure if that changed.

repos/spack_repo/builtin/packages/py_scipy/package.py
```diff
@@ -186,6 +199,9 @@ def archive_files(self):
 
     @run_before("install")
     def set_fortran_compiler(self):
```
adamjstewart, line 197: Is this setup.cfg logic required at all in scipy 1.9+? If not, we can only use this for 1.8 and older.

repos/spack_repo/builtin/packages/py_scipy/package.py
rgommers: I think both, but I can't quite remember. Right now, the versions are back in sync I believe, my Apple Clang tells me it's at 21.0.0 which is up-to-date for LLVM. There was a point where it was four versions behind IIRC.

The check never gave us problems, so this `conflicts` should be safe to keep either way.

repos/spack_repo/builtin/packages/py_scipy/package.py
```diff
@@ -214,28 +231,29 @@ def setup_build_environment(self, env: EnvironmentModifications) -> None:
...
 
+        if "fortran" in self.spec:
+            if spec.satisfies("%gcc"):
+                fortran_std = "legacy"
+            else:
+                fortran_std = "none"
+
+            settings["setup-args"]["-Dfortran_std"] = fortran_std
+        else:
+            settings["setup-args"]["-D_without-fortran"] = True
```
rgommers: Note that this option will be valid for 1.18.x only; it's already deleted in `main` together with the last trace of `.f` code.

repos/spack_repo/builtin/packages/py_scipy/package.py
rgommers: It's not very experimental, just private (starts with `_`) - I'm sure the implementation is solid. The effect it has is to disappear the niche deprecated `scipy.odr` submodule. That should be fine for folks opting into it, it's rarely used.

rgommers:
> didn't test the ilp64 changes.

SciPy itself should be fine with that change, however without also setting `cython-blas-abi=lp64`, this will be breaking for scikit-learn<1.9.0 (xref https://github.com/scikit-learn/scikit-learn/pull/33784) and statsmodels (similar fix, not yet released). So it might be wise to hold off on that for one more release, so the downstream fixes are out. Or, default to `cython-blas-abi=lp64` for now, because that will make the effect not leak beyond SciPy-internal usage.

rgommers:
> ould also love to get accelerate working but have no idea where to start



> Would also love to get accelerate working but have no idea where to start, need to learn more about packaging macOS system libs.

Doesn't that depend on https://github.com/spack/spack/pull/49395? That actually look pretty manageable. It will work for SciPy even without that, but lots of packages need BLAS and most won't be handling the symbol mangling correctly, so I'd be surprised if larger environments work without a common shim layer.

repos/spack_repo/builtin/packages/py_scipy/package.py
adamjstewart: If it's that quick, I'm thinking about removing the variant. Just need to decide whether to make it the default or not. I'm not in a huge rush to remove the Fortran requirement, but as a macOS user, I am excited.

rgommers: Leaving it alone makes sense to me - it's a private option, and I added it for very specific reasons (allowing easier ILP64 builds in 1.18.x/main without having to deal with Fortran). Waiting 6 more months isn't that long.

repos/spack_repo/builtin/packages/py_scipy/package.py
rgommers: It's numpy.distutils-specific, so only <1.9 indeed

rgommers (approved):
LGTM, should be mergeable if CI is happy

adamjstewart:
> Or, default to cython-blas-abi=lp64 for now, because that will make the effect not leak beyond SciPy-internal usage.

This is indeed the default for now.

kshea21 (approved):
Looks good to me, thanks guys

PR #5281, opened 2026-06-20
new-package spirv headers + llvm-amdgpu air-gapped

williampiat3:
This is a step towards making py-torch +rocm builds possible on air-gapped systems, I need to make llvm-amdgpu installable on air-gapped system.
ISSUE: llvm-amdgpu is trying to pull spirv-headers from github
FIX: add spirv-headers to spack and make llvm-amdgpu use it directly

### DIFF
--- a/repos/spack_repo/builtin/packages/llvm_amdgpu/package.py
+++ b/repos/spack_repo/builtin/packages/llvm_amdgpu/package.py
@@ -114,6 +114,7 @@ class LlvmAmdgpu(CMakePackage, LlvmDetection, CompilerPackage):
     depends_on("libdrm", when="@7.1:")
     depends_on("libelf", when="@7.1:")
     depends_on("xxd", when="@7.1:")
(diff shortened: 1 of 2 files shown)


williampiat3:
QUESTION: should I use `resource` in llvm-amdgpu instead of making an extra package?

afzpatel:
> QUESTION: should I use `resource` in llvm-amdgpu instead of making an extra package?

Yes that would be better. `SPIRV-LLVM_TRANSLATOR` is not grabbing any of the releases but rather a specific commit here:
https://github.com/ROCm/SPIRV-LLVM-Translator/blob/rocm-7.2.3/spirv-headers-tag.conf

I've added the changes for this in my PR if you want to take a look:
https://github.com/spack/spack-packages/pull/5244/changes/1a71c43b8e04cc99410623f926ec52dd613497f5

williampiat3:
Then I should probably close this PR!

williampiat3:
closing because @afzpatel has similar changes in #5244

PR #5282, opened 2026-06-20
Package update: air-gapped aotriton (should be split / made smaller)

williampiat3:
This PR aims at making aotriton installable on air gapped system, indeed cmake was creating a virtual env and trying to pull sources from pypi. We avoid this by directly adding build deps in aotriton

I also had build errors that needed a patch, for these ones I extensively used LLMs to make it work on my machine. I would be cautious before merging that it does not generate build errors on other machines or versions


williampiat3:
converted to draft because the changes are not mergeable as is: too many patches I really don't feel confident on this one

PR #5284, opened 2026-06-21
g4data: Fix missing package dep

sethrj:
Failed tabdo in vim because tabs were hidden :'(

FIxes g4 installation from #5223 (maybe need a new audit?)


wdconinc:
> maybe need a new audit?

Or we enable geant4 data installs in CI after all?

wdconinc (approved):

PR #5291, opened 2026-06-22
New version for py-nanobind + switch to pypi for released versions

williampiat3:
New (and old) versions for py-nanobind
Also I changed the released version to fetch source on pypi, it is prefered as stated in the documentation: https://spack.readthedocs.io/en/latest/build_systems/pythonpackage.html#downloading
It is also more convenient for air-gapped systems not to rely on a github commit with submodules.


chrisrichardson (approved):

RikkiButler20 (approved):
LGTM

PR #5294, opened 2026-06-22
GauXC: Update for use with CP2K

mkrack:

### DIFF
--- a/repos/spack_repo/builtin/packages/gauxc/package.py
+++ b/repos/spack_repo/builtin/packages/gauxc/package.py
@@ -32,7 +32,12 @@ class Gauxc(CMakePackage, CudaPackage):
     version("1.1", sha256="17de077fb23e44d03b0ed14dcd8117c01e5b3431fbefa2352d751639ada7f91c")
 
     variant("c", default=True, description="Build with C API support", when="@1.2.dev2: ~master")
-    variant("fortran", default=True, description="Build with Fortran support", when="@dev20260608 ~master")
+    variant(
+        "fortran",
+        default=True,
+        description="Build with Fortran support",
+        when="@dev20260608 ~master",
+    )
     variant("host", default=True, description="Build with host integrator")
     variant("cuda", default=False, description="Build with CUDA support")
     variant("hip", default=False, description="Build with HIP support")
--- a/repos/spack_repo/builtin/packages/gauxc/package.py
+++ b/repos/spack_repo/builtin/packages/gauxc/package.py
@@ -23,7 +23,7 @@ class Gauxc(CMakePackage, CudaPackage):
     license("BSD-3-Clause")
 
     version("master", branch="master")
-    version("dev20260522", branch="skala", commit="69ee054054c642bce0d9b5e8c9c0c13afa97b774")
+    version("dev20260608", branch="skala", commit="36aeab80bf7b2245087bdfeffd8e61b11206b3bd")
     version(
         "1.2.dev2",
         sha256="6659d00522b443f2557a960a9ba0217449d2ee25df809634d6becca23eb0f1ff",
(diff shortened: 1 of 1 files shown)


mkrack:
@awvwgk please, could you review this PR. It includes the latest version of the spack recipe used by CP2K locally. It would be great, if we could sync that recipe with the upstream version so that the local spack recipe in the CP2K tree could be dropped before the next CP2K release.

repos/spack_repo/builtin/packages/gauxc/package.py
```diff
@@ -18,52 +18,111 @@ class Gauxc(CMakePackage, CudaPackage):
...
+    variant("host", default=True, description="Build with host integrator")
+    variant("cuda", default=False, description="Build with CUDA support")
+    variant("hip", default=False, description="Build with HIP support")
+    variant("mpi", default=False, description="Build with MPI support")
+    variant("openmp", default=True, description="Build with OpenMP support")
+    variant("tests", default=False, description="Build with unit tests")
+    variant("gau2grid", default=True, description="Build with Gau2Grid collocation")
+    variant("hdf5", default=True, description="Build with HDF5 support")
+    variant("onedft", default=False, description="Build with ONEDFT functional")
+    variant("skala", default=False, description="Build with Skala support")
```
awvwgk: As mentioned before, `onedft` and `skala` are the same feature in GauXC. It is unfortunate that during the development we didn't rename `onedft` to `skala` in time for GauXC and now have both names.

repos/spack_repo/builtin/packages/gauxc/package.py
```diff
@@ -18,52 +18,111 @@ class Gauxc(CMakePackage, CudaPackage):
...
 
+    variant("c", default=True, description="Build with C API support")
+    variant("fortran", default=True, description="Build with Fortran support")
+    variant("host", default=True, description="Build with host integrator")
+    variant("cuda", default=False, description="Build with CUDA support")
+    variant("hip", default=False, description="Build with HIP support")
+    variant("mpi", default=False, description="Build with MPI support")
+    variant("openmp", default=True, description="Build with OpenMP support")
+    variant("tests", default=False, description="Build with unit tests")
+    variant("gau2grid", default=True, description="Build with Gau2Grid collocation")
```
awvwgk: This is not an actual option that can be used, using `~gau2grid` will abort the build immediately.

repos/spack_repo/builtin/packages/gauxc/package.py
```diff
@@ -18,52 +18,111 @@ class Gauxc(CMakePackage, CudaPackage):
...
+    variant("hdf5", default=True, description="Build with HDF5 support")
+    variant("onedft", default=False, description="Build with ONEDFT functional")
+    variant("skala", default=False, description="Build with Skala support")
+    variant("fast_rsqrt", default=False, description="Build with fast RSQRT support")
+    variant("blas_prefer_ilp64", default=False, description="Prefer ILP64 for host BLAS")
+    variant("link_cuda_static", default=False, description="Link GauXC with static CUDA libs")
+    variant("magma", default=False, description="Build with MAGMA linear algebra support")
+    variant("nccl", default=False, description="Build with NCCL collectives")
+    variant("cutlass", default=False, description="Build with CUTLASS linear algebra support")
+    variant("pic", default=True, description="Build position independent code")
```
awvwgk, line 52: This package has `+pic` as an option but no `+shared`.

repos/spack_repo/builtin/packages/gauxc/package.py
```diff
@@ -18,52 +18,111 @@ class Gauxc(CMakePackage, CudaPackage):
     git = "https://github.com/wavefunction91/GauXC.git"
     url = "https://github.com/wavefunction91/GauXC/archive/refs/tags/v1.1.tar.gz"
 
-    maintainers("awvwgk")
+    maintainers("awvwgk", "mkrack")
 
     license("BSD-3-Clause")
 
     version("master", branch="master")
+    version("dev20260522", branch="skala", commit="69ee054054c642bce0d9b5e8c9c0c13afa97b774")
```
awvwgk: This commit is not on the Skala branch but a PR https://github.com/wavefunction91/GauXC/pull/203

repos/spack_repo/builtin/packages/gauxc/package.py
```diff
@@ -18,52 +18,111 @@ class Gauxc(CMakePackage, CudaPackage):
...
     version(
         "1.2.dev2",
         sha256="6659d00522b443f2557a960a9ba0217449d2ee25df809634d6becca23eb0f1ff",
         url="https://github.com/microsoft/skala/releases/download/v1.1.1/gauxc-skala-r2.tar.gz",
     )
     version("1.1", sha256="17de077fb23e44d03b0ed14dcd8117c01e5b3431fbefa2352d751639ada7f91c")
 
+    variant("c", default=True, description="Build with C API support")
```
awvwgk: This variant is not available in 1.1 yet but only on the master branch.

repos/spack_repo/builtin/packages/gauxc/package.py
```diff
@@ -18,52 +18,111 @@ class Gauxc(CMakePackage, CudaPackage):
...
+    variant("fortran", default=True, description="Build with Fortran support")
```
awvwgk: This variant is not available yet in the master branch

repos/spack_repo/builtin/packages/gauxc/package.py
mkrack: You are right, no need to expose `pic` as variant here.

repos/spack_repo/builtin/packages/gauxc/package.py
mkrack: Thanks for the hint. I will remove the `gau2grid` variant and only keep a hard dependency to the `gau2grid` package.

repos/spack_repo/builtin/packages/gauxc/package.py
mkrack: I will remove the `onedft` variant.

mkrack:
@awvwgk Thanks for the review! I hope that I applied all your change requests properly.

What's the timeline for a new GauXC release including all changes from the `skala` branch?

mkrack:
@awvwgk The CP2K spack recipe applies the `+pic` variant via `prefer:` as default. It turns out that CP2K can only be built without relocation errors during link time when the `+pic` option is passed to the GauXC CMake build. The current recipe version [works with CP2K](https://github.com/cp2k/cp2k/pull/5445) using `dev20260608`.

PR #5299, opened 2026-06-22
Gau2grid: Update for use with CP2K

mkrack:


mkrack:
@awvwgk please, could you review this PR. It includes the latest version of the spack recipe used by CP2K locally. It would be great, if we could sync that recipe with the upstream version so that the local spack recipe in the CP2K tree could be dropped before the next CP2K release.

repos/spack_repo/builtin/packages/gau2grid/package.py
```diff
@@ -9,32 +9,51 @@
...
 
 class Gau2grid(CMakePackage):
-    """Gau2Grid is a library for efficient evaluation of Gaussian
-    basis functions on grids."""
+    """Gau2Grid is a library for the fast computation of a Gaussian and
+    its derivative on a grid.
+    """
 
-    homepage = "https://gau2grid.readthedocs.io"
+    homepage = "https://github.com/psi4/gau2grid"
```
awvwgk: I would prefer to keep the link to the documentation here.

repos/spack_repo/builtin/packages/gau2grid/package.py
```diff
@@ -9,32 +9,51 @@
...
     git = "https://github.com/psi4/gau2grid.git"
-    url = "https://github.com/psi4/gau2grid/archive/refs/tags/v2.0.8.tar.gz"
+    url = "https://github.com/psi4/gau2grid/archive/v2.0.9.tar.gz"
```
awvwgk: Why drop the `refs/tags` from the URL?

mkrack: You are right, it is safer to stick to the modern `refs/tags` format instead of relying on the legacy short path.

mkrack:
@awvwgk Thanks for your quick review! I have applied your change requests.

awvwgk (approved):

repos/spack_repo/builtin/packages/gau2grid/package.py
RMeli: @mkrack FYI, there is no need to change the version in the URL; Spack substitutes the version with the one in the spec (one could put `v0.0.0` as a placeholder).

repos/spack_repo/builtin/packages/gau2grid/package.py
```diff
@@ -9,32 +9,51 @@
...
         args = [
-            self.define("MAX_AM", 8),
-            self.define("ENABLE_XHOST", False),
-            self.define("INSTALL_PYMOD", False),
+            self.define("MAX_AM", spec.variants["max_am"].value),
+            self.define("ENABLE_XHOST", spec.satisfies("+xhost")),
+            self.define("BUILD_FPIC", spec.satisfies("+pic")),
+            self.define("BUILD_SHARED_LIBS", spec.satisfies("+shared")),
+            self.define("ENABLE_GENERIC", spec.satisfies("+generic")),
+            self.define("DISABLE_PRAGMA", spec.satisfies("+no_pragma")),
```
RMeli:
```suggestion
            self.define_from_variant("ENABLE_XHOST", "xhost"),
            self.define_from_variant("BUILD_FPIC", "pic"),
            self.define_from_variant("BUILD_SHARED_LIBS", "shared"),
            self.define_from_variant("ENABLE_GENERIC", "generic"),
            self.define_from_variant("DISABLE_PRAGMA", "no_pragma"),
```

RMeli (dismissed):

repos/spack_repo/builtin/packages/gau2grid/package.py
mkrack: @RMeli You are right, it is not needed to update the placeholder in url, but it is arbitrary and thus `v2.0.9` will work fine like `v0.0.0`.

RMeli (approved):

PR #5301, opened 2026-06-23
Update package: air gapped hipblas common

williampiat3:
add dependency to rocm-cmake to avoid pulling it from github


repos/spack_repo/builtin/packages/hipblas_common/package.py
```diff
@@ -25,6 +25,10 @@ def url_for_version(self, version):
             url = "https://github.com/ROCm/rocm-libraries/archive/rocm-{0}.tar.gz"
         return url.format(version)
 
+    # if rocm-cmake is not installed it pulls from the web
+    # needed for air-gapped environments
+    depends_on("rocm-cmake", type="build")
```
afzpatel: Could you add a for-loop to this so each hipblas-common version depends on the same rocm-cmake version

afzpatel (approved):
LGTM

bernhardkaindl (approved):

PR #5304, opened 2026-06-23
nnpack and qnnpack: Support air-gapped build

williampiat3:
This updates NNPACK and QNNPACK to work in air-gapped systems

### DIFF
--- a/repos/spack_repo/builtin/packages/nnpack/package.py
+++ b/repos/spack_repo/builtin/packages/nnpack/package.py
@@ -46,42 +46,42 @@ class Nnpack(CMakePackage):
     resource(
         name="peachpy",
         git="https://github.com/Maratyszcza/PeachPy.git",
-        branch="master",
+        commit="349e8f836142b2ed0efeb6bb99b1b715d87202e9",
         destination="deps",
         placement="peachpy",
     )
     resource(
         name="cpuinfo",
-        git="https://github.com/Maratyszcza/cpuinfo.git",
-        branch="master",
+        git="https://github.com/pytorch/cpuinfo.git",
+        commit="315d03cacc51bfabe316057b0d3466e13bce88a0",
         destination="deps",
         placement="cpuinfo",
     )
     resource(
         name="fp16",
         git="https://github.com/Maratyszcza/FP16.git",
-        branch="master",
+        commit="782eea126dc5c755827be751a099eb01826175cf",
         destination="deps",
         placement="fp16",
     )
     resource(
         name="fxdiv",
         git="https://github.com/Maratyszcza/FXdiv.git",
-        branch="master",
+        commit="63058eff77e11aa15bf531df5dd34395ec3017c8",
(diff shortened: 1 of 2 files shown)


bernhardkaindl:
@williampiat3, is this still a draft, do you continue with it?

williampiat3:
It took me some time, in a nutshell:
* The @run_before("cmake") needed to have opcodes available for PeachPy (the fix is a bit wobbly... locally changing the PYTHONPATH)
* I changed PeachPy ressource directory to match the one used by PyTorch
* I had to find a commit of FP16 that wasn't generating errors in the build

That was a lot of efforts for a package that has no dependencies but it might be added to the py-torch recipe to avoid having py-torch build it

williampiat3:
@spackbot fix style

repos/spack_repo/builtin/packages/nnpack/package.py
```diff
@@ -95,9 +100,26 @@ class Nnpack(CMakePackage):
...
+            f"{pythonpath}:{old_pythonpath}" if old_pythonpath else pythonpath
+        )
+        try:
+            with working_dir(join_path(deps_dir, "peachpy")):
+                python("setup.py", "generate")
+        finally:
+            if old_pythonpath is None:
+                del os.environ["PYTHONPATH"]
+            else:
+                os.environ["PYTHONPATH"] = old_pythonpath
```
adamjstewart: Ditch all the defensive AI-generated code and use:
```python
def setup_environment(self, env):
    ...
    env.prepend_path("PYTHONPATH", pythonpath)
```

adamjstewart (changes requested):
QNNPACK could technically be a separate PR and be merged immediately.

repos/spack_repo/builtin/packages/nnpack/package.py
williampiat3: I removed the PYTHONPATH change and added a setup_build_environment, this worked but the @run_before("cmake") uses setuptools and works only for py-setuptools@:67 which is rather old.

Pytorch does not install PeachPy using setuptools before building NNPACK: they simply use NNPACK's cmake given PeachPy source dir.

Therefore I am trying to replicate this because it will allow to remove the setuptools@:67   constraint that might be a little bit too restrictive

I have a much simpler recipe than the one I suggested in this PR that builds but I am running the test suite to determine if this is really working

adamjstewart: Simpler is always better, newer setuptools would also be nice. Setuptools has a CVE and I'm trying to deprecate all but the latest version of setuptools.

williampiat3:
@spackbot fix style

williampiat3:
The recipe builds in the current state.
The build did not use opcodes even when tests are built therefore I removed it 
This removes the need for setuptools@:67 in the dependencies

adamjstewart (approved):
Much simpler now, thanks!

PR #5313, opened 2026-06-23
xarray@:2024 requires pandas@:2.1

climbfuji:
## Description

This PR fixes
```
xarray/coding/times.py line 358, in to_timedelta_unboxed
    assert result.dtype == "timedelta64[ns]"
AssertionError
```
by pinning `py-pandas` to versions 2.1 and earlier for `py-xarray` versions 2024 and earlier. 

`py-pandas@2.2` introduced non-nanosecond datetime/timedelta resolutions, but `xarray@2024.7.0`'s `to_timedelta_unboxed` still hard-asserts `[ns]`.


repos/spack_repo/builtin/packages/py_xarray/package.py
```diff
@@ -88,8 +88,8 @@ class PyXarray(PythonPackage):
         depends_on("py-packaging@21.3:", when="@2023.7:")
         depends_on("py-packaging@20:", when="@0.21:")
         depends_on("py-pandas@2.2:", when="@2025.7:")
-        depends_on("py-pandas@2.0:", when="@2024.7:")
-        depends_on("py-pandas@1.4:", when="@2023.7:")
+        depends_on("py-pandas@2.0:2.1", when="@2024.7:")
+        depends_on("py-pandas@1.4:2.1", when="@2023.7:")
```
adamjstewart: When needs upper bounds, otherwise this applies to all versions.

climbfuji: Actually, all that is needed is removing the colons, since we only know 2023.7.0 and 2024.7.0 in spack!

climbfuji:
```suggestion
        depends_on("py-pandas@2.0:2.1", when="@2024.7")
        depends_on("py-pandas@1.4:2.1", when="@2023.7")
```

adamjstewart (approved):

PR #5326, opened 2026-06-25
eccodes: enforce libaec version & set static/shared lib correctly

sibcse:
- starting with eccodes 2.47.0, the minimum required libaec version is 1.1.4: https://confluence.ecmwf.int/spaces/ECC/pages/656639695/ecCodes+version+2.47.0+released
- since eccodes 2.46.0, CMakeLists.txt uses ENABLE_USE_SHARED_LIB_AEC; set this in package.py based on libaec's specs instead of using the default: https://github.com/ecmwf/eccodes/blob/c10e8be0c3e27dd2be36a909b06d5d3be6889708/CMakeLists.txt#L158-L166


climbfuji (approved):
Looks reasonable to me, but I am not the primary maintainer of the package

victoria-cherkas (approved):

bernhardkaindl (approved):

PR #5332, opened 2026-06-25
dyninst: add conflict for dyninst@13 with tbb@2023

mwkrentel:
Intel-tbb 2023.0.0 added a patch that causes dyninst parseapi to segfault with tbbmalloc_proxy.  Now fixed in dyninst master, but :13 fails.

See: https://github.com/dyninst/dyninst/pull/2304

ping @hainest


hainest (approved):
Looks good. Thanks, Mark!

PR #2333, opened 2025-11-07
acts: remove git-lfs dependency

wdconinc:
This PR removes the `acts` dependency on `git-lfs` entirely. The `git-lfs` dependency was originally added for versions 12–13, but support for those deprecated versions (all below `38.x.x`) has since been removed from the recipe. With no remaining versions that require `git-lfs`, the dependency is no longer needed.

Now `acts` doesn't depend on `go` anymore.

PR #5339, opened 2026-06-25
acts: cleanup the last remaining dep for the removed versions

bernhardkaindl:
The initial version of this PR was opened by @wdconinc as #2333:

This PR removes the `acts` dependency on `git-lfs` entirely. The `git-lfs` dependency was originally added for versions 12–13, but support for those deprecated versions (all below `38.x.x`) has since been removed from the recipe. With no remaining versions that require `git-lfs`, the dependency is no longer needed.


wdconinc (approved):
LGTM, but I also submitted the original PR :-) Leaving up to @stephenswat to confirm :-)

stephenswat (approved):
LGTM. :)

PR #5340, opened 2026-06-26
[scorep, otf2, cubelib, cubew] Add new releases

wrwilliams:
PR for the upcoming Score-P 10 release.

@spackbot fix style


wrwilliams:
@spackbot fix style

wrwilliams:
@spackbot fix style

wrwilliams:
@spackbot fix style

wrwilliams:
@spackbot fix style

wrwilliams:
@spackbot fix style

repos/spack_repo/builtin/packages/scorep/fix-intel-mpi-linker-error.patch
melven, line 1: Hi @wrwilliams

I've confirmed the checksums.

Question: why is there a new patch file? It seems to me, it downloads the patch from gitlab - or am I missing something?

Code in the package.py:
```python
   patch(
        "https://gitlab.com/score-p/scorep/-/commit/6d5534de9746e76a28dd9452c3efaa45890067cd.diff",
        when="@10.0 +mpi",
        sha256="20946ff71fb226bbe76b4f6dfff6cea127211d58162247d8e0bebba745917e7d",
    )
```

melven (changes requested):

repos/spack_repo/builtin/packages/scorep/package.py
```diff
@@ -143,11 +161,7 @@ def configure_args(self):
         cname = self.clean_compiler(spec.compiler.name)
         config_args.extend(["--with-nocross-compiler-suite={0}".format(cname)])
 
-        if self.version >= Version("4.0"):
-            config_args.append("--with-cubew=%s" % spec["cubew"].prefix.bin)
-            config_args.append("--with-cubelib=%s" % spec["cubelib"].prefix.bin)
-        else:
-            config_args.append("--with-cube=%s" % spec["cube"].prefix.bin)
+        config_args.append("--with-cube=%s" % spec["cube"].prefix.bin)
```
melven: Sorry, I don't understand this change here:

There is no dependency to `cube`, shouldn't it be to `cubew` and `cubelib`?

It reads as if the wrong path of the if was removed...

melven: I also get an error here:
```
[x] hyfjpim scorep@10.0 failed: /tmp/zoel_ml/spack-stage/spack-stage-scorep-10.0-hyfjpimfbxvvomx67yqjpumo3mb27vod-ucueivhm.log (5s)
-- lines 83 to 102 --
    File "/home/zoel_ml/spack_tmp/lib/spack/spack/spec.py", line 3762, in __getitem__
      edge = next((e for e in order if e.spec.name == name or name in e.virtuals))
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  StopIteration

  The above exception was the direct cause of the following exception:

    File "/home/zoel_ml/spack-packages/repos/spack_repo/builtin/build_systems/autotools.py", line 622, in configure
      options += self.configure_args()
                 ^^^^^^^^^^^^^^^^^^^^^
    File "/home/zoel_ml/spack_tmp/lib/spack/spack/builder.py", line 270, in _adapter
      return bind_method()
             ^^^^^^^^^^^^^
    File "/home/zoel_ml/spack-packages/repos/spack_repo/builtin/packages/scorep/package.py", line 164, in configure_args
      config_args.append("--with-cube=%s" % spec["cube"].prefix.bin)
                                            ~~~~^^^^^^^^
    File "/home/zoel_ml/spack_tmp/lib/spack/spack/spec.py", line 3764, in __getitem__
      raise KeyError(f"No spec with name {name} in {self}") from e
  KeyError: 'No spec with name cube in scorep@10.0/hyfjpimfbxvvomx67yqjpumo3mb27vod'

```

melven (changes requested):

repos/spack_repo/builtin/packages/scorep/package.py
wrwilliams: Quite correct, fixed by fd600ee305f6944bb752d3d603d51467d3084b78.

repos/spack_repo/builtin/packages/scorep/fix-intel-mpi-linker-error.patch
wrwilliams: It does download from GitLab.

10.0 was tagged prematurely and has a regression: our Intel MPI adapters at the 10.0 tag do not have a NEEDED entry for libmpi, which can lead to situations where e.g. someone building with CMake would be able to build a nonfunctional target.

There are other bug fixes that are going into 10.1 and we don't want to deprecate 10.0 if we can patch it, but per discussion on other package MRs, it is apparently preferred these days that patches come from actual commits--so as soon as the patch became available in the repo, I used that as the source.

melven: That's fine but the .patch file is not needed and can be removed again now, right?

wrwilliams: Ah, yes. See 2f2cd202ce.

repos/spack_repo/builtin/packages/scorep/package.py
melven: Works great now, thanks:
```
> spack find scorep
-- linux-ubuntu24.04-zen3 / %c,cxx,fortran=gcc@16.1.0 -----------
scorep@10.0
==> 1 installed package
```
Also compile and runs fine using a simple openmp test program on my system.

melven (approved):

repos/spack_repo/builtin/packages/scorep/package.py
```diff
@@ -32,6 +33,11 @@ class Scorep(AutotoolsPackage):
         when="@7:8.4 +gcc-plugin",
         sha256="d20b3046ba6a89ad9c106bcf372bceb1bd9ab780d4c7dd9e7373f0099b92d933",
     )
+    patch(
+        "https://gitlab.com/score-p/scorep/-/commit/6d5534de9746e76a28dd9452c3efaa45890067cd.diff",
+        when="@10.0 +mpi",
```
bernhardkaindl, line 43: @wrwilliams, this patch does not apply when I try to build with `+mpi`:
```py
spack install scorep@10.0+mpi
[x] nkiypls scorep@10.0 failed: /tmp/bkaindl/spack-stage/spack-stage-scorep-10.0-nkiypls6dfad3c4763dj6qzsg2v3k63b-uuxwyzt6.log (1s)
-- lines 1 to 6 --
  ==> Using cached archive: /home/bkaindl/gh/spack-core/var/spack/cache/_source-cache/archive/1e/1e96fe2414bfc9eb868619501a5f62943b554eb1358628610e8e9e01036dacf2.tar.gz
  ==> Using cached archive: /home/bkaindl/gh/spack-core/var/spack/cache/_source-cache/archive/20/20946ff71fb226bbe76b4f6dfff6cea127211d58162247d8e0bebba745917e7d
> 1 out of 1 hunk FAILED -- saving rejects to file ChangeLog.rej
  ==> Patch https://gitlab.com/score-p/scorep/-/commit/6d5534de9746e76a28dd9452c3efaa45890067cd.diff failed.
  Command exited with status 1:
      /usr/bin/patch -s -p 1 -i /tmp/bkaindl/spack-stage/spack-stage--nkiypls-patch-20946ff71fb226bbe76b4f6dfff6cea127211d58162247d8e0bebba745917e7d/spack-src/6d5534de9746e76a28dd9452c3efaa45890067cd.diff -d .
==> Error: The following packages failed to install:
scorep@10.0/nkiypls6dfad3c4763dj6qzsg2v3k63b: /tmp/bkaindl/spack-stage/spack-stage-scorep-10.0-nkiypls6dfad3c4763dj6qzsg2v3k63b-uuxwyzt6.log
```

wrwilliams: ...the ChangeLog hunk fails to apply? Odd. That's the diff that's in our repo for the 10.0->10.1 path.

I see four paths available:

1) Put a clean patch somewhere (patches repo I guess?) that has just the minimum to make 10.0/intel MPI work, accept it will be technically different from 10.1's solution. This would also cover the 1b) case of using the various file manipulation/string substitution options inside Spack to recreate the patch manually.
2) Somehow within the package class (file filtering etc) trim the ChangeLog chunk from the patch so it applies cleanly
3) Assuming that this is a whitespace issue, force `--ignore-whitespace` on application
4) Punt on 10.0 getting into Spack, start the package with 10.1 as the first good 10.x

Thoughts?

bernhardkaindl: @wrwilliams, the reason the changelog diff does not apply is:
```py
diff --git a/ChangeLog b/ChangeLog
index 80ae992df02cb2fa23b49ad643b8cac9e2bd1fa1..b97f3b7df3d5d62142d46d7f72d35f030aa48dec 100644
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,5 +1,10 @@
 ------------------- Upcoming version 10.1 ----------------------------
 
+Bugfixes:
+
+- Prevent potential unresolved references/symbols errors in
+  conjunction with Intel MPI.
+
 ------------------- Released version 10.0 ----------------------------
```

Whereas the ChangeLog file of 10.0 starts with:
```py
------------------- Released version 10.0 ----------------------------
```
So if you prepend this (with a empty line afte the "Upcoming") to the file before patching it, the patch might apply:
```py
 ------------------- Upcoming version 10.1 ----------------------------
 
```

bernhardkaindl (approved):
LGTM.

@wrwilliams - BTW, just as a comment:

With the new post-10.0 patch added for the ChangeLog (and the version), I just want to note that it sets the version to 10.1beta, AFAICS:

```patch
diff --git a/build-config/VERSION b/build-config/VERSION
index 0b8a8e3e807b497bdd8e2ef73b433d1af730c402..dc287b4e97f8dda9b57f27029c818b375d417a95 100644
--- a/build-config/VERSION
+++ b/build-config/VERSION
@@ -15,8 +15,8 @@ library interface version:
     http://sourceware.org/autobook/autobook/autobook_91.html#SEC91
 
 package.major=10
-package.bugfix=0
-package.suffix=
+package.bugfix=1
+package.suffix=beta
 
 library.current=15
 library.revision=0
```
If you like to reset the version in this file to 10.0:

These two lines could be reverted back to 10.0, using two filter_file() calls that would normally be done in a do_patch() function.

These calls that just replace these lines to restore the original lines from 10.0. But as the package builds, it would be ok for me to merge.

If you want or need the version to be 10.0, I can help with coming up with a `def patch(self):` function with these two `filter_file()` calls, but it should be quite simple.

Otherwise, I'm ready to merge as-is.

AFAICS, if I see no other comments and revisit this PR at some point,
I think might merge it, but I wanted to let you know of the version info update in this patch.

wrwilliams:
> But as the package builds, it would be ok for me to merge.
> 
I am fine with a 10.0 patched towards 10.1 having a 10.1-beta tag, at least as a one-off. Please feel free to merge at your leisure.

PR #5349, opened 2026-06-27
Update packages

awvwgk:
- add new versions in TOML Fortran, Jonquil and mctc-lib
- fix variants, licenses and dependencies

### DIFF
--- a/repos/spack_repo/builtin/packages/dftd4/package.py
+++ b/repos/spack_repo/builtin/packages/dftd4/package.py
@@ -38,6 +38,7 @@ class Dftd4(MesonPackage, CMakePackage):
     version("3.1.0", sha256="b652aa7cbf8d087c91bcf80f2d5801459ecf89c5d4176ebb39e963ee740ed54b")
     version("3.0.0", sha256="a7539d68d48d851bf37b79e37ea907c9da5eee908d0aa58a0a7dc15f04f8bc35")
 
+    variant("shared", default=True, description="Build shared libraries")
     variant("openmp", default=True, description="Use OpenMP parallelisation")
     variant(
         "python",
@@ -45,8 +46,6 @@ class Dftd4(MesonPackage, CMakePackage):
         when="build_system=meson",
         description="Build Python extension module",
     )
-    with when("build_system=cmake"):
-        variant("shared", default=True, description="Build shared libraries")
 
     depends_on("c", type="build")
     depends_on("fortran", type="build")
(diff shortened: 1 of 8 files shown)


awvwgk:
@spackbot fix style

repos/spack_repo/builtin/packages/jonquil/package.py
```diff
@@ -32,6 +34,7 @@ class Jonquil(MesonPackage, CMakePackage):
 
     depends_on("fortran", type="build")
     depends_on("meson@0.57.2:", type="build", when="build_system=meson")
+    depends_on("meson@0.60.0:", type="build", when="build_system=meson @0.3.1:")
```
mtaillefumier: it is probably better to just do 

depends_on("meson", type="build", when="build_system=meson") 

and remove specific version of meson unless there are constraints coming from the library.

repos/spack_repo/builtin/packages/toml_f/package.py
```diff
@@ -36,6 +40,7 @@ class TomlF(MesonPackage, CMakePackage):
 
     depends_on("fortran", type="build")  # generated
     depends_on("meson@0.57.2:", type="build", when="build_system=meson")
+    depends_on("meson@0.60.0:", type="build", when="build_system=meson @0.5.1:")
```
mtaillefumier: same here. it is better to avoid pinning a version as spack will always take the latest version it knows by default. spack curent version is 1.11.1 so safe to use

depends_on("meson", ...)

mtaillefumier (changes requested):
besides two minor suggestions this PR looks ok to me

repos/spack_repo/builtin/packages/jonquil/package.py
awvwgk: The lower bounds are left from back when most distributions still only shipped meson 0.49 if at all. I guess they are not needed anymore today.

Growl1234:
Could you please add "shared" variant to tblite recipe?

repos/spack_repo/builtin/packages/dftd4/package.py
```diff
@@ -96,6 +96,7 @@ def meson_args(self):
 class CMakeBuilder(cmake.CMakeBuilder):
     def cmake_args(self):
         return [
+            self.define_from_variant("BUILD_SHARED_LIBS", "shared"),
```
mtaillefumier: the line is repeated.

awvwgk:
```suggestion
```

mtaillefumier (approved):

RMeli (approved):

PR #5350, opened 2026-06-27
opencascade: Make sure TCL libraries are picked up on mac.

luca-heltai:
On a Mac, compilation of OCC fails because TCL libraries are not properly detected. This fix helps cmake find the correct libraries on Mac, and should not arm on linux and windows.


bernhardkaindl:
Hi, @luca-heltai, you wrote in the PR description:

> On a Mac, compilation of OCC fails because TCL libraries are not properly detected. This fix helps cmake find the correct libraries on Mac, and should not arm on linux and windows.

To ensure that this does not affect other platforms and make the purpose specific, you could have put the changes for Darwin below:
```py
        if self.spec.satisfies("platform=darwin"):
```
I think I would prefer that.

bernhardkaindl:
I'll do a build on Linux to check if it causes a build issue, if it does, it is good to merge.

bernhardkaindl (approved):
`opencascade+tk` was still successful on Ubuntu 24.04 with this PR:
```py
spack find -v opencascade+tk
-- linux-ubuntu24.04-skylake / %c,cxx=gcc@13.3.0 ----------------
opencascade@7.9.3+application_framework+data_exchange+draw~ffmpeg~freeimage~freetype~ipo+modeling_algorithms+modeling_data~rapidjson~tbb+tk+visualization~vtk build_system=cmake build_type=Release generator=make
```

PR #5351, opened 2026-06-27
taskflow: update versions and force cxx17 for @3:

luca-heltai:
The original  CMakeLists.txt does not enforce std 17 for task flow 3. This PR fixes this, and adds the latest 3.X versions, and the 4.X versions.


bernhardkaindl (approved):
Built the new versions.

PR #5353, opened 2026-06-27
set `SPACK_PREFIX_MAP_ARGS` for `-ffile-prefix-map=<stage>=.`  and `SPACK_BUILD_ID_ARGS` for  `--build-id` injection at compiler-wrapper level (companion to spack/compiler-wrapper#19)

SebastianPaucar:
This PR sets `SPACK_DEBUG_PREFIX_MAP` in `setup_dependent_build_environment` (in `repos/spack_repo/builtin/packages/compiler_wrapper/package.py`) to the staging source path of the package being compiled. `cc.sh` in `compiler-wrapper` repo reads this variable and injects `-ffile-prefix-map=<staging>=.`,  normalizing DWARF paths for build reproducibility.  `develop`-based packages are excluded.

This PR is a companion to https://github.com/spack/compiler-wrapper/pull/19

### DIFF
--- a/repos/spack_repo/builtin/packages/compiler_wrapper/package.py
+++ b/repos/spack_repo/builtin/packages/compiler_wrapper/package.py
@@ -229,6 +229,16 @@ class CompilerWrapper(Package):
             extra_rpaths = dedupe(extra_rpaths)
             env.set("SPACK_COMPILER_EXTRA_RPATHS", ":".join(extra_rpaths))
 
+        # Set SPACK_DEBUG_PREFIX_MAP to the staging path
+        # Develop packages are excluded
+        if not dependent_spec.variants.get("dev_path"):
+            try:
+                staging_src = dependent_spec.package.stage.source_path
+                env.set("SPACK_DEBUG_PREFIX_MAP", staging_src)
+            except Exception:
+                # Stage not yet set up
+                pass
+
         env.set("SPACK_ENABLE_NEW_DTAGS", self.enable_new_dtags)
         env.set("SPACK_DISABLE_NEW_DTAGS", self.disable_new_dtags)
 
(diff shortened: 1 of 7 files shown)


haampie:
Can this be done without changing `cc.sh`?

repos/spack_repo/builtin/packages/compiler_wrapper/package.py
```diff
@@ -229,6 +229,16 @@ def setup_dependent_build_environment(
...
             env.set("SPACK_COMPILER_EXTRA_RPATHS", ":".join(extra_rpaths))
 
+        # Set SPACK_DEBUG_PREFIX_MAP to the staging path
+        # Develop packages are excluded
+        if not dependent_spec.variants.get("dev_path"):
+            try:
+                staging_src = dependent_spec.package.stage.source_path
+                env.set("SPACK_DEBUG_PREFIX_MAP", staging_src)
+            except Exception:
+                # Stage not yet set up
```
haampie: Why do you wrap this in try/catch?

SebastianPaucar:
> Can this be done without changing `cc.sh`?

It can, but I added the `-ffile-prefix-map` injection into `cc.sh` to hide the flag from the build system. Maybe injecting via `SPACK_*FLAGS` from `package.py` and avoiding touching `cc.sh` could leak the stage info in logs. Could you clarify what you have in mind? I am happy to work on it immediately

repos/spack_repo/builtin/packages/compiler_wrapper/package.py
SebastianPaucar: I added it out of caution, thinking the stage might not always be set up at this call site. I will look into it in more detail and remove it if it is safe.

PR #5354, opened 2026-06-28
Update libxsmm recipe and add version 2.0.0

Growl1234:
Add optional CMake build support for `libxsmm@2.0:` (version 2.0 is coming soon). The CMake path uses the upstream installation rules and maps `+shared` to `BUILD_SHARED_LIBS`. Makefile-only variants are kept unavailable with CMake so that unsupported options cannot be silently ignored.

~This would have better be merged after 2.0 is out and version information here is updated accordingly. And this pinged [`main-2023-11`](https://github.com/spack/spack-packages/blob/d73d9c24ad8f52a7db9af5014ad1d11cfe0e02cc/repos/spack_repo/builtin/packages/libxsmm/package.py#L29) "version" could also be removed then.~ ***[Update: 2.0.0 is released.]***

### DIFF
--- a/repos/spack_repo/builtin/packages/cp2k/package.py
+++ b/repos/spack_repo/builtin/packages/cp2k/package.py
@@ -362,6 +362,9 @@ class Cp2k(MakefilePackage, CMakePackage, CudaPackage, ROCmPackage):
         # require libxsmm-1.11+ since 1.10 can leak file descriptors in Fortran
         depends_on("libxsmm@1.11:")
         depends_on("libxsmm@1.17:", when="@9.1:")
+        # CP2K <= 2026.1 requires the legacy libxsmmext interface,
+        # which was removed in libxsmm 2.0.
+        depends_on("libxsmm@:1.17-cp2k", when="@:2026.1")
         # use pkg-config (support added in libxsmm-1.10) to link to libxsmm
 
     # Several packages provide "opencl" (incl. ICD/loader), e.g., "cuda"
--- a/repos/spack_repo/builtin/packages/cp2k/package.py
+++ b/repos/spack_repo/builtin/packages/cp2k/package.py
@@ -364,7 +364,7 @@ class Cp2k(MakefilePackage, CMakePackage, CudaPackage, ROCmPackage):
         depends_on("libxsmm@1.17:", when="@9.1:")
         # CP2K <= 2026.1 requires the legacy libxsmmext interface,
         # which was removed in libxsmm 2.0.
-        depends_on("libxsmm@:1.17-cp2k", when="@:2026.1")
+        depends_on("libxsmm@:1", when="@:2026.1")
         # use pkg-config (support added in libxsmm-1.10) to link to libxsmm
 
     # Several packages provide "opencl" (incl. ICD/loader), e.g., "cuda"
--- a/repos/spack_repo/builtin/packages/dbcsr/package.py
+++ b/repos/spack_repo/builtin/packages/dbcsr/package.py
@@ -125,6 +125,9 @@ class Dbcsr(CMakePackage, CudaPackage, ROCmPackage):
 
     with when("smm=libxsmm"):
         depends_on("libxsmm@1.11:")
+        # DBCSR <= 2.9.1 requires the legacy libxsmmext interface,
+        # which was removed in libxsmm 2.0.
+        depends_on("libxsmm@:1.17-cp2k", when="@:2.9.1")
 
     depends_on("cmake@3.10:", type="build")
     depends_on("cmake@3.12:", type="build", when="@2.1:")
--- a/repos/spack_repo/builtin/packages/dbcsr/package.py
+++ b/repos/spack_repo/builtin/packages/dbcsr/package.py
@@ -127,7 +127,7 @@ class Dbcsr(CMakePackage, CudaPackage, ROCmPackage):
         depends_on("libxsmm@1.11:")
         # DBCSR <= 2.9.1 requires the legacy libxsmmext interface,
         # which was removed in libxsmm 2.0.
-        depends_on("libxsmm@:1.17-cp2k", when="@:2.9.1")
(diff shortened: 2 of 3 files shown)


Growl1234:
@spackbot fix style

Growl1234:
Libxsmm 2.0.0 is out now, so I just updated the recipe. @hfp could you have a look?

Growl1234:
There is problem with `1.17-cp2k` version flag.

Growl1234:
Some generated CI jobs are failing during GitLab artifact upload with:

```text
FATAL: flag provided but not defined: -timeout
```

This occurs before the package build or test payload is run, and affects multiple unrelated `*-generate` jobs, so it appears to be a shared GitLab Runner/artifact-upload issue rather than a package-specific failure.

Anyone have a look at it?

Growl1234:
Currently one of the two CP2K building tests on CI failed, but unfortunately I cannot see the full log as it exceeds the storage limit...

Given both are based on `x86_64`, I suspect it's a false positive. Let me try pushing the commit again.

mtaillefumier (approved):
LGTM

Growl1234:
Gentle ping:)
@hfp @RMeli could you have a look?

repos/spack_repo/builtin/packages/libxsmm/package.py
```diff
@@ -5,35 +5,32 @@
...
     git = "https://github.com/libxsmm/libxsmm.git"
 
     maintainers("hfp")
 
     license("BSD-3-Clause")
 
-    # 2.0 release is planned for Jan / Feb 2024. This commit from main is added
-    # as a stable version that supports other targets than x86. Remove this
-    # after 2.0 release.
-    version("main-2023-11", commit="0d9be905527ba575c14ca5d3b4c9673916c868b2")
```
RMeli, line 29: I think this needs to be deprecated first.

repos/spack_repo/builtin/packages/libxsmm/package.py
```diff
@@ -5,35 +5,32 @@
...
+
     version("main", branch="main")
-    version("1.17-cp2k", commit="6f883620f58afdeebab28039fc9cf580e76a5ec6")
-    version(
-        "1.17",
-        sha256="8b642127880e92e8a75400125307724635ecdf4020ca4481e5efe7640451bb92",
-        preferred=True,
-    )
+    version("2.0.0", sha256="7e532dc5520f864ce6d7f44f3fd50365e3edb23da97dbdc54fd53845d86a290b")
+    version("1.17-cp2k", commit="e0c4a2389afba36c453233ad7de07bd92c715bec")
```
RMeli: Can we deprecate this version, or is it really necessary for CP2K? If it is just to match the toolchain version, I'd deprecate it and then remove it down the line.

repos/spack_repo/builtin/packages/libxsmm/package.py
```diff
@@ -68,36 +65,59 @@ class Libxsmm(MakefilePackage):
...
+    with when("build_system=makefile"):
+        # A recent `as` is needed to compile libxsmm until version 1.17
+        # (<https://github.com/spack/spack/issues/28404>), but not afterwards
+        # (<https://github.com/spack/spack/pull/21671#issuecomment-779882282>).
+        depends_on("binutils+ld+gas@2.33:", type="build")
+
+    # CMake support is introduced with 2.0. Keep it available for the moving
+    # development branch, but prevent selection for the older fixed sources.
+    conflicts("build_system=cmake", when="@:1.17")
+    conflicts("build_system=cmake", when="@1.17-cp2k")
```
RMeli: I think it would be cleaner to use a `conditional()` when defining the available build systems.

```
    build_system("makefile", conditional("cmake", when="@2:"), default="makefile")
```

Also, seeing the two conflicts I'm wondering: does `@:1.17-cp2k` used elsewhere works as intended? (I need to double check the version rules...)

repos/spack_repo/builtin/packages/libxsmm/package.py
Growl1234: Honestly I don't know how this version was originally introduced, but the existing `6f883620f58afdeebab28039fc9cf580e76a5ec6` commit failed with CP2K build, and previously CP2K's local Spack recipe used the same version as toolchain, so I modified it here. I think it's fine to deprecate and remove it because it wasn't actually used by DBCSR and CP2K in the official recipes.

Growl1234:
Hi @RMeli, could you have a look at this again?

repos/spack_repo/builtin/packages/cp2k/package.py
```diff
@@ -362,6 +362,9 @@ class Cp2k(MakefilePackage, CMakePackage, CudaPackage, ROCmPackage):
         # require libxsmm-1.11+ since 1.10 can leak file descriptors in Fortran
         depends_on("libxsmm@1.11:")
         depends_on("libxsmm@1.17:", when="@9.1:")
+        # CP2K <= 2026.1 requires the legacy libxsmmext interface,
+        # which was removed in libxsmm 2.0.
+        depends_on("libxsmm@:1.17-cp2k", when="@:2026.1")
```
RMeli:
```suggestion
        depends_on("libxsmm@:1", when="@:2026.1")
```

repos/spack_repo/builtin/packages/dbcsr/package.py
```diff
@@ -125,6 +125,9 @@ class Dbcsr(CMakePackage, CudaPackage, ROCmPackage):
 
     with when("smm=libxsmm"):
         depends_on("libxsmm@1.11:")
+        # DBCSR <= 2.9.1 requires the legacy libxsmmext interface,
+        # which was removed in libxsmm 2.0.
+        depends_on("libxsmm@:1.17-cp2k", when="@:2.9.1")
```
RMeli:
```suggestion
        depends_on("libxsmm@:1", when="@:2.9.1")
```

RMeli (approved):

PR #5357, opened 2026-06-28
Revert "Add GitHub Copilot instructions for Spack package repository"

haampie:
Reverts spack/spack-packages#1243

We don't commit editor related config files.


alalazo (approved):
Also, we are discussing AI policies during the next TSC - so another reason not to merge AI related PRs before any decision is taken.

alalazo:
@bernhardkaindl @wdconinc fyi

wdconinc:
> Also, we are discussing AI policies during the next TSC - so another reason not to merge AI related PRs before any decision is taken.

We've been discussing AI policies for 10 months in the TSC, and I don't think that what we will discuss in the TSC has any impact on this PR. This revert just comes across as poorly motivated (AGENTS.md is an editor config? Should we remove git blame suppression too?) and more based on a personal preference than the outcome of a motivated discussion. None of these objections (neither the need for discussion in the TSC nor the notion that AGENTS.md are editor config files) were raised before in the original PR.

alalazo:
> This revert just comes across as poorly motivated (AGENTS.md is an editor config? Should we remove git blame suppression too?) and more based on a personal preference than the outcome of a motivated discussion. 

Just saw this message. I approved the revert because it's my understanding that we didn't commit any AGENTS.md or CLAUDE.md until now on purpose. So the idea is really to get more consensus around PRs like that before merging them.

PR #5360, opened 2026-06-28
osg-ca-certs: depends on gmake and perl, type build

wdconinc:
This adds to `osg-ca-certs` the build dependencies `gmake` and `perl`, which are needed by the build script, https://github.com/opensciencegrid/osg-certificates/blob/master/build-certificates-dir.sh.


alecbcs (approved):

PR #5363, opened 2026-06-29
py-pspamm: Use spack's defaullt to install the package

Thomas-Ulrich:
Else the package is not found by seissol.


kshea21:
Do you have thoughts on this change, @ravil-mobile ?

bernhardkaindl (approved):

PR #5366, opened 2026-06-29
Vecgeom: new version v2.1.0

sethrj:
- [release notes](https://gitlab.cern.ch/VecGeom/VecGeom/-/releases/v2.1.0)
- deprecated ancient veccore versions (v0.8.2. has been preferred for a long time by vecgeom)
- unmark 1.x as preferred


wdconinc (approved):
LGTM.

sethrj:
Heads up that for some reason the URLs are returning 403 when fetched through spack/curl, even though I can download them through my browser (even when in private mode, so it shouldn't be cookie/login-related…)

wdconinc:
Yeah, I saw the other PR addressing that failure mode. Let me know when that fix is in or the CERN gitlab server stops serving 403. I can't override the CI checks I'm afraid.

PR #5382, opened 2026-06-30
r-made4: new package

w8jcik:

### DIFF
--- a/repos/spack_repo/builtin/packages/r_ade4/package.py
+++ b/repos/spack_repo/builtin/packages/r_ade4/package.py
@@ -22,6 +22,8 @@ class RAde4(RPackage):
 
     license("GPL-2.0-or-later")
 
+    version("1.7-24", sha256="59509dcf781b2b69a97e225f53f2f76d0e0d5bfb04c5b6198007c035c4bd6d2a")
+    version("1.7-23", sha256="3092c5a68f68f24fc28387407f43b939862a460635250609f2409e359476677f")
     version("1.7-22", sha256="007df54e83a2a6cb8d6da8006f0aace011e7eaa7744dc5f8230ac2c002b393b4")
     version("1.7-20", sha256="ddc75788b55902f17dabda209eb67c378aa0233767697e00fa3f8fc85e4bad7b")
     version("1.7-19", sha256="c85111881ff93a261ca34a5b5ef313d1c0ff9b84f0b23596bc56cfa696f165ef")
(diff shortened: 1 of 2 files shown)


repos/spack_repo/builtin/packages/r_made4/package.py
```diff
@@ -0,0 +1,31 @@
...
+
+    with default_args(get_full_repo=True):
+        version("1.86.0", commit="e95e18a3e90a95f49c864ccfcce7250bd55ef41f")  # bioc 3.23
+        version("1.84.0", commit="34e97a7f79ecaa16d65ac2060413d4b5dbcd583f")  # bioc 3.22
+        version("1.82.0", commit="af8e708e74dff93eee6e67c9c2d73e7941be783f")  # bioc 3.21
+        version("1.80.0", commit="3db6e0c36d353732ce857203f77ee4414c57c26c")  # bioc 3.20
+        version("1.78.0", commit="981e55a052f8809f524b9069c7125dc29d79efb4")  # bioc 3.19
+        version("1.76.0", commit="4fa62322cf560d2720055f48c1dcf46f5bc5d745")  # bioc 3.18
+        version("1.74.0", commit="d5b34b546d47e42523c61afc4c77e3b8c2c0f479")  # bioc 3.17
+        version("1.72.0", commit="4f008b4fdcd1cdbda818b81ffa9efceaaf6a823e")  # bioc 3.16
```
bernhardkaindl: @w8jcik, do you need all these new versions?

```suggestion
    with default_args(get_full_repo=True):
        version("1.86.0", commit="e95e18a3e90a95f49c864ccfcce7250bd55ef41f")  # bioc 3.23
```

Spack accumulates versions over time because unless deprecating them for various reasons including to keep the size of the concretisation job manageable. New packages should just add the versions that are really needed.

repos/spack_repo/builtin/packages/r_ade4/package.py
```diff
@@ -22,6 +22,8 @@ class RAde4(RPackage):
 
     license("GPL-2.0-or-later")
 
+    version("1.7-24", sha256="59509dcf781b2b69a97e225f53f2f76d0e0d5bfb04c5b6198007c035c4bd6d2a")
+    version("1.7-23", sha256="3092c5a68f68f24fc28387407f43b939862a460635250609f2409e359476677f")
```
bernhardkaindl: While at it (see the other comment on new vesrions):
```suggestion
    version("1.7-24", sha256="59509dcf781b2b69a97e225f53f2f76d0e0d5bfb04c5b6198007c035c4bd6d2a")
```

Spack accumulates versions over time because unless deprecating them for various reasons including to keep the size of the concretisation job manageable, it does not remove versions to not break existing spack uses. Like for new packages it would be sufficient to just add the versions that are really needed.

Of course keep the dependencies for 1.7-23, because such versions could be built just be specifying that version directly, as far as I'm aware.

bernhardkaindl (commented):
Added two conversations for questions I've had during review.

repos/spack_repo/builtin/packages/r_made4/package.py
w8jcik: I left 3.21 that I need and latest 3.23. Is it fine?

bernhardkaindl (approved):
@w8jcik Looks good to me now and built the new versions on Ubuntu 24.04 a a quick sanity check.

w8jcik:
Thank you for the review and merging.

PR #5383, opened 2026-06-30
r-affy and r-affyio: add @1.86,1.90 and @1.78,1.82

w8jcik:

### DIFF
--- a/repos/spack_repo/builtin/packages/r_affy/package.py
+++ b/repos/spack_repo/builtin/packages/r_affy/package.py
@@ -17,16 +17,19 @@ class RAffy(RPackage):
     bioc = "affy"
 
     with default_args(get_full_repo=True):
+        version("1.90.0", commit="7ed56e774351b32c16e0f1a6314faa2b0a79f36c")  # bioc 3.23
+        version("1.86.0", commit="263e70742f3085699839b7762b99f7dfe734a573")  # bioc 3.21
         version("1.78.0", commit="cc7eac358b6e10ee86a7a93d2e436758f6fbd9b5")
-        version("1.76.0", commit="3bb309388d5d6402c356d4a5270ee83c5b88942f")
+        version("1.76.0", commit="3bb309388d5d6402c356d4a5270ee83c5b88942f")  # bioc 3.16
         version("1.74.0", commit="2266c4a46eda7e5b64f7f3e17e8b61e7b85579ff")
         version("1.72.0", commit="3750b4eb8e5224b19100f6c881b67e568d8968a2")
 
-    depends_on("c", type="build")  # generated
+    depends_on("c", type="build")
 
     depends_on("r-biocgenerics@0.1.12:", type=("build", "run"))
     depends_on("r-biobase@2.5.5:", type=("build", "run"))
     depends_on("r-affyio@1.13.3:", type=("build", "run"))
     depends_on("r-biocmanager", type=("build", "run"))
     depends_on("r-preprocesscore", type=("build", "run"))
-    depends_on("r-zlibbioc", type=("build", "run"))
+
+    depends_on("r-zlibbioc", type=("build", "run"), when="@:1.85.0")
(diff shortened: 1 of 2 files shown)


bernhardkaindl:
On these `r-raffyio` versions:
```py
        version("1.70.0", commit="95560567e27088863c64e868a8e5069fc725b8d7")  # bioc 3.17
        version("1.68.0", commit="33080c5eeb14c0ca40f0d231706af4e0c2c1ef8b")  # bioc 3.16      
```
@w8jcik, I'm having trouble to compile these older:
```py
.../r-4.5.3-.../rlib/R/include" ...  -c read_pgf.c -o read_pgf.o
  read_cdf_xda.c: In function 'read_cdf_qcunit':
> read_cdf_xda.c:245:24: warning: implicit declaration of function 'Calloc'; did you mean 'calloc'? [-Wimplicit-function-declaration]
    245 |   my_unit->qc_probes = Calloc(my_unit->n_probes,cdf_qc_probe);
        |                        ^~~~~~
        |                        calloc
> read_cdf_xda.c:245:49: error: expected expression before 'cdf_qc_probe'
    245 |   my_unit->qc_probes = Calloc(my_unit->n_probes,cdf_qc_probe);
        |                                                 ^~~~~~~~~~~~
  read_cdf_xda.c: In function 'read_cdf_unit':
> read_cdf_xda.c:289:49: error: expected expression before 'cdf_unit_block'
    289 |   my_unit->unit_block = Calloc(my_unit->nblocks,cdf_unit_block);
        |                                                 ^~~~~~~~~~~~~~
> read_cdf_xda.c:300:78: error: expected expression before 'cdf_unit_cell'
    300 |     my_unit->unit_block[i].unit_cells = Calloc(my_unit->unit_block[i].ncells,cdf_unit_cell);
        |                                                                              ^~~~~~~~~~~~~
  read_cdf_xda.c: In function 'dealloc_cdf_xda':
> read_cdf_xda.c:332:5: warning: implicit declaration of function 'Free'; did you mean 'free'? [-Wimplicit-function-declaration]
    332 |     Free(my_cdf->probesetnames[i]);
        |     ^~~~
        |     free
  read_cdf_xda.c: In function 'read_cdf_xda':
> read_cdf_xda.c:420:62: error: expected expression before 'char'
    420 |   my_cdf->header.ref_seq = Calloc(my_cdf->header.len_ref_seq,char);
        |                                                              ^~~~
```
There are hundreds of these `expected expression before 'char'`.

The same errors occur when I try to build other older versions like `r-preprocesscore@1.62` and older.

Update: This is an R version problem where r@4.5 removed those Macros. A workaround would be to define the macros in spack building using r@4.5 like this...:
```c
#define Calloc(n,t) ((t *)R_Calloc((n),t))
#define Free R_Free
```
Workaround:
```py
spack install r-affyio@1.70 ^r@4.4
```

Suggestion: Do you need these older versions or can you start deprecating the oldest versions of these packages as you add new versions?

bernhardkaindl (approved):
@w8jcik approving, but please see: https://github.com/spack/spack-packages/pull/5383#issuecomment-4979529466

w8jcik:
> Do you need these older versions or can you start deprecating the oldest versions of these packages

I noticed a trend in Spack that packages are deprecated in large numbers. In particular there was a large deprecation wave when the `spack-packages` was split from `spack`. Although I understand maintenance reasons I believe that deprecation for no reason is problematic for the HPC/scientific users. Scientists keep some very old packages in use and they don't have time capacity, funding, sometimes skills to migrate.

In particular I was disappointed with Spack upon deprecation of older [GROMACS](https://packages.spack.io/package.html?name=gromacs) versions which was causing me some huge practical issues.

My point is that I would not deprecate something that someone might still need. That includes even 10 years old packages.

> Update: This is an R version problem where r@4.5 removed those Macros. A workaround would be to define the macros in spack building using r@4.5 like this...:

We could add the patch that you suggested, but wouldn't it be better to just add a constraint on R version? That would be the a logical solution that requires no maintenance.

w8jcik:
How about adding this constraint?
- https://github.com/spack/spack-packages/pull/5609

The issue is with Bioconductor in general which is only intended to be used with particular version of R.  

Spack doesn't include constraint on R, because Bioconductor doesn't guarantee incompatibility, it only doesn't guarantee compatibility.

PR #5391, opened 2026-06-30
py-jaxtyping: add v0.3.9

LydDeb:

### DIFF
--- a/repos/spack_repo/builtin/packages/py_jaxtyping/package.py
+++ b/repos/spack_repo/builtin/packages/py_jaxtyping/package.py
@@ -15,15 +15,19 @@ class PyJaxtyping(PythonPackage):
 
     license("Apache-2.0")
 
+    version("0.3.9", sha256="f8c02d1b623d5f1b6665d4f3ddaec675d70004f16a792102c2fc51264190951d")
     version("0.2.33", sha256="9a9cfccae4fe05114b9fb27a5ea5440be4971a5a075bbd0526f6dd7d2730f83e")
     version("0.2.19", sha256="21ff4c3caec6781cadfe980b019dde856c1011e17d11dfe8589298040056325a")
 
     depends_on("py-hatchling", type="build")
 
     with default_args(type=("build", "run")):
+        depends_on("python@3.11:", when="@0.3.8:")
         depends_on("python@3.9:3", when="@0.2.33:")
         depends_on("python@3.8:3", when="@0.2.19")
-        depends_on("py-typeguard@2.13.3", when="@0.2.33:")
+        depends_on("py-wadler-lindig@0.1.3:", when="@0.2.37:")
+        depends_on("py-typing-extensions", when="@0.2.36^python@:3.9")
+        depends_on("py-typeguard@2.13.3", when="@0.2.33:0.2.34")
         depends_on("py-typeguard@2.13.3:", when="@0.2.19")
 
         # Historical dependencies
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/py_jaxtyping/package.py
```diff
@@ -15,15 +15,19 @@ class PyJaxtyping(PythonPackage):
...
 
     depends_on("py-hatchling", type="build")
 
     with default_args(type=("build", "run")):
+        depends_on("python@3.11:", when="@0.3.8:")
         depends_on("python@3.9:3", when="@0.2.33:")
         depends_on("python@3.8:3", when="@0.2.19")
-        depends_on("py-typeguard@2.13.3", when="@0.2.33:")
+        depends_on("py-wadler-lindig@0.1.3:", when="@0.2.37:")
+        depends_on("py-typing-extensions", when="@0.2.36^python@:3.9")
```
alecbcs:
```suggestion
        depends_on("py-typing-extensions", when="@0.2.36 ^python@:3.9")
```

repos/spack_repo/builtin/packages/py_jaxtyping/package.py
```diff
@@ -15,15 +15,19 @@ class PyJaxtyping(PythonPackage):
...
     license("Apache-2.0")
 
+    version("0.3.9", sha256="f8c02d1b623d5f1b6665d4f3ddaec675d70004f16a792102c2fc51264190951d")
     version("0.2.33", sha256="9a9cfccae4fe05114b9fb27a5ea5440be4971a5a075bbd0526f6dd7d2730f83e")
     version("0.2.19", sha256="21ff4c3caec6781cadfe980b019dde856c1011e17d11dfe8589298040056325a")
 
     depends_on("py-hatchling", type="build")
 
     with default_args(type=("build", "run")):
+        depends_on("python@3.11:", when="@0.3.8:")
```
alecbcs:
```suggestion
        depends_on("python@3.11:3", when="@0.3.8:")
```

alecbcs (changes requested):
Couple of quick nitpicks, otherwise looks good to me. Thanks @LydDeb

LydDeb:
Thanks for the review @alecbcs

repos/spack_repo/builtin/packages/py_jaxtyping/package.py
```diff
@@ -15,15 +15,19 @@ class PyJaxtyping(PythonPackage):
...
 
     with default_args(type=("build", "run")):
+        depends_on("python@3.11:3", when="@0.3.8:")
         depends_on("python@3.9:3", when="@0.2.33:")
         depends_on("python@3.8:3", when="@0.2.19")
-        depends_on("py-typeguard@2.13.3", when="@0.2.33:")
+        depends_on("py-wadler-lindig@0.1.3:", when="@0.2.37:")
+        depends_on("py-typing-extensions", when="@0.2.36 ^python@:3.9")
+        depends_on("py-typeguard@2.13.3", when="@0.2.33:0.2.34")
         depends_on("py-typeguard@2.13.3:", when="@0.2.19")
```
alecbcs: I think there's something funky with these depends_on directives. This is saying the package depends on 

```
py-typeguard >= 2.13.3 when 0.2.19
py-typeguard == 2.13.3 when 0.2.33 -> 0.2.34
```

So this applies a very tight constraint for those two versions and then drops the dependency entirely after.

alecbcs:
```suggestion
        depends_on("py-typeguard@2.13.3", when="@:0.2.34")
```

I'd suggest changing the dependency to this since we can assume that older versions of `jaxtyping` weren't actually compatible with any newer versions of `typeguard` but the upper bound was only added later on when the incompatibility was found.

alecbcs (approved):
Looks good to me. Thanks @LydDeb!

PR #5413, opened 2026-07-01
TetGen: add CMake support

ltroska:
This PR adds CMake as a build system for TetGen.


cmelone (changes requested):
hi @ltroska, Spack expects source and build system files to be available from the tarball or git repository, not included in the spack repo as you've done in this PR. 

check out the packaging guide here and let me know if you have any questions:

https://spack.readthedocs.io/en/latest/packaging_guide_creation.html

PR #5415, opened 2026-07-01
`admixtools` installation escapes build directory

Sibgatulin:
# Issue
Parts of the installation process use the current working directory, where the user ran `spack install admixtools`:

## Symptoms
1. build artefacts (`admixtables/`) are left in the current directory 

    <details> 

    <summary> Building in <code>~/</code> </summary>

    ```
    $ cd ~/
    $ ls admixtables/
    ls: cannot access 'admixtables/': No such file or directory
    $ spack install admixtools@7.0.2 %gcc@14.1.0 %openblas
    [+] / (external glibc-2.38-tfghf6luj545o6gwct3qgiitlwtre7wm)
    [+] /mpcdf/soft/SLE_15/packages/x86_64/gcc/14.1.0 (external gcc-14.1.0-aleiywcnxoqwjfdqz2xq6z4vi76wuh66)
    [+] /geany/fs/scratch/resib/spack/opt/spack/linux-zen4/compiler-wrapper-1.1.0-rm6a2ekkxuodgply3j5kbgt4jnmqpthx
    [+] /geany/fs/scratch/resib/spack/opt/spack/linux-zen4/gcc-runtime-14.1.0-zeve346kgkr5tzt4xddedgx44ctiajxm
    [+] /geany/fs/scratch/resib/spack/opt/spack/linux-zen4/openblas-0.3.33-eqc27nyuza6gubsnx72l7akvqwlacu4b
    [+] /geany/fs/scratch/resib/spack/opt/spack/linux-zen4/gmake-4.4.1-ymcpt75lijc324xrz2td4lp2saqhz6gh
    [+] /geany/fs/scratch/resib/spack/opt/spack/linux-zen4/gsl-2.8-7fthp53pmaw7rrqgyit2ykyenzi33jlu
    ==> No binary for admixtools-7.0.2-bpdeyvz2t4dzip7osehsjhporlk7dj65 found: installing from source
    ==> Installing admixtools-7.0.2-bpdeyvz2t4dzip7osehsjhporlk7dj65 [8/8]
    ==> Using cached archive: /geany/fs/scratch/resib/spack/var/spack/cache/_source-cache/archive/d1/d1dc1963e01017f40e05e28009008e14388a14a3facc75cff46653da585bd91e.tar.gz
    ==> No patches needed for admixtools
    ==> admixtools: Executing phase: 'edit'
    ==> admixtools: Executing phase: 'build'
    ==> admixtools: Executing phase: 'install'
[...]

PR #5417, opened 2026-07-01
admixtools: add v8.0.2

Sibgatulin:
The last three releases are not compatible with the recipe as it is. 

Bits of the Makefile for context:
```makefile
HOMEL=$(PWD)
TOP=../bin
BIN=$(HOMEL)/../bin
```
and 
```makefile
install:	all dirs tables
	mkdir -p $(TOP)
	cp $(PROGS) $(TOP)
	cp $(PROGS2) $(TOP)
	cp $(PROGS3) $(TOP)
	cp $(PROGS3) $(TOP)
	cp ../perlsrc/* $(TOP) 
	cp fxtract $(BIN)
	cp -r script $(BIN) 
```

the last 3 lines of `install` target were introduced in v8.0.0 and the two of them use $BIN instead of $TOP.
[Normally](https://github.com/DReichLab/AdmixTools/blob/683c89db5ae2d1dc819f45e52d912ed774c83687/README.INSTALL#L9)  `$BIN` and `$TOP` evaluate to the same folder (and I don't quite understand why the two of them).
However, because Makefile so far uses `PWD` and not `CURWD`, `$BIN != $TOP` under `spack install` (see #5415). Yet, Makefile never `mkdir $(BIN)`, only `mkdir -p $(TOP)`and this, in turn, means that `cp fxtract $(BIN)` *creates* `$BIN` to be an executable file (a copy of `fxtract`) and not a folder. This means that the next call, fails with
`cp: cannot overwrite non-directory '...' with directory 'script'`.
IMHO, easiest fix is to change $(BIN) to $(TOP), consistently with the rest of `install` target, which this PR does.
<details>

<summary>Once #5415 lands, it can be simplified </summary>

```diff
@@ -56,12 +56,9 @@ class Admixtools(MakefilePackage):
         )
 
         makefile.filter("HOMEL=$(PWD)", "HOMEL=$(CURDIR)", string=True)
-        makefile.filter("TOP=../bin", "TOP=./bin")
-        makefile.filter("cp fxtract $(BIN)", "cp fxtract $(TOP)", string=True)
-        makefile.filter("cp -r script $(BIN)", "cp -r script $(TOP)", string=True)
 
     def install(self, spec, prefix):
         with working_dir("src"):
             make("install")
-            install_tree("bin", prefix.bin)
+            install_tree("../bin", prefix.bin)
             install("twtable", prefix.bin)
```
</details>

### DIFF
--- a/repos/spack_repo/builtin/packages/admixtools/package.py
+++ b/repos/spack_repo/builtin/packages/admixtools/package.py
@@ -15,6 +15,9 @@ class Admixtools(MakefilePackage):
     homepage = "https://github.com/DReichLab/AdmixTools"
     url = "https://github.com/DReichLab/AdmixTools/archive/v7.0.2.tar.gz"
 
+    version("8.0.2", sha256="fea3eaabc5c47aa85dbc4346b6be0c377249064ccba087c246cbc7bec4b18777")
+    version("8.0.1", sha256="fb9347542cf847f8f6e0cf5abc62e2ff615de4a53c20fa8d0b7d9e4bc4216f12")
+    version("8.0.0", sha256="8779dd94c923dfaf2e6449f008dd45671c121158262fdaa0925b87b4f91650c1")
     version("7.0.2", sha256="d1dc1963e01017f40e05e28009008e14388a14a3facc75cff46653da585bd91e")
     version("7.0.1", sha256="182dd6f55109e9a1569b47843b0d1aa89fe4cf4a05f9292519b9811faea67a20")
     version("7.0", sha256="c00faab626f02bbf9c25c6d2dcf661db225776e9ed61251f164e5edeb5a448e5")
@@ -42,6 +45,8 @@ class Admixtools(MakefilePackage):
         )
 
         makefile.filter("TOP=../bin", "TOP=./bin")
+        makefile.filter("cp fxtract $(BIN)", "cp fxtract $(TOP)", string=True)
+        makefile.filter("cp -r script $(BIN)", "cp -r script $(TOP)", string=True)
 
     def install(self, spec, prefix):
         with working_dir("src"):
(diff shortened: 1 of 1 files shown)


bernhardkaindl (changes requested):
@Sibgatulin, @8.0.0 doesn't complile with gcc:

```py
  cc -g  -p -pg -L./nicksrc  qp4diff.o nicksrc/libnick.a qpsubs.o mcio.o mapfile.o ldsubs.o admutils.o egsubs.o regsubs.o geno.o  -lgsl -lopenblas -lm -lnick -o qp4diff
> /usr/bin/ld: admutils.o:/tmp/bkaindl/spack-stage/spack-stage-admixtools-8.0.0-mzq5xmgvrwxxw7k2z7q2cg4me2aakel6/spack-src/src/packit.h:3: multiple definition of `packlen'; snpunion.o:/tmp/bkaindl/spack-stage/spack-stage-admixtools-8.0.0-mzq5xmgvrwxxw7k2z7q2cg4me2aakel6/spack-src/src/snpunion.c:67: first defined here
> /usr/bin/ld: admutils.o:/tmp/bkaindl/spack-stage/spack-stage-admixtools-8.0.0-mzq5xmgvrwxxw7k2z7q2cg4me2aakel6/spack-src/src/packit.h:4: multiple definition of `rlen'; snpunion.o:/tmp/bkaindl/spack-stage/spack-stage-admixtools-8.0.0-mzq5xmgvrwxxw7k2z7q2cg4me2aakel6/spack-src/src/snpunion.c:67: first defined here
```

Could you remove 8.0.0? And do you need 8.0.1 when you have 8.0.2?

Sibgatulin:
@bernhardkaindl thanks for looking into it and testing the versions. To be honest, I included the interim releases for completeness, but now I can see that they were followed up by bugfix releases for a reason. Let's only keep the latest stable release at this point

bernhardkaindl (approved):

PR #4828, opened 2026-05-13
fckit: fix error: no member named 'lookup' in 'eckit::system::Library'

harshula:
* Fixed in 0.14.2: https://github.com/ecmwf/fckit/pull/84
* Fixes error:
```
/tmp/root/spack-stage/spack-stage-fckit-0.14.1-o2amrf5abawjgy2dhws4n55poyh7b7e7/spack-src/src/fckit/Log.cc:176:31: error: no member named 'lookup' in 'eckit::system::Library'; did you mean 'eckit::system::LibraryManager::lookup'?
  176 |         if ( Channel& debug = Library::lookup( libname ).debugChannel() ) {
      |                               ^~~~~~~~~~~~~~~
      |                               eckit::system::LibraryManager::lookup
```

PR #5426, opened 2026-07-01
Update versions of ECMWF libraries

fmahebert:
Adds new versions of ECMWF libraries from early June 2026
- eckit 2.0.7
- fckit 0.14.3
- fiat 2.0.0
- ectrans 1.8.0
- atlas 0.46.0

A few additional notes,
- fiat patch doesn't appear needed (not sure since when, I put in 1.6.99 conservatively), but this could use another pair of eyes
- fckit contraints were a bit too strict, so I relaxed them


repos/spack_repo/builtin/packages/eckit/package.py
```diff
@@ -92,8 +93,10 @@ class Eckit(CMakePackage):
 
     # Build issues with cmake 3.20, not sure about 3.21
     depends_on("cmake@3.12:3.19,3.22:", type="build")
+    depends_on("cmake@3.18:3.19,3.22:", when="@1.33:", type="build")
     depends_on("ecbuild@3.5:", when="@:1.20.99", type="build")
-    depends_on("ecbuild@3.7:", when="@1.21:", type="build")
+    depends_on("ecbuild@3.7:", when="@1.21:1.30,1.32:", type="build")
+    depends_on("ecbuild@3.11:", when="@1.31:1.31.4", type="build")
```
climbfuji, line 99: That's interesting. Versions 1.31.0 to 1.31.4 need a newer ecbuild than 1.32?

repos/spack_repo/builtin/packages/ectrans/package.py
```diff
@@ -88,7 +89,8 @@ class Ectrans(CMakePackage):
     )
 
     # https://github.com/ecmwf-ifs/ectrans/issues/316
-    patch("find_lapack.patch", when="@1.5:")
+    patch("find_lapack_v15.patch", when="@1.5:1.7")
+    patch("find_lapack_v18.patch", when="@1.8")
```
climbfuji, line 93: Thanks for carrying this forward. Unfortunate that they still didn't fix this upstream.

climbfuji (approved):

repos/spack_repo/builtin/packages/eckit/package.py
fmahebert: Yes. Based on reading git history, it seems the ECMWF team found it too cumbersome to bump up the ecbuild version on some systems (the phrase "dependency hell" appears), so they rolled back the requirements.

Technically, this is true in some other repos as well, so more logic like this could be added elsewhere. But since we don't always enforce the ecbuild _version_ in these package definition files, I didn't bother to make the changes everywhere. I _think_ that if eckit's ecbuild requirement are satisfied, the other repos' typically will be too. I worry that getting every interdependency exactly right is an endless diminishing returns problem, and I don't know exactly how far to push it for a project like spack.

repos/spack_repo/builtin/packages/fckit/package.py
```diff
@@ -51,7 +52,7 @@ class Fckit(CMakePackage):
     depends_on("eckit@:1.23 +mpi", when="@:0.10 +eckit")
     # https://github.com/ecmwf/fckit/pull/84
     depends_on("eckit@1.24:1.33 +mpi", when="@0.11:0.14.1 +eckit")
-    depends_on("eckit@2: +mpi", when="@0.14.2: +eckit")
+    depends_on("eckit@1.26: +mpi", when="@0.14.2: +eckit")
```
srherbener, line 55: Why did the version for eckit go back to an older one? Was the eckit@2: spec incorrect?

srherbener (commented):
Thanks for the update! I just have the one comment.

repos/spack_repo/builtin/packages/fckit/package.py
fmahebert: Good spot.

You can see some discussion here https://github.com/spack/spack-packages/pull/4828#discussion_r3509499752 around this version requirement change.

In short, my understanding and testing suggests that eckit@2 is too restrictive, and empirically I find that eckit@1.26 is the one that allows fckit to build.

srherbener: Sounds good @fmahebert. Thanks for the explanation!

srherbener (approved):

bernhardkaindl (approved):
Approving with two maintainer reviews and passing installs on Ubuntu 24.04 with %gcc@13.3 and %clang@18 and %clang@22

PR #4809, opened 2026-05-12
py-pyarrow: add v23.0.1, v24.0.0

paulgessinger:

PR #5435, opened 2026-07-02
py-pyarrow: add @20:24; arrow: add @24; xsimd: add @14.3; py-libcst: add @1.8.6; py-pyyaml-ft: new package

w8jcik:
# Build grid

## py-libcst

image

## py-pyarrow

image

...


w8jcik:
@spackbot re-run pipeline

w8jcik:
@spackbot re-run pipeline

repos/spack_repo/builtin/packages/arrow/package.py
```diff
@@ -55,8 +55,11 @@ class Arrow(CMakePackage, CudaPackage):
     depends_on("boost@1.60: +filesystem +system")
     depends_on("brotli", when="+brotli")
     depends_on("bzip2", when="+bz2")
-    depends_on("cmake@3.2.0:", type="build")
-    depends_on("cmake@3.25.0:", type="build", when="@20:")
+    depends_on("cmake@2.7:", type="build")
+    depends_on("cmake@3.2:", when="@0.17:", type="build")
+    depends_on("cmake@3.5:", when="@11:", type="build")
```
manuelakuhn: Shouldn't this be 

```suggestion
    depends_on("cmake@3.2:", type="build")
    depends_on("cmake@3.5:", when="@3:", type="build")
```
see here
https://github.com/apache/arrow/blob/apache-arrow-0.8.0/cpp/CMakeLists.txt#L18
https://github.com/apache/arrow/blob/apache-arrow-3.0.0/cpp/CMakeLists.txt#L18

Or am I looking in the wrong CMakeList.txt?

thomas-bouvier:
Hey @w8jcik would it be possible to add v24 from this PR too https://github.com/spack/spack-packages/pull/4809? Thanks!

manuelakuhn (commented):
The python part looks good to me.

repos/spack_repo/builtin/packages/py_pyarrow/package.py
```diff
@@ -42,8 +46,12 @@ class PyPyarrow(PythonPackage):
     depends_on("c", type="build")
     depends_on("cxx", type="build")
 
+    depends_on("python@3.10:", type=("build", "run"), when="@22:")
+    depends_on("python@3.9:", type=("build", "run"), when="@18:")
+
```
thomas-bouvier, line 52: I think `depends_on("python", type=("build", "run"))` should be added to cover all cases.

repos/spack_repo/builtin/packages/arrow/package.py
thomas-bouvier: I agree with @manuelakuhn, I do not understand where do the changes come from

repos/spack_repo/builtin/packages/arrow/package.py
```diff
@@ -55,8 +55,11 @@ class Arrow(CMakePackage, CudaPackage):
...
+    depends_on("cmake@3.16:", when="@13:", type="build")
+    depends_on("cmake@3.25:", when="@20:", type="build")
```
thomas-bouvier, line 62: These look correct to me

thomas-bouvier (changes requested):

repos/spack_repo/builtin/packages/py_pyarrow/package.py
w8jcik: This package inherits from `PythonPackage`

```
class PyPyarrow(PythonPackage):
```

It means that it has `depends_on("python", type=("build", "run"))` implicitly.

When one generates a new Python package with

```
spack create -t python test
```

a message appears that there is no need to specify implicit dependencies

```
$ cat packages/py_test/package.py
...
    # FIXME: Only add the python/pip/wheel dependencies if you need specific versions
    # or need to change the dependency type. Generic python/pip/wheel dependencies are
    # added implicitly by the PythonPackage base class.
    # depends_on("python@2.X:2.Y,3.Z:", type=("build", "run"))
    # depends_on("py-pip@X.Y:", type="build")
    # depends_on("py-wheel@X.Y:", type="build")
```

Python packages in general don't explicitly say that they depend on Python. Should I still add it?

repos/spack_repo/builtin/packages/arrow/package.py
w8jcik: > I do not understand where do the changes come from

Sorry, my bad, I noticed that the CMake version constraints are incomplete comparing to Python module and copied over the ranges from the Python package. I had a wrong assumption that Python package and C++ part coming from the same repository have the same version requirements. I was checking it afterwards, but apparently not well enough.

I checked it again and yes you are right @manuelakuhn. It is exactly as you suggested

```
    depends_on("cmake@3.2:", type="build")
    depends_on("cmake@3.5:", when="@3:", type="build")
    depends_on("cmake@3.16:", when="@13:", type="build")
    depends_on("cmake@3.25:", when="@20:", type="build")
```

Applied the change.

repos/spack_repo/builtin/packages/arrow/package.py
w8jcik: Corrected, as described in https://github.com/spack/spack-packages/pull/5435#discussion_r3593782094. Sorry for the mistake.

w8jcik:
> would it be possible to add v24 from this PR too https://github.com/spack/spack-packages/pull/4809? Thanks!

@thomas-bouvier I have added `(py-)arrow@24` as requested, although it required additional changes to the package and to it's dependencies.

w8jcik:
@spackbot re-run pipeline

repos/spack_repo/builtin/packages/py_pyarrow/package.py
```diff
@@ -42,30 +47,50 @@ class PyPyarrow(PythonPackage):
...
         depends_on("py-setuptools-scm@8:+toml", when="@17:")
         depends_on("py-setuptools-scm", when="@16")
         depends_on("py-setuptools-scm@:7", when="@0.15:15")
-        depends_on("py-setuptools@64:", when="@17:")
-        depends_on("py-setuptools@40.1:", when="@10.0.1:")
-        depends_on("py-setuptools@38.6:", when="@7:")
-        depends_on("py-setuptools")
+
+        depends_on("py-setuptools@77:", when="@23:23")
```
manuelakuhn: This can be simplified to
```suggestion
        depends_on("py-setuptools@77:", when="@23")
```

repos/spack_repo/builtin/packages/xsimd/package.py
manuelakuhn, line 51: This has also restrictions on the cmake versions, see [`CMakeLists.txt`](https://github.com/xtensor-stack/xsimd/blob/14.2.0/CMakeLists.txt#L12)
```
depends_on("cmake@3.13:", type="build", when="@14.2:")
depends_on("cmake@3.8:", type="build", when="@12:")
depends_on("cmake@3.1:", type="build")
```

repos/spack_repo/builtin/packages/arrow/package.py
```diff
@@ -86,6 +89,7 @@ class Arrow(CMakePackage, CudaPackage):
     depends_on("utf8proc@2.7.0: +shared", when="+compute")
     depends_on("utf8proc@2.7.0: +shared", when="+gandiva")
     depends_on("utf8proc@2.7.0: +shared", when="+python")
+    depends_on("xsimd@14.1:", when="@24:")
```
manuelakuhn: The dependency is correct, but where is the version coming from? I could only find a version limit in the conda specs for the ci and that is 14.0 and not 14.1:
https://github.com/apache/arrow/blob/apache-arrow-24.0.0/ci/conda_env_cpp.txt#L50

repos/spack_repo/builtin/packages/py_libcst/package.py
```diff
@@ -15,12 +15,28 @@ class PyLibcst(PythonPackage):
...
+
+    with default_args(type="build"):
+        depends_on("py-setuptools-rust", type="build")
+        depends_on("py-setuptools-scm", type="build")
+        depends_on("py-setuptools", type="build")
+        depends_on("rust", type="build")
+
+    with default_args(type=("build", "run")):
+        depends_on("py-pyyaml@6.0.3:", when="@1.8.6: ^python@3.14:")
+        # depends_on("py-pyyaml-ft@8:", when="@1.8: ^python@3.13")
```
manuelakuhn: What is the reason why you did not add the dependency here. I know that the package is not in spack yet, but it seems to be a small step to just build it (looks pretty straight forward to me at first glance).

repos/spack_repo/builtin/packages/arrow/package.py
manuelakuhn: Thanks!

repos/spack_repo/builtin/packages/xsimd/package.py
```diff
@@ -10,15 +10,16 @@
...
     """C++ wrappers for SIMD intrinsics"""
 
-    homepage = "https://quantstack.net/xsimd"
-    url = "https://github.com/QuantStack/xsimd/archive/3.1.0.tar.gz"
-    git = "https://github.com/QuantStack/xsimd.git"
+    homepage = "https://xsimd.readthedocs.io"
+    url = "https://github.com/xtensor-stack/xsimd/archive/refs/tags/14.3.0.tar.gz"
+    git = "https://github.com/xtensor-stack/xsimd"
 
     maintainers("ax3l")
```
ax3l, line 17: Thx! LGTM :)

repos/spack_repo/builtin/packages/xsimd/package.py
w8jcik: Thank you for noticing. Added:

```
    depends_on("cmake@3.13:", type="build", when="@14.2:")
    depends_on("cmake@3.10:", type="build", when="@14:")
    depends_on("cmake@3.8:", type="build", when="@12:")
    depends_on("cmake@3.1:", type="build")
```

repos/spack_repo/builtin/packages/py_libcst/package.py
w8jcik: Added `py-pyyaml-ft` package.

repos/spack_repo/builtin/packages/arrow/package.py
w8jcik: I misinterpreted these lines

```
  // TODO(xsimd) bug fixed in xsimd 14.1.0
  // https://github.com/xtensor-stack/xsimd/pull/1266
#if XSIMD_VERSION_MAJOR < 14 || ((XSIMD_VERSION_MAJOR == 14) && XSIMD_VERSION_MINOR == 0)
```

Still, there is a constraint here

```cmake
if(ARROW_USE_XSIMD)
  resolve_dependency(xsimd
                     FORCE_ANY_NEWER_VERSION
                     TRUE
                     IS_RUNTIME_DEPENDENCY
                     FALSE
                     REQUIRED_VERSION
                     "14.0.0")
```

https://github.com/apache/arrow/blob/apache-arrow-24.0.0/cpp/cmake_modules/ThirdpartyToolchain.cmake#L2671

Changed `@14.1:` to `@14:`.

w8jcik:
## Updated build grid

### py-pyarrow

image

repos/spack_repo/builtin/packages/py_libcst/package.py
manuelakuhn: I saw that you went the extra step by making it work with the `libyaml` dependency instead of going pure python. That was more work that I wanted for you, but thank you very much!

repos/spack_repo/builtin/packages/arrow/package.py
manuelakuhn: Thanks

manuelakuhn (dismissed):
LGTM and builds fine on rocky linux 8 with gcc@15.

Thanks you for all the work you put into this!

manuelakuhn:
@spackbot re-run pipeline

manuelakuhn:
@thomas-bouvier and @ax3l do you want to take a last look as well, or can this be merged?

repos/spack_repo/builtin/packages/py_libcst/package.py
w8jcik: You are welcome. I just adapted the `py-pyyaml` package. The `libyaml` seems to be handled in the same way in both `py-pyyaml` and `py-pyyaml-ft`, so I assume it is valid.

repos/spack_repo/builtin/packages/py_pyyaml_ft/package.py
```diff
@@ -0,0 +1,66 @@
...
+    variant("libyaml", default=True, description="Use libYAML bindings")
+
+    depends_on("python@3.13:", type=("build", "run"))
+
+    depends_on("libyaml", when="+libyaml", type="link")
+
+    with default_args(type="build"):
+        depends_on("py-setuptools", type="build")
+        depends_on("py-packaging", type="build")
+        depends_on("py-cython@3.1:", type="build")
```
thomas-bouvier: I think these `type="build"` could be removed

thomas-bouvier (approved):
LGTM, thank you for your work @w8jcik

repos/spack_repo/builtin/packages/py_pyyaml_ft/package.py
w8jcik: Removed as you suggested

manuelakuhn (approved):

PR #5436, opened 2026-07-02
r-biocparallel: add v1.42.2 and v1.46.0

w8jcik:

### DIFF
--- a/repos/spack_repo/builtin/packages/r_biocparallel/package.py
+++ b/repos/spack_repo/builtin/packages/r_biocparallel/package.py
@@ -18,14 +18,8 @@ class RBiocparallel(RPackage):
 
     with default_args(get_full_repo=True):
         version("1.46.0", commit="a331dfd2695daec754304505bdbd7358eef009fa")  # bioc 3.23
-        version("1.44.0", commit="3d6f2f614a8cb6d052c3c89cb328cc7d454ff565")  # bioc 3.22
         version("1.42.2", commit="dbadae1fcbb3141202376d65c9a1224736c8825c")  # bioc 3.21
-        version("1.40.1", commit="db5b1458c78b3b8f4c3ea71fa4e0930dbf5960f1")  # bioc 3.20
-        version("1.38.0", commit="d180bc0969a0f719e01ba127af45173c5bd8e27f")  # bioc 3.19
-        version("1.36.0", commit="ba4ec2934579d5d1a1e2d0e69d44342221e49013")  # bioc 3.18
-        version("1.34.2", commit="77491b20e0c827e7dc9af56ab58824855ad75430")  # bioc 3.17
         version("1.34.0", commit="f3bbc0a2d38da034c50ca1e4704fc4ee99a2dc9e")
-        version("1.32.6", commit="994f4e72557f4fc47118f9b092ee448d7d5db299")  # bioc 3.16
         version("1.32.1", commit="6c85dbad596a74a6d3022173a4a11c6b81a4a2c2")
         version("1.30.4", commit="1229ebe9f6d8305f9f61e562464f83f9ba86e699")
         version("1.30.2", commit="e7e109f7a94dbfbc50f926be030c7ad8c1a053db")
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/r_biocparallel/package.py
```diff
@@ -29,11 +37,15 @@ class RBiocparallel(RPackage):
...
-    depends_on("cxx", type="build")  # generated
+    depends_on("cxx", type="build")
 
+    depends_on("r@4.1:", type=("build", "run"), when="@1.41.4:")
     depends_on("r@3.5.0:", type=("build", "run"), when="@1.28.3:")
-    depends_on("r-futile-logger", type=("build", "run"))
-    depends_on("r-snow", type=("build", "run"))
-    depends_on("r-codetools", type=("build", "run"), when="@1.30.4:")
+
+    depends_on("r-bh@1.87:", type=("build", "run"), when="@1.41.4:")
```
RikkiButler20:
```suggestion
    depends_on("r-bh@1.87:", type=("build", "run"), when="@1.40.4:")
```
I think you've got the wrong version

RikkiButler20 (changes requested):

repos/spack_repo/builtin/packages/r_biocparallel/package.py
w8jcik: Looking at the metadata file I believe that the version range `@1.41.4:` is correct

```sh
git clone https://git.bioconductor.org/packages/BiocParallel
cd BiocParallel
git diff 70bb9f91...ff5f3923 DESCRIPTION
```

```diff
diff --git a/DESCRIPTION b/DESCRIPTION
index 68fb74e..8fd95d1 100644
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Package: BiocParallel
 Type: Package
 Title: Bioconductor facilities for parallel evaluation
-Version: 1.41.2
+Version: 1.41.4
 Authors@R: c(
     person("Martin", "Morgan",
         email = "mtmorgan.bioc@gmail.com",
@@ -34,7 +34,7 @@ BugReports: https://github.com/Bioconductor/BiocParallel/issues
 biocViews: Infrastructure
 License: GPL-2 | GPL-3
 SystemRequirements: C++11
-Depends: methods, R (>= 3.5.0)
+Depends: methods, R (>= 4.1.0)
 Imports: stats, utils, futile.logger, parallel, snow, codetools
 Suggests: BiocGenerics, tools, foreach, BBmisc, doParallel,
     GenomicRanges, RNAseqData.HNRNPC.bam.chr14,
@@ -56,6 +56,6 @@ Collate: AllGenerics.R DeveloperInterface.R prototype.R
     progress.R ipcmutex.R worker-number.R
     utilities.R rng.R bpinit.R reducer.R worker.R
     bpoptions.R cpp11.R BiocParallel-defunct.R
-LinkingTo: BH, cpp11
+LinkingTo: BH (>= 1.87.0), cpp11
 VignetteBuilder: knitr
 RoxygenNote: 7.1.2
```

There is also no `1.40.4`. The authors went from `1.40.0` to `1.41.0`.

Although I might be missing something.

RikkiButler20: I meant to suggest v1.40.1.
I got the version from the Description in on github. Am I looking in the wrong place?
https://github.com/Bioconductor/BiocParallel/blob/db5b1458c78b3b8f4c3ea71fa4e0930dbf5960f1/DESCRIPTION#L59

RikkiButler20 (changes requested):

repos/spack_repo/builtin/packages/r_biocparallel/package.py
w8jcik: image

We might be both right. The authors changed this dependency twice. You are referring to the tip of BioC 3.20 (in pink on the picture) and I am referring to the trunk (violet).

A precise description of this situation is following

```diff
depends_on("r-bh@1.87:", type=("build", "run"), when="@1.40.1,1.41.4:")
```

That is all assuming that these `DESCRIPTION` files are accurate, which might not be the case anyway.

I got confused by this, thank you for noticing.

Updated the PR. Is the current solution fine?

RikkiButler20: Yeah this fix is good.

RikkiButler20 (approved):

PR #5437, opened 2026-07-02
libxs: Sync with CP2k recipe

mkrack:
The following changes are proposed:
- Fix typo in homepage url
- Prefer release url over auto-generated archive url for download
- Set default for variant fortran to false as in CMakeLists.txt
- Remove the not existing libxsmm option and the corresponding dependency
- Add variant tests and add dependence on BLAS for +tests
- Remove redundant shared variant LIBXS_SHARED from the CMake args

### DIFF
--- a/repos/spack_repo/builtin/packages/libxs/package.py
+++ b/repos/spack_repo/builtin/packages/libxs/package.py
@@ -11,37 +11,35 @@ class Libxs(CMakePackage):
     """LIBXS is a portable C library providing building blocks for memory
     operations, numerics, synchronization, and more -- with a focus on
     performance and minimal dependencies. Targets x86-64, AArch64, and RISC-V;
-    requires only a C89 compiler. Originally developed as part of LIBXSMM.
-    """
+    requires only a C89 compiler. Originally developed as part of LIBXSMM."""
+
+    homepage = "https://libxs.readthedocs.io/en/latest"
+    git = "https://github.com/hfp/libxs.git"
+    url = "https://github.com/hfp/libxs/releases/download/1.0.0/libxs-1.0.0.tar.gz"
 
     maintainers("hfp", "mkrack", "mtaillefumier")
-
-    homepage = "hhttps://libxs.readthedocs.io/en/latest"
-    url = "https://github.com/hfp/libxs/archive/refs/tags/1.0.tar.gz"
-    git = "https://github.com/hfp/libxs.git"
 
     license("BSD-3-Clause", checked_by="mkrack")
 
     version("main", branch="main")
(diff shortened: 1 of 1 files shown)


mtaillefumier (approved):
This comment is not blocking. 

There is no need to create a variant for building the tests as enabling the tests in `cmake` only build them.  

One can use this instead
```
define("BUILD_TESTING", pkg.run_tests),
```
as it allows the user to run the tests without installing them
```
spack install --test=root <package>
```
The check method also needs to be defined. the following example is for `ctest` but the method can be adapted to the specific testing framework used in `libxs`.
```
class CMakeBuilder(BuilderWithDefaults):
    build_time_test_callbacks = ["check"]

    def check(self):
        with fs.working_dir(self.build_directory):
            if self.generator == "Unix Makefiles":
                self.pkg._if_make_target_execute("test", jobs_env="CTEST_PARALLEL_LEVEL")
                self.pkg._if_make_target_execute("check")
            elif self.generator == "Ninja":
                self.pkg._if_ninja_target_execute("test", jobs_env="CTEST_PARALLEL_LEVEL")
                self.pkg._if_ninja_target_execute("check")
```

mkrack:
> This comment is not blocking.
> 
> There is no need to create a variant for building the tests as enabling the tests in `cmake` only build them.
> 
> One can use this instead
> 
> ```
> define("BUILD_TESTING", pkg.run_tests),
> ```
> 
> as it allows the user to run the tests without installing them
> 
> ```
> spack install --test=root <package>
> ```
> 
> The check method also needs to be defined. the following example is for `ctest` but the method can be adapted to the specific testing framework used in `libxs`.
> 
> ```
> class CMakeBuilder(BuilderWithDefaults):
>     build_time_test_callbacks = ["check"]
> 
>     def check(self):
>         with fs.working_dir(self.build_directory):
>             if self.generator == "Unix Makefiles":
>                 self.pkg._if_make_target_execute("test", jobs_env="CTEST_PARALLEL_LEVEL")
>                 self.pkg._if_make_target_execute("check")
>             elif self.generator == "Ninja":
>                 self.pkg._if_ninja_target_execute("test", jobs_env="CTEST_PARALLEL_LEVEL")
>                 self.pkg._if_ninja_target_execute("check")
> ```

@mtaillefumier Thanks for the hint! Would you mind if we postpone such an elaboration of the recipe to a follow-up PR, because we do not perform a testing of that kind on the CP2K side?

mtaillefumier:
No problem on my side. As I said it is a side comment worth keeping in mind for later.

RMeli (approved):

PR #5438, opened 2026-07-02
hpctoolkit: add support for jemalloc

mwkrentel:
HPCToolkit and Dyninst switched from using TBB malloc_proxy to jemalloc. This patch adds the spack support for that when both packages allow it. We don't want to run both, so disable jemalloc for older Dyninst.

Switch hpctoolkit package maintainer to blue42u (I'm retired).

ping @blue42u @hainest

### DIFF
--- a/repos/spack_repo/builtin/packages/hpctoolkit/package.py
+++ b/repos/spack_repo/builtin/packages/hpctoolkit/package.py
@@ -23,7 +23,7 @@ class Hpctoolkit(AutotoolsPackage, MesonPackage):
 
     homepage = "https://hpctoolkit.org"
     git = "https://gitlab.com/hpctoolkit/hpctoolkit.git"
-    maintainers("mwkrentel")
+    maintainers("blue42u")
 
     tags = ["e4s"]
 
@@ -196,6 +196,7 @@ class Hpctoolkit(AutotoolsPackage, MesonPackage):
     depends_on("xz+pic libs=static", type="link", when="@:2023.08")
     depends_on("yaml-cpp@0.7.0: +shared", when="@2022.10:")
     depends_on("googletest@1.8.1: +gmock", type="test", when="@2025:")
+    depends_on("jemalloc", when="@develop: ^dyninst@master:")
 
     depends_on("zlib-api")
     depends_on("zlib+shared", when="^[virtuals=zlib-api] zlib")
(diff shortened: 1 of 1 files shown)


mwkrentel:
@spackbot fix style

repos/spack_repo/builtin/packages/hpctoolkit/package.py
```diff
@@ -196,6 +196,7 @@ class Hpctoolkit(AutotoolsPackage, MesonPackage):
     depends_on("xz+pic libs=static", type="link", when="@:2023.08")
     depends_on("yaml-cpp@0.7.0: +shared", when="@2022.10:")
     depends_on("googletest@1.8.1: +gmock", type="test", when="@2025:")
+    depends_on("jemalloc", when="@develop: ^dyninst@master:")
```
hainest: Do you want to use jemalloc even after dyninst@14 is released? If so, I'd change this to


```suggestion
    depends_on("jemalloc", when="@develop: ^dyninst@13:")
```

and similarly below.

mwkrentel: @hainest At present, jemalloc for dyninst 13 would be wrong because dyninst 13
pulls in TBB malloc proxy.  At best, it would build jemalloc and then discard it.

My plan is to change the dependency to `^dyninst @14:` if/when Dyninst
14 is ever released.  But I want to wait until Dyninst 14 exists.
Any word on when that might be?

blue42u: Speaking as a future maintainer of this package, I'd rather have the dependency pinned down to `^dyninst@14:` or `^dyninst@13.1:` early, even if those versions aren't released yet. Saves a PR cycle later and will "just work" day-0 when the versions do release.

For the same reason, I'd rather use `@2026.1:` as the HPCToolkit version range here instead of `@develop:`.

repos/spack_repo/builtin/packages/hpctoolkit/package.py
```diff
@@ -480,6 +481,13 @@ def meson_args(self):
         if spec.satisfies("@:2024.01"):
             args.append(f"--native-file={self.gen_prefix_file()}")
 
+        # Dyninst before master June 2026 pulls in tbbmalloc_proxy,
+        # and we don't really want to run both.
+        if spec.satisfies("@develop:"):
+            args.append(
+                "-Djemalloc=" + ("enabled" if spec.satisfies("^dyninst@master:") else "disabled")
```
blue42u:
```suggestion
        if spec.satisfies("@2026.1:"):
            args.append(
                "-Djemalloc=" + ("enabled" if spec.satisfies("^dyninst@13.1:") else "disabled")
```

See prior comment for rationale.

mwkrentel: @blue42u  @hainest  I refined the dependencies.  Spack seems to be Ok with 
mentioning versions that don't exist, so Ok.

Make sure that toolkit 2026.1.0 is either derived from develop or else
cherry picks the jemalloc commit.

blue42u (approved):
LGTM

mwkrentel:
@hainest Are we ready to merge?

hainest (approved):

hainest:
> @hainest Are we ready to merge?

LGTM. Thanks!

PR #5445, opened 2026-07-03
geant4: add examples variant

madbaron:
This PR adds a variant to geant4 to allow a build without examples.


RikkiButler20 (commented):
Looks good to me.
@drbenmorgan @sethrj Do you have any feedback?

sethrj (approved):
LGTM thanks. The option was added in 2017 so I think it might be ignored for very old versions of G4, but that's not too important.

PR #5457, opened 2026-07-05
neovim: add v0.12.4

mcmehrtens:
- add v0.12.4
- removed the jemalloc dependency
  - Neovim got rid of all jemalloc support in https://github.com/neovim/neovim/pull/9526 back in 2019
  - all the versions in the Neovim Spack package seem to have no jemalloc dependency
- changed lua-mpack dependency from `when="@0:10"` to `when="@:0.9"`
  - starting in 0.10.0, Neovim [vendors its own lua-mpack](https://github.com/neovim/neovim/blob/27fb62988e922c2739035f477f93cc052a4fee1e/src/nvim/CMakeLists.txt#L369) vs. [:0.9 where it was still needed as an external dependency](https://github.com/neovim/neovim/blob/8744ee8783a8597f9fce4a573ae05aca2f412120/CMakeLists.txt#L171).
- changed gperf to a build-time dependency since it is a build-, not link-, time dependency (no libraries to link against!)
- reordered 0.10.1 so it was in its proper order (descending order)
- libuv: only version 1.28.0 is required (see [0.5.1](https://github.com/neovim/neovim/blob/v0.5.1/CMakeLists.txt#L384) and [0.12.4](https://github.com/neovim/neovim/blob/v0.12.4/src/nvim/CMakeLists.txt#L32))
  - I think this was previously set to the version that Neovim pinned, not its true dependency floor
- msgpack-c: only version 1.0.0 is required (see [0.5.1](https://github.com/neovim/neovim/blob/v0.5.1/CMakeLists.txt#L387) and [0.10.4](https://github.com/neovim/neovim/blob/v0.10.4/src/nvim/CMakeLists.txt#L35))
  - same situation as libuv; I think this was set to the pinned version, not the required version
 
additional updates:
- deprecate v0.11 and older (nothing depends on Neovim)
- add @mcmehrtens to maintainers
- add supplier field


mcmehrtens:
I did successfully build and test all versions and variants on a Linux x86_64 box with the latest version of Spack.

bernhardkaindl:
@mcmehrtens, because you need to restart CI (it appears the high-load situation causes some issues), could you possibly deprecate some older versions? I assume spack does not need to support so many old versions of neovim?

mcmehrtens:
> could you possibly deprecate some older versions?

@bernhardkaindl gladly! How far back would you like me to go?

bernhardkaindl:
@mcmehrtens, whatever you feel comfortable with!

At least in this repo, I don't see `depends_on()` for neovim (that could depend on a version range), so if you don't need specific older versions, my personal suggestion would be the idea to just make the recipe simpler for the spack solver (less overhead, just as one of many/all, not specifically neovim) and future package maintainance simple/less versions to look at.

So I'd personally deprecate all versions 0.11 and older versions, or whatever would make sense with a view to cleanup old versions, because once versions are deprecated for one spack release, the deprecated versions can be removed in the next version cycle. Anyway, this is just an opportunity to start the process, so take this just as an idea.

mcmehrtens:
@bernhardkaindl take a look at the latest diffs—deprecated @:0.11, added supplier, and added myself as a maintainer.

Let me know if anything else sticks out to you.

bernhardkaindl (approved):
```py
spack find neovim
-- linux-ubuntu24.04-skylake / %c=gcc@13.3.0 --------------------
neovim@0.12.4
```

PR #5462, opened 2026-07-06
r-languageserver, r-lintr, r-collections and r-xmlparsedata: new packages

w8jcik:
Installation of `r-languageserver` offers R language support in VSCode and other LSP-enabled editors. The other three packages are the dependencies of `r-languageserver` that were missing from `spack-packages`.

### DIFF
--- a/repos/spack_repo/builtin/packages/r_collections/package.py
+++ b/repos/spack_repo/builtin/packages/r_collections/package.py
@@ -1,0 +1,20 @@
+# Copyright Spack Project Developers. See COPYRIGHT file for details.
+#
+# SPDX-License-Identifier: (Apache-2.0 OR MIT)
+
+from spack_repo.builtin.build_systems.r import RPackage
+
+from spack.package import *
+
+
+class RCollections(RPackage):
+    """Provides high performance container data types such as queues, stacks, deques, dicts
+    and ordered dicts."""
+
+    cran = "collections"
+
+    license("MIT")
+
+    version("0.3.12", sha256="60e63ee65bc1889e54a008410f53cb5b643f2fffdb926c3ed12316094709c60d")
+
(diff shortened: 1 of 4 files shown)


bernhardkaindl:
@w8jcik I built all added packages, there is only one minor issue:

Building the dependency `r-fs` (at least as r-fs@2.1.0) failed without libuv development package installed on the host:
```py
  --------------------------- [CONFIGURE] --------------------------------
  Configuration failed because libuv was not found. Try installing:
   * deb: libuv1-dev (Debian, Ubuntu, etc)
   * rpm: libuv-devel (Fedora, EPEL)
   * brew: libuv (OSX)
  Alternatively set environment variable USE_BUNDLED_LIBUV=1 to build a static
  version of libuv that is included with this package.
  -------------------------- [ERROR MESSAGE] ---------------------------
> <stdin>:1:10: fatal error: uv.h: No such file or directory
  compilation terminated.
  --------------------------------------------------------------------
  ERROR: configuration failed for package 'fs'
```
After installing the libuv1-devel package, the installation worked, but of course the idea would be that the dependency should be part of the spack build. I'd plan to approve nonetheless, but if you do a new PR for r packages, if it is easy to do, this could be added to such PR.

I'm not asking to add it to this PR because it's GitLab CI has already passed and there is no need to spend the CI resources to re-run CI just for this minor issue. It would be better to add it to some PR that might anyway need a new GitLab CI pipeline run in the future when it is submitted.

bernhardkaindl (approved):
@w8jcik, also see https://github.com/spack/spack-packages/pull/5462#issuecomment-4985886909

w8jcik:
This https://github.com/spack/spack-packages/pull/5608 should fix missing `libuv`.

PR #5464, opened 2026-07-06
(py-)xgboost: add v2.1.4

w8jcik:
The latest 2.x release supports newer CUDA and newer GPUs  
https://github.com/dmlc/xgboost/releases/tag/v2.1.4

### DIFF
--- a/repos/spack_repo/builtin/packages/py_xgboost/package.py
+++ b/repos/spack_repo/builtin/packages/py_xgboost/package.py
@@ -20,6 +20,7 @@ class PyXgboost(PythonPackage):
     license("Apache-2.0")
     maintainers("adamjstewart")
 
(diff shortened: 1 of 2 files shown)


w8jcik:
@spackbot re-run pipeline

adamjstewart (approved):
2.1.4 is quite old, could also add 3.3.0 if you want.

w8jcik:
> 2.1.4 is quite old, could also add 3.3.0 if you want.

Thank you for merging.

Yes I am aware that there are newer versions, but there seems to be many changes between 2.x and 3.x which I am not confident to package. I needed xgboost as a dependency of [PyProphet](https://github.com/PyProphet/pyprophet) which is satisfied by 2.1.4.

PR #5486, opened 2026-07-07
py-torch: ensure setuptools is not unnecessarily constrained for 2.10:

wdconinc:
This PR fixes what (I think) is a typo in `py-torch` introduced with the addition of v2.10. Currently, for v2.10 and later, setuptools is still constrained to at most v79.


adamjstewart (approved):

wdconinc:
@spackbot run pipeline

PR #5493, opened 2026-07-07
py-psyclone: add v3.2.0, v3.2.1, v3.2.2, v3.3.1

arporter:


arporter:
This PR is blocked by #5490 as it needs the py_fparser package to be updated first.

repos/spack_repo/builtin/packages/py_psyclone/package.py
```diff
@@ -44,18 +48,23 @@ class PyPsyclone(PythonPackage):
     # Current dependencies
     depends_on("py-setuptools", type="build")
     depends_on("py-pyparsing", type=("build", "run"))
-    depends_on("py-fparser@0.2.0:", type=("build", "run"), when="@3.0.0:")
+    depends_on("py-fparser@0.2.4", type=("build", "run"), when="@3.3")
```
bernhardkaindl: @arporter
This appears to remove the dependency on py-fparser for @3.0 to @3.2 if I'm right.

I guess you might have just wanted to add the new line, not add the old line?
```suggestion
    depends_on("py-fparser@0.2.0:", type=("build", "run"), when="@3.0.0:")
    depends_on("py-fparser@0.2.4", type=("build", "run"), when="@3.3")
```
The PR is still building 3.0.0 (I've an unrelated builder error of a dependency of 3.1.0), but I just wanted to ask if the replacement of the line was intentional and instead just the new line should be added.

arporter: Oops, well spotted thanks. I've updated the 'historical dependencies' now so that 3.2 onwards depends on 0.2.1 of fparser.

arporter: Correction, that didn't seem to work either so I've changed it so that 3.2 depends on 0.2.1 of fparser.

arporter: `spack spec py_psyclone@3.3` now looks correct. As does `@3.2`.

arporter:
Thanks @bernhardkaindl, should be ready for another look assuming the CI is happy.

repos/spack_repo/builtin/packages/py_psyclone/package.py
```diff
@@ -44,18 +48,23 @@ class PyPsyclone(PythonPackage):
     # Current dependencies
     depends_on("py-setuptools", type="build")
     depends_on("py-pyparsing", type=("build", "run"))
-    depends_on("py-fparser@0.2.0:", type=("build", "run"), when="@3.0.0:")
+    depends_on("py-fparser@0.2.4", type=("build", "run"), when="@3.3:")
```
bernhardkaindl: @arporter, did you check the `fparser` dependencies of the versions before `@3.3`? AFACIS, '@3.0' to '@3.2' also still need `fparser` but 0.2.0/0.2.1:

https://github.com/stfc/PSyclone/blob/v3.0.0/setup.py#L164
https://github.com/stfc/PSyclone/blob/v3.1.0/setup.py#L164
```py
install_requires=['pyparsing', 'fparser>=0.2.0', 'configparser',
```

While this says >=, as the newer requires are exact. you might want to use exact depdenency version, but if you agree that it shall be >= you can also use `@0.2.0:` for the spack spec:
```py
    depends_on("py-fparser@0.2.0:", type=("build", "run"), when="@3.2:")
```
vs
```py
    depends_on("py-fparser@0.2.0", type=("build", "run"), when="@3.2:")
```

'@3.2' seems to have updated from 0.2.0 to 0.2.1:

https://github.com/stfc/PSyclone/blob/v3.2.0/setup.py#L172
```py
        install_requires=['pyparsing', 'fparser>=0.2.1', 'configparser',
```

With 3.2.2, they changed to a pinned version:
https://github.com/stfc/PSyclone/blob/3.2.2/setup.py#L172
```py
        install_requires=['pyparsing', 'fparser==0.2.1', 'configparser',
```

Example:
```suggestion
    depends_on("py-fparser@0.2.0", type=("build", "run"), when="@3.0:")
    depends_on("py-fparser@0.2.1", type=("build", "run"), when="@3.2:")
    depends_on("py-fparser@0.2.4", type=("build", "run"), when="@3.3:")
```

bernhardkaindl (commented):
@arporter, can you comment on the different dependencies on `fparser`?

Should these older dependencies be kept and maintained like I propose in this review?

bernhardkaindl (changes requested):
@arporter, pleease see review the new comment below https://github.com/spack/spack-packages/pull/5493#pullrequestreview-4678608645

repos/spack_repo/builtin/packages/py_psyclone/package.py
arporter: Now I have:
version >=3.3 depends on fparser 0.2.4
3.2.* depends on fparser 0.2.1
3.0<= version < 3.2 depends on fparser 0.2.0
which covers it I think.

repos/spack_repo/builtin/packages/py_psyclone/package.py
```diff
@@ -44,18 +48,23 @@ class PyPsyclone(PythonPackage):
...
+    depends_on("py-sympy@1.13.3", type=("build", "run"), when="@3.3:")
     depends_on("py-termcolor", type=("build", "run"))
 
     # Historical dependencies
     depends_on("py-six", type=("build", "run"), when="@2.0.0:2.3.1")
     depends_on("py-jsonschema@3.0.2", type=("build", "run"), when="@2.1.0:2.4.0")
 
     # Test cases fail without compatible versions of py-fparser:
+    depends_on("py-fparser@0.2.1", type=("build", "run"), when="@3.2")
+    depends_on("py-fparser@0.2.0", type=("build", "run"), when="@3.0:")
```
bernhardkaindl: @arporter Sorry, due to using a pinned version dependency here (`"py-fparser@0.2.0"`) you cannot use an overlapping `when=` as `0.2.0` `and` `0.2.1` would apply to @3.2 and newer.

Therefore, as long as you pin the version, you must use a non-overlapping `when`:
```suggestion
    depends_on("py-fparser@0.2.0", type=("build", "run"), when="@3.0:3.1")
```
Otherwise, py-psyclone@3.2 can't be concretized because of the conflict with the line above.

bernhardkaindl (changes requested):
Also, as there is a dependency on py-setuptools-scm-git-archive, which does not build with newer setuptools, add an upper version constraint:
```py
--- a/repos/spack_repo/builtin/packages/py_setuptools_scm_git_archive/package.py
+++ b/repos/spack_repo/builtin/packages/py_setuptools_scm_git_archive/package.py
@@ -23,5 +23,5 @@ class PySetuptoolsScmGitArchive(PythonPackage):
     version("1.1", sha256="6026f61089b73fa1b5ee737e95314f41cb512609b393530385ed281d0b46c062")
     version("1.0", sha256="52425f905518247c685fc64c5fdba6e1e74443c8562e141c8de56059be0e31da")
 
+    depends_on("py-setuptools@:80", type="build")
     depends_on("py-setuptools-scm@:7", type=("build", "run"))
```
Without this, py-setuptools-scm-git-archive fails to build with the error 
```py
  ModuleNotFoundError: No module named 'pkg_resources'
```
from the removed `pkg_resources` (https://setuptools.pypa.io/en/stable/history.html#v82-0-0) module in `setuptools` hidden in the build log. 

As one can't build `py-psyclone` fresh with no older setuptools already installed, please apply this fix for the dependency tree and the requested fix to not overlap the pinned version dependency.

repos/spack_repo/builtin/packages/py_psyclone/package.py
arporter: Oh, apologies. I thought it would sort this out automatically.

arporter:
Thanks @bernhardkaindl and apologies for the to-and-fro on this - I'm not very familiar with Spack. Hopefully I've understood correctly: I've updated the py-setuptools-scm-git-archive package to limit it to <= v80 of py-setuptools. I'd also misunderstood how Spack coped with overlapping dependencies and have fixed those.

I also realised that we didn't actually list py-setuptools-scm-git-archive as a PSyclone dependency (even though it used to be) so I've added that.

repos/spack_repo/builtin/packages/py_psyclone/package.py
bernhardkaindl: Oh, on the `py-fparser` dependencies, I was confused. I started this with "This appears to remove the dependency on py-fparser for @3.0 to @3.2 if I'm right.`, I indeed didn't read this correctly:

```diff
     # Current dependencies
     depends_on("py-setuptools", type="build")
     depends_on("py-pyparsing", type=("build", "run"))
-    depends_on("py-fparser@0.2.1:", type=("build", "run"), when="@3.2.0:")
+    depends_on("py-fparser@0.2.4", type=("build", "run"), when="@3.3")
... 
     # Test cases fail without compatible versions of py-fparser:
+    depends_on("py-fparser@0.2.1", type=("build", "run"), when="@3.2")
```

The problem was that at `...`, more lines exist in the initial commit of the PR, so I didn't connect the two patch hunks as a move with a slight modification.

This is the reason why depends_on() blocks are usually grouped by dependency, not grouped by current dependencies and older versions.

I guess for the future, it would be better to move the
```py
depends_on("py-fparser@0.2.4", type=("build", "run"), when="@3.3")
```
to the block of `"py-fparser@` dependencies to keep them all together.

repos/spack_repo/builtin/packages/py_psyclone/package.py
bernhardkaindl: That would be nice, but the way it works is that more narrow rules effectively add on top of other, more wide rules and the more narrow rule apparently wins.

repos/spack_repo/builtin/packages/py_psyclone/package.py
```diff
@@ -44,18 +48,24 @@ class PyPsyclone(PythonPackage):
...
-    depends_on("py-sympy@1.6.1:", type=("build", "run"), when="@2.2.0:")
+    # Anything more recent than sympy 1.13.3 currently causes problems
+    # https://github.com/stfc/PSyclone/issues/3474
+    depends_on("py-sympy@1.13.3", type=("build", "run"))
     depends_on("py-termcolor", type=("build", "run"))
 
     # Historical dependencies
     depends_on("py-six", type=("build", "run"), when="@2.0.0:2.3.1")
     depends_on("py-jsonschema@3.0.2", type=("build", "run"), when="@2.1.0:2.4.0")
+    depends_on("py-setuptools-scm-git-archive", type="build", when="@:3.3.0")
```
bernhardkaindl: > Thanks @bernhardkaindl and apologies for the to-and-fro on this - I'm not very familiar with Spack. Hopefully I've understood correctly: I've updated the py-setuptools-scm-git-archive package to limit it to <= v80 of py-setuptools. I'd also misunderstood how Spack coped with overlapping dependencies and have fixed those.

No problem, it is something we should be able to automatically check, I hope that check could be done soon.

> I also realised that we didn't actually list py-setuptools-scm-git-archive as a PSyclone dependency (even though it used to be) so I've added that.

When I said that there is a dependency on py-setuptools-scm-git-archive, I didn't mean to say it is a direct dependency, I'm sorry that I didn't clearly explain that when saying that. It is a dependency of py-fparser, but only for a range of the versions of py-fparser:

```py
grep archive repos/spack_repo/builtin/packages/py_fparser/package.py
    depends_on("py-setuptools-scm-git-archive", type="build", when="@0.1:0.2.2")
```

As psyclone does not seem to use it directly (at least I can't find it directly in its setup.py/pyproject.toml files and not through), please remove it again:

```suggestion
```

PS: Unfotunately, only after you removed that line, you can see if and where the depedency chain pulls it into the build. Eg. building py-psyclone@3.3.1 does not pull it into the build because it uses the newer py-fparser@2.4 versions that no longer use `py-setuptools-scm-git-archive`. But `@3.2.1` pull it though the older version of `^py-fparser@0.2.1`
```py
spack spec py-psyclone@3.2.1 |grep -B7 git-archive|sed 's/ build_system.*//'
...
 -       ^py-fparser@0.2.1
[+]          ^py-importlib-metadata@9.0.0
[+]              ^py-setuptools-scm@9.2.2+toml
[+]                  ^git@2.53.0+man+nls+perl+subtree~tcltk
[+]                      ^curl@8.20.0~gssapi~ldap+libidn2~librtmp~libssh~libssh2+nghttp2
[+]              ^py-zipp@3.23.1
[+]                  ^py-coherent-licensed@0.5.2
[+]          ^py-setuptools-scm-git-archive@1.4.1
```

bernhardkaindl (changes requested):
I think this should be the final change needed :-)

repos/spack_repo/builtin/packages/py_psyclone/package.py
```diff
@@ -44,18 +48,24 @@ class PyPsyclone(PythonPackage):
     # Current dependencies
     depends_on("py-setuptools", type="build")
     depends_on("py-pyparsing", type=("build", "run"))
-    depends_on("py-fparser@0.2.0:", type=("build", "run"), when="@3.0.0:")
+    depends_on("py-fparser@0.2.4", type=("build", "run"), when="@3.3:")
```
bernhardkaindl: Final recommendation (a suggestion, not required, but I'd recommend it):
Put all depdencies on the same package together.

Part 1: Remove it from where it is mixed into the list of other dependencies:
```suggestion
```

repos/spack_repo/builtin/packages/py_psyclone/package.py
```diff
@@ -44,18 +48,24 @@ class PyPsyclone(PythonPackage):
...
     depends_on("py-termcolor", type=("build", "run"))
 
     # Historical dependencies
     depends_on("py-six", type=("build", "run"), when="@2.0.0:2.3.1")
     depends_on("py-jsonschema@3.0.2", type=("build", "run"), when="@2.1.0:2.4.0")
+    depends_on("py-setuptools-scm-git-archive", type="build", when="@:3.3.0")
+    depends_on("py-sympy@1.6.1:", type=("build", "run"), when="@2.2:3.2")
 
     # Test cases fail without compatible versions of py-fparser:
+    depends_on("py-fparser@0.2.1", type=("build", "run"), when="@3.2")
```
bernhardkaindl, line 61: Part 2: Move it on top of the block of `py-fparser` dependencies:
```suggestion
    depends_on("py-fparser@0.2.4", type=("build", "run"), when="@3.3:")
    depends_on("py-fparser@0.2.1", type=("build", "run"), when="@3.2")
```

repos/spack_repo/builtin/packages/py_psyclone/package.py
arporter: Apologies again - my brain doesn't seem to be working very well at the minute. I don't know why I decided that it was a direct dependence. I've removed that line again now.

repos/spack_repo/builtin/packages/py_psyclone/package.py
arporter: Yes, good idea. Done now.

repos/spack_repo/builtin/packages/py_psyclone/package.py
arporter: Done, plus some other tidying.

arporter:
Thanks @bernhardkaindl, should be ready for another look now, CI permitting.

repos/spack_repo/builtin/packages/py_psyclone/package.py
bernhardkaindl: No worries, my message wasn't clear at the time because at that point, I didn't investigate why this indirect dependency broke my test build.

repos/spack_repo/builtin/packages/py_psyclone/package.py
bernhardkaindl: Ah, that looks a lot better!

bernhardkaindl (approved):
LGTM, successfully built these versions on a minimal Ubuntu 24.04 host:
```py
$ gh extension install spack/gh-spack-pr;gh spack-pr run 5493
Passed specs:
- `py-psyclone@=3.3.1`
- `py-psyclone@=3.2.2`
- `py-psyclone@=3.2.1`
- `py-psyclone@=3.2.0`
```

Merged, I'm glad this PR can be ticked off ;)

PR #5501, opened 2026-07-07
libxml2: update handling for iconv

psakievich:
Inspecting installed configure scripts for libxml2 showed invalid include paths: `-I//include`. This was traced to using `glibc` as the `iconv` provider. Update the configure args to use libxml2's built-in option for `libc` vendored `iconv`.

Reasoning and inspection of libxml2s build system were AI assisted through direct source inspection. Contribution is the authors.


repos/spack_repo/builtin/packages/libxml2/package.py
```diff
@@ -211,12 +211,24 @@ def import_module_test(self):
...
+    def _iconv_option(self):
+        """Special handling for iconv
+        iconv can be vendored from libx and in these cases the assumption of a prefix
+        is not valid. libxml2's build system allows 3 options (yes, no, or a prefix)
+        to account for this. When iconv comes from glibc or musl pass 'yes' as the
+        build system intends.
+        """
+        iconv = self.spec["iconv"]
+        if iconv.name in ("glibc", "musl"):
+            return "yes"
```
AlexanderRichert-NOAA, line 223: Should there be an 'else no' case?

psakievich: Not here because we require `^iconv` when `build_system=autotools` (see L60).

AlexanderRichert-NOAA: Ah I see, I misread the logic here, this makes sense. Thanks.

AlexanderRichert-NOAA (approved):

psakievich:
Rebased, first gitlab-ci pipeline failure was due to an inability to fetch source code for a package downstream of libxml2.

cmelone (approved):

PR #5503, opened 2026-07-07
py-numba: add 0.66.0, py-llvmlite: add 0.48.0

mathomp4:
Adding py-numba 0.66.0 and py-llvmlite 0.48.0 which essentially adds `llvm@22` support.

This is mainly done so I can use an already-installed `llvm@22` (from brew or by hand) without needing to build it via spack


mathomp4:
Looks like no maintainers. I requested approvers from what github suggested

mathomp4:
> Looks like no maintainers. I requested approvers from what github suggested

I did the "right" thing and added myself as a maintainer. At least then someone will be pinged for updates 🙂

manuelakuhn (approved):

PR #5504, opened 2026-07-07
openblas: fix dylib install name for cmake build on macOS

mathomp4:
When building with `build_system=cmake` on macOS, the install name of `libopenblas.dylib` was set to the build-stage path rather than the install prefix. This caused all downstream packages (e.g. `py-numpy`) to fail at import time with "Library not loaded: <build_stage_path>".

Add `CMAKE_INSTALL_NAME_DIR` set to `spec.prefix.lib` when `platform=darwin` so the dylib embeds the correct install-time path.

The makefile build system handles this correctly and is the default on macOS, which is why this has not been caught upstream.

---

I had help finding this fix with Claude Sonnet 4.6

### DIFF
--- a/repos/spack_repo/builtin/packages/openblas/package.py
+++ b/repos/spack_repo/builtin/packages/openblas/package.py
@@ -558,10 +558,7 @@ class MakefileBuilder(makefile.MakefileBuilder):
         if self.spec.satisfies("target=m1:"):
             # Use custom simplified target for macOS ARM procesors:
             # GENERIC target results in SIGILL
-            make_defs += [
-                "TARGET=VORTEX",
-                "NO_SVE=1",
-            ]
+            make_defs += ["TARGET=VORTEX", "NO_SVE=1"]
         else:
             # Try to intelligently add target and architecture flags
             make_defs += self._microarch_target_args()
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/openblas/package.py
```diff
@@ -691,6 +688,13 @@ def cmake_args(self):
             self.define("CMAKE_POLICY_DEFAULT_CMP0042", "NEW"),
         ]
 
+        # On macOS the cmake build sets the dylib install name to the build-stage
+        # path instead of the install prefix, causing downstream packages (e.g.
+        # py-numpy) to embed a broken rpath.  Setting CMAKE_INSTALL_NAME_DIR fixes
+        # the install name to the real installed lib directory.
+        if self.spec.satisfies("platform=darwin"):
+            cmake_defs += [self.define("CMAKE_INSTALL_NAME_DIR", self.spec.prefix.lib)]
```
alecbcs, line 696: Is there any downside to setting this variable when not on MacOS?

sethrj:
IIRC the cmake for openblas is something of a mess…

repos/spack_repo/builtin/packages/openblas/package.py
mathomp4: I think the issue is more due to RPATH handling on macOS? At least Claude seemed to imply that. So I suppose that's why I decided to protect it.

I guess if someone out there with a Linux box handy could try a build of `py-numpy` say along with `openblas` built as:
```yaml
  openblas:
    prefer:
    - build_system=cmake
    variants: ~dynamic_dispatch
```
If this patch is needed, the build would fail.

mathomp4:
> IIRC the cmake for openblas is something of a mess…

@sethrj Maybe true. But at this point, I sort of keep my CMake build on macOS as default as I think I'm the one that finds these bugs! 😄

repos/spack_repo/builtin/packages/openblas/package.py
bernhardkaindl: @mathomp4, as an answer for your comment, I ran this command successfully on Ubuntu 24.04 with `%c,cxx,fortran=gcc@13.3.0` (it built `openblas~dynamic_dispatch` using cmake and then `py-numpy@2.5.0` using it), so as far as I understand, this PR does not need further changes for Linux
```py
spack install py-numpy@2.5.0 ^openblas~dynamic_dispatch build_system=cmake
```

bernhardkaindl (approved):

PR #5509, opened 2026-07-07
fortran: disable F77/FC if no fortran virtual, hdf5-blosc: Dedicated plugin dir and use libtool for installs

### LINKED ISSUE #5507: hdf5-blosc: Ill-formatted install breaks hdf5 plugin loader on large installs
HDF5's plugin loader walks every library in `HDF5_PLUGIN_PATH` searching for symbols of the plugin it is expecting. Currently `hdf5-blosc` installs it's plugin inside the `prefix.lib` directory. When this gets combined into a view it can be installed next to hundreds of dynamic libraries. This leads to nasty runtime errors since hdf5's own libraries are also installed here.
 
Proposed fix is to install plugins in a dedicated `prefix.lib.hdf5_plugins` directory to isolate the consequences of hdf5's plugin-loader design.

psakievich:
- Isolate plugin installation to `prefix.lib.hdf5_plugins`
- Use `libtool --mode=install` to install instead of a custom copy script that leaked some rpath information from the build dir.

Example below where the first entry is still pointing to the stage directory:
``` shell
  $ readelf -d $store/hdf5-blosc-master-<hash>/lib/libblosc_plugin.so.0.0.0 | grep RPATH
  (RPATH)  Library rpath: [$stage/spack-stage-hdf5-blosc-master-<hash>/spack-src/src/.libs:
                           $store/hdf5-blosc-master-<hash>/lib:
                           $store/gcc-runtime-14.2.0-<hash>/lib:
                           $store/zstd-1.5.2-<hash>/lib: ...]
```
`libtool --mode=install` is designed to update the rpath. It also makes sense to install via `libtool` since we are building with it as well.

Diagnosis and fixes in this PR were AI assisted.

Todo:
- [ ] Test on Darwin

Closes #5507

### DIFF
--- a/repos/spack_repo/builtin/packages/hdf5_blosc/package.py
+++ b/repos/spack_repo/builtin/packages/hdf5_blosc/package.py
@@ -10,22 +10,6 @@ from spack_repo.builtin.build_systems.generic import Package
 from spack.package import *
 
 
-def _install_shlib(name, src, dst):
-    """Install a shared library from directory src to directory dst"""
-    if sys.platform == "darwin":
-        shlib0 = name + ".0.dylib"
-        shlib = name + ".dylib"
-        install(join_path(src, shlib0), join_path(dst, shlib0))
-        symlink(shlib0, join_path(dst, shlib))
-    else:
-        shlib000 = name + ".so.0.0.0"
-        shlib0 = name + ".so.0"
-        shlib = name + ".dylib"
-        install(join_path(src, shlib000), join_path(dst, shlib000))
-        symlink(shlib000, join_path(dst, shlib0))
-        symlink(shlib0, join_path(dst, shlib))
-
-
 class Hdf5Blosc(Package):
     """Blosc filter for HDF5"""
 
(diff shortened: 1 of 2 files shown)


bernhardkaindl:
@psakievich, could you review the suggestions from the new review?

psakievich:
@spackbot fix style

bernhardkaindl (dismissed):

psakievich:
Doesn't build on vanilla Mac (dylib). I think the issue is libtool is thinking there is a fortran compiler even though there isn't one. Then libtool is built without support of dynamic libraries because its configure script says the fortran compiler can't support building them.

haampie:
Using `$libtool/bin/libtool` is problematic, cause it is a configured script for the compiler toolchain libtool was built with. If you mix compilers between hdf5-bloc and libtool, you get a misconfigured libtool.

Normally libtool is generated as part of `./configure`. If that's missing, idk what the simplest way is to generate it other than pulling in autotools deps and an ad-hoc `configure.ac`.

Another option suggested by gemini is a build dep on `slibtool`. Dunno if that has the same issue, if so, it'd be a `resource` instead of a `depends_on`, but I hope it could be a build dep.

PR #5519, opened 2026-07-08
proteus: new package

cmelone:
https://github.com/Olympus-HPC/proteus

@zwfink @ggeorgakoudis

### DIFF
--- a/repos/spack_repo/builtin/packages/proteus/package.py
+++ b/repos/spack_repo/builtin/packages/proteus/package.py
@@ -2,9 +2,10 @@
 #
 # SPDX-License-Identifier: (Apache-2.0 OR MIT)
 
+from spack_repo.builtin.build_systems.cmake import CMakePackage
 from spack_repo.builtin.build_systems.cuda import CudaPackage
 from spack_repo.builtin.build_systems.rocm import ROCmPackage
-from spack_repo.builtin.build_systems.cmake import CMakePackage
+
 from spack.package import *
 
 
--- a/repos/spack_repo/builtin/packages/proteus/package.py
+++ b/repos/spack_repo/builtin/packages/proteus/package.py
@@ -1,0 +1,131 @@
+# Copyright Spack Project Developers. See COPYRIGHT file for details.
+#
+# SPDX-License-Identifier: (Apache-2.0 OR MIT)
+
+from spack_repo.builtin.build_systems.cuda import CudaPackage
+from spack_repo.builtin.build_systems.rocm import ROCmPackage
+from spack_repo.builtin.build_systems.cmake import CMakePackage
+from spack.package import *
+
+
+class Proteus(CMakePackage, CudaPackage, ROCmPackage):
+    """
+    Proteus: A Programmable Just-In-Time (JIT) Compiler based on LLVM.
+    It embeds seamlessly into existing C++ codebases and accelerates
+    CUDA, HIP, and host-only C/C++ applications.
+    """
+
+    homepage = "https://github.com/Olympus-HPC/proteus"
+    url = "https://github.com/Olympus-HPC/proteus/archive/refs/tags/v2026.07.0.tar.gz"
+    git = "https://github.com/Olympus-HPC/proteus.git"
+
+    maintainers("ZwFink", "ggeorgakoudis")
+
+    license("Apache-2.0 WITH LLVM-exception")
+
+    version("main", branch="main")
+    version(
+        "2026.07.0",
+        sha256="4131ed4fc932784d3a5a978fd48475a92c2acc439cb3511eccd7d799016650f2",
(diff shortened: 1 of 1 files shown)


bernhardkaindl:
I noticed that the package does not build without `zstd`, and it also warns when `libedit` is not installed. [...] (rest removed for brevity as this is an obsolete discussion)

cmelone:
> I noticed that the package does not build without zstd, and it also warns when libedit is not installed

can you post the error messages? 

also, please refrain from changing the draft status of my PRs in the future

repos/spack_repo/builtin/packages/proteus/package.py
```diff
@@ -0,0 +1,132 @@
...
+    requires("%[virtuals=c,cxx] llvm-amdgpu", when="+rocm")
+    requires("%[virtuals=c,cxx] llvm", when="+cuda")
+
+    # CUDA and HIP dependencies.
+    depends_on("cuda@12:", when="+cuda")
+    depends_on("hip@6.2:", when="+rocm")
+
+    # MPI dependency.
+    depends_on("mpi", when="+mpi")
+
```
bernhardkaindl, line 92: @cmelone To trigger this error, you need to use a external `llvm+clang` package.

This also happens when you do `spack compiler find` and you've clang version 18.x, 19.x or 20.x installed:

When using a spack-internal `llvm+clang` build, `llvm` brings `zstd` as a dependency into the build but not the external llvm+clang:

Note: `%llvm@18` must not be built in this spack build already, so the external llvm+clang is used (you can also use llvm-19 or llvm-20 for this, whatever is not installed in spack)
```py
spack uninstall -y -a --dependents llvm+clang@18
```

Then prepare the external clang compiler and add it:
```py
$ sudo apt install -y clang-18 libclang-18-dev # required for libclangBasic.a
$ spack compiler find
==> Added 3 new compilers to /home/bkaindl/.spack/packages.yaml
    llvm@20.1.2  llvm@19.1.1  llvm@18.1.3
    $HOME/.spack/packages.yaml
```

Then, `zstd` is not pulled into the build by internal `llvm`:
```py
[x] bccdreb proteus@2026.07.0 failed
-- lines 35 to 47 --
  -- Found LLVM libraries: LLVM
  -- Found Threads: TRUE
  -- Configuring done (7.0s)
> CMake Error at /usr/lib/llvm-18/lib/cmake/llvm/LLVMExports.cmake:73 (set_target_properties):
  
      zstd::libzstd_shared
  
    but the target was not found.  Possible reasons include:
  
-- lines 53 to 67 --
    CMakeLists.txt:106 (include)
```
At this point, llvm@18 should be shown as an external package:
```
$ spack find -p llvm+clang@18
llvm@18.1.3  /usr
==> 1 installed package
```

At the same time, also LibEdit is warned as not found:

```py
-- Could NOT find LibEdit (missing: LibEdit_INCLUDE_DIRS LibEdit_LIBRARIES) 
-- Could NOT find zstd (missing: zstd_LIBRARY zstd_INCLUDE_DIR) 
```

```suggestion

    depends_on("zstd")
    depends_on("libedit") # Optional dependency for Proteus, for consistency of builds.
```

bernhardkaindl (commented):
@cmelone, here is the review.

repos/spack_repo/builtin/packages/proteus/package.py
cmelone: I think there is something wrong with your external clang/llvm setup. I've tested this with external llvm installations on two machines: one didn't have zstd support and the other did. Proteus compiles fine with an LLVM w/o LLVM_ENABLE_ZSTD.

bernhardkaindl: Thanks for taking the time to test it.

I actually don't think reproducing the exact failure is the important part here. This might be related to many facors. For example, a build host might have zstd installed somewhere or some other recipe in your specific build could pull it into the concretisation.

My error came from `CMake` `LLVMSupport` target which requires zstd, so this is not just affecting `proteus` but all packages that use `LLVMSupport`.

The external `LLVMSupport` is generic when clang is used as an external compiler, which adds its base LLVM as a external package.

cmelone: to sum up our discussion offline:

- proteus does not depend on zstd, it uses whatever compression algorithm was bundled with the provided LLVM installation
- any issues with externally built LLVM are outside the purview of the proteus spack package

bernhardkaindl (approved):
We agreed offline that I can take my finding to a packager discussion:
The issue I ran into is generic to how clang compilers are added as external llvm packages to spack with the `+clang` variant. We agreed I should discuss this on the overall packager discussion level as the issue is not specific to this package.

PR #5525, opened 2026-07-09
autotools: remove macos deployment target

alecbcs:
When macOS jumped from 10.16 to 11, we patched the autotools build system to hardcode the deployment target at 10.16, since many autotools scripts at the time assumed a 10.xx macOS version and broke otherwise.

This workaround is now instead causing build failures on macOS 26/27. The latest apple-clang@21 emits a warning whenever the deployment target is now set below 11 (since Apple Silicon is required for new macOS releases you'll no longer be able to compile for an SDK before M-series support). In several packages (like ruby) this warning crashes the test program in configure, causing the build to fail outright.

Since the original compatibility issue is now several years old, I removed the environment hardcode and rebuilt 100+ autotools-based packages (bash, coreutils, aspell, sed, etc.) to check for regressions. All of them built successfully.

As such, I think it's finally time we remove this from the autotools builder to fix compatibility with newer MacOS versions.

### DIFF
--- a/repos/spack_repo/builtin/build_systems/autotools.py
+++ b/repos/spack_repo/builtin/build_systems/autotools.py
@@ -8,7 +8,6 @@ from typing import Callable, List, Optional, Set, Tuple, Union
 
 from spack.package import (
     BuilderWithDefaults,
-    EnvironmentModifications,
     Executable,
     FileFilter,
     InstallError,
@@ -16,7 +15,6 @@ from spack.package import (
     PackageBase,
     Prefix,
     Spec,
-    Version,
     apply_macos_rpath_fixups,
     build_system,
     compiler_spec,
@@ -29,7 +27,6 @@ from spack.package import (
     force_remove,
     is_exe,
     keep_modification_time,
-    macos_version,
     mkdirp,
     register_builder,
     run_after,
--- a/repos/spack_repo/builtin/build_systems/autotools.py
+++ b/repos/spack_repo/builtin/build_systems/autotools.py
@@ -841,12 +841,6 @@ To resolve this problem, please try the following:
             with open(self._removed_la_files_log, mode="w", encoding="utf-8") as f:
                 f.write("\n".join(libtool_files))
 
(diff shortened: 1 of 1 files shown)


haampie:
Note that no rebuilds are triggered, also not when merged.

Probably good to do a full rebuild of the Darwin pipelines.

alecbcs:
All MacOS pipelines are currently disabled at UO. I don't know that we have a timeline yet for when those will come back, but I agree when we re-enable those we should perform a full rebuild.

I still think it's worth merging this in the mean time to fix packages like Ruby that currently fail to build on MacOS 26.

cmelone (approved):

haampie:
We could scan the relevant tarballs in the source mirror to find problematic configure files.

alecbcs:
Discussed with @haampie offline. We looked at scanning all versions across all autotools packages to determine which were built with libtool@:2.4.6. In testing, though, we found that packages built with these older libtool versions (e.g. aspell@0.60.8.2, berkeley-db@18.1.40) still build successfully on macOS without the compatibility layer.

Digging in further, Harmen found that this only becomes an issue if a package purposefully builds shared libraries with unresolved symbols (e.g. plugin architectures) and those packages may need the compatibility layer added back on a per-package basis. This significantly reduces the scope of affected packages.

As such, we've decided to go ahead and merge this PR as-is, and will add the compatibility layer back for individual packages if/when they turn out to need it.

haampie:
To be precise, if you get a failure like

```
$ cc -dynamiclib -o libplugin.dylib plugin.c
Undefined symbols for architecture arm64:
  "_host_counter", referenced from:
      _plugin_value in plugin-5dee83.o
ld: symbol(s) not found for architecture arm64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

you need ldflags like

```
$ cc -dynamiclib -Wl,-undefined -Wl,dynamic_lookup -o libplugin.dylib plugin.c
```

Setting `MACOSX_DEPLOYMENT_TARGET=10.X` was a workaround to get that in old configure scripts that did not anticipate a version 11.x

PR #5530, opened 2026-07-09
GauXC: add skala branch and install SKALA resource file

mkrack:
This updated recipe has been tested as a local spack recipe with CP2K

### DIFF
--- a/repos/spack_repo/builtin/packages/gauxc/package.py
+++ b/repos/spack_repo/builtin/packages/gauxc/package.py
@@ -5,24 +5,27 @@
 
 from spack_repo.builtin.build_systems.cmake import CMakePackage, generator
 from spack_repo.builtin.build_systems.cuda import CudaPackage
+from spack_repo.builtin.build_systems.rocm import ROCmPackage
 
 from spack.package import *
 
 
-class Gauxc(CMakePackage, CudaPackage):
+class Gauxc(CMakePackage, CudaPackage, ROCmPackage):
     """GauXC is a modern, modular C++ library for the evaluation of quantities related to
     the exchange-correlation (XC) energy (e.g. potential, etc) in the Gaussian basis set
-    discretization of Kohn-Sham density function theory (KS-DFT) on heterogenous architectures."""
+    discretization of Kohn-Sham density function theory (KS-DFT) on heterogenous
+    architectures."""
 
     homepage = "https://github.com/wavefunction91/GauXC"
     git = "https://github.com/wavefunction91/GauXC.git"
-    url = "https://github.com/wavefunction91/GauXC/archive/refs/tags/v1.1.tar.gz"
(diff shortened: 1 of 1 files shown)


mkrack:
@RMeli Please, could you review this PR.

repos/spack_repo/builtin/packages/gauxc/package.py
```diff
@@ -5,71 +5,130 @@
...
+    variant("cutlass", default=False, description="Build with CUTLASS linear algebra support")
+    variant("pic", default=True, description="Build position independent code")
+
+    conflicts("+magma", when="~cuda ~rocm", msg="MAGMA requires CUDA or HIP")
+    conflicts("+nccl", when="~cuda", msg="NCCL requires CUDA")
+    conflicts("+cutlass", when="~cuda", msg="CUTLASS requires CUDA")
+    conflicts("+fortran", when="~c @1.2.dev2:", msg="Fortran bindings require C API")
+
+    # Skala resource file
+    skala_version = "1.1"
```
RMeli, line 60: Should this be a multi-valued variant (only available `when="+skala"`) with the possible versions available? I'm not yet familiar with this package, but hard-coding this doesn't seem ideal.

awvwgk: I would prefer to not include any checkpoint files in the spack package of GauXC, but refer the user to the HuggingFace repository for downloading the necessary files. Including the checkpoint here would make the spack distribution of GauXC different from other distributions which come without the checkpoint.

mkrack: Thank you very much for reviewing.
@awvwgk I understand your concerns regarding the inclusion of a specific resource file in the recipe. I will close this request.

RMeli: @awvwgk I understand your concern, but what about allowing both possibilities? Something like the following:
```python
    allowed_skala_parameters = ["none", "1.1"]

    variant("skala_parameters", default="none", values=allowed_skala_parameters,  multi=False)

    # Skala resource file
    if not self.spec.satisfies("skala_parameters=none")
      skala_version = self.spec.variants["skala_parameters"].value
      skala_file = f"skala-{skala_version}.fun"
      resource(
          name="skala_fun",
          url=f"https://huggingface.co/microsoft/skala-{skala_version}/resolve/main/{skala_file}",
          sha256="0c8432ac3f03c8f1276372df9aca5b7ee7f8939d47a8789eb158976e89aa0606",
          expand=False,
          placement=skala_file,
          when="+skala",
      )
```

So by default checkpoints are not included. But the `skala_parameters` variant still allow to download them automatically and place them in the right place for convenience. What do you think?

awvwgk: I would prefer that we implement this as part of the upstream distribution, rather than in each package manager individually.

PR #5532, opened 2026-07-09
labels: automatically tag build-systems, fix maintainers

alecbcs:
- automatically tag build system changes with existing `build-systems` label
- automatically tag license changes with existing `licenses` label
- fix maintainers regex to look for `maintainers()` directive and not list
- make labels that are not `new-*` also trigger on deleted lines


kwryankrattiger (approved):
LGTM

PR #5534, opened 2026-07-09
Python: Correct impossible virtual

psakievich:
`gettext` depends on `iconv` and therefore can't be a provider of `iconv`.

The providers of `iconv` are `libiconv`, `musl` and `glibc`. Of these, `libiconv` is the only one where we want to require `gettext`. `gettext` (provides `libintl`) is an optional dependency of `python` and python is configured to build without it.
If `gettext` is detected then `python` will use it, otherwise it builds without it.

Links:
https://github.com/python/cpython/issues/61428
https://github.com/astral-sh/python-build-standalone/blob/main/docs/technotes.rst#gettext--locale-module

### DIFF
--- a/repos/spack_repo/builtin/packages/python/package.py
+++ b/repos/spack_repo/builtin/packages/python/package.py
@@ -220,7 +220,7 @@ class Python(Package):
         depends_on("pkgconfig", type="build")
         depends_on("gettext +libxml2", when="+libxml2")
         depends_on("iconv", when="~libxml2")
-        depends_on("gettext ~libxml2", when="~libxml2 ^[virtuals=iconv]gettext")
+        depends_on("gettext ~libxml2", when="~libxml2 ^[virtuals=iconv] libiconv")
 
         # Optional dependencies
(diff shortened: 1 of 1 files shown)


adamjstewart:
@alalazo this is getting into the weeds of the virtual syntax you designed, does this make more sense now?

psakievich:
I'm testing locally to make sure it is properly enforced such that I can still concretize without depending on `gettext`.

psakievich:
Okay confirmed if I have
```yaml
specs:
- python ^[virtuals=iconv] glibc
packages:
  python:
    require: 
    - ~libxm2
```

I can get the `gettext` dependency to drop.

Interestingly

```yaml
specs:
- python 
packages:
  python:
    iconv:
      prefer:
      - libiconv
    require: 
    - ~libxm2
    - ^[virtuals=iconv] glibc
```

Doesn't pick up the requirement correctly in the environment I'm working in. That's a tangent though.

alecbcs (approved):
Makes sense and looks good to me

PR #5513, opened 2026-07-08
py-plaid: add v1.0.0

casenave:
add [plaid v1.0.0](https://github.com/PLAID-lib/plaid/blob/main/CHANGELOG.md)

Depends on:
- #3981
  - #3502

PR #5539, opened 2026-07-10
py-trame: add v3.10.2 -> v3.9.1

casenave:
Required by [#5513](https://github.com/spack/spack-packages/pull/5513)


johnwparent (commented):
Do we have a need for all these non latest patch versions? If not, lets just add the latest patch versions for each minor version series.

casenave:
> Do we have a need for all these non latest patch versions? If not, lets just add the latest patch versions for each minor version series.

Sure, I removed the non latest patch versions.

johnwparent (approved):
Looks great, thanks!

bernhardkaindl (approved):

PR #5544, opened 2026-07-10
uwtools: add v2.10.0, v2.11.0, v2.9.1

eap:
These package version updates are needed for the spack-stack python update work noted in this issue https://github.com/JCSDA/spack-stack/issues/2005


repos/spack_repo/builtin/packages/uwtools/package.py
```diff
@@ -24,24 +24,29 @@ class Uwtools(PythonPackage):
...
-    depends_on("python@3.9:3.11")
+    depends_on("python@3.10:3.14", when="@2.11:")
+    depends_on("python@3.9:3.13", when="@2.9:2.10")
+    depends_on("python@3.9:3.11", when="@:2.8")
     depends_on("py-setuptools", type="build")
     depends_on("py-f90nml@1.4")
     depends_on("py-jinja2@3.1")
     depends_on("iotaa@1.1", when="@2.6")
     depends_on("iotaa@1.2", when="@2.7")
     depends_on("iotaa@1.3:2.0", when="@2.8:")
```
RikkiButler20:
```suggestion
    depends_on("iotaa@1.3:1.0", when="@2.8:")
```
Could you fix the the upper bounds? Everything else looks great.

RikkiButler20 (changes requested):
@maddenp-cu

repos/spack_repo/builtin/packages/uwtools/package.py
maddenp-cu: Hi @RikkiButler20 -- I don't (yet) understand what this means and how I can assist. If I'm reading it right, the old line says that `uwtools` version 2.8 requires `iotaa` between versions 1.3 and 2.0. Is that right? That agrees with [the version pin](https://github.com/ufs-community/uwtools/blob/2.8.2/recipe/meta.yaml#L20C7-L20C21) for that version of `uwtools`. The suggested change above looks like it would be an empty range -- greater than 1.3 but less than 1.0? I'm not very familiar with Spack, though, so I could be misunderstanding.

Is it the the case that `iotaa` needs a lower bound than `2.0`, though? That seems like it could be possible. There could have been breaking changes >1.3 but <2.0. I think this could be strictly `1.3` (no upper bound specified) and be fine if that's what's happening. Does that help?

RikkiButler20: I understand the confusion. You are right about the versions required of `iotaa` be between 1.3 and 2.0, but its the version notation that is wrong.
The number after the colon (:1) specifies the upper major version boundary (exclusive). So @1.3:2 actually means ">=1.3 and <=2.0", not "greater than 1.3 but less than 2.0". Does this make sense?

Here's a link to the documentation on specifying version constraints in packages.
[Package Guide: Version Constraints](https://spack.readthedocs.io/en/latest/packaging_guide_creation.html#specifying-version-constraints)

maddenp-cu: I think I just realized that your request was directed to @eap, who opened the PR. Sorry for the confusion.

I am a maintainer of both `uwtools` and its `iotaa` dependency, and so am happy to advise: I think specifying `1.3` as the sole `iotaa` version (no range, no upper bound) for `uwtools` `2.8` would be fine.

If there's anything else I can do to help, please let me know.

eap: Yes the inclusive/exclusive version spec of "2.0" is what I was trying to fix here. Although looking at iotaa we there isn't even have a 2.0 release yet so the entire concern is unnecessary. I'm taking the suggestion offered above to use 1.3 strictly since, as I understand it, spack translates this to `>=1.3`.

```suggestion
    depends_on("iotaa@1.3", when="@2.8:")
```

Apologies for my tardiness here, I opened this PR before a vacation!

RikkiButler20:
Hey @eap, your changes got pulled into #5631 since it was adding a new version to `uwtools` too.

I'm going to close this one out to keep things clean, but could you jump over to that other PR and give it a quick review to make sure everything looks good to you? Thanks for the solid work.

maddenp-cu (approved):

PR #5546, opened 2026-07-10
libarchive: add CMake builder and Windows support

johnwparent:
Adds CMake builder (and via that, Windows support) to libarchive. 

Adds a couple windows specific variant options as well

### DIFF
--- a/repos/spack_repo/builtin/packages/libarchive/package.py
+++ b/repos/spack_repo/builtin/packages/libarchive/package.py
@@ -105,7 +105,11 @@ class AutotoolsBuilder(autotools.AutotoolsBuilder):
 
         # Handle crypto explicitly: "cng" is Windows/CMake-only and unknown to configure
         for opt in ("mbedtls", "nettle", "openssl"):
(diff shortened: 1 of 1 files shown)


johnwparent:
@spackbot fix style

haampie:
Is there a way to build CMake from sources on Windows? This is the so manieth CMake dependency that now depends on CMake.

johnwparent:
> Is there a way to build CMake from sources on Windows? This is the so manieth CMake dependency that now depends on CMake.

Unfortunately for packaging sanity, a lot of CMake's dependencies only have cmake systems on Windows. 
Ignoring curl for a second (I'm working on this now), I think it should, IIUC, `cmake+ownlibs` breaks the cycle for libarchive and similar packages (notably, not curl) so the solver will resolve to using that variant, right? (I tested this locally a while back and this was the case then, please correct me if that's changed or I was wrong initially).

I'm working on adding a break in that cycle for curl, as without either an external curl or external cmake, no cmake will not build from source on Windows via spack right now, which is quite bad.

PR #5549, opened 2026-07-10
py-mistune: add v3.3.3, deprecate CVEs

adamjstewart:


alecbcs (approved):
Looks good to me. Thanks @adamjstewart

PR #5550, opened 2026-07-11
geant4: add version 11.4.2

plexoos:
This PR adds the Geant4 11.4.2 release to the package recipe and updates the source checksum accordingly. No dependency
or variant changes were needed for this bump.


sethrj (approved):
Checksum is valid. Thanks!

PR #5554, opened 2026-07-11
llvm: require usable libclang before detecting external +clang

### LINKED ISSUE #5553: LLVM compiler discovery can register external `llvm+clang` even when `libclang` is not build-usable
## Bug: LLVM compiler discovery can register external `llvm+clang` even when `libclang` is not build-usable

- Fixed by #5561 
- Slack discussion: https://spackpm.slack.com/archives/D0BG6A6H4D7/p1783765910667869 (#packaging)
### Summary

`spack compiler find` can auto-register a system LLVM installation as an external `llvm+clang` package based on the presence of a detected clang compiler, even when the host installation cannot actually compile/link dependent packages against `libclang`.

This can cause packages with `depends_on("llvm+clang")` to select the auto-detected external LLVM and fail later during CMake configuration or linking.

### Description

When Spack discovers compilers, it may also register external packages associated with those compilers. For clang, Spack can add an external `llvm` package with compiler metadata under `extra_attributes.compilers`.

The detected external LLVM spec may also include variants such as `+clang`.

However, `+clang` is used by packages that need the clang/libclang development interface, not just a working clang executable. On Linux distributions, it is common for the compiler executable and some LLVM runtime libraries to be installed without all development files or transitive development dependencies required to build new software against `libclang`.

For example, a system may have:

- a working `clang-18` executable
- LLVM and clang development packages installed
- `libclang` libraries present
- missing transitive development dependencies needed by LLVM's exported CMake targets, such as `libzstd-dev`

In that situation, Spack may still register the external LLVM as `llvm@18+clang`, causing dependent packages to select it. The failure is then deferred until the dependent package tries to use LLVM/Clang through CMake.

### Expected behavior

Spack should only auto-register an external LLVM as `+clang` when the detected installation is usable by dependent packages that require the clang/libclang development interface.

If `libclang` cannot be compiled, linked, and used successfully, the external LLVM should either not be marked `+clang`, or should otherwise not satisfy `depends_on("llvm+clang")`.

### Actual behavior

Spack registers the system LLVM as external `llvm@18+clang` after detecting the system clang compiler, even though packages that use LLVM/Clang through CMake fail because a transitive dependency target is missing.

### Reproducer

Precondition: no existing Spack-built `llvm@18+clang` should satisfy the dependency, so the auto-detected external LLVM is selected.

```console
$ spack uninstall -y -a --dependents llvm+clang@18
$ grep -A5 llvm ~/.spack/packages.yaml
# no llvm external present
```

Install a system clang compiler and clang development package, but remove the zstd development package:

```console
$ sudo apt install -y clang-18
$ sudo apt remove -y libzstd-dev
$ sudo apt install -y libclang-18-dev
```

Run compiler discovery:

```console
$ spack compiler find
==> Added 1 new compiler to /home/bkaindl/.spack/packages.yaml
    llvm@18.1.3
    $HOME/.spack/packages.yaml
```

Spack now shows the external LLVM as satisfying `llvm+clang@18`:

```console
$ spack find -p llvm+clang@18
llvm@18.1.3  /usr
==> 1 installed package
```

The generated external package entry includes `+clang`:

```yaml
llvm:
  externals:
  - spec: llvm@18.1.3+clang~flang~lld~lldb
    prefix: /usr
    extra_attributes:
      compilers:
        c: /usr/bin/clang
        cxx: /usr/bin/clang++
```

Now install a package that depends on `llvm+clang`:

```console
```

The build fails during CMake configuration:

```console
-- Found LLVM libraries: LLVM
-- Found Threads: TRUE
-- Configuring done

CMake Error at /usr/lib/llvm-18/lib/cmake/llvm/LLVMExports.cmake:73 (set_target_properties):

    zstd::libzstd_shared

  but the target was not found.
```

Relevant backtrace:

```console
CMakeLists.txt:106 (include)
```

### Why this matters

The external package was added automatically by compiler discovery, without the user explicitly declaring that `/usr` provides a usable `llvm+clang` development installation.

Because the external is marked `+clang`, Spack can choose it for packages that require `llvm+clang`. The failure then appears later in an unrelated package build, even though the root cause is that the auto-detected external LLVM does not actually provide a complete build-usable clang/libclang development interface.

### Possible fixes

The "clang" variant likely should not be changed, at least it is not easy: llvm.spec providing `"c"` and `"cxx"` compilers depends on the `"+clang"` variant, and this is apparently not easy to change, and `+clang`/`~clang` would change for existing hosts if we change the detection.
- I tried this in #5554 and changes for the spack solver, which failed so far.

Instead, llvm could add a new `+libclang` variant which determines if `llvm` provides `libclang` development functionality and for external packages, before setting it to `+libclang`, verify that the detected installation can compile, link, and run a minimal program using `libclang`, using include and library paths from the paired `llvm-config`
- Implemented by #5561

bernhardkaindl:
- Fixes #5553 
- Slack discussion: https://spackpm.slack.com/archives/D0BG6A6H4D7/p1783765910667869 (#packaging)
## Summary
This changes external `llvm+clang` detection during compiler discovery to be capability-based.

When Spack discovers a system clang compiler and auto-registers an external LLVM package, it now advertises `+clang` only if the detected installation can compile, link, and run a minimal program using `libclang`.

This aligns with the intended meaning of `+clang` for dependent packages that need the libclang development interface, rather than the clang executable, which is a separate concern from the libclang development interface.

## Motivation

In practice, many common distro installations provide working `clang` executable, but unrelated to the compiler, using `libclang` may fail due to:

- missing development headers,
- missing link-time dependencies,
- missing transitive dependencies such as `zstd`,

In those cases, Spack would have still selected the external LLVM for `depends_on("llvm+clang")`, only for the dependent package to fail later during build or link.

## Implementation

To address this issue in external LLVM detection, Spack now:

1. locates the paired `llvm-config`
2. reads include and library paths from it
3. compiles a minimal `clang-c` test program against `libclang`
4. links with the dependency set needed by this LLVM configuration, including the zstd library, to match CMake LLVMConfig behavior.
5. executes the test binary and checks that it succeeds

If any step fails, the external LLVM is registered without `+clang`.

## Behavior Change

Before:

- `+clang` could be inferred from clang executable presence alone

After:

- `+clang` is advertised only when `libclang` is usable end to end

This keeps automatic compiler discovery intact while avoiding false-positive external `llvm+clang` registrations.

## Scope

This is limited to clang-related variant detection for external LLVM registration. Non-clang variant detection behavior is unchanged.

## Testing

Tested on a host where compiler discovery auto-adds an external LLVM from system clang. The previous executable-based detection was insufficient on that host, while the probe-based detection correctly withheld `+clang` unless `libclang` was build-usable.


bernhardkaindl:
Archiving this PR, I think it is better to pivot this to add +libclang instead and change the few packages that compile against libclang deend on `llvm+libclang`.

bernhardkaindl:
The GitLab Pipeline was not cancelled when closing this PR, hoping to cancel it by draft status?

PR #5062, opened 2026-06-02
reframe: add v4.10.0

jscook2345:
https://github.com/reframe-hpc/reframe/releases/tag/v4.10.0

PR #5556, opened 2026-07-11
reframe: Fix dependencies, add v1.10.0, v1.10.1, add analytics variant

### LINKED ISSUE #5409: reframe: missing dependencies & missing versions
I'd like to report some issue with `reframe` and its dependencies.

### Missing dependencies

When using reframe installed with spack, I get `No module named <pkg>` errors for `ClusterShell` and `fasteners`. Reproduce with: `from reframe.utility.typecheck import Duration`.

Indeed, the two are not listed among the dependencies in the recipe here, while they are indeed in the GitHub requirements, see [those](https://github.com/reframe-hpc/reframe/blob/v4.9.0/requirements.txt) for v4.9.0.

I think one should add something like:
```python
    depends_on("py-clustershell", type="run")
    depends_on("py-fasteners", when="@4.9.0:", type="run")
```

### `py-clustershell` and recent python
That actually led me to another issue with `ClusterShell`. Indeed, its latest release, 1.9.3, builds with `distutils`, which is deprecated in python 3.12. I was forced to used develop version. I had to add to `py_clustershell/package.py`:
```python
    git = "https://github.com/clustershell/clustershell.git"

    conflicts("@:1.9", when="%python@3.12:", msg="Old versions rely on distutils, deprecated in python 3.12")
```

### Sanity checks with recent releases of `reframe`
Finally, going back to reframe, new releases have been published, 4.10.x. I tried a couple of days ago with 4.10.0 and installation failed because they seem to have changed a bit the structure of the tar and the [sanity checks](https://github.com/spack/spack-packages/blob/develop/repos/spack_repo/builtin/packages/reframe/package.py#L165) fail; indeed, `sanity_check_is_file = ["bin/reframe"]` fails because there exist no folder bin any more.

### Maintainers
@victorusu @vkarak

vkarak:
Closes #5409.
Replaces #5062.

### DIFF
--- a/repos/spack_repo/builtin/packages/reframe/package.py
+++ b/repos/spack_repo/builtin/packages/reframe/package.py
@@ -27,6 +27,8 @@ class Reframe(Package):
     license("BSD-3-Clause")
 
     version("develop", branch="develop")
+    version("4.10.1", sha256="6af76c3dd2e100d18cbfb45721dd69115f54643e493f58707953841a1f0b51b3")
+    version("4.10.0", sha256="3fddf2ab8642a11d35d878ecf07a8912ee4d21f4ab37db8e53ca6c130b1819de")
     version("4.9.3", sha256="44efde35528a36bec330926d769dbdb9277ed46c98a40a4f914bcbe5132f60ff")
     version("4.9.2", sha256="a224434033cf94e4c74654fdc6204c9a8d5a0575060addc47ec1de5242148fdf")
     version("4.9.1", sha256="82efaf74218d7f518a13393c066502ff61dfb713d040b33298fb01c3cd5e4182")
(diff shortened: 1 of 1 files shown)


bernhardkaindl:
Hi @vkarak, I see that @RMeli also reported this issue in #5409:

----

### `py-clustershell` and recent python
That actually led me to another issue with `ClusterShell`. Indeed, its latest release, 1.9.3, builds with `distutils`, which is deprecated in python 3.12. I was forced to used develop version. I had to add to `py_clustershell/package.py`:
```python
    git = "https://github.com/clustershell/clustershell.git"

    conflicts("@:1.9", when="%python@3.12:", msg="Old versions rely on distutils, deprecated in python 3.12")
```
---

Yep, due to `@:1.9` still using `distutils`, clustershell needs a conflics for with `^python@3.12:`, it does not build with Python 3.12 or newer:

- The newly released v1.10 supports newer Python versions.
- BTW, distutils was deprecated sincy Python 3.10 and completely removed with Python 3.12, so the conflict msg can be shortened.
- I opened #5558 to fix this dependency, can you review it?

bernhardkaindl (approved):

PR #5558, opened 2026-07-11
py-clustershell: add v1.10 to support Python 3.12 and newer

bernhardkaindl:
Together with #5556 for `reframe`, this fixes the remaining open bugs in #5409:

  - As reported in #5409, the existing older version os py-clustershell still use `distutils`.
  - The new version fixes that.
  - The URL of the homepage has changed, so while at it fix it too.
  - The file `url` has also been updated accordingly.
  
Tested with a recent master version (likely very similar) by @RMeli in #5409.

To: @vkallesh, related to your PR #5556 (py-clustershell is a dependency of `reframe`)
Cc: @jscook2345: JFYI - feel free to review, test, add a non-maintainer approval

### DIFF
--- a/repos/spack_repo/builtin/packages/py_clustershell/package.py
+++ b/repos/spack_repo/builtin/packages/py_clustershell/package.py
@@ -12,17 +12,20 @@ class PyClustershell(PythonPackage):
     node groups and execute commands on cluster nodes in parallel.
     """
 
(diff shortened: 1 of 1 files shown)


vkarak (approved):
Lgtm.

RMeli (approved):
Not sure if the tag to me was correct, but LGTM. =)

PR #5561, opened 2026-07-12
RFC on llvm: Add libclang variant with external detection

### LINKED ISSUE #5553: LLVM compiler discovery can register external `llvm+clang` even when `libclang` is not build-usable

bernhardkaindl:
Add a libclang variant and detect whether an external LLVM installation provides usable libclang development support during compiler discovery.

When Spack discovers a system clang compiler or registers an external LLVM package, detection uses llvm-config to query include, library, and binary paths, then compiles, links, and runs a minimal clang-c program against libclang.

This lets detected externals distinguish between a usable clang compiler and a usable libclang development installation. Common distro installs can provide clang while still lacking libclang headers, link-time libraries, or transitive development dependencies such as zstd.

In those cases the external is still detected, but with ~libclang so dependents that need libclang can fall back to a Spack-built LLVM.

The detection tests model both positive and negative libclang probe behavior. Positive fixtures provide llvm-config scripts with matching --version, include, library, and bindir responses, and mock clang executables create runnable probe binaries.

Negative fixtures cover explicit ~libclang detection for:

- a generated probe executable that fails when run,
- a mock clang compile step that fails,
- a derived clang executable that does not exist,
- an llvm-config --bindir result that points to a missing directory, and
- an llvm-config failure while querying probe paths.

Also updated the filter for Executables like lldb-vscode-X: Only apply it to the file names not the full path to avoid filtering out directories with these names which would break some environments.

Tested with `spack -d audit externals`

### DIFF
--- a/repos/spack_repo/builtin/packages/llvm/package.py
+++ b/repos/spack_repo/builtin/packages/llvm/package.py
@@ -3,6 +3,7 @@
 # SPDX-License-Identifier: (Apache-2.0 OR MIT)
 import os
 import re
+import tempfile
 
 from spack_repo.builtin.build_systems.cmake import CMakePackage, generator
 from spack_repo.builtin.build_systems.compiler import CompilerPackage
@@ -22,12 +23,14 @@ class LlvmDetection(PackageBase):
     def filter_detected_exes(cls, prefix, exes_in_prefix):
         # Executables like lldb-vscode-X are daemon listening on some port and would hang Spack
         # during detection. clang-cl, clang-cpp, etc. are dev tools that we don't need to test
+        # and can be filtered out. Only apply it to the file names not the full path to avoid
+        # filtering out directories with these names which would break some environments.
         reject = re.compile(
             r"-(vscode|cpp|cl|ocl|gpu|tidy|rename|scan-deps|format|refactor|offload|"
             r"check|query|doc|move|extdef|apply|reorder|change-namespace|"
             r"include-fixer|import-test|dap|server|PerfectShuffle)"
         )
-        return [x for x in exes_in_prefix if not reject.search(x)]
+        return [x for x in exes_in_prefix if not reject.search(os.path.basename(x))]
 
 
 class Llvm(CMakePackage, CudaPackage, LlvmDetection, CompilerPackage):
@@ -106,6 +109,8 @@ class Llvm(CMakePackage, CudaPackage, LlvmDetection, CompilerPackage):
     variant(
         "clang", default=True, description="Build the LLVM C/C++/Objective-C compiler frontend"
     )
+    variant("libclang", default=True, description="Provide the libclang development interface")
+    conflicts("+libclang", when="~clang")
 
     variant("flang", default=False, description="Build the LLVM Fortran compiler frontend ")
 
@@ -709,13 +714,19 @@ class Llvm(CMakePackage, CudaPackage, LlvmDetection, CompilerPackage):
 
     @classproperty
     def executables(cls):
-        return super().executables + [r"^ld\.lld(-\d+)?$", r"^lldb(-\d+)?$"]
+        return super().executables + [
+            r"^ld\.lld(-\d+)?$",
+            r"^lldb(-\d+)?$",
+            r"^llvm-config(-\d+(\.\d+)*)?$",
+        ]
 
     @classmethod
     def determine_version(cls, exe):
         try:
             compiler = Executable(exe)
             output = compiler(cls.compiler_version_argument, output=str, error=str)
+            if compiler.name.startswith("llvm-config"):
+                return output.strip()
             if "Apple" in output:
                 return None
             if "AMD" in output:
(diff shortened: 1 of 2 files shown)


rbberger:
Should there be a conditional for `provides("libllvm")`?

bernhardkaindl:
> Should there be a conditional for provides("libllvm")?

I think this can only be possible if a conditional `provides` could be in `packages.yaml`?

Nitpick: `libLLVM` this is "just" LLVM, no C frontend) is different from `libclang` (the C frontend of LLVM)

For example, it would have to be something that can be expressed  in `~/.spack/packages.yaml` and be depended on using `depends_on("llvm@18:20 +libclang")` or `depends_on("libclang@18:20")`. I don't think adding several conditional provides to `~/.spack/packages.yaml` is a good idea, variants I think are good for that:

```
head ~/.spack/packages.yaml
packages:
  llvm:
    externals:
    - spec: llvm@18.1.3+clang~flang~libclang~lld~lldb
      prefix: /usr
      extra_attributes:
        compilers:
          c: /usr/bin/clang
          cxx: /usr/bin/clang++    
    - spec: llvm@20.1.1+clang~flang+libclang~lld~lldb
      prefix: /usr
      extra_attributes:
        compilers:
          c: /usr/bin/clang
          cxx: /usr/bin/clang++

File: .ci/gitlab/.gitlab-ci.yml [hunk 1]
```diff
@@ -116,7 +116,7 @@ default:
     - mkdir "${CI_PROJECT_DIR}/jobs_scratch_dir/${SPACK_CI_STACK_NAME}"
     - spack
       config blame > "${CI_PROJECT_DIR}/jobs_scratch_dir/${SPACK_CI_STACK_NAME}/spack.yaml.blame"
-    - spack -v  --color=always
+    - spack -d -v  --color=always
```
rbberger: Should this `-d` stay?

bernhardkaindl: @rbberger, thanks, I added this commit to get verbose logs, but it does not belong here. Removed.

bernhardkaindl:
Note / TODO: I should rename the new `libclang` variant I propose for LLVM to `libclang-dev` because it is about the development library support of the package, not the libclang itself (which is always present, it is where clang lives). This also avoids conflicting variants:
- #5856

repos/spack_repo/builtin/packages/llvm/package.py
```diff
@@ -1197,6 +1252,13 @@ def post_install(self):
         spec = self.spec
         define = self.define
 
+        if spec.satisfies("~libclang"):
+            clang_c_headers = join_path(self.prefix.include, "clang-c")
+            if os.path.exists(clang_c_headers):
+                remove_linked_tree(clang_c_headers)
+            if os.path.isdir(self.prefix.lib):
+                force_remove(*find(self.prefix.lib, "libclang.*", recursive=False))
+
```
sethrj, line 1261: This seems drastic; if post-install detects these it should probably be an error message. It seems to me that leaving these files in place is less likely to cause issues than removing them aggressively.

sethrj (commented):
Was this PR generated by claude/AI?

repos/spack_repo/builtin/packages/llvm/package.py
bernhardkaindl: This is only there because I expect that LLVM/Clang might have no option to disable installing development headers and libraries.

The issue here is that Linux distributions split LLVM and Clang packages into runtime/compiler and development packages, and spack detects "llvm" based on the compiler programs existing, not if the development libraries are provided by an external LLVM installation. This clashes when packages request "llvm" and expect the LLVM installation to have the development libraries like they normally exist in a spack installation of LLVM, which causes the build of such packages to fail if the llvm/clang development packages aren't installed for the external llvm/clang packages imported from the host.

For consistency with external packages without libclang-development headers and libraries, when the libclang variant is actively disabled, this would aims to align the build what is provided by an external package detected as `~libclang`. Specifying `~libclang` could be useful when an environment wants to provide a specific llvm package build (but other builds in parallel it), but do not want not provide its using its libclang to link against, in order to prevent other packages that link against libclang to become linked to this specific version. The only way to prevent this for sure is to also not install the libclang development library for the ~libclang build.

I think this should explain that there is nothing drastic and aggressive there, just consistency. Probably, I should add a comment about that above this code. I might also have to refine it a bit more to not delete the actual libclang.so.x.x library used as clang runtime of the clang compiler itself and add a trivial smoke tests after the install to ensure the clang compiler is working.

I'll convert the PR to draft status to prevent merge until this check is in place and add an `RFC:` prefix to encourage further comments and discussion.

I'm also pondering of the variant should be named `libclang-dev` communicate in the name that this variant about providing the libclang development library, because there is another PR that also adds a libclang variant to a different package, not LLVM.

sethrj: I understand the need for a variant (for externally installed versions at least). But how do you know, e.g., whether no future version of the `clang` executable will link against `libclang*`?  Are you sure your approach will work if someone does, e.g.,  `~libclang +link_llvm_dylib`?

I think it would be consistent to *prevent* a `~libclang` variant if llvm prevents you from disabling it inline: raising InstallError after or something.

kwryankrattiger:
---

In general I am not a huge fan of just layering on more variants to solve packaging problems, I would prefer that the packages represent detectable/relational constructs on their own and only use variants to toggle features/extensions that can actually be influenced by Spack during the build. And even then, I think there is a lot to be said about Spack needing to have more of an opinion on what flags are grouped behind a single variant vs mapping every build option to a variant.

---

There has been discussion in the past to have llvm not provide compilers or tools directly, and have meta packages for clang/flang/lld/etc. that depend on llvm with the require variants.

That would also solve this issue, so we could detect `clang` which implies nothing about an llvm or libllvm or development environment.

```python
class Clang(Package):

    depends_on("llvm +clang", type=("build", "run"))
    provides("c", "cxx")

    def install(self):
            # Symlink/copy clang components from llvm 
```

PR #5567, opened 2026-07-13
geomodel: add v6.28.0, v6.29.0

stephenswat:
This commit adds versions 6.28.0 and 6.29.0 of the GeoModel package.


alecbcs (approved):
Looks good to me. Thanks @stephenswat 

Will wait a few days for @wdconinc to do a maintainer review. Feel free to ping me if you don't hear back.

wdconinc (approved):
LGTM. Only thing that caught my eye in the release notes ("Added an explicit dependency on XercesC to GeoModelXml") is already a dependency here.

PR #5568, opened 2026-07-13
libgpiod: add v2.2.4-v2.3.1, support meson build, fix snapshot urls

davekeeshan:
## Summary
- libgpiod switched from autotools to meson starting with the 2.3 release
  series. The package now selects the build system per version via
  `build_system()` and provides both an `AutotoolsBuilder` (`@:2.2`) and a
  `MesonBuilder` (`@2.3:`, default).
- Add versions 2.2.4, 2.2.5, 2.3.0 and 2.3.1.
- Fix `url_for_version` for `x.y.0` releases: kernel.org tags these without the
  trailing ".0" (e.g. `v2.3`), so use `up_to(2)` instead of `up_to(1)`, which
  previously generated a non-existent `libgpiod-v2` URL.
- Refresh the sha256 checksums for older cgit snapshots (`<= 2.2.0` and `1.6.x`)
  that no longer matched the tarballs currently served by kernel.org.

All 18 versions from 1.6.3 through 2.3.1 had their recorded sha256 verified
against the official kernel.org release tarballs. Builds were validated across
both paths — meson for 2.3+ (2.3.1, 2.3.0) and autotools for <= 2.2 (2.2.5,
2.0.0, 1.6.3).

### DIFF
--- a/repos/spack_repo/builtin/packages/libgpiod/package.py
+++ b/repos/spack_repo/builtin/packages/libgpiod/package.py
@@ -3,12 +3,14 @@
 # SPDX-License-Identifier: (Apache-2.0 OR MIT)
 
 
+from spack_repo.builtin.build_systems import autotools, meson
 from spack_repo.builtin.build_systems.autotools import AutotoolsPackage
+from spack_repo.builtin.build_systems.meson import MesonPackage
 
 from spack.package import *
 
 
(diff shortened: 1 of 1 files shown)


bernhardkaindl (approved):
Built all versions successfuly with:
```py
gh extension install spack/gh-spack-pr;gh spack-pr run 5568
```

PR #5579, opened 2026-07-14
snls: new package

dtaller:
snls package, https://github.com/llnl/SNLS .
Although I am making this pull request, several LLNL employees contributed to this recipe, including blake and guy bergel. Thanks for your help!

### DIFF
--- a/repos/spack_repo/builtin/packages/snls/package.py
+++ b/repos/spack_repo/builtin/packages/snls/package.py
@@ -28,6 +28,17 @@ class Snls(CMakePackage, CudaPackage, ROCmPackage):
     version("v0.4.0", sha256="dfedf72272ef8bbf41d62ae2e8f2b1e48653b1ec364a19b6a832e2dd2c19d7ea", url="https://github.com/llnl/SNLS/archive/refs/tags/v0.4.0.tar.gz")
 
     variant("shared", default=True, description="build shared libs")
+    variant("tests", default=False, description="Build with tests enabled")
+    variant(
+        "cxxstd",
+        default="20",
+        values=("11", "14", "17", "20"),
+        description="C++ standard to build with",
+    )
+
+    @property
+    def cxx_std(self):
+        return self.spec.variants.get("cxxstd").value
 
     depends_on("blt", type=("build"))
     depends_on("c", type=("build"))
@@ -44,10 +55,11 @@ class Snls(CMakePackage, CudaPackage, ROCmPackage):
         """
         spec = self.spec
         args = [
-            self.define("ENABLE_TESTS", False),
-            self.define("ENABLE_GTEST", False),
+            self.define_from_variant("ENABLE_TESTS", "tests"),
+            self.define_from_variant("ENABLE_GTEST", "tests"),
             self.define("USE_BATCH_SOLVERS", True),
             self.define_from_variant("BUILD_SHARED_LIBS", "shared"),
+            self.define("BLT_CXX_STD", f"c++{self.cxx_std}"),
             self.define("BLT_SOURCE_DIR", spec["blt"].prefix),
             self.define("RAJA_DIR", spec["raja"].prefix.lib.cmake.raja),
             self.define("CAMP_DIR", spec["camp"].prefix.lib.cmake.camp),
--- a/repos/spack_repo/builtin/packages/snls/package.py
+++ b/repos/spack_repo/builtin/packages/snls/package.py
@@ -9,14 +9,14 @@ from spack_repo.builtin.build_systems.rocm import ROCmPackage
 from spack.package import *
 
 
-class Snls(CMakePackage, CudaPackage, ROCmPackage):
-    """Small solver for Nonlinear systems of equations"""
+class Snls(CMakePackage):
+    """Small Non-Linear Solver for solving non-linear systems of equations"""
 
     homepage = "https://github.com/llnl/SNLS"
     url = "https://github.com/llnl/SNLS/archive/refs/tags/v0.4.4.tar.gz"
     git = "https://github.com/llnl/SNLS.git"
 
-    maintainers("rblake-llnl")
+    maintainers("rcarson3")
 
     license("BSD-3-Clause", checked_by="rblake-llnl")
 
@@ -49,21 +49,22 @@ class Snls(CMakePackage, CudaPackage, ROCmPackage):
 
     variant("shared", default=True, description="build shared libs")
     variant("tests", default=False, description="Build with tests enabled")
+    variant("batch_solver", default=True, description="enable batch solver")
     variant(
         "cxxstd",
-        default="20",
-        values=("11", "14", "17", "20"),
+        default="17",
+        values=("17", "20"),
         description="C++ standard to build with",
     )
 
     depends_on("blt", type=("build"))
-    depends_on("c", type=("build"))
+    depends_on("c", type=("build"), when="+tests")
     depends_on("cxx", type=("build"))
-    depends_on("chai")
-    depends_on("umpire")
-    depends_on("raja")
-    depends_on("camp")
-    depends_on("fmt")
+    depends_on("chai", when="+batch_solver")
+    depends_on("umpire", when="+batch_solver")
+    depends_on("raja", when="+batch_solver")
+    depends_on("camp", when="+batch_solver")
+    depends_on("fmt", when="+batch_solver")
 
     @property
     def cxx_std(self):
@@ -77,7 +78,7 @@ class Snls(CMakePackage, CudaPackage, ROCmPackage):
         args = [
             self.define_from_variant("ENABLE_TESTS", "tests"),
             self.define_from_variant("ENABLE_GTEST", "tests"),
-            self.define("USE_BATCH_SOLVERS", True),
+            self.define_from_variant("USE_BATCH_SOLVERS", "batch_solver"),
             self.define_from_variant("BUILD_SHARED_LIBS", "shared"),
             self.define("BLT_CXX_STD", f"c++{self.cxx_std}"),
             self.define("BLT_SOURCE_DIR", spec["blt"].prefix),
--- a/repos/spack_repo/builtin/packages/snls/package.py
+++ b/repos/spack_repo/builtin/packages/snls/package.py
@@ -50,6 +50,7 @@ class Snls(CMakePackage):
     variant("shared", default=True, description="build shared libs")
     variant("tests", default=False, description="Build with tests enabled")
     variant("batch_solver", default=True, description="enable batch solver")
+    variant("use_raja_only", default=False, description="only use raja")
     variant(
         "cxxstd",
         default="17",
@@ -57,12 +58,15 @@ class Snls(CMakePackage):
         description="C++ standard to build with",
     )
 
+    conflicts("+batch_solver", when="+use_raja_only")
+
     depends_on("blt", type=("build"))
     depends_on("c", type=("build"), when="+tests")
     depends_on("cxx", type=("build"))
     depends_on("chai", when="+batch_solver")
     depends_on("umpire", when="+batch_solver")
     depends_on("raja", when="+batch_solver")
+    depends_on("raja", when="+use_raja_only")
     depends_on("camp", when="+batch_solver")
     depends_on("fmt", when="+batch_solver")
 
@@ -80,6 +84,7 @@ class Snls(CMakePackage):
             self.define_from_variant("ENABLE_GTEST", "tests"),
             self.define_from_variant("USE_BATCH_SOLVERS", "batch_solver"),
             self.define_from_variant("BUILD_SHARED_LIBS", "shared"),
+            self.define_from_variant("USE_RAJA_ONLY", "use_raja_only"),
             self.define("BLT_CXX_STD", f"c++{self.cxx_std}"),
             self.define("BLT_SOURCE_DIR", spec["blt"].prefix),
             self.define("RAJA_DIR", spec["raja"].prefix),
--- a/repos/spack_repo/builtin/packages/snls/package.py
+++ b/repos/spack_repo/builtin/packages/snls/package.py
@@ -56,10 +56,6 @@ class Snls(CMakePackage, CudaPackage, ROCmPackage):
         description="C++ standard to build with",
     )
 
-    @property
-    def cxx_std(self):
-        return self.spec.variants.get("cxxstd").value
-
     depends_on("blt", type=("build"))
     depends_on("c", type=("build"))
     depends_on("cxx", type=("build"))
@@ -68,6 +64,10 @@ class Snls(CMakePackage, CudaPackage, ROCmPackage):
     depends_on("raja")
     depends_on("camp")
     depends_on("fmt")
+
+    @property
+    def cxx_std(self):
+        return self.spec.variants.get("cxxstd").value
 
     def cmake_args(self):
         """
--- a/repos/spack_repo/builtin/packages/snls/package.py
+++ b/repos/spack_repo/builtin/packages/snls/package.py
@@ -81,10 +81,10 @@ class Snls(CMakePackage, CudaPackage, ROCmPackage):
             self.define_from_variant("BUILD_SHARED_LIBS", "shared"),
             self.define("BLT_CXX_STD", f"c++{self.cxx_std}"),
             self.define("BLT_SOURCE_DIR", spec["blt"].prefix),
-            self.define("RAJA_DIR", spec["raja"].prefix.lib.cmake.raja),
-            self.define("CAMP_DIR", spec["camp"].prefix.lib.cmake.camp),
+            self.define("RAJA_DIR", spec["raja"].prefix),
+            self.define("CAMP_DIR", spec["camp"].prefix),
             self.define("UMPIRE_DIR", spec["umpire"].prefix),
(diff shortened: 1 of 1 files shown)


dtaller:
@spackbot fix style

dtaller:
@spackbot fix style

repos/spack_repo/builtin/packages/snls/package.py
```diff
@@ -0,0 +1,90 @@
...
+    depends_on("c", type=("build"))
+    depends_on("cxx", type=("build"))
+    depends_on("chai")
+    depends_on("umpire")
+    depends_on("raja")
+    depends_on("camp")
+    depends_on("fmt")
+
+    @property
+    def cxx_std(self):
```
gberg617: Was this actually needed for anything?

repos/spack_repo/builtin/packages/snls/package.py
```diff
@@ -0,0 +1,90 @@
...
+
+from spack_repo.builtin.build_systems.cmake import CMakePackage
+from spack_repo.builtin.build_systems.cuda import CudaPackage
+from spack_repo.builtin.build_systems.rocm import ROCmPackage
+
+from spack.package import *
+
+
+class Snls(CMakePackage, CudaPackage, ROCmPackage):
+    """Small solver for Nonlinear systems of equations"""
```
rcarson3: Update comment to "Small Non-Linear Solver for solving non-linear systems of equations"

repos/spack_repo/builtin/packages/snls/package.py
```diff
@@ -0,0 +1,90 @@
...
+
+    homepage = "https://github.com/llnl/SNLS"
+    url = "https://github.com/llnl/SNLS/archive/refs/tags/v0.4.4.tar.gz"
+    git = "https://github.com/llnl/SNLS.git"
+
+    maintainers("rblake-llnl")
```
rcarson3: "rblake-llnl" is not a maintainer of the library. He also hasn't ever contributed to this library either. You should update it to my github handle "rcarson3" as I am the principal developer / maintainer of it.

repos/spack_repo/builtin/packages/snls/package.py
```diff
@@ -0,0 +1,90 @@
...
+        values=("11", "14", "17", "20"),
+        description="C++ standard to build with",
+    )
+
+    depends_on("blt", type=("build"))
+    depends_on("c", type=("build"))
+    depends_on("cxx", type=("build"))
+    depends_on("chai")
+    depends_on("umpire")
+    depends_on("raja")
```
gberg617: I believe we can still build without any RAJA/CHAI/Umpire support. I think this depends on `USE_BATCH_SOLVERS` defined. @rcarson3 may know more about this.

repos/spack_repo/builtin/packages/snls/package.py
```diff
@@ -0,0 +1,90 @@
...
+    )
+
+    variant("shared", default=True, description="build shared libs")
+    variant("tests", default=False, description="Build with tests enabled")
+    variant(
+        "cxxstd",
+        default="20",
+        values=("11", "14", "17", "20"),
+        description="C++ standard to build with",
+    )
```
rcarson3, line 57: SNLS requires C++17 by default and will not compile with anything below that. So, I would remove C++11 and C++14. So, I would probably make C++17 the default at least until we require some feature within C++20.

repos/spack_repo/builtin/packages/snls/package.py
```diff
@@ -0,0 +1,90 @@
...
+# SPDX-License-Identifier: (Apache-2.0 OR MIT)
+
+from spack_repo.builtin.build_systems.cmake import CMakePackage
+from spack_repo.builtin.build_systems.cuda import CudaPackage
+from spack_repo.builtin.build_systems.rocm import ROCmPackage
+
+from spack.package import *
+
+
+class Snls(CMakePackage, CudaPackage, ROCmPackage):
```
gberg617: I haven't tested my package with Cuda/Hip yet.  Did you confirm that this package works on the relevant platforms?

repos/spack_repo/builtin/packages/snls/package.py
```diff
@@ -0,0 +1,90 @@
...
+    variant("tests", default=False, description="Build with tests enabled")
+    variant(
+        "cxxstd",
+        default="20",
+        values=("11", "14", "17", "20"),
+        description="C++ standard to build with",
+    )
+
+    depends_on("blt", type=("build"))
+    depends_on("c", type=("build"))
```
rcarson3: We actually don't need C at all except if tests are enabled and we are building with HIP do to fun reasons related to GTest...

repos/spack_repo/builtin/packages/snls/package.py
```diff
@@ -0,0 +1,90 @@
...
+    depends_on("cxx", type=("build"))
+    depends_on("chai")
+    depends_on("umpire")
+    depends_on("raja")
+    depends_on("camp")
+    depends_on("fmt")
```
rcarson3: These are all optional build requirements. The RAJA Portability Suite is only required if the `USE_BATCH_SOLVERS` CMake option is turned on which enables the batch solver suite, some additional forall/view type abstractions, and access to the memory manager functionality. This option is also the only one that would need `fmt` as Umpire makes use of it internally.

RAJA is only required if the `USE_RAJA_ONLY` CMake option is turned on which just enables our forall abstraction layers and SubView types internally. Note, `USE_BATCH_SOLVERS` and `USE_RAJA_ONLY` should not be allowed at the same time. If a user is asking for the batch solvers that will bring in everything that the `USE_RAJA_ONLY` would enable and therefore only the batch solver options should be enabled.

repos/spack_repo/builtin/packages/snls/package.py
dtaller: I used this as a shortcut below self.define("BLT_CXX_STD", f"c++{self.cxx_std}"),

repos/spack_repo/builtin/packages/snls/package.py
dtaller: I'll remove this. I originally got this recipe from someone else, and it had this, but I can delete it until we test gpus.

repos/spack_repo/builtin/packages/snls/package.py
dtaller: Will change

repos/spack_repo/builtin/packages/snls/package.py
dtaller: changing

repos/spack_repo/builtin/packages/snls/package.py
dtaller: removing 11 and 14, making 17 default

repos/spack_repo/builtin/packages/snls/package.py
dtaller: Adding batch_solver variant, depend on this conditional on that variant

repos/spack_repo/builtin/packages/snls/package.py
dtaller: I removed this package inheriting from hip and cuda as per Guy's comment (since I have not tested that with spack). But I can make this when="+tests"

dtaller:
@spackbot fix style

repos/spack_repo/builtin/packages/snls/package.py
dtaller: I added a Use_raja_only variant, and a conflict conflicts("+batch_solver", when="+use_raja_only"), and raja will be used if one of those options is on. Let me know if what I added is what you intended or if I misunderstood.

rcarson3: I think it should at least function how we might expect now. One thing would be if the:

```
depends_on("raja", when="+batch_solver")
depends_on("raja", when="+use_raja_only")
```

is the standard spack way of doing this or if there is like some sorta logical or equivalent like maybe something below

```
depends_on("raja", when="+batch_solver +use_raja_only")
```

dtaller: I am pretty sure that depends_on("raja", when="+batch_solver +use_raja_only") would work like an "and", requiring both +batch_solver and +use_raja_only to be on at the same time which is not what we want. What I have now on two lines is like an "or", in effect. 

I tried looking and did not see a way to do an "or" on the same line, but probably when a member of the spack team looks over this they will be able to tell us if there is a cleaner way

repos/spack_repo/builtin/packages/snls/package.py
```diff
@@ -0,0 +1,94 @@
...
+    variant("batch_solver", default=True, description="enable batch solver")
+    variant("use_raja_only", default=False, description="only use raja")
+    variant(
+        "cxxstd",
+        default="17",
+        values=("17", "20"),
+        description="C++ standard to build with",
+    )
+
+    conflicts("+batch_solver", when="+use_raja_only")
```
cmelone:
```suggestion
    conflicts("+batch_solver +use_raja_only")
```

may be a cleaner way to express two incompatible variants. also, adding an `msg` arg can be helpful to explain to end users

repos/spack_repo/builtin/packages/snls/package.py
```diff
@@ -0,0 +1,94 @@
...
+
+    depends_on("blt", type=("build"))
+    depends_on("c", type=("build"), when="+tests")
+    depends_on("cxx", type=("build"))
+    depends_on("chai", when="+batch_solver")
+    depends_on("umpire", when="+batch_solver")
+    depends_on("raja", when="+batch_solver")
+    depends_on("raja", when="+use_raja_only")
+    depends_on("camp", when="+batch_solver")
+    depends_on("fmt", when="+batch_solver")
```
cmelone: does snls depend on fmt directly, or is it a transitive dependency (via umpire)?

repos/spack_repo/builtin/packages/snls/package.py
```diff
@@ -0,0 +1,94 @@
...
+    variant(
+        "cxxstd",
+        default="17",
+        values=("17", "20"),
+        description="C++ standard to build with",
+    )
+
+    conflicts("+batch_solver", when="+use_raja_only")
+
+    depends_on("blt", type=("build"))
```
cmelone: add a dependency on your required cmake version

repos/spack_repo/builtin/packages/snls/package.py
cmelone: you can also set `spec.variants["cxxstd"].value` within `cmake_args`

repos/spack_repo/builtin/packages/snls/package.py
```diff
@@ -0,0 +1,94 @@
...
+            self.define_from_variant("USE_BATCH_SOLVERS", "batch_solver"),
+            self.define_from_variant("BUILD_SHARED_LIBS", "shared"),
+            self.define_from_variant("USE_RAJA_ONLY", "use_raja_only"),
+            self.define("BLT_CXX_STD", f"c++{self.cxx_std}"),
+            self.define("BLT_SOURCE_DIR", spec["blt"].prefix),
+            self.define("RAJA_DIR", spec["raja"].prefix),
+            self.define("CAMP_DIR", spec["camp"].prefix),
+            self.define("UMPIRE_DIR", spec["umpire"].prefix),
+            self.define("CHAI_DIR", spec["chai"].prefix),
+            self.define("FMT_DIR", spec["fmt"].prefix),
```
cmelone: these should be guarded for optional dependencies

something like:
```python
if "+batch_solver" in spec or "+use_raja_only" in spec:
    args.append(self.define("RAJA_DIR", spec["raja"].prefix))
```

and the same for the rest of the dependencies guarded on `+batch_solver`

cmelone (changes requested):

repos/spack_repo/builtin/packages/snls/package.py
cmelone: correct, there is no OR mechanism within the `when` arg

repos/spack_repo/builtin/packages/snls/package.py
dtaller: Changed, added a message

repos/spack_repo/builtin/packages/snls/package.py
dtaller: Added

repos/spack_repo/builtin/packages/snls/package.py
dtaller: Added spec.satisfies(+variant) blocks here

dtaller:
@spackbot fix style

repos/spack_repo/builtin/packages/snls/package.py
dtaller: I think that snls does not use fmt directly (no called to fmt:: or includes in its source code). However, if I get rid of the FMT_DIR below then the build system dies and it complains that it needs to find fmt, so I was not able to remove this

repos/spack_repo/builtin/packages/snls/package.py
cmelone: @dtaller since you're not using `cxx_std` anywhere else, can you move this variable to `cmake_args`?

repos/spack_repo/builtin/packages/snls/package.py
```diff
@@ -0,0 +1,104 @@
...
+
+    conflicts(
+        "+batch_solver +use_raja_only",
+        msg="raja_only option does not allow snls batch_solver to work",
+    )
+
+    depends_on("cmake@3.21:", type="build")
+    depends_on("blt", type=("build"))
+    depends_on("c", type=("build"), when="+tests")
+    depends_on("cxx", type=("build"))
```
cmelone:
```suggestion
    with default_args(type="build"):
        depends_on("cmake@3.21:")
        depends_on("blt")
        depends_on("cxx")
        depends_on("c", when="+tests")
```

you might also consider using this paradigm to clean up the +batch_solver depends_on statements:

```
with when("+batch_solver"):
```

repos/spack_repo/builtin/packages/snls/package.py
cmelone: I was able to install snls 0.4.0-0.4.4 find without passing FMT_DIR. I got this warning but the spack install process didn't exit, are you seeing fatal errors?

```
CMake Warning at cmake/thirdpartylibraries/SetupThirdPartyLibraries.cmake:65 (message):
  FMT_DIR was not provided.  This is a requirement for camp as of v2024.02.0.
```

dtaller:
@spackbot fix style

repos/spack_repo/builtin/packages/snls/package.py
dtaller: Changed

repos/spack_repo/builtin/packages/snls/package.py
dtaller: Changed, moved this

repos/spack_repo/builtin/packages/snls/package.py
dtaller: I removed the dependency on fmt.

I am using snls as part of a project at LLNL, and it seems that a teammate took snls and modified cmake/thirdpartylibraries/SetupThirdPartyLibraries.cmake to make this warning fatal, so it is fatal for me.

but that is not part of the public snls, I can fix things or have a patch on my end internally.

dtaller:
@spackbot fix style

dtaller:
Just to reiterate, responding to this:

```
I was able to install snls 0.4.0-0.4.4 find without passing FMT_DIR. I got this warning but the spack install process didn't exit, are you seeing fatal errors?

CMake Warning at cmake/thirdpartylibraries/SetupThirdPartyLibraries.cmake:65 (message):
  FMT_DIR was not provided.  This is a requirement for camp as of v2024.02.0.


```

I removed the dependency on fmt.

I am using snls as part of a project at LLNL, and it seems that a teammate took snls and modified cmake/thirdpartylibraries/SetupThirdPartyLibraries.cmake to make this warning fatal, so it is fatal for me.

but that is not part of the public snls, I can fix things or have a patch on my end internally.

cmelone (approved):
LGTM, I'll wait on a final review from @rcarson3 and/or @gberg617 before merging

rcarson3 (approved):
Looks good to me now

repos/spack_repo/builtin/packages/snls/package.py
```diff
@@ -0,0 +1,101 @@
...
+            self.define_from_variant("USE_RAJA_ONLY", "use_raja_only"),
+            self.define("BLT_CXX_STD", f"c++{spec.variants.get('cxxstd').value}"),
+            self.define("BLT_SOURCE_DIR", spec["blt"].prefix),
+        ]
+
+        if spec.satisfies("+batch_solver") or spec.satisfies("+use_raja_only"):
+            args.append(self.define("RAJA_DIR", spec["raja"].prefix))
+
+        if spec.satisfies("+batch_solver"):
+            args.append(self.define("RAJA_DIR", spec["raja"].prefix))
```
cmelone: this line is redundant with L93
```suggestion
```

cmelone (changes requested):

repos/spack_repo/builtin/packages/snls/package.py
dtaller: Removed

cmelone (approved):

PR #5582, opened 2026-07-14
run_ctest_after_install_step with reviewer fix

jeffmauldin:
change seacas package.py to run ctest after install, and not run ctest in check() callback


jeffmauldin:
@tokusanya
Hey Tolu. This is jeff. You should get an email with this comment. Maybe it will trigger a spackbot thing.

jeffmauldin:
Couldn't get tolu review triggered

PR #5584, opened 2026-07-15
r-gage: new package

w8jcik:


bernhardkaindl (approved):
Built both versions on Ubuntu 24.04

PR #5585, opened 2026-07-15
r-topgo: add v2.60.1, v2.64.0

w8jcik:


bernhardkaindl (approved):
Built the new versions on Ubuntu 24.04

Issue #3077, opened 2026-01-21
Installation issue: ginkgo 1.11.0 with cuda


MarcelKoch:
### Steps to reproduce the issue

Run spack-packages CI pipeline for the new Ginkgo version, see #2767.

### Error message

<details><summary>Error message</summary>

<pre>
-- The CXX compiler identification is GNU 11.4.0
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /home/software/spack/[padded-to-256-chars]/linux-neoverse_v2/compiler-wrapper-1.0-evq3teks3gdqyoc4gukdcxgkntkyy5lf/libexec/spack/gcc/g++ - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Could NOT find PAPI (missing: PAPI_LIBRARY PAPI_INCLUDE_DIR sde) (Required is at least version "7.0.1.0")
-- The CUDA compiler identification is NVIDIA 13.0.88 with host compiler GNU 11.4.0
-- Detecting CUDA compiler ABI info
-- Detecting CUDA compiler ABI info - done
-- Check for working CUDA compiler: /home/software/spack/[padded-to-256-chars]/linux-neoverse_v2/cuda-13.0.2-72trsaxvd3jgxtziqo3hnwnligz7zpsa/bin/nvcc - skipped
-- Detecting CUDA compile features
-- Detecting CUDA compile features - done
-- Found CUDAToolkit: /home/software/spack/[padded-to-256-chars]/linux-neoverse_v2/cuda-13.0.2-72trsaxvd3jgxtziqo3hnwnligz7zpsa/targets/sbsa-linux/include;/home/software/spack/[padded-to-347-chars]/linux-neoverse_v2/cuda-13.0.2-72trsaxvd3jgxtziqo3hnwnligz7zpsa/targets/sbsa-linux/include/cccl (found version "13.0.88")
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD - Success
-- Found Threads: TRUE
-- The CUDA compiler supports the following architectures: 100;103;110;120;121;75;80;86;87;88;89;90
-- Found NVTX: /home/software/spack/[padded-to-256-chars]/linux-neoverse_v2/cuda-13.0.2-72trsaxvd3jgxtziqo3hnwnligz7zpsa/targets/sbsa-linux/include/nvtx3
[...]

PR #5591, opened 2026-07-15
cccl: new package

Gusser93:
The CUDA Core Compute Libraries (CCCL) are a collection of important libraries for developing CUDA applications. They are shipped with the CUDA toolkit, but an own package allows using the newer features like the CCCL runtime on older hardware which only supports CUDA 12.

I tested the package locally without problems. I plan to include variants in the future after more testing.

Could help with #3077

@spackbot fix style

### DIFF
--- a/repos/spack_repo/builtin/packages/cccl/package.py
+++ b/repos/spack_repo/builtin/packages/cccl/package.py
@@ -39,7 +39,7 @@ class Cccl(CMakePackage, CudaPackage):
     variant("cuda", default=True, description="Build with CUDA")
 
     depends_on("cmake@3.21:", type="build")
-    depends_on("cxx")
-    depends_on("cuda@12:", when="@3")
+    depends_on("cxx", type="build")
+    depends_on("cuda@12:", type=("build", "link"), when="@3")
 
     requires("+cuda")
--- a/repos/spack_repo/builtin/packages/cccl/package.py
+++ b/repos/spack_repo/builtin/packages/cccl/package.py
@@ -3,28 +3,14 @@
 # SPDX-License-Identifier: (Apache-2.0 OR MIT)
 
 from spack_repo.builtin.build_systems.cmake import CMakePackage
+from spack_repo.builtin.build_systems.cuda import CudaPackage
 
 from spack.package import *
 
 
-class Cccl(CMakePackage):
+class Cccl(CMakePackage, CudaPackage):
     """CUDA Core Compute Libraries (CCCL)
-
-    Welcome to the CUDA Core Compute Libraries (CCCL) where our mission is to make CUDA more
-    delightful.
-
-    This repository unifies three essential CUDA C++ libraries into a single, convenient
-    repository:
-
-    - Thrust,
-    - CUB,
-    - libcudacxx
-
-    The goal of CCCL is to provide CUDA C++ developers with building blocks that make it easier to
-    write safe and efficient code. Bringing these libraries together streamlines your development
-    process and broadens your ability to leverage the power of CUDA C++.
-
-    CCCL is a header-only library.
+    CCCL provide the following header-only CUDA libraries: Thrust, CUB, and libcudacxx.
     """
 
     homepage = "https://github.com/NVIDIA/cccl"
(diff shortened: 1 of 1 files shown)


Gusser93:
@spackbot fix style

repos/spack_repo/builtin/packages/cccl/package.py
```diff
@@ -0,0 +1,48 @@
...
+
+    - Thrust,
+    - CUB,
+    - libcudacxx
+
+    The goal of CCCL is to provide CUDA C++ developers with building blocks that make it easier to
+    write safe and efficient code. Bringing these libraries together streamlines your development
+    process and broadens your ability to leverage the power of CUDA C++.
+
+    CCCL is a header-only library.
```
cmelone: please shorten this description -- I usually grab them under the "About" section on GitHub

repos/spack_repo/builtin/packages/cccl/package.py
```diff
@@ -0,0 +1,48 @@
...
+        "Apache-2.0 AND Apache-2.0 WITH LLVM-exception AND BSL-1.0 AND LicenseRef-scancode-bsd-unmodified AND BSD-3-Clause",
+        checked_by="gusser93",
+    )
+
+    version("3.3.4", sha256="9d5ae91a71f971c69a16ec139c6882c2c19f74a862c3d90ceaa3c9e8f327e5a6")
+    version("3.3.3", sha256="7aed8bd89049bb75261cc9633e4471e1fcf5fbb5eb5b1aeb3f82ee07e9f60395")
+    version("3.3.2", sha256="7bf03b4f3ab4db8b5781613564a01cf19682e50afc58bb06ced53cd049a52965")
+    version("3.3.1", sha256="95355e7d492d70604705330c12afef785c76048e1084852ceeb31522e2dbf223")
+
+    depends_on("cmake@3.21:", type="build")
```
cmelone, line 41: please add the dependencies for this package. at the very least, you'll need a cxx compiler, cuda (inherit from CudaPackage), and a CMake version reflected in CMakeLists.txt

https://spack.readthedocs.io/en/latest/packaging_guide_build.html#environment-variables

repos/spack_repo/builtin/packages/cccl/package.py
```diff
@@ -0,0 +1,48 @@
...
+    homepage = "https://github.com/NVIDIA/cccl"
+    url = "https://github.com/NVIDIA/cccl/releases/download/v3.3.4/cccl-src-v3.3.4.tar.gz"
+    git = "https://github.com/NVIDIA/cccl"
+
+    supplier = "NVIDIA"
+
+    maintainers("gusser93")
+
+    license(
+        "Apache-2.0 AND Apache-2.0 WITH LLVM-exception AND BSL-1.0 AND LicenseRef-scancode-bsd-unmodified AND BSD-3-Clause",
```
cmelone: I'm not seeing the source of `LicenseRef-scancode-bsd-unmodified`

cmelone (changes requested):

repos/spack_repo/builtin/packages/cccl/package.py
Gusser93: Added it because the license for `thrust::complex` stated in [lines 351-383 in the LICENSE file](https://github.com/NVIDIA/cccl/blob/main/LICENSE#L351-L383). The only reference for this BSD 2-clause with this unmodified copyright notice requirement I found was in the [scancode licensedb](https://scancode-licensedb.aboutcode.org/bsd-unmodified.html). I couldn't find a SPDX approved version which allowed this requirement.

repos/spack_repo/builtin/packages/cccl/package.py
Gusser93: Added the dependencies. Regarding the CMake version: I checked the necessary CMakeLists.txt files for the provided versions and 3.21 was the smallest one I could find. For example Thrust needs 3.21 since CCCL 3.2.

Gusser93:
@spackbot fix style

repos/spack_repo/builtin/packages/cccl/package.py
```diff
@@ -0,0 +1,45 @@
...
+    version("3.3.4", sha256="9d5ae91a71f971c69a16ec139c6882c2c19f74a862c3d90ceaa3c9e8f327e5a6")
+    version("3.3.3", sha256="7aed8bd89049bb75261cc9633e4471e1fcf5fbb5eb5b1aeb3f82ee07e9f60395")
+    version("3.3.2", sha256="7bf03b4f3ab4db8b5781613564a01cf19682e50afc58bb06ced53cd049a52965")
+    version("3.3.1", sha256="95355e7d492d70604705330c12afef785c76048e1084852ceeb31522e2dbf223")
+
+    variant("cuda", default=True, description="Build with CUDA")
+
+    depends_on("cmake@3.21:", type="build")
+    depends_on("cxx")
+    depends_on("cuda@12:", when="@3")
```
cmelone: see the feedback I left on the `nvbench` PR. also, I don't think the `when` directive is necessary at the moment

Gusser93: Done. Regarding the `when`: It's not needed right now, but I've added it as a reminder for future updates, along with a comment. CUDA is evolving rather quickly at the moment, and there are rumours about CUDA 14, so I wanted to be prepared.  
If this approach is inappropriate or flawed, I will remove it and add the relevant code later when necessary.

cmelone (approved):

PR #5592, opened 2026-07-15
nvbench: new package

Gusser93:
nvbench is a library to write performance benchmarks for CUDA kernel.

I tested the package locally without problems. I plan to include variants in the future after more testing.

### DIFF
--- a/repos/spack_repo/builtin/packages/nvbench/package.py
+++ b/repos/spack_repo/builtin/packages/nvbench/package.py
@@ -27,8 +27,8 @@ class Nvbench(CudaPackage, CMakePackage):
     variant("cuda", default=True, description="Build with CUDA")
 
     depends_on("cmake@4:", type="build")
-    depends_on("cxx")
-    depends_on("cuda@12")
+    depends_on("cxx", type="build")
+    depends_on("cuda@12:", type=("build", "link"))
 
     requires("+cuda")
     requires("%gcc@7:", when="%gcc")
--- a/repos/spack_repo/builtin/packages/nvbench/package.py
+++ b/repos/spack_repo/builtin/packages/nvbench/package.py
@@ -3,11 +3,12 @@
 # SPDX-License-Identifier: (Apache-2.0 OR MIT)
 
 from spack_repo.builtin.build_systems.cmake import CMakePackage
+from spack_repo.builtin.build_systems.cuda import CudaPackage
 
 from spack.package import *
 
 
-class Nvbench(CMakePackage):
+class Nvbench(CudaPackage, CMakePackage):
     """NVBench is a C++17 library designed to simplify CUDA kernel benchmarking.
     It solves challenges inherent to CUDA kernel benchmarking."""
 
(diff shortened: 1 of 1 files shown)


Gusser93:
@spackbot fix style

repos/spack_repo/builtin/packages/nvbench/package.py
```diff
@@ -0,0 +1,25 @@
...
+    git = "https://github.com/NVIDIA/nvbench"
+
+    supplier = "NVIDIA"
+
+    maintainers("gusser93")
+
+    license("Apache-2.0 WITH LLVM-exception", checked_by="gusser93")
+
+    version("main", branch="main")
+    depends_on("cmake@4:", type="build")
```
cmelone, line 29: nvbench seems to have additional dependencies -- https://github.com/NVIDIA/nvbench#supported-compilers-and-tools

cmelone (changes requested):

repos/spack_repo/builtin/packages/nvbench/package.py
Gusser93: I added the dependencies. Regarding the different CMake version: the [used rapids CMake script](https://raw.githubusercontent.com/rapidsai/rapids-cmake/branch-25.12/RAPIDS.cmake) requires 4.0 which is not reflected in the README.

Gusser93:
@spackbot fix style

repos/spack_repo/builtin/packages/nvbench/package.py
```diff
@@ -0,0 +1,35 @@
...
+
+    license("Apache-2.0 WITH LLVM-exception", checked_by="gusser93")
+
+    version("main", branch="main")
+
+    variant("cuda", default=True, description="Build with CUDA")
+
+    depends_on("cmake@4:", type="build")
+    depends_on("cxx")
+    depends_on("cuda@12")
```
cmelone: to specify greater than or equal to dependency ranges, you'd do something like `cuda@12:`

repos/spack_repo/builtin/packages/nvbench/package.py
```diff
@@ -0,0 +1,35 @@
...
+
+    supplier = "NVIDIA"
+
+    maintainers("gusser93")
+
+    license("Apache-2.0 WITH LLVM-exception", checked_by="gusser93")
+
+    version("main", branch="main")
+
+    variant("cuda", default=True, description="Build with CUDA")
```
cmelone, line 27: since you inherit from CudaPackage, I don't think you need to add another variant directive here

repos/spack_repo/builtin/packages/nvbench/package.py
cmelone: are these build/link/run dependencies? add the type

cmelone (changes requested):

repos/spack_repo/builtin/packages/nvbench/package.py
Gusser93: Don't I need this to override the [default from the CudaPackage](https://github.com/spack/spack-packages/blob/develop/repos/spack_repo/builtin/build_systems/cuda.py#L69) which is `False`? I took this from similar packages like [nvshmem](https://github.com/spack/spack-packages/blob/develop/repos/spack_repo/builtin/packages/nvshmem/package.py#L46).

cmelone: yeah you're right, my mistake

cmelone (approved):

PR #5599, opened 2026-07-15
flex: fix libfl link failure on macOS

mcmehrtens:
- see #5525


alecbcs (approved):
Looks good to me. Thanks @mcmehrtens

PR #5600, opened 2026-07-15
libevent: fix dylib link failure on macOS

mcmehrtens:
- see #5525

### DIFF
--- a/repos/spack_repo/builtin/packages/libevent/package.py
+++ b/repos/spack_repo/builtin/packages/libevent/package.py
@@ -49,6 +49,14 @@ class Libevent(AutotoolsPackage):
 
     conflicts("+openssl", when="@:2.0")
 
+    def setup_build_environment(self, env: EnvironmentModifications) -> None:
+        # The extra, pthreads, and openssl libs are linked without libevent_core, so they
+        # only link as dylibs when libtool passes -undefined dynamic_lookup, which it does
+        # only for MACOSX_DEPLOYMENT_TARGET=10.*
+        # See https://github.com/spack/spack-packages/pull/5525
+        if self.spec.platform == "darwin":
+            env.set("MACOSX_DEPLOYMENT_TARGET", "10.16")
+
     def url_for_version(self, version):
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/libevent/package.py
```diff
@@ -49,6 +49,14 @@ class Libevent(AutotoolsPackage):
 
     conflicts("+openssl", when="@:2.0")
 
+    def setup_build_environment(self, env: EnvironmentModifications) -> None:
+        # The extra, pthreads, and openssl libs are linked without libevent_core, so they
+        # only link as dylibs when libtool passes -undefined dynamic_lookup, which it does
+        # only for MACOSX_DEPLOYMENT_TARGET=10.*
+        # See https://github.com/spack/spack-packages/pull/5525
+        if self.spec.platform == "darwin":
+            env.set("MACOSX_DEPLOYMENT_TARGET", "10.16")
```
alecbcs, line 60:
```suggestion
    @when("@:2.1.12")
    def setup_build_environment(self, env: EnvironmentModifications) -> None:
        # The extra, pthreads, and openssl libs are linked without libevent_core, so they
        # only link as dylibs when libtool passes -undefined dynamic_lookup, which it does
        # only for MACOSX_DEPLOYMENT_TARGET=10.*
        # Fixed in v2.1.13 and later versions
        # See https://github.com/spack/spack-packages/pull/5525
        if self.spec.platform == "darwin":
            env.set("MACOSX_DEPLOYMENT_TARGET", "10.16")
```

Quick suggestion to only apply this mitigation when <= v2.1.12 since it looks like it's been fixed in v2.1.13.

alecbcs (changes requested):
One quick suggestion, otherwise looks good to me. Thanks @mcmehrtens!

mcmehrtens:
@alecbcs good catch—thanks!

alecbcs (approved):
Looks good to me. Thanks @mcmehrtens!

PR #5603, opened 2026-07-15
r-rtracklayer: new versions, added new r-seqinfo dependency

emwjacobson:
I've added a handful of missing versions from rtracklayer. r-seqinfo also becomes a dependency of r-rtracklayer starting with version 1.70.1, so I added it as a new package.


bernhardkaindl:
@emwjacobson, there is a conflict in the r-seqinfo recipe you need to fix.

emwjacobson:
Looks like someone else added the r-seqinfo package in parallel. I went ahead and merged their package into this PR. Should be good to go now.

alecbcs (approved):
Looks good to me. Thanks @emwjacobson

PR #5606, opened 2026-07-16
r-genomeinfodb: add @1.44.3,1.48; r-seqinfo: new package

w8jcik:


bernhardkaindl (approved):
Built all new versions on Ubuntu 24.04

PR #5610, opened 2026-07-16
Update patch condition pytorch 2.13

williampiat3:
Issue with air gapped build has not been fixed upstream for now

### DIFF
--- a/repos/spack_repo/builtin/packages/py_torch/package.py
+++ b/repos/spack_repo/builtin/packages/py_torch/package.py
@@ -611,10 +611,9 @@ class PyTorch(PythonPackage, CudaPackage, ROCmPackage):
     # This forwards six source folder path to NNPACK which forwards it to PeachPy
     # for versions @2.5:2.11
     patch("air_gapped_nnpack_cmake_older.patch", when="@2.5:2.11")
-    # for version @2.12: (env forwarding mecanism changed) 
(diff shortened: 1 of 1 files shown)


williampiat3:
@adamjstewart : I need to extend the air-gapped patch for for 2.13 it has not been fixed upstream despite a PR and an issue...

williampiat3:
See here for issue upstream: https://github.com/pytorch/pytorch/issues/188652
See here for PR upstream: https://github.com/pytorch/pytorch/pull/188263

adamjstewart (commented):
Can you add a comment linking to the upstream PR? Let's not put an upper bound on the patch until it's merged. I'll notice this immediately when I try to add a new version and the patch no longer applies.

williampiat3:
@spackbot fix style

williampiat3:
> Can you add a comment linking to the upstream PR? Let's not put an upper bound on the patch until it's merged. I'll notice this immediately when I try to add a new version and the patch no longer applies.

noted! I changed the recipe accordingly. sorry for putting an upper bound in the first place I genuinely thought that the upstream PR would be merged faster

adamjstewart (approved):

adamjstewart:
Upstream development seems to have slowed down thanks to AI, I have my own PRs now left unreviewed for months.

PR #5611, opened 2026-07-16
r-keggrest: add v1.48.1, v1.52.2

w8jcik:


bernhardkaindl (approved):
Built all new versions on Ubuntu 24.04

PR #5620, opened 2026-07-16
python packages: standardize url_for_version() methods

alecbcs:
I noticed we had a couple different competing `url_for_version()` method formats over the python ecosystem.

This PR attempts to standardize the methods so that it's easier to users to pull examples from the existing packages when updating other packages.

@adamjstewart I'm not sure if you would be interested in taking a look at these.

### DIFF
--- a/repos/spack_repo/builtin/packages/py_accessible_pygments/package.py
+++ b/repos/spack_repo/builtin/packages/py_accessible_pygments/package.py
@@ -27,11 +27,9 @@ class PyAccessiblePygments(PythonPackage):
     depends_on("py-hatch-vcs", type="build", when="@0.0.5:")
 
     def url_for_version(self, version):
-        url = (
-            "https://pypi.org/packages/source/a/accessible-pygments/accessible{}pygments-{}.tar.gz"
-        )
+        url = "https://pypi.org/packages/source/a/{0}/{0}-{1}.tar.gz"
         if version < Version("0.0.5"):
-            separator = "-"
+            name = "accessible-pygments"
         else:
-            separator = "_"
-        return url.format(separator, version)
+            name = "accessible_pygments"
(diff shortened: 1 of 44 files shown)


adamjstewart:
1. I prefer the old style
2. ~Many of these changes are incorrect, the PyPI package name doesn't change, only the tarball name~

alecbcs:
> 2. Many of these changes are incorrect, the PyPI package name doesn't change, only the tarball name

The upstream PyPI reports 200 ok if you change the package name as well and the checksums match.

E.g.
```
> spack -d checksum --verify py-cwl-utils 0.37
...
==> [2026-07-16-16:00:41.493948] Checking existence of https://files.pythonhosted.org/packages/source/c/cwl_utils/cwl_utils-0.37.tar.gz
==> [2026-07-16-16:00:41.532987] Found 1 version of py-cwl-utils
    [100%]  360.97 KB @   42.5 MB/s
==> [2026-07-16-16:00:41.723791] Checksummed 1 version of py-cwl-utils:
==> [2026-07-16-16:00:41.723884] Verified 1 of 1

  0.37  [=] Correct
```

adamjstewart:
Does that also apply to changes in capitalization or just hyphenation changes?

alecbcs:
Package names appear to always be normalized by PyPi so we just need to match the same case / separator as is used in the actual tarball name for `{name}` and it'll work.

adamjstewart (approved):
Assuming everything works, I'm okay with this change, although it may be a better use of time to instead rework things to actually use PyPI's API so we never need to guess the URL.

alecbcs:
Let's merge this for now as a short term fix to help improve consistency across the package repo. I'll follow up with another PR later on when I get some time that can look into potentially replacing this by instead using the PyPI API.

PR #5627, opened 2026-07-17
bufr: Add meson dependency for newer python

mathomp4:
In trying to build `bufr` with Python 3.13, it was found it now needs to depend on `meson` due to use of `f2py`


bernhardkaindl (approved):
Confirmed, fixes the missing meson depdency.

PR #5637, opened 2026-07-17
py-mcp: new package (with missing deps, py-pyjwt 2.10.1)

aprozo:
Adds the official [Model Context Protocol (MCP) python SDK](https://github.com/modelcontextprotocol/python-sdk) and its missing dependencies, one commit per package:

- **py-pyjwt: add v2.10.1** — mcp needs `pyjwt>=2.10.1`
- **py-httpx-sse: new package** (0.4.0)
- **py-sse-starlette: new package** (2.1.3)
- **py-mcp: new package** (1.28.1) 
- **py-restrictedpython** (8.5)- not an mcp part, needed by MCP servers that sandbox code exec

PR #5638, opened 2026-07-17
supergateway: new package

aprozo:
Adds [supergateway](https://github.com/supercorp-ai/supergateway): bridges MCP (Model Context Protocol) stdio servers to SSE / streamable HTTP and vice versa. The npm registry tarball ships a prebuilt `dist/`, so the install uses `npm install --global --ignore-scripts` (same npm pattern as `claude-code`).

Companion to #5637 (py-mcp) — used to expose the MCP servers packaged for the EIC's eic-shell ([eic/eic-spack#986](https://github.com/eic/eic-spack/pull/986)) as local HTTP services, but general-purpose for any MCP deployment. Verified: builds and serves streamable-HTTP MCP sessions in that stack.


bernhardkaindl:
See my https://github.com/spack/spack-packages/pull/5639#issuecomment-5284467080

PR #5639, opened 2026-07-17
opencode, bun: new packages

aprozo:
Adds [opencode](https://opencode.ai) ([anomalyco/opencode](https://github.com/anomalyco/opencode)), an open-source AI coding agent CLI — useful on headless systems (farm nodes, CI debugging) where a graphical client is not available — together with the `bun` package it needs to build.

Built from the tagged source archive. opencode is a bun workspace: the lockfile is `bun.lock`, dependencies use bun's `catalog:` / `workspace:` protocols plus `patchedDependencies`, and the build is `Bun.build({ compile: ... })`, which links the sources against the bun runtime into a single executable — so npm cannot drive it. The build steps follow the nixpkgs derivation: `bun install` over the workspaces the CLI pulls in, then `bun --bun ./script/build.ts --single --skip-install` in `packages/opencode`.

`bun` itself is installed from the upstream release archives, in the same shape as the existing `deno` package: bun is Zig plus a patched JavaScriptCore and needs an existing bun to bootstrap. It is a separate commit and can be split into its own PR if preferred.

Dependencies, each confirmed by shimming the tool out of `PATH` and re-running the build: `bun` and `node-js` at build time (the embedded web UI is built with vite, which runs on node), `ripgrep` at run time (otherwise opencode downloads an `rg` binary into the user cache at first use). `git` and `npm` are not needed.

Verified on linux-x86_64: clean `spack install opencode` → `bin/opencode`, `opencode --version` reports `1.18.3`. Companion to #5637; also exercised in the EIC eic-shell stack ([eic/eic-spack#986](https://github.com/eic/eic-spack/pull/986)).

### DIFF
--- a/repos/spack_repo/builtin/packages/bun/package.py
+++ b/repos/spack_repo/builtin/packages/bun/package.py
@@ -1,0 +1,64 @@
+# Copyright Spack Project Developers. See COPYRIGHT file for details.
+#
+# SPDX-License-Identifier: (Apache-2.0 OR MIT)
+
+import platform
+
+from spack_repo.builtin.build_systems.generic import Package
+
+from spack.package import *
+
+
+class Bun(Package):
+    """Bun is an all-in-one JavaScript runtime, bundler, transpiler and package manager."""
+
+    homepage = "https://bun.com/"
+    url = "https://github.com/oven-sh/bun/releases/download/bun-v1.3.14/bun-linux-x64.zip"
+    supplier = "Organization: Oven"
+
+    maintainers("aprozo")
+
+    license("MIT AND LGPL-2.1-only", checked_by="aprozo")
+
+    skip_version_audit = ["platform=windows"]
+
+    sanity_check_is_file = ["bin/bun"]
+
+    bun_versions = {
+        "1.3.14": {
+            "linux": {
+                "x86_64": "951ee2aee855f08595aeec6225226a298d3fea83a3dcd6465c09cbccdf7e848f",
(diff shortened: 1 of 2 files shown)


alecbcs:
@aprozo instead of installing opencode from the binary releases we should look at building it from source archives.

An example package that should be pretty close to this one is [pi/package.py](https://github.com/spack/spack-packages/blob/develop/repos/spack_repo/builtin/packages/pi/package.py).

aprozo:
Thanks a lot, @alecbcs , will do!

aprozo:
@alecbcs, done but had to add bun as well

bernhardkaindl (approved):

bernhardkaindl:
@aprozo, great to have working opencode in spack, well done!"

I'll merge and I assume you can use the provided bun to also build the other package from source?:
- #5638

PS, this got two review comments by Wouter, can you implement them?
You already created a url_for_version for this PR, you can use it again to implement Wouter's suggestion:
- #5637

bernhardkaindl:
@aprozo, I enabled auto-merge, it depends on GitLab CI passing (it is one of the GitHup CI check steps visible in the PR)

In case the GitLab CI pipeline wasn't started (e.g still saying "GitLab CI is disabled for draft PRs"), the direct ping start it would be to send @spackbot this message (just a PR comment with only this command without the quote):
> @spackbot run pipeline

Ah, saw the command despit the quote and is making an attempt to start it ;) (might not start now as the pipeline/workers are super-busy now), if not started by tomorrow, repeat the command again.

PR #5640, opened 2026-07-17
Updates camp to v2026.07.0

bechols97:
Updates camp to version 2026.07.0 and updates RAJA develop to use newer camp for CI testing


smithsg84:
This PR causes RAJA develop builds to fail.   Need RAJA camp updates in RAJA before updating spack-packages.  Will resubmit when dependency ordering is fixed.

PR #5642, opened 2026-07-17
py-pandas-datareader: add 0.11.0, 0.11.1

mathomp4:
More in my series of building things with Python 3.13, this adds py-pandas-datareader 0.11.0. I also added a patch 0.10.0 in case that version is explicitly needed.

Also clean up the PyPI source URL handling for `py-pandas-datareader`.

PyPI changed the sdist filename from `pandas-datareader-...tar.gz` to `pandas_datareader-...tar.gz` in `0.11.x`, which meant `spack checksum py-pandas-datareader` would not discover the new releases correctly from the default package URL.

This update switches the package's default PyPI template to the current `0.11.x` naming scheme and adds a `url_for_version()` override for older releases, so both interactive `spack checksum py-pandas-datareader` and targeted checksums for `0.11.0` / `0.11.1` work cleanly.

### DIFF
--- a/repos/spack_repo/builtin/packages/py_pandas_datareader/package.py
+++ b/repos/spack_repo/builtin/packages/py_pandas_datareader/package.py
@@ -11,30 +11,36 @@ class PyPandasDatareader(PythonPackage):
     """Up-to-date remote data access for pandas. Works for multiple versions of pandas"""
 
     homepage = "https://pypi.org/project/pandas-datareader"
-    pypi = "pandas-datareader/pandas-datareader-0.10.0.tar.gz"
+    pypi = "pandas-datareader/pandas_datareader-0.11.1.tar.gz"
     git = "https://github.com/pydata/pandas-datareader.git"
 
     maintainers("climbfuji")
 
     license("BSD-3-Clause", checked_by="climbfuji")
 
-    version(
-        "0.11.1",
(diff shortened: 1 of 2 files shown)


repos/spack_repo/builtin/packages/py_pandas_datareader/package.py
```diff
@@ -18,13 +18,23 @@ class PyPandasDatareader(PythonPackage):
 
     license("BSD-3-Clause", checked_by="climbfuji")
 
+    version(
+        "0.11.1",
+        url="https://files.pythonhosted.org/packages/58/86/288382815ce0b52585e8fbcdd8167cdbc056e2d750dea84fcc2ba5c36748/pandas_datareader-0.11.1.tar.gz",
```
climbfuji: There might be a better way to do this (url_for_version). I think these url issues are all because - became _ or vice versa.

mathomp4: Yeah, I wanted something that worked, but I'll see if I can figure this out.

mathomp4: Okay. Pushed an update. `spack checksum` will now only show the new urls by default but that's probably better.

climbfuji (approved):

bernhardkaindl (approved):
Approving after maintainer approval and successful builds of all 3 verisons on Ubuntu 24.04

PR #5644, opened 2026-07-17
gmap_gsnap: version bump, improved simd formatting

emwjacobson:
This bumps the version of gmap-gsnap.

I also noticed that the simd version could use some stricter limits on the x86_64 versions as avx2 isnt supported until x86_64_v3. I also noticed that when the default avx2 version could't be satisfied that it would choose a variant in-order, so I attempted to reorder them from most-optimized to least.

Maintainer is listed as @snehring


bernhardkaindl (approved):

bernhardkaindl (approved):

snehring:
Looks fine.

PR #5653, opened 2026-07-19
professor: add v2.5.6

wdconinc:
This PR adds to `professor` the newer version v2.5.6. This also fixes a few issues which likely explain why I could never get professor to build in the HEP stack: the Makefile seems to pick up the `$PWD` where you issue the spack command, unless we add a working_dir and/or replace the use of explicit `$(PWD)` with `$(CURDIR)` (https://stackoverflow.com/questions/52437728/what-is-the-difference-between-pwd-and-curdir). This now also explicitly calls the `./configure` script (home-grown, not autotools).

Test build (`dev_path` with a fake 2.5.1, but it was a clean clone):
```
t7tqfkn professor@2.5.1+interactive build_system=generic dev_path=/home/wdconinc/git/professor
```

### DIFF
--- a/repos/spack_repo/builtin/packages/professor/package.py
+++ b/repos/spack_repo/builtin/packages/professor/package.py
@@ -19,8 +19,6 @@ class Professor(Package):
     maintainers("mjk655")
 
     version("2.5.6", sha256="7537f23078bd56f00e67e1f96c7a24026b255cc26907ad5d5234b8371e49b3c7")
-    version("2.4.2", sha256="d1a602f2301dac5a165d8cb5baf31d8a1a3d92f30a7069cda66b98c595a4fa21")
-    version("2.3.4", sha256="2c293479342d3a74a6261e7139677bf3580ec34c7bc6a389854fc26497c65ad8")
     version("2.3.3", sha256="60c5ba00894c809e2c31018bccf22935a9e1f51c0184468efbdd5d27b211009f")
 
     variant(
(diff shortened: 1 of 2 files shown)


bernhardkaindl (dismissed):

repos/spack_repo/builtin/packages/professor/package.py
```diff
@@ -33,15 +38,22 @@ class Professor(Package):
...
         env.set("PROF_VERSION", self.spec.version)
 
     def install(self, spec, prefix):
-        make()
-        make("PREFIX={0}".format(prefix), "install")
+        with working_dir(self.stage.source_path):
+            configure = Executable("./configure")
+            configure(
+                "--prefix={0}".format(prefix), "--with-eigen={0}".format(spec["eigen"].prefix)
+            )
```
wdconinc: It seems `configure` was only added in 2.5.0.

wdconinc:
@spackbot run pipeline

wdconinc:
@spackbot run pipeline

wdconinc:
@spackbot run pipeline

wdconinc:
@spackbot run pipeline

wdconinc:
This now builds correctly (sigh): https://gitlab.spack.io/spack/spack-packages/-/jobs/23504815

bernhardkaindl (approved):
Noted: Also adds it to the HEP stack's CI build

PR #5656, opened 2026-07-20
motif: fix build with GCC 14 and newer.

nicolamori:
- Patch the code for building with GCC >= 14 (mainly add missing headers).
- Set gnu17 dialect for GCC >= 15 which defaults to gnu23 and generates errors.


bernhardkaindl (approved):
Confirmed: Fixes build with `%gcc@14` and `%gcc@15` on Ubuntu 24.04.

Further tested by building `xnedit+motif`
```py
spack find motif xnedit+motif
-- linux-ubuntu24.04-skylake / %c,cxx=gcc@14.2.0 ----------------
motif@2.3.8  xnedit@1.6.3

-- linux-ubuntu24.04-skylake / %c,cxx=gcc@15.2.0 ----------------
motif@2.3.8  xnedit@1.6.3
```
Also tested by building `cernlib` with `%gcc@14` based using #5660.

PR 5562 tried to fix motif for the new gcc versions too, but downgrading missing includes to warnings didn't work for me.

PR #5658, opened 2026-07-20
xbae: fix build with GCC 14 and newer.

nicolamori:
- Patch the code for building with GCC >= 14:
  - add missing headers
  - replace true and false with xm_true and xm_false as names for variables
- Set gnu17 dialect for GCC >= 15 which defaults to gnu23 and generates errors.


bernhardkaindl (approved):
Confirmed: Fixes build with `%gcc@14` and `%gcc@15`.

PR #5659, opened 2026-07-20
boost: missing upper version limit for cobalt library variant

Giuncan:
`cobalt` variant can be enabled starting from 1.84 version, but in the recipe is missing the `:` to indicate all higher versions.
The typo was not present in older versions of the recipe, and was introduced with the merge of this [pull request](https://github.com/spack/spack-packages/pull/4016) that added version restrinctions for other variants.


hainest (approved):
Good catch. Apologies for missing that in the previous PR.

PR #5661, opened 2026-07-20
libxsmm: scope binutils+gas dependency to @:1.16

Sbozzolo:
The package comment already notes that a recent `as` (from binutils +gas) is only needed to build libxsmm before version 1.17, not afterwards. The dependency was nevertheless unconditional.


bernhardkaindl (approved):
Hi @hfp, reading up on https://github.com/spack/spack/pull/21671#issuecomment-779882282 and following, and the report by @Sbozzolo (and the related PR which this PR blocks), I read this as two sources saying that this should be the correct change, so I think I can approve and merge. If there is any issue, please ping me!

PR #5662, opened 2026-07-20
hdf: fix for GCC 15+ build

mathomp4:
This PR sets `-std=gnu17` when building with GCC 15+ as they now default to `gnu23` and `hdf` raises errors. (This has been done to many other packages, see https://github.com/spack/spack-packages/pull/5656 for example.)

### DIFF
--- a/repos/spack_repo/builtin/packages/hdf/package.py
+++ b/repos/spack_repo/builtin/packages/hdf/package.py
@@ -172,7 +172,7 @@ class Hdf(AutotoolsPackage):
             ):
                 flags.append("-Wno-error=implicit-int")
 
-            if (self.spec.satisfies("%gcc@15:")):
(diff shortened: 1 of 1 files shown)


bernhardkaindl (approved):
@lrknox We haven't seen a review from you but the PR looks good to me and 9 days have passed since the review invitation, so more than enough time was given for review.

As this PR will need to re-run the GitLab CI pipeline, there would still be many hours until it can be merged. If the merge shall be delayed add a comment and in case auto-merge is enabled you might be able to cancel it, I guess (or set the PR to draft if needed).

bernhardkaindl:
Rebased on current develop to ensure GitLab CI runs using / on top of the current develop branch.
This might also delay the start of the pipeline a few hours, but I enabled auto-merge so it should be merged automatically after CI passes.

PR #5663, opened 2026-07-20
ncview: fix for GCC 15+ build

mathomp4:
This PR sets `-std=gnu17` when building with GCC 15+ as they now default to `gnu23` and `ncview` raises errors. (This has been done to many other packages, see https://github.com/spack/spack-packages/pull/5656 for example.)

### DIFF
--- a/repos/spack_repo/builtin/packages/ncview/package.py
+++ b/repos/spack_repo/builtin/packages/ncview/package.py
@@ -25,6 +25,13 @@ class Ncview(AutotoolsPackage):
     depends_on("libpng")
     depends_on("libxaw")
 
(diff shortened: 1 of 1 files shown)


mathomp4:
Since there are no maintainers, I've added @climbfuji as a reviewer (last commit) and @bernhardkaindl (who did something similar for motif).

bernhardkaindl (approved):
Confirmed: Fixes build with %gcc@15.

PS: To clarify, I didn't make the gcc-15 fix for motif, it was done by nicolamori.
I only reviewed, confirmed it, approved and merged it.

bernhardkaindl:
@spackbot run pipeline

PR #5666, opened 2026-07-20
new package: exacmech

dtaller:
Although I am the one upstreaming this, other LLNL employees contributed. Thanks for your work!

### DIFF
--- a/repos/spack_repo/builtin/packages/exacmech/package.py
+++ b/repos/spack_repo/builtin/packages/exacmech/package.py
@@ -91,7 +91,7 @@ class Exacmech(CMakePackage, CudaPackage, ROCmPackage):
             self.define("CHAI_DIR", self.spec["chai"].prefix),
             self.define("UMPIRE_DIR", self.spec["umpire"].prefix),
             self.define("BLT_SOURCE_DIR", self.spec["blt"].prefix),
-            self.define("BLT_CXX_STD", f"c++{spec.variants.get('cxxstd').value}"),
+            self.define("BLT_CXX_STD", f"c++{self.spec.variants.get('cxxstd').value}"),
             self.define_from_variant("BUILD_SHARED_LIBS", "shared"),
             self.define_from_variant("ENABLE_GTEST", "tests"),
             self.define("ENABLE_MINIAPPS", "OFF"),
--- a/repos/spack_repo/builtin/packages/exacmech/package.py
+++ b/repos/spack_repo/builtin/packages/exacmech/package.py
@@ -1,10 +1,11 @@
 # Copyright Spack Project Developers. See COPYRIGHT file for details.
-# 
+#
 # SPDX-License-Identifier: (Apache-2.0 OR MIT)
 
 from spack_repo.builtin.build_systems.cmake import CMakePackage
 from spack_repo.builtin.build_systems.cuda import CudaPackage
 from spack_repo.builtin.build_systems.rocm import ROCmPackage
+
 from spack.package import *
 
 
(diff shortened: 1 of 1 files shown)


dtaller:
@spackbot fix style

repos/spack_repo/builtin/packages/exacmech/package.py
```diff
@@ -0,0 +1,92 @@
...
+        sha256="3a2b229b493cfb3490c4a4cbe280c32de8f361f0aa4b9a2a84412dcbfa7e5db6",
+        url="https://github.com/llnl/ExaCMech/archive/refs/tags/v0.2.0.tar.gz",
+    )
+
+    variant("openmp", default=False, description="Enable OpenMP support")
+    variant("shared", default=False, description="Enables the build of shared libraries")
+
+    depends_on("blt", type="build")
+    depends_on("c", type="build")
+    depends_on("cxx", type="build")
```
rcarson3: Similar to the earlier SNLS PR, you'll want to have this require C++17 or greater.

repos/spack_repo/builtin/packages/exacmech/package.py
```diff
@@ -0,0 +1,92 @@
...
+    def cmake_args(self):
+        args = [
+            self.define("RAJA_DIR", join_path(self.spec["raja"].prefix, "lib/cmake/raja")),
+            self.define("SNLS_DIR", self.spec["snls"].prefix),
+            self.define("CAMP_DIR", self.spec["camp"].prefix),
+            self.define("CHAI_DIR", self.spec["chai"].prefix),
+            self.define("UMPIRE_DIR", self.spec["umpire"].prefix),
+            self.define("BLT_SOURCE_DIR", self.spec["blt"].prefix),
+            self.define_from_variant("BUILD_SHARED_LIBS", "shared"),
+            self.define("ENABLE_SNLS_V03", "ON"),
```
rcarson3: This option can be removed or at least make it dependent on version <= 0.3.4. Since with v0.4 and newer we moved to using standard cmake imports for SNLS (v0.4)

repos/spack_repo/builtin/packages/exacmech/package.py
```diff
@@ -0,0 +1,92 @@
...
+        "v0.2.0",
+        sha256="3a2b229b493cfb3490c4a4cbe280c32de8f361f0aa4b9a2a84412dcbfa7e5db6",
+        url="https://github.com/llnl/ExaCMech/archive/refs/tags/v0.2.0.tar.gz",
+    )
+
+    variant("openmp", default=False, description="Enable OpenMP support")
+    variant("shared", default=False, description="Enables the build of shared libraries")
+
+    depends_on("blt", type="build")
+    depends_on("c", type="build")
```
rcarson3: This is only required if the tests are enabled, and we're building with HIP due to similar issues that I mentioned in the SNLS PR.

repos/spack_repo/builtin/packages/exacmech/package.py
```diff
@@ -0,0 +1,92 @@
...
+    depends_on("cxx", type="build")
+    depends_on("snls")
+    depends_on("raja")
+    depends_on("camp")
+    depends_on("chai")
+    depends_on("umpire")
```
rcarson3: These are really only required for GPU builds or when SNLS is built with the batch solver option enabled (the GPU builds require SNLS to be built with the batch solvers in order to take advantage of it's abstraction layers for the RAJA Portability Suite).

dtaller:
@spackbot fix style

repos/spack_repo/builtin/packages/exacmech/package.py
dtaller: Added batch solver option passed to snls, protect these with block requiring batch solver

repos/spack_repo/builtin/packages/exacmech/package.py
dtaller: Fixed

repos/spack_repo/builtin/packages/exacmech/package.py
dtaller: Changed, only add this when we are at those versions

repos/spack_repo/builtin/packages/exacmech/package.py
dtaller: added cxxstd variant, starts at C++17

dtaller:
@spackbot fix style

repos/spack_repo/builtin/packages/exacmech/package.py
```diff
@@ -0,0 +1,107 @@
...
+        depends_on("c", when="+rocm +tests")
+        depends_on("cxx")
+
+    depends_on("snls")
+    with when("+batch_solver"):
+        depends_on("snls+batch_solver")
+        depends_on("raja")
+        depends_on("camp")
+        depends_on("chai")
+        depends_on("umpire")
```
rcarson3: If the batch solvers is not turned on then the (snls+use_raja_only)  should be used as that is required as of v0.4. However, we always need snls and RAJA no matter what.

dtaller: Ah ok, no worries. I just pushed a change

rcarson3 (approved):

repos/spack_repo/builtin/packages/exacmech/package.py
```diff
@@ -0,0 +1,108 @@
...
+    )
+    version(
+        "v0.3.0",
+        sha256="c879c18c0947f6a6c921b6784ebf436ed75b2af061be428b6eaf60f30b26697d",
+        url="https://github.com/llnl/ExaCMech/archive/refs/tags/v0.3.0.tar.gz",
+    )
+    version(
+        "v0.2.0",
+        sha256="3a2b229b493cfb3490c4a4cbe280c32de8f361f0aa4b9a2a84412dcbfa7e5db6",
+        url="https://github.com/llnl/ExaCMech/archive/refs/tags/v0.2.0.tar.gz",
```
cmelone: spack can usually extrapolate the version url from the top level `url` definition. were you having trouble fetching versions without these args set?

repos/spack_repo/builtin/packages/exacmech/package.py
```diff
@@ -0,0 +1,108 @@
...
+    variant("shared", default=False, description="Enables the build of shared libraries")
+    variant("tests", default=False, description="Build with tests enabled")
+    variant("batch_solver", default=True, description="enable snls batch solver")
+    variant(
+        "cxxstd",
+        default="17",
+        values=("17", "20"),
+        description="C++ standard to build with",
+    )
+
```
cmelone, line 40: can you define the version of cmake this package depends on?

repos/spack_repo/builtin/packages/exacmech/package.py
```diff
@@ -0,0 +1,108 @@
...
+        depends_on("blt")
+        depends_on("c", when="+rocm +tests")
+        depends_on("cxx")
+
+    depends_on("raja")
+    with when("+batch_solver"):
+        depends_on("snls+batch_solver")
+        depends_on("camp")
+        depends_on("chai")
+        depends_on("umpire")
```
cmelone, line 52: naive question: does exacmech depend on camp, chai, or umpire directly? if any of them are not required by exacmech itself, the dependency on `snls+batch_solver` is sufficient

repos/spack_repo/builtin/packages/exacmech/package.py
dtaller: Added

repos/spack_repo/builtin/packages/exacmech/package.py
dtaller: I see things like this in the exacmech build system, so I would be hesitant to remove them. 

`list(APPEND ECMECH_DEPENDS chai umpire) `

```
ecmech_fill_depends_list(LIST_NAME  ECMECH_DEPENDS
                        DEPENDS_ON RAJA camp snls)
```

I also see chai::managed_ptr in the code directly,

dtaller:
@spackbot fix style

repos/spack_repo/builtin/packages/exacmech/package.py
dtaller: Honestly I just have a script that fetches and prints out these version statements with the url included. I'll try removing the urls now and see if anything breaks...

dtaller: Nope, [ci / prechecks / verify-checksums (pull_request)](https://github.com/spack/spack-packages/actions/runs/29848204649/job/88694240895?pr=5666)
ci / prechecks / verify-checksums (pull_request)

fails with ==> Error: Package exacmech does not have a valid URL for version v0.4.3

dtaller: I'll revert this, just a sec

dtaller: Reverted

cmelone: the fix is to remove the `v` prefix in your version directives. i tested this locally and `spack checksum` worked fine

ie 
```
version("0.4.3", sha256="eee...")
```

dtaller:
@spackbot fix style

repos/spack_repo/builtin/packages/exacmech/package.py
dtaller: Ok, just pushed this without the v. Thanks for investigating.

Is the fact that it failed with the v a bug that we should report as an issue either in spack or in spack-packages? Or is that intended?

cmelone: i believe it's the standard in spack. because (most) github releases start with `v`, spack uses whatever comes after as the version identifier

cmelone (approved):

PR #5672, opened 2026-07-21
ncview: version bump, fixed versions URL

emwjacobson:
Bumps the versions of ncview, also fixes the `url_for_version` function as from first glace it wouldn't have ever worked for version 2.1.8...

No maintainer for ncview.


hainest (approved):
Verified SHAs.

PR #5678, opened 2026-07-21
matrix-switch: version bump, fixed detection of lapack/scalapack

emwjacobson:
I was running into problems getting matrix-switch installed using netlib-scalapack as the scalapack provider. A fix to this appeared to be depending on C when netlib-scalapack was the provider.

Additionally I noticed that during program compilation, it'd continue to choose OpenBLAS as the Scalapack provider, regardless of what the scalapack provider was chosen to be. I added some tooling to hopefully be a bit more strict with that (Sca)LAPACK libraries are chosen for compilation.

Also added a couple versions. While version 1.5.0 is _technically_ available as well, it appeared to have some function name conflicts so I opted to leave it out rather than writing a patch file for the outdated version.

Maintainer is @RMeli

### DIFF
--- a/repos/spack_repo/builtin/packages/matrix_switch/package.py
+++ b/repos/spack_repo/builtin/packages/matrix_switch/package.py
@@ -50,12 +50,12 @@ class MatrixSwitch(CMakePackage):
         ]
 
         if self.spec.satisfies("+lapack"):
-            args.append(self.define("LAPACK_LIBRARY", self.spec['lapack'].libs.ld_flags))
-            args.append(self.define("LAPACK_LIBRARY_DIR", self.spec['lapack'].prefix.lib))
+            args.append(self.define("LAPACK_LIBRARY", self.spec["lapack"].libs.ld_flags))
+            args.append(self.define("LAPACK_LIBRARY_DIR", self.spec["lapack"].prefix.lib))
 
         if self.spec.satisfies("+scalapack"):
-            args.append(self.define("SCALAPACK_LIBRARY", self.spec['scalapack'].libs.ld_flags))
-            args.append(self.define("SCALAPACK_LIBRARY_DIR", self.spec['scalapack'].prefix.lib))
+            args.append(self.define("SCALAPACK_LIBRARY", self.spec["scalapack"].libs.ld_flags))
+            args.append(self.define("SCALAPACK_LIBRARY_DIR", self.spec["scalapack"].prefix.lib))
 
         if self.spec.satisfies("+dbcsr"):
             args.append(self.define("DBCSR_ROOT", self.spec["dbcsr"].prefix))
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/matrix_switch/package.py
```diff
@@ -29,6 +31,7 @@ class MatrixSwitch(CMakePackage):
     variant("dbcsr", default=False, when="+mpi", description="Build with DBCSR interface.")
 
     depends_on("fortran", type="build")  # generated
```
RMeli:
```suggestion
    depends_on("fortran", type="build")
```

repos/spack_repo/builtin/packages/matrix_switch/package.py
```diff
@@ -46,6 +49,14 @@ def cmake_args(self):
             self.define_from_variant("WITH_DBCSR", "dbcsr"),
         ]
 
+        if self.spec.satisfies("+lapack"):
+            args.append(self.define("LAPACK_LIBRARY", self.spec["lapack"].libs.ld_flags))
+            args.append(self.define("LAPACK_LIBRARY_DIR", self.spec["lapack"].prefix.lib))
+
+        if self.spec.satisfies("+scalapack"):
+            args.append(self.define("SCALAPACK_LIBRARY", self.spec["scalapack"].libs.ld_flags))
+            args.append(self.define("SCALAPACK_LIBRARY_DIR", self.spec["scalapack"].prefix.lib))
```
RMeli: Worth setting `SCALAPACK_DETECTION=OFF` as well?

repos/spack_repo/builtin/packages/matrix_switch/package.py
```diff
@@ -46,6 +49,14 @@ def cmake_args(self):
             self.define_from_variant("WITH_DBCSR", "dbcsr"),
         ]
 
+        if self.spec.satisfies("+lapack"):
+            args.append(self.define("LAPACK_LIBRARY", self.spec["lapack"].libs.ld_flags))
+            args.append(self.define("LAPACK_LIBRARY_DIR", self.spec["lapack"].prefix.lib))
```
RMeli: Worth setting `LAPACK_LINKER_FLAG` and `LAPACK_DETECTION=OFF` too? 

I'm also wondering if it's worth doing the same with BLAS at this point.

emwjacobson: I assume the intention of setting `LAPACK_LINKER_FLAG` would be to have `LAPACK_LIBRARY=/path/to/libopenblas.so` and `LAPACK_LINKER_FLAG=-lopenblas` (or the respective paths for other lapack implementations)? If so, then my latest commit should address this.

RMeli (approved):
Thanks @emwjacobson

PR #5689, opened 2026-07-21
ExodusII shared build fix

gberg617:
add SEACAS CMake option to disable shared build when the shared variant is set to false


gsjaardema (commented):
looks good to me

gberg617:
@spackbot fix style

cmelone (approved):

PR #5691, opened 2026-07-22
perl-statistics-regression: new package

w8jcik:

### DIFF
--- a/repos/spack_repo/builtin/packages/perl_statistics_regression/package.py
+++ b/repos/spack_repo/builtin/packages/perl_statistics_regression/package.py
@@ -15,6 +15,8 @@ class PerlStatisticsRegression(PerlPackage):
 
     license("GPL-1.0-or-later")
 
+    maintainers("w8jcik")
+
     version("0.53", sha256="2cc53e3996dda4dceb5d7a794ba15107adb7a6614e19b485c0b8e47a5ab8b69a")
 
     depends_on("perl-module-install", type="build")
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/perl_statistics_regression/package.py
```diff
@@ -0,0 +1,22 @@
...
+from spack.package import *
+
+
+class PerlStatisticsRegression(PerlPackage):
+    """Weighted linear regression (line+plane fitting)"""
+
+    homepage = "https://metacpan.org/pod/Statistics::Regression"
+    url = "https://cpan.metacpan.org/authors/id/I/IA/IAWELCH/Statistics-Regression-0.53.tar.gz"
+
+    license("GPL-1.0")
```
bernhardkaindl: Hi @w8jcik, "GPL-1.0" is not a valid SPDX license string, it must be "-only" or "-or-later": https://spdx.org/licenses/

As https://metacpan.org/pod/Statistics::Regression#LICENSE says "This module is released for free public use under a GPL license.", please change to:

```suggestion
    license("GPL-1.0-or-later")
```

bernhardkaindl (changes requested):

repos/spack_repo/builtin/packages/perl_statistics_regression/package.py
bernhardkaindl: Also, maybe I'd like to suggest (at your discretion) you might add yourself as maintainer so you'd get Cc-ed to PRs that would do anything to the package, e.g. remove or deprecate it.

w8jcik: I have added myself as a maintainer and applied the change to SPDX, as you suggested.
Thank you for the review and for spotting the mistake.

bernhardkaindl (approved):

PR #5694, opened 2026-07-22
mumps: Fix package.py to use mipf90 for linking on darwin other than …

marcalexanderschweitzer:
…10.8/10.9

The current package.py claims to need system FC for linking on 10.8 and 10.9 but does not check for version. On 25.x at least I need mpif90 for linking to avoid segfaults.

### DIFF
--- a/repos/spack_repo/builtin/packages/mumps/package.py
+++ b/repos/spack_repo/builtin/packages/mumps/package.py
@@ -5,7 +5,6 @@
 import glob
 import os
 import sys
-import platform
 
 from spack_repo.builtin.build_systems.generic import Package
 
@@ -67,6 +66,7 @@ class Mumps(Package):
         description="Allow BLAS calls in OpenMP regions "
         + "(warning: might not be supported by all multithread BLAS)",
     )
+    variant("pkgconfig", default=False, description="Create unofficial pkgconfig files")
 
     depends_on("c", type="build")  # generated
     depends_on("fortran", type="build")  # generated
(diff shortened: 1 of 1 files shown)


jcortial-safran (changes requested):
Why was the pkgconfig variant removed ?

marcalexanderschweitzer:
Sorry, my mistake - too many simultaneous edits/versions. Removing pkgconf variant was not intended.

repos/spack_repo/builtin/packages/mumps/package.py
```diff
@@ -339,16 +340,24 @@ def write_makefile_inc(self):
...
-                makefile_conf.extend(
-                    [
-                        "LIBEXT=.dylib",
-                        "AR=%s -dynamiclib -Wl,-install_name -Wl,%s/$(notdir $@)"
-                        " -undefined dynamic_lookup %s -o "
-                        % (os.environ["FC"], prefix.lib, inject_libs),
-                        "RANLIB=echo",
-                    ]
-                )
-            else:
```
jcortial-safran, line 351: I think this final 'else' was removed by mistake.

jcortial-safran (changes requested):

jcortial-safran (changes requested):
It seems like the new code can be factored. The only difference I see is FC / mpifc.

jcortial-safran:
@spackbot fix style

jcortial-safran (changes requested):
It seems like the code can be factored. The only difference I see is FC / mpifc.

jcortial-safran:
@spackbot re-run pipeline

PR #5697, opened 2026-07-22
setuptools: deprecate versions with CVEs

adamjstewart:
This should test the limits of what we're willing to deprecate vs. ignore. I highly doubt most of CI will concretize without older setuptools.


adamjstewart:
It seems that Spack is not yet ready for this, there are just too many packages that still require ancient setuptools.

PR #5699, opened 2026-07-22
require patchelf at 'which'-time to improve errors

sethrj:
This improves a couple of packages that require `patchelf` to fix up library rpaths. Without `required=True` the `which` command will return None (causing `TypeError` later) or for older spack versions just fail with a less clear error message.

Inspired by an `external` for patchelf that pointed to `/usr` but actually wasn't there.


adamjstewart (approved):
Should probably be done for most `which` uses. I wonder if we should make this the default instead?

alecbcs (approved):
Looks good to me. Thanks!

gartung (approved):

alecbcs:
@spackbot run pipeline

alecbcs (approved):

PR #5701, opened 2026-07-22
py-parameterized: add v0.8.1, v0.9.0

haraldkl:
The py-parameterized version 0.7.1 seems to depend on an old setuptools version that is not available in the builtin spack packages anymore. But there are newer versions available that also work with newer setuptools. This MR drops 0.7.1 and adds 0.8.1 and 0.9.0 with updated dependencies.

### DIFF
--- a/repos/spack_repo/builtin/packages/py_parameterized/package.py
+++ b/repos/spack_repo/builtin/packages/py_parameterized/package.py
@@ -11,14 +11,13 @@ class PyParameterized(PythonPackage):
     """Parameterized testing with any Python test framework."""
 
     homepage = "https://github.com/wolever/parameterized"
-    pypi = "parameterized/parameterized-0.7.1.tar.gz"
+    pypi = "parameterized/parameterized-0.8.1.tar.gz"
 
     version("0.9.0", sha256="7fc905272cefa4f364c1a3429cbbe9c0f98b793988efb5bf90aac80f08db09b1")
     version("0.8.1", sha256="41bbff37d6186430f77f900d777e5bb6a24928a1c46fb1de692f8b52b8833b5c")
-    version("0.7.1", sha256="6a94dbea30c6abde99fd4c2f2042c1bf7f980e48908bf92ead62394f93cf57ed")
 
-    depends_on("py-setuptools", type="build")
+    depends_on("py-setuptools@57:", type="build")
 
-    with when("@0.9.0"):
+    with when("@0.9.0:"):
         depends_on("python@3.7:")
         depends_on("py-setuptools@61.2:", type="build")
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/py_parameterized/package.py
```diff
@@ -11,8 +11,13 @@ class PyParameterized(PythonPackage):
     """Parameterized testing with any Python test framework."""
 
     homepage = "https://github.com/wolever/parameterized"
-    pypi = "parameterized/parameterized-0.7.1.tar.gz"
+    pypi = "parameterized/parameterized-0.8.1.tar.gz"
 
-    version("0.7.1", sha256="6a94dbea30c6abde99fd4c2f2042c1bf7f980e48908bf92ead62394f93cf57ed")
```
RikkiButler20, line 16: Why did you remove `v0.7.1`

repos/spack_repo/builtin/packages/py_parameterized/package.py
```diff
@@ -11,8 +11,13 @@ class PyParameterized(PythonPackage):
...
+    version("0.9.0", sha256="7fc905272cefa4f364c1a3429cbbe9c0f98b793988efb5bf90aac80f08db09b1")
+    version("0.8.1", sha256="41bbff37d6186430f77f900d777e5bb6a24928a1c46fb1de692f8b52b8833b5c")
 
-    depends_on("py-setuptools", type="build")
+    depends_on("py-setuptools@57:", type="build")
```
RikkiButler20, line 19: Where did you find the version constraint for `py-setuptools` at v0.8.1? I couldn't find it.

repos/spack_repo/builtin/packages/py_parameterized/package.py
haraldkl: It requires an older setuptools, then what is available in the spack packages repository. Not sure, which one, but I tried the oldest available one in the setuptools package, and this results in an error (same error as with the latest setuptools version) during installation.

repos/spack_repo/builtin/packages/py_parameterized/package.py
haraldkl: By trying to install it with different setuptools versions. This is the earliest, that did not give me an error during installation. Older versions of setuptools resulted in failed installations for me.

haraldkl: No, actually it is just the oldest version I validated to work, as it is the oldest in the setuptools package, can't say if older versions would work or not.

RikkiButler20:
@spackbot re-run pipeline

RikkiButler20 (approved):
Looks good.

PR #5704, opened 2026-07-22
coolprop: add v8.0.0

staneuski:
CoolProp is a thermophysical property database and wrappers for a selection of programming environments, read more here:
- https://coolprop.org/
- https://github.com/CoolProp/CoolProp

### DIFF
--- a/repos/spack_repo/builtin/packages/coolprop/package.py
+++ b/repos/spack_repo/builtin/packages/coolprop/package.py
@@ -1,0 +1,93 @@
+# Copyright Spack Project Developers. See COPYRIGHT file for details.
+#
+# SPDX-License-Identifier: (Apache-2.0 OR MIT)
+import glob
+import os
+
+from spack_repo.builtin.build_systems.cmake import CMakePackage
+
+from spack.package import *
+
+
+class Coolprop(CMakePackage):
+    """CoolProp is a C++ library that implements pure and pseudo-pure fluid
+    equations of state and transport properties for 122 components, as well
+    as mixture properties using high-accuracy Helmholtz-energy-based mixing
+    rules.
+    """
+
+    homepage = "https://coolprop.org/"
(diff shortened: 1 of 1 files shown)


johnwparent (commented):
I think you perhaps meant to submit this as a builtin package under  `repos/spack_repo/builtin` not as a test package.

johnwparent (approved):
Looks good, I think some very minor cleanup could be applied to the install_links method, but it's all just code org nits. Not going to block this on that.

repos/spack_repo/builtin/packages/coolprop/package.py
```diff
@@ -0,0 +1,83 @@
...
+    def install_links(self):
+        """CoolProp installs into shared_library/<system>/<bitness>bit/ (or the
+        equivalent static_library/... layout) rather than the usual lib/,
+        include/. Add the conventional symlinks so the package is usable like
+        any other spack-installed library.
+        """
+        with working_dir(self.prefix):
+            if os.path.isdir("shared_library"):
+                if not os.path.exists("lib"):
+                    symlink(os.path.join("shared_library", "Linux", "64bit"), "lib")
```
johnwparent: Just noticed this hardcodes linux, do you want this to build on Windows? (it does build, I just built it). if so, please re-write this method, if not, please conflict with `platform=windows` (unless "linux" as a directory is created on all platforms...)

Further this will error with: 
```
spack.util.filesystem.SymlinkError: The source path (shared_library\Linux\64bit) is not relative to the link path (lib). Resulting link would be broken so not making link.
```

On windows, likely due to the fact that the link paths are both relative and our logic can't prove they share a relative path base.

johnwparent (changes requested):

johnwparent:
Sorry for the approval whiplash, I noticed the "linux" components after I hit submit and then tried to build it on Windows instead of mac.

johnwparent (approved):
Looks good and confirmed builds on Windows + Mac.

PR #5706, opened 2026-07-22
herwig3: add CT14(?n)lo pdfsets as build-time resources

wdconinc:
This PR modifies `herwig3` to not require `lhapdfsets` as a build dependency, but to pull in the specific pdf sets it requires (CT14(?n)lo) as resources. This avoids imposing requirements on `lhapdfsets` that are not actually needed (or desired) at build time. The resources downloaded as pdf sets are not installed, and only serve to satisfy the installation which uses them as defaults.

Needs:
- [x] https://github.com/spack/spack-packages/pull/5718


wdconinc:
This now installs fine in CI as well (now that `herwig3` has been added to the HEP stack), demonstrating that the approach with resources works fine and they don't need to be placed elsewhere as copilot claims.

bernhardkaindl (approved):

PR #5707, opened 2026-07-22
py-torch: honor disabled CMake features in 2.12.0:2.13.0

mathomp4:
## AI Disclosure

GPT 5.6-Terra helped me a lot with diagnosing and creating this patch (as well as the upstream issue)

## Summary

~Upstream report: https://github.com/pytorch/pytorch/issues/190800~ Fixed in upstream `main`

PyTorch 2.12.0 to 2.13.0 does not reliably preserve selected `USE_*` environment variables in its CMake cache when configured with Intel oneAPI. In particular, `USE_XPU=OFF`, `USE_KINETO=OFF`, and `USE_MKLDNN=OFF` can become `ON`.

This enables Kineto's XPU/PTI configuration unexpectedly and causes configuration to fail when PTI is not installed.

Add a patch that records the Spack-selected feature values in the CMake cache before PyTorch's later option defaults are evaluated.

This issue was observed when building PyTorch 2.12.0 with Intel oneAPI: PyTorch ignored Spack’s disabled `USE_XPU`, `USE_KINETO`, and `USE_MKLDNN` environment selections, enabled Kineto’s XPU/PTI path, and failed to configure without PTI.

The patch preserves the feature selections already exported by the Spack recipe in CMake’s cache. It does not enable a new backend or otherwise change the selected variants.

### Version Note

An analysis by GPT 5.6 Terra shows only v2.12.0 to 2.13.0 have a need for this patch.

  | PyTorch version | Forwarding implementation | Patch needed? |
  |---|---|---|
  | 2.10.0, 2.11.0 | `setup_helpers/cmake.py` collects `USE_*` variables and passes them as `-D` arguments | No |
  | 2.12.0, 2.12.1, 2.13.0 | `cmake/EnvVarForwarding.cmake`; identical older implementation | Yes |
  | current `main` | Fixed by https://github.com/pytorch/pytorch/pull/188242 | No |

## Testing

- Built `py-torch@2.12.0` with Intel oneAPI 2024.2.
- Verified the CMake cache preserves disabled XPU, Kineto, and MKLDNN settings.
- Verified configuration no longer enters Kineto's XPU/PTI path.


adamjstewart:
Excuse my ignorance, but what is CMakeCache and why should we care?

mathomp4:
> Excuse my ignorance, but what is CMakeCache and why should we care?

Ah. `CMakeCache.txt` is CMake’s generated "state" for a build directory. When you run CMake (as in `cmake -B build ...`) that file records the resolved value of, well, everything. In the earth system model I support, a *lot*:

```
> wc -l CMakeCache.txt
2263 CMakeCache.txt
```

In there are things like:

```cmake
//CXX compiler
CMAKE_CXX_COMPILER:FILEPATH=/usr/local/intel/oneapi/2024/compiler/2024.2/bin/icpx
...
//esmf.mk file found
ESMF_FOUND:BOOL=TRUE
...
//Use CODATA 2018 Constants
USE_CODATA_2018_CONSTANTS:BOOL=OFF

//Use ExtData2G
USE_EXTDATA2G:BOOL=ON
...
```

Those last one are like the `USE_XPU`, `USE_KINETO`, and `USE_MKLDNN` options this patch is trying to target. Those values are *cached* and are what CMake actually uses when it does its stuff. 

Now, for `py-torch`, the Spack recipe *deliberately* sets `USE_XPU=OFF`:

https://github.com/spack/spack-packages/blob/21903c4c7ed734fdb8899858cee76ae5f6c6eb21/repos/spack_repo/builtin/packages/py_torch/package.py#L702-L703

and also has:

https://github.com/spack/spack-packages/blob/21903c4c7ed734fdb8899858cee76ae5f6c6eb21/repos/spack_repo/builtin/packages/py_torch/package.py#L753

https://github.com/spack/spack-packages/blob/21903c4c7ed734fdb8899858cee76ae5f6c6eb21/repos/spack_repo/builtin/packages/py_torch/package.py#L788

and I want to build as:

```
 -   zvfyrzb      ^py-torch@2.12.0~caffe2~cuda
  +custom-protobuf~debug+distributed+fbgemm~gloo~kineto~metal
  ~mkldnn+mpi+nnpack+numa+numpy+openmp+qnnpack~rocm+tensorpipe
  ~test~ucc~valgrind+xnnpack build_system=python_pip
  commit=0d62256a2b23365f8e1604297eb23a6545102aa8 
  patches:=176c062,f3649e6,f4cd8de platform=linux 
  os=sles15 target=zen3 %c,cxx=oneapi@2024.2.0
```

So the Spack recipe says `USE_XPU=OFF` and I have `~kineto` and `~mkldnn` and it's like the `py-torch` CMake setup doesn't seem to honor those as I detail in https://github.com/pytorch/pytorch/issues/190800.

Then my reading of `py-torch` is that the recipe "re-defines" `enable_or_disable`:

https://github.com/spack/spack-packages/blob/21903c4c7ed734fdb8899858cee76ae5f6c6eb21/repos/spack_repo/builtin/packages/py_torch/package.py#L675-L690

And my reading of that is that `enable_or_disable(foo)` is translated to either `USE_FOO=ON` or `USE_FOO=OFF` depending on `+foo` or `~foo`.

Thus, the explicit setting of `USE_XPU=OFF` and the two `enable_or_disable` settings all should result in:
```
USE_XPU=OFF USE_MKLDNN=OFF USE_KINETO=OFF
```

Spack says "we want `USE_XPU=OFF`" but the `py-torch` CMake doesn't seem to correctly forward that through so its CMake says "You are using Intel oneAPI, so I'm turning on XPU support" and that would be "okay" I guess if I actually had any Intel GPUs or whatever. But the oneAPI install on my cluster has none so we don't even install the XPU/PTI/whatever bits that PyTorch's Kineto is using.

In the pytorch issue, I show that even if you explicitly set:
```
USE_XPU=OFF
USE_KINETO=OFF
USE_MKLDNN=OFF
```
in the environment, it doesn't get passed down into CMake. CMake still sets all those on due to its own internal rules, etc. My patch instead sort of *forces* CMakeCache to set those to whatever the environment says, which should override CMake.

As GPT 5.6 Terra said to me when I asked:

> The patch is not intended to make the cache an externally visible interface. It ensures the values PyTorch already receives from the environment are established as CMake options early enough that PyTorch’s later default option() declarations do not replace them. The cache is simply the direct, inspectable evidence that the requested configuration was lost.

adamjstewart:
Yeah, we set env vars, and PyTorch's setup.py does some weird logic where env vars are translated to cmake args during the call. It's possible something is going wrong there. Can you also open a PR with PyTorch and tag me? Would be nice to know what the PyTorch maintainers think of the proposed fix before making a decision within Spack.

mathomp4:
Ahh an update. I think PyTorch `main` has since implemented the general fix in https://github.com/pytorch/pytorch/pull/188242 by @xinyazhang (3 days ago!)

My reading is it now forwards all `USE_*`/`BUILD_*` environment variables into the CMake cache with `FORCE`, specifically to prevent option defaults or an earlier cache from shadowing the environment. 

And an analysis by GPT 5.6 Terra says:

  | PyTorch version | Forwarding implementation | Patch needed? |
  |---|---|---|
  | 2.10.0, 2.11.0 | `setup_helpers/cmake.py` collects `USE_*` variables and passes them as `-D` arguments | No |
  | 2.12.0, 2.12.1, 2.13.0 | `cmake/EnvVarForwarding.cmake`; identical older implementation | Yes |
  | current `main` | Fixed by https://github.com/pytorch/pytorch/pull/188242 | No |

I've updated this proposed patch to only target 2.12 to 2.13 as any newer versions should work.

I've also drafted this PR as I'm waiting to run my test with PyTorch `main`. Unfortunately, it takes about 1h30m for some reason to fully clone PyTorch and all its submodules on my cluster where I have oneAPI. Hopefully I can confirm.

NOTE: It's possible there are more things from `enable_or_disable` that might not work as expected then in v2.12 and v2.13. I'm not sure what a good general solution would be. Well, other than just "en masse" overwriting the `cmake/EnvVarForwarding.cmake` with the one by @xinyazhang from https://github.com/pytorch/pytorch/pull/188242

adamjstewart:
Could we copy the subset of the patch from main and use it instead of your patch?

mathomp4:
> Could we copy the subset of the patch from main and use it instead of your patch?

Hmm... that might be best, I suppose. Let me try that out locally in my 2.12 clone of pytorch.

mathomp4:
So, with the clone of PyTorch, before:

```
> grep -E '^(USE_XPU|USE_KINETO|USE_MKLDNN):' build/CMakeCache.txt
USE_KINETO:BOOL=ON
USE_MKLDNN:BOOL=ON
USE_XPU:BOOL=ON
```

and after the patch:

```
> grep -E '^(USE_XPU|USE_KINETO|USE_MKLDNN):' build/CMakeCache.txt
USE_KINETO:STRING=OFF
USE_MKLDNN:BOOL=OFF
USE_XPU:STRING=OFF
```

Nice. I'm now doing tests in spack. So fun 20 minute build...back soon.

mathomp4:
@adamjstewart Okay. I think this is good. I mean, it builds on my mac so the patch is at least not malformed.

I'm submitted build with oneAPI again on my cluster, but that might be a while. It's in the SLURM queue so it could complete in one hour or ten.

adamjstewart (approved):
Thanks for the fix!

Issue #1477, opened 2025-09-15
Installation issue: py-numpy@1.26.3 with Intel oneAPI (icx) 2025.2.1


climbfuji:
### Steps to reproduce the issue

There is a regression in Intel oneAPI 2025.2.1 that prevents building the auto-generated dispatch code in py-numpy. I will create a ticket with Intel to look into this, potentially this is related to a regression discovered when building py-scipy with Intel oneAPI 2025.2.x (and/or 2025.1.y, I need to narrow it down) that, according to the Inte compiler team, will be fixed in Intel oneAPI 2026.

Note: The build system chain spack -> pip -> toml -> meson is astonishingly complex to understand or modify.

```console
$ spack spec -l py-numpy@1.26.4
 -   o4ftduw  py-numpy@1.26.4 build_system=python_pip patches:=873745d arch=linux-almalinux9-x86_64_v3 %c,cxx=oneapi@2025.2.1
[+]  i6k7ldz      ^compiler-wrapper@1.0 build_system=generic arch=linux-almalinux9-x86_64_v3
[e]  4ywflif      ^glibc@2.34 build_system=autotools arch=linux-almalinux9-x86_64_v3
[e]  5zsheof      ^intel-oneapi-compilers@2025.2.1~amd+envmods~nvidia build_system=generic arch=linux-almalinux9-x86_64_v3
[e]  g5tylao      ^intel-oneapi-mkl@2025.2~cluster+envmods~gfortran~ilp64+shared build_system=generic mpi_family=none threads=none arch=linux-almalinux9-x86_64_v3
[+]  35at7gy      ^intel-oneapi-runtime@2025.2.1 build_system=generic arch=linux-almalinux9-x86_64_v3
[e]  svye46d          ^gcc-runtime@11.5.0 build_system=generic arch=linux-almalinux9-x86_64_v3
[+]  3yqagxy      ^py-cython@3.0.12 build_system=python_pip arch=linux-almalinux9-x86_64_v3 %c,cxx=oneapi@2025.2.1
[+]  ncmmc47          ^py-setuptools@69.2.0 build_system=generic arch=linux-almalinux9-x86_64_v3
[+]  wvkzmzd      ^py-meson-python@0.16.0 build_system=python_pip arch=linux-almalinux9-x86_64_v3 %c=oneapi@2025.2.1
[+]  twukedn          ^meson@1.7.0 build_system=python_pip patches:=0f0b1bd arch=linux-almalinux9-x86_64_v3
[...]

PR #5711, opened 2026-07-22
add numpy oneapi2025 fix

gberg617:
This PR attempts to fix build issues with py-numpy@1.26 using Intel's OneAPI 2025 compiler (see https://github.com/spack/spack-packages/issues/1477).  The existing workaround patches Meson's build file to lower the compiler optimization level to -O1. However, this doesn't remove optimization flags that are specified in the Spack environment. The proposed fix here manually injects -O1 in flag_handler and removes other optimization flags that were previously set, if they exist.

### DIFF
--- a/repos/spack_repo/builtin/packages/py_numpy/package.py
+++ b/repos/spack_repo/builtin/packages/py_numpy/package.py
@@ -266,6 +266,14 @@ class PyNumpy(PythonPackage):
             if gcc_version <= Version("5.1"):
                 flags.append(self.compiler.c99_flag)
 
+        if self.spec.satisfies("@1.26 ^intel-oneapi-compilers@2025.2") and name in (
+            "cflags",
+            "cxxflags",
+            "fflags",
+        ):
+            flags = [flag for flag in flags if not flag.startswith("-O")]
+            flags.append("-O1")
+
         return (flags, None, None)
 
     def _blas_lapack_pkg_config_mkl(self, spec) -> str:


rgommers (commented):
> The proposed fix here manually injects -O1 in flag_handler and removes other optimization flags that were previously set, if they exist.

Is that a reasonable thing to do? If I as a user explicitly give specific flags to override the defaults, I probably don't want them silently ignored. Rather get an error, or at least a warning to tell me I'm getting `-O1` rather than what I asked for, so I can decide to use either a more recent version or a different/better compiler toolchain.

The patch is harmless to other compilers, so if this seems fine to @adamjstewart it's fine with me. It's more a "how does Spack do this usually" question.

adamjstewart:
I'm also not sure. @becker33 @alalazo?

becker33:
It might be worth showing a warning, but I think this is generally good (for two primary reasons).

1. When users write `spack install foo cflags=-O3` they *usually* don't care specifically about the optimizations provided by `-O3` in a particular compiler, they just want it to be as optimized as they can easily get it.
2. Users often add optimization flags to their compiler configuration. This is really the key use case for something like this: If the user's compiler description or their toolchain adds compiler flags, they'll want to be able to apply that compiler or toolchain to numpy even though it has this bug.

The `flag_handler` method was originally written primarily to allow for passing the flags to the build system to allow the build system to back off on individual files that don't support higher optimization levels. This seems analogous.

adamjstewart (approved):

adamjstewart:
@spackbot run pipeline

PR #5309, opened 2026-06-23
hipblaslt: Support air-gapped build

williampiat3:
I believe this one is not in another PR but I might be mistaken
This needs PR #5291 to be completed first that's why it is a draft.

PR #5715, opened 2026-07-23
umpire: raise fmt upper bound to 12.1 for @2024.02.0:2025

zackgalbreath:
PR #5309 added the following dependency between hipblaslt and fmt:

```
depends_on("fmt@11.1:", when="@7.0:")
```

This conflicted with umpire@2024:'s constraint on fmt@9.1:11.0. As a consequence of this, the concretizer fell back to umpire@2023.06.0 for the E4S stack because this version doesn't declare a dependency on fmt at all.


johnwparent (approved):
LGTM

PR #5724, opened 2026-07-23
omm: Added v1.4.1, v1.5.0. Fixed dependencies and dependency detection.

emwjacobson:
This is a very similar fix to #5678 where dependencies weren't properly being detected.

Also bumped the versions while I was at it, and noticed some compilation problems with different versions of omm and matrix-switch. I locked the versions of omm to some specific matrix-switch versions where needed.

Listed maintainer is @RMeli


RMeli (approved):

RMeli:
Thanks @emwjacobson, I haven't tested/updated the ELSI ecosystem in a while.

PR #5727, opened 2026-07-23
Python: fix bad nuget link and improve python detection

johnwparent:
Python (like many windows builds) decided to require another version of itself to build itself. If it cannot find an existing system python (it uses the python launcher to do this, so if you didn't explicitly click on that part of the installer, easy to not have), it downloads one with nuget.

Python 3.14 had a broken nuget link (seems like msft changed it on them), patch in the correct link from the upstream python PR.

Additionally, set the env variable to Spack's python to prevent this download in the first place.

### DIFF
--- a/repos/spack_repo/builtin/packages/python/package.py
+++ b/repos/spack_repo/builtin/packages/python/package.py
@@ -251,6 +251,12 @@ class Python(Package):
         depends_on("libxcrypt", when="+crypt")
 
     patch(
+        "https://github.com/python/cpython/commit/106eb532ea3b243423e62a702719e9d3c0e40c16.patch?full_index=1",
+        sha256="d01422e6c6cb1df21161f5fa261cfd12853756abbddf816543934f3fe96a5827",
+        when="@3.14:3.14.4 platform=windows"
+        )
+
+    patch(
         "https://bugs.python.org/file44413/alignment.patch",
         when="@3.6",
         sha256="d39bacde16128f380933992ea7f237ac8f70f9cdffb40c051aca3be46dc29bdf",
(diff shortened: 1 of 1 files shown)


adamjstewart (commented):
Is the patch needed if this env var is set? The only thing I dislike more than using sys.executable is breaking air-gapped installs, so I would prefer the env var if both aren't needed.

johnwparent:
> Is the patch needed if this env var is set? The only thing I dislike more than using sys.executable is breaking air-gapped installs, so I would prefer the env var if both aren't needed.

The code the patch is resolving is not executed if the env var is set. That said, IMO, the patch is still good because currently the URL being patched points to a dead link with unknown redirects and the resulting download is unconditionally executed by the build system. Perhaps I'm just being paranoid, but I'd rather avoid the random dead url, even if we're not going to use it.

(the current behavior without the env var is to download, and attempt to execute, the bing.com home page)

> I dislike more than using sys.executable

Can you elaborate? I can't recall sys.executable being regarded as a smell in general, but it is weird in Spack. You're definitely the python expert here, is this a general python bad practice, a spack issue or just a personal preference? I'm happy to avoid it,  I'm just trying to learn more.

johnwparent:
Python on windows is currently building a number of external packages it fetches from the internet (https://github.com/python/cpython-source-deps).
I'm working changing this for spack in a local branch, but it probably won't be ready for a bit, and this unblocks building python 3.14 on Windows in the short term.

adamjstewart:
> The code the patch is resolving is not executed if the env var is set. That said, IMO, the patch is still good because currently the URL being patched points to a dead link with unknown redirects and the resulting download is unconditionally executed by the build system. Perhaps I'm just being paranoid, but I'd rather avoid the random dead url, even if we're not going to use it.

I guess I'm still leaning towards removing just to keep the package.py simpler, especially since it's been fixed upstream already. These are only build tools right, they don't get installed?

> I can't recall sys.executable being regarded as a smell in general

This has come up in a few discussions in Spack, usually around trying to resolve cyclic dependencies (libxml2 I think) and reproducibility of builds in general. It's generally considered bad practice and we try to avoid it.

johnwparent:
> I guess I'm still leaning towards removing just to keep the package.py simpler, especially since it's been fixed upstream already. These are only build tools right, they don't get installed?

No, python's build downloads dependencies it builds and links against on Windows, a few of which we can't actually build with Spack on Windows right now.

But fair enough re:removing the line in the package recipe

nuget itself is just another package manager.

>> sys.executable
> It's generally considered bad practice and we try to avoid it.

Well, good to know its generally considered bad practice at least, thanks

PR #5729, opened 2026-07-24
julia 1.12: Fix LLVM's shared library symbol version

RemiLacroix-IDRIS:
Julia 1.12 depends on LLVM 18.1.8 and the Julia ecosystem expects the LLVM's shared library symbol version to be "major.minor" (see https://github.com/JuliaLang/julia/blob/v1.12.6/deps/llvm-ver.make#L20).


RemiLacroix-IDRIS:
Note that this caused some errors like
```
/.../llvm/18.1.8/gcc-14.2.0-cgghm22mn6bzoqvwuf6tm4rs3lp4b6hq/lib/libLLVM.so.18.1jl: version `JL_LLVM_18.1' not found
```
when installing some Julia packages.

giordano (approved):

bernhardkaindl (approved):

Issue #5738, opened 2026-07-24
Installation issue: netcdf-c build_system=cmake libtirpc issue


payerle:
### Steps to reproduce the issue

I am running into errors when linking shared libs during the build of netcdf-d:

I am trying to build
```console
$ spack --env tptest find -lvfdcI netcdf-c
...
-- linux-rhel8-zen2 / %c,cxx=gcc@13.4.0 -------------------------
 -   i2n3vrs netcdf-c@4.10.0+blosc+byterange+dap~fsync+hdf4+ipo~jna~logging~mpi+nczarr_zip+optimize~parallel-netcdf+pic+shared+szip+zstd build_system=cmake build_type=RelWithDebInfo generator=make patches:=9a22f9f
[+]  clcmbca     bzip2@1.0.8~debug+pic+shared build_system=generic
[+]  aveg6xi         diffutils@3.12 build_system=autotools
[+]  g5qgddj     c-blosc@1.21.6+avx2+ipo build_system=cmake build_type=RelWithDebInfo generator=make
[+]  rv4gmuf         lz4@1.10.0+pic build_system=makefile libs:=shared,static
[+]  utuggfs         snappy@1.2.2+ipo+pic+shared build_system=cmake build_type=RelWithDebInfo generator=make
[+]  komdk4j     cmake@3.31.9~doc+ncurses+ownlibs~qtgui build_system=generic build_type=RelWithDebInfo
[+]  4fr2zga         ncurses@6.5-20250705+symlinks+termlib abi=5 build_system=autotools patches:=7a351bc
[+]  c4lvqa5     compiler-wrapper@1.0 build_system=generic
[+]  5iwsjnc     curl@8.15.0+gssapi~ldap+libidn2~librtmp~libssh+libssh2+nghttp2 build_system=autotools libs:=shared,static tls:=mbedtls,openssl
[+]  vjvgdcm         krb5@1.21.3+shared build_system=autotools
[+]  rcuuwdi             findutils@4.10.0 build_system=autotools patches:=440b954
[+]  un24op4             gettext@0.23.1+bzip2+curses+git+libunistring+libxml2+pic+shared+tar+xz build_system=autotools
[+]  ac3qlxk                 tar@1.35 build_system=autotools zip=pigz
[+]  xdqbnzg                     pigz@2.8 build_system=makefile
[+]  p2qbo6v             libedit@3.1-20240808 build_system=autotools
[+]  cwtascr         libidn2@2.3.7 build_system=autotools
[...]

PR #5739, opened 2026-07-24
netcdf-c: Fix linking with hdf~shared+external-xdr^libtirpc when build_sytem=cmake

payerle:
This seems to fix issue #5738

This is a somewhat ugly hack
I have tried to mimic the "hack" in AutotoolsBuilder by adding the appropriate linker flags to link to libtirpc.so.  Some of this is hardcoded, but I am not sure how to generalize.  In particular, libtirpc appears to be the only current provider of the "rpc" virtual, and something like rpclib does not include XDR routines.

@spackbot fix style

### DIFF
--- a/repos/spack_repo/builtin/packages/netcdf_c/package.py
+++ b/repos/spack_repo/builtin/packages/netcdf_c/package.py
@@ -473,21 +473,13 @@ class CMakeBuilder(AnyBuilder, cmake.CMakeBuilder):
                 self.define("PLUGIN_INSTALL_DIR", pathlib.Path(self.prefix.plugins).as_posix())
             )
 
-        linker_flags = []
-        if "+hdf4" in self.spec:
-            hdf = self.spec["hdf"]
-            if "~shared" in hdf:
-                if "+external-xdr ^libtirpc" in hdf:
-                    # This is somewhat ugly
-                    libdir = hdf["rpc"].prefix.lib
-                    lib = "tirpc"
-                    linker_flags.append(f"-L{libdir}")
-                    linker_flags.append(f"-l{lib}")
-
-        if linker_flags:
-            lflagstr = " ".join(linker_flags)
-            base_cmake_args.append(self.define("CMAKE_EXE_LINKER_FLAGS", lflagstr))
-            base_cmake_args.append(self.define("CMAKE_MODULE_LINKER_FLAGS", lflagstr))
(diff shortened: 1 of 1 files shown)


payerle:
@spackbot fix style

bernhardkaindl:
I guess something based on `spack install netcdf-c+hdf4 hdf+external-xdr~shared` might reproduce it too.

@payerle, can you test this version of the fix (it should have add the same effect), according to my review, it is a non-functional change and converts it to user the current spack `satisfies()` API:

```py
        # With static hdf+external-xdr we need to specify spack's libtirpc explicitly
        # to avoid undefined references to xdr functions. Fixes spack-packages#5738:
        if self.spec.satisfies("+hdf4"):
            if self.spec["hdf"].satisfies("~shared +external-xdr ^libtirpc"):
                tirpc = f"-L{self.spec['libtirpc'].prefix.lib} -ltirpc"
                base_cmake_args.append(self.define("CMAKE_EXE_LINKER_FLAGS", tirpc))
                base_cmake_args.append(self.define("CMAKE_MODULE_LINKER_FLAGS", tirpc))
```

- This assumes that `self.spec["hdf"].satisfies("~shared +external-xdr ^libtirpc")` works like the nested check.
- Single f-string "-L{...} -ltirpc" vs. two separate appends — same output

With the comment and the improved maintainability, it's not as ugly anymore.

These are the wonders of using static libraries if the build does not use additonal magic like `pkg-config` files giving it the needed linker flags for the other static libraries to link with. If the CMake build would properly use a CMake module / `pkg-config` or a (and I thin libtool's `.la` files) I think this would work without helping the packages like this. So, I'd call this just needing a little help.

bernhardkaindl:
@payerle, I can convert https://github.com/spack/spack-packages/pull/5739#issuecomment-5154968411 into a applyable suggestion, it only refactors the your change, but I'd leave that to you for now.

With it, I think this would be fine to approve.

payerle:
I have applied the suggested edits from @bernhardkaindl 

My comments re ugliness had to do with:
1) the "little help" you refered to, and
2) the fact that we are hard-coding to libtirpc, especially as rpc is technically a virtual package (even if spack only has a single provider currently)

bernhardkaindl:
@spackbot run pipeline

bernhardkaindl (approved):

PR #5741, opened 2026-07-24
hadoop-xrootd: add v1.0.10

jmcarcell:


repos/spack_repo/builtin/packages/hadoop_xrootd/package.py
```diff
@@ -10,13 +10,14 @@
 class HadoopXrootd(MavenPackage):
     """Connector between Hadoop and XRootD protocols (EOS compatible)."""
 
-    homepage = "https://gitlab.cern.ch/db/hadoop-xrootd"
+    homepage = "https://gitlab.cern.ch/hadoop/utils/hadoop-xrootd"
```
wdconinc, line 13: Is there no way to make this a public repo? It's harder to review.

jmcarcell: I asked for them to make it public; the sources are already public anyway through lcgpackages.

wdconinc: Thanks. It's easier to review based on a compare revisions on a gitlab server than pulling two tarballs.

repos/spack_repo/builtin/packages/hadoop_xrootd/package.py
```diff
@@ -10,13 +10,14 @@
...
     url = "https://lcgpackages.web.cern.ch/tarFiles/sources/hadoop-xrootd-v1.0.7.tar.gz"
 
     maintainers("haralmha")
 
     license("Apache-2.0")
 
+    version("1.0.10", sha256="1895bb3875ed03838de220e4c50d5519a4a8ef3b9c31da3be101c9aaa1b095fd")
```
wdconinc: Does this version have a corresponding tag? I see tags v1.0.7, but not v1.0.10, and instead there is 1.0.10-0. If there will be multiple 1.0.10 releases, maybe we need a 4th component to the version?

jmcarcell:
```suggestion
    version("1.0.10-0", sha256="1895bb3875ed03838de220e4c50d5519a4a8ef3b9c31da3be101c9aaa1b095fd")
```

jmcarcell: Added; I don't know if there will be more tags - I think it's not being developed that much, the previous tags being from 2024.

wdconinc (approved):

PR #5749, opened 2026-07-26
PEXSI: Allow OpenMP support for 2.0.0

Growl1234:
PEXSI 2.0.0 has also `PEXSI_ENABLE_OPENMP` option whose default is `OFF`, so enable it in Spack.

And, change default value of `fortran` to `True` as Spack recipe of PEXSI already requires Fortran compiler unconditionally.


bernhardkaindl (approved):

bernhardkaindl (commented):
@Growl1234, as a future suggestion: If `~fortran` does not need a fortran compiler (to be checked), this change could be made:
```diff
-        depends_on("fortran", type="build")  # generated
+        depends_on("fortran", type="build", when="+fortran")
```
Because this PR already passed GitLab CI, and I see not further suggestion that would be needed, I can leave this as a possible further streamlining of the recipe.

PR #5755, opened 2026-07-27
vasp: The +mpi variant of wannier90 3.x does not work with VASP

nboelte:
@snehring

### DIFF
--- a/repos/spack_repo/builtin/packages/vasp/package.py
+++ b/repos/spack_repo/builtin/packages/vasp/package.py
@@ -66,7 +66,7 @@ class Vasp(MakefilePackage, CudaPackage):
     depends_on("wannier90", when="+wannier90")
     # The +mpi variant of wannier90 3.x does not work with VASP
     # https://vasp.at/wiki/Makefile.include#Wannier90_(optional)
-    depends_on("wannier90~mpi", when="+wannier90 %wannier90@3:")
+    conflicts("^wannier90@3:+mpi", msg="The +mpi variant of wannier90 3.x does not work with VASP")
     # at the very least the nvhpc mpi seems required
     requires("^nvhpc+mpi+lapack+blas", when="%nvhpc")
 
--- a/repos/spack_repo/builtin/packages/vasp/package.py
+++ b/repos/spack_repo/builtin/packages/vasp/package.py
@@ -66,7 +66,7 @@ class Vasp(MakefilePackage, CudaPackage):
     depends_on("wannier90", when="+wannier90")
     # The +mpi variant of wannier90 3.x does not work with VASP
     # https://vasp.at/wiki/Makefile.include#Wannier90_(optional)
-    conflicts("^wannier90@3:+mpi")
+    depends_on("wannier90~mpi", when="+wannier90 %wannier90@3:")
     # at the very least the nvhpc mpi seems required
     requires("^nvhpc+mpi+lapack+blas", when="%nvhpc")
 
--- a/repos/spack_repo/builtin/packages/vasp/package.py
+++ b/repos/spack_repo/builtin/packages/vasp/package.py
@@ -63,7 +63,9 @@ class Vasp(MakefilePackage, CudaPackage):
     depends_on("hdf5+fortran+mpi", when="+hdf5")
     depends_on("libbeef", when="+libbeef")
     depends_on("libxc~fhc+fortran", when="+libxc")
-    depends_on("wannier90", when="+wannier90")
+    # The +mpi variant of wannier90 3.x does not work with VASP
+    # https://vasp.at/wiki/Makefile.include#Wannier90_(optional)
+    depends_on("wannier90~mpi", when="+wannier90")
     # at the very least the nvhpc mpi seems required
     requires("^nvhpc+mpi+lapack+blas", when="%nvhpc")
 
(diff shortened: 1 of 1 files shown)


snehring:
It'd be better to only restrict `wannier90~mpi` when `wannier90@3:`

repos/spack_repo/builtin/packages/vasp/package.py
```diff
@@ -63,7 +63,9 @@ class Vasp(MakefilePackage, CudaPackage):
     depends_on("hdf5+fortran+mpi", when="+hdf5")
     depends_on("libbeef", when="+libbeef")
     depends_on("libxc~fhc+fortran", when="+libxc")
-    depends_on("wannier90", when="+wannier90")
+    # The +mpi variant of wannier90 3.x does not work with VASP
+    # https://vasp.at/wiki/Makefile.include#Wannier90_(optional)
+    depends_on("wannier90~mpi", when="+wannier90")
```
bernhardkaindl: @snehring wrote:
> It'd be better to only restrict wannier90~mpi when wannier90@3:

I was about to suggest in effect the same to make the change match the justification:
```suggestion
    depends_on("wannier90~mpi", when="+wannier90@3:")
```

nboelte: I don't think that you can have a version number on the variant, but i can add a conflict maybe. I was thinking how to do this cleanly before but couldn't really figure out how to do it without a conflict.

nboelte:
@snehring I used a conflict instead to only restrict wannier90@3:

repos/spack_repo/builtin/packages/vasp/package.py
```diff
@@ -64,6 +64,9 @@ class Vasp(MakefilePackage, CudaPackage):
     depends_on("libbeef", when="+libbeef")
     depends_on("libxc~fhc+fortran", when="+libxc")
     depends_on("wannier90", when="+wannier90")
+    # The +mpi variant of wannier90 3.x does not work with VASP
+    # https://vasp.at/wiki/Makefile.include#Wannier90_(optional)
+    conflicts("^wannier90@3:+mpi")
```
snehring:
```suggestion
    depends_on("wannier90~mpi", when="+wannier90 %wannier90@3:")
```
see if that works for you

nboelte: That works great! Thank you!

snehring: ugh, the audit doesn't like it. I'll think about this a bit.

snehring: I'd really like to take care of this with a depends_on, but audit says no. So let's do the conflicts
```
conflicts("%wannier90@3:+mpi", msg="MPI enabled wannier90 3.x+ is not supported by vasp")
```
but we'll make sure to tell the user why it conflicts in the message.

repos/spack_repo/builtin/packages/vasp/package.py
```diff
@@ -64,6 +64,9 @@ class Vasp(MakefilePackage, CudaPackage):
     depends_on("libbeef", when="+libbeef")
     depends_on("libxc~fhc+fortran", when="+libxc")
     depends_on("wannier90", when="+wannier90")
+    # The +mpi variant of wannier90 3.x does not work with VASP
+    # https://vasp.at/wiki/Makefile.include#Wannier90_(optional)
+    depends_on("wannier90~mpi", when="+wannier90 %wannier90@3:")
```
bernhardkaindl: Spack's CI prechecks failed with `vasp: dependency on 'wannier90~mpi' when '+wannier90 %wannier90@3:' is self-referential`.

Besides the self-referential, normally, `%` is for compilers, so I don't know if that is correct: `%wannier90@3:`

I agree the best maybe only way way is to express this as a conflict, like the PR title says it, like like @nboelte wrote it, but there is room to add a `msg` too:

```suggestion
    conflicts("^wannier90@3:+mpi", msg="The +mpi variant of wannier90 3.x does not work with VASP")
```

snehring: My understanding is ```%``` denotes a direct dependency, whereas ```^``` denotes a dependency anywhere in the tree.

bernhardkaindl: > My understanding is % denotes a direct dependency, whereas ^ denotes a dependency anywhere in the tree.
https://spack.readthedocs.io/en/latest/spec_syntax.html#spec-syntax

Ack, I stand corrected. But with `spack -d audit packages vasp`, the `depends_on()` does not appear to work so far.

I was wrong in suggesting a `depends_on()` would be better. As said in the title "[vasp: The +mpi variant of wannier90 3.x does not work with VASP](https://github.com/spack/spack-packages/pull/5755#top)", this maps to a conflict directly.

nboelte: Yes that works. I have had bad experiences with conflicts in the past, but maybe the concretizer is better now.

snehring (approved):

snehring:
lgtm

bernhardkaindl (approved):
Ack, perfect - with `msg` and a comment with URL.

As snehring approved, and I see nothing further to improve here, I approve too.

PR #5760, opened 2026-07-28
libidn2: Fix intel classic build

nboelte:
This fixes a build error with the intel classic compiler:

> error #3895: expected a comma (the one-argument version of static_assert is not enabled in this mode)

### DIFF
--- a/repos/spack_repo/builtin/packages/libidn2/package.py
+++ b/repos/spack_repo/builtin/packages/libidn2/package.py
@@ -37,3 +37,9 @@ class Libidn2(AutotoolsPackage, GNUMirrorPackage):
 
     # in-source build fails
     build_directory = "spack-build"
+
+    def setup_build_environment(self, env: EnvironmentModifications) -> None:
+        super().setup_build_environment(env)
+
+        if self.spec.satisfies("%intel"):
+            env.append_flags("CFLAGS", "-std=c11")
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/libidn2/package.py
```diff
@@ -37,3 +37,9 @@ class Libidn2(AutotoolsPackage, GNUMirrorPackage):
 
     # in-source build fails
     build_directory = "spack-build"
+
+    def setup_build_environment(self, env: EnvironmentModifications) -> None:
+        super().setup_build_environment(env)
+
+        if self.spec.satisfies("%intel"):
+            env.append_flags("CFLAGS", "-std=c11")
```
bernhardkaindl: Using `git grep`, I find uses of the newer spack `flag_hander()` doing this I think (untested):
```suggestion
    def flag_handler(self, name, flags):
        if name == "cflags" and self.spec.satisfies("%intel"):
            flags.append("-std=c11")
        return (flags, None, None)
```
Documentation to check:
https://spack.readthedocs.io/en/latest/packaging_guide_build.html#compiler-flags

nboelte: Ah, i thought `flag_handler()` is only supported by some build systems. Thank you! I changed the PR accordingly :)

nboelte: (And i did test it using intel@2021.10.0)

bernhardkaindl (approved):

PR #5761, opened 2026-07-28
gettext: Fix intel classic build

nboelte:
This fixes a build error with the intel classic compiler:

> error https://github.com/spack/spack-packages/pull/3895: expected a comma (the one-argument version of static_assert is not enabled in this mode)

### DIFF
--- a/repos/spack_repo/builtin/packages/gettext/package.py
+++ b/repos/spack_repo/builtin/packages/gettext/package.py
@@ -100,6 +100,8 @@ class Gettext(AutotoolsPackage, GNUMirrorPackage):
         # this goes together with gl_cv_libxml_force_included=no
         if name == "ldflags" and self.spec.satisfies("+libxml2"):
             flags.append("-lxml2")
+        if name == "cflags" and self.spec.satisfies("%intel"):
+            flags.append("-std=c11")
         return (flags, None, None)
 
     @classmethod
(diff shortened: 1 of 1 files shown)


nboelte:
Sorry forgot to tag @michaelkuhn

repos/spack_repo/builtin/packages/gettext/package.py
```diff
@@ -184,3 +184,9 @@ def setup_dependent_build_environment(
         # In addition to prefix/share/aclocal, newer versions of gettext also
         # install m4 files into prefix/share/gettext/m4.
         env.append_path("ACLOCAL_PATH", self.prefix.share.gettext.m4)
+
+    def setup_build_environment(self, env: EnvironmentModifications) -> None:
+        super().setup_build_environment(env)
+
+        if self.spec.satisfies("%intel"):
+            env.append_flags("CFLAGS", "-std=c11")
```
michaelkuhn: I think it might be easier and cleaner to add this to the `flag_handler` method: https://github.com/spack/spack-packages/blob/develop/repos/spack_repo/builtin/packages/gettext/package.py#L99-L103

nboelte: Good point. I will test that now.

nboelte: It works fine using `flag_handler`, i changed the PR accordingly. Thank you for catching that, i was updating several packages and missed the existing method.

michaelkuhn (approved):
LGTM

PR #5764, opened 2026-07-28
openms: new package

w8jcik:
The package describes OpenMS versions 3.1 to 3.5, but also not yet released 3.6.

The part about 3.6 might be confusing.

As of now the dependency description for 3.6 is accurate and can be installed by referring to development branch

```sh
spack install openms@git.develop=3.6.0-dev+pyopenms
spack load openms@3.6.0-dev
```

or better with the current commit hash

```sh
spack install openms@git.e9ba7c27c03a4b89f93972bd7b927f84a132e492=3.6.0-e9ba7c2-dev+pyopenms
spack load openms@3.6.0-e9ba7c2-dev
```

I have included the description of 3.6 because I already need 3.6, and it is unlikely for the dependencies to change much. The authors create releases at long intervals, which is impractical (the last release was 7 months ago).

https://github.com/OpenMS/OpenMS/blob/develop/CHANGELOG

I will update the recipe when 3.6 is released, most likely just a version entry will need to be added.

The other versions 3.1 to 3.5 work without further complications.

### DIFF
--- a/repos/spack_repo/builtin/packages/openms/package.py
+++ b/repos/spack_repo/builtin/packages/openms/package.py
@@ -1,0 +1,140 @@
+# Copyright Spack Project Developers. See COPYRIGHT file for details.
+#
+# SPDX-License-Identifier: (Apache-2.0 OR MIT)
+
+import shutil
+
+from spack_repo.builtin.build_systems.cmake import CMakePackage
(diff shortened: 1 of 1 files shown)


bernhardkaindl (approved):
LGTM with maintainer.

PR #5776, opened 2026-07-28
paraview: add version 6.2.0-RC1

cquammen:


vicentebolea:
@cquammen lets make the previous final release as the default while we are still in RCs for the current release

cquammen:
> @cquammen lets make the previous final release as the default while we are still in RCs for the current release

Sounds good. Does that mean swapping the new line with the previous release line?

vicentebolea:
Nope just add the preferred=True param to the function call that you added

[image: logo shows large letter K with the word Kitware below it]
<http://www.kitware.com>

Vicente Bolea, MSc.

*Senior R&D Engineer*

cquammen:
> Nope just add the preferred=True param to the function call that you added

I think you meant make 6.1.1 be preferred while we are in the RC phase. I have done that in the latest push.

vicentebolea (approved):

RikkiButler20 (approved):

PR #5782, opened 2026-07-28
add cxxstd variant to camp

dtaller:
The most recent version of camp has concepts, a C++20 feature, so I added a variant to specify the version so I can build with spack.


dtaller:
@spackbot fix style

dtaller:
Converting to draft. I think the cxxstd fix may need to go into packages that use camp's header files, such as axom

PR #5790, opened 2026-07-29
r-rcpp: add 1.1.1-1.1, 1.1.2

emwjacobson:
Updates r-rcpp with 2 latest versions.

Versions up to 1.1.1 (_not_ including 1.1.1-1.1) have a conflict with r@4.6, so I added that as well despite r@4.6.0 not (yet) being added in Spack.


bernhardkaindl (approved):
@emwjacobson and @w8jcik, jfyi:

I built both new versions and checked rlib/R/library/Rcpp/NEWS.Rd of them is from the respective version.

PR #5791, opened 2026-07-29
axom: Add cxxstd variant for C++20, python variant improvements

bmhan12:
This PR:
- Adds `cxxstd` variant for C++20
- Marks `cpp14` variant as deprecated, no longer a hard requirement
  - Relevant  Discussion Comment: https://github.com/llnl/axom/pull/1912#discussion_r3573447174
- Improves python installation for `python` variant

Relates to:
- LLNL/axom#1912 (C++20 changes)
- LLNL/axom#1896 (python changes)

### DIFF
--- a/repos/spack_repo/builtin/packages/axom/package.py
+++ b/repos/spack_repo/builtin/packages/axom/package.py
@@ -168,7 +168,13 @@ class Axom(CachedCMakePackage, CudaPackage, ROCmPackage):
     variant(
         "cxxstd",
         default="20",
-        values=("11", "14", "17", "20"),
+        values=(
+            conditional("11", when="@:0.6.1"),
+            conditional("14", when="@:0.11.0"),
+            conditional("17", when="@:0.14.0"),
+            "20",
+        ),
+        multi=False,
         description="C++ standard to build with",
     )
 
@@ -344,13 +350,6 @@ class Axom(CachedCMakePackage, CudaPackage, ROCmPackage):
     # -----------------------------------------------------------------------
     # Conflicts
     # -----------------------------------------------------------------------
-
-    # C++14 required as of 0.6.2
-    conflicts("cxxstd=11", when="@0.6.2:")
-    # C++17 required as of 0.12.0
-    conflicts("cxxstd=14", when="@0.12.0:")
-    # C++20 required as of unreleased 0.15.0
-    conflicts("cxxstd=17", when="@0.15.0:")
 
     # Conduit's cmake config files moved and < 0.4.0 can't find it
     conflicts("^conduit@0.7.2:", when="@:0.4.0")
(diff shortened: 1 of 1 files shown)


dtaller:
Thanks! I appreciate it!

repos/spack_repo/builtin/packages/axom/package.py
```diff
@@ -150,20 +154,24 @@ class Axom(CachedCMakePackage, CudaPackage, ROCmPackage):
...
     variant("devtools", default=False, description=varmsg)
 
+    variant(
+        "cxxstd",
+        default="20",
+        values=("11", "14", "17", "20"),
+        description="C++ standard to build with",
+    )
+
     # -----------------------------------------------------------------------
```
dtaller, line 181: You may want to add conflicts if someone sets +cpp14 but also cxxstd=17 for example

bmhan12: The `cpp14` variant in its current deprecated state is a misnomer.
It is a no-op the majority of the time, and it is only relevant in a narrow case where you want to build with C++14 with axom <= 0.6.1 & with blt <= 0.5.1 (with or without CUDA).
I've left the variant in but defaulted to false in order to not break some environments in the shared environments that have `+cpp14` explicitly defined (configuration would break when spack determines variant no longer exists).

dtaller (approved):

repos/spack_repo/builtin/packages/axom/package.py
```diff
@@ -499,25 +525,41 @@ def initconfig_hardware_entries(self):
...
             if spec.satisfies("+fortran"):
-                link_remove_list = []
+                link_lib_remove_list = []
+                link_dir_remove_list = []
+
+                if self.cxx_std == "20":
+                    link_dir_remove_list += [
+                        "/opt/rh/gcc-toolset-12/root/usr/lib/gcc/x86_64-redhat-linux/12"
+                    ]
+                    link_dir_remove_list += ["/opt/rh/gcc-toolset-12/root/usr/lib64"]
```
becker33: Can you give me more context on what's going on here? I don't think we should need to hardcode a RHEL path to get this working in the general case.

repos/spack_repo/builtin/packages/axom/package.py
```diff
@@ -853,10 +916,61 @@ def test_install_using_cmake(self):
...
+                    import_path.append(conduit_py)
+
+        for dep in ("py-numpy", "py-mpi4py"):
+            if self.spec.satisfies("^{0}".format(dep)):
+                dep_py = join_path(self.spec[dep].prefix, python_platlib)
+                if os.path.isdir(dep_py):
+                    import_path.append(dep_py)
+
+        imports = "import axom.sidre as s; import numpy; print('axom.sidre', s.__version__)"
+        python = Executable(join_path(self.spec["python"].prefix.bin, "python3"))
```
becker33:
```suggestion
        python = self["python"].command
```

becker33 (changes requested):

repos/spack_repo/builtin/packages/axom/package.py
bmhan12: > Can you give me more context on what's going on here?

@becker33 - Removed the hardcoded path logic. 
Found a better way to handle the issue this code snippet was trying to solve.

The issue was fortran was injecting the wrong versioned gcc-toolset link path.
It was injecting the path above, instead of version 13 that was correctly provided by C++ and C.
The better solution is to define `fflags: --gcc-toolchain=/opt/rh/gcc-toolset-13/root/usr`in the spack enviroment file, removing the need for the `BLT_CMAKE_IMPLICIT_LINK_DIRECTORIES_EXCLUDE` logic: https://github.com/llnl/axom/pull/1949#discussion_r3770872846

becker33 (approved):

PR #5792, opened 2026-07-29
omega-h: add v11.2.0-scorec and a python variant

Sichao25:
This PR adds Omega_h version 11.2.0-scorec. It also supports python variant from now.

### DIFF
--- a/repos/spack_repo/builtin/packages/omega_h/package.py
+++ b/repos/spack_repo/builtin/packages/omega_h/package.py
@@ -92,7 +92,9 @@ class OmegaH(CMakePackage, CudaPackage):
     variant("gmsh", default=False, description="Use Gmsh C++ API")
     variant("kokkos", default=False, description="Use Kokkos")
     variant("cuda", default=False, description="Enable CUDA backend", when="@:10.10.0")
-    variant("python", default=False, description="enable python interfaces", when="@11.2.0-scorec:")
+    variant(
+        "python", default=False, description="enable python interfaces", when="@11.2.0-scorec:"
+    )
 
     depends_on("cxx", type="build")
     depends_on("c", type="build", when="+mpi")
--- a/repos/spack_repo/builtin/packages/omega_h/package.py
+++ b/repos/spack_repo/builtin/packages/omega_h/package.py
@@ -77,7 +77,11 @@ class OmegaH(CMakePackage, CudaPackage):
     variant("shared", default=True, description="Build shared libraries")
     variant("mpi", default=True, description="Activates MPI support")
     variant("zlib", default=True, description="Activates ZLib support")
-    variant("trilinos", default=False, description="Use Kokkos and SEACASExodus from trilinos")
+    variant(
+        "trilinos",
+        default=False,
+        description="Use Kokkos and SEACASExodus from trilinos",
+    )
     variant(
         "exodus",
         default=False,
@@ -93,7 +97,10 @@ class OmegaH(CMakePackage, CudaPackage):
     variant("kokkos", default=False, description="Use Kokkos")
     variant("cuda", default=False, description="Enable CUDA backend", when="@:10.10.0")
     variant(
-        "python", default=False, description="enable python interfaces", when="@11.2.0-scorec:"
+        "python",
+        default=False,
+        description="enable python interfaces",
+        when="@11.2.0-scorec:",
     )
 
     depends_on("cxx", type="build")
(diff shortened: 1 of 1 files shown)


Sichao25:
@spackbot fix style

repos/spack_repo/builtin/packages/omega_h/package.py
```diff
@@ -82,6 +87,7 @@ class OmegaH(CMakePackage, CudaPackage):
     variant("gmsh", default=False, description="Use Gmsh C++ API")
     variant("kokkos", default=False, description="Use Kokkos")
     variant("cuda", default=False, description="Enable CUDA backend", when="@:10.10.0")
+    variant("python", default=False, description="enable python interfaces")
```
cwsmith: I think we should limit this to the 11.2.0-scorec version and later since the interface was in an unknown state prior to this release.

Sichao25:
@spackbot fix style

cwsmith (approved):

bernhardkaindl:
@Sichao25 as agreed with @cwsmith in #4672, I merged the pre-existing earlier PR first because of the `git` conflict between them and I think it is easier to rebase this new PR on top of the merge: https://github.com/spack/spack-packages/pull/4672#pullrequestreview-4872151933

As the view of this PR on Github now shows, the conflict now needs to be resolved in this PR. Can you do that?

If possible, it would be great if you could do it with a `git rebase` on top of the latest develop branch (not using a `git merge`) because then your branch keep the real `git` history of `develop` as base, and your changes stay on top of the history, which can make the review much easier if one needs to look at your commits to the PR. Thanks, Bernhard

PS: Also, there seems to be a `libomega_h.so: undefined reference to `nc_set_log_level'` triggered by `spack install omega-h +exodus +mpi +shared` now. It looks like only `netcdf-c` provides it but `parallel-netcdf` not, which I saw used in the build log of `+exodus +mpi`.

jacobmerson:
@Sichao25 please add the following line to this PR which with the updates in https://github.com/SCOREC/omega_h/pull/228 will enable the `PyOmega_h` python package to directly be used in a spack environment without any manual modification of `PYTHONPATH` or `LD_LIBRARY_PATH`.

`extends('python', when='+python')`

See the use in PCMS for example: https://github.com/jacobmerson/pcms-spack/blob/e9a7bd00201c871da0a4ffe14e17c3b5e6d2c86d/spack_repo/pcms/packages/pcms/package.py#L58

Sichao25:
@bernhardkaindl  Thanks for the notification. I've rebased the branch. However, I wasn't able to reproduce the `libomega_h` link error with `spack install +exodus +mpi +shared` alone. Could you let me know where can I find more details on which full spec triggers this?

bernhardkaindl:
@Sichao25, thanks for checking. In a new docker container, I did a completely fresh install of spack `develop`, setup https://cache.spack.io as buildcache and this latest PR branch on ARM64 today (using the and it didn't reproduce there.

bernhardkaindl (dismissed):

repos/spack_repo/builtin/packages/omega_h/package.py
```diff
@@ -127,6 +150,8 @@ class OmegaH(CMakePackage, CudaPackage):
     # https://gcc.gnu.org/bugzilla/show_bug.cgi?id=86610
     conflicts("%gcc@8:8.2", when="@:9.22.1")
 
+    conflicts("+python", when="~shared", msg="python requires shared build")
+
     def patch(self):
         if "@:9.34.8" in self.spec:
```
bernhardkaindl: @Sichao25: @cwsmith sent me the fix for the issue in https://github.com/spack/spack-packages/pull/4672#issuecomment-5205184292.

This `filter_file()` applies it to the versions that need it (tested to build both affected versions) for `+exodus` (when `+mpi +shared` are also enabled):
```suggestion
    def patch(self):
        # https://github.com/SCOREC/omega_h/commit/4dd682ef16ebf2502239ad06883e9f10c611f1c4
        if self.spec.satisfies("@10.8.6-scorec:11.0.0-scorec" ):
            filter_file(r"nc_set_log_level\(5\);", "", "src/Omega_h_exodus.cpp")
        if self.spec.satisfies("@:9.34.8"):
```
I guessed it would be worth a suggestion to fix it.

Due to the GitHub-wide actions workflows outage yesterday, the CI pipelines for this PR didn't run - and applying this suggestion (assuming posting the suggestion applies my change correctly) would re-trigger CI to run again too.

cwsmith: Thank you.  This patch makes sense to me.  Deprecating scorec versions of omegah prior to 10.8.6-scorec (the last version prior to dropping direct cuda support https://github.com/SCOREC/omega_h/issues/182) would also be helpful in the long run.

bernhardkaindl (approved):

PR #5793, opened 2026-07-29
Msmpi: Quote source tree paths for robocopy

johnwparent:
Msmpi's msbuild build system leverages robocopy. Previously msbuild was more tolerant about paths with spaces and handed them off to robocopy nicely, but more recent versions of visualstudio vendor an msbuild that is less tolerant, and passes broken paths to robocopy (if they're unqouted paths with spaces).
This breaks msmpi installs for spack with install trees or stages with spaces. 

Instead just quote the paths that robocopy will be moving via a patch.

Patch is not upstreamed as the project (msmpi) seems to be non active outside of patching vulnerabilities.


bernhardkaindl (approved):
Confirmed: This fixes the build of `msmpi` (an msbuild.MSBuildPackage)

PR #5794, opened 2026-07-29
nlohmann_json_schema_validator: add v2.4.0

Sbozzolo:
Add the latest released version


repos/spack_repo/builtin/packages/nlohmann_json_schema_validator/package.py
RikkiButler20, line 30:
```suggestion
    depends_on("cmake@3.14:", type="build", when="@2.3:")
    depends_on("cmake@3.2:", type="build")
```

RikkiButler20 (changes requested):
Update the cmake minimum required version

repos/spack_repo/builtin/packages/nlohmann_json_schema_validator/package.py
Sbozzolo: Thanks for catching that!

RikkiButler20 (approved):
Looks good

PR #5805, opened 2026-07-30
thrift: Properly send openssl variant to cmake build

paulgessinger:
This PR propagates the `openssl` variant into the CMake build. With `WITH_OPENSSL` unset, thrift's CMake code will unconditionally lookup openssl and flip `WITH_OPENSSL=ON` silently. If spack's openssl is not in the dependency tree, this will find whatever openssl might be installed on the system: not what we want.

When `WITH_OPENSSL` is synced to `openssl`, this is not the case, and with an explicit `OFF` thrift will also not use the speculatively found openssl to build.


thomas-bouvier (approved):
LGTM

RikkiButler20 (approved):

PR #5807, opened 2026-07-30
Ruff: add v0.16.0

adamjstewart:
This new minor version enables a much broader set of rules by default. We should consider adopting this in Spack as well @johnwparent


johnwparent (approved):
Looks good to me, thanks! 

Will look into the expanded default rules & bumping the version we use in Spack for style.

PR #5816, opened 2026-07-30
kokkos: add v5.2.0

lucbv:
Enabling version 5.2.0 of the Kokkos Ecosystem

Outside of the new versions and sha256, I am updating our list of AMD supported architectures in accordance with the new architectures enabled in Kokkos Core.


rbberger (dismissed):

rbberger:
@spackbot run pipeline

tpadioleau (approved):
Thanks @lucbv!

- Should we add a conflict for 
> Enable C++23 support with NVCC (available since CUDA 13.3.0) https://github.com/kokkos/kokkos/pull/9264

- Is there anything we should do about nextsilicon backend in Spack ? Is there a compiler in Spack that handles this backend ?

lucbv:
@tpadioleau 

- We can add that around line 357, I am not sure I recall the syntax, but based on the surounding code I would go with ```conflict("+cuda", when="cxxstd=23 ^cuda@:13.3.0")```?
- For NextSilicon, nothing is ready at this point we **do not** want to make it more available than it is currently

tpadioleau:
> * We can add that around line 357, I am not sure I recall the syntax, but based on the surounding code I would go with `conflict("+cuda", when="cxxstd=23 ^cuda@:13.3.0")`?

I think we also need to add the version of Kokkos, it may be easier that way:
```python
requires("@5.2: ^cuda@13.3:", when="+cuda cxxstd=23")
```
But actually, this CUDA 13.3 condition is only related to `nvcc`, right ? Compiling with clang should not have this restriction ?

> * For NextSilicon, nothing is ready at this point we **do not** want to make it more available than it is currently

Sure

rbberger:
Can we add the kokkos-tools release in here now?

lucbv:
Yes, I need to get back to this. Also since the discussion regarding cuda and C++23 is starting to be larger than just the regular updating for a release I would like to do the same thing as before and we can tackle a refactor in a second PR.
I do not like the idea of having release related activities being held up for too long.
Also I feel that doing some of these changes should happen at the time the PR is merged against core and marked as a develop only feature in our package. At release time we would only need to change the "develop" into "X.Y:".

rbberger:
@tpadioleau compiling with LLVM has a different condition. CUDA 13 only works with LLVM 22+. This isn't the host compiler compatibility that CUDA advertises, but the internal headers in LLVM and linker changes for CUDA 13 that only made it into the 22.x.

tpadioleau (approved):
thanks

repos/spack_repo/builtin/packages/kokkos/package.py
```diff
@@ -264,7 +265,10 @@ class Kokkos(CMakePackage, CudaPackage, ROCmPackage):
         "gfx950": ("amd_gfx950", "@5.1.0:"),
         "gfx1030": ("navi1030", None),
         "gfx1100": ("navi1100", "@4.1.00:"),
+        "gfx1101": ("amd_gfx1101", "@5.2.0:"),
         "gfx1103": ("amd_gfx1103", "@4.5.00:"),
+        "gfx1151": ("amd_gfx1151", "@5.2.0:"),
+        "gfx1152": ("amd_gfx1152", "@5.2.0:"),
```
amd-vkale, line 271: We probably want gfx1251 at some point here as well, but in any case, this is good and thanks for adding.

amd-vkale (approved):

repos/spack_repo/builtin/packages/kokkos/package.py
lucbv: That's something that would need to happen in Kokkos Core first, as long as it is not supported their, no point in adding it to the spack package.

rbberger (dismissed):
conflicts and requires will need some more work outside this PR, but otherwise LGTM.

rbberger:
@spackbot fix style

rbberger (approved):

PR #5821, opened 2026-07-31
cudss: Properly support CUDA 13

RemiLacroix-IDRIS:
The recipe pretended to be compatible with CUDA 12+ but it would always install precompiled libraries depending on CUDA 12.

Use the same mechanism as cuDNN recipe, that is adding the CUDA version as a suffix to the cuDSS version.

### DIFF
--- a/repos/spack_repo/builtin/packages/cudss/package.py
+++ b/repos/spack_repo/builtin/packages/cudss/package.py
@@ -10,7 +10,17 @@ from spack_repo.builtin.build_systems.generic import Package
 from spack.package import *
 
 _versions = {
-    "0.7.1": {
+    "0.7.1-13": {
+        "Linux-x86_64": (
+            "84b34ebe7fad40ec10f2aab2957a63b6070bd8ce16e3ada3e6bcac7317256347",
+            "https://developer.download.nvidia.com/compute/cudss/redist/libcudss/linux-x86_64/libcudss-linux-x86_64-0.7.1.4_cuda13-archive.tar.xz",
+        ),
+        "Linux-aarch64": (
+            "4085361899eaf08f6b707321b33a2d5263ac986015527bef8e500f2c102f0258",
+            "https://developer.download.nvidia.com/compute/cudss/redist/libcudss/linux-aarch64/libcudss-linux-aarch64-0.7.1.4_cuda13-archive.tar.xz",
+        ),
+    },
+    "0.7.1-12": {
         "Linux-x86_64": (
             "946571d9ea164f948e402dd97a14541cb90fbec800336cfa7ae644af5937632f",
             "https://developer.download.nvidia.com/compute/cudss/redist/libcudss/linux-x86_64/libcudss-linux-x86_64-0.7.1.4_cuda12-archive.tar.xz",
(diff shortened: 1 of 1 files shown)


RikkiButler20:
Looks good to me.
@ddement Any feedback?

imciner2:
As I said in my update https://github.com/spack/spack-packages/pull/5823 (I had actually started that earlier in the week, but only could finish it now), CUDA 13 needs to use the SBSA build, not the Jetson builds, and in fact the Jetson builds don't even have a CUDA 13 version for the new 0.8 release.

RemiLacroix-IDRIS:
@RikkiButler20 @bernhardkaindl : Let me know if I should backport the changes suggested by @imciner2 if he doesn't have time to rebase his PR.

bernhardkaindl (commented):
Hi @RemiLacroix-IDRIS, it looks like @imciner2 doesn't have time to rebase his PR, so I notify you suggested to see if you can backport the changes suggested by @imciner2 into this PR.

RemiLacroix-IDRIS:
Done, hopefully without breaking anything!

repos/spack_repo/builtin/packages/cudss/package.py
```diff
@@ -68,13 +110,20 @@ class Cudss(Package):
         if pkg:
             version(ver, sha256=pkg[0], url=pkg[1])
 
+            ver_split = ver.split("-")
+            if len(ver_split) == 2:
+                _, cuda_ver = ver_split
+                depends_on("cuda@{}".format(cuda_ver), when="@{}".format(ver))
```
bernhardkaindl, line 116: @RemiLacroix-IDRIS, I did a diff with #5823 and copied this additional from #5823:
```suggestion
            ver_parts = ver.split("-")
            if len(ver_parts) > 1:
                cudss_ver = ver_parts[0]
                cuda_ver = Version(ver_parts[1])
                long_ver = f"{cudss_ver}-{cuda_ver}"
                depends_on(f"cuda@{cuda_ver}", when=f"@{long_ver}")
```
The difference I see is this handles more more than two ver_parts, eg.:
```py
"<cudss_ver>-<cuda_ver>-<subver>"
```
While no version in this format has been added yet, it looks more robust and used the more modern and shorter f-string mechanism.

repos/spack_repo/builtin/packages/cudss/package.py
```diff
@@ -68,13 +110,20 @@ class Cudss(Package):
...
+
+    # Versions without CUDA 13 builds
+    depends_on("cuda@12", when="@:0.6")
+
```
bernhardkaindl, line 120: #5823 also declares the 0.5 versions present in the recipe to be limited ot CUDA 12, which looks correct I guess:
```suggestion
    depends_on("cuda@12", when="@0.5")

```

bernhardkaindl (changes requested):
Diffed from #5823, I only see one missing line:
```py
    depends_on("cuda@12", when="@0.5")
```
The other diff I found is cosmetic, but I like the update to use an f-string better, and to be robust against a 3rd version component like "<cudss_ver>-<cuda_ver>-<subver>" looks like an improvement.

repos/spack_repo/builtin/packages/cudss/package.py
RemiLacroix-IDRIS: I just merged the two lines into one with an upper bound.

repos/spack_repo/builtin/packages/cudss/package.py
RemiLacroix-IDRIS: I sort of did that on purpose because I am not really sure what would be the use of handling the extra dash.

The artificially added CUDA version should always come last I think. So it felt like the logic wouldn't quite work if CUDSS was to use versions with a dash in them.

If you feel it's better, I can make that change though or add an assert with "len < 2" so that's it's clear it's currently not supported.

RemiLacroix-IDRIS: @bernhardkaindl Any thoughts?

bernhardkaindl: @RemiLacroix-IDRIS, sorry for the long delay, had to attend to other things. Will merge!

bernhardkaindl (approved):
The review comments have been answered.

AFAICS, #5823 has been merged into this PR and can be closed.

I guess this means we can close the conflicting #5823.

PR #5822, opened 2026-07-31
cusparselt: Reenable CUDA 13 support

RemiLacroix-IDRIS:
Use the same mechanism as the cuDNN recipe.

I have renamed the CUDA version suffix to match what is done in the cuDNN recipe but I can revert that part if the majority thinks we should keep it as is.


RikkiButler20 (commented):
Looks good to me. @thomas-bouvier any feedback?

thomas-bouvier (approved):
LGTM, thanks Rémi

RikkiButler20 (approved):

PR #5824, opened 2026-07-31
ollama: add 0.32.5 and fix `ollama -v` to show the version

imciner2:
Adds the most recent ollama version and updates the Go version dependency (based on the upstream repo, 0.23.1 was the first to require Go 1.26.0).

This also fixes the build to pass the version number into the build system to embed it in the final binary. This is based on the way it is done in the upstream ollama build steps.

Now it will show
```
$ ollama -v
Warning: could not connect to a running Ollama instance
Warning: client version is 0.32.5
```
instead of 
```
$ ollama -v
Warning: could not connect to a running Ollama instance
Warning: client version is 0.0.0
```

### DIFF
--- a/repos/spack_repo/builtin/packages/ollama/package.py
+++ b/repos/spack_repo/builtin/packages/ollama/package.py
@@ -20,6 +20,7 @@ class Ollama(GoPackage, CudaPackage):
     # A shell script is run by `go generate` which assumes source is in a git
     # repo.  So we must use git VCS and not tarballs and defeat source caching.
     with default_args(no_cache=True):
+        version("0.32.5", commit="eec8e0b9458b8a01be0c216a9cc53eefde24ef50")
         version("0.20.7", commit="8d0dcf4b6daf8d7833c8b55108e5b45063795e57")
         version("0.13.1", commit="5317202c38437867bc6c9ed21ffc5c949ab6794c")
         version("0.12.11", commit="c1149875234a51aa1e5e60b74f3807f5982c60fa")
@@ -52,6 +53,7 @@ class Ollama(GoPackage, CudaPackage):
     depends_on("go@1.23.4:", type="build", when="@0.5.2:")
     depends_on("go@1.24.0:", type="build", when="@0.5.13:")
     depends_on("go@1.24.1:", type="build", when="@0.12.10:")
+    depends_on("go@1.26.0:", type="build", when="@0.23.1:")
     depends_on("git", type="build")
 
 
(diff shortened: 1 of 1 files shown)


teaguesterling (approved):
LGTM

imciner2:
I fixed one of the style checks, but the other is getting thrown because the linker line is extremely long. Can that stay one line, or do I need to find a way to split it?

imciner2:
Actually, this isn't quite ready yet. They changed how it interacts with llama-server stuff in 0.30 apparently, so now I think I need to add a new CMake build step to build the llama-server that ollama wants.

teaguesterling:
Thanks for working on this!

bernhardkaindl:
@imciner2, just a reminder that you said that this isn't quite ready yet:
> Actually, this isn't quite ready yet. They changed how it interacts with llama-server stuff in 0.30 apparently, so now I think I need to add a new CMake build step to build the llama-server that ollama wants.

I've set the draft flag for this PR to indicate this state of the PR for when iterating of the PRs formal review. Once it is is fully fixed, please update it (you should have a button to set the PR to ready again). Once it is finished and is ready for final, formal spack-maintainer review for merging it.

brettviren (approved):
Looks reasonable to me.

imciner2:
Ok, I think this is ready now.

Basically, starting with 0.30 Ollama has changed how llama-cpp is used, and so now it needs to be built differently. There is a CMake build system that will build everything correctly (the local dev docs say to use CMake for everything: https://docs.ollama.com/development), so I have moved the package recipe to using that for 0.30+, and enabled the CUDA 13 backend for llama-cpp when it is getting built. (there are more possible backends, like cuda_v12, and some rocm ones), so there is more potential for refactoring/configuration here, but I don't have good ways to test those.

teaguesterling:
@spackbot fix style

Do you mind running the formatter if this doesn't work?

bernhardkaindl:
@imciner2, looks good so far. To get CI working, you'll need to split this string across two lines:
(I think you can make make it two quoted strings and become one or a +)
```py
return [
    f"-ldflags=-w -s -X=github.com/ollama/ollama/version.Version={self.spec.version.string} "
    f"-X=github.com/ollama/ollama/server.mode=release"
]
```

Install ruff (https://docs.astral.sh/ruff/installation/) and run `ruff check` on the file, if it passes the format is accepted by CI.

imciner2:
It actually still complained about the line length of the first string, so I just went ahead and shortened the version variable by using a local, and so now it doesn't complain with the strings split.

bernhardkaindl (dismissed):
@imciner2, Started server and client, looks great!

bernhardkaindl:
@spackbot fix style

bernhardkaindl (approved):

PR #5828, opened 2026-07-31
scorep: add v10.1

wrwilliams:
New bugfix release for Score-P. No changes to package logic--the 10.0 patches should be already version-specific.


RikkiButler20 (approved):
Looks good.

PR #5830, opened 2026-07-31
bigdft: Improve CUDA support

RemiLacroix-IDRIS:
Properly handle the cuda_arch variant in order to support other architectures than the builtin SM20 default.

### DIFF
--- a/repos/spack_repo/builtin/packages/bigdft_chess/package.py
+++ b/repos/spack_repo/builtin/packages/bigdft_chess/package.py
@@ -111,6 +111,14 @@ class BigdftChess(AutotoolsPackage, CudaPackage):
             args.append(f"--with-cuda-path={spec['cuda'].prefix}")
             args.append(f"--with-cuda-libs={spec['cuda'].libs.link_flags}")
 
+            cuda_arch = [x for x in spec.variants["cuda_arch"].value if x and x != "none"]
+            if cuda_arch:
+                args.append(
+                    "NVCC_FLAGS={0} --compiler-options {1}".format(
+                        " ".join(self.cuda_flags(cuda_arch)), self.compiler.cxx_pic_flag
+                    )
+                )
+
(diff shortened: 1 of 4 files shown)


RemiLacroix-IDRIS:
@spackbot fix style

bernhardkaindl:
@iamashwin99 Hello Ashwin, do you need more time for review? (Review invitation was 7 days ago)

iamashwin99:
I haven't checked it yet, sorry.  I need a little more time to compile it and test it. Do you want to pass on the review to some one else?

bernhardkaindl:
> I haven't checked it yet, sorry. I need a little more time to compile it and test it. Do you want to pass on the review to some one else?

Hi @iamashwin99, great that you're checking, I'll wait for your feedback!

iamashwin99:
@RemiLacroix-IDRIS Have you tried compiling this without a Cuda Arch? I'm trying to test my hypothesis that because the string `'none'` which is the default value of `cuda_arch` evaluates as truthy in Python, it won't be filtered out by the list comprehension `[x for x ... if x]`. As a result, the `if cuda_arch: ` condition will trigger and pass `['none']` into `self.cuda_flags()` .  I am not yet able to test this locally and I was wondering since you have the setup could you please check this?


If it does fail, filtering out the 'none' string directly should resolve it:
```python
cuda_arch = [x for x in spec.variants["cuda_arch"].value if x and x != "none"]
```

RemiLacroix-IDRIS:
If I explicitly set `cuda_arch=none` then it indeed doesn't work so I have repushed with the suggested fix.

iamashwin99:
LGTM, Thanks for the fix and sorry for the late review.

iamashwin99 (approved):

bernhardkaindl (approved):
Approving per mainainer approval.

PR #5823, opened 2026-07-31
cudss: add 0.8.0, add CUDA 13 versions, fix the CUDA dependencies

imciner2:
Summary by @bernhardkaindl:
- #5821 is aimed into the same direction
- This PR also adds 0.8.0 and seems to add more logic that looks like to be helpful too.
- It also adds @imciner2 as an additional maintainer, great! 👍 
- It looks like this PR superseedes 5821 (but it needs a `git` conflict resolved)
----
Implement both CUDA 12 and CUDA 13 versions of the library, following the same method used by cuDNN (appending `-12` or `-13` to the end of the version to denote which CUDA version to use). CUDA 13 aarch64 uses the new SBSA build instead of the Jetson build.

Also, add the newest 0.8.0 version.

PR #5831, opened 2026-07-31
cudss: support CUDA 13

Sbozzolo:
Follow the cuDNN package's CUDA-suffixed version scheme so each precompiled cuDSS archive depends on the matching CUDA major.


RemiLacroix-IDRIS:
Funny how we all had the same idea today... See #5821 and #5823.

Sbozzolo:
I'll close mine :)

PR #5832, opened 2026-07-31
zookeeper: add v3.8.6

RatneshVaibhav:
Adds version 3.8.6 of zookeeper (latest patch release on the 3.8.x line).

Tested locally with `spack install zookeeper@3.8.6` — installs successfully.


repos/spack_repo/builtin/packages/zookeeper/package.py
```diff
@@ -28,4 +29,4 @@ def install(self, spec, prefix):
     def setup_run_environment(self, env: EnvironmentModifications) -> None:
         env.set("ZOOBINDIR", self.prefix.bin)
         env.set("ZOOCFGDIR", ".")
-        env.set("ZOO_LOG_DIR", ".")
+        env.set("ZOO_LOG_DIR", ".")
```
RikkiButler20: I think you removed the last line of the file

RatneshVaibhav: Thanks for catching that! I've added the missing trailing newline and pushed the fix.

RikkiButler20 (approved):
Looks good. Thanks for the update.

RatneshVaibhav:
Thanks for the review!

PR #5833, opened 2026-07-31
ROCm: Add the Therock 7.14.0 release to all packages

afzpatel:

### DIFF
--- a/repos/spack_repo/builtin/packages/amdsmi/package.py
+++ b/repos/spack_repo/builtin/packages/amdsmi/package.py
@@ -102,7 +102,7 @@ class Amdsmi(CMakePackage):
         tag="esmi_pkg_ver-4.2",
         commit="8da6df879b0acafbcbe78e5b54af81a9e51dce6d",
         placement="projects/amdsmi/esmi_ib_library",
-        when="@7.2:",
+        when="@7.2",
     )
     resource(
         name="esmi_ib_library",
--- a/repos/spack_repo/builtin/packages/amdsmi/package.py
+++ b/repos/spack_repo/builtin/packages/amdsmi/package.py
@@ -34,6 +34,7 @@ class Amdsmi(CMakePackage):
             url = "https://github.com/ROCm/rocm-systems/archive/refs/tags/therock-{0}.{1}.tar.gz"
             return url.format(version[0], version[1])
 
+    version("7.14.0", sha256="8cadf0d5c0f53f334b7b940a78619d1746c913b26ae719e2a09e20a6f7128330")
     version("7.13.0", sha256="86162d975c59c2f43eb79187378a9b10615db5c1d73441e7e0b7621a7ef8962c")
     version("7.2.3", sha256="e90cfd8694af28a56433c8827a581ee12a4ba835f0d952436741d9e0f3f8685b")
     version("7.2.1", sha256="201f19174eafbace2f7abf0d1178ebb17db878191276aba6d23f0e1758b0e10f")
@@ -73,10 +74,36 @@ class Amdsmi(CMakePackage):
     depends_on("libdrm@:2.4.124", when="@:7.2")
     depends_on("libdrm@2.4.131:", when="@7.13:")
     depends_on("py-pyyaml")
+    depends_on("libnl", when="@7.14:")
+    depends_on("libmnl", when="@7.14:")
 
     depends_on("googletest@1.14:", type="build", when="@6.4:")
     depends_on("googletest@1.16:", type="build", when="@7.2:")
 
+    resource(
+        name="esmi_ib_library",
+        git="https://github.com/amd/esmi_ib_library.git",
+        tag="esmi_pkg_ver-5.2.1.1",
+        commit="d494a3194ceb4cc4dbb2debf9fcbe8773c6d3bef",
+        placement="projects/amdsmi/esmi_ib_library",
+        when="@7.14",
+    )
+    resource(
+        name="esmi_ib_library",
+        git="https://github.com/amd/esmi_ib_library.git",
+        tag="esmi_pkg_ver-5.1.1",
+        commit="6cf2435dd15519ab4148c36751de7a9efc27d8dc",
+        placement="projects/amdsmi/esmi_ib_library",
+        when="@7.13",
+    )
+    resource(
+        name="esmi_ib_library",
+        git="https://github.com/amd/esmi_ib_library.git",
+        tag="esmi_pkg_ver-5.1.1",
+        commit="6cf2435dd15519ab4148c36751de7a9efc27d8dc",
+        placement="projects/amdsmi/esmi_ib_library",
+        when="@7.13",
+    )
     resource(
         name="esmi_ib_library",
         git="https://github.com/amd/esmi_ib_library.git",
--- a/repos/spack_repo/builtin/packages/amdsmi/package.py
+++ b/repos/spack_repo/builtin/packages/amdsmi/package.py
@@ -99,14 +99,6 @@ class Amdsmi(CMakePackage):
     resource(
         name="esmi_ib_library",
         git="https://github.com/amd/esmi_ib_library.git",
-        tag="esmi_pkg_ver-5.1.1",
-        commit="6cf2435dd15519ab4148c36751de7a9efc27d8dc",
-        placement="projects/amdsmi/esmi_ib_library",
-        when="@7.13",
-    )
-    resource(
-        name="esmi_ib_library",
-        git="https://github.com/amd/esmi_ib_library.git",
         tag="esmi_pkg_ver-4.2",
         commit="8da6df879b0acafbcbe78e5b54af81a9e51dce6d",
         placement="projects/amdsmi/esmi_ib_library",
--- a/repos/spack_repo/builtin/packages/llvm_amdgpu/package.py
+++ b/repos/spack_repo/builtin/packages/llvm_amdgpu/package.py
@@ -387,7 +387,7 @@ class LlvmAmdgpu(CMakePackage, LlvmDetection, CompilerPackage):
                     llvm_version = "20"
                 args.append(
                     self.define(
-                        "ROCM_DEVICE_LIBS_BITCODE_INSTALL_LOC_NEW",
+                        "ROCM_DEVICE_LIBS_BITCODE_INSTALL_LOC",
                         f"lib/clang/{llvm_version}/lib/amdgcn",
                     )
                 )
--- a/repos/spack_repo/builtin/packages/llvm_amdgpu/package.py
+++ b/repos/spack_repo/builtin/packages/llvm_amdgpu/package.py
@@ -247,7 +247,7 @@ class LlvmAmdgpu(CMakePackage, LlvmDetection, CompilerPackage):
             sha256=d_shasum,
             when=f"@{d_version}",
         )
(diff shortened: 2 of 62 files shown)


repos/spack_repo/builtin/packages/llvm_amdgpu/package.py
```diff
@@ -371,7 +387,7 @@ def cmake_args(self):
                     llvm_version = "20"
                 args.append(
                     self.define(
-                        "ROCM_DEVICE_LIBS_BITCODE_INSTALL_LOC",
+                        "ROCM_DEVICE_LIBS_BITCODE_INSTALL_LOC_NEW",
```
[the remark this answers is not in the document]
afzpatel: added suggested change

repos/spack_repo/builtin/packages/rocprofiler_systems/package.py
```diff
@@ -232,7 +240,8 @@ class RocprofilerSystems(ROCmLibrary, CMakePackage):
         when="@7.1.1:",
     )
     depends_on("libiberty+pic", when="+internal-dyninst")
-    depends_on("intel-tbb@2019:2020.3", when="~internal-tbb")
+    depends_on("intel-tbb@2019:2020.3", when="@:7.13 ~internal-tbb")
+    depends_on("intel-tbb@:2021.3", when="~internal-tbb")
```
bernhardkaindl: Hi Afzal Batel @afzpatel, I reviewed the progress of the PR it appears to look good.

I reviewed the overlap found but suppressed as not important in the last GitHub Copilot review:

For ROCm **≤7.13 with `~internal-tbb`**, **both** rules apply simultaneously:

```py
- Rule from line 243: `intel-tbb@2019:2020.3`
- Rule from line 244: `intel-tbb@:2021.3`
```

Spack intersects these → resolves to `intel-tbb@2019:2020.3` (which happens to be correct). So it **works**, but it's expressing the same constraint twice via two overlapping rules — which is redundant and potentially confusing.

### What the Copilot review comment suggests

Scope line 244 to 7.14+:
```suggestion
    depends_on("intel-tbb@2019:2020.3", when="@:7.13 ~internal-tbb")
    depends_on("intel-tbb@:2021.3", when="@7.14: ~internal-tbb")
```
Now, the rules are **non-overlapping**: each version range has exactly one constraint, making the intent explicit.

### Was the Copilot finding correct?

**Yes, it's valid.** The bug is not a runtime failure — Spack resolves it correctly — but it's a design smell: the unconditional `when="~internal-tbb"` on line 244 was intended only for 7.14+ but shadows the stricter `@:7.13` constraint above it, acting as a background depencency, making the 1st line the retrospectively the overlapping dependency The fix is simple and makes the intent clearer by not overlapping.

Personally, I'd say that overlapping ranges are normal in spack, and I'm neutral about it, but in this case, I was confused because the order is reversed to what I normally see. 

My preferred order is to swap the lines, newer versions first, as commonly seen in recipes:
```py
    depends_on("intel-tbb@:2021.3", when="@7.14: ~internal-tbb")
    depends_on("intel-tbb@2019:2020.3", when="@:7.13 ~internal-tbb")
```
This fixes the slightly confusing order for me. The background (if using overlapping) is set first by removing the `@7.14:` from it, but kept for clarity, and it is then partly overlapped or refined for the older versions.

With 7.15, another "higher" dependency for 7.15 would be on top, and it will then be  progressively refined for these, then older versions (with or without overlap).

However, as this is not a runtime failure and only what could be seen as stylistic difference to what I see most often in other recipes, which makes the current version appear as a bit of a code smell.

As it is not more, than that and it could be put on a "remember the final review of the previous ROCm" list of things to optionally consider, I give my approval.

bernhardkaindl (dismissed):
Approving with a small comment for later.

bernhardkaindl:
@srekolam, @renjithravindrankannath, @cgmb, did you have enough time for review?
(review was invited on Aug 4, or on Aug 6 for @cgmb)

Are you fine with the PR? Do you see anything missing?

repos/spack_repo/builtin/packages/amdsmi/package.py
```diff
@@ -73,10 +74,28 @@ def url_for_version(self, version):
...
+        when="@7.14",
+    )
+    resource(
+        name="esmi_ib_library",
+        git="https://github.com/amd/esmi_ib_library.git",
+        tag="esmi_pkg_ver-5.1.1",
+        commit="6cf2435dd15519ab4148c36751de7a9efc27d8dc",
+        placement="projects/amdsmi/esmi_ib_library",
+        when="@7.13",
+    )
```
scheibelp, line 98: This overlaps with the resource below I think (both have the same placement).

repos/spack_repo/builtin/packages/mivisionx/package.py
```diff
@@ -167,14 +170,21 @@ def patch(self):
...
                 "utilities/runcl/CMakeLists.txt",
                 string=True,
             )
+        if self.spec.satisfies("@7.14:"):
+            filter_file(
+                r"crypto",
+                "{0}".format(self.spec["openssl"].libs),
+                "utilities/runvx/CMakeLists.txt",
+                string=True,
+            )
```
scheibelp, line 186: Minor: this looks the same except for not applying to `utilities/runcl/CMakeLists.txt` - I'd recommend consolidating that

repos/spack_repo/builtin/packages/rdc/package.py
```diff
@@ -67,7 +68,8 @@ def url_for_version(self, version):
     depends_on("grpc@1.55.0+shared", when="@:6.0")
     depends_on("grpc@1.59.1+shared", when="@6.1")
     depends_on("grpc@1.61.2+shared", when="@6.2:6.4")
-    depends_on("grpc@1.67.1 cxxstd=17 +shared", when="@7.0:")
+    depends_on("grpc@1.67.1 cxxstd=17 +shared", when="@7.0:7.13")
+    depends_on("grpc@1.78.1+shared", when="@7.14:")
```
scheibelp, line 72: Question: I'd normally assume that if earlier versions need c++17, that later versions would need that or later, but is that not the case here?

scheibelp (changes requested):
I have a couple suggestions (one is minor and can be discarded) and a question. Thanks!

repos/spack_repo/builtin/packages/rocjpeg/package.py
```diff
@@ -66,12 +72,19 @@ class Rocjpeg(ROCmLibrary, CMakePackage):
...
+                r"${ROCM_PATH}/lib/llvm/bin/clang++",
+                "{0}/bin/clang++".format(self.spec["llvm-amdgpu"].prefix),
+                "CMakeLists.txt",
+                string=True,
+            )
+
+    @property
+    def root_cmakelists_dir(self):
+        if self.spec.satisfies("@7.13.0:"):
+            return join_path(super().root_cmakelists_dir, "projects", "rocdecode")
```
srekolam: join_path(super().root_cmakelists_dir, "projects", "rocjpeg")

srekolam (changes requested):

repos/spack_repo/builtin/packages/rocjpeg/package.py
afzpatel: updated, thanks!

repos/spack_repo/builtin/packages/mivisionx/package.py
afzpatel: updated, thanks!

repos/spack_repo/builtin/packages/amdsmi/package.py
afzpatel: updated, thanks!

repos/spack_repo/builtin/packages/rocprofiler_systems/package.py
afzpatel: thanks for the suggestion! I updated it

renjithravindrankannath:
Update rocm-examples 7.14. Tag is available - https://github.com/ROCm/rocm-examples/releases/tag/therock-7.14.

srekolam (approved):

repos/spack_repo/builtin/packages/rocm_cmake/package.py
renjithravindrankannath, line 87:
```suggestion
        "7.13.0",
        "7.14.0",
```

repos/spack_repo/builtin/packages/rdc/package.py
afzpatel: I was able to build with cxxstd=11 so I don't think it's a requirement now

afzpatel:
@spackbot rerun pipeline

renjithravindrankannath:
rocalution@7.14 can be updated.

repos/spack_repo/builtin/packages/mivisionx/package.py
```diff
@@ -167,14 +170,20 @@ def patch(self):
                 "model_compiler/python/nnir_to_clib.py",
                 string=True,
             )
-        if self.spec.satisfies("@6.2:"):
+        if self.spec.satisfies("@6.2:7.13"):
```
scheibelp, line 173: Question: This overlaps with the above check `if self.spec.satisfies("@:7.13 +hip"):` - is that intended? i.e. `@7.13`  will apply all the above `filter_file` operations, plus this

repos/spack_repo/builtin/packages/mivisionx/package.py
```diff
@@ -167,14 +170,20 @@ def patch(self):
...
                 "{0}".format(self.spec["openssl"].libs),
-                "utilities/runvx/CMakeLists.txt",
                 "utilities/runcl/CMakeLists.txt",
                 string=True,
             )
+        if self.spec.satisfies("@7.13:"):
+            filter_file(
+                r"crypto",
+                "{0}".format(self.spec["openssl"].libs),
+                "utilities/runvx/CMakeLists.txt",
```
scheibelp, line 184: Originally this modification was applied for all of `@6.2:` and I think the current update removes that for some versions (`@6.3:7.1`).  I think this may have been to address https://github.com/spack/spack-packages/pull/5833/changes#r3755302751. You can undo that if it makes it easier to restore what versions these are applied for. I was originally thinking you could compute the set of files based on versions, and then make one call to `filter_file` with `*args`

scheibelp (commented):
I had a couple more questions about `mivisionx`. Every other package looks good to me.

repos/spack_repo/builtin/packages/mivisionx/package.py
afzpatel: Thanks for catching that! I've updated it so it applies on runvx/CMakeLists.txt for all version `@6.2:`. It should not apply on `runcl/CMakeLists.txt` for 7.14.0

repos/spack_repo/builtin/packages/mivisionx/package.py
afzpatel: yes both are meant to be applied for 7.13

afzpatel: Neither should be applied for 7.14

renjithravindrankannath (approved):
LGTM

afzpatel:
@scheibelp please review again

scheibelp (approved):

PR #5184, opened 2026-06-12
autoconf: add v2.73

alalazo:

PR #5838, opened 2026-07-31
autoconf: add v2.73

qwertos:


RikkiButler20 (approved):
LGTM

bernhardkaindl:
@spackbot run pipeline

The failure I saw was due to ftpmirror.gnu.org outage.

bernhardkaindl:
Hi @qwertos, I looked at the pipeline fail: It looks like you'd need to fix the build of other packages (related or not related makes no difference) to have CI pass on this change. It might help to set the current spack version as the preferred, using

```py
   version(
      "2.72",
      ...,
      preferred=True
   )
```

That way you also avoid causing any issue outside of CI, because most packages are not actually tested in CI.

For your reference, here is the same change, nothing else. It also fails: https://github.com/spack/spack-packages/pull/5184

alalazo:
Duplicate of #5184

alalazo:
If this fixes the package before #5184, let's close #5184

PR #5849, opened 2026-07-31
doxygen: Windows support

johnwparent:
Adds Windows support to doxygen builds

Adds missing fmt dep
Add proper bison/flex usage for windows

CMake args to support fmt detection

### DIFF
--- a/repos/spack_repo/builtin/packages/doxygen/package.py
+++ b/repos/spack_repo/builtin/packages/doxygen/package.py
@@ -97,17 +97,22 @@ class Doxygen(CMakePackage):
     depends_on("cmake@3.14:", type="build", when="@1.10:")
 
     depends_on("python", type="build")  # 2 or 3 OK; used in CMake build
-    depends_on("iconv")
-    depends_on("flex", type="build")
-    depends_on("bison", type="build")
-    # code.l just checks subminor version <=2.5.4 or >=2.5.33
-    # but does not recognize 2.6.x as newer...could be patched if needed
-    depends_on("flex@2.5.39", type="build", when="@1.8.10")
-    depends_on("bison@2.7:", type="build", when="@1.8.10:")
+    for plat in ["linux", "darwin", "freebsd"]:
+        with when(f"platform={plat}"):
+            depends_on("iconv")
+            depends_on("flex", type="build")
+            depends_on("bison", type="build")
+            # code.l just checks subminor version <=2.5.4 or >=2.5.33
+            # but does not recognize 2.6.x as newer...could be patched if needed
+            depends_on("flex@2.5.39", type="build", when="@1.8.10")
+            depends_on("bison@2.7:", type="build", when="@1.8.10:")
+    with when("platform=windows"):
+        depends_on("winbison", type="build")
 
     # originally bundled dependencies
     depends_on("spdlog", when="@1.9.8:")
     depends_on("sqlite", when="@1.10:")
+    depends_on("fmt", when="@1.14:")
 
     # optional dependencies
     depends_on("graphviz", when="+graphviz", type="run")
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/doxygen/package.py
```diff
@@ -97,17 +97,22 @@ def determine_variants(cls, exes, version_str):
...
 
     depends_on("python", type="build")  # 2 or 3 OK; used in CMake build
-    depends_on("iconv")
-    depends_on("flex", type="build")
-    depends_on("bison", type="build")
-    # code.l just checks subminor version <=2.5.4 or >=2.5.33
-    # but does not recognize 2.6.x as newer...could be patched if needed
-    depends_on("flex@2.5.39", type="build", when="@1.8.10")
-    depends_on("bison@2.7:", type="build", when="@1.8.10:")
+    for plat in ["linux", "darwin", "freebsd"]:
```
sethrj, line 100: This should also include `cray` I think… see `SPACK_ROOT/lib/spack/spack/platforms`. But maybe it'd be easier to just do what a bunch of other packages do and write:
```
    if sys.platform != "win32":
        # original recipe
    else:
        # winbison
```
but I'm not sure what the implications of this are (or whether you're building in wsl, or using cygwin, or if it's still win32 for 64-bit or ARM processors, or …

repos/spack_repo/builtin/packages/doxygen/package.py
```diff
@@ -163,8 +168,16 @@ def patch(self):
...
             self.define("use_sys_sqlite3", self.spec.satisfies("@1.10:")),
             self.define("use_sys_fmt", self.spec.satisfies("@1.14:")),
         ]
+        if self.spec.satisfies("@1.14:"):
+            args.append(
+                self.define(
+                    "fmt_DIR",
+                    join_path(self.spec["fmt"].prefix, "lib", "cmake", "fmt"),
+                )
+            )
```
sethrj, line 182: The builtin `CMAKE_PREFIX_PATH` update that spack performs means this shouldn't be required: the fact that `use_sys_fmt` has been provided means that it *should* find it if it's there. I think it worked up to this point without the `depends_on("fmt")` due to a transitive dependency on `spdlog`.

sethrj (changes requested):

repos/spack_repo/builtin/packages/doxygen/package.py
johnwparent: Wsl is a real linux VM, so it behaves exactly like every other linux install.
We do not support cygwin with Spack, nor will we unless someone _really_ wants me to (and makes a really convincing argument why we should) or contributes it themselves. 

It is still win32 for ARM processors. 

I probably added a lot of those sys.platform checks, that was likely wrong of me, I've been doing back and removing them. I do not think they're generally what we want for modeling specs. 

The sys.platform approach is _probably_ fine for dependencies like this, but for variants it can lead to easy to make mistakes and bugs when specifying variants, that lead to hard to diagnose solver failures (i.e. paraview 6 depends on a variant of seacas that was not defined at all on windows and segfaults clingo). 

Either way, I think it's more correct to specify the constraints at the spec level.

> This should also include cray I think… see SPACK_ROOT/lib/spack/spack/platforms

I used to do this, but have not been for some time after someone suggested I stop, IIRC the reasoning was something like we don't still detect cray as a platform

johnwparent: Yeah the cray platform module is now a single method, I do not think `cray` should be in this list. Happy to be wrong though.

repos/spack_repo/builtin/packages/doxygen/package.py
johnwparent: `use_sys_fmt` just means doxygen won't try to external_project_add a copy of fmt, it doesn't mean anything for whether doxygen will find `fmt`'s config module. 

You're right though, it should have picked up the module, the find_package explicitly looks for the config mode. I did require this change to build on a fedora and Windows. I'll look deeper.

PR #5850, opened 2026-07-31
opencv: supply posix paths on Windows

johnwparent:
All ocv needs to build on Windows is converting the cmake args to use posix paths.

### DIFF
--- a/repos/spack_repo/builtin/packages/opencv/package.py
+++ b/repos/spack_repo/builtin/packages/opencv/package.py
@@ -27,6 +27,7 @@ class Opencv(CMakePackage, CudaPackage):
 
     version("master", branch="master")
     version("4.13.0", sha256="1d40ca017ea51c533cf9fd5cbde5b5fe7ae248291ddf2af99d4c17cf8e13017d")
+    version("4.12.0", sha256="44c106d5bb47efec04e531fd93008b3fcd1d27138985c5baf4eafac0e1ec9e9d")
     with default_args(deprecated=True):
         # https://www.cvedetails.com/cve/CVE-2025-53644/
         version(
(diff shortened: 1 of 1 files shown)


johnwparent:
@spackbot fix style

adamjstewart:
Never seen this before. Is this required for all CMake packages?

johnwparent:
> Never seen this before. Is this required for all CMake packages?

I completely missed this comment, sorry for the delayed response! 

Great question.

CMake in general does not do well when handed Windows paths. Escaped windows paths can be ok if you're careful and there's no too many IO layers, but posix paths are safer. 

Most CMake packages in Spack don't pass so many paths to CMake on the CLI, but for those that do, yes. 

It's a bit tougher to solve from the builder side of things since we can't know what a variable is supposed to be (and I'm 100% sure some package out there relies on a backslash on the CLI to function for some reason).
I've played around with adding a new command to the cmake builder interface, something like `define_path(...` but it's never really gone anywhere.

I dislike having ad-hoc casts like this, but even if we did have internal path consistency inside Spack, passing paths like this out of spack into tooling would still require some layer of translation.

adamjstewart:
I actually quite like the `define_path` idea. That would at least give us a single location to document the Windows weirdness and point to CMake docs. Using something like `self._posix_path` feels wrong and I'm worried about whether or not private attrs are even tracked in Spack's package API versioning.

johnwparent:
Ok cool! I'll PR the `define_path` approach, glad you're interested.

 > I'm worried about whether or not private attrs are even tracked in Spack's package API versioning.

Moot if we got with the CMake based approach, which I think we agree is the right choice, and I agree that ad-hoc methods like this are a bit weird, but I'm still kinda curious about this.

Are private package apis part of the `spack.package` api? Or do we make guarantees about the stability of the interfaces associated with the packages themselves for downstream repos inheriting the builtin packages? In which case, I think I still would consider a private method not part of the API, but that may be my c++ bias.

PR #5852, opened 2026-07-31
silo: build system and version constraint for Windows

johnwparent:
Silo builds out of the box on Windows, for versions older than 4.11.1, just reflect that with the appropriate version constraints.


patrickb314 (approved):
This change looks appropriate to handle Windows, and it passes all CI checks, so I'm approving.

bernhardkaindl (approved):

PR #5853, opened 2026-08-01
lammps: deprecation of versions older than 2024

rbberger:
- also adds missing conflict for kokkos@5 in older versions
- this is in preparation removing anything that depends on Kokkos 3


RikkiButler20 (approved):
LGTM

PR #5855, opened 2026-08-01
sra-tools: constrain libxml2 version to work around "error: expected ';', ',' or ')' before 'ATTRIBUTE_UNUSED'" and related errors

mikerenfro:
In sra-tools@3.3.0, and possibly earlier versions, depending on libxml2@2.15: cause build errors such as:

```
/tmp/spack1/spack-stage/spack-stage-sra-tools-3.3.0-bvl4bnfxt3gk3pzsii2rfbzhgqmvhw2i/spack-src/libs/kxml/xml.c:174:52: error: expected ';', ',' or ')' before 'ATTRIBUTE_UNUSED'
  174 | static void s_xmlGenericErrorDefaultFunc(void *ctx ATTRIBUTE_UNUSED,
      |                                                    ^~~~~~~~~~~~~~~~
```

libxml2@:2.13 does not exhibit the same behavior.

### DIFF
--- a/repos/spack_repo/builtin/packages/sra_tools/package.py
+++ b/repos/spack_repo/builtin/packages/sra_tools/package.py
@@ -23,7 +23,7 @@ class SraTools(CMakePackage):
 
     depends_on("openjdk")
     depends_on("flex@2.6:")
-    depends_on("libxml2")
+    depends_on("libxml2@:2.13")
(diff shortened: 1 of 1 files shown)


bernhardkaindl (approved):
Issue confirmed confirmed and the fix works.

PR #5856, opened 2026-08-01
py-tensorflow: add libclang variant

wdconinc:
This PR adds to `py-tensorflow` a new variant `libclang` that decides whether to depend on `py-libclang` or `llvm +clang +python`. The last `libclang` [on PyPI](https://pypi.org/project/libclang/) (and in spack) is now 18.1.1, and there are more recent llvm releases that can be used with tensorflow. LLVM, the project, [has taken over](https://github.com/sighingnow/libclang/issues/84) the management of the python bindings but it hasn't resulted in a newer PyPI release.

This new variant defaults to the previous behavior, but in environments that include an `llvm` for other reasons, it may result in different concretization outcomes nonetheless (since the preference for newer versions may be weighted stronger than deviations from default variants).

(Not marking as draft to trigger a gitlab pipeline to see what it does for its concretization.)

### DIFF
--- a/repos/spack_repo/builtin/packages/py_tensorflow/package.py
+++ b/repos/spack_repo/builtin/packages/py_tensorflow/package.py
@@ -179,6 +179,11 @@ class PyTensorflow(Package, CudaPackage, ROCmPackage, PythonExtension):
     variant("mpi", default=False, description="Build with MPI support")
     variant("android", default=False, description="Configure for Android builds")
     variant("ios", default=False, description="Build with iOS support (macOS only)")
+    variant(
+        "libclang",
+        default=True,
+        description="Use py-libclang instead of LLVM's bundled Python clang bindings",
+    )
     variant("monolithic", default=False, description="Static monolithic build")
     variant("numa", default=False, description="Build with NUMA support")
     variant(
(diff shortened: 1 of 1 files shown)


wdconinc:
@bernhardkaindl alerted me to #5561 but I'm not sure if there's that much actual semantic overlap, so maybe we should avoid the duplicate variant names?

adamjstewart:
I agree that it's been frustrating that py-libclang hasn't been updated in a long time. Since py-libclang depends on llvm anyway, I wonder if we shouldn't just always use llvm and drop the variant.

wdconinc:
> I agree that it's been frustrating that py-libclang hasn't been updated in a long time. Since py-libclang depends on llvm anyway, I wonder if we shouldn't just always use llvm and drop the variant.

I'd be perfectly fine with that too. Waiting for some comments from others.

bernhardkaindl:
> @bernhardkaindl alerted me to https://github.com/spack/spack-packages/pull/5561 but I'm not sure if there's that much actual semantic overlap, so maybe we should avoid the duplicate variant names?

@wdconinc, alternatively, and maybe anyway, I should probably rename the new `libclang` variant I propose for LLVM to `libclang-dev` because it is about the development library support of the package, not the libclang itself (which is always present, it is where clang lives).

bernhardkaindl:
@wdconinc, I'll use another solution for #5561, so this PR does not conflict with it!

PR #5858, opened 2026-08-02
win-wdk: detect the machine-wide Windows Kits install instead of installing it

bernhardkaindl:
Summary
-------

Make `win-wdk` an external-detection-only package, consistent with its sibling `win-sdk` and `wgl` recipes, and point `WDKContentRoot` at the registered Windows Kits root instead of the Spack prefix.

Problem
-------

The WDK is not a standalone, relocatable product. It is one component of the single machine-wide Windows Kits installation rooted at `KitsRoot10` and registered under `HKLM\SOFTWARE\Microsoft\Windows Kits\Installed Roots`.

The recipe tried to install it into a Spack prefix with `/features + /quiet /installpath <prefix>`. Once any Kit component is registered, which is the normal state on any machine that has Visual Studio and the Windows SDK, the bootstrapper refuses that prefix outright:

```ml
e000: ERROR: Cannot set the path to <prefix> because another installation has
already been detected. Additional programs will have to be installed to the
previously selected install path C:\Program Files (x86)\Windows Kits\10\.
i007: Exit code: 0x3e9, restarting: No
```

Where the installer does proceed, it writes machine-wide rather than into the prefix, so nothing lands in the Spack prefix and the install is rejected after the fact:

```py
==> win-wdk: Executing phase: 'install'
  File "lib/spack/spack/builder.py", line 631, in sanity_check_prefix
    raise spack.error.InstallError(f"Install failed for {pkg.name}.  Nothing was installed!")
spack.error.InstallError: Install failed for win-wdk.  Nothing was installed!
```

There is no arrangement of installer arguments that fixes this, because the payload is machine-wide by design. A prefix populated this way could also never be relocated or redistributed through a build cache. Separately,
`setup_dependent_build_environment` exported `WDKContentRoot` pointing at the Spack prefix, which does not contain the Kit.

Solution
--------

Mark the package `has_code = False` and drop the download URLs, checksums, the installer invocation, and the hook that renamed the downloaded bootstrapper. `install()` now raises with guidance to install the WDK from Microsoft, matching the SDK version already on the system, and to register it with `spack external find win-wdk`. This mirrors how `win-sdk` and `wgl` already represent machine-wide Windows components.

External detection through `mmos.lib` is unchanged and continues to resolve both the version and the `plat` variants. `setup_dependent_build_environment` now resolves `KitsRoot10` from the registry and falls back to the prefix, so dependents such as `msmpi` receive the real Kit location instead of an unrelated path.

### DIFF
--- a/repos/spack_repo/builtin/packages/win_wdk/package.py
+++ b/repos/spack_repo/builtin/packages/win_wdk/package.py
@@ -3,13 +3,13 @@
 # SPDX-License-Identifier: (Apache-2.0 OR MIT)
 
 
-import glob
 import os
 import re
 
 from spack_repo.builtin.build_systems.generic import Package
 
 from spack.package import *
+from spack.util.windows_registry import RegistryError
 
 
 class WinWdk(Package):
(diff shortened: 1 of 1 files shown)


bernhardkaindl:
Converted to draft because @johnwparent has work in progress that reworks all of the Windows SDK, WDK, WGL, and MSVC installation, to allow them to be spack installable or at least play nicer with non spack installed versions which we should see soon.

PR #5860, opened 2026-08-03
py-resampy: fix the ^python@3.1.2: conflict typo and update the deps

bernhardkaindl:
#### Fix the dependencies of `0.3.0:` and correct the Python conflict range for the 0.2 releases.

Version 0.4.3 was added in 8d880cd9a2 but the recipe kept the dependency set of 0.2.2, and the conflict guarding the removal of the `imp` module was incorrectly written as `^python@3.1.2:`

- `imp` was removed in 3.12, so the needed range is updated.

- Also, resampy dropped `six` in 0.3 and updated the lower bounds of `numpy`, `numba`, and `scipy`, needs newer build dependencies and needs `importlib-resources` for Python < 3.9.

### DIFF
--- a/repos/spack_repo/builtin/packages/py_resampy/package.py
+++ b/repos/spack_repo/builtin/packages/py_resampy/package.py
@@ -11,7 +11,7 @@ class PyResampy(PythonPackage):
     """Efficient sample rate conversion in python"""
 
     homepage = "https://github.com/bmcfee/resampy"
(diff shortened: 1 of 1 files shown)


payerle (approved):
Sorry, the python@3.1.2 instead of python@3.12 was a mental lapse on my part.

Thank you for the version updates and I will try to do better in the future.

wdconinc (approved):

PR #5862, opened 2026-08-03
py-pytecplot: add v1.7.4

### LINKED ISSUE #5735: py-pytecplot: Bad URL and missing versions
Current `py-pytecplot` recipe uses an URL with [`.zip` extension](https://github.com/spack/spack-packages/blob/develop/repos/spack_repo/builtin/packages/py_pytecplot/package.py#L21). It seems that it does not work any more. Changing it to `tar.gz` seems to solve the issue. With that, `spack version` can also find all the recent [versions published on PyPI](https://pypi.org/project/pytecplot/#history).

RiMillo:
Address seems to have changed from `.zip` to `.tar.gz`. Fixing that allowed one to get new versions. The additional version have been obtained with `spack checksum -a`.

close #5735

### DIFF
--- a/repos/spack_repo/builtin/packages/py_pytecplot/package.py
+++ b/repos/spack_repo/builtin/packages/py_pytecplot/package.py
@@ -14,10 +14,22 @@ class PyPytecplot(PythonPackage):
     Tecplot 360 macro language is helpful, but not required."""
 
     homepage = "https://www.tecplot.com/docs/pytecplot/"
-    pypi = "pytecplot/pytecplot-1.4.2.zip"
+    pypi = "pytecplot/pytecplot-1.4.2.tar.gz"
 
     license("Frameworx-1.0")
 
+    version("1.7.4", sha256="cb4456f004267140a35925a4b46c41a20ec15051fcb2404ab3d6077d6cec18a7")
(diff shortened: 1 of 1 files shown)


RiMillo:
@saisharan0103

RiMillo:
@spackbot help

RiMillo:
@spackbot request review

RiMillo:
@spackbot request review

RiMillo:
Not sure why the checksum of a single version failed while everything was alright the first time I pushed it

bernhardkaindl:
@RiMillo, to fix #5735, you don't need all those new versions.

Unless you've a specific project that needs one of those, remove all versions this diff adds except for the latest version.

Also, I guess it would be nice if the previous versions arent specificaly required by a recipe, if those could be set to deprecated using `with default_args(deprecated=True):`

RiMillo:
> Unless you've a specific project that needs one of those, remove all versions this diff adds except for the latest version.
> 
> Also, I guess it would be nice if the previous versions arent specificaly required by a recipe, if those could be set to deprecated using `with default_args(deprecated=True):`

Hi @bernhardkaindl, thanks for your message. I removed as requested all new versions but latest one, and changed the title accordingly. I also deprecated the only version that was already there, although, I'm not sure I understood the second part of your message.

bernhardkaindl (approved):

bernhardkaindl:
@RiMillo, thanks all fine! I didn't look at close before, ack!

Issue #5866, opened 2026-08-03
Improve handling of cxxstd variants in RADIUSS packages


adrienbernede:
Raja expresses constraints on variant values using conflicts:

https://github.com/spack/spack-packages/blob/d0045f62713febbaf5bcec4b6506e6fae37ae171/repos/spack_repo/builtin/packages/raja/package.py#L260

Umpire expresses constraints on variant values with repeated definitions of the variant:

https://github.com/spack/spack-packages/blob/d0045f62713febbaf5bcec4b6506e6fae37ae171/repos/spack_repo/builtin/packages/umpire/package.py#L238
(Namely, detail the exact compatibility between versions and C++ standards without using conflicts).

Spack also provides "conditional" values to express constraints directly in the variant definition. 
https://spack.readthedocs.io/en/latest/packaging_guide_creation.html#conditional-possible-values

The goal here is to determine whether on solution is better than the other and why.

_Originally posted by @adrienbernede in https://github.com/spack/spack-packages/issues/5716#issuecomment-5167463052_

adrienbernede:
If I could, I would assign it to myself :)

PR #5870, opened 2026-08-03
RAJA suite packages: Convert conflicts to cxxstd variant conditionals

adrienbernede:
This PR translates constraints on the cxxstd variant to the conditional values syntax.

This use case is documented here:
https://spack.readthedocs.io/en/latest/packaging_guide_creation.html#conditional-possible-values

## Notes:
Solves #5866 

In particular, I tried to use conditional values AND conflicts, when appropriate:

1. Conditional values = "this value exists, but only here." Adds optimization guidance (penalties) to the solver. Lets the solver prefer the default value.
2. Conflicts = "this combination is invalid." Hard constraints that remove solutions from the search space. No optimization guidance.
3. They are complementary, not redundant. Conflicts handle dropping/forbidding. Conditional values handle version-dependent availability.
6. Avoid multiple version-gated variant definitions — the old pattern of variant("cxxstd", ..., when="@...") repeated for each version range is verbose, confusing, and produces multiple definitions the solver must reconcile.


repos/spack_repo/builtin/packages/chai/package.py
```diff
@@ -155,8 +155,15 @@ class Chai(CachedCMakePackage, CudaPackage, ROCmPackage):
     variant(
         "cxxstd",
         default="17",
-        values=("11", "14", "17", "20", "23"),
         description="C++ standard to build with",
+        values=(
+            conditional("11", when="@:1.0"),
```
adayton1: I think 1.0 should be changed to 2.4.0.

adrienbernede: Agreed!
```suggestion
            conditional("11", when="@:2.4.0"),
```

adrienbernede:
@adayton1 github mistakenly attributed your suggestion to myself. I owe you one ;)

adayton1 (approved):

repos/spack_repo/builtin/packages/care/package.py
```diff
@@ -126,11 +126,15 @@ class Care(CachedCMakePackage, CudaPackage, ROCmPackage):
     variant(
         "cxxstd",
         default="17",
-        values=("11", "14", "17", "20"),
         description="C++ standard to build with",
+        values=(
+            conditional("11", when="@:0.15.1"),
```
davidbeckingsale, line 131: How does conditional interact with variant default? e.g. older versions used to have different default C++ standard

adrienbernede: Default is preferred, so we lose the information "what is the preferred value _per version_" as it was expressed in Umpire packages only.

Conditionals add a _possible value_ for the variant under a condition, _not a constraint, not a preferrence_.

Our projects have defined c++ standard requirements. The current implementation reflects that:
- e.g., cxxstd 14 in not defined for versions that require c++17.
- we assuming using a newer variant is not a problem
- we can further restrict the values if newer standards break builds of older versions.

## Note:

This is the opposite to what is done in boost, which is taken as an example in spack documentation:
Boost cxxstd variant: https://github.com/spack/spack-packages/blob/2cea2c4fd477292766a4ef4bc191d235710c7454/repos/spack_repo/builtin/packages/boost/package.py#L215
Spack docs on conditional values in variants: https://spack.readthedocs.io/en/latest/packaging_guide_creation.html#conditional-possible-values

However, boost way of doing this seems backwards:
- they conditionally introduce new standard (instead of conditionally restricting older ones).
- they default to (spack will prefer) the oldest supported value unless a conflict excludes it.

bernhardkaindl (commented):
Hi @adrienbernede, as reasoned in https://github.com/spack/spack-packages/pull/5961#pullrequestreview-4953760166, I merged #5961.

In https://github.com/spack/spack-packages/pull/5961#discussion_r3793032587, @smithsg84 mentioned  that he'll work with you to update the conflicts about the `cxxstd` variants to variant conditionals in this PR.

With the merge of #5961, two `git` conflicts with the merge of it have now surfaced, but this was expected. I hope that it should be relatively easy for you to incorporate the merged changes (rebase this PR on the new develop merge of #5961) and as mentioned, @smithsg84 said he'll work with you on that.

smithsg84:
@bernhardkaindl, Thanks for review/merging #5961.   Adrien and I will get the conflicts resolved on this PR, as you note the conflicts are trivial.

adrienbernede:
@smithsg84 @bernhardkaindl conflicts solved.

bernhardkaindl (approved):
@smithsg84, @adrienbernede, @davidbeckingsale, @adayton1 Is this ready for merge?

adrienbernede:
@bernhardkaindl yes it is!

bernhardkaindl (approved):
Will merge! cc: @smithsg84, @adrienbernede, @davidbeckingsale, @adayton1

PR #5872, opened 2026-08-03
sfcgal: add 2.3.0, update dependency constraints

theAeon:
Seems this one was basically abandoned and nonfunctional. Attempted to touch it up.

### DIFF
--- a/repos/spack_repo/builtin/packages/sfcgal/package.py
+++ b/repos/spack_repo/builtin/packages/sfcgal/package.py
@@ -10,34 +10,42 @@ from spack.package import *
 class Sfcgal(CMakePackage):
     """
     SFCGAL is a C++ wrapper library around CGAL with the aim of supporting
-    ISO 19107:2013 and OGC Simple Features Access 1.2 for 3D operations. SFCGAL
-    provides standard compliant geometry types and operations, that can be
+    ISO 19107:2013 and OGC Simple Features Access 1.2 for 3D operations.
+    It provides standard compliant geometry types and operations, that can be
     accessed from its C or C++ APIs.
     """
 
     homepage = "http://www.sfcgal.org/"
-    url = "https://gitlab.com/sfcgal/SFCGAL/-/archive/v1.5.1/SFCGAL-v1.5.1.tar.gz"
-    # URL for versions up to 1.3.8
-    old_github_urlbase = "https://github.com/Oslandia/SFCGAL/archive/v{0}.tar.gz"
+    url = "https://gitlab.com/sfcgal/SFCGAL/-/archive/v2.3.0/SFCGAL-v2.3.0.tar.gz"
+    old_github_urlbase = "https://github.com/Oslandia/SFCGAL/archive/v{0}.tar.gz" # up to 1.3.8
 
     license("LGPL-2.0-or-later")
 
-    version("1.5.1", sha256="ea5d1662fada7de715ad564dc810c3059024ed81ae393f5352489f706fdfa3b1")
-    version("1.4.1", sha256="1800c8a26241588f11cddcf433049e9b9aea902e923414d2ecef33a3295626c3")
-    version(
-        "1.3.8",
-        sha256="5154bfc67a5e99d95cb653d70d2b9d9293d3deb3c8f18b938a33d68fec488a6d",
-        url=old_github_urlbase.format("1.3.8"),
-    )
-    version(
-        "1.3.7",
-        sha256="30ea1af26cb2f572c628aae08dd1953d80a69d15e1cac225390904d91fce031b",
-        url=old_github_urlbase.format("1.3.7"),
-    )
+    version("2.3.0", sha256="5f6aa1838e5ae31523ebf410cde0240b7a88d7e062b7ffff945e4fae2aaba0fa")
+    version("2.2.0", sha256="bb6bb77ddb58523d8c229764de23699f99c1a7011d873419afd2a67df85602a2")
 
-    depends_on("cxx", type="build")  # generated
+    with default_args(deprecated=True):
+        version("1.5.1", sha256="ea5d1662fada7de715ad564dc810c3059024ed81ae393f5352489f706fdfa3b1")
+        version("1.4.1", sha256="1800c8a26241588f11cddcf433049e9b9aea902e923414d2ecef33a3295626c3")
+        version(
+            "1.3.8",
+            sha256="5154bfc67a5e99d95cb653d70d2b9d9293d3deb3c8f18b938a33d68fec488a6d",
+            url=old_github_urlbase.format("1.3.8"),
+        )
+        version(
+            "1.3.7",
+            sha256="30ea1af26cb2f572c628aae08dd1953d80a69d15e1cac225390904d91fce031b",
+            url=old_github_urlbase.format("1.3.7"),
+        )
 
-    depends_on("cmake@2.8.6:", type="build")
+    variant("eigen", default=False, description="Enable Eigen support")
+
+    with default_args(type="build"):
+        depends_on("c")
+        depends_on("cxx")
+        depends_on("cmake@2.8.6:")
+        depends_on("nlohmann-json@3.11:", when="@2.3:")
+
     # Ref: https://oslandia.github.io/SFCGAL/installation.html, but starts to work @4.7:
     # Ref: https://gitlab.com/sfcgal/SFCGAL/-/blob/v1.5.1/NEWS?ref_type=tags
     # and looking at CMakeLists.txt find_package(CGAL) declaration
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/sfcgal/package.py
```diff
@@ -21,7 +21,11 @@ class Sfcgal(CMakePackage):
     old_github_urlbase = "https://github.com/Oslandia/SFCGAL/archive/v{0}.tar.gz"
 
     license("LGPL-2.0-or-later")
-
+    version("2.3.0", sha256="5f6aa1838e5ae31523ebf410cde0240b7a88d7e062b7ffff945e4fae2aaba0fa")
+    version("2.2.0", sha256="bb6bb77ddb58523d8c229764de23699f99c1a7011d873419afd2a67df85602a2")
+    version("2.1.0", sha256="cb73a0496c61a5c7bf0ccc68c42e4378bfc441b242e9dee894067e24d2e21d0f")
+    version("2.0.0", sha256="11843953f49e7e4432c42fd27d54e1ff7ca55d0cc72507725c2a5d840c2c6535")
+    version("1.5.2", sha256="b946b3c20d53f6e2703046085f0fcfea6c1a4081163f7bedd30b1195801efdd2")
```
bernhardkaindl: Hi @theAeon, spack accumulates versions over time, but unless there is a real need to add lots of versions at once, it would be better to only add the latest version (those need to be parsed, checked, maintained)

Suggestion: If you don't actively need those added versions, could you remove them from the PR?
```suggestion
```
Of course, if you could need them or specific versions, add them. In any case, keep the dependency version constraints for them because then, you can still build them by requesting their specific version if needed: `spack install packagename@version`, even if the version isn't listed.

repos/spack_repo/builtin/packages/sfcgal/package.py
```diff
@@ -35,7 +39,9 @@ class Sfcgal(CMakePackage):
         url=old_github_urlbase.format("1.3.7"),
     )
```
bernhardkaindl: @theAeon, because the package was basically abandoned and nonfunctional, this would be the best opportunity to deprecated the old versions which were previously non-functional:

```suggestion
    with default_args(deprecated=True):
        version("1.5.1", sha256="ea5d1662fada7de715ad564dc810c3059024ed81ae393f5352489f706fdfa3b1")
        version("1.4.1", sha256="1800c8a26241588f11cddcf433049e9b9aea902e923414d2ecef33a3295626c3")
        version(
            "1.3.8",
            sha256="5154bfc67a5e99d95cb653d70d2b9d9293d3deb3c8f18b938a33d68fec488a6d",
            url=old_github_urlbase.format("1.3.8"),
        )
        version(
            "1.3.7",
            sha256="30ea1af26cb2f572c628aae08dd1953d80a69d15e1cac225390904d91fce031b",
            url=old_github_urlbase.format("1.3.7"),
        )
```

They can be built with --deprecated, and after the next spack release with these deprecations in place, the the old versions and code for them removed, which would make the package slimmer less effort to maintain.

repos/spack_repo/builtin/packages/sfcgal/package.py
theAeon: Gotcha. Feels a bit weird to me to not hardcode sha256sums from a security perspective, but if that's how the project handles it generally, that's fine.

bernhardkaindl: Ah, understand! I'm sorry mis-remembered `spack install packagename@version` doesn't work, it would have to be a package (or some environment) that would have to dicate a version.

I think the idea is (I assume https://spack.readthedocs.io/en/latest/ might explain more) that versions which are not listed are not supported and I did see a warning when spack tried to fetch a version without a sha265sum. I guess it might be configurable in the config yaml.

bernhardkaindl (dismissed):

repos/spack_repo/builtin/packages/sfcgal/package.py
```diff
@@ -21,21 +21,30 @@ class Sfcgal(CMakePackage):
...
 
     license("LGPL-2.0-or-later")
-
+    version("2.3.0", sha256="5f6aa1838e5ae31523ebf410cde0240b7a88d7e062b7ffff945e4fae2aaba0fa")
     version("1.5.1", sha256="ea5d1662fada7de715ad564dc810c3059024ed81ae393f5352489f706fdfa3b1")
     version("1.4.1", sha256="1800c8a26241588f11cddcf433049e9b9aea902e923414d2ecef33a3295626c3")
     version(
         "1.3.8",
         sha256="5154bfc67a5e99d95cb653d70d2b9d9293d3deb3c8f18b938a33d68fec488a6d",
         url=old_github_urlbase.format("1.3.8"),
```
bernhardkaindl: @theAeon, my 2nd suggestion should also have removed the version lines it is replacing:
```suggestion
```
This could have been an oversight by me but I also heard that GitHub suggestions suffered from a bug recently.

I any case, after you apply a change re-test the recipe for the versions that are changes to ensure they still work (you need to install the deprecated versions with `spack install --deprecated` and you fully understand the change.

This also caused a syntax error because the 1st, duplicate version statement of 1.3.8 was now incomplete (the `)`-line was removed by the change):
```py
    version(
        "1.3.8",
        sha256="5154bfc67a5e99d95cb653d70d2b9d9293d3deb3c8f18b938a33d68fec488a6d",
        url=old_github_urlbase.format("1.3.8"),
```
Here is the CI check error for it:
```py
ruff check repos/spack_repo/builtin/packages/sfcgal/package.py
invalid-syntax: Expected `)`, found newline
  --> repos/spack_repo/builtin/packages/sfcgal/package.py:30:48
   |
28 |         "1.3.8",
29 |         sha256="5154bfc67a5e99d95cb653d70d2b9d9293d3deb3c8f18b938a33d68fec488a6d",
30 |         url=old_github_urlbase.format("1.3.8"),
   |                                                ^
31 |     with default_args(deprecated=True):
32 |         version("1.5.1", sha256="ea5d1662fada7de715ad564dc810c3059024ed81ae393f5352489f706fdfa3b1")
   |

E501 Line too long (110 > 99)
```
This means you didn't test-build the package as after this commit, because the Python syntax of it was broken, and spack would have told you the exact place and reason:

```py
spack install --overwrite -y sfcgal
==> Error: cannot load package 'sfcgal' from the 'builtin' repository: '(' was never closed (package.py, line 27)
```
The incomplete suggestion would possibly have been my fault or a simple glitch on GitHub, but please check the build after applying suggestions. Use `git pull` to pull the new commit from your GitHub repo's PR branch if you applied it using the "Apply suggestion" button on github.com.

repos/spack_repo/builtin/packages/sfcgal/package.py
```diff
@@ -21,21 +21,30 @@ class Sfcgal(CMakePackage):
...
+            url=old_github_urlbase.format("1.3.7"),
+        )
 
-    depends_on("cxx", type="build")  # generated
+    variant("eigen", default=False, description="Enable Eigen support")
+    depends_on("c", type="build")
+    depends_on("cxx", type="build")
 
     depends_on("cmake@2.8.6:", type="build")
     # Ref: https://oslandia.github.io/SFCGAL/installation.html, but starts to work @4.7:
```
bernhardkaindl, line 49: To make this less confusing and more readable, the best way would be to leave an an empty line beween the versions, the variants (even if only one) and the dependencies.

This the best practise which I can see from other packages and the guidance I found. While at it, also group the build dependencies as one block and separate the build dependencies below from the start of the block of other dependencies.

Even better, use a context manager to apply `type="build"` once for all build dependencies:
```suggestion
    variant("eigen", default=False, description="Enable Eigen support")

    with default_args(type="build"):
        depends_on("cxx")
        depends_on("cmake@2.8.6:")

# Ref: https://oslandia.github.io/SFCGAL/installation.html, but starts to work @4.7:
```
This helps a lot when reading the final recipe! 👍

repos/spack_repo/builtin/packages/sfcgal/package.py
```diff
@@ -47,13 +56,18 @@ class Sfcgal(CMakePackage):
...
-    depends_on("cgal@5.6", when="@1.5")
+    depends_on("cgal@5.3", when="@1.4:1.5")
+    depends_on("cgal@5.6:6.0", when="@2.0:2.2")
+    depends_on("cgal@5.6:6.2", when="@2.3")
+
     depends_on(
-        "boost@1.54.0:+chrono+filesystem+program_options+serialization+system+test+thread+timer"
+        "boost@1.54.0:+chrono+filesystem+program_options+serialization+system+test+thread+timer", when ="@1.3"
     )
+    depends_on("boost@1.74.0:+chrono+filesystem+program_options+serialization+system+thread+timer", when="@1.4:")
```
bernhardkaindl: These long lines are not accepted by `ruff check` of the file in spack.
Please install the ruff formatter https://docs.astral.sh/ruff/installation/ and run the formatter after you fixed the syntax error that was left over after applying the initial review comments (A formatter can't fix the syntax error because of the duplicated versions lines):
```py
ruff format repos/spack_repo/builtin/packages/sfcgal/package.py
```
If you don'e want to de that, should look like this afterwards:
```suggestion
        "boost@1.54.0:+chrono+filesystem+program_options+serialization+system+test+thread+timer",
        when="@1.3",
    )
    depends_on(
        "boost@1.74.0:+chrono+filesystem+program_options+serialization+system+thread+timer",
        when="@1.4:",
```
After the syntax error is fixed and pushed, also `@spackbot fix style` could do that for you, but it is much slower and the duplicated CI runs cause unneeded churn, so I'd highly encourage you to install ruff, apply all suggestions I made here yourself in your IDE/editor (thay way, you can do them all at once, without triggering more CI runs) and let `ruff format` do the formatting (yes, like `black`, `ruff format` is an opinionated formatter, and share the same format, only differnet configs, so you'll get "the" only correct and only format accepted by CI, so run `ruff format` on it). Then commit all changes, check everything is comitted and then test the updated commit to build all versions with `spack install --overwrite -y` and the depreacated version with `--deprecated`, only then, push it to GitHub for review. 👍 💯%🥇

repos/spack_repo/builtin/packages/sfcgal/package.py
bernhardkaindl: applied now, but keeping the 1.5.1 and 1.4.1 not deprecated.

repos/spack_repo/builtin/packages/sfcgal/package.py
bernhardkaindl: fixed

repos/spack_repo/builtin/packages/sfcgal/package.py
bernhardkaindl: fixed.

repos/spack_repo/builtin/packages/sfcgal/package.py
bernhardkaindl: Applied, except "c" is needed, it fails without. Kept "c".

bernhardkaindl (approved):

PR #5873, opened 2026-08-03
Kokkos: Conflict non supported archs for CUDA and HIP

JBludau:
This PR changes the kokkos package to only display the CUDA and HIP archs we actually support.
Furthermore, it causes an error if users do not supply a gpu arch but enable the backend (added for HIP and SYCL, CUDA was already in place).


rbberger:
@JBludau this is ok, but I believe the "early" warnings never trigger in this case since the acceptable values are limited.

PR #5887, opened 2026-08-04
spack: add v0.22.0 -> v1.2.2

cedricjourdain:
We offer our users different versions of Spack; depending on the version used, Spack’s configuration varies depending on whether we provided it to them at a given point in time or whether they have used a previous version of Spack. That is why, within our software stack, we have deployed several versions of Spack so that users can use an older version of Spack.


cedricjourdain:
@haampie Is that a bad idea?

PR #5891, opened 2026-08-04
petsc: support multiple CUDA architectures

Sbozzolo:
Pass the full cuda_arch list to PETSc's CUDA architecture option.

commit in 3.19 that adds support for multi arch: https://gitlab.com/petsc/petsc/-/commit/f23e50abecd9f61bbbae4cd7d74ef3d78d3a5b23


bernhardkaindl (approved):
Appoving, looks good to me.

balay:
`with-cuda-gencodearch` was changed to `cuda-arch` in petsc-3.16. However `with-cuda-gencodearch`  continues to work [so this change is fine]

I guess we need to update this code [to switch to using  `cuda-arch`] at some point..

PR #5650, opened 2026-07-19
py-wxpython: add v4.2.5

wdconinc:
This PR upgrades and fixes `py-wxpython` to v4.2.4 and v4.2.5. Since `py-wxpython` pins a specific version of `wxwidgets` we add the necessary versions. These new versions add support for free-threaded python (i.e. it now compiles on those pythons). URL on PyPI changed.

Test build (against free-threaded python, fast due to ccache):
```
[+] exwg3lw py-wxpython@4.2.5 /opt/software/linux-skylake/py-wxpython-4.2.5-exwg3lwwqy7ae7tohqnmvertb73pojzh (4m01s)
```

PR #5893, opened 2026-08-04
py-wxwidgets: Fix setuptools and Python bounds of older releases

bernhardkaindl:
Hi @wdconinc, as mentioned in it, a follow-up after the review of #5650 updating this package:

wxPython 4.1 only supported up to Python 3.9, add a constraint for it.

Setuptools 81 removed the 8th argument from copy_file, and older wxPython 4.2 versions used that. Add a constraint for it.

Move all build dependencies below single context manager.


wdconinc (approved):
Thanks for the fix. This looks good to me.

PR #5894, opened 2026-08-04
blaspp: support multiple CUDA architectures

Sbozzolo:
Pass the full cuda_arch list to native CUDA builds.

I also updated the cmake version to sync up with the cmakefile


mgates3 (approved):
Looks good. Thanks.

bernhardkaindl (approved):
Approving per maintainer approval.

PR #5896, opened 2026-08-04
meson: py-setuptools is a build-time-only dep of Meson

bernhardkaindl:
AFAICS, Meson's runtime dependencies were always only Python and Ninja
shown on the top of https://github.com/mesonbuild/meson/blob/master/README.md

I verified that no version of Meson used setuptools at runtime using
```py
git clone https://github.com/mesonbuild/meson.git
git -C meson log -p --pretty=oneline . ':(exclude)setup.py' | grep ' setuptools'
```
so it should be removed from the runtime dependencies.

### DIFF
--- a/repos/spack_repo/builtin/packages/meson/package.py
+++ b/repos/spack_repo/builtin/packages/meson/package.py
@@ -39,7 +39,7 @@ class Meson(PythonPackage):
         version("1.8.2", sha256="6b878fb0f6f0318cbd54e13539f89a1a8305791668e8e93ffd59d82722888dac")
         version("1.7.0", sha256="a6ca46e2a11a0278bb6492ecd4e0520ff441b164ebfdef1e012b11beb848d26e")
 
-    depends_on("py-setuptools@42:", type=("build", "run"))
+    depends_on("py-setuptools@42:", type="build")
     depends_on("ninja@1.8.2:", type="run")
 
     # By default, Meson strips the rpath on installation. This patch disables


eli-schwartz (approved):
The odd runtime dependency was added in https://github.com/spack/spack/pull/15674 seemingly with the rationale that it is needed "because setuptools generated entrypoint scripts import pkg_resources".

bernhardkaindl:
@eli-schwartz, thanks for the approval!

> The odd runtime dependency was added in https://github.com/spack/spack/pull/15674 seemingly with the rationale that it is needed "because setuptools generated entrypoint scripts import pkg_resources".

Thanks, yes the discussion later confirmed in 2023 that this wasn't the case (anymore).

`py-setuptools@81:` also removed `pkg_resources`, so with those newer setuptools that entry point doesn't exist anymore either.

There are packages that use setuptools at runtime, in their non-generated python code, but it is those very specific packages that would need the runtime dependency, not Meson itself: As Meson itself is only a build time dependency of such packages, the run dependency would not have helped indirectly either.

eli-schwartz:
Yup. I was mildly surprised that individual packages would seemingly be adding manual dependencies for this at all -- it seems like the kind of thing that should be automatically injected by a build class, and then removed globally in a single action once setuptools stopped doing that.

eli-schwartz:
> There are packages that use setuptools at runtime, in their non-generated python code, but it is those very specific packages that would need the runtime dependency, not Meson itself:

Old versions of meson did do this too incidentally, but only:
- as a side effect of importing the stdlib distutils, for versions of cpython that removed distutils
- to extract a small piece of data needed when compiling python extension modules

So it's also not really about that, since a) that is better declared via py-meson-python, b) it would be python-version-specific in addition to meson-version-specific.

See https://github.com/spack/spack-packages/blob/d2a77cbf4d0c0f14bc7e66bcaee1e9cb73778bc9/repos/spack_repo/builtin/packages/py_meson_python/package.py#L52

michaelkuhn (approved):
LGTM

PR #5897, opened 2026-08-04
py-omegaconf: add 2.3.1 and update version constraints:

bernhardkaindl:
`py-omegaconf`: add 2.3.1 (to not depend on old `py-setuptools`) and update version constraints:
- `py-omegaconf` before `2.3.1` depends on `pkg_resources` by `py-setuptools@:80`
- sha265 of `2.2.2` has changed (verified)
- `2.1.0` only supports up to Python 3.9
- `2.2.2` only supports up to Python 3.10
- Refactor build dependencies to use `default_args(type="build")`

Cc: @calebrob6, @LydDeb

### DIFF
--- a/repos/spack_repo/builtin/packages/py_omegaconf/package.py
+++ b/repos/spack_repo/builtin/packages/py_omegaconf/package.py
@@ -14,19 +14,30 @@ class PyOmegaconf(PythonPackage):
     """
 
     homepage = "https://github.com/omry/omegaconf"
-    pypi = "omegaconf/omegaconf-2.3.0.tar.gz"
+    pypi = "omegaconf/omegaconf-2.3.1.tar.gz"
 
     maintainers("calebrob6")
 
     license("BSD-3-Clause")
 
+    version("2.3.1", sha256="e5e7de64aeebeddaf8e6d3f7a783b32ac2a01c0fbd9c878012caecb891a1f42a")
     version("2.3.0", sha256="d5d4b6d29955cc50ad50c46dc269bcd92c6e00f5f90d23ab5fee7bfca4ba4cc7")
-    version("2.2.2", sha256="10a89b5cb81887d68137b69a7c5c046a060e2239af4e37f20c3935ad2e5fd865")
-    version("2.1.0", sha256="a08aec03a63c66449b550b85d70238f4dee9c6c4a0541d6a98845dcfeb12439d")
+    version("2.2.2", sha256="65c85b2a84669a570c70f2df00de3cebcd9b47a8587d3c53b1aa5766bb096f77")
+    with default_args(deprecated=True):
+        version("2.1.0", sha256="a08aec03a63c66449b550b85d70238f4dee9c6c4a0541d6a98845dcfeb12439d")
 
-    depends_on("py-setuptools", type="build")
-    depends_on("py-pytest-runner", when="@2.1", type="build")
+    # https://github.com/omry/omegaconf/releases#release-v2.1.0
+    conflicts("python@3.10:", when="@:2.1", msg="Use Omegaconf >= 2.2.2 for Python 3.10 and newer")
+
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/py_omegaconf/package.py
```diff
@@ -24,7 +24,7 @@ class PyOmegaconf(PythonPackage):
     version("2.2.2", sha256="10a89b5cb81887d68137b69a7c5c046a060e2239af4e37f20c3935ad2e5fd865")
     version("2.1.0", sha256="a08aec03a63c66449b550b85d70238f4dee9c6c4a0541d6a98845dcfeb12439d")
 
-    depends_on("py-setuptools", type="build")
+    depends_on("py-setuptools@59.6:80", type="build")  # for pkg_resources
```
LydDeb: The 'pkg_resources' module is not present in newer versions of setuptools.

I do indeed see the `pkg_resources` import in version 2.3.0 (cf. https://github.com/omry/omegaconf/blob/350bdb632865c5dd2286f2f6521acefe4abd843d/setup.py#L13 )
However, it is not present in version 2.3.1 and on the 'main' branch.
Note: it reappears on v2.4.0dev4 (cf. https://github.com/omry/omegaconf/blob/28ad5c0d10470dff7c2a1a7622df0e84d7aa276b/setup.py#L14 )

LydDeb (approved):

repos/spack_repo/builtin/packages/py_omegaconf/package.py
bernhardkaindl: Thanks for confirming!

As "py-setuptools@59.6:80" is only a build time dependency, it is not a constraint affecting other packages in the DAG or other builds, so we can use it unconditionally without adding new versions and fleshing out which `when=` versions don't need it.

wdconinc: I do think restricting this to the versions that need it is useful. For example, someone may have `py-setuptools@81:` in their spack, site, or environment packages.yaml preferences, and that applies to build deps as well. I have exactly that line :-)

bernhardkaindl: Hi @wdconinc, done!

### This required to add the current release 2.3.1 which fixed this issue:

Updated the PR:

`py-omegaconf`: add 2.3.1 (to not depend on old `py-setuptools`) and update version constraints:
- `py-omegaconf` before `2.3.1` depends on `pkg_resources` by `py-setuptools@:80`
- sha265 of `2.2.2` has changed (verified)
- `2.1.0` only supports up to Python 3.9: https://github.com/omry/omegaconf/releases#release-v2.1.0
- `2.2.2` only supports up to Python 3.10: https://github.com/omry/omegaconf/releases#release-v2.3.0
- Refactor build dependencies to use `default_args(type="build")`

bernhardkaindl:
@wdconinc updated the PR as suggested in https://github.com/spack/spack-packages/pull/5897#discussion_r3720904160 to not always depend on `py-setuptools@:80`:

### This required to add the current release 2.3.1 which fixed this issue:

Updated the PR:

`py-omegaconf`: add 2.3.1 (to not depend on old `py-setuptools`) and update version constraints:
- `py-omegaconf` before `2.3.1` depends on `pkg_resources` by `py-setuptools@:80`
- sha265 of `2.2.2` has changed (verified)
- `2.1.0` only supports up to Python 3.9: https://github.com/omry/omegaconf/releases#release-v2.1.0
- `2.2.2` only supports up to Python 3.10: https://github.com/omry/omegaconf/releases#release-v2.3.0
- Refactor build dependencies to use `default_args(type="build")`

@wdconinc, thanks for the review, can you re-review it now?

wdconinc (approved):

PR #5900, opened 2026-08-04
py-nvtx: Pass spack's nvtx install location to the build

bernhardkaindl:
py-nvtx needs the nvtx headers, but they weren't passed to the build.

Draft: LydDeb will submit another, superseding PR which will also contain a version update.


LydDeb (approved):
Starting with version 0.2.11, the include directory is packaged in the PyPI source distribution.
This PR therefore successfully resolves the installation issue for the version currently available in the recipe (0.2.10).

bernhardkaindl:
Superseded by #5926

PR #5901, opened 2026-08-04
heffte: update FFTW dependency to use fftw-api

cedricjourdain:
Switch the +fftw dependency from a hardcoded fftw@3.3.8: to the fftw-api virtual package. Depending directly on fftw prevents using alternative FFTW-API providers (e.g. amdfftw, cray-fftw, etc)

### DIFF
--- a/repos/spack_repo/builtin/packages/heffte/package.py
+++ b/repos/spack_repo/builtin/packages/heffte/package.py
@@ -70,11 +70,7 @@ class Heffte(CMakePackage, CudaPackage, ROCmPackage):
 
     # Only the fftw@3.3.8: backend implementation is verified in current testing.
     # Alternative fftw-api providers such as amdfftw and cray-fftw have not been tested.
-    depends_on(
-        "fftw-api",
-        when="+fftw",
-        type=("build", "run")
-    )
+    depends_on("fftw-api", when="+fftw", type=("build", "run"))
     depends_on("intel-oneapi-mkl", when="+mkl", type=("build", "run"))
     # mkl renamed dfti.hpp to dft.hpp in 2026.0, which breaks heffte@2.4.1 and earlier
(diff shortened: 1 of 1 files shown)


mkstoyanov (approved):
On a base principle of things, I'm OK with the change. However, using alternative APIs quirks come up in the edge cases and we have not tested all of the possibilities.

I would suggest that we add a comment in the spack package, only fftw x.y.z is tested. Just to have that documented.

repos/spack_repo/builtin/packages/heffte/package.py
```diff
@@ -68,7 +68,12 @@ class Heffte(CMakePackage, CudaPackage, ROCmPackage):
 
     depends_on("mpi", type=("build", "run"))
 
-    depends_on("fftw@3.3.8:", when="+fftw", type=("build", "run"))
+    # Only the fftw@3.3.8 and rocfft@6.4.3 backend implementation used in our current testing is verified.
```
mkstoyanov: The PR isn't touching the rocfft backend, why is it there?

Otherwise, this is what I had in mind.

cedricjourdain: No, it’s just that, in my case, I made these changes so that I could use rocfft instead of fftw. If you’d prefer, I can remove the reference to rocfft@6.4.3

mkstoyanov: We have tested newer versions of rocfft, so the comment is not accurate in this regard. We have not tested alternatives to fftw, such as amdfftw and cray-fftw. That's what the comment should reflect.

alalazo (dismissed):

alalazo:
@spackbot fix style

mkstoyanov (approved):

bernhardkaindl (approved):
Approving per maintainer approval.

PR #5904, opened 2026-08-04
yafyaml: skip rpath for NAG compiler

mathomp4:
NAG's nested linker-argument syntax cannot represent Spack's padded install-prefix placeholder in an rpath, causing the build to fail.

yaFyaml only installs static libraries, so build rpaths are not needed and CMAKE_SKIP_RPATH can be set safely.


bernhardkaindl:
@tclune, do you want to review this PR and need more time to review it? Thanks, Bernhard

tclune (approved):

bernhardkaindl (approved):
Approving per maintainer approval

PR #5905, opened 2026-08-04
cargo-c: add thru v0.10.24

wdconinc:
This PR adds to `cargo-c` the versions through v0.10.24. All versions are added since there's a close connection between MSRV and new `cargo-c` releases, and this allows for the most recent version depending on what `rust` version people may prefer (e.g. as external).

Test build:
```
[+] 5hzw3bq cargo-c@0.10.22 /opt/software/linux-skylake/cargo-c-0.10.22-5hzw3bqnxpvqppwbum2pjdc4zbb3vrcq (11m26s)
```


RikkiButler20 (approved):
LGTM

PR #5906, opened 2026-08-04
py-scipy: disable GNU version script for NAG compiler

mathomp4:
NAG's linker forwards GNU linker flags (like --version-script) to GCC without the required -Wl, prefix, causing the link step to fail. Add a patch to scipy's meson.build that excludes nagfor from the version-script code path, matching the existing clang-cl exclusion.

A PR has been submitted upstream to scipy to fix this in the source directly: https://github.com/scipy/scipy/pull/25804

~The patch applies cleanly to scipy 1.17.0, 1.17.1, 1.18.0, and main. I've done `@1.17:` since, well, I'm not sure if it'll get into v2.0.0 or not. Depends on the upstream.~

scipy 2.0 won't depend on Fortran (see https://github.com/scipy/scipy/pull/25804#issuecomment-5181768203) So I've clamped to v1.

Note: GPT 5.6 Terra helped me figure out this fix.

### DIFF
--- a/repos/spack_repo/builtin/packages/py_scipy/nag_disable_version_script.patch
+++ b/repos/spack_repo/builtin/packages/py_scipy/nag_disable_version_script.patch
@@ -1,0 +1,5 @@
+--- a/meson.build
++++ b/meson.build
+@@ -168 +168 @@
+-if cc.get_id() != 'clang-cl'
++if cc.get_id() != 'clang-cl' and ff.get_id() != 'nagfor'
--- a/repos/spack_repo/builtin/packages/py_scipy/package.py
+++ b/repos/spack_repo/builtin/packages/py_scipy/package.py
@@ -189,6 +189,10 @@ class PyScipy(PythonPackage):
         when="@1.8.0:1.14.0",
     )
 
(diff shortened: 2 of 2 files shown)


mathomp4:
Update! Per @j-bowhay [on my scipy PR](https://github.com/scipy/scipy/pull/25804#issuecomment-5181768203):

> FYI, from SciPy 2.0 a Fortran compiler will no longer be needed.

So I've clamped the patch to v1

rgommers (commented):
The patch seems fine to me. The PR description says it applies cleanly to 1.17.x and 1.18.x, which is the kind of thing I'd worry about. This linker script is also present on older SciPy versions, so the 1.17 lower bound seems more like a "I don't care about older versions" rather than it being a problem introduced in 1.17

The `1.17:1` syntax still looks foreign to me, but that's probably just me:) Might want to spell it as `1.17:1.18` for clarity perhaps.

Either way, I'm happy if @adamjstewart is happy

mathomp4:
@rgommers i can take a look and see when the current meson.build existed. The patch should be good for any version from then on.

adamjstewart (approved):

PR #5907, opened 2026-08-04
pipx: Update and modernize

ebagrenrut:
Overhauls the `pipx` package, which has not been actively
maintained. I am adopting it, and have updated it to support `pipx` up
through the latest release. In addition to introducing newer versions of
`pipx`, and their dependencies, a `backend` variant is introduced to
account for the fact that `pipx` supports both `pip` and `uv` as
backends. `pipx` since 1.12.0 has defaulted to the `uv` backend, so this
updated package does the same. The user can require it use the `pip`
backend by adding `backend=pip` to their spec.

@adamjstewart I am interested in your review, when you have the time.

### DIFF
--- a/repos/spack_repo/builtin/packages/pipx/package.py
+++ b/repos/spack_repo/builtin/packages/pipx/package.py
@@ -71,7 +71,8 @@ class Pipx(PythonPackage):
     # When using uv backend, py-pip is needed for building, else it is needed for building and
     # running
     depends_on("py-pip", when="backend=uv", type="build")
-    depends_on("py-pip", type=("build", "run"))
+    depends_on("py-pip", when="backend=pip", type=("build", "run"))
+    depends_on("py-pip", when="@:1.11", type=("build", "run"))
 
     depends_on("py-platformdirs@2.1:", when="@1.3:", type=("build", "run"))
 
@@ -84,12 +85,11 @@ class Pipx(PythonPackage):
     depends_on("py-uv@0.4:", when="@1.12: backend=uv", type="run")
 
     def setup_run_environment(self, env):
-        if self.spec.satisfies("backend=pip"):
-            # If uv is not desired, force pip backend. This avoids pipx using any other uv found on
-            # PATH.
-            env.set("PIPX_DEFAULT_BACKEND", "pip")
+        if self.spec.satisfies("@1.12:"):
+            backend = self.spec.variants["backend"].value
 
-        elif self.spec.satisfies("backend=uv"):
-            # Ensure Spack-installed uv is used
-            env.set("PIPX_DEFAULT_BACKEND", "uv")
-            env.set("PIPX_UV_BINARY", join_path(self.spec["py-uv"].prefix.bin, "uv"))
+            env.set("PIPX_DEFAULT_BACKEND", backend)
+
+            if backend == "uv":
+                # Ensure Spack-installed uv is used
+                env.set("PIPX_UV_BINARY", self.spec["py-uv"].prefix.bin.uv)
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/pipx/package.py
```diff
@@ -15,14 +15,81 @@ class Pipx(PythonPackage):
...
+    depends_on("py-hatchling@0.15:", when="@:1.3.1", type="build")
+
+    depends_on("py-importlib-metadata@3.3:", when="^python@:3.7", type=("build", "run"))
+
+    depends_on("py-packaging@20:", type=("build", "run"))
+
+    # When using uv backend, py-pip is needed for building, else it is needed for building and
+    # running
+    depends_on("py-pip", when="backend=uv", type="build")
+    depends_on("py-pip", type=("build", "run"))
```
adamjstewart: This code does not match the comment. At the moment, all versions/variants require pip for build/run. I think you want:
```suggestion
    depends_on("py-pip", when="backend=pip", type=("build", "run"))
    depends_on("py-pip", when="@:1.11", type=("build", "run"))
```

repos/spack_repo/builtin/packages/pipx/package.py
```diff
@@ -15,14 +15,81 @@ class Pipx(PythonPackage):
...
+
+    def setup_run_environment(self, env):
+        if self.spec.satisfies("backend=pip"):
+            # If uv is not desired, force pip backend. This avoids pipx using any other uv found on
+            # PATH.
+            env.set("PIPX_DEFAULT_BACKEND", "pip")
+
+        elif self.spec.satisfies("backend=uv"):
+            # Ensure Spack-installed uv is used
+            env.set("PIPX_DEFAULT_BACKEND", "uv")
```
adamjstewart: Could simplify the first half of this to:
```python
env.set("PIPX_DEFAULT_BACKEND", spec.variants["backend"].value)
```
Might need to guard with:
```python
if self.spec.satisfies("@1.12:"):
```
since the variant didn't always exist.

repos/spack_repo/builtin/packages/pipx/package.py
```diff
@@ -15,14 +15,81 @@ class Pipx(PythonPackage):
...
+            env.set("PIPX_UV_BINARY", join_path(self.spec["py-uv"].prefix.bin, "uv"))
```
adamjstewart: Simpler:
```suggestion
            env.set("PIPX_UV_BINARY", self.spec["py-uv"].prefix.bin.uv)
```

adamjstewart (commented):
Didn't check the deps, just some obvious bugs

repos/spack_repo/builtin/packages/pipx/package.py
ebagrenrut: Resolved in commit 6a77635.

repos/spack_repo/builtin/packages/pipx/package.py
ebagrenrut: Resolved in commit 6a77635.

repos/spack_repo/builtin/packages/pipx/package.py
ebagrenrut: Resolved in commit 6a77635.

ebagrenrut:
@adamjstewart Thanks for the great suggestions!

Regarding the dependencies, I've been maintaining this package internal to my org for the past couple of years. Just to ensure I was handling he deps properly, I did go back through the `pyproject.toml` for all of the included releases to ensure I had things as accurate as I could make them. Certainly, I welcome any corrections.

adamjstewart (approved):

Issue #3533, opened 2026-02-24
Installation issue: gcc@15.2.0 +nvptx


JonathanFrassineti:
### Steps to reproduce the issue

@alalazo @michaelkuhn 

In an environment with gcc@11.5.0 as the system compiler:
```console
spack:
  concretizer:
    targets:
      host_compatible: false
    unify: true
  packages:
    all:
      require: target=zen4
      variants:
      - +cuda
      - cuda_arch=90
  definitions:
  - current_compiler:
    - gcc@11.5.0 +bootstrap -nvptx
  - cuda_compiler:
    - cuda@13.0.2
  - gcc_compiler:
    - gcc@15.2.0 -bootstrap +nvptx +binutils
  - all:
    - $gcc_compiler
  - out_of_matrix:
    - $cuda_compiler
  specs:
  - matrix:
    - - $all
    - - $%current_compiler
  - $out_of_matrix
  view: false
  modules:
    default:
      enable: []
```
Concretization output:
```
 -   gt243rr  gcc@15.2.0+binutils~bootstrap~graphite~mold+nvptx~piclibs~strip build_system=autotools build_type=RelWithDebInfo languages:='c,c++,fortran' platform=linux os=rhel9 target=zen4 %c,cxx=gcc@11.5.0
[+]  ued5ku3      ^gawk@5.3.1~nls build_system=autotools platform=linux os=rhel9 target=zen4 %c=gcc@11.5.0
[+]  woc52vk      ^gmp@6.3.0+cxx build_system=autotools libs:=shared,static platform=linux os=rhel9 target=zen4 %c,cxx=gcc@11.5.0
[+]  f5wqmi7      ^mpc@1.3.1 build_system=autotools libs:=shared,static platform=linux os=rhel9 target=zen4 %c=gcc@11.5.0
[+]  vxgnls5        ^mpfr@4.2.1 build_system=autotools libs:=shared,static patches:=3ec29a6 platform=linux os=rhel9 target=zen4 %c=gcc@11.5.0
[+]  ja3p4nl            ^autoconf-archive@2023.02.20 build_system=autotools platform=linux os=rhel9 target=zen4
[+]  xyiihtb         ^nvptx-tools@2023-09-13 build_system=autotools commit=c321f1a3573dd89a12e3291d690207685a34df6e platform=linux os=rhel9 target=zen4 %c,cxx=gcc@11.5.0

```
Install gcc:
```console
$ spack install -j8 gcc > install.log
```

### Error message

[...]

PR #5921, opened 2026-08-05
gcc: Fix build of `gcc@13: +nvptx ^cuda@13:`

mikerenfro:
Related to #3533, this seems to address the issue both with `sm_52` and a similar issue with `sm_30`. Without something like this patch, it's impossible to build a new gcc with nvptx and CUDA 13.

Edits completely welcome.


bernhardkaindl (approved):
Ack: Much needed, I could not get any `gcc +nvptx ^cuda@13` build so far without this PR.

I guess some contraints which other GCC +nvptx versions don't support cuda@13 would alo be good.

PR #5863, opened 2026-08-03
readline: switch url for patch download

mkrack:
ftpmirror.gnu.org does not work any longer

----
Added by @bernhardkaindl:

Update:
- https://ftpmirror.gnu.org/readline/readline-8.3-patches/readline83-003 is responding normally again, but it was in mostly broken state for at last 16 hours.
- We might try and see if we can update to support fallback to ftp.gnu.org for patch() urls as well as we apparently do for the main `url` of the recipe.

---
https://ftpmirror.gnu.org has had more outages recently than before:

Here are some stats you can refresh:
- https://notopening.com/site/ftp.gnu.org - Works since some downtime last week
- https://notopening.com/site/ftpmirror.gnu.org - Was down for many hours.

For me, https://ftpmirror.gnu.org/readline/readline-8.3-patches/readline83-003 currently alternates beween:
- Rarely: Working, but with a long delay of 30 seconds, 37 seconds, 48 seconds, or more.
- Most often: Timeout
- Also often: 502: Bad Gateway.

We have reports in Slack that https://ftpmirror.gnu.org didn't work on:
- July 9 and 10
- April 24, 27 and 29
- September 2, 2025
- March 24, 2025

PR #5922, opened 2026-08-05
bash: switch patch download url away from ftpmirror.gnu.org

RatneshVaibhav:
Same underlying issue as #5863(readline)  - bash's patch loop is still uses ftpmirror.gnu.org for constructing the URLs, which has a documented history of intermittent outages(see #5863). 
Applying the same fix: switching to ftp.gnu.org for patch downloads. 
Tested locally with `spack install bash`, patches fetch and apply successfully.

### DIFF
--- a/repos/spack_repo/builtin/packages/bash/package.py
+++ b/repos/spack_repo/builtin/packages/bash/package.py
@@ -194,7 +194,7 @@ class Bash(AutotoolsPackage, GNUMirrorPackage):
     ):
         ver = Version(verstr)
         patch(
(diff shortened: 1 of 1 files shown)


bernhardkaindl (approved):
Ack as needed, ftpmirror.gnu.org is down again and multiple GitLib CI pipelines failed because of ftpmirror.gnu.org:

https://notopening.com/site/ftpmirror.gnu.org

RatneshVaibhav:
Thanks for the update! Good to know it's still ongoing. That's actually why I was thinking it might be worth checking a few more GNU mirror–dependent packages to see if they're affected in the same way as bash and readline. I'm happy to look into a few more if that would be helpful.

PR #5923, opened 2026-08-05
ci: audit packages against released Spack versions

alalazo:
Make the Spack checkout a matrix axis of the audit job, and run the audits against v1.2.2 as well as the commit pinned in `.ci/env`. This gives us some confidence that we won't introduce utterly breaking changes in packages, for older Spack versions we still support (e.g. use of new directives introduced on `spack/spack` develop etc.).

### DIFF
--- a/.github/actions/checkout-spack/action.yml
+++ b/.github/actions/checkout-spack/action.yml
@@ -5,6 +5,10 @@ inputs:
     description: Path to checkout Spack repository
     required: false
     default: spack-core
+  ref:
+    description: 'Spack ref to checkout, defaults to SPACK_CHECKOUT_VERSION from .ci/env'
+    required: false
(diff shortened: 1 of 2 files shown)


haampie (approved):

kwryankrattiger:
Could you also add the latest 1.1 and 1.0?

alalazo:
@kwryankrattiger I'm waiting to do that because locally I had issues - so I want first to secure v1.2.2, then look at the other two.

kwryankrattiger:
> I'm waiting to do that because locally I had issues - so I want first to secure v1.2.2, then look at the other two.

That's fine. It would be nice to also have a PR with the issues visible, but that can be a follow up.

kwryankrattiger (approved):

PR #5926, opened 2026-08-05
py_nvtx: add version 0.2.15 and manage nvtx location for 0.2.10

LydDeb:
This PR is relative to #5900 
The NVTX header directory has been included in the source code since version 0.2.11, and it is no longer necessary to define NVTX_PREFIX.

### DIFF
--- a/repos/spack_repo/builtin/packages/py_nvtx/package.py
+++ b/repos/spack_repo/builtin/packages/py_nvtx/package.py
@@ -14,6 +14,7 @@ class PyNvtx(PythonPackage):
     pypi = "nvtx/nvtx-0.2.10.tar.gz"
 
     license("Apache-2.0")
+    maintainers("LydDeb")
 
     version("0.2.15", sha256="2287d3be05b85661deb386f878d1f536c2e532774aa9ec7a50c434942ed81ae5")
     version("0.2.10", sha256="58b89cd69079fda1ceef8441eec5c5c189d6a1ff94c090a3afe03aedd0bbd140")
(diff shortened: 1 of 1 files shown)


bernhardkaindl (approved):

bernhardkaindl (commented):
@LydDeb There is a missed change that should have been aded:

Source metadata uses the newer PEP 639 string form for the license. This is a version mismatch in the Python build backend: It looks like older setuptools reject that declaration:

```py
    File "/tmp/spack-install/linux-x86_64_v4/py-setuptools-69.2.0-jww7czdeaw7h6xn3dgg5ataxeasuraxx/lib/python3.14/site-packages/setuptools/config/pyprojecttoml.py", line 57, in validate
      raise ValueError(f"{error}\n{summary}") from None
  ValueError: invalid pyproject.toml config: `project.license`.
  configuration error: `project.license` must be valid exactly by one definition (2 matches found):

      - keys:
          'file': {type: string}
        required: ['file']
      - keys:
          'text': {type: string}
        required: ['text']
```
@LydDeb, this fixes the isse:
```py
depends_on("py-setuptools@77:", type="build", when="@0.2.15:")
```

Also, pyproject.toml says Cython >= 3.1 is required:
https://github.com/NVIDIA/NVTX/blob/dev/python/pyproject.toml#L25
```py
[build-system]
requires = ["setuptools", "wheel", "Cython>=3.1"]
```
That means the cython dep should be updated or one for the newer versions that need newer cython should be added:
```
depends_on("py-cython@3.1:", type="build")
```
@LydDeb, could you please submite a new PR with these fixes?

PR #5927, opened 2026-08-05
freetype: deprecate versions with CVEs

adamjstewart:


adamjstewart:
@spackbot fix style

tpadioleau:
oh boy! :)

michaelkuhn (approved):
LGTM

A quick grep also doesn't show any packages depending on older FreeType versions, so this shouldn't break anything.

PR #5930, opened 2026-08-05
py-pybtex-docutils: add v1.0.1, v1.0.2, v1.0.3

sethrj:
This adds versions 1.0.1–1.0.3 which include a compatibility update for the latest sphinx. Without this update, sphinxcontrib-bibtex fails with:
```
      File "…/pybtex/richtext.py", line 894, in render
        return backend.format_href(self.url, text, self.external)
               ~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    TypeError: Backend.format_href() takes 3 positional arguments but 4 were given
```


RikkiButler20 (approved):
LGTM

PR #5933, opened 2026-08-05
mfem: constrain compatible cuDSS versions

Sbozzolo:
See https://github.com/mfem/mfem/issues/5441.

I also added the lower bounds by the compatible syntax.


bernhardkaindl (approved):
Hi @tzanio, @v-dobrev, if you want to review, do you need more time to check it?

PR #5934, opened 2026-08-05
gcc: add v14.4.0

jcortial-safran:
Gcc 14.4 has been released in June, but I could not find any related pull request or issue.


michaelkuhn:
Thanks! Looks like I missed this. Could you please also remove 14.3.0?

jcortial-safran:
> Thanks! Looks like I missed this. Could you please also remove 14.3.0?

Directly remove 14.3 ? Not even deprecate first ?

michaelkuhn:
Sorry, I meant deprecating. You can simply move it down to the `with default_args(deprecated=True)` block.

michaelkuhn (approved):
Works for me (Fedora 44, GCC 16)

PR #5937, opened 2026-08-05
py-melissa-online

abhishek1297:
From melissa version 3+, all builds are done via scikit-build-core and therefore, I created a new package `py-melissa-online` that is a combination of `melissa` and `py-melissa-core` packages.

How to discontinue older packages?


bernhardkaindl:
> How to discontinue older packages?

I guess you could deprecate all versions of the old packages.
Then, after the next spack release is out, you remove the recipes with only deprecated versions.

BTW: If the currently deprecated versions are already deprecated in the lastest spack release using deprecated=True, you could already remove those sooner or now.

AFAICS, in the case of melissa, that should be all. If you then want the new package to provide melissa, you can add a provides("melissa") to it, but you'd only need that if there are packages which depend on it.

bernhardkaindl (approved):

PR #5945, opened 2026-08-06
hpcserver: new package

laksono:
hpcserver is a server-side component of HPCToolkit to allow hpcviewer to remotely access databases.
- All builds are done using Java and Apache Maven.
- The installation is done by untarring the package instead of using Apache Maven

See https://gitlab.com/hpctoolkit/database for details.

Assisted-by: Claude Sonnet 4.6

### DIFF
--- a/repos/spack_repo/builtin/packages/hpcserver/package.py
+++ b/repos/spack_repo/builtin/packages/hpcserver/package.py
@@ -1,0 +1,85 @@
+# Copyright Spack Project Developers. See COPYRIGHT file for details.
+#
+# SPDX-License-Identifier: (Apache-2.0 OR MIT)
+
+import glob
+import os
+
+from spack_repo.builtin.build_systems.maven import MavenPackage
+
+from spack.package import *
+
+
+class Hpcserver(MavenPackage):
+    """
+    remote performance analysis. It runs on remote hosts holding HPCToolkit
+    measurement databases, performing server-side parsing, indexing, and
+    serving of performance data to GUI tools like hpcviewer, avoiding the need
+    to download large profile directories locally
+    """
+
+    homepage = "https://gitlab.com/hpctoolkit/database"
+    git = "https://gitlab.com/hpctoolkit/database.git"
+    url = "https://gitlab.com/hpctoolkit/database/-/archive/1.4.0/database-1.4.0.tar.gz"
+
+    maintainers("laksono")
+
+    license("Apache-2.0")
+
+    version("develop", branch="main")
+    version("1.4.0", sha256="661806f9d2e2a4ebcc99f54853cb8d5188fc6a3ce0287af2c5ca93631461a373")
+    version("1.3.1", sha256="beafbadd409720456833dcf9044f1ca34df41ac9c0be62c3432fa476344581ab")
+    version("1.3.0", sha256="5658c4e043482ac01b8015a7aeac20e5fe54b170567a12f05a5ba71a532617f0")
+
+    # -----------------------------------------------------------------------
+    # Dependencies
+    # -----------------------------------------------------------------------
+    # Java is needed both to build (Maven runs on the JVM) and at runtime
+    # (the server is a Java application).
+    depends_on("java@17:25", type=("build", "run"))
+
+    # -----------------------------------------------------------------------
+    def install(self, spec, prefix):
+        """Unpack the assembled tarball and bind the Spack JRE into the launcher."""
+
+        # Locate the produced binary tarball.  Maven puts it under
+        # server/target/ with the naming convention:
+        #   org.hpctoolkit.db.server-<VERSION>-bin.tar.gz
+        pattern = join_path(
+            "server",
+            "target",
+            "org.hpctoolkit.db.server-*-bin.tar.gz",
+        )
+        tarballs = glob.glob(pattern)
+        if not tarballs:
+            raise InstallError(
+                "Could not find the hpcserver binary tarball matching:\n"
+                f"  {pattern}\n"
+                "Check that 'mvn clean package' completed successfully."
+            )
+        # Take the most-recently modified match (there should be exactly one).
+        tarball = max(tarballs, key=os.path.getmtime)
+
+        # The tarball unpacks into a directory named
+        # org.hpctoolkit.db.server-<VERSION>/  (no '-bin' suffix after extraction).
+        # We install that directory tree directly into prefix.
+        mkdirp(prefix)
+        tar = which("tar")
+        tar("xf", tarball, "-C", prefix, "--strip-components=1")
+
+        # Bind the Spack-provided JRE so the launcher script uses the correct
+        # Java regardless of what is in the user's PATH at runtime.
+        bind_java = join_path(prefix, "share", "hpcserver", "bind-java.sh")
+        if not os.path.isfile(bind_java):
+            raise InstallError(
+                f"bind-java.sh not found at expected location:\n  {bind_java}\n"
+                "Check the tarball layout – the install prefix structure may have changed."
+            )
+
+        # bind-java.sh must be run from inside the installation directory.
+        with working_dir(prefix):
+            bash = which("bash")
+            bash(bind_java)
+
+    # -----------------------------------------------------------------------
+    sanity_check_is_file = ["bin/hpcserver"]


repos/spack_repo/builtin/packages/hpcserver/package.py
```diff
@@ -0,0 +1,133 @@
...
+
+    # -----------------------------------------------------------------------
+    # Build phases
+    # -----------------------------------------------------------------------
+    phases = ["build", "install"]
+
+    # -----------------------------------------------------------------------
+    def build(self, spec, prefix):
+        """Run 'mvn clean package' to compile and assemble the server tarball."""
+        mvn = which("mvn")
```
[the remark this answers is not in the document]
laksono: Solved  in the commit 1dbf3cea17a352f1f22ec0c83c19ce11b6e754ac

repos/spack_repo/builtin/packages/hpcserver/package.py
```diff
@@ -0,0 +1,132 @@
...
+    maintainers("laksono")
+
+    version("develop", branch="main")
+    version("1.4.0", tag="1.4.0", commit="1493ca471d2ad6d43648e3e5d076fc0cba7c6093")
+    version("1.3.1", tag="1.3.1", commit="fc8282855bc83a6cb13fb586392b6df811e453d1")
+    version("1.3.0", tag="1.3.0", commit="9bc9be2274dee7261602aab506051497ea11c8be")
+    version("1.2.0", tag="1.2.0", commit="bc562f658ec4a04fc469b40b020b40aa96bf0f59")
+    version("1.1.1", tag="1.1.1", commit="d41fbfb0782ae706c18a7f67d75e5d5f42f4c2f5")
+    version("1.1.0", tag="1.1.0", commit="cfe7c727e258ceaa05da24eb712c3aa802416010")
+    version("1.0.0", tag="1.0.0", commit="b7173db2bb68a2be19dd4ea1b5087d94c6bfe360")
```
cmelone: for releases, we prefer the source to be a tarball from something like gitlab/github releases. you'd get them on this page

https://gitlab.com/hpctoolkit/database/-/releases

see these docs: https://spack.readthedocs.io/en/latest/packaging_guide_creation.html#source-code-and-versions

repos/spack_repo/builtin/packages/hpcserver/package.py
```diff
@@ -0,0 +1,132 @@
...
+#
+# SPDX-License-Identifier: (Apache-2.0 OR MIT)
+
+import glob
+import os
+
+from spack.package import *
+
+
+class Hpcserver(Package):
```
cmelone: have you considered using the `MavenPackage` class? might simplify some of your directives

https://spack.readthedocs.io/en/latest/build_systems/mavenpackage.html

laksono: But we are using `MavenPackage`:

```python
    # Java is needed both to build (Maven runs on the JVM) and at runtime
    # (the server is a Java application).
    depends_on("java@17:25", type=("build", "run"))

    # Maven is the build tool; only needed during the build phase.
    depends_on("maven@3.6:", type="build")
```

cmelone: inheriting from `MavenPackage` is best practice. you can still override the `java`/`maven` version ranges, but it will eliminate the need to manually define the phases and other behavior that your package may be reimplementing.

https://github.com/spack/spack-packages/blob/78e56ec137005f8b422209fcef31787de1f254b8/repos/spack_repo/builtin/build_systems/maven.py

see these examples: https://github.com/search?q=repo%3Aspack%2Fspack-packages+MavenPackage&type=code

repos/spack_repo/builtin/packages/hpcserver/package.py
```diff
@@ -0,0 +1,132 @@
...
+
+        # bind-java.sh must be run from inside the installation directory.
+        with working_dir(prefix):
+            bash = which("bash")
+            bash(bind_java)
+
+    # -----------------------------------------------------------------------
+    def setup_run_environment(self, env):
+        """Expose the hpcserver launcher on PATH."""
+        env.prepend_path("PATH", self.prefix.bin)
```
cmelone: spack will do this by default when the package is loaded

repos/spack_repo/builtin/packages/hpcserver/package.py
laksono: Thanks for the pointer. Fixed in the latest commit

repos/spack_repo/builtin/packages/hpcserver/package.py
```diff
@@ -0,0 +1,97 @@
...
+    on remote hosts where HPCToolkit performance measurement databases are stored,
+    performing server-side parsing, indexing, and serving of performance data to
+    GUI analysis tools like hpcviewer over network connections. By offloading
+    database handling to the host system, hpcserver eliminates the need to
+    download large performance profile directories to local machines.
+    """
+
+    homepage = "https://gitlab.com/hpctoolkit/database"
+    git = "https://gitlab.com/hpctoolkit/database.git"
+    url = "https://gitlab.com/hpctoolkit/database/-/archive/1.4.0/database-1.4.0.tar.gz"
```
cmelone, line 23: please add a license directive https://spack.readthedocs.io/en/latest/packaging_guide_creation.html#license-information

repos/spack_repo/builtin/packages/hpcserver/package.py
```diff
@@ -0,0 +1,97 @@
...
+        if not os.path.isfile(bind_java):
+            raise InstallError(
+                f"bind-java.sh not found at expected location:\n  {bind_java}\n"
+                "Check the tarball layout – the install prefix structure may have changed."
+            )
+
+        # bind-java.sh must be run from inside the installation directory.
+        with working_dir(prefix):
+            bash = which("bash")
+            bash(bind_java)
```
cmelone, line 82: if I understand correctly, this sets `JAVA_HOME` to the Java install it was built with? 

this is not necessary as the Java providers will set this in `setup_run_environment` automatically after this package is loaded

this warning also indicates it should not be added to spack

> This script has a very specific use case and isn't suited for automated
installation or downstream packaging.

repos/spack_repo/builtin/packages/hpcserver/package.py
```diff
@@ -0,0 +1,97 @@
...
+        """Quick sanity check: verify that the expected launcher exists."""
+        launcher = join_path(self.prefix, "bin", "hpcserver")
+        if not os.path.exists(launcher):
+            # Some releases may use a .sh extension or place it elsewhere.
+            # Emit a warning rather than a hard error so packaging experiments
+            # are not blocked.
+            tty.warn(
+                f"Expected launcher not found at {launcher}.\n"
+                "You may need to add the correct subdirectory to PATH manually."
+            )
```
cmelone: this entire block can be replaced with 

```
sanity_check_is_file = ["bin/hpcserver"]
```

repos/spack_repo/builtin/packages/hpcserver/package.py
```diff
@@ -0,0 +1,97 @@
...
+        )
+        tarballs = glob.glob(pattern)
+        if not tarballs:
+            raise InstallError(
+                "Could not find the hpcserver binary tarball matching:\n"
+                f"  {pattern}\n"
+                "Check that 'mvn clean package' completed successfully."
+            )
+        # Take the most-recently modified match (there should be exactly one).
+        tarball = max(tarballs, key=os.path.getmtime)
```
cmelone, line 61: if there's only one expected output, why not just access the first element of the iterable?

repos/spack_repo/builtin/packages/hpcserver/package.py
```diff
@@ -0,0 +1,97 @@
...
+        """Unpack the assembled tarball and bind the Spack JRE into the launcher."""
+
+        # Locate the produced binary tarball.  Maven puts it under
+        # server/target/ with the naming convention:
+        #   org.hpctoolkit.db.server-<VERSION>-bin.tar.gz
+        pattern = os.path.join(
+            self.stage.source_path,
+            "server",
+            "target",
+            "org.hpctoolkit.db.server-*-bin.tar.gz",
```
cmelone, line 51: is there a deterministic identifier for the tarball? from my testing it looks like `{version}-SNAPSHOT` but let me know if that's wrong. it would be nicer to have that than globbing

repos/spack_repo/builtin/packages/hpcserver/package.py
```diff
@@ -0,0 +1,97 @@
...
+    version("develop", branch="main")
+    version("1.4.0", sha256="661806f9d2e2a4ebcc99f54853cb8d5188fc6a3ce0287af2c5ca93631461a373")
+    version("1.3.1", sha256="beafbadd409720456833dcf9044f1ca34df41ac9c0be62c3432fa476344581ab")
+    version("1.3.0", sha256="5658c4e043482ac01b8015a7aeac20e5fe54b170567a12f05a5ba71a532617f0")
+
+    # -----------------------------------------------------------------------
+    # Dependencies
+    # -----------------------------------------------------------------------
+    # Java is needed both to build (Maven runs on the JVM) and at runtime
+    # (the server is a Java application).
```
cmelone, line 38:
```suggestion
```

extraneous details

cmelone (changes requested):

repos/spack_repo/builtin/packages/hpcserver/package.py
laksono: The goal of `bind-java.sh` is to ensure `hpcserver` can be launched remotely using the current JRE (or JRE set by Spack).

So, this call to `bind-java.sh` does NOT set `JAVA_HOME`; it specifies which JRE hpcserver needs to run. I believe this script is fine for automated installation. If `JAVA_HOME` is not set, hpcserver will use the default JRE, if any.

repos/spack_repo/builtin/packages/hpcserver/package.py
laksono: You're right, that works with the release version (1.4.0, 1.3.1,...), but unfortunately, it does not work with the `develop` version as there is no `develop-SNAPSHOT`.

repos/spack_repo/builtin/packages/hpcserver/package.py
```diff
@@ -0,0 +1,88 @@
...
+
+class Hpcserver(MavenPackage):
+    """hpcserver is the backend server component of HPCToolkit's client-server
+    architecture, designed to facilitate remote performance analysis. It executes
+    on remote hosts where HPCToolkit performance measurement databases are stored,
+    performing server-side parsing, indexing, and serving of performance data to
+    GUI analysis tools like hpcviewer over network connections. By offloading
+    database handling to the host system, hpcserver eliminates the need to
+    download large performance profile directories to local machines.
+    """
```
alecbcs:
```suggestion
    """Backend server component of HPCToolkit's client-server architecture for
       remote performance analysis. It runs on remote hosts holding HPCToolkit
       measurement databases, performing server-side parsing, indexing, and
       serving of performance data to GUI tools like hpcviewer, avoiding the need
       to download large profile directories locally."""
    """
```

The current package description is longer than we normally include. What would y'all think about this instead?

repos/spack_repo/builtin/packages/hpcserver/package.py
```diff
@@ -0,0 +1,88 @@
...
+    depends_on("java@17:25", type=("build", "run"))
+
+    # -----------------------------------------------------------------------
+    def install(self, spec, prefix):
+        """Unpack the assembled tarball and bind the Spack JRE into the launcher."""
+
+        # Locate the produced binary tarball.  Maven puts it under
+        # server/target/ with the naming convention:
+        #   org.hpctoolkit.db.server-<VERSION>-bin.tar.gz
+        pattern = os.path.join(
```
alecbcs:
```suggestion
        pattern = join_path(
```

`spack.package` already provides a `join_path()` method that is commonly used across the repository. Suggesting here to stick with the common format.

repos/spack_repo/builtin/packages/hpcserver/package.py
```diff
@@ -0,0 +1,88 @@
...
+            self.stage.source_path,
```
alecbcs:
```suggestion
```

This line is unnecessary as `install()` already treats the stage's source path as its working directory.

repos/spack_repo/builtin/packages/hpcserver/package.py
alecbcs: Is there a configuration flag we can pass to Maven to set the produced tarball's path deterministically?

repos/spack_repo/builtin/packages/hpcserver/package.py
```diff
@@ -0,0 +1,88 @@
...
+        )
+        tarballs = glob.glob(pattern)
+        if not tarballs:
+            raise InstallError(
+                "Could not find the hpcserver binary tarball matching:\n"
+                f"  {pattern}\n"
+                "Check that 'mvn clean package' completed successfully."
+            )
+        # Take the most-recently modified match (there should be exactly one).
+        tarball = max(tarballs, key=os.path.getmtime)
```
alecbcs, line 61: @laksono can you explain a little here about why hpcserver's Maven build is so different than all of the other maven packages in the repository that don't produce/re-unpack tarballs in their builds?

repos/spack_repo/builtin/packages/hpcserver/package.py
laksono: I am fine with it. Thanks for the suggestion.

repos/spack_repo/builtin/packages/hpcserver/package.py
laksono: Two reasons:
1. `mvn install` will install all libraries in the repository: `local`, `client`, `protocol`, and `server`; while this custom function will only install the `server` library (and dependencies). It is harmless to install all libraries, but I just don't think it's the proper way.
2. We can customize the Maven's install phase to install just the `server` part, but it will be in a future release.

PR #5949, opened 2026-08-06
Add Spack v1.1.1 to the audit job

alalazo:
Due to a change in spack/spack#52113 some redundant conflicts in `boost` are flagged in v1.1.1 and are not in v1.2.2. The difference is that in v1.1.1 we were merging the conflict with its when clause before checking.

Since these conflicts just make the audits fail, but they don't compromise anything else - boost can still be installed, the error messages don't get worse, etc. here we remove those conflicts to make the new checks pass.

### DIFF
--- a/.github/workflows/audit.yaml
+++ b/.github/workflows/audit.yaml
@@ -25,7 +25,7 @@ jobs:
         - { os: macos-latest, shell: bash }
         # An empty ref means the commit pinned in .ci/env, the others are the
         # released versions we keep the packages repository compatible with
-        spack_ref: ['', 'v1.2.2']
+        spack_ref: ['', 'v1.2.2', 'v1.1.1']
     defaults:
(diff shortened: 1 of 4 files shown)


alalazo:
The `genesis` failure seems real. It's being reported on v1.1.1 and not on v1.2.2 because of spack/spack#51908

alalazo:
The `samrai` failure was real too :disappointed:

alalazo:
@kwryankrattiger This takes care of Spack v1.1.1 I had to fix a few packages, in a way that is not disruptive for users - remove redundant directives, fix a few package bugs etc.

alalazo:
`samrai` was not reported on the current `develop` because of a bug: https://github.com/spack/spack/pull/52845

kwryankrattiger (approved):

PR #5950, opened 2026-08-06
opari2: add 2.0.10, deprecate previous versions

wrwilliams:
OPARI2 2.0.10 fixes crash bugs when processing particular C++ codes. In general we want Score-P (and any other downstream users, if they exist) to stay on the latest OPARI2 version, but the old versions will _build_ with the older, buggy OPARI2s.

OPARI2 1.0 is ancient and unused and the deprecation is 100% pre-removal. OPARI2 < 2.0.6 is not depended on by any Spack-supported Score-P and can be removed next cycle after deprecation AFAIK.

### DIFF
--- a/repos/spack_repo/builtin/packages/opari2/package.py
+++ b/repos/spack_repo/builtin/packages/opari2/package.py
@@ -21,17 +21,21 @@ class Opari2(AutotoolsPackage):
     homepage = "https://www.vi-hps.org/projects/score-p"
     url = "https://perftools.pages.jsc.fz-juelich.de/cicd/opari2/tags/opari2-2.0.8/opari2-2.0.8.tar.gz"
 
-    version("2.0.9", sha256="d57139f757c5666afaaead45ed3d0954a9b98c4a6cef6b22afe672707cffd779")
-    version("2.0.8", sha256="196e59a2a625e6c795a6124c61e784bad142f9f38df0b4fa4d435ba9b9c19721")
-    version("2.0.7", sha256="e302a4cc265eb2a4aa27c16a90eabd9e1e58cb02a191dd1c4d86f9a0df128715")
-    version("2.0.6", sha256="55972289ce66080bb48622110c3189a36e88a12917635f049b37685b9d3bbcb0")
-    version("2.0.5", sha256="9034dd7596ac2176401090fd5ced45d0ab9a9404444ff767f093ccce68114ef5")
-    version("2.0.4", sha256="f69e324792f66780b473daf2b3c81f58ee8188adc72b6fe0dacf43d4c1a0a131")
-    version("2.0.3", sha256="7e2efcfbc99152ee6e31454ef4fb747f77165691539d5d2c1df2abc4612de86c")
-    version("2.0.1", sha256="f49d74d7533f428a4701cd561eba8a69f60615332e81b66f01ef1c9b7ee54666")
-    version("2.0", sha256="0c4e575be05627cd001d692204f10caef37b2f3d1ec825f98cbe1bfa4232b0b7")
-    version("1.1.4", sha256="b80c04fe876faaa4ee9a0654486ecbeba516b27fc14a90d20c6384e81060cffe")
-    version("1.1.2", sha256="8405c2903730d94c828724b3a5f8889653553fb8567045a6c54ac0816237835d")
+    # OPARI2 2.x gets bugfix releases that preserve backwards compatibility for all Score-P versions
+    # still in Spack. 1.x is ancient and is deprecated in anticipation of removal.
+    # Deprecating here as a _slightly_ gentler approach than hard-bumping the version dependencies downstream.
+    version("2.0.10", sha256="49d9526bf76ebf7836e62f70850d4d605b08910dbabddac3f7479b509abce887")
+    version("2.0.9", sha256="d57139f757c5666afaaead45ed3d0954a9b98c4a6cef6b22afe672707cffd779", deprecated=True)
+    version("2.0.8", sha256="196e59a2a625e6c795a6124c61e784bad142f9f38df0b4fa4d435ba9b9c19721", deprecated=True)
+    version("2.0.7", sha256="e302a4cc265eb2a4aa27c16a90eabd9e1e58cb02a191dd1c4d86f9a0df128715", deprecated=True)
+    version("2.0.6", sha256="55972289ce66080bb48622110c3189a36e88a12917635f049b37685b9d3bbcb0", deprecated=True)
+    version("2.0.5", sha256="9034dd7596ac2176401090fd5ced45d0ab9a9404444ff767f093ccce68114ef5", deprecated=True)
+    version("2.0.4", sha256="f69e324792f66780b473daf2b3c81f58ee8188adc72b6fe0dacf43d4c1a0a131", deprecated=True)
+    version("2.0.3", sha256="7e2efcfbc99152ee6e31454ef4fb747f77165691539d5d2c1df2abc4612de86c", deprecated=True)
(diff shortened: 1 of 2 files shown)


wrwilliams:
If it's better form to hard-update the `scorep` dependencies to a consistent `opari2@2.0.10:` I can do that instead--just let me know.

repos/spack_repo/builtin/packages/opari2/package.py
```diff
@@ -21,17 +21,65 @@ class Opari2(AutotoolsPackage):
...
+    version(
+        "1.1.4",
+        sha256="b80c04fe876faaa4ee9a0654486ecbeba516b27fc14a90d20c6384e81060cffe",
+        deprecated=True,
+    )
+    version(
+        "1.1.2",
+        sha256="8405c2903730d94c828724b3a5f8889653553fb8567045a6c54ac0816237835d",
+        deprecated=True,
+    )
```
bernhardkaindl: Hi @wrwilliams, lets check which recipes use opari2:
```py
git grep opari2 | grep -v /opari2/
repos/spack_repo/builtin/packages/eztrace/package.py:    depends_on("opari2")
repos/spack_repo/builtin/packages/scorep/package.py:    depends_on("opari2@2.0.9", when="@9:")
---------------------------------------------------------------------------^^^^^
repos/spack_repo/builtin/packages/scorep/package.py:    depends_on("opari2@2.0.6:", when="@7:")
repos/spack_repo/builtin/packages/scorep/package.py:            "--with-opari2=%s" % spec["opari2"].prefix.bin,
```
This `shows scorep@9:` uses `"opari2@2.0.9"` hard-coded without an open-ended version range! This meams deprecating `"opari2@2.0.9"` would make scorep require `--deprecated`, and I think this should be avoided, it could break all environments using scorep, AFAICS.

Instead, as a 1st measure, I'ld suggest to keep 2.0.9 as non-deprecated for now.
This would unblock this PR as far as I could so so far.

As a 2nd measure, to improve the scorep depedendency, could you check if all `"scopreo@9:` releases (so all releases since version 9) would work at least also with `"2.0.10"` as well and update the versioned dependency of those versions in the scorep recipe?

Often, it is just an oversight of the submitter and the review to just add the dependency without a ranmge, e.g. the open ended dependency would be:
```off
-   depends_on("opari2@2.0.9", when="@9:")
+   depends_on("opari2@2.0.9:", when="@9:")
```
But ususally it would be safer, to clamp the upper range a bit, depending on incomptable changes could be made e.g. with 2.1.x already:
```off
-   depends_on("opari2@2.0.9", when="@9:")
+   depends_on("opari2@2.0.9:2.0", when="@9:")
```

Or, if breaking changes would only happen with 3.x:
```off
-   depends_on("opari2@2.0.9", when="@9:")
+   depends_on("opari2@2.0.9:2", when="@9:")
```
One of those would be great, as far as I can see. You may add this to this PR as well and a scorep maintainer should be consulted / pinged here, preferably, I think.

1. Ah, I see you are the maintainer of `scorep` 👍 , so you should know!
2. Secondly, would you consider to add yourself as maintainer of `opari2` too?
3. `git grep` only found the dependees shown above, so you should only check if you can update the pinned dependency version of scorep, the other looked like they're open-ended. You can remove the comment about hard-updating other version ranges, only the `scorep` version dependency needs to be taken care of, and of course run a test of scorep with the new version to check that it builds (and hopefully works) if you update the version dep to allow it use 2.0.10 to be sure.

So this would be the version to keep 2.0.9 un-deprecated and use a context manager to deprecate. The line-length configruation of `ruff format` for this repo has been updated to not require breaking the lines with this method:
```
ruff format repos/spack_repo/builtin/packages/opari2/package.py
1 file left unchanged
```
Suggestion:
```suggestion
    version("2.0.9", sha256="d57139f757c5666afaaead45ed3d0954a9b98c4a6cef6b22afe672707cffd779")
    with default_args(deprecated=True):
        version("2.0.8", sha256="196e59a2a625e6c795a6124c61e784bad142f9f38df0b4fa4d435ba9b9c19721")
        version("2.0.7", sha256="e302a4cc265eb2a4aa27c16a90eabd9e1e58cb02a191dd1c4d86f9a0df128715")
        version("2.0.6", sha256="55972289ce66080bb48622110c3189a36e88a12917635f049b37685b9d3bbcb0")
        version("2.0.5", sha256="9034dd7596ac2176401090fd5ced45d0ab9a9404444ff767f093ccce68114ef5")
        version("2.0.4", sha256="f69e324792f66780b473daf2b3c81f58ee8188adc72b6fe0dacf43d4c1a0a131")
        version("2.0.3", sha256="7e2efcfbc99152ee6e31454ef4fb747f77165691539d5d2c1df2abc4612de86c")
        version("2.0.1", sha256="f49d74d7533f428a4701cd561eba8a69f60615332e81b66f01ef1c9b7ee54666")
        version("2.0", sha256="0c4e575be05627cd001d692204f10caef37b2f3d1ec825f98cbe1bfa4232b0b7")
        version("1.1.4", sha256="b80c04fe876faaa4ee9a0654486ecbeba516b27fc14a90d20c6384e81060cffe")
        version("1.1.2", sha256="8405c2903730d94c828724b3a5f8889653553fb8567045a6c54ac0816237835d")
```

wrwilliams: Well, of course I know a `scorep` maintainer; he's me. :)

Context manager is a good call, 2.0.9 deprecation is fully intended, I'll fix `scorep` to have open-ended deps and add that here as well if that's OK.

wrwilliams: I should also add, WRT the future of OPARI2: we'll keep it alive until `libgomp` supports OMPT at a comparable level to LLVM's `libomp`, but nothing beyond bugfix versions are planned until then--at which point we intend to retire the project.

repos/spack_repo/builtin/packages/opari2/package.py
```diff
@@ -21,17 +21,22 @@ class Opari2(AutotoolsPackage):
     homepage = "https://www.vi-hps.org/projects/score-p"
     url = "https://perftools.pages.jsc.fz-juelich.de/cicd/opari2/tags/opari2-2.0.8/opari2-2.0.8.tar.gz"
 
+    # OPARI2 2.x gets bugfix releases that preserve backwards compatibility for all Score-P versions
+    # still in Spack. 1.x is ancient and is deprecated in anticipation of removal.
+    # Deprecating here as a _slightly_ gentler approach than hard-bumping the version dependencies downstream.
```
bernhardkaindl: Hi @wrwilliams, CI fails on these lines (https://github.com/spack/spack-packages/actions/runs/31095316153/job/92595903100?pr=5950) and the autofix can't fix these comments for you:
```py
==> running ruff check
E501 Line too long (100 > 99)
  --> repos/spack_repo/builtin/packages/opari2/package.py:24:100
   |
22 |     url = "https://perftools.pages.jsc.fz-juelich.de/cicd/opari2/tags/opari2-2.0.8/opari2-2.0.8.tar.gz"
23 |
24 |     # OPARI2 2.x gets bugfix releases that preserve backwards compatibility for all Score-P versions
   |                                                                                                    ^
25 |     # still in Spack. 1.x is ancient and is deprecated in anticipation of removal.
26 |     # Deprecating here as a _slightly_ gentler approach than hard-bumping the version dependencies downstream.
```
AFAICS, this should fix the CI error:
```suggestion
    # The OPARI2 2.x bugfix releases preserve backwards compatibility for all Score-P versions
    # still in Spack. 1.x is ancient and is deprecated in anticipation of removal.
```
The 3rd line should not be needed. Deprecating and then removing the deprecated versions after the next spack release is actually wanted to keep recipes maintainable.

bernhardkaindl (approved):

PR #5952, opened 2026-08-06
r-fs: fix libuv as link dependency

manuelakuhn:
`libuv` also has to be a link dependency, otherwise I get the error:
```
> Error: package or namespace load failed for 'fs' in dyn.load(file, DLLpath = DLLpath, ...):
   unable to load shared object '/spack/opt/spack/linux-skylake/r-fs-2.1.0-wlcopzvsux56cocy6n36crxcrnivhadq/rlib/R/library/00LOCK-spack-src/00new/fs/libs/fs.so':
>   libuv.so.1: cannot open shared object file: No such file or directory
> Error: loading failed
```


w8jcik (approved):
I also seen these errors recently, but didn't realize what is the issue.

Good catch.

bernhardkaindl (approved):

PR #5956, opened 2026-08-06
augustus: include cstdint to comply with c++11 standard

mikerenfro:
There are several instances of augustus using `uint64_t` and related types, but not including `cstdint`. Older compilers can tolerate this, but newer versions of GCC don't.

### DIFF
--- a/repos/spack_repo/builtin/packages/augustus/cstdint.patch
+++ b/repos/spack_repo/builtin/packages/augustus/cstdint.patch
@@ -1,0 +1,49 @@
+--- a/auxprogs/homGeneMapping/include/sqliteDB.hh    2026-07-14 18:30:54.427118315 +0000
++++ b/auxprogs/homGeneMapping/include/sqliteDB.hh 2026-07-14 18:32:43.140850329 +0000
+@@ -10,6 +10,7 @@
+ #ifndef __SQLITEDB_H__
+ #define __SQLITEDB_H__
+
++#include <cstdint>
+ #include <string>
+ #include <vector>
+ #include <sqlite3.h>
+--- a/auxprogs/homGeneMapping/src/sqliteDB.cc      2026-07-14 18:31:10.264225061 +0000
++++ b/auxprogs/homGeneMapping/src/sqliteDB.cc   2026-07-14 18:33:19.025091519 +0000
+@@ -11,6 +11,7 @@
+  **********************************************************************/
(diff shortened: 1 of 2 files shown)


bernhardkaindl:
Hi @mikerenfro,

Checked if this works for all versions with gcc-14 (dpkg of Ubuntu 24.04)
```py
for i in `spack versions --safe augustus|tac`;do spack install augustus@=$i %gcc@14;echo $i ;done
...
spack find augustus
-- linux-ubuntu24.04-skylake / %c,cxx=gcc@14.3.0 ----------------
augustus@3.4.0  augustus@3.5.0
==> 2 installed packages
```
The older versions fail with other errors like
```c
> genome.cc:817:47: error: 'class SeqIntKey' has no member named 'getKey'
> genome.cc:523:18: error: request for member 'insert' in '((Genome*)this)->Genome::gfHash', which is of non-class type 'int'
    523 |     ret = gfHash.insert(pair< int, map< uint_fast64_t, list< GeneFeature* > > >(seqID,m));
```
So the patch is sufficient for the two newer versions, not the older ones.

3.5.0 failed with %gcc@15 and @16:
```py
> filterBam.cc:338:40: error: 'using std::__shared_ptr_access<BamAlignmentRecord, __gnu_cxx::_S_atomic, false, false>::element_type = class BamAlignmentRecord' {aka 'class BamAlignmentRecord'} has no member named 'countCigarOperations'
```
Only 3.4.0 was successful with gcc-15 and 16

I think for these, maybe a conflict is the last resort?
```py
    conflicts("%gcc@15:", when="@:3.5.0")
    conflicts("%gcc@14:", when="@:3.3")
```
No pressure: Could deprecating some older versions be a non-problem for users?

mikerenfro:
I must have started this patch when I was using gcc@13 by default, and never checked it after I migrated to 15 or 16. I'm also about 99% certain we'll soon move to Biocontainers and [container-mod](https://github.com/TuftsRT/container-mod) for the vast majority of our bioinformatics applications.

I have no idea which versions could be deprecated: our user population often asks for very old versions of things to replicate results from other articles.

But I'll at least update the branch to handle what you found.

bernhardkaindl:
@mikerenfro, I opened #6083 to use `gcc -include stdint.h`, it appears to work as well as the patch, but needs another maintainer to approve.

PR #5958, opened 2026-08-06
lcov: add v2.4 and v2.5

jcortial-safran:
Since release 2.5, lcov requires sphinx to build the documentation. There seems to be no easy way to disable the dependency short of patching the Makefiles.


jcortial-safran:
@KineticTheory For some reason you were not automatically added as a reviewer and I cannot do it myself...

bernhardkaindl (approved):
@jcortial-safran, as these are just two new versions of lcov, the basic coverage tool, I think I can approve and merge them.

PR #5962, opened 2026-08-07
glpk: fix build for %gcc@:14, %clang

bernhardkaindl:
Fix the build of glpk with compiler != %gcc@15:
The patch for gcc@15: must not be applied for other compilers, AFAICS.


wdconinc (approved):
IMHO this applies a patch where maybe we should instead apply a compiler flag to set C standard back to gnu17 or so. The bool type and constants are just one thing that c23 introduced.

But I assume it's indeed correct that if we are going to apply a patch it should be to gcc@15: only.

bernhardkaindl:
@manuelakuhn, see https://github.com/spack/spack-packages/pull/5962#pullrequestreview-4891651717.

Another alternative for the future could "`g++ --include stdint.h ...`", I guess.

PR #5974, opened 2026-08-07
bowtie2: Prevent compilation errors with gcc@:11 for version @2.5.5

emwjacobson:
`bowtie@2.5.5` introduces the use of `__builtin_cpu_supports("x86-64-v3")`, which fails for versions of GCC <= 11. A build flag was added which should avoid this block entirely.

This already seems to be fixed in the master branch, so I'd be fairly confident in keeping it strictly locked to the 2.5.5 version.

Possibly relevant readings:
https://github.com/BenLangmead/bowtie2/issues/522


bernhardkaindl (approved):
Ack, verifed the problem and the fix. Restricted to the affected x86_64 platform and the old gcc compilers.

PR #5978, opened 2026-08-07
`[WIP]` Replace self.compiler with self[cxx].cxx [or fortran flags, etc] for mixed-compiler builds

dtaller:
As per discussions with @becker33 , self[cxx].cxx  is favored over self.compiler.cxx . 

Greg said:
"You want 

self[“cxx”].cxx
self[“fortran”].fortran

self.compiler can return the wrong compiler when you’re combining different c/cxx and fortran compilers. We have it for backwards compatibility but you should avoid it in new code. "


dtaller:
@spackbot fix style

repos/spack_repo/builtin/packages/ace/package.py
```diff
@@ -34,9 +34,9 @@ def edit(self, spec, prefix):
         # Dictionary mapping: compiler-name : ACE config-label
         supported = {"intel": "_icc", "gcc": ""}
 
-        if self.compiler.name not in supported:
+        if self.spec.compiler.name not in supported:
```
dtaller, line 37: Is it safe to replace cases with "compiler.name" with sef["cxx"].cxx? AI says it is not "self["cxx"].cxx is the compiler executable path, not the compiler family name. It’s enough when you want to pass the compiler through directly, but not when we need to answer “which compiler is this?” ... self["cxx"].cxx is less robust for symlinked compiler names. Short version: self["cxx"].cxx by itself is not sufficient for compiler-identity checks; it needs either path normalization or a different compiler-family API."

bernhardkaindl:
Hi @dtaller, a great change I guess! Because it's big I've a few thoughts about it:

As far as I saw, bigger changes have in the past mostly done by core developers, so I think it should at have a core developer guiding this, but only by self-assignment of course.

Just ideas:
- Ideally, first update the documentation updated to no longer give self.compiler as example:
  https://github.com/search?q=repo%3Aspack%2Fspack+self.compiler+path%3Alib%2Fspack%2Fdocs&type=code
- Add a note referring about this to other PS that mention self.compiler: https://github.com/spack/spack-packages/pulls?q=is%3Apr+is%3Aopen+self.compiler
- For reference, it would be best if the explanation was referred to and reachable from the PR.
- Provide how the changes were created, to be sure: A script, show it source, or a KI model?
  - This would allow the changes to be recreated locally using that script so the PR branch could be compared against them.

dtaller:
This was done using codex -m gpt-5.4 -c model_reasoning_level="xhigh" .
I had codex make a session log file SELF_COMPILER_SESSION_LOG.md , which at least gives some documentation for my general prompt and a starting point for others to continue

dtaller:
I opened a small MR https://github.com/spack/spack/pull/52870 with just the documentation changes you suggested

PR #5990, opened 2026-08-08
r-libpressio: newer versions need newer gcc; libpressio-tools: needs c compiler

w8jcik:
* Prevents use of GCC <= 9 with newer versions or `r-libpressio`.
* Adds C compiler as dependency of `libpressio-tools` otherwise all versions fail to build.


bernhardkaindl (approved):
Approving as only fixing the contraints of the build.

PR #5993, opened 2026-08-09
HDF5: add versions 2.0.0, 2.1.1, and 2.2.0

Growl1234:
- Add versions 2.0.0, 2.1.1, and 2.2.0
- Update develop branches to select
- Drop preferred from 1.14.6 (attempt; might be reverted if CI fails)

### DIFF
--- a/repos/spack_repo/builtin/packages/hdf5/package.py
+++ b/repos/spack_repo/builtin/packages/hdf5/package.py
@@ -33,16 +33,31 @@ class Hdf5(CMakePackage):
 
     # The 'develop' version is renamed so that we could uninstall (or patch) it
     # without affecting other develop version.
-    version("develop-2.0", branch="develop")
+    version("develop-2", branch="develop")
     version("develop-1.14", branch="hdf5_1_14")
     version("develop-1.12", branch="hdf5_1_12")
     version("develop-1.10", branch="hdf5_1_10")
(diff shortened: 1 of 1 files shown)


hyoklee (approved):

Growl1234:
Looks like the failure might be due to imcompatibility with HDF5 2.x versions? I'll restore the `preferred` for `1.14.6` for now...

hyoklee (approved):

lrknox (changes requested):
The added develop-* branches will access minor or patch release branches that will soon disappear.  These develop-2.*.* versions are essentially duplicates of the 2.2.0, 2.1.1, and 2.0.0 versions added below.  There should be just one version("develop-2", branch="develop) for the develop branch, which will have code for version 2.x.x until the future next major version 3.0.0 release.

The previously existing develop-* branches were support branches for versions 1.8, 1.10, 1.12, 1.14; each of those was a different major version.  HDF5 made the switch to semantic versioning as of version 2.0.0, changing the version number format to be Semantic Versioning compliant (https://semver.org).  HDF5 versions from 2.0.0 forward will be Major.Minor.patch from the previous HDF5 1.Major.Minor.  Probably the version string for the develop branch should be "develop-2".

hyoklee (approved):

lrknox (approved):

RikkiButler20 (approved):

PR #5997, opened 2026-08-10
numdiff: Fix build for gcc@15+: Set C standard to C17 [TODO: use `def flag_handler`]

gshipman:
Added C17 standard flag to configure arguments to avoid conflicts with legacy code due to GCC 15+.

### DIFF
--- a/repos/spack_repo/builtin/packages/numdiff/package.py
+++ b/repos/spack_repo/builtin/packages/numdiff/package.py
@@ -31,6 +31,10 @@ class Numdiff(AutotoolsPackage):
     def configure_args(self):
         spec = self.spec
         args = []
+        # Fix for GCC 15+ defaulting to C23 where 'false', 'true', 'bool' are keywords
+        # Force C17 standard to avoid conflicts with legacy code
+        cflags = "-std=c17"
+        
         if "+nls" in spec:
             args.append("--enable-nls")
         else:


repos/spack_repo/builtin/packages/numdiff/package.py
```diff
@@ -31,6 +31,10 @@ class Numdiff(AutotoolsPackage):
     def configure_args(self):
         spec = self.spec
         args = []
+        # Fix for GCC 15+ defaulting to C23 where 'false', 'true', 'bool' are keywords
+        # Force C17 standard to avoid conflicts with legacy code
+        cflags = "-std=c17"
+
```
bernhardkaindl, line 37: Hi @gshipman:
```suggestion
```
Setting the C std to 17 (for gcc@15:) would be the easiest approach, but this unless other ways do not work for the package `configure_args()` is not longer the preferred place do do this in current spack. In this current form, the new cflags variable this adds is never returned or applied, so this diff is is dead code.

The preferred way would be like it is done in #5942 (if it works), which is to add an implementation flag_hander() function, and which then defines the flag to be set, but only for the compilers it is needed for.

bernhardkaindl (changes requested):

repos/spack_repo/builtin/packages/numdiff/package.py
bernhardkaindl, line 30: Based on #5942, this is the modern way to do this in spack, but using std=c17 instead of gnu17 and based on your submitted explanatory comment:
```suggestion

    def flag_handler(self, name, flags):
        # Fix for GCC 15+ which is defaulting to C23 where 'false', 'true', 'bool' are keywords
        if name == "cflags" and self.spec.satisfies("%gcc@15:"):
            flags.append("-std=c17")
        return (flags, None, None)

```
This fixes the build of `numdiff ~gmp ~nls` on all compilers I tested so far.
However, `numdiff +gmp` fails a multiple definition linking error.

bernhardkaindl (changes requested):

bernhardkaindl:
Hi @gshipman, I've set the draft flag only temporary delay running the full CI before the PR is ready for CI. It will be enabled after the fix is verified and the change is approved.

repos/spack_repo/builtin/packages/numdiff/package.py
gshipman: Thanks @bernhardkaindl , I will give this a try.

repos/spack_repo/builtin/packages/numdiff/package.py
gshipman: Trying this version (flag_handler). Thx!

PR #6003, opened 2026-08-10
relion: fix +gui X11 deps, PYTHON_EXE_PATH, and a stray -D__INTEL_COM…

FlyinTeller:
- depends_on libx11/xproto when +gui: cmake_args() references self.spec['libx11']/['xproto'] but neither was declared, crashing with KeyError during an actual +gui install.
- PYTHON_EXE_PATH now points at python-venv's interpreter instead of the bare python package: the bare interpreter has none of py-relion/topaz-3dem/model-angelo/dynamight's own site-packages on its sys.path, so every relion_python_* wrapper script fails at runtime with ModuleNotFoundError even though the build succeeds.
- Strip -D__INTEL_COMPILER from cmake/BuildTypes.cmake's nvcc flags for +cuda builds: meant only to stop fftw3.h emitting a __float128 typedef nvcc can't parse, already unnecessary since fftw 3.3.11 guards that on !defined(__CUDACC__). Left in place it also fools CUDA 12.9's own libcu++ headers into an Intel-only branch invalid under GCC 13, breaking every .cu file with "use of built-in trait '__remove_reference(_Tp)' in function signature; use library traits instead" (a real, reproduced build failure). Note: this exact issue was independently fixed upstream in RELION 5.1.0 (relion@5e18e811) - this patch is still needed for the 5.0.1 release this recipe builds.


repos/spack_repo/builtin/packages/relion/package.py
```diff
@@ -187,6 +205,24 @@ def patch(self):
...
+            # 07ef78dc1b273a40fb4f7db1797d12d3423b1f40 ("fix #18 (disable
+            # float128 for CUDACC)", 2014-07-08) - first released in FFTW
+            # 3.3.5, so the define is redundant for any FFTW this recipe
+            # would reasonably resolve today (enforced by the depends_on
+            # floor added below). Left in place it instead actively breaks
+            # things: CUDA 12.9's own libcu++ headers branch on
+            # __INTEL_COMPILER too, taking a path invalid under GCC 13 and
+            # failing every .cu file ("use of built-in trait
+            # '__remove_reference(_Tp)' in function signature; use library
+            # traits instead") - reproduced with gcc@13.4 + cuda@12.9.2.
```
bernhardkaindl: You should rewrite all these comments.

They make it think an LLM might have formatted them to conform to 77 column width.

- This is wrong:
  - Spack allows a width of up to 99 columns, Use them, especially for such long comments.
  - It also misses some punctiations and flow improvements.
    For example from my limited understanging, it should be:
    ```patch
    -    Left in place it instead actively breaks things:
    +    If left in place, it would actively break the following:
    ```

These comments also look like justificaitons that belong to the PR description,
not a code comment because they describe the change, which the PR should describe.
Instead, to comments should describe the new status quo of the code after the change.
Long descriptions of changes or newly added / changed should start with explainin answering the "Why" of the new code or the change in case of the PR description.

You might experiment with adding comments on the changes using review comments in GitHub "Files changes" tab instead.

bernhardkaindl (changes requested):
I think all comments should be made structured, use the extended text width beyond 77 chars, up to 99 chars per line instead to allow for more visual structure.

repos/spack_repo/builtin/packages/relion/package.py
```diff
@@ -187,6 +205,24 @@ def patch(self):
         if self.spec.satisfies("%nvhpc"):
             filter_file("-std=c99", "-c99", "src/apps/CMakeLists.txt")
 
+        if "+cuda" in self.spec:
+            # cmake/BuildTypes.cmake unconditionally adds -D__INTEL_COMPILER
```
bernhardkaindl: BTW, if this comment is reformatted and kept, it an be moved above the `if`:
This would give it 4 more columns it can use as one indent level less above the `if`
and as mentioned it should structured, not a wall of bricks and use the up to 99 columns.

repos/spack_repo/builtin/packages/relion/package.py
FlyinTeller: Thanks for the detailed feedback, I rewrote the comments in the amended commit

repos/spack_repo/builtin/packages/relion/package.py
FlyinTeller: done

repos/spack_repo/builtin/packages/relion/package.py
```diff
@@ -187,6 +201,20 @@ def patch(self):
...
+        # __CUDACC__ itself) since commit 07ef78dc1b273a40fb4f7db1797d12d3423b1f40 ("fix #18
+        # (disable float128 for CUDACC)", 2014-07-08), first released in FFTW 3.3.5 - so the
+        # define is redundant for any FFTW this recipe would resolve today (enforced by the
+        # depends_on floor above). If left in place, it actively breaks the following instead:
+        # CUDA 12.9's own libcu++ headers branch on __INTEL_COMPILER too, taking a path invalid
+        # under GCC 13 and failing every .cu file ("use of built-in trait
+        # '__remove_reference(_Tp)' in function signature; use library traits instead") -
+        # reproduced with gcc@13.4 + cuda@12.9.2.
+        if "+cuda" in self.spec:
+            filter_file(r"-D__INTEL_COMPILER ", "", "cmake/BuildTypes.cmake")
```
bernhardkaindl, line 216: I still find this too hard to read for humans.
- Firstly, don't need long-form git hashes in comments. Short ones would suffice.
- Instead of referring to it in prose, simply link to the commit to let the link do the talking.
- Start the paragraph with the problem that filter solves on a high level: e.g. "Fix compilation of CUDA 12.9's own libcu++ headers branch".

Given that, can the filter_file() be restricted to older CUDA versions so the filter can be removed by review only when CUDA 12.9 is removed in the future?
```suggestion
        # Fix compilation of CUDA 12.9's own libcu++ headers branch". See ...
        if self.spec.satisfies("+cuda@:12.9"):
            filter_file(r"-D__INTEL_COMPILER ", "", "cmake/BuildTypes.cmake")
```
In general, please use `self.spec.satisfies(...):` for new code like seen above in the existing code.

Use imperative style "Fix ..." not defensive "If left in place, .."

Please apply this to all added comments.

bernhardkaindl (changes requested):

PR #6004, opened 2026-08-10
py-relion: numpy2/aretomo-optional patches, corrected version ranges,…

FlyinTeller:
- numpy2-str0-compat.patch: scripts/eer_trajectory_handler.py uses np.str0, removed in numpy 2.0 (renamed to np.str_).
- aretomo-optional-import.patch: align_tilt_series/__init__.py eagerly imports both the AreTomo and IMOD alignment backends at module load - without py-lil-aretomo installed, the whole command fails to start even for IMOD-only use. Makes the AreTomo import optional.
- New +aretomo variant (default off): AreTomo is a closed-source prebuilt binary hard-tied to CUDA 10.1/10.2/11.1-11.8, conflicting with py-torch's CUDA 12+ requirement. IMOD-based alignment is the new default, via the new py-yet-another-imod-wrapper dependency.
- python range widened to 3.10:3.13.
- Version bounds across the dependency list come from a cross-repo API-compatibility audit of relion + relion-classranker + relion-blush + DynaMight + topaz + model-angelo (dual-venv hasattr/signature diffing, not just declared metadata).
- Added missing runtime deps found via ModuleNotFoundError at actual execution: py-scipy (get_particle_poses), py-dill (scripts/filament_selection.in), py-morphosamplers (get_particle_poses/spheres.py).
- py-pint version floor added (0.24:): pint's numpy-quantity glue code calls np.cumproduct at import time, removed in numpy 2.0 - broke napari's own import and cascaded into every relion_python_tomo_* wrapper that touches napari.
- py-torch bumped from the old non-Blackwell 2.0.1 pin to 2.7.1:2.13.0 (2.0.1 predates Blackwell kernel support).
- dynamight, topaz-3dem, model-angelo, py-relion-blush, py-relion-classranker wired in as runtime dependencies.


bernhardkaindl:

No historical / change descriptoins in the comments, they shall describe the new state after the change.

You might experiment with adding comments on the changes using review comments in GitHub "Files changes" tab instead.

FlyinTeller:
> @FlyinTeller, please rewrite the comments, with justifcaiton / why first, concise to the point. Use up to 99 columns per line, which is allowed by `ruff format` in this repo.
> 
> No historical / change descriptoins in the comments, they shall describe the new state after the change.
> 
> You might experiment with adding comments on the changes using review comments in GitHub "Files changes" tab instead.

@bernhardkaindl Finished reworking the comments.

repos/spack_repo/builtin/packages/py_relion/aretomo-optional-import.patch
```diff
@@ -0,0 +1,19 @@
...
++    from .aretomo import aretomo_cli
++except ImportError:
++    # AreTomo support is optional (py-lil-aretomo/aretomo pull in a hard, old
++    # CUDA requirement that conflicts with the rest of a modern build) - fall
++    # back gracefully so the IMOD-based backends (fiducials/patch tracking)
++    # still work without it.
++    aretomo_cli = None
+ from ._job_utils import write_global_output
+ 
+ # Then import the cli, it will be decorated with all programs (subcommands)
```
bernhardkaindl, line 19: I understand the description that the aretomo dependency does not work at all because it would pull in a hard, old CUDA requirement that conflicts with the rest of a modern build.

To me, this suggests that when aretomo isn't usable this patch can be replaced by a simple filter_file command in a `def patch(self):` of the recipe:
```py
    def patch(self, --fixme-add-args-and-test--):
        if self.spec.satisfies("~aretomo"):
            filter_file(
               "from .aretomo import aretomo_cli",
               "aretomo_cli = None",
               "src/tomography_python_programs/align_tilt_series/__init__.py"
            )
```

```suggestion
```

repos/spack_repo/builtin/packages/py_relion/package.py
```diff
@@ -26,40 +26,84 @@ class PyRelion(PythonPackage, CudaPackage):
     version("5.0.1", sha256="acbf898e96513b092514a56ff2a255c69a795e7a6f04131eacc8f55e2a900c23")
     version("5.0.0", sha256="5d02d529bfdb396204310b35963f35e5ec40ed9fd10bc88c901119ae7d7739fc")
 
+    # scripts/eer_trajectory_handler.py uses np.str0, removed in numpy 2.0 (renamed to np.str_).
+    patch("numpy2-str0-compat.patch")
```
bernhardkaindl, line 30: The description suggests:
```suggestion
    patch("numpy2-str0-compat.patch", when="^py-numpy@2:")
```
But even better, use https://github.com/spack/spack-packages/pull/6004/changes#r3803730686

repos/spack_repo/builtin/packages/py_relion/package.py
```diff
@@ -26,40 +26,84 @@ class PyRelion(PythonPackage, CudaPackage):
...
+
+    # align_tilt_series/__init__.py eagerly imports both the AreTomo and IMOD alignment
+    # backends at module load - without py-lil-aretomo installed, the whole command fails to
+    # start even for IMOD-only use. Makes the AreTomo import optional (try/except).
+    patch("aretomo-optional-import.patch")
+
```
bernhardkaindl, line 36:
```suggestion
```
See https://github.com/spack/spack-packages/pull/6004/changes#r3803662153

repos/spack_repo/builtin/packages/py_relion/numpy2-str0-compat.patch
```diff
@@ -0,0 +1,47 @@
...
+             hot_xs = tmp[0, :]
+             hot_ys = tmp[1, :]
+ 
+-        traj_star['hot_pixels']['rlnCoordinateX'] = list(hot_xs.astype(np.str0))
+-        traj_star['hot_pixels']['rlnCoordinateY'] = list(hot_ys.astype(np.str0))
++        traj_star['hot_pixels']['rlnCoordinateX'] = list(hot_xs.astype(np.str_))
++        traj_star['hot_pixels']['rlnCoordinateY'] = list(hot_ys.astype(np.str_))
+ 
+     return traj_star
+
```
bernhardkaindl, line 47: Like above, it looks like this can be replaced with a single

```py
    if self.spec.satisfies("^py-numpy@2:"):
       filter_file("np.str0", "np.str_", "scripts/eer_trajectory_handler.py"
```
(test yourself)

repos/spack_repo/builtin/packages/py_relion/package.py
```diff
@@ -26,40 +26,84 @@ class PyRelion(PythonPackage, CudaPackage):
...
 
     with default_args(type=("build", "run")):
-        depends_on("py-torchvision@0.15.2")
-        depends_on("py-tqdm@4.65.0")
-        depends_on("py-mrcfile@1.4.3")
+        # Floors below come from a cross-repo API-compatibility audit of
+        # environment_blackwell.yml (relion, relion-classranker, relion-blush, DynaMight,
+        # topaz, model-angelo) - lowest verified-working version for each. No ceilings unless
+        # a specific break is known: "verified compatible up to X" during that audit isn't
+        # evidence that X+1 breaks anything.
```
bernhardkaindl, line 58: IMHO, this comment does not say where the audited environment_blackwell.yml can be found. Just add the link to the audited yml file and say that the minium versions are the lowest verified-working version for each.

You don't need to justify that the version pins are in reality just minimums or snapshots from one version that happened to be working, and this kind of pinning isn't sustainable in spack. But that is clear and does not have to be said in a recipe comment. That belongs to the PR description only.
```suggestion
```

repos/spack_repo/builtin/packages/py_relion/package.py
```diff
@@ -26,40 +26,84 @@ class PyRelion(PythonPackage, CudaPackage):
...
+        depends_on("py-napari+all@0.4.18:")
+        # 5.15.x is PyQt5's own stable minor line; PyQt6 is a different API entirely, so cap
+        # before crossing that boundary.
+        depends_on("py-pyqt5@5.15.9:5")
+        depends_on("py-typer@0.9.0:")
+        depends_on("py-biopython@1.81:")
+        depends_on("py-seaborn@0.12.2:")
+        # numpy<2.0 doesn't support python@3.13. The only removed-symbol breaks found in the
+        # audit pool (this file's own np.str0, fixed via patch above; DynaMight's np.product,
+        # irrelevant here) are both fixed, so no ceiling.
```
bernhardkaindl, line 82: It looks like you gave an AI the target to make the relion work with python@3.13, which is good, but again that goal belongs in the PR description, it does not have to be said here. Please also remove apparently irrelevant comments like "DynaMight's np.product, irrelevant here".
```suggestion
```

repos/spack_repo/builtin/packages/py_relion/package.py
```diff
@@ -26,40 +26,84 @@ class PyRelion(PythonPackage, CudaPackage):
...
+        depends_on("py-numpy@1.26.1:")
         depends_on("py-click@:8.1")
         depends_on("py-mdocfile")
         depends_on("py-rich")
         depends_on("py-einops")
-        depends_on("py-lil-aretomo")
+        depends_on("py-lil-aretomo", when="+aretomo")
+        # Default tilt-series alignment backend (IMOD fiducials/patch tracking); AreTomo is opt-in.
```
bernhardkaindl, line 89:
```suggestion
        depends_on("py-lil-aretomo", when="+aretomo")
        # Default tilt-series alignment backend (IMOD fiducials/patch tracking)
```
That "aretomo" is opt-in is already clear from the default=False variant for it.

repos/spack_repo/builtin/packages/py_relion/package.py
```diff
@@ -26,40 +26,84 @@ class PyRelion(PythonPackage, CudaPackage):
...
+        depends_on("py-yet-another-imod-wrapper")
         depends_on("py-makefun")
         depends_on("py-lru-dict")
+        # scripts/filament_selection.in does `import dill as pickle` directly.
+        depends_on("py-dill")
```
bernhardkaindl, line 94: See https://github.com/spack/spack-packages/pull/6006#discussion_r3803493316: This does not mean that this script is installed. I assume yes, but it is not listed in pyproject.toml. Could you check that and clarify upstream maybe?

repos/spack_repo/builtin/packages/py_relion/package.py
```diff
@@ -26,40 +26,84 @@ class PyRelion(PythonPackage, CudaPackage):
...
+        # pint's numpy-quantity glue code calls np.cumproduct at import time, removed in numpy
+        # 2.0 (renamed to cumprod) - breaks napari's own import (napari imports pint) and
+        # cascades into every relion_python_tomo_* wrapper that touches napari. Fixed in
+        # pint 0.24.
+        depends_on("py-pint@0.24:")
+        # get_particle_poses/spheres.py imports morphosamplers directly.
+        depends_on("py-morphosamplers")
```
bernhardkaindl, line 101: Just go by https://github.com/3dem/relion/blob/master/pyproject.toml#L39, no need to give the file importing it as a comment for each dependency fix.

https://spack.readthedocs.io/en/latest/build_systems/pythonpackage.html#dependencies says:
> ## dependencies under `[project]`
> 
> These packages are required for building and installation.
> You can add them with type=("build", "run").

That means: Please iterate over https://github.com/3dem/relion/blob/master/pyproject.toml#L25-L39 and add them inside a context manager for "build" and "run":

```suggestion
    with default_args(type=("build", "run"):
        depends_on("py-numpy@1.26.1:")
        ...
        depends_on("py-morphosamplers")
```
I'm stopping the review here, please apply this to all project dependencies as documented.

bernhardkaindl (changes requested):

bernhardkaindl:
@FlyinTeller, sorry for the amount of review, but there is a lot in this PR at once. You'll get there.

It guess might be easier and review would be easier when targeting a much smaller change at first, to fix the build with numpy@1 in a first PR and only once that is clean and merged to go for numpy@2 and python3.13

PR #6005, opened 2026-08-10
model-angelo: missing torch dependency, esmfold gating, pyhmmer/numpy…

FlyinTeller:
- py-torch added: model-angelo imports torch directly (41 files), previously only available transitively by accident.
- py-fair-esm constrained to ~esmfold: model-angelo only uses esm.pretrained.* for checkpoint loading, never ESMFold structure prediction.
- py-pyhmmer bumped to the exact 0.11.0 pin: earlier versions have no cp313 wheel and fail to build from source on python@3.13 (removed CPython C-API symbols).
- py-numpy ceiling widened past <2.0 to 2.2.6 (numpy<2.0 doesn't support python@3.13).


bernhardkaindl:

No change descriptoins in the comments, they shall describe the new state after the change.

FlyinTeller:
> @FlyinTeller, please rewrite the comments, with justifcaiton / why first, concise to the point. Use up to 99 columns per line, which is allowed by `ruff format` in this repo.
> 
> No change descriptoins in the comments, they shall describe the new state after the change.

@bernhardkaindl Done :)

repos/spack_repo/builtin/packages/model_angelo/package.py
```diff
@@ -21,14 +21,23 @@ class ModelAngelo(PythonPackage):
...
+    # model-angelo only uses esm.pretrained.* for checkpoint loading, never ESMFold structure
+    # prediction, so the optional deepspeed/pytorch-lightning/... stack isn't needed.
+    depends_on("py-fair-esm@1.0.3~esmfold", type=("build", "run"))
+    # pyhmmer<0.11 has no cp313 wheel and fails to build from source on python@3.13 (removed
+    # CPython C-API symbols). 0.11.0 is the lowest version with a cp313 wheel.
+    depends_on("py-pyhmmer@0.11.0", type=("build", "run"))
     depends_on("py-loguru", type=("build", "run"))
-    depends_on("py-numpy@1.24.4:1", type=("build", "run"))
+    # numpy<2.0 doesn't support python@3.13. No ceiling beyond that - not aware of a specific
+    # break past 2.2.6.
```
bernhardkaindl, line 42: @FlyinTeller, Please remove this comment, currently it is confusing, as it is not clear from the context that this is about supporting Python 3.13.

The only point I see would be that this change says that model-angelo supports `numpy@2:` without a version limit, and that change speaks for itself and needs no comment.

repos/spack_repo/builtin/packages/model_angelo/package.py
```diff
@@ -21,14 +21,23 @@ class ModelAngelo(PythonPackage):
...
     depends_on("py-mrcfile", type=("build", "run"))
     depends_on("py-pandas", type=("build", "run"))
-    depends_on("py-fair-esm@1.0.3", type=("build", "run"))
-    depends_on("py-pyhmmer@0.7.1", type=("build", "run"))
+    # model-angelo only uses esm.pretrained.* for checkpoint loading, never ESMFold structure
+    # prediction, so the optional deepspeed/pytorch-lightning/... stack isn't needed.
+    depends_on("py-fair-esm@1.0.3~esmfold", type=("build", "run"))
+    # pyhmmer<0.11 has no cp313 wheel and fails to build from source on python@3.13 (removed
+    # CPython C-API symbols). 0.11.0 is the lowest version with a cp313 wheel.
+    depends_on("py-pyhmmer@0.11.0", type=("build", "run"))
```
bernhardkaindl, line 39: The only comment that would be needed here would be why to hard-code the version again (before, it was 0.7.1, now 0.11.0): As this shows, there was never a fixed version pin requirement in the first place. Therefore I recommand to cleanup the comment (it is enough to tell this in the PR description) and do not pin the version - allow newer versions unless there is a real reason for the pin - in which case the pinnning would require a justifying comment, I think:
```suggestion
    depends_on("py-pyhmmer@0.11:", type=("build", "run"))
```

bernhardkaindl (changes requested):

PR #6006, opened 2026-08-10
py-relion-classranker: fix numpy/matplotlib version conflicts

FlyinTeller:
Edit by @bernhardkaindl:
- This PR fixes concretization of `py-relion-classranker` in spack by fixing the current incorrect pinned version dependency for `py-numpy@1.24.4` which, in reality, is not exact, it is only the minimum, so the range must be open-ended by appending the missing `:` to the required minimum version.
- py-matplotlib does not show to to required as it is only used in tools that are neither installer installed, nor used or tested in any way.

----
Submitter's description:
- py-numpy widened from an exact 1.24.4 pin (conflicting with py-relion's own numpy floor) to a range matching py-relion/model-angelo.
- py-matplotlib added: classranker's own training/test_*.py import it directly, with a floor past 3.8.3 (which caps numpy at <2.0).

### DIFF
--- a/repos/spack_repo/builtin/packages/py_relion_classranker/package.py
+++ b/repos/spack_repo/builtin/packages/py_relion_classranker/package.py
@@ -26,4 +26,9 @@ class PyRelionClassranker(PythonPackage):
 
     depends_on("py-torch@2.0.1:", type=("build", "run"))
     depends_on("py-torchvision@0.15.2:", type=("build", "run"))
-    depends_on("py-numpy@1.24.4", type=("build", "run"))
+    # classranker's requirements.txt only declares numpy>=1.24.4 (no ceiling); numpy<2.0
+    # doesn't support python@3.13, so the floor is raised to match py-relion's own numpy floor.
+    depends_on("py-numpy@1.26.1:", type=("build", "run"))
+    # classranker's training/test_*.py import matplotlib directly. Floor is past 3.8.3, which
+    # caps numpy at <2.0.
+    depends_on("py-matplotlib@3.9.0:", type=("build", "run"))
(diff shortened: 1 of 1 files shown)


bernhardkaindl:

No change descriptoins in the comments, they shall describe the new state after the change.

repos/spack_repo/builtin/packages/py_relion_classranker/package.py
```diff
@@ -26,4 +26,9 @@ class PyRelionClassranker(PythonPackage):
 
     depends_on("py-torch@2.0.1:", type=("build", "run"))
     depends_on("py-torchvision@0.15.2:", type=("build", "run"))
-    depends_on("py-numpy@1.24.4", type=("build", "run"))
+    # 1.26.1 matches py-relion's own numpy floor: numpy<2.0 doesn't support python@3.13.
+    # classranker's requirements.txt itself only declares numpy>=1.24.4, with no ceiling.
+    depends_on("py-numpy@1.26.1:", type=("build", "run"))
+    # classranker's training/test_*.py import matplotlib directly; matplotlib up to 3.8.3 caps
+    # numpy at <2.0, conflicting with the numpy floor above.
+    depends_on("py-matplotlib@3.9.0:", type=("build", "run"))
```
bernhardkaindl, line 34: Hi @FlyinTeller,

1. The constraints of dependents of this package are not relevant to the dependencies of this package. 
2. It is true that classranker's training/test_*.py imports matplotlib, but those files are not installed and no test uses them. Hence, the claimed dependency is not a actual dependency.
   ```sh
   $ opt/spack/linux-x86_64_v4/py-relion-classranker-20250108-*
   $ grep -r matplotlib lib/python3.13/site-packages/relion_classranker
   (no finds)
   ```

The only bug here is that the numpy dep was pinned to a single version. Just add the missing ":".
Remove every part that is relevant for this package itself.

The only comment this PR would have needed is that the dependency that the package declares in `setup.cfg` on py-numpy's only has 1.24.4 as minimum version and this fix is needed to build the package in (it does not concretize without this fix):

```suggestion
    depends_on("py-numpy@1.24.4:", type=("build", "run"))
```

In case you used AI (e.g. Claude Code), please ensure that submitted PRs:
- explain the issue fixed in terms on how that fixes the issue from spack's perspective.
- and apply only changes validated to have the needed impact on the installed package.

bernhardkaindl (changes requested):

PR #6007, opened 2026-08-10
topaz-3dem: numpy2 compat patch, torch version alignment, drop unused…

FlyinTeller:
- topaz-numpy2-copy-compat.patch: np.array(x, copy=False) at 15 call sites across 8 files - numpy<2.0 treats this as "avoid a copy when possible", numpy>=2.0 as "never copy, raise if one would be required", a silent behavior change. np.asarray(x) is numpy's documented migration-guide replacement.
- py-torch range changed from 1:2.3 (conflicting with py-relion's own torch range) to 2.7.1:2.13.0.
- py-future removed: unused - every "future"-looking match in topaz's source is a stdlib `from __future__ import` statement.


bernhardkaindl:

FlyinTeller:
@bernhardkaindl did as requested

repos/spack_repo/builtin/packages/topaz_3dem/package.py
```diff
@@ -19,13 +19,21 @@ class Topaz3dem(PythonPackage):
...
     depends_on("py-torchvision", type=("build", "run"))
     depends_on("py-numpy@1.11:", type=("build", "run"))
     depends_on("py-pandas", type=("build", "run"))
     depends_on("py-scikit-learn@0.19.0:", type=("build", "run"))
     depends_on("py-scipy@0.17.0:", type=("build", "run"))
     depends_on("py-pillow@6.2.0:", type=("build", "run"))
-    depends_on("py-future", type=("build", "run"))
     depends_on("py-scikit-image", type=("build", "run"))
+    # No py-future dependency: every "future"-looking match in topaz's source is a stdlib
+    # `from __future__ import` statement, not this package.
```
bernhardkaindl, line 39: Remove the comment: Future was only for Python 2 to 3 migration. Need to for a comment justify. If there are old version that still needs it, please deprecate those and make it when= for these old versions instead.

repos/spack_repo/builtin/packages/topaz_3dem/package.py
```diff
@@ -19,13 +19,21 @@ class Topaz3dem(PythonPackage):
 
     version("20220325", commit="14b2bc331768b67b3267397523de57c707fa1253")
 
+    # np.array(x, copy=False) at 15 call sites across 8 files: numpy<2.0 treats this as "avoid
+    # a copy when possible", numpy>=2.0 as "never copy, raise if one would be required" - a
+    # silent behavior change. np.asarray(x) is numpy's documented migration-guide replacement.
+    patch("topaz-numpy2-copy-compat.patch")
```
bernhardkaindl, line 25: Is it possible to replace the parch with a filer_file() call?

Shorten the comment to 1-2 lines if possible.

repos/spack_repo/builtin/packages/topaz_3dem/package.py
```diff
@@ -19,13 +19,21 @@ class Topaz3dem(PythonPackage):
...
+
     depends_on("py-setuptools", type="build")
-    depends_on("py-torch@1:2.3", type=("build", "run"))
+    # No ceiling: only verified up to 2.13.0, no specific break known beyond it. A ceiling here
+    # would conflict with py-relion's own torch range.
```
bernhardkaindl, line 29: Replace with a reference / link to the setup.cfg or pyproject.toml. It is the default authority for deps.

bernhardkaindl (changes requested):
Please check the new review comments.

PR #6008, opened 2026-08-10
py-fair-esm: gate ESMFold-only dependencies behind +esmfold

FlyinTeller:
Only ESMFold structure prediction needs deepspeed/pytorch-lightning/dm-tree/omegaconf/ml-collections; plain checkpoint loading (esm.pretrained.*) doesn't. Previously unconditional, dragging the whole stack into every consumer regardless of whether ESMFold is used.

### DIFF
--- a/repos/spack_repo/builtin/packages/py_fair_esm/package.py
+++ b/repos/spack_repo/builtin/packages/py_fair_esm/package.py
@@ -22,11 +22,10 @@ class PyFairEsm(PythonPackage):
 
     variant("esmfold", default=True, description="Enable dependencies for OpenFold")
     depends_on("py-biopython@1.79:", when="+esmfold", type=("build", "run"))
-
-    depends_on("py-deepspeed", type=("build", "run"))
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/py_fair_esm/package.py
```diff
@@ -21,12 +21,15 @@ class PyFairEsm(PythonPackage):
     depends_on("py-setuptools@59.5.0:", type=("build"))
 
     variant("esmfold", default=True, description="Enable dependencies for OpenFold")
+    # Only ESMFold structure prediction needs these; plain checkpoint
+    # loading (esm.pretrained.*) doesn't. Upstream leaves them
+    # unconditional, dragging deepspeed/pytorch-lightning/... into every
+    # consumer regardless of whether ESMFold is used.
```
bernhardkaindl: Such justifications are not added to recipes in spack, they are in the PR description
```suggestion
```

PR #6009, opened 2026-08-10
py-napari: drop unconditional/ungated py-sphinx dependency

FlyinTeller:
sphinx (a doc-build tool, not needed at install/run time) was declared unconditionally, with no "docs"-style variant gating it.

### DIFF
--- a/repos/spack_repo/builtin/packages/py_napari/package.py
+++ b/repos/spack_repo/builtin/packages/py_napari/package.py
@@ -74,7 +74,6 @@ class PyNapari(PythonPackage):
         depends_on("py-scipy@1.5.4:")
         depends_on("py-superqt@0.7.4:", when="@0.6:")
         depends_on("py-superqt@0.4.1:")
-        depends_on("py-sphinx@4.3.0:4", when="@0.4:0.5")
         depends_on("py-tifffile@2022.7.28:", when="@0.6:")
         depends_on("py-tifffile@2020.2.16:")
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/py_napari/package.py
```diff
@@ -74,7 +74,9 @@ class PyNapari(PythonPackage):
         depends_on("py-scipy@1.5.4:")
         depends_on("py-superqt@0.7.4:", when="@0.6:")
         depends_on("py-superqt@0.4.1:")
-        depends_on("py-sphinx@4.3.0:4", when="@0.4:0.5")
+        # sphinx (a doc-build tool, not needed at install/run time) was
+        # declared here unconditionally, with no "docs"-style variant
+        # gating it - removed.
```
bernhardkaindl: Makes sense. However, the process says the maintainer should be asked, I guess.

@Markus92, as maintainer, do you agree? I shall give you about a week for a comment, ok?

I'll set the draft status until @Markus92 has given a comment or 👍 👎 reaction to not run CI until then.

The comment for the removal only is needed in the PR description, not in the comment in the recipe, so I suggest to remove it:
```suggestion
```

bernhardkaindl (approved):
Given time for review by maintainer, but 3 business days passed. I rebased the PR branch onto the latest develop to ensure PR runs with the latest fixes for the CI pipelines, they should start after a few hours and I'll enable auto-merge after CI pass.

PR #6010, opened 2026-08-10
py-makefun: widen stale setuptools ceiling

FlyinTeller:
No evidence makefun needs anything past basic setuptools support; removed the @39.2:71 upper bound.

### DIFF
--- a/repos/spack_repo/builtin/packages/py_makefun/package.py
+++ b/repos/spack_repo/builtin/packages/py_makefun/package.py
@@ -20,5 +20,7 @@ class PyMakefun(PythonPackage):
 
     version("1.16.0", sha256="e14601831570bff1f6d7e68828bcd30d2f5856f24bad5de0ccb22921ceebc947")
 
-    depends_on("py-setuptools@39.2:71", type="build")
+    # was @39.2:71 - no evidence makefun needs anything past basic
+    # setuptools support, removed the stale ceiling.
+    depends_on("py-setuptools@39.2:", type="build")
     depends_on("py-setuptools-scm", type="build")


repos/spack_repo/builtin/packages/py_makefun/package.py
```diff
@@ -20,5 +20,7 @@ class PyMakefun(PythonPackage):
 
     version("1.16.0", sha256="e14601831570bff1f6d7e68828bcd30d2f5856f24bad5de0ccb22921ceebc947")
 
-    depends_on("py-setuptools@39.2:71", type="build")
+    # was @39.2:71 - no evidence makefun needs anything past basic
+    # setuptools support, removed the stale ceiling.
+    depends_on("py-setuptools@39.2:", type="build")
```
bernhardkaindl, line 25: To get the stated reason for this upper bound, all you had to do was to look at the source:
https://github.com/smarie/python-makefun/blob/1.16.0/pyproject.toml#L3
```py
    "setuptools>=39.2,<72",   # later versions do not read 'tests_require' from setup.cfg anymore
```
Points:
- It does what upstream declared.
- setuptools is only a build dependency, it does not apply to other packages.
- besides the potentially extra setuptools build, there is no further cost
- and other packages might also need an older version, so it should be near-zero.
- There is also evidence that `setuptools@81:` is not working with version 1.16.0 of the package: https://github.com/smarie/python-makefun/blob/1.16.0/setup.py#L19
  ```py
  import pkg_resources

  pkg_resources.require("setuptools>=39.2")
  pkg_resources.require("setuptools_scm")
  ```

  - `pkg_resources` is only provided until `setuptools@:80`
  - even if `:71` would not apply for spack `:80` does apply:
  ```py
  $ spack install -v py-makefun ^py-setuptools@81:
  ...
    ModuleNotFoundError: No module named 'pkg_resources'
  ```

Please reopen the PR if this range needs to be fixed so solve a problem.

bernhardkaindl (changes requested):

bernhardkaindl:
See the review for details: https://github.com/spack/spack-packages/pull/6010#discussion_r3755128527

Summary:
- The stated range is correct, there is a reference why this upper bound exists.
- There is evidence to the contrary that an upper bound is required: `import pkg_resources`

Closing as I don't another valid change in this PR.

repos/spack_repo/builtin/packages/py_makefun/package.py
FlyinTeller: Thank you for the review.

The line you quote from https://github.com/smarie/python-makefun/blob/1.16.0/pyproject.toml#L3 mentions that the cap at 71 is only to be able to read `test_require`, which is not relevant for a spack installed package. 

I agree that `setuptools@81:` is a good point, also mentioned at https://github.com/smarie/python-makefun/issues/122

I would suggest to actually fix both and apply a similar fix to https://github.com/alpinelinux/aports/commit/9763023f192c6efe3eeb76289f82fd5daa0eb51b to completely remove the setuptools cap

I hit an actual issue when trying to concretize `relion` on our system, which depends on `py-relion` which in turn depends on `py-torch` and `py-makefun`. Pytorch has quite strict setuptools requirements, and it failed to concretize with different setuptools versions, hence I had to broaden the range  for this package.

PR #6011, opened 2026-08-10
tsne-cuda: propagate cuda_arch to faiss, widen cmake ceiling

FlyinTeller:
- faiss (also a CudaPackage) otherwise defaults to cuda_arch=none, which its own cmake_args() passes through as -DCMAKE_CUDA_ARCHITECTURES=none - newer CMake rejects that literal string.
- cmake ceiling widened from 3.30 to 4.3.3: builds cleanly once the faiss cuda_arch fix above is in place.

### DIFF
--- a/repos/spack_repo/builtin/packages/tsne_cuda/package.py
+++ b/repos/spack_repo/builtin/packages/tsne_cuda/package.py
@@ -29,11 +29,8 @@ class TsneCuda(CMakePackage, CudaPackage, PythonExtension):
 
     depends_on("c", type="build")
     depends_on("cxx", type="build")
-    # was @3.20:3.30 ("CMake 3.31 messes a bit too much with find CUDA") -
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/tsne_cuda/package.py
```diff
@@ -29,14 +29,23 @@ class TsneCuda(CMakePackage, CudaPackage, PythonExtension):
 
     depends_on("c", type="build")
     depends_on("cxx", type="build")
-    depends_on("cmake@3.20:3.30", type="build")  # CMake 3.31 messes a bit too much with find CUDA
+    # No ceiling: with faiss getting a real cuda_arch (see below), builds are clean through
+    # cmake 4.3.3, the highest version tested; no specific break is known beyond it.
```
bernhardkaindl: Remove. No need to justify fixing the CMake version constraint as a comment.

PR #6012, opened 2026-08-10
pbzip2: fix Makefile CXXFLAGS/LDFLAGS injection and clang-family lite…

FlyinTeller:
This Makefile assigns CXXFLAGS/LDFLAGS with plain `=`, so neither depends_on-provided include paths nor flag_handler()-injected flags reach the compiler:
- bzip2's include path was never seen ("bzlib.h file not found") - now patched directly into the Makefile text, same as PREFIX is handled.
- pbzip2 uses C99 PRIuMAX macros without the C++11-required space (e.g. "%"PRIuMAX); clang-family compilers (including AMD's aocc) treat this as an error, suppressed via -Wno-reserved-user-defined-literal.


bernhardkaindl (approved):
Ack, checked, confirmed. Enabled workflows for auto-merge.

PR #6014, opened 2026-08-10
py-pandas: fix Cython>=3.1.0 requirement cutoff

FlyinTeller:
pandas' own pyproject.toml requires Cython>3.1.0 starting at 3.0.1, not 3.0.2+ as this recipe had it - 3.0.1 fell through to an older Cython incompatible with its @cython.critical_section directive (added in Cython 3.1).


adamjstewart (changes requested):
Can you delete the comment?

FlyinTeller:
@adamjstewart done

adamjstewart (approved):

PR #6015, opened 2026-08-10
py-grpcio: cap re2 version to avoid a C++14/17 cxxstd conflict

FlyinTeller:
re2 has no upper bound by default. Recent re2 (2025-11-05+) needs std::optional (C++17) in its public header, but grpcio's own setup.py hardcodes -std=c++14 for this version bracket - and abseil-cpp's cxxstd=14 default (shared with re2 on the same DAG node) can't be raised to 17 without conflicting with other consumers. Caps re2 at the last version still using absl::optional.


bernhardkaindl:
@FlyinTeller, JFYI, I don't see a submission of a re2 >= 2025-11-05 or similar for spack yet.

When adding that version and building it with gcc@16, of Ubuntu's ppa, the build still succeeds:
```py
spack install py-grpcio ^re2@2025-11-05
[+] tdi6qcs re2@2025-11-05 /home/bkaindl/spack/spack/opt/spack/linux-x86_64_v4/re2-2025-11-05-tdi6qcsuxj7qdxvvexwx22nmbvy24ipj (5s)
$ spack find re2   py-grpcio
-- linux-ubuntu24.04-x86_64_v4 / %c,cxx=gcc@16.0.1 --------------
py-grpcio@1.81.1  py-grpcio@1.81.1

-- linux-ubuntu24.04-x86_64_v4 / %cxx=gcc@16.0.1 ----------------
re2@2024-07-02  re2@2025-11-05
==> 4 installed packages
```
How does the described problem manifest? (for testing / checking)

repos/spack_repo/builtin/packages/py_grpcio/package.py
```diff
@@ -109,6 +109,13 @@ class PyGrpcio(PythonPackage):
     depends_on("zlib-api", when="@1.33.1:")
     depends_on("c-ares", when="@1.33.1:")
     depends_on("re2+shared", when="@1.34:")
+    # re2 has no upper bound by default. Recent re2 (2025-11-05+) needs
+    # std::optional (C++17) in its public header, but grpcio's own
+    # setup.py hardcodes -std=c++14 for this version bracket - and
+    # abseil-cpp's cxxstd=14 default (shared with re2 on the same DAG node)
+    # can't be raised to 17 without conflicting with other consumers. Cap
+    # re2 at the last version still using absl::optional instead.
+    depends_on("re2@:2024-07-02", when="@1.47:1.64")
```
bernhardkaindl: @FlyinTeller, PS: This is a good comment that describes why the new dependency constraint is needed. It could have taken less lines if all 99 columns would be used, but it is fine.

However, for the PR, it would be interesting to know what happens if "needs std::optional (C++17) in its public header" is violated. It apparently still builds but I guess the downstream users could run into issues. It sounds like you might want to add those new re2 versions as well, maybe with the last 2024 version marked as "`preferred=True`", I think untils it is tested that the new version does not cause issues.

FlyinTeller:
@bernhardkaindl 

The new re2 versions are now submitted in #6123 (2025-07-17 through 2025-11-05).

 While preparing that PR I checked the upstream tarballs: the absl::optional → std::optional switch in re2/re2.h actually happened in 2025-08-05, not 2025-11-05 — I've updated the comment in this PR accordingly. The @:2024-07-02 cap itself is unaffected.

On how the problem manifests: your test used `py-grpcio@1.81.1`, which is outside the capped bracket: for `@1.71:` this recipe pins `abseil-cpp cxxstd=17` and `grpcio` builds as `C++17`, so new `re2` works fine there (as your install shows). The cap only applies to `@1.47:1.64`, where grpcio's `setup.py` hardcodes `-std=c++14` and `abseil` (shared with `re2` on the same DAG node) is pinned to `cxxstd=14`. In that bracket a new `re2` fails at compile time twice over: the `re2` recipe derives `CMAKE_CXX_STANDARD` from `abseil's` cxxstd, so `re2` itself gets built as `C++14`, and 
 `grpcio's` extension compiles `re2`/`re2.h` at `C++14`. `std::optional` isn't available in either case. To reproduce: remove the cap and try `spack install py-grpcio@1.64.3 ^re2@2025-08-05`.
 
 Once #6123 lands, that combination is rejected at concretization for the old bracket, since it adds `conflicts("^abseil-cpp cxxstd=14")` for `re2@2025-08-05`.

bernhardkaindl (approved):
Without the clarifications from the PR description of #6123, the comment added with this PR change is missing vital context.

@FlyinTeller, as you showed that you used Claude Code, please review that what it writes does make sense not just to you but ot someone who have not done the anaysis behind the change. I'm recording this additional comment from the PR description from #6123 here for clarity:

> Starting with 2025-08-05, the public header re2/re2.h uses `std::optional` instead of `absl::optional`, ...

PR #6021, opened 2026-08-10
lp-solve: Fixed compiling with newer compilers

emwjacobson:
Newer compilers, specifically gcc 14+, didn't like compiling lp-solve. This fixes that. Tested with gcc versions 11 through 16.


bernhardkaindl (approved):
Ack, saw this too, fixes it, compiles also with gcc-9 and -10.

PR #6034, opened 2026-08-11
accfft and opensubdiv: Fix CUDA arch filtering.

RemiLacroix-IDRIS:
Explicitly setting `cuda_arch=none` would cause the CUDA arch not to be correctly filtered.

Brought up by @iamashwin99 in #5830.


bernhardkaindl (approved):

iamashwin99 (approved):
Since its the same idea and fix as in https://github.com/spack/spack-packages/pull/5830, looks good to me!

PR #6039, opened 2026-08-11
geant4-data: refactor to clarify version-dependent changes

sethrj:
This inverts the loops for the G4 datasets to reduce duplication (copy-pasting from one version to the next, risking omitting a dataset and highlighting changes in the datasets). I used claude to do the boring part.

The inversion also allows us to specify exact dataset version requirements easily, which fixes similar errors as #5223 (mismatch between g4 cmake variables and install paths). Using exact versions identified dataset requirements for **nonexistent** versions of G4 (viz. 10.3.0), so those have been fixed.

I verified some of the older and newer versions on the g4 dev repo with (e.g.) `git show geant4-10-03-patch-01:cmake/Modules/Geant4DatasetDefinitions.cmake` and `git show geant4-11-00:cmake/Modules/G4DatasetDefinitions.cmake`. Additionally I tested a couple of versions of the data with installs. Finding dataset version mismatches with a local `geant4` is easy: build with `-DGEANT4_INSTALL_DATADIR=${PREFIX}/share/data` (where `PREFIX` is the latest geant4-data install location) and see what datasets it lists and complains about. For the latest dev version (which isn't yet in spack):
```
-- Reusing dataset G4NDL (4.7.1)
-- Pre-configuring dataset G4EMLOW (9.0)
-- Reusing dataset PhotonEvaporation (6.1.2)
-- Reusing dataset RadioactiveDecay (6.1.2)
-- Reusing dataset G4PARTICLEXS (4.2)
-- Reusing dataset G4PII (1.3)
-- Reusing dataset RealSurface (2.2)
-- Reusing dataset G4SAIDDATA (2.0)
-- Reusing dataset G4ABLA (3.3)
-- Reusing dataset G4INCL (1.3)
-- Reusing dataset G4ENSDFSTATE (3.0)
-- Reusing dataset G4CHANNELING (2.0)
  *WARNING*
    Geant4 has been pre-configured to look for datasets
    in the directory:

    /opt/spack/var/spack/environments/geant4-dev/.spack-env/view/share/data

    but the following datasets are NOT present on disk at
    that location:

    G4EMLOW (9.0)
```


repos/spack_repo/builtin/packages/geant4_data/package.py
```diff
@@ -99,193 +99,134 @@ class Geant4Data(BundlePackage):
...
-        for _d in _dsets:
-            depends_on(_d, type=("build", "run"), when=_vers)
-
-    _datasets_tendl = {
-        "11.0:11.4": "g4tendl@1.4",
-        "10.4:10.7": "g4tendl@1.3.2",
-        "10.3:10.3": "g4tendl@1.3",
+    # Optional datasets with independent variants
+    _optional_datasets = {
+        "g4tendl": {
```
wdconinc, line 207: Should there be an open ended entry here?

sethrj: Good catch! double checking the other cases for that too.

wdconinc (approved):
LGTM. Leaving a few days for @drbenmorgan to review.

PR #6048, opened 2026-08-11
podio: add arrow variant

wdconinc:
This PR adds to `podio` a new variant `arrow` ("I/O backend" is a bit aspirational since it's in-memory only, but I've kept the same description as for the SIO backend).


wdconinc:
@spackbot fix style

jmcarcell (approved):

RikkiButler20 (approved):
LGTM

PR #4737, opened 2026-05-07
fms: add 2025.03, 2026.01, and 2026.01.01

rem1776:
Updates the fms package for the past 3 releases. Includes patches to disable the unit tests building by default.

Fixes #3596, https://github.com/NOAA-GFDL/FMS/issues/1865,

PR #6050, opened 2026-08-11
fms: patch cmake for +yaml

AlexanderHrabski-NOAA:
When compiling an application that depends on FMS 2025.03's yaml features using fms@2025.03+yaml, the build fails at the linking stage due to missing symbols from libyaml. This bug [was fixed in the upstream](https://github.com/NOAA-GFDL/FMS/pull/1822/changes#diff-5d3d35a2f345719a38a400a0160578378b3136b52db27fad76d1f5edc39a84f7), however the fix never made it into the affected 2025 (and older) versions.

This PR patches the missing `target_link_libraries`. I've confirmed this works for versions `2025.02:2025.04` (#4737 adds `2025.04:2026.01.01`). The patch should apply to older affected versions, but my testing was blocked by a separate bug. So, this PR only includes those versions I've tested. If support for older versions is desirable, I can open a new issue to sort out the separate bug. I can attach the MRE I used for testing on request.

Depends on #4737

### DIFF
--- a/repos/spack_repo/builtin/packages/fms/fms-link-libyaml.patch
+++ b/repos/spack_repo/builtin/packages/fms/fms-link-libyaml.patch
@@ -1,0 +1,13 @@
+--- a/CMakeLists.txt
++++ b/CMakeLists.txt
+@@ -417,6 +417,10 @@
+                                          NetCDF::NetCDF_Fortran
+                                          MPI::MPI_Fortran)
+
++  if(WITH_YAML)
++    target_link_libraries(${libTgt} PRIVATE ${LIBYAML_LIBRARIES})
++  endif()
++
+   if(OpenMP_Fortran_FOUND)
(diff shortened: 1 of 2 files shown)


climbfuji (approved):
Looks correct to me. Just ran into this issue today with the UFS!

mathomp4 (approved):

AlexanderHrabski-NOAA:
Thanks for having a look! I'll rebase once #4737 is in.

bernhardkaindl (approved):

bernhardkaindl:
Based on the linked PRs, this is ready for merge. I haven't seen a conflict, so a rebase of the PR does not appear to be needed.

 @AlexanderRichert-NOAA, @Hang-Lei-NOAA, @climbfuji, @edwardhartnett, I assume you all just wait for the merge, will merge soon.

PR #6051, opened 2026-08-11
kokkos: allow more values for deprecated_code

rbberger:
- `kokkos deprecated_code=none`
- `kokkos deprecated_code=auto` (default, use CMake defaults)
- `kokkos deprecated_code=current` (major_version)
- `kokkos deprecated_code=previous` (major_version-1)
- `kokkos deprecated_code=all` (major_version & major_version-1)

Alternative to https://github.com/spack/spack-packages/pull/5035

### DIFF
--- a/repos/spack_repo/builtin/packages/kokkos/package.py
+++ b/repos/spack_repo/builtin/packages/kokkos/package.py
@@ -154,7 +154,6 @@ class Kokkos(CMakePackage, CudaPackage, ROCmPackage):
         "debug": [False, None, "Activate extra debug features - may increase compiletimes"],
         "debug_bounds_check": [False, None, "Use bounds checking - will increase runtime"],
         "debug_dualview_modify_check": [False, "@:4", "Debug check on dual views"],
-        "deprecated_code": [False, "@:4", "Whether to enable deprecated code"],
         "hpx_async_dispatch": [False, "@:4", "Whether HPX supports asynchronous dispath"],
         "tuning": [False, None, "Create bindings for tuning tools"],
         "tests": [False, None, "Build for tests"],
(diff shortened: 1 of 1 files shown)


JBludau (approved):
For context: We discussed this in our build and packaging working group meeting and decided that we want the deprecation to follow the defaults that we set in CMake if users don't request otherwise.
As Kokkos has a grace period for new major releases, where both old and new API is enabled, we opted from changing the binary variant to allow using the default with "auto".
Nevertheless, users can select specific versions or nothing by using the other options.
We think this makes the deprecation options more consistent.

We did opt to not add a new option for this, as we think deprecated code should be used deliberately and that should be expressed by users

repos/spack_repo/builtin/packages/kokkos/package.py
```diff
@@ -344,11 +343,10 @@ class Kokkos(CMakePackage, CudaPackage, ROCmPackage):
         variant("cxxstd", default="17", values=("17", "20", "23"), when="@4")
         variant("cxxstd", default="20", values=("20", "23"), when="@5:")
 
-    # FIXME regroup variants and conflicts
     variant(
         "deprecated_code",
-        default=True,
-        when="@5:",
+        default="auto",
+        values=("none", "auto", "all", "previous", "current"),
```
nmm0, line 349: I think we should clarify in the description what these options do.

* none: disable all deprecated code
* auto: leave to cmake defaults (i believe this is the same as "current" but we shouldn't specify that)
* all: enable code deprecated in the provided spec major version and the major version prior to that
* current: enable code only deprecated in the provided spec major version
* previous: enable code only deprecated in the major version prior to the spec's major version

JBludau: I think it is clear enough, but I might be biased

dalg24 (commented):
Do we need to worry about backward compatibility?
What happens if someone configure expecting that it is still a boolean option?

JBludau:
> Do we need to worry about backward compatibility? What happens if someone configure expecting that it is still a boolean option?

They will see an error super early in their pipeline. It at least looks like other packages in the CI are not using it.
We think the transition to the new option is easy enough and would only be required if the defaults are not already sufficient.

repos/spack_repo/builtin/packages/kokkos/package.py
bernhardkaindl: @rbberger, just sending a reminder that this is an open conversation in this thread which needs to be resolved before a final review/merge.

Also, there is a backward compatibility conversation in the main thread of the PR.

cedricchevalier19:
We have discussed about this with @tpadioleau and we are quite sceptical about introducing such a complex option.
Do we have users asking for this?

IMO, we should not expose `deprecated_code` and simply enable the "=current". I think it is the most classical behaviour for consumers: the soon to be deleted code is flagged as deprecated but still available.

rbberger:
Kokkos Build and Packaging WG wants more time discussing this. Changing back to draft.

PR #6056, opened 2026-08-12
mpsolve: regenerate parser w/ bison to fix gcc@14 build

d-torrance:
The parser distributed in the tarball was generated using byacc and compiling w/ gcc 14 gives us "implicit declaration" errors for `yyparse`.

See https://github.com/robol/MPSolve/issues/48 and the discussion in https://github.com/spack/spack-packages/pull/5033 w/ @bernhardkaindl.


RikkiButler20 (approved):
LGTM

PR #6058, opened 2026-08-12
flux: Aug26 updates for core/sched/pmix

sam-maloney:
Adds releases from previous 2 months.

Also adds an environment variable for `flux-pmix` to add `pmix` to the list of PMI methods tried during bootstrap, with the default `simple` method left in first position (i.e. it will try `pmix` automatically but only if `simple` fails first). I'm not particularly familiar with Spack, so hopefully that is the correct way to set it :)
```
env.set("FLUX_PMI_CLIENT_METHODS", "simple pmix libpmi2 libpmi single")
```


grondo (approved):
Thanks! LGTM!

RikkiButler20 (approved):

PR #6059, opened 2026-08-12
py-gevent: add v26.8.0

adamjstewart:


adamjstewart:
@alalazo is this CI failure related to the simpler solve we perform in order to bootstrap clingo the first time to perform a proper solve? Do we need to hardcode something to use older cffi?

alalazo (approved):

alalazo:
Rebasing to check if this was due to a buggy version of Spack develop that was briefly used last week.

alalazo (dismissed):

adamjstewart:
Same CI failure as before

adamjstewart:
I think there was a bug in some of my `depends_on` versions, should be fixed now.

repos/spack_repo/builtin/packages/py_gevent/package.py
```diff
@@ -41,10 +43,16 @@ class PyGevent(PythonPackage):
...
 
     with default_args(type=("build", "run")):
+        depends_on("python@3.10:", when="@26.8:")
         depends_on("python@3.9:", when="@24.10.1:")
         depends_on("python@3.8:", when="@23.7.0:")
         depends_on("python@:3.10", when="@:21.12")
 
+        depends_on("py-setuptools@40.8:", when="@20.5.1:")
+        # https://github.com/gevent/gevent/commit/2a9cdcd19702154039aa0f074ee3cb11f63c239e
+        depends_on("py-setuptools@:80", when="@:26.4")
```
sethrj: @adamjstewart did you see https://github.com/spack/spack-packages/blob/2f927f10f40921c88fe7da695081fcfffd0e7cb4/repos/spack_repo/builtin/packages/py_gevent/package.py#L64-L65 ? I think "excluding versions for incompatibility" might be modeled better with `conflicts` in this case: the `depends_on` ended up causing surprise concretizations of older python packages that require newer setuptools.

sethrj (commented):
Discussing https://github.com/spack/spack-packages/pull/6091/changes#r3841659195

repos/spack_repo/builtin/packages/py_gevent/package.py
adamjstewart: Yes, but we only check for `depends_on` in our audit checks, so if someday older setuptools is removed, we wouldn't know to remove older gevent unless we use depends_on. In general, depends_on should be preferred unless there is something that can't easily be modeled without conflicts.

sethrj: Makes sense. Should we add an audit check for `conflicts`?

repos/spack_repo/builtin/packages/py_gevent/package.py
sethrj: You'll want to delete this now with the change above.

repos/spack_repo/builtin/packages/py_gevent/package.py
sethrj: Oh yeah, also should setuptools be *run* dependencies? I think just "build"… the requirement for 40.8 is now in the `build` section.

sethrj (changes requested):

repos/spack_repo/builtin/packages/py_gevent/package.py
adamjstewart: I'm not sure how to audit check that, @alalazo?

sethrj (approved):
py-gevent changes lgtm!

PR #6062, opened 2026-08-12
fflas-ffpack: improve blas/lapack detection

d-torrance:
We properly pass the correct BLAS/LAPACK build flags to the configure script, and also apply some patches from upstream to fix the build when using netlib-lapack.

This should fix the issues encountered by @bernhardkaindl in #5033.

### DIFF
--- a/repos/spack_repo/builtin/packages/fflas_ffpack/package.py
+++ b/repos/spack_repo/builtin/packages/fflas_ffpack/package.py
@@ -27,6 +27,46 @@ class FflasFfpack(AutotoolsPackage):
     depends_on("cxx", type="build")
     depends_on("pkgconfig", type="build")
 
-    depends_on("blas")
     depends_on("givaro")
     depends_on("lapack")
+
+    # require blas w/o threads
+    depends_on("blas")
+    for pkg in [
+        "acfl",
+        "amdblis",
+        "armpl-gcc",
+        "atlas",
+        "blis",
+        "essl",
+        "intel-oneapi-mkl",
+        "nvpl-blas",
+        "openblas",
+    ]:
+        depends_on(f"{pkg} threads=none", when=f"^[virtuals=blas] {pkg}")
+
+    # fix build w/ netlib-lapack
+    with when("@2.5.0"):
+        # https://github.com/linbox-team/fflas-ffpack/pull/384
+        patch(
+            "https://github.com/linbox-team/fflas-ffpack/commit/2cd3a04050f2fa447d8a9c8265f32f25f4cb6546.patch?full_index=1",
+            sha256="86186030f6d82f4ce2ae5acc5de23568bd057ef177d31ff6a717e7cc7697aa60",
+        )
+        # https://github.com/linbox-team/fflas-ffpack/pull/385
+        patch(
+            "https://github.com/linbox-team/fflas-ffpack/commit/97ee718ebc054d3dda020223a7d777786f7ad924.patch?full_index=1",
+            sha256="15b956fa5188f5c6f684c578004230f7f996144cc63429323d3b89e296d509e0",
+        )
+        # https://github.com/linbox-team/fflas-ffpack/pull/403
+        patch(
+            "https://github.com/linbox-team/fflas-ffpack/commit/9994f6aa5fc2e3677b14c3d12671ba46ce5a58b5.patch?full_index=1",
+            sha256="d3562239a4e04b5411e638ab5752afbe7649a38d97fd27f07928fb129a2648c6",
+        )
+
+    def configure_args(self):
+        blas_headers = self.spec["blas"].headers + self.spec["lapack"].headers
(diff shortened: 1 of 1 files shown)


bernhardkaindl (commented):
@d-torrance, ack, this fixes the build with `^blas=intel-oneapi-mkl`:
```py
$ spack install fflas-ffpack ^blas=intel-oneapi-mkl
$ spack find -v -d fflas-ffpack |grep -e ffla -e blas -e intel
fflas-ffpack@2.5.0 build_system=autotools patches:=15b956f,8618603,d356223
    intel-oneapi-mkl@2026.1.0~cluster+envmods~gfortran~ilp64+shared build_system=generic mpi_family=none threads=none
```
However, as I mentioned in the macaulay2 PR, fflas-ffpack's README.md says that one should make surce that a single-threaded BLAS is used.
I didn't test, but would this be safer to ensure no MPI or OpenMP version of BLAS gets used?
```py
    depends_on("blas threads=none")
```
Note: This is only a question or reminder, and not tested in any way,
so if you make a change, please test it (you're the expert, I only observed what they wrote)

d-torrance:
I don't think all the blas providers have a `threads` option.  And it seems like the ones that do default to `none` already, e.g.:

https://github.com/spack/spack-packages/blob/63c38504a144bf372ba429c2559e44f9456dbee0/repos/spack_repo/builtin/packages/openblas/package.py#L136-L142
https://github.com/spack/spack-packages/blob/63c38504a144bf372ba429c2559e44f9456dbee0/repos/spack_repo/builtin/packages/intel_oneapi_mkl/package.py#L200-L206

bernhardkaindl:
> I don't think all the blas providers have a threads option.
> And it seems like the ones that do default to none already, e.g.:

When for some reason, one of the blas providers is already built,
unless there is a constraint not to use it, the clingo solver will by default (unless `spack install --fresh` is used) attempt to use that `blas` provider:
- At this point, the default variant value of `threads=none` is no longer relevant because the package is already installed.
- And as the dependency does not specify a variant, the clingo solver will happily try to use whatever variant is installed. If both variants are installed, it is undefined which one will used for a given build:
```c++
spack find -v intel-oneapi-mkl
-- linux-ubuntu24.04-x86_64_v4 / %c=clang@18.1.3 ----------------
intel-oneapi-mkl@2026.1.0~cluster+envmods~gfortran~ilp64+shared
    build_system=generic mpi_family=none threads=openmp

-- linux-ubuntu24.04-x86_64_v4 / %c=rocmcc@7.2.3 ----------------
intel-oneapi-mkl@2026.1.0~cluster+envmods~gfortran~ilp64+shared
    build_system=generic mpi_family=none threads=none
==> 2 installed packages
```
It may take the last one (`threads=none`) but it not defined, it could just as well be the 1st one with `threads=openmp`.

> I don't think all the blas providers have a threads option.

Maybe, but I think they should get one if they don't have it to be compatible with this variant which `openblas` etc provide, even if they are only singlethreaded. If they are only multithreaded, then I think the missing variant would reject the package correctly.

As it is a big improvment, I don't want to hold PR back.
So I'll approve it as good progress, but I'll approve with this as review comment.

bernhardkaindl (dismissed):
As mentioned in https://github.com/spack/spack-packages/pull/6062#issuecomment-5282554196, I'm approving with this comment as review comment.

Summary: If only supports single-threaded blas implementations, I think a mechanism to declare that should be used to prevent builds which don't properly work or don't compile for this reason. As discussed, I it could be that not all BLAS recipes have a variant `threads=none`, which was my 1st idea (and maybe the variant does not work on the virtual blas depenency.

But, I think I've found a working example now where `repos/spack_repo/builtin/packages/strumpack/package.py` adds conditional `depends_on` for specific blas packages when they are selected as the blas provider:
```py
    depends_on("lapack")
    for blas in ("openblas", "amdblis", "blis"):
        ...
        depends_on(f"{blas} threads=none", when=f"~openmp ^[virtuals=blas] {blas}")
```
This adds a 3 depdencies which use `"^[virtuals=blas] {blas}"` in the when condition to only trigger this dependency when the blas virtual is the given `{blas}` (it defines one of those for each of the 3 blas providers).

While that's not as elegant as globally specifying `blas threads=none` which might not work, this possibility adresses a list of known recipes with a known `threads=` variant and requests `threads=none` from the recipe when the recipe is selected as the `"blas"` virtual.

I'd add `"intel-oneapi-mkl"` to this list of blas providers with a threads variant: Then would say that at least the kown cases if potentially getting the wrong variant would be adressed.

I'm adding an approval, but want to bring this option that is in use forward. Upate: I've found a solution that is actually in use in a spack recipe, and mention it in my review comment for approval.

d-torrance:
I just force-pushed a new version adding the `threads=none` variants to the blas dependency as suggested.

bernhardkaindl (approved):
Looks great!

PR #6066, opened 2026-08-12
Remove gtest as a dependency of STK for Trilinos16.*

cedelmaier:
Some downstream applications break when Trilinos +stk is always built with gtest, when gtest is pulled in as a different dependency in downstream applications. There was a brief moment where this was not the case, commit: 1a79dd1ae9ff26bd0dda5a6fdc10d95df46c6ff0. This makes it so that Trilinos +stk can be built without gtest.

### DIFF
--- a/repos/spack_repo/builtin/packages/trilinos/package.py
+++ b/repos/spack_repo/builtin/packages/trilinos/package.py
@@ -383,7 +383,7 @@ class Trilinos(CMakePackage, CudaPackage, ROCmPackage):
     with when("~gtest"):
         conflicts("+minitensor")
         conflicts("+rol")
-        conflicts("+stk")
+        conflicts("+stk", when="@17:")
     # this does alleviate needing: conflicts("+test", when="@17: ~gtest")
     # see https://github.com/spack/spack-packages/pull/3361 for explanation
 


cgcgcg:
@cedelmaier Trilinos used to have its own copy of gtest. We removed it for 17.0.0. Can you explain how the gtest dependency breaks other packages?

cedelmaier:
@cgcgcg We have a downstream package that compiles in Trilinos 16.1.0 with +stk, using the TriBITS build system under the hood (like Trilinos does). One issue with this is that we have our own gtest added as a dependency in that downstream project, and then because of how TriBITS orders the includes, we run into de-synchronization issues with gtest between our own dependency version and if Trilinos pulls in its own; the two gtests conflict with each other.

cgcgcg:
This sounds like the issue is that Trilinos is exposing the gtest package to its downstream dependencies. We should stop doing that, or at least make it optional.

cedelmaier:
Exactly! I think that would solve our problems. My PR should just make it optional (and then nobody gets it downstream), but isolating it so that none of the downstream consumers could see it would be ideal. That's how I discovered it in the first place. I also had a spack-packages version (briefly) that didn't have gtest turned on by default, which is why I only just discovered this as I was building new docker images for our testing pipeline.

cgcgcg:
@cedelmaier I have a hard time finding where this comes from. Can you post an error message or a reproducer?

cedelmaier:
Here are the environments that I am set up in (including spack, spack-packages, and trilinos specs/versions).
spack: 1.2.2 (3e19345b6e12f5ff1b874f4059622fc6a1fd804a)
spack-packages: releases/v2026.06
trilinos (note this trilinos cannot be compiled without +gtest when +stk)
==> In environment tril1610_cpu (1 root spec)
[+] trilinos@16.1.0+belos~boost+exodus+hdf5+kokkos+mpi+openmp+shared+stk+zoltan+zoltan2 cxxstd=17

-- linux-ubuntu24.04-broadwell / %c,cxx,fortran=gcc@13.3.0 ------
trilinos@16.1.0~adelus~adios2+amesos+amesos2+anasazi+aztec~basker+belos~boost~chaco~complex~cuda~cuda_constexpr~cuda_rdc~cusparse~debug~dtk+epetra+epetraext~epetraextbtf~epetraextexperimental~epetraextgraphreorderings+exodus+explicit_template_instantiation~float+fortran+gtest+hdf5~hypre+ifpack+ifpack2~intrepid~intrepid2~ipo~isorropia+kokkos~mesquite~minitensor+ml+mpi+muelu~mumps~nox+openmp~pamgen~panzer~phalanx~piro~python~rocm~rocm_rdc~rol+sacado~scorec+shards+shared~shylu+stk~stokhos~stratimikos~strumpack~suite-sparse~superlu-dist~teko~tempus~test~threadsafe~thyra+tpetra~trilinoscouplings~wrapper~x11+zoltan+zoltan2 build_system=cmake build_type=Release cxxstd=17 generator=make gotype=long_long patches:=fad9394

I get the error (copying the command as well)

[ 10%] Building CXX object mundy/tests/unit_tests/CMakeFiles/Mundy_UnitTestTest.dir/UnitTestTest.cpp.o
cd /home/testuser/MuNDy/build/mundy/tests/unit_tests && /usr/bin/ccache /home/testuser/spack/var/spack/environments/tril1610_cpu/.spack-env/view/bin/mpicxx -DKOKKOS_DEPENDENCE -I/home/testuser/MuNDy/build -I/home/testuser/MuNDy/mundy/core/utils/src -I/home/testuser/MuNDy/build/mundy/core/utils -I/home/testuser/MuNDy/mundy/core/math/src -I/home/testuser/MuNDy/build/mundy/core/math -I/home/testuser/MuNDy/mundy/core/geom/src -I/home/testuser/MuNDy/build/mundy/core/geom -I/home/testuser/MuNDy/mundy/core/mech/src -I/home/testuser/MuNDy/build/mundy/core/mech -I/home/testuser/MuNDy/mundy/mesh/src -I/home/testuser/MuNDy/build/mundy/mesh -I/home/testuser/MuNDy/mundy/search/src -I/home/testuser/MuNDy/build/mundy/search -isystem /home/testuser/spack/var/spack/environments/tril1610_cpu/.spack-env/view/include -isystem /home/testuser/spack/opt/spack/linux-broadwell/hwloc-2.13.0-n76pvvjqdoz2jxyzizvlbm5u5aefvjop/include -isystem /home/testuser/MuNDy/mundy_deps_cpu/include -isystem /home/testuser/spack/opt/spack/linux-broadwell/openblas-0.3.33-iaiqbwt5jsajvpdnzfxenpssstptcceg/include -isystem /home/testuser/spack/opt/spack/linux-broadwell/netcdf-c-4.10.0-ufiyavz3wa2k2cuxila7gnffpla5yj7h/include -isystem /home/testuser/spack/opt/spack/linux-broadwell/hdf5-1.14.6-edp4f3fpbxgw227t6mdxyh3fpi4flpzb/include -isystem /home/testuser/spack/opt/spack/linux-broadwell/parallel-netcdf-1.14.1-kwuobonlnoch3u2hijnz3op5wie6mhnf/include -isystem /home/testuser/spack/opt/spack/linux-broadwell/metis-5.1.0-44ozssid4jhfz4urdbmuk6bezsvebl6l/include -isystem /home/testuser/spack/opt/spack/linux-broadwell/parmetis-4.0.3-ya3qgk2vjwh746zbmy6xpdplhvaebgws/include -isystem /home/testuser/spack/opt/spack/linux-broadwell/zlib-ng-2.3.3-o5ht7spsrghlqd2dt5ptovinnmm4lyez/include -isystem /home/testuser/spack/opt/spack/linux-broadwell/cgns-4.5.0-c7jktkzl52zrnuqx7rzlieb23ri2uzre/include -isystem /home/testuser/MuNDy/mundy_deps_cpu/include/ArborX -isystem /home/testuser/MuNDy/mundy_deps_cpu/include/ArborX/details -isystem /home/testuser/MuNDy/mundy_deps_cpu/include/ArborX/geometry -isystem /home/testuser/MuNDy/mundy_deps_cpu/include/ArborX/kokkos_ext -pedantic -Wall -Wno-long-long -Wwrite-strings   -O3 -march=native -O3 -DNDEBUG -std=c++20 -fPIE -march=core-avx2 -mtune=core-avx2 -mrtm -fopenmp -MD -MT mundy/tests/unit_tests/CMakeFiles/Mundy_UnitTestTest.dir/UnitTestTest.cpp.o -MF CMakeFiles/Mundy_UnitTestTest.dir/UnitTestTest.cpp.o.d -o CMakeFiles/Mundy_UnitTestTest.dir/UnitTestTest.cpp.o -c /home/testuser/MuNDy/mundy/tests/unit_tests/UnitTestTest.cpp
In file included from /home/testuser/MuNDy/mundy_deps_cpu/include/gmock/internal/gmock-internal-utils.h:52,
                 from /home/testuser/MuNDy/mundy_deps_cpu/include/gmock/gmock-actions.h:146,
                 from /home/testuser/MuNDy/mundy_deps_cpu/include/gmock/gmock.h:56,
                 from /home/testuser/MuNDy/mundy/tests/unit_tests/UnitTestTest.cpp:24:
/home/testuser/spack/var/spack/environments/tril1610_cpu/.spack-env/view/include/gtest/gtest.h:1116:18: error: redefinition of 'class testing::internal::RE'
 1116 | class GTEST_API_ RE {
      |                  ^~
In file included from /home/testuser/MuNDy/mundy_deps_cpu/include/gmock/internal/gmock-port.h:58,
                 from /home/testuser/MuNDy/mundy_deps_cpu/include/gmock/internal/gmock-internal-utils.h:51:
/home/testuser/MuNDy/mundy_deps_cpu/include/gtest/internal/gtest-port.h:986:32: note: previous definition of 'class testing::internal::RE'
  986 | class GTEST_API_ [[nodiscard]] RE {
      |                                ^~
/home/testuser/spack/var/spack/environments/tril1610_cpu/.spack-env/view/include/gtest/gtest.h:1182:6: error: multiple definition of 'enum testing::internal::GTestLogSeverity'
 1182 | enum GTestLogSeverity {
      |      ^~~~~~~~~~~~~~~~
/home/testuser/MuNDy/mundy_deps_cpu/include/gtest/internal/gtest-port.h:1050:6: note: previous definition here

and then there are lots of these errors from the gtest being included in our tests, all redefinition errors it looks like.

cedelmaier:
@cgcgcg It looks like its pulling in gtest from the view include in the example I posted.

/home/testuser/spack/var/spack/environments/tril1610_cpu/.spack-env/view/include/gtest/gtest.h

which looks like its the result of TriBITS pulling in the whole view include:

-isystem /home/testuser/spack/var/spack/environments/tril1610_cpu/.spack-env/view/include

cgcgcg:
I'm still confused. To me, it looks like Trilinos is using the gtest dependency that it declared. Where are we doing something that breaks downstream apps?

cedelmaier:
@cgcgcg It's that Trilinos exposes its gtest to downstream applications, and I copied a snippet of what our application does (via TriBITS) to pull in Trilinos (Kokkos, Kokkos-kernels, and STK actually). So when we have our own gtest as a dependency, it sees both versions. I can set up a minimal example if that would help (and make sure that it still has the same problems).

cgcgcg:
@cedelmaier Yes, I think that would be helpful. I am trying to understand if the issue is in Trilinos cmake itself or in the Trilinos spack package.

cedelmaier:
@cgcgcg Well, it's an example, I'm not sure it's minimal, but at least it's a start.

https://github.com/cedelmaier/TrilinosGtestExample

I made sure that the cmake stage errors because we try to pull in google test ourself in the build script, and it gets exposed to the version that Trilinos already included. This is an easy version of a test of this than the full cmake+TriBITS version that I send you a snippet of before. At least this one just fails at the cmake stage, not the build stage. Let me know if you need anything else, or have any other questions, or what I can help with.

cgcgcg:
@cedelmaier Thank you very much. I can reproduce the issue.

cgcgcg:
@alanw0 FYI. It looks like the issue is that STKUnit_test_utils and STKNGP_TEST declare library dependencies on gtest.

alanw0:
@cgcgcg yes those stk sub-packages (STKUnit_test_utils and STKNGP_TEST) do have dependencies on gtest. Those two sub-packages are test-utility packages. We could make the gtest dependency optional, i.e., the files in those dirs that actually reference gtest could be excluded from the build. Maybe a better possibility is that we simply disable those packages when gtest is not found. (I would have thought we already do that, but maybe not...)

cgcgcg:
@alanw0 I changed the `gtest` dependency of `STKUnit_test_utils` and `STKNGP_TEST` from `LIB_REQUIRED_TPLS` to `TEST_REQUIRED_TPLS` and installed a Trilinos that only enabled `STK`. This seems to work with the reproducer provided by @cedelmaier.

alanw0:
> @alanw0 I changed the `gtest` dependency of `STKUnit_test_utils` and `STKNGP_TEST` from `LIB_REQUIRED_TPLS` to `TEST_REQUIRED_TPLS` and installed a Trilinos that only enabled `STK`. 

I'm not sure that will work for users (such as exawind) who want to use those stk libraries even if Trilinos tests are not enabled...

cgcgcg:
@alanw0 Yeah, this seems somewhat tricky to deal with.

bartlettroscoe:
So, just to be clear, is the problem that downstream CMake projects are pulling in all of the STK subpackages and libraries, and some of these STK libraries (rightly) are linked to the Trilinos TPL gtest and the downstream CMake project wants to use a different version of gtest?  Could could downstream CMake projects not pull in all of STK and just pick the STK subpackages that they need and include the one that has the gtest dependency?

NOTE: It is **NOT** recommended for downstream CMake projects to build all of Trilinos and then just pull all of it in with `find_package(Trilinos)` like is done here:

https://github.com/cedelmaier/TrilinosGtestExample/blob/00953f7d8da09d9c3c4d3d7672d9e5db01e8e0d6/raw_cmake/CMakeLists.txt#L78

Instead, you should select the specific Trilinos packages that you need.  And just exclude the STK subpackage that has this gtest dependency.

Could we make it so that the Spack build of Trilinos, by default, does **not** enable these STK subpackages that depend on `gtest`?  Otherwise, a TPL dependency like gtest that is exported through installed libraries is extremely problematic (which is why Trilinos remove the internal gtest).  IMHO, we should never have installed libraries that include a gtest dependency.

Another option is for downstream CMake projects to keep using `find_package(Triilnos)` to pull in everything but stop linking to all of the Trilinos libs or the entire `<ParentPackage>::all_libs` target and instead link to the individual `<SubPackage>::all_libs` target or even the individual library targets (just look at the installed `<Package>ConfigTargets.cmake` file).  The below example is linking to **all** Trilinos libraries:

https://github.com/cedelmaier/TrilinosGtestExample/blob/00953f7d8da09d9c3c4d3d7672d9e5db01e8e0d6/raw_cmake/CMakeLists.txt#L80

The other solution is to have downstream CMake projects have an option to use an external gtest.  For example:

- https://github.com/xsdk-project/xsdk-community-policies/blob/master/package_policies/M12.md

However, that does not seem fair to impose on downstream projects like that and that is not scalable.

So I see a few different options to fix this:

1. Don't have the Spack build of Trilinos enable these STK subpackages that depend on gtest for the libs.
2. Require downstream CMake project to pull in only the STK subpackages that they need with `find_package()` and exclude the ones that depend on gtest.
3. Downstream projects move to linking against `<SubPackage>::all_libs` as needed.
4. Update downstream CMake projects to have an option and allow the usage of a gtest provided externally.

Options, options, options ...

alanw0:
Thanks @bartlettroscoe , if someone configures trilinos without providing gtest, then the stk sub-packages that need gtest should automatically disable, unless those sub-packages have been explicitly enabled. (I think.) I will check to see why that is not happening, if indeed it is not happening.

bartlettroscoe:
> Thanks @bartlettroscoe , if someone configures trilinos without providing gtest, then the stk sub-packages that need gtest should automatically disable, unless those sub-packages have been explicitly enabled. (I think.) I will check to see why that is not happening, if indeed it is not happening.

@alanw0, what I am suggesting is that the Spack Trilinos package be set up so that just setting `trilinos+stk+gtest` does not, by default, enable the STK subpackages that depend on gtest.  The user should have to specify `trilinos+stk-test-gtest` or some other explicit enable to get those problematic STK subpackages that install libs that link to gtest.  Then Spack package and downstream customers don't get hit by this.  Then the Spack Sierra build can set `trilinos+stk-test-gtest`.  Or, the STK subpackages that depend on gtest could be built in a different Spack package

alanw0:
@bartlettroscoe sierra doesn't enable stk when building trilinos, since stk lives separately in sierra. But anyway...

However, the exawind folks at NREL (well both the project and the lab have different names now but I digress) would do something like `trilinos+stk-test+gtest` which should build/install these problematic subpackages. At the same time, other users should be able to do `trilinos+stk-test-gtest` as you say, and the problematic subpackages should *NOT* be built/installed, thus not impacting/hindering the user. (Does that agree with what you are saying?)

bartlettroscoe:
@alanw0 

> However, the exawind folks at NREL (well both the project and the lab have different names now but I digress) would do something like `trilinos+stk-test+gtest` which should build/install these problematic subpackages.

Yes, but perhaps  `trilinos+stk+stk-test+gtest`?

> At the same time, other users should be able to do `trilinos+stk-test-gtest` as you say, and the problematic subpackages should _NOT_ be built/installed, thus not impacting/hindering the user. (Does that agree with what you are saying?)

I think you mean  `trilinos+stk`, missing the `+stk-test-gtest`, correct?  I am also arguing that the default for `trilinos+stk` should be to **not** enable the STK subpackages that install libraries that depend on gtest.   Does that sound reasonable?  That way, users will not get hit by this by default and they will need to try harder to get an incompatible gtest lib included in their CMake project.

Question: If STK does not enable these STK subpackages that depend on gtest, does that mean that none of the other STK subpackages get any tests enabled?  I know Spack does not enable tests in Trilinos currently but in in a future world where there is more testing done with Spack packages, it could (in a test/integration build of  Spack).  There is a way to handle this situation but it might require a minor TriBITS tweak (i.e., allow you to assign entire subpackage as test-only in some configurations so that other package's tests can depend on them but they don't get installed and they don't get lumped into the `<Package>::all_libs` target).

alanw0:
@bartlettroscoe 

> Question: If STK does not enable these STK subpackages that depend on gtest, does that mean that none of the other STK subpackages get any tests enabled?

It is probably true that if those particular sub-packages are not enabled, then none of the stk tests get enabled. The general structure is that (for example) the tests for the `stk_mesh` subpackage are located in `stk_unit_tests/stk_mesh` (`stk_unit_tests` is a separate subpackage). The reason for this is that the tests have more dependencies. In this example, `stk_unit_tests/stk_mesh` can depend on `stk_io` and `stk_unit_test_utils`, and `gtest`, whereas `stk_mesh` can't depend on any of those.

Regarding the pseudo-spack-spec questions, when I say `trilinos+stk-test-gtest` I mean stk is enabled but trilinos tests are not enabled and gtest is not enabled. If that is the same as `trilinos+stk` due to defaults, then yes I agree with you.

bartlettroscoe:
@alanw0 

> Regarding the pseudo-spack-spec questions, when I say `trilinos+stk-test-gtest` I mean stk is enabled but trilinos tests are not enabled and gtest is not enabled. If that is the same as `trilinos+stk` due to defaults, then yes I agree with you.

Okay, there is the confusion, to remove an option in spack, you use `~<option>`, not `-<option>`.  See:

- https://spack.readthedocs.io/en/latest/spec_syntax.html

The char `-` is allowed to be part of a token name in Spack Spec.

So I think you mean `trilinos+stk~test~gtest` enable STK but disable tests and disable gtest?

cedelmaier:
@alanw0 @bartlettroscoe @cgcgcg As an aside, the library that we're compiling (not the mock example that I sent) pulls in Teuchos, STK, Tpetra, Belos (and some others), but not the entire Trilinos package. The original issue was that if I try to pull in STK in this manner, I still see the gtest that it was compiled with. My original question was just making it so that you could select if you wanted to install trilinos+stk~gtest or trilinos+stk+gtest, and make it so that you were not forced to have trilinos+stk+gtest always.

alanw0:
@bartlettroscoe 
> 
> > Regarding the pseudo-spack-spec questions, ... 
> Okay, there is the confusion, to remove an option in spack, you use `~<option>`, not `-<option>`.

ok now we're getting somewhere. I know how an actual spack-spec works, I used the term "pseudo-spack-spec" because I thought you were interchanging `-` and `~`.
So what did you mean when you originally said `trilinos+stk-test-gtest` ? I interpreted that as `trilinos+stk~test~gtest`.

To answer @cedelmaier , yes I think you should be able to do either `trilinos+stk~gtest` or `trilinos+stk+gtest`, and the former would automatically disable any stk subpackages that depend on gtest.

My belief is that that does work now, because I just configured trilinos without gtest, and I see the following output:
```
-- Setting Trilinos_ENABLE_STKNGP_TEST=OFF because STKNGP_TEST has a required library dependence on disabled package gtest
-- Setting Trilinos_ENABLE_STKUnit_test_utils=OFF because STKUnit_test_utils has a required library dependence on disabled package gtest
```

cedelmaier:
@alanw0 Interesting. I just tried it in one of my docker images, and ran into the same conflict that I was having originally. I found one version of spack and spack-packages that pinned it so that I could build it as `trilinos+stk~gtest`, but that doesn't work for me with the following (just pulled and made sure were up to date today):

spack: 1.3.0.dev0 (99940142424248ceaf45cc0e3e8126e88bc06165)
spack-packages (v2.2): 44ab024161cdeab9192145b529a4844e28b7dfd6

alanw0:
@cedelmaier 
> 
> spack: 1.3.0.dev0 (99940142424248ceaf45cc0e3e8126e88bc06165) spack-packages (v2.2): [44ab024](https://github.com/spack/spack-packages/commit/44ab024161cdeab9192145b529a4844e28b7dfd6)

ok, I'll look into this and see if we made some recent cmake changes that caused my version to work correctly. (I used trilinos/develop this morning, I don't know if it's easy for you to try that.)

cedelmaier:
@alanw0 I just tried with the most recent spack (downloaded a moment ago), and I still see an error when trying to configure trilinos@16.1+stk~gtest. I'm copying the spack version that it reports, along with the spack-packages repo hash.

spack version: 1.3.0.dev0 (5385604e4d0cac68f13bc072931b50fbbb0bb5f8)
spack-packages version (v2.2): 7318c6ac1a452a5e9f433d6de81841a036114e8f

```bash
spack spec trilinos@16.1+belos~boost+exodus~gtest+hdf5+kokkos+mpi+openmp+shared+stk+zoltan+zoltan2 cxxstd=17 %c=gcc@13 %cxx=gcc@13 %fortran=gcc@13 ^kokkos@4.5.01+openmp+serial+shared ^kokkos-kernels@4.5.01+blas+lapack+openmp+serial+shared 
==> Error: failed to concretize `trilinos@16.1+belos~boost+exodus~gtest+hdf5+kokkos+mpi+openmp+shared+stk+zoltan+zoltan2 cxxstd=17 %c,cxx,fortran=gcc@13 ^kokkos@4.5.01+openmp+serial+shared ^kokkos-kernels@4.5.01+blas+lapack+openmp+serial+shared` for the following reasons:
       1. trilinos: '+stk' conflicts with '~gtest'
  Analyzing the cause of the failure, this may take a moment...
==> Error:      2. trilinos: '+stk' conflicts with '~gtest'
        required because conflict is triggered when +stk 
          required because trilinos@16.1+belos~boost+exodus~gtest+hdf5+kokkos+mpi+openmp+shared+stk+zoltan+zoltan2 cxxstd=17 %c,cxx,fortran=gcc@13 ^kokkos@4.5.01+openmp+serial+shared ^kokkos-kernels@4.5.01+blas+lapack+openmp+serial+shared requested explicitly 
        required because conflict constraint ~gtest 
          required because trilinos@16.1+belos~boost+exodus~gtest+hdf5+kokkos+mpi+openmp+shared+stk+zoltan+zoltan2 cxxstd=17 %c,cxx,fortran=gcc@13 ^kokkos@4.5.01+openmp+serial+shared ^kokkos-kernels@4.5.01+blas+lapack+openmp+serial+shared requested explicitly 
```

alanw0:
@cedelmaier yeah, I didn't think newer spack would help. Rather, I was thinking newer trilinos (e.g. 17.1.2 or master) might help. But I'll try to check it this afternoon, and let you know what I find.

alanw0:
> @cedelmaier yeah, I didn't think newer spack would help. Rather, I was thinking newer trilinos (e.g. 17.1.2 or master) might help. But I'll try to check it this afternoon, and let you know what I find.

@cedelmaier sorry for the delay...
I confirmed that configuring trilinos with `trilinos+stk~gtest` does indeed have trouble, even with 17.1.1. So unfortunately this will require a small development task to fix.

Regarding the change you are proposing in this pull-request (adjusting the `conflicts` line): if that unblocks you, and @cgcgcg approves, then I approve.

cedelmaier:
@alanw0 @cgcgcg That does unblock me, but I can also stay pinned to a previous version of spack and spack-packages for now. My fix didn't seem to cause any problems with at least building our library, but I'm not sure how something like ExaWind would deal with it, or if they depend on having gtest exposed through Trilinos for some reason.

alanw0:
After doing a few spack builds, I would like to suggest adjusting your change to be this:
`+        conflicts("+stk", when="@:17.0.0")`
which means that the conflict between `~gtest` and `+stk` applies for trilinos versions up to 17.0.0. But there is no conflict for later versions. i.e., using the spec `trilinos@17.1.2+stk~gtest` appears to concretize and build successfully, as does `17.2.1`.

cedelmaier:
@alanw0 So `trilinos@16.1.0+stk~gtest` would still error, correct? That's what I was hoping to avoid until we can move to 17.X for our library.

alanw0:
> @alanw0 So `trilinos@16.1.0+stk~gtest` would still error, correct? That's what I was hoping to avoid until we can move to 17.X for our library.

Ah, ok. Yes this change would still leave versions 16.x.y giving conflicts. So my interpretation of your change, which is `conflicts("+stk", when="@17:")`, would remove the conflicts for earlier versions, but continue conflicting for versions after 17.
Perhaps a better change then is to simply remove the line, since my experiments show that it's safe to remove the conflict for later versions.

cedelmaier:
I'm totally fine with just removing the conflict altogether (although it's not up to me).

PR #6067, opened 2026-08-12
dorado: add 2.1.1 and remove conditional constraint

theAeon:
Tested against gcc 14, 2.1.1 seems to build.

### DIFF
--- a/repos/spack_repo/builtin/packages/dorado/package.py
+++ b/repos/spack_repo/builtin/packages/dorado/package.py
@@ -49,7 +49,7 @@ class Dorado(CMakePackage, CudaPackage):
     depends_on("zlib-api")
 
     conflicts("%gcc@:8", msg="Dorado requires at least gcc@9 to compile.")
-    conflicts("%gcc@13:", when="@:2",  msg="Dorado will not build with gcc@13 and newer.")
+    conflicts("%gcc@13:", when="@:2", msg="Dorado will not build with gcc@13 and newer.")
 
     patch("hdf5-libaec-0.5.3.patch", when="@:0.5.3")
     patch("hdf5-libaec-0.6.3.patch", when="@0.6:0.9")
--- a/repos/spack_repo/builtin/packages/dorado/package.py
+++ b/repos/spack_repo/builtin/packages/dorado/package.py
@@ -18,6 +18,7 @@ class Dorado(CMakePackage, CudaPackage):
 
     maintainers("snehring")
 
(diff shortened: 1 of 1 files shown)


theAeon:
@spackbot fix style

repos/spack_repo/builtin/packages/dorado/package.py
```diff
@@ -18,7 +18,7 @@ class Dorado(CMakePackage, CudaPackage):
 
     maintainers("snehring")
 
-    version("2.1.0", commit="48ab35d456dab9b08318c7413fd993ecee523a88", submodules=True)
+    version("2.1.1", commit="d66c17c657a55eaa2a6e3b5551cc83fd0a4a28b5", submodules=True)
```
bernhardkaindl, line 21: Please see the documentation for deprecating versions:
https://spack.readthedocs.io/en/latest/packaging_guide_creation.html#deprecate

Briefly, spack normally does not like to remove versions unless there is a very good reason. If a version shall be obsoleted simply because it is not preferred anymore or not used because of CVEs, bugs etc, it can be deprecated. See the documentaiton for the details.

It means you can't remove the currently latest version without good reason.
Here is what you should do:
1. Keep the current latest version (2.1.0).
   1.1 That is unless 2.1.0 shall no longer be used because of CVE or a severe bug.
   In that case, deprecate this version:
   with default_args(deprecated=true):
   ```py
       version("2.1.0", commit="48ab35d456dab9b08318c7413fd993ecee523a88", submodules=True)
   ```
   Also move other versions affect by such severe issues into this context.

2. Add the newer 2.1.1 on top of the versions.

If you think that the some versions can be deprecated as well because they'are no longer supported, superseded or so, add 
```
with default_args(deprecated=true):
   version ...
```
For example, if the 0.x.x version 1.x.x jump is a good barrier what to deprecate, and you don't expect users requiring those versions, you can move some or most of these older versions inside such context manager's context that gives them the deprecated=true argument.

bernhardkaindl (commented):
In short: You can't remove the current latest version (2.1.0) I'd say based on the review guidelines. If it has a severe issue, you can deprecate it and other affected versions. See the documentation link I give in the review for the official guidline to follow.

repos/spack_repo/builtin/packages/dorado/package.py
theAeon: Oh, that's my bad, I thought that I introduced that from a bunch of branches i was juggling. Will fix.

bernhardkaindl (approved):

PR #6069, opened 2026-08-12
ci: add nightly job to sync source mirror with pkgs in builtin

alecbcs:
Add a minimal nightly GitHub Actions CI/CD job that compares the state of the package repository to the mirrored source cache and mirrors any artifacts that are missing from the cache. (As of the time of posting there are 439 missing package versions that would be uploaded by this workflow.)

### DIFF
--- a/.github/workflows/bin/find-missing-mirror-specs.py
+++ b/.github/workflows/bin/find-missing-mirror-specs.py
@@ -21,6 +21,7 @@ from typing import Dict, List
 
 import spack.repo
 import spack.spec
+from spack.util import tty
 from spack.version import StandardVersion
 
 
@@ -62,6 +63,10 @@ def main(sha256_file: str) -> None:
     # limit to a maximum of 100 specs at a time due to GitHub
     # runner disk space limitations. Skipped specs will be
     # retried on the next scheduled job.
+    specs_to_output_num = len(specs_to_output)
+    if specs_to_output_num > 100:
+        tty.warn(f"Limiting to first 100 missing specs. Detected {specs_to_output_num} missing.")
+
     for spec in specs_to_output[:100]:
         print(spec)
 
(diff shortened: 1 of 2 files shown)


.github/workflows/bin/find-missing-mirror-specs.py
```diff
@@ -0,0 +1,67 @@
...
+            specs_to_output.append(version_spec)
+
+    # filter out non-redistributable packages
+    specs_to_output = [
+        spec for spec in specs_to_output if repo.get_pkg_class(spec.name).redistribute_source(spec)
+    ]
+
+    # output specs one per line for use by `spack mirror create`
+    for spec in specs_to_output:
+        print(spec)
```
scheibelp, line 71: Request: limit this to something like 100-500 and print a warning if we go over that (tty.warn? something that prints to stderr)

scheibelp (changes requested):
Other points (not requests for this PR):

- (request for later) If a patch/resource is dropped, this doesn't catch that: if we already placed the "main" version tarball in a mirror, and a resource was missed, then this job will also miss that (I see though that patches/resources generally *are* included, just not in this rare case). I think this can be addressed later.
- (not a request) What about wrong checksums (i.e. if someone specifies the wrong checksum for a version)? I think the behavior would be OK: this script will collect every wrong checksum, download the version (which should fail, but `spack mirror` generally proceeds on in this case). We'd want to clean those up though.

.github/workflows/bin/find-missing-mirror-specs.py
alecbcs: Good idea! Added in the last two commits.

tgamblin:
@scheibelp are your review comments addressed?

scheibelp (approved):

tgamblin (approved):

PR #6071, opened 2026-08-13
Fix missing depends_on("c") in 21 packages without maintainer

bernhardkaindl:
Hi, while reviewing and building packages for reviews, I saw a few cases of missing `depends_on("c")`,
where `depends_on("cxx")` was generated, but `depends_on("c")` not. But, "c" was in fact also needed.

To fix most similar packages in one or a few PRs, I grepped the potentially affected packages (with the generated "cxx" line, and without "c") and started building all I could build.

- [x] This PR is restricted to only packages without maintainers:
  - With 21 packages, this PR has a lot of files on its own.

Checked:
- [x] The last recipe in this PR has a manual change to add the missing pkgconfig dependency too.


ax3l (approved):

wdconinc (approved):
LGTM, and sound methodology.

PR #6073, opened 2026-08-13
Fix missing depends_on("c") in 13 packages with maintainers

bernhardkaindl:
Hi, while reviewing and building packages for reviews, I saw a few cases of missing `depends_on("c")`, where `depends_on("cxx")` was generated and `depends_on("c")` was not generated. In some cases, "c" was in fact also needed.

This is the 2nd PR in this series (the 1st PR is for packages without maintainers):
- #6071
- This PR is for packages with maintainers

Dear maintainers: You'd only have to look at your package(s) in this PR. Even if your package doesn't use C, often CMake or configure scripts check for a C compiler (unless the check for "C" is removed) which causes spack to fail if `depends_on("c")` isn't added.

Approach:

To fix most similar packages in one or a few PRs, I grepped the potentially affected packages (with the generated "cxx" line, and without "c") and started building all I could build.

- [x] This PR is restricted to only packages with maintainers:
  - The mainainer-less packages are in #6071.

Checked:
- [x] The xeus recipe also adds pkgconfig and openssl to fix the build on a system without them preinstalled on the host.

### DIFF
--- a/repos/spack_repo/builtin/packages/amqp_cpp/package.py
+++ b/repos/spack_repo/builtin/packages/amqp_cpp/package.py
@@ -32,7 +32,8 @@ class AmqpCpp(CMakePackage):
 
     conflicts("+tcp", when="platform=darwin", msg="TCP module requires Linux")
 
-    depends_on("cxx", type="build")  # generated
+    depends_on("c", type="build")
+    depends_on("cxx", type="build")
 
     depends_on("cmake@3.5:", type="build")
     depends_on("openssl@1.1.1:", when="+tcp", type=("build", "link", "run"))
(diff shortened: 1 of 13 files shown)


ax3l (approved):
LGTM for Isaac.
Likely just a CMake default to enable C and CXX by default together, even for C++ projects.

bonachea (approved):
LGTM for sympack

bernhardkaindl:
Thanks for the approval and the comment so far!
Because this PR only adds missing build dependencies and doesn't change your packages in a functional way, I think it can be merged.

Thanks for the reviews, this PR just means that with the new spack core which requires depends_on("c") for the C compiler, your packages will still build.

PR #6076, opened 2026-08-13
Fix missing depends_on("c") and similar dependencies in 10 recipes

bernhardkaindl:
Hi, while reviewing and building packages for reviews, I saw a few cases of missing `depends_on("c")`, where `depends_on("cxx")` was generated and `depends_on("c")` was not generated. In some cases, "c" was in fact also needed.

This is the 3rd PR in this series, the previous PRs are:
- #6071
- #6073

Dear maintainers: You'd only have to look at your package(s) in this PR. Even if your package doesn't use C, often CMake or configure scripts check for a C compiler (unless the check for "C" is removed) which causes spack to fail if `depends_on("c")` isn't added.

Approach taken:

To fix most similar packages in one or a few PRs, I grepped the potentially affected packages (with the generated "cxx" line, and without "c") and started building all packages I could build.

Checked:
- [x] Two recipes needed further manual fixes, but are limited to build dependences and versions (e.g. since boost-89, the component boost_system isn't in the compiled library anymore even though it still exists in the headers-only part, as far as I could understand)

### DIFF
--- a/repos/spack_repo/builtin/packages/brigand/package.py
+++ b/repos/spack_repo/builtin/packages/brigand/package.py
@@ -25,7 +25,8 @@ class Brigand(CMakePackage):
     version("1.1.0", sha256="afdcc6909ebff6994269d3039c31698c2b511a70280072f73382b26855221f64")
     version("1.0.0", sha256="8daf7686ff39792f851ef1977323808b80aab31c5f38ef0ba4e6a8faae491f8d")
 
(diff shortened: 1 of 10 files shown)


xylar (approved):
tempestremap change makes sense.

vijaysm (approved):
Looks good to me as well.

michaelkuhn (approved):
LGTM

PR #6083, opened 2026-08-13
Augustus: include stdint.h using the -include flag (fixes gcc-14 to 16)

bernhardkaindl:
Fix the build of augustus with gcc-14 to 16:

- include stdint.h using the -include flag

This is a smaller version of this PR to not require a patch for the fix:
- #5956 (Cc: @mikerenfro)

Alternatively, #5956 could be merged, but I wanted to fix to be tiny.

Cc: @emwjacobson (as you just made the previous fixes):
FYI, in case you build with gcc@14 or newer, this fixes it!


sethrj (changes requested):
This should be addressed with an upstream pull request and fixed via a patch directive, e.g.:
```c++
        patch(
            "https://github.com/Gaius-Augustus/Augustus/pull/NNNNNN.diff?full_index=1",
            sha256="",
            when="@:VVVV",
        )
```

PR #6087, opened 2026-08-13
kokkos: filter compiler wrappers also in lib

rbberger:
On some systems the CMake files are installed in lib instead of lib64. This adds the missing filter logic.

### DIFF
--- a/repos/spack_repo/builtin/packages/kokkos/package.py
+++ b/repos/spack_repo/builtin/packages/kokkos/package.py
@@ -403,7 +403,7 @@ class Kokkos(CMakePackage, CudaPackage, ROCmPackage):
         filter_compiler_wrappers(
             "KokkosConfigCommon.cmake",
             relative_root=os.path.join(libdir, "cmake", "Kokkos"),
-        ) 
+        )
 
     # sanity check
     sanity_check_is_file = [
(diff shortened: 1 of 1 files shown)


lucbv (approved):

repos/spack_repo/builtin/packages/kokkos/package.py
```diff
@@ -399,6 +399,9 @@ class Kokkos(CMakePackage, CudaPackage, ROCmPackage):
     # Filter spack-generated files that may include links to the
     # spack compiler wrappers
     filter_compiler_wrappers("kokkos_launch_compiler", relative_root="bin")
+    filter_compiler_wrappers(
+        "KokkosConfigCommon.cmake", relative_root=os.path.join("lib", "cmake", "Kokkos")
+    )
     filter_compiler_wrappers(
         "KokkosConfigCommon.cmake", relative_root=os.path.join("lib64", "cmake", "Kokkos")
     )
```
masterleinad:
```suggestion
    for libdir in ("lib", "lib64"):
        filter_compiler_wrappers(
            "KokkosConfigCommon.cmake",
            relative_root=os.path.join(libdir, "cmake", "Kokkos"),
        ) 
```

rbberger:
@spackbot fix style

masterleinad (approved):

tpadioleau (approved):
Ok for me but I would like to know if this filtering can have an impact on packages built with Spack and relying on the Kokkos launcher ?

rbberger:
@tpadioleau yes. Within Spack this can lead to missing flags that are injected by the Spack compiler wrapper (like rpaths). The only fix I can think of long term is making installed KokkosCommonConfig.cmake aware of operating within or outside the Spack build environment.

RikkiButler20 (approved):

PR #6088, opened 2026-08-13
add depends_on(c) directives to mergiraf, typos, and difftastic

cmelone:
without a C dependency, components of these packages may fail to build if the default system compiler is selected (like Cray CCE). adding C as a dep allows for more control over the compiler used and injected build flags


alecbcs (approved):
Looks good to me. Thanks @cmelone

PR #6091, opened 2026-08-14
pyzmq: remove dependency on old setuptools

sethrj:
A transient dependency through py-gevent was limiting installs to setuptools@0.80. The dependency of zmq on gevent is *only* needed for a runtime `pyzmq.green` module access, which ipykernel doesn't use. Only `circus` seems to use `pyzmq.green`, and no one on github uses `pyzmq.ssh`.

### DIFF
--- a/repos/spack_repo/builtin/packages/py_gevent/package.py
+++ b/repos/spack_repo/builtin/packages/py_gevent/package.py
@@ -44,10 +44,6 @@ class PyGevent(PythonPackage):
         depends_on("python@3.8:", when="@23.7.0:")
         depends_on("python@:3.10", when="@:21.12")
 
-        depends_on("py-setuptools@40.8:", when="@20.5.1:")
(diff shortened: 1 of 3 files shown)


alalazo (approved):

alalazo:
Giving some time to @manuelakuhn for review

repos/spack_repo/builtin/packages/py_gevent/package.py
```diff
@@ -44,10 +45,6 @@ class PyGevent(PythonPackage):
         depends_on("python@3.8:", when="@23.7.0:")
         depends_on("python@:3.10", when="@:21.12")
 
-        depends_on("py-setuptools@40.8:", when="@20.5.1:")
-        # https://github.com/pypa/distutils/pull/335
-        depends_on("py-setuptools@:80")
```
adamjstewart, line 49: Not 100% convinced this is the right approach. Let's discuss in #6059

PR #6093, opened 2026-08-14
shapelib: add tests variant

mathomp4:
In doing a build test on an air-gapped node where I'd pre-fetched my sources, I found I couldn't build `shapelib`. By default, their CMake seems to enable testing as long as `bash` can be found (see https://github.com/OSGeo/shapelib/blob/6e3ac42088274d3a68c0c35aa42e0ddf7f80104e/CMakeLists.txt#L207-L213).

This then triggers a `FetchContent` call to download Google benchmark and GoogleTest which weren't prefetched and crash.

So this PR adds a `tests` variant which defaults to False.

### DIFF
--- a/repos/spack_repo/builtin/packages/shapelib/package.py
+++ b/repos/spack_repo/builtin/packages/shapelib/package.py
@@ -24,3 +24,12 @@ class Shapelib(CMakePackage):
 
     depends_on("c", type="build")  # generated
     depends_on("cxx", type="build")  # generated
+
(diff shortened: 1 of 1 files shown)


mathomp4:
Hmm. No maintainer. I'm going to mention @rouault to see if he has objections. 

I also request review from @haampie since, well, GitHub recommended it. 😄

RikkiButler20 (approved):
LGTM

PR #6095, opened 2026-08-14
ci: update SPACK_EXTRA_MIRROR to use branch ref instead of stack name

zackgalbreath:
The University of Oregon is maintaining a local copy of our top-level develop buildcache to speed up download times and reduce data egress charges.

Change the way this environment variable is used by removing the stack name and replacing it with the base ref (currently `develop`). This fixes a layout problem and allows us to maintain separate develop vs. release buildcache copies at UO if the need arises.


zackgalbreath:
evidence that this change results in the updated mirror URL that we expect:

```
local: https://uo-spack-mirror.e4s.io:9002/spack-binaries/develop
```

https://gitlab.spack.io/spack/spack-packages/-/jobs/23940247#L722

kwryankrattiger (approved):

PR #6100, opened 2026-08-14
mpsolve: ensure that we have bison 3

d-torrance:
This is a quick followup to #6056, which fixed a build issue in mpsolve.  The parser shipped with the tarball is broken with newer versions of GCC, so we now regenerate it with bison first.

*However*, we need to ensure that we have a new enough bison, as otherwise, we'll just regenerate the same broken parser we had in the first place!


RikkiButler20 (approved):
LGTM

PR #6102, opened 2026-08-14
Patch to set TAUROOT to prefix, not build, dir

wspear:


eugeneswalker (approved):

eugeneswalker:
Should we merge now or wait for further review?

wspear:
I think this should be safe. It's already in TAU develop, and it's only applied to one release.

Issue #3322, opened 2026-02-09
Installation issue: flint


jagot:
### Steps to reproduce the issue

# Version 1

```console
$ spack spec -I flint
 -   o4yjsqg  flint@3.4.0 build_system=autotools platform=darwin os=tahoe target=m3 %c,cxx=apple-clang@17.0.0
[e]  qftzyfe      ^apple-clang@17.0.0 build_system=bundle platform=darwin os=tahoe target=aarch64
[+]  dwzpelv      ^compiler-wrapper@1.0 build_system=generic platform=darwin os=tahoe target=m3
[e]  y64bfyx      ^gmake@3.81~guile build_system=generic platform=darwin os=tahoe target=aarch64
[+]  xszt6ur      ^gmp@6.3.0+cxx build_system=autotools libs:=shared,static platform=darwin os=tahoe target=m3 %c,cxx=apple-clang@17.0.0
[e]  amrnftu          ^autoconf@2.72 build_system=autotools platform=darwin os=tahoe target=aarch64
[e]  etrteq5          ^automake@1.18.1 build_system=autotools platform=darwin os=tahoe target=aarch64
[+]  52m24cx          ^compiler-wrapper@1.0 build_system=generic platform=darwin os=tahoe target=m3
[e]  73c3hes          ^gmake@3.81~guile build_system=generic platform=darwin os=tahoe target=aarch64
[e]  ljtjo2o          ^libtool@2.5.4 build_system=autotools platform=darwin os=tahoe target=aarch64
[+]  smz2tkk          ^m4@1.4.20+sigsegv build_system=autotools platform=darwin os=tahoe target=m3 %c,cxx=apple-clang@17.0.0
[-]  rulmyap              ^compiler-wrapper@1.0 build_system=generic platform=darwin os=tahoe target=m3
[-]  ec36fte              ^diffutils@3.10 build_system=autotools platform=darwin os=tahoe target=m3 %c=apple-clang@17.0.0
[+]  elenvao                  ^libiconv@1.18 build_system=autotools libs:=shared,static platform=darwin os=tahoe target=m3 %c=apple-clang@17.0.0
[+]  3d5zhqf              ^libsigsegv@2.14 build_system=autotools platform=darwin os=tahoe target=m3 %c=apple-clang@17.0.0
[+]  pdklxby      ^gnuconfig@2024-07-27 build_system=generic platform=darwin os=tahoe target=m3
[...]

PR #6103, opened 2026-08-15
flint: fix git + macos builds

d-torrance:
The "main" build from git was failing because we didn't have the necessary autotools dependencies to generate the configure script.

Also add a fix for builds on macOS, which were failing because of the `/var` -> `/private/var` symlink confusing the build system into thinking it was an out-of-tree build when it was really an in-tree build (see https://github.com/flintlib/flint/issues/2799 and https://github.com/flintlib/flint/pull/2800).

### DIFF
--- a/repos/spack_repo/builtin/packages/flint/package.py
+++ b/repos/spack_repo/builtin/packages/flint/package.py
@@ -41,6 +41,23 @@ class Flint(AutotoolsPackage):
 
     depends_on("m4", type="build")
 
+    # generate configure script when building from git
+    with when("@main"):
+        depends_on("autoconf", type="build")
+        depends_on("automake", type="build")
+        depends_on("libtool", type="build")
+
+    # https://github.com/flintlib/flint/pull/2800
+    @when("@3.4.0:")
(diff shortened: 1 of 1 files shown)


d-torrance:
Just noticed that the macOS build was an existing issue:

Closes: #3322

repos/spack_repo/builtin/packages/flint/package.py
```diff
@@ -41,6 +41,23 @@ class Flint(AutotoolsPackage):
...
 
+    # generate configure script when building from git
+    with when("@main"):
+        depends_on("autoconf", type="build")
+        depends_on("automake", type="build")
+        depends_on("libtool", type="build")
+
+    # https://github.com/flintlib/flint/pull/2800
+    @when("@3.4.0:")
+    @run_before("configure")
```
alalazo:
```suggestion
    @run_before("configure", when="@3.4.0")
```

alalazo (changes requested):

alalazo (approved):

PR #5980, opened 2026-08-07
gcc: add v16.2.0

rbberger:

PR #6112, opened 2026-08-17
gcc: disable debug info checking for `gcc@16 +bootstrap` according to upstream bug report

yizeyi18:
Both in https://gcc.gnu.org/bugzilla/show_bug.cgi?id=125598 and #5980 people say their gcc 16.2 fails in bootstrapping with non-default build config; the GCC side report claims that it may be caused by debug section only, so this PR was made.

This PR disables debug section comparison during bootstrapping GCC 16 with gcc's default build config. 

EDIT: test on my local machine shows that with this patch on `spack.mk`, `+graphite +bootstrap` also passes so a `when` was appended to the conflict.

### DIFF
--- a/repos/spack_repo/builtin/packages/gcc/package.py
+++ b/repos/spack_repo/builtin/packages/gcc/package.py
@@ -877,6 +877,11 @@ class Gcc(AutotoolsPackage, GNUMirrorPackage, CompilerPackage):
             # Improve the build time for stage 2 a bit by enabling -O1 in stage 1.
             # Note: this is ignored under ~bootstrap.
             f.write("STAGE1_CFLAGS += -O1\n")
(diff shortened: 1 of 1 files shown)


michaelkuhn (approved):
Works for me (Fedora 44, GCC 16.1)

yizeyi18:
It seems the ci pipeline is affected by github's error.
@spackbot re-run pipeline

bernhardkaindl (approved):
```py
spack find -v gcc@16.2
-- linux-ubuntu24.04-x86_64_v4 / %c,cxx=gcc@13.3.0 --------------
gcc@16.2.0+binutils+bootstrap+futex+graphite+libsanitizer~mold~nvptx~piclibs~profiled~strip build_system=autotools build_type=RelWithDebInfo languages:='c,c++,fortran'
```

PR #6118, opened 2026-08-17
CMake: add define path method

johnwparent:
CMake should be handed posix style paths. It is more robust, safer, and recommended by the CMake developers.

Rather than relying on CMakePackages to ad-hoc cast their paths to posix, provide an interface around `define` that implicitly casts paths to posix before defining them as a CMake arg.

The method expects to be handed a pathlike object

Intended as a QOL improvement for CMake packages supporting Windows.

### DIFF
--- a/repos/spack_repo/builtin/build_systems/cmake.py
+++ b/repos/spack_repo/builtin/build_systems/cmake.py
@@ -7,7 +7,7 @@ import pathlib
 import platform
 import re
 import sys
-from typing import Any, List, Optional, Tuple
+from typing import Any, List, Optional, Tuple, Union
 
 from spack.package import (
     BuilderWithDefaults,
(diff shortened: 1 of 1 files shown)


johnwparent:
cc @adamjstewart since you had expressed interest.

adamjstewart (approved):
Would be worth adding links to the docs that recommend POSIX paths even on Windows, otherwise this looks good to me. Curious what other maintainers think.

johnwparent:
good call on the pointer to CMake docs, thanks for the feedback! Also very curious as to what others think.

My end goal w.r.t. paths and Spack is we ingest platform style paths, convert to pathlib, represent as posix where possible, and then convert to Windows backslashes at boundaries where the paths leave spack to be consumed by other things, where appropriate. That is to say, ideally, this method may become facile, but I could also see value in being explicit about defining paths and needing other path specific conventions for CMake defines. 

In the meantime, this is a valuable add IMO

PR #6121, opened 2026-08-17
legion: disable parallel build for `+rocm`

scheibelp:
Extract legion changes for rocm build from #4411

I noticed a problem with an intermittent build error (some sort of race condition) in CI when building `legion+rocm` in #4411

### DIFF
--- a/repos/spack_repo/builtin/packages/legion/package.py
+++ b/repos/spack_repo/builtin/packages/legion/package.py
@@ -366,7 +366,7 @@ class Legion(CMakePackage, CudaPackage, ROCmPackage):
 
     @property
     def parallel(self):
-        return not self.spec.satisfies("+rocm")
+        return not self.spec.satisfies("@:26.06 +rocm")
 
     def flag_handler(self, name, flags):
         if name == "cxxflags":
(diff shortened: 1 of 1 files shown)


rbberger:
This should be revisited with the upcoming Legion/Realm September releases.

elliottslaughter:
Legion master has already merged the fix. Can we version bound this to 26.06 and earlier?

scheibelp:
> Legion master has already merged the fix. Can we version bound this to 26.06 and earlier?

I've updated the constraint so that it only disables parallel builds for `@:26.06 +rocm`

bernhardkaindl (dismissed):

bernhardkaindl (commented):
PS: I see legion is used in CI, but I don't see that +rocm is enabled, so I think it should not change CI duration:
```py
git grep '\<legion\>' .ci/
.ci/gitlab/configs/linux/ci.yaml:      - legion
grep rocm .ci/gitlab/configs/linux/ci.yaml
```

bernhardkaindl (approved):
@alalazo I un-pushed my accidential push, but now it needs another approver to merge.

rbberger (approved):

rbberger:
@spackbot run pipeline

rbberger (approved):

PR #6124, opened 2026-08-18
py-atb: new package

Growl1234:
Add a Spack package recipe for [pyATB](https://github.com/pyatb/pyatb), a Python package for ab-initio tight-binding calculations.

The recipe builds pyATB from source with Spack-provided dependencies, including Eigen, OpenBLAS, and mpi4py, and patches the upstream build configuration to use the external Eigen installation and OpenBLAS-provided LAPACKE interface.

### DIFF
--- a/repos/spack_repo/builtin/packages/py_atb/package.py
+++ b/repos/spack_repo/builtin/packages/py_atb/package.py
@@ -1,0 +1,76 @@
+# Copyright Spack Project Developers. See COPYRIGHT file for details.
+#
+# SPDX-License-Identifier: (Apache-2.0 OR MIT)
+
+from spack_repo.builtin.build_systems.python import PythonPackage
+
+from spack.package import *
+
+
+class PyAtb(PythonPackage):
+    """Python ab initio tight-binding simulation package."""
+
+    homepage = "https://pyatb.github.io/pyatb/"
+    git = "https://github.com/pyatb/pyatb.git"
+
+    maintainers("Growl1234")
+
+    license("GPL-3.0-only", checked_by="RMeli")
+
+    version("main", branch="main")
+    version("1.1.2", commit="1b35d9f1f5ee1e6aa8b24c5ccf3c92bd19d32e36")
+
+    depends_on("cxx", type="build")
+
+    with default_args(type="build"):
+        depends_on("py-setuptools@42:")
+        depends_on("py-pybind11")
+        depends_on("eigen")
+
+    # pyATB directly uses the BLAS/LAPACK/LAPACKE interfaces provided by
+    # OpenBLAS. Keep the ABI at the LP64, unsuffixed interface expected by
+    # the current pyATB sources.
+    depends_on(
+        "openblas~ilp64 symbol_suffix=none",
+        type=("build", "link"),
+    )
+
+    with default_args(type=("build", "run")):
+        depends_on("py-numpy@1.17:")
+        depends_on("py-scipy@1.5:")
+        depends_on("py-mpi4py@3.1:")
+        depends_on("py-matplotlib@2.2.2:")
+        depends_on("py-ase")
+
+    def patch(self):
+        # Use Spack's Eigen instead of the bundled git submodule.
+        filter_file(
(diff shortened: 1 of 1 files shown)


Growl1234:
@RMeli Would you like to take a look at this one?

repos/spack_repo/builtin/packages/py_atb/package.py
```diff
@@ -0,0 +1,71 @@
...
+
+class PyAtb(PythonPackage):
+    """Python ab initio tight-binding simulation package."""
+
+    homepage = "https://pyatb.github.io/pyatb/"
+    git = "https://github.com/pyatb/pyatb.git"
+
+    maintainers("Growl1234")
+
+    license("GPL-3.0-only")
```
RMeli:
```suggestion
    license("GPL-3.0-only", checked_by="RMeli")
```

repos/spack_repo/builtin/packages/py_atb/package.py
```diff
@@ -0,0 +1,71 @@
...
+
+    version("main", branch="main")
+    version("1.1.2", commit="1b35d9f1f5ee1e6aa8b24c5ccf3c92bd19d32e36")
```
RMeli, line 21: For a Python package, I think the preferred way is to download from PyPI.

https://spack.readthedocs.io/en/latest/build_systems/pythonpackage.html#pypi-vs-github

repos/spack_repo/builtin/packages/py_atb/package.py
```diff
@@ -0,0 +1,71 @@
...
+
+    depends_on("cxx", type="build")
+
+    with default_args(type="build"):
+        depends_on("py-setuptools@42:")
+        depends_on("py-wheel")
```
RMeli:
```suggestion
```
It's already a build dependency in `PythonPackage`, so there is no need to add it here (no particular version constraint is required).

https://spack.readthedocs.io/en/latest/build_systems/pythonpackage.html#dependencies

repos/spack_repo/builtin/packages/py_atb/package.py
```diff
@@ -0,0 +1,71 @@
...
+            "libraries = ['openblas']",
+            "setup.py",
+            string=True,
+        )
+
+        # Do not hard-code GCC's libgomp. Let the compiler driver select the
+        # corresponding OpenMP runtime at link time.
+        filter_file(
+            "extra_link_args = ['-lgomp']",
+            "extra_link_args = ['-fopenmp']",
```
RMeli: Only works with GCC and Clang. I'd add `conflicts("%intel-oneapi-compilers")` and `conflicts("%intel-oneapi-compilers-classic")` etc. above.

RMeli (changes requested):

repos/spack_repo/builtin/packages/py_atb/package.py
Growl1234: PyPI currently only provides binary wheels for pyATB and does not provide an sdist, so I used the GitHub source instead, following the exception mentioned in the linked documentation.

repos/spack_repo/builtin/packages/py_atb/package.py
Growl1234: Would it be preferable to replace both hard-coded OpenMP flags with `self.compiler.openmp_flag`? That said I'm trying to reach upstream to confirm whether Intel compilers can work.

RMeli: Yes, that would be a nicer solution. If Intel compilers do not work the conflicts are still needed, but they could be eventually removed down the line and `self.compiler.openmp_flag` will work.

Growl1234:
Looks like two CI jobs stucked possibly due to GitHub Action bug, as I encountered such case in other repos as well? Is it possible to re-run them?

RMeli:
GitHub has indeed been a bit flaky yesterday. I restarted the jobs. :crossed_fingers:

RMeli (approved):

PR #6125, opened 2026-08-18
swig: add v4.5.0

RiMillo:
Maintainers: @sethrj


sethrj (approved):
Checksum's good, swig doesn't have dependencies. Thanks!

PR #6129, opened 2026-08-18
flint: apply macos bug fix in all versions since 3.4.0

d-torrance:
This is a quick followup to #6103.

I just noticed that @alalazo's suggestion in https://github.com/spack/spack-packages/pull/6103#discussion_r3803939351 switched `@3.4.0:` -> `@3.4.0`.  The bug that this part of the formula is fixing is present in all flint versions since 3.4.0, so we need that trailing colon.

I should have looked more carefully before I clicked the "Apply suggestion" button!  :)


alalazo:
Sorry for the typo :slightly_smiling_face:

alalazo (approved):

PR #6134, opened 2026-08-18
mapl: add v2.71.0

mathomp4:
This adds MAPL v2.71.0

I also made a fix suggested by `ruff` based on:

UP031: Use format specifiers instead of percent format

### DIFF
--- a/repos/spack_repo/builtin/packages/mapl/package.py
+++ b/repos/spack_repo/builtin/packages/mapl/package.py
@@ -41,10 +41,11 @@ class Mapl(CMakePackage):
 
     # Remember if there is a new ESMA_cmake, to update the resources too
     version(
-        "2.70.0",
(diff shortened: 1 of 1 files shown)


climbfuji (approved):

mathomp4:
I wonder if I should trim the `maintainers` here. I mean, I'm pretty much the maintainer now. No need to bother everyone else. Thoughts, @climbfuji ?

climbfuji:
I'm ok to stay on, even if it's just to be aware of new versions. Not sure if @edwardhartnett or @Hang-Lei-NOAA are [still] interested.

mathomp4:
> I'm ok to stay on, even if it's just to be aware of new versions. Not sure if @edwardhartnett or @Hang-Lei-NOAA are [still] interested.

I removed both. I mean, for now 4 is enough.

climbfuji (approved):

RikkiButler20 (approved):

PR #6143, opened 2026-08-19
kahip: build with NONATIVEOPTIMIZATIONS to respect Spack's target flags

jhale:
Upstream forces `-march=native` via `add_definitions()` unless `NONATIVEOPTIMIZATIONS` is set, overriding Spack's target-derived flags and breaking build cache consumers with a narrower CPU.


jhale:
I used Claude Sonnet 5.0 to draft this PR.

This was picked up using KaHiP in build caches on machines with different architectures.

garth-wells (approved):

PR #6149, opened 2026-08-19
scorep: new variant to enable xray feature

pearzt:
Adds a new variant `xray` to enable the XRay-based instrumentation feature that was introduced with Score-P 10.

### DIFF
--- a/repos/spack_repo/builtin/packages/scorep/package.py
+++ b/repos/spack_repo/builtin/packages/scorep/package.py
@@ -57,7 +57,12 @@ class Scorep(AutotoolsPackage):
     variant(
         "llvm-plugin", default=True, description="Enable LLVM compiler plugin", when="@9.0: ^llvm"
     )
-    variant("xray", default=False, description="Enable instrumetation via LLVM XRay", when="@10.0: %llvm")
+    variant(
+        "xray",
+        default=False,
+        description="Enable instrumetation via LLVM XRay",
+        when="@10.0: %llvm",
+    )
     variant(
         "binutils",
         default=True,
@@ -133,7 +138,7 @@ class Scorep(AutotoolsPackage):
     # Score-P first has support for ROCm 6.x as of v8.4
     conflicts("hip@6.0:", when="@:8.3+hip")
 
-    # XRay-based instrumentation is mutually exclusive 
+    # XRay-based instrumentation is mutually exclusive
     # with the LLVM instrumentation plugin
     conflicts("+xray", when="+llvm-plugin")
 
--- a/repos/spack_repo/builtin/packages/scorep/package.py
+++ b/repos/spack_repo/builtin/packages/scorep/package.py
@@ -55,7 +55,10 @@ class Scorep(AutotoolsPackage):
     variant("hip", default=False, description="Enable ROCm/HIP support", when="@8.0:")
     variant("gcc-plugin", default=True, description="Enable gcc-plugin", when="%gcc")
     variant(
-        "llvm-plugin", default=True, description="Enable LLVM compiler plugin", when="@9.0: %c,cxx=llvm"
+        "llvm-plugin",
+        default=True,
+        description="Enable LLVM compiler plugin",
+        when="@9.0: %c,cxx=llvm",
     )
     variant(
         "xray",
--- a/repos/spack_repo/builtin/packages/scorep/package.py
+++ b/repos/spack_repo/builtin/packages/scorep/package.py
@@ -55,7 +55,7 @@ class Scorep(AutotoolsPackage):
     variant("hip", default=False, description="Enable ROCm/HIP support", when="@8.0:")
     variant("gcc-plugin", default=True, description="Enable gcc-plugin", when="%gcc")
     variant(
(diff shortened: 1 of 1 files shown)


pearzt:
@spackbot fix style

repos/spack_repo/builtin/packages/scorep/package.py
```diff
@@ -57,6 +57,12 @@ class Scorep(AutotoolsPackage):
     variant(
         "llvm-plugin", default=True, description="Enable LLVM compiler plugin", when="@9.0: ^llvm"
     )
+    variant(
+        "xray",
+        default=False,
+        description="Enable instrumetation via LLVM XRay",
+        when="@10.0: %llvm",
+    )
```
wrwilliams, line 65: Direct/transitive dependency here should be consistent.

I think what we want is "when LLVM provides at least one language virtual" for both, yeah? And that's not actually expressed by either of these.

wrwilliams (changes requested):
Safe to assume that LLVM supports Xray out-of-the-box or do we need a variant there as well?

pearzt:
> Safe to assume that LLVM supports Xray out-of-the-box or do we need a variant there as well?

I think it's safe to assume. `compiler-rt` needs to be built (which the Spack package does) and `COMPILER_RT_BUILD_XRAY` can be used to control whether XRay is enabled (which defaults to true and is not touched by the Spack package).

repos/spack_repo/builtin/packages/scorep/package.py
pearzt: Good point. I don't fully understand "when LLVM provides at least one language virtual", though. Shouldn't all language virtuals be provided by LLVM before allowing to build with XRay support?

wrwilliams: Since we currently expect only homogeneous toolchains in the Spack package, these should be equivalent but "all supported" is the most correct expression (ie `+xray ~fortran` is of course legal). The only reasonable corner cases here are things like custom compilers in Score-P for Clang/gfortran if we added support for that to the package.

pearzt: So, if I understand you correctly, `%c,cxx=llvm` models what we want, right?

wrwilliams: I think `%c,cxx=llvm` and `with when=+fortran` guarding a `%fortran=llvm`? But yes, that sounds right for the c/c++ cases.

(Related: should `compiler_instrumentation` become a single variant with `xray/llvm-plugin/gcc-plugin/instrument-functions/tcollect` values? Out of scope here but that may be nicer for users.)

pearzt: AFAIK, XRay currently does not support Fortran at all. (Mostly because nobody got to integrating it into the flang compiler driver yet, I think, not because of any major roadblocks.) So, I guess we don't really care about the Fortran compiler here? Should we still add the guarded `%fortran=llvm` in preparation for future LLVM version potentially lifting this restriction?

And yes, I think an explicit variant for different instrumentation plugins would be nice and would better express that those are mutually exclusive.

wrwilliams: Let's just go with C/C++ for now then; we can get fancy later when we need it.

pearzt: Alright, I just updated the PR. It's still not consistent with the constraint for `llvm-plugin`, though. Should we fix this here? We'd probably need the special case for Fortran there, right?

wrwilliams: It's probably better to fix here so that the `conflict` declaration doesn't (well, shouldn't) hit any weird corner cases. (Or I can do another PR to fix that and we merge that one first if you want.) I'm fine without the Fortran special case there at least to start with. scorep with Clang+Gfortran will require several other changes if anyone wants that to actually work under Spack.

repos/spack_repo/builtin/packages/scorep/package.py
```diff
@@ -57,6 +57,12 @@ class Scorep(AutotoolsPackage):
     variant(
         "llvm-plugin", default=True, description="Enable LLVM compiler plugin", when="@9.0: ^llvm"
```
wrwilliams:
```suggestion
        "llvm-plugin", default=True, description="Enable LLVM compiler plugin", when="@9.0: %c,cxx=llvm"
```

repos/spack_repo/builtin/packages/scorep/package.py
pearzt: I just accepted your suggested change. In the future, having the special case for Fortran would be useful, since otherwise someone might try to compile with `+llvm-plugin +fortran %c,cxx=llvm %fortran=gcc` which would probably to weird results, I guess. It can probably be elegantly expressed if you should decide to go with a `compiler_instrumentation` variant.

pearzt:
@spackbot fix style

wrwilliams (approved):
LGTM now.

RikkiButler20 (approved):

PR #6150, opened 2026-08-19
ascent: fix CUDA 13 build and allow raja@develop

biddisco:
This PR fixes two issues that prevent `ascent@develop +cuda` from building with CUDA 13 and `raja@develop`.

1. **CUDA 13 compatibility**
   CUDA 13.0 removed `cudaDeviceProp.memoryClockRate`. The bundled BLT smoke test `blt_cuda_runtime_smoke.cpp` still references it, causing:
   ```text
   error: 'struct cudaDeviceProp' has no member named 'memoryClockRate'
   Add a patch that guards the memoryClockRate usage with #if defined(CUDART_VERSION) && CUDART_VERSION < 13000.
2. Allow raja@develop
The dependency depends_on("raja@2024.02.1:2025.03.1", when="@0.9.3:") also matched @develop, blocking raja@develop. Narrow it to @0.9.3:0.9.5.
3. Match CUDA C++ standard to RAJA
With raja@develop built as C++20, the bundled camp headers require C++20 concepts, but CUDA files were compiled with CMake's default CMAKE_CUDA_STANDARD=17. Set CMAKE_CUDA_STANDARD from RAJA's cxxstd variant when +raja.
Tested by building ascent@develop +cuda with cuda@13.0.2 and raja@develop.

### DIFF
--- a/repos/spack_repo/builtin/packages/ascent/package.py
+++ b/repos/spack_repo/builtin/packages/ascent/package.py
@@ -235,7 +235,8 @@ class Ascent(CMakePackage, CudaPackage, ROCmPackage):
     #######################
     with when("+raja"):
         depends_on("raja")
-        depends_on("raja@2024.02.1:2025.03.1", when="@0.9.3:")
+        # develop needs raja@develop, so restrict the release range to 0.9.x
+        depends_on("raja@2024.02.1:2025.03.1", when="@0.9.3:0.9.5")
         depends_on("raja+openmp", when="+openmp")
         depends_on("raja~openmp", when="~openmp")
         depends_on("raja+rocm", when="+rocm")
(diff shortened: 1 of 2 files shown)


repos/spack_repo/builtin/packages/ascent/package.py
alalazo, line 249: Isn't this the same bug on `umpire`?

repos/spack_repo/builtin/packages/ascent/package.py
```diff
@@ -173,6 +173,10 @@ class Ascent(CMakePackage, CudaPackage, ROCmPackage):
         sha256="0dc417d8a454d235cdeb9e0f0bb527dc3c42a1eb6ae80e8bd5b33ead19198329",
     )
 
+    # cudaDeviceProp.memoryClockRate was removed in CUDA 13.0
+    patch("ascent-blt-cuda13-memory-clock.patch", when="@:0.9.5 +cuda")
+    patch("ascent-blt-cuda13-memory-clock.patch", when="@develop +cuda")
```
alalazo, line 178: The name of the patch suggests this is a CUDA 13 problem, but the patch applies to all CUDA versions

alalazo (changes requested):

PR #6151, opened 2026-08-19
libfabric: fix null deref in ofi_import_monitor_unsubscribe

biddisco:
`ofi_import_monitor_unsubscribe()` in `prov/util/src/import_mem_monitor.c` asserts `impmon.impfid` before dereferencing it, but the function can be called when no imported monitor has been set up. Replace the assert with a null check / early return to avoid a crash.

The imported memory monitor file was introduced in libfabric v2.1.0 and the issue is still present in v2.6.0 and `main`, so scope the patch with `when="@2.1.0:"`.

### DIFF
--- a/repos/spack_repo/builtin/packages/libfabric/package.py
+++ b/repos/spack_repo/builtin/packages/libfabric/package.py
@@ -157,6 +157,9 @@ class Libfabric(AutotoolsPackage, CudaPackage, ROCmPackage):
     # Fix for the inline assembly problem for the Nvidia compilers
     # https://github.com/ofiwg/libfabric/pull/7665
     patch("nvhpc-symver.patch", when="@1.6.0:1.14.0 %nvhpc")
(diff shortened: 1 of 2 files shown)


repos/spack_repo/builtin/packages/libfabric/package.py
```diff
@@ -157,6 +157,9 @@ class Libfabric(AutotoolsPackage, CudaPackage, ROCmPackage):
     # Fix for the inline assembly problem for the Nvidia compilers
     # https://github.com/ofiwg/libfabric/pull/7665
     patch("nvhpc-symver.patch", when="@1.6.0:1.14.0 %nvhpc")
+    # Fix null deref in ofi_import_monitor_unsubscribe when no imported monitor
+    # is active. File introduced in v2.1.0; still unfixed as of v2.6.0/main.
+    patch("mr_unsubscribe.patch", when="@2.1.0:")
```
rbberger, line 162: Was this reported upstream and if so is there a PR?

PR #6152, opened 2026-08-19
py-dm-tree: 0.1.10 needs abseil-cpp@20250814: for absl:MutexLock(mutex)

bernhardkaindl:
py-dm-tree@0.1.10: started to use `absl:MutexLock(mutex)`:
https://github.com/google-deepmind/tree/compare/0.1.9...0.1.10#diff-ae58474be80ffb81dab28f1870a4514f935ede4508766a4461f7b88563330c4fR118
 
`absl:MutexLock(mutex)` started to be provided with `abseil-cpp@20250814:`, so update the unversioned dependency for `abseil-cpp` to use this as the minimum version.

This should fix the HEP stack GitLab CI pipelines for the `develop` branch which are currently failing because of this missing constraint for `py-dm-tree@0.1.10`:
- https://gitlab.spack.io/spack/spack-packages/-/jobs/24059703
- https://gitlab.spack.io/spack/spack-packages/-/jobs/24059178
- https://gitlab.spack.io/spack/spack-packages/-/jobs/24058804

### DIFF
--- a/repos/spack_repo/builtin/packages/py_dm_tree/package.py
+++ b/repos/spack_repo/builtin/packages/py_dm_tree/package.py
@@ -41,7 +41,8 @@ class PyDmTree(CMakePackage, PythonExtension):
         depends_on("python@:3.11", when="@:0.1.8")
         depends_on("python@:3.10", when="@:0.1.7")
 
-    depends_on("abseil-cpp cxxstd=17", type="link", when="@0.1.10:")
+    # 0.1.10 added using absl::MutexLock lock(mutex_) which needs abseil-cpp@20250814.1:
+    depends_on("abseil-cpp@20250814.1: cxxstd=17", type="link", when="@0.1.10:")
     depends_on("abseil-cpp cxxstd=14", type="link", when="@0.1.8:0.1.9")
     depends_on("py-attrs@18.2.0:", type=("build", "run"), when="@0.1.9:")
     depends_on("py-wrapt@1.11.2:", type=("build", "run"), when="@0.1.9:")


repos/spack_repo/builtin/packages/py_dm_tree/package.py
```diff
@@ -41,7 +41,8 @@ class PyDmTree(CMakePackage, PythonExtension):
         depends_on("python@:3.11", when="@:0.1.8")
         depends_on("python@:3.10", when="@:0.1.7")
 
-    depends_on("abseil-cpp cxxstd=17", type="link", when="@0.1.10:")
+    # 0.1.10 added using absl::MutexLock lock(mutex_) which needs abseil-cpp@20250814.1:
+    depends_on("abseil-cpp@20250814.1: cxxstd=17", type="link", when="@0.1.10:")
```
RikkiButler20, line 45:
```suggestion
    depends_on("abseil-cpp@20260107.1: cxxstd=17", type="link", when="@0.1.10:")
```
I think you've got the wrong version.
[py-dm-tree CMakeLists.txt](https://github.com/google-deepmind/tree/blob/9330c9565222ca83905aabe9c3ad737037328269/tree/CMakeLists.txt#L92)

RikkiButler20 (changes requested):

repos/spack_repo/builtin/packages/py_dm_tree/package.py
bernhardkaindl: Hi @RikkiButler20, thanks for the review!

[py-dm-tree CMakeLists.txt](https://github.com/google-deepmind/tree/blob/9330c9565222ca83905aabe9c3ad737037328269/tree/CMakeLists.txt#L92) has:
```cmake
FetchContent_Declare(absl URL ${ABSEIL_REPO}/archive/refs/tags/20260107.1.tar.gz ...)
```

This is not a minimum required version, it is the version they used for their fetcher when abseil is not installed and the package needs to fetch it.

This would normally be a newer version than they actually require as mininum.

The version that added the required `absl:MutexLock(mutex)` is the version that I added as a minimum version.

Our current build will use the `abseil-cpp@20260107.1` (the same version they fetch when it is not installed) as it is the only newer version that spack as a checksum of.

Thus, the minimum only specifies what would be the minimum version, not what would normally be picked, which currently is the version that the FetchContent_Declare also would fetch.

The minimum version I specified in the PR is correct, but it is only the absolute minimum and spack will use the fetched version as it is the only version available for this constraint currently.

RikkiButler20: Yep. I'm with you now. 😃

RikkiButler20 (approved):

Issue #6098, opened 2026-08-14
Installation issue: gromacs: TypeError when specifying openmp_max_threads


payerle:
### Steps to reproduce the issue

I encountered a very non-helpful error message when trying to concretize gromacs with a value for the variant openmp_max_threads.  

E.g., running
```console
$ spack spec -lI gromacs openmp_max_threads=none
==> Error: '<' not supported between instances of 'str' and 'bool'
```

Similarly,
```console
$ spack spec -lI gromacs openmp_max_threads=64
==> Error: '<' not supported between instances of 'str' and 'bool'
```

When run with the --debug flag, (see output file in Additional Information section), I see that the error is in the execution of "sorted(self.variant_values_from_specs)" in the file lib/spack/spack/solver/asp.py.

I get the same results when attempting to concretize an environment with gromacs as a root spec and a requires setting the openmp_max_threads for gromacs (in which case the error message is particularly less than helpful, as the error message (even with debug flag) does not specify which of the ~100 root specs is the cause of the problem, much less which variant, etc).  I had to add debugging code to determine the issue was related to package variants, much less which package).

I believe there are two issues here:
1) gromacs appears to have an issue with the openmp_max_threads flag.  I have looked at the recipe, and I am quite confused at what the recipe is trying to do:
1a) the value is used to set the CMake flag GMX_OPENMP_MAX_THREADS which seems to want a string representation of an integer.
1b) the default value of "none" seems to also suggest a string type
1c) There is a conflicts line (around line 182) with "+openmp_max_threads", i.e. treating the variant as a boolean

2) in my opinion, the main spack code should be more robust in handling such issues from package recipes.  I have not so far been able to reproduce the TypeError exception with another package/variant, so it seems like this issue is due to something funky with the gromacs package.  But it does seem to be at least theoretically plausible that another current or future package recipe could similarly tickle a TypeError in this way, and so spack should produce better diagnostics in this case.  I others agree, I will open an issue in the spack repo.

### Error message

<details><summary>Error message</summary>

<pre>
==> Error: '<' not supported between instances of 'str' and 'bool'
</pre></details>

### Information on your system

```console
$ spack debug report
* **Spack:** 1.2.0.dev0
* **Builtin repo:** https://github.com/spack/spack-packages/commit/119680aeee8ea802c6111b7167583bddef97e82f
* **Python:** 3.6.8
* **Platform:** linux-rhel8-zen2
```

### Additional information

[...]

PR #6158, opened 2026-08-19
gromacs: make variant openmp_max_threads conditional on +openmp

payerle:
This should fix issue #6098

The issue described in #6098 appears to be manifested because of the conflict on openmp_max_threads being set when ~openmp.  That conflict used "+openmp_max_threads", causing a TypeError is openmp_max_threads is set to any value.

This PR removes the conflict, and makes the openmp_max_threads variant conditional on +openmp, thereby making the conflict moot.

### DIFF
--- a/repos/spack_repo/builtin/packages/gromacs/package.py
+++ b/repos/spack_repo/builtin/packages/gromacs/package.py
@@ -179,8 +179,8 @@ class Gromacs(CMakePackage, CudaPackage):
     )
 
     variant(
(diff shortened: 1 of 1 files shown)


payerle:
@spackbot fix style

bernhardkaindl (approved):

mabraham:
Thanks, I was working on this last week and came to the same solution but didn't get to upload a PR!

PR #6161, opened 2026-08-20
visit: add v3.5.0

vicentebolea:


cyrush:
One challenge is that VisIt 3.5.0 uses Qt6, and the Qt spack packages were refactored into smaller packages starting with Qt6 -- due to the modular changes between Qt5 and Qt6

vicentebolea:
@cyrush CI passed! :tada:

cyrush:
Oh wow, how do we see the results of those visit builds?
(I couldn't find them in gitlab ci easily)

vicentebolea:
> Oh wow, how do we see the results of those visit builds? (I couldn't find them in gitlab ci easily)

Good catch, the changes in the deps made the concretizer to prefer visit 3.4.1 which in turn resulted in no new builds. I am temporarily forcing visit 3.5.0 in the dav-sdk stack to test this new version.

kwryankrattiger:
Looks like HDF5 API compatibility issue.

```
src/databases/OpenPMD/OpenPMDClasses/PMDFile.C:303:34: error: too many arguments to function 'herr_t H5Oget_info_by_name1(hid_t, const char*, H5O_info1_t*, hid_t)'
```

VisIt is setting the API version explicitly here, duplicate of the command line we use but that should be ok.

```
src/databases/OpenPMD/OpenPMDClasses/PMDParticle.h:26: warning: "H5_USE_18_API" redefined
```

kwryankrattiger:
> I am temporarily forcing visit 3.5.0 in the dav-sdk stack to test this new version.

We should probably do that anyway, older versions of VTK are going to get removed before long and that means most of the versions of VisIt will also have to go. Forcing the latest is not a bad idea, at very least it will keep testing the VTK upper bound for VisIt in CI.

vicentebolea:
> Looks like HDF5 API compatibility issue.
> 
> ```
> src/databases/OpenPMD/OpenPMDClasses/PMDFile.C:303:34: error: too many arguments to function 'herr_t H5Oget_info_by_name1(hid_t, const char*, H5O_info1_t*, hid_t)'
> ```
> 
> VisIt is setting the API version explicitly here, duplicate of the command line we use but that should be ok.
> 
> ```
> src/databases/OpenPMD/OpenPMDClasses/PMDParticle.h:26: warning: "H5_USE_18_API" redefined
> ```

Thanks for pointing this out, how did you get this errors? in the CI job it says: `Job's log exceeded limit of`

kwryankrattiger:
> Thanks for pointing this out, how did you get this errors? in the CI job it says: Job's log exceeded limit of

You can get the full logs from the artifacts under `jobs_scratch_dir/<stack-name>/logs/spack-build-out.txt.gz`. If the job fails for infrastructure reasons that isn't always available, but when it is a build error it should be there.

Here is the full log of the latest failure.
[spack-build-out.visit.txt.gz](https://github.com/user-attachments/files/31497164/spack-build-out.visit.txt.gz)

```
src/databases/OpenPMD/OpenPMDClasses/PMDIteration.C:143:50: error: too many arguments to function 'herr_t H5Oget_info_by_name1(hid_t, const char*, H5O_info1_t*, hid_t)'
...
src/databases/OpenPMD/OpenPMDClasses/PMDIteration.C:363:58: error: too many arguments to function 'herr_t H5Oget_info_by_name1(hid_t, const char*, H5O_info1_t*, hid_t)'
```

This looks like the signature for `H5Oget_info_by_name2` which was introduced in 1.10.3, so I am curious how this was building before.

kwryankrattiger (approved):

PR #6166, opened 2026-08-20
viskores: bump new version v1.2.0

vicentebolea:


johnwparent (approved):
Looks good to me!

It could be nice to have some reference to release notes for version bump PRs in the description so we can record them in the commit that gets merged to develop and folks looking through history can have that reference. Not a requirement by any means, so feel free to ignore.

kmorel (approved):

PR #6168, opened 2026-08-20
BLT: drop language requirements

johnwparent:
BLT is a Package, that simply copies itself into the install tree, there is no building, and thus, no requirement on any compilers.

Remove the language dependencies.

### DIFF
--- a/repos/spack_repo/builtin/packages/blt/package.py
+++ b/repos/spack_repo/builtin/packages/blt/package.py
@@ -85,10 +85,6 @@ class Blt(Package):
         when="@0.7.1",
     )
 
-    depends_on("c", type="build")  # generated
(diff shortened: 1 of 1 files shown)


white238:
@becker33 can you think of any side effects of this? One thing could be if we add tests they wont be able to run. We havent gotten around to doing this yet but should.

johnwparent:
FWIW I don't think the tests would run as the recipe stands now as the language requirement is `build`. 
I'm happy to add tests if it's straightforward.

white238:
Totally fine. I can also fix it when we get around to adding the tests. We have an effort to add the necessary plumbing in our packages to run the tests and havent gotten to BLT yet.

white238 (approved):

bernhardkaindl (approved):

PR #6169, opened 2026-08-20
bdsim: add v1.7.8

tmadlener:
Add version 1.7.8 for bdsim. This is necessary to build with c++23. See https://github.com/bdsim-collaboration/bdsim/pull/141 I am not sure what should go into a `conflict` to make that a bit more obvious for older versions.

@jmcarcell any idea?

### DIFF
--- a/repos/spack_repo/builtin/packages/bdsim/package.py
+++ b/repos/spack_repo/builtin/packages/bdsim/package.py
@@ -24,6 +24,7 @@ class Bdsim(CMakePackage):
     license("GPL-3.0-or-later")
 
     version("master", branch="master")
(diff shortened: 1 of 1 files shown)


jmcarcell:
I'd leave it as it is since not that many people are using it. It's complicated and it isn't clear if it's using the one from Geant4 always or the one from ROOT. Adding `cxxstd` doesn't work obviously. If any conflict were to be set it would be something like `geant4 cxxstd=23` and possibly `root cxxstd=23` before 1.7.8.

jmcarcell (approved):

RikkiButler20 (approved):
LGTM

PR #6172, opened 2026-08-20
py-onnxruntime: Allow builds with gcc16 for 1.22.2

tmadlener:
Apply patch that landed upstream after 1.22.


wdconinc:
Can we use https://github.com/microsoft/onnxruntime/commit/f7619dc93f592ddfc10f12f7145f9781299163a0 by url instead?

tmadlener:
Yes we can. (one does get lazy when handing things off to the LLMs).

wdconinc (approved):

PR #6173, opened 2026-08-20
acts: allow builds with more recent numpys

tmadlener:
This is a first shot at allowing to build acts with newer versions of numpy. I have successfully built the current main branch with this. I *think* this should be possible from acts@46 onwards, but I am not entirely sure.

### DIFF
--- a/repos/spack_repo/builtin/packages/acts/package.py
+++ b/repos/spack_repo/builtin/packages/acts/package.py
@@ -241,7 +241,8 @@ class Acts(CMakePackage, CudaPackage):
     # robust solution is available we pretend that these packages are also
     # run- and link-time dependencies.
     depends_on("python@3.12:", when="@44:")
-    depends_on("py-numpy @2.2", when="@44:")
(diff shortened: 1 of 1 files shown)


wdconinc:
Is this just a matter of building or also of correctness? I've remove the 2.2 pinning a long time ago locally and not had any issues :-)

tmadlener:
Good question. For now (and for me) it has been mostly about building. I am not sure I actually exercise anything of acts at the moment that needs numpy.

paulgessinger:
I'm fairly sure it was never an ACTS-level restriction, but a consequence of how the dependencies are declared in the spack package. The codegen, that I believe is what pulls this in, doesn't care which numpy version is used.

tmadlener:
Should we consider dropping it entirely then? Otherwise you will not be able to build acts with c++23 via spack.

paulgessinger:
I remember some context was here: https://github.com/spack/spack-packages/pull/3064/changes#r2827919551

PR #6179, opened 2026-08-21
py-ipyvue: add v3.0.0

jeremyfix:
This is adding ipyvue version 3.0.0 

The sha can be verified from the pypi page : https://pypi.org/project/ipyvue/#ipyvue-3.0.0.tar.gz 

It brings support for Vue 3 (see https://github.com/widgetti/ipyvue/releases/tag/v3.0.0 ) , Vue 2 support has ended on Dec 31, 2023 (see https://v2.vuejs.org/eol/) .


repos/spack_repo/builtin/packages/py_ipyvue/package.py
bernhardkaindl, line 27: https://github.com/widgetti/ipyvue/blob/master/pyproject.toml#L1C1-L6C2
I assume therefore:
```suggestion
    with default_args(type=("build", "run")):
        depends_on("py-ipywidgets@7:", when=@:1)
        depends_on("py-wheel", when=@3:)
        depends_on("py-jupyterlab@4:", when=@3:)
```
Or not? Can you check?

Theh end of setup.py suggests support is only up to Python 3.11? Could you run it with newer Python versions?
https://github.com/widgetti/ipyvue/blob/master/setup.py#L243

PR #6188, opened 2026-08-21
ICON Subclassing without parsing configure args

leclairm:
In order to make extending the ICON package recipe easier, we suggest the following:

- Store the variants to be en-or-disabled in the `EN_DIS_ABLE_FLAGS` class variable.
- Create 3 new attributes, `single_args`, `flags` and `config_libs` to host configure args.
- Populate them in the new `set_configure_args` method.
- `configure_args` becomes a small method that concatenate the content of these 3 attributes in the final `list[str]`.

Doing this would greatly improve the subclassing experience of the ICON package recipe like it is done [here](https://github.com/C2SM/spack-c2sm/blob/main/repos/spack_repo/c2sm/packages/icon_nwp/package.py) where parsing of the configure args is necessary. With this PR, we could modify `self.single_args` and `self.flags` where we need, e.g. with
```python
    def set_configure_args(self) -> None:
        super().set_configure_args()
        # Modify self.single_args, self.flags and self.config_libs
```
or
```python
    @run_before("configure")
    def extend_configure_args(self) -> None:
        # Modify self.single_args, self.flags and self.config_libs
```
Also extending  `EN_DIS_ABLE_FLAGS`  would help by keeping the logic in the parent class.

### DIFF
--- a/repos/spack_repo/builtin/packages/icon/package.py
+++ b/repos/spack_repo/builtin/packages/icon/package.py
@@ -147,13 +147,9 @@ class Icon(AutotoolsPackage):
     for __x in nvidia_targets.keys():
         depends_on("cuda", when="gpu={0}".format(__x))
 
-    def __init__(self, spec: spack.spec.Spec) -> None:
-        super().__init__()
-        self.single_args: list[str] = []
-        self.flags: dict(str, list[str]) = defaultdict(list)
-
     def set_configure_args(self) -> None:
-        self.single_args.append("--disable-rpaths")
+        self.flags: dict(str, list[str]) = defaultdict(list)
+        self.single_args: list[str] = ["--disable-rpaths"]
         libs = LibraryList([])
 
         for x in [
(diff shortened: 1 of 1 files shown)


Try2Code:
hi!
I hope u realize that any spack package will always lag behind the icon building - by definition.

but a little off-topic question: why is a subclass of icon actually needed? I was not aware of it

leclairm:
Hi!
> hi! I hope u realize that any spack package will always lag behind the icon building - by definition.

Of course I do. I don't get why that question?

> but a little off-topic question: why is a subclass of icon actually needed? I was not aware of it

Ah, I didn't know this wasn't known! We rely on a subclass at MCH and C2SM since years where we put different variants or behaviours not necessarily suitable for upstreaming. Also for being able to quickly give access to a feature if necessary. Now that ICON is open source, we can probably upstream quite a bit of these to the main ICON recipe. I was hoping to do that in a second stage.

leclairm:
@Try2Code Do you otherwise have some comments on the content of the PR? Is something not suitable from your side?

PR #6189, opened 2026-08-22
seacas: Allow building applications without matio

clintonstimpson:
This enables building multiple seacas applications without the ones requiring matio, if matio is excluded.


repos/spack_repo/builtin/packages/seacas/package.py
```diff
@@ -389,6 +389,11 @@ def cmake_args(self):
                 ]
             )
 
+            app_dependencies = {
+                    'Exo2mat' : ['+matio'],
+                    'Mat2exo' : ['+matio'],
+                    }
```
bernhardkaindl: This will not be accepted by the CI check running `ruff format`. 
```py
ruff format repos/spack_repo/builtin/packages/seacas/package.py
1 file reformatted
```
The change would look this this:
```suggestion
                "Exo2mat": ["+matio"],
                "Mat2exo": ["+matio"],
            }
```
This is by cut-and-paste, so it could be slightly off too.

Preferably, can you install ruff and run `ruff format repos/spack_repo/builtin/packages/seacas/package.py` before pushing?

Here are the different installation methods: https://docs.astral.sh/ruff/installation/

clintonstimpson: Ok, its fixed now.  Thanks.

johnwparent (approved):

PR #6190, opened 2026-08-22
qt-tools: add new uitools variant

clintonstimpson:


repos/spack_repo/builtin/packages/qt_tools/package.py
```diff
@@ -82,4 +87,5 @@ def cmake_args(self):
             self.define_qt_feature_from_variant("designer"),
             self.define_qt_feature_from_variant("zstd", "designer"),
             self.define_qt_feature_from_variant("linguist"),
+            self.define_qt_feature_from_variant("uitools"),
```
wdconinc, line 90: Does this actually work? There is no `QT_FEATURE_uitools` variable anywhere and the build of `uitools` is simply gated by the target Qt::Widgets (https://github.com/qt/qttools/blob/dev/src/CMakeLists.txt#L24). You should get the `uitools` feature with `qt-base +widgets`.

clintonstimpson: Oh, you're right that it looks suspicious.  I'll have to go back and find out why doing this made my application build.
Let's disregard this pull request.

PR #6191, opened 2026-08-22
py-dotmatch: Add 0.2.2

dnncha:
## Description
Add DotMatch 0.2.2, a known-target short-DNA assignment tool for FASTQ and other short-DNA target sets.

## Checks
- PyPI sdist URL and SHA256 verified against the published archive.
- Package module loads with Spack 1.3.0.dev0 package API.
- `python3 -m py_compile repos/spack_repo/builtin/packages/py_dotmatch/package.py`
- `git diff --check`

## Dependencies
- Python 3.9 or newer
- setuptools 77 or newer for the PEP 517 build
- tomli at runtime for Python 3.10 and older

This adds a distribution route only; package downloads or user adoption are not claimed by this PR.

### DIFF
--- a/repos/spack_repo/builtin/packages/py_dotmatch/package.py
+++ b/repos/spack_repo/builtin/packages/py_dotmatch/package.py
@@ -1,0 +1,22 @@
+# Copyright Spack Project Developers. See COPYRIGHT file for details.
+#
+# SPDX-License-Identifier: (Apache-2.0 OR MIT)
+
+from spack_repo.builtin.build_systems.python import PythonPackage
+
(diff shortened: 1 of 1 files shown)


dnncha:
Additional validation: DotMatch 0.2.2 was installed from the published sdist in a clean Python 3.9 virtual environment with `setuptools>=77` and `wheel`, using `pip --no-build-isolation --no-deps` as Spack does. Import and smoke checks passed for `dotmatch` and `quickdna`.

bernhardkaindl (approved):

PR #6193, opened 2026-08-23
linbox: new package

d-torrance:
We add [linbox](https://linalg.org/), a C++ library for exact computation algebra.

We also make a small update to givaro, one of linbox's dependencies, so that the correct gmp flags end up in givaro's pkg-config file.

### DIFF
--- a/repos/spack_repo/builtin/packages/linbox/package.py
+++ b/repos/spack_repo/builtin/packages/linbox/package.py
@@ -1,0 +1,46 @@
+# Copyright Spack Project Developers. See COPYRIGHT file for details.
+#
+# SPDX-License-Identifier: (Apache-2.0 OR MIT)
+
+from spack_repo.builtin.build_systems.autotools import AutotoolsPackage
+
+from spack.package import *
+
+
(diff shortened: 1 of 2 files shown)


repos/spack_repo/builtin/packages/linbox/package.py
```diff
@@ -0,0 +1,48 @@
...
+    variant("ntl", default=False, description="Enable NTL support")
+    depends_on("ntl", when="+ntl")
+
+    def configure_args(self):
+        args = []
+        for dep in ["flint", "fplll", "mpfr", "ntl"]:
+            if self.spec.satisfies(f"^{dep}"):
+                args.append(f"--with-{dep}={self.spec[dep].prefix}")
+            else:
+                args.append(f"--without-{dep}")
```
alalazo: Here you can probably use `self.with_or_without`

d-torrance: That's slick -- thanks for the pointer!

I've updated the PR accordingly

alalazo (approved):

PR #6194, opened 2026-08-23
macaulay2: drop -lquadmath from FLIBS hack

d-torrance:
It was causing linking errors on ARM machines, where libquadmath isn't available.  It was really only there because we needed something nonempty to avoid `AC_FC_LIBRARY_LDFLAGS` trying to generate them (resulting in https://github.com/Macaulay2/M2/issues/4074), and `-lquadmath` was there on x86_64.  The build works just fine without it.


RikkiButler20 (approved):
LGTM

PR #6196, opened 2026-08-23
py-scipy: add v1.18.1

adamjstewart:
https://github.com/scipy/scipy/releases/tag/v1.18.1

### DIFF
--- a/repos/spack_repo/builtin/packages/py_pythran/package.py
+++ b/repos/spack_repo/builtin/packages/py_pythran/package.py
@@ -19,6 +19,7 @@ class PyPythran(PythonPackage):
     license("BSD-3-Clause")
     maintainers("rgommers")
 
+    version("0.19.0", sha256="18a096daa8b4bbe3ace06ce5d7325be5ff57aec4cf67b91834cf1b757b3e5392")
     version("0.18.1", sha256="8803ed948bf841a11bbbb10472a8ff6ea24ebd70e67c3f77b77be3db900eccfe")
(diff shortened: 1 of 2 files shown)


rgommers (commented):
Thanks @adamjstewart. Looks good, except it misses a version bump for `pythran` for python>=3.15:

https://github.com/scipy/scipy/blob/23008e81add15196d49e443e3b2975e9faf49f96/pyproject.toml#L30-L33

adamjstewart:
See if this makes sense, my understanding is that the constraint should actually be inside of pythran and was only added to scipy due to missing python_requires metadata in past pythran releases.

rgommers:
Yes, that is strictly better indeed, that should work. The one nit I have there is that previous such changes in `pythran` with newer versions requiring a higher minimum Python version were handled with `conflicts()` rather than `depends_on()`:

https://github.com/adamjstewart/spack-packages/blob/5791dc665b010eaf52f9380cc146ae9e61814eac/repos/spack_repo/builtin/packages/py_pythran/package.py#L88-L91

Probably makes sense to make that uniform?

adamjstewart:
Can't remember why I did it that way. Depends_on is technically preferred because we have audit checks to make sure we remove ancient versions of packages that can no longer be built (e.g., if python 3.10 is removed someday). Let me update all of them.

rgommers (approved):
LGTM now!

RikkiButler20 (approved):
LGTM

PR #6202, opened 2026-08-24
octave: fix opengl/fltk boolean-logic bug and missing gl/glu deps for +qt

gonzabotmdp:
## Problem

`octave`'s configure-args logic for `--without-opengl` has a boolean-logic bug:

```python
if "~opengl" and "~fltk" in spec:
    config_args.extend(["--without-opengl", "--without-framework-opengl"])
```

Python parses this as `bool("~opengl") and ("~fltk" in spec)`. `"~opengl"` is a
non-empty string literal, always truthy, so the check silently collapses to just
`"~fltk" in spec` -- `--without-opengl` gets added any time `~fltk` is set,
regardless of whether `+opengl` (or `+qt`, which pulls in `qt+opengl`) is
actually active.

Separately, `+qt` alone (without `+opengl`/`+fltk`) has no `depends_on` on
`gl`/`glu`. octave's own GUI code (`libgui/graphics/gl-select.cc`) calls the
GL API directly, so without these deps the build has no GL/GLU include/lib
paths and fails with undeclared GL symbols, even though `qt+opengl` is already
a transitive dependency.

## Fix

- Correct the condition to `"~opengl" in spec and "~fltk" in spec and "~qt" in spec`.
- Add `depends_on("gl", when="+qt")` and `depends_on("glu", when="+qt")`.

## Testing

Building `octave+qt` (no `+fltk`) previously failed with undeclared GL symbols
in `gl-select.cc`; with `gl`/`glu` deps added and the corrected condition
(so `--without-opengl` is no longer wrongly injected), it builds and links
successfully. Verified on our own AOCC/GCC14 HPC stack (Spack 1.2.2) before
upstreaming.

### DIFF
--- a/repos/spack_repo/builtin/packages/octave/package.py
+++ b/repos/spack_repo/builtin/packages/octave/package.py
@@ -336,11 +336,12 @@ class Octave(AutotoolsPackage, GNUMirrorPackage):
         else:
             config_args.append("--disable-java")
 
-        # NOTE: the original condition here was ,
-        # which Python parses as  -- the
-        # non-empty string literal "~opengl" is always truthy, so this always
-        # reduced to just , silently adding --without-opengl
-        # any time ~fltk was set, regardless of +opengl/+qt actually being on.
+        # The original condition here was the string "~opengl" ANDed with
+        # the membership test "~fltk" in spec, which Python parses as
+        # bool("~opengl") and ("~fltk" in spec) -- the non-empty string
+        # literal is always truthy, so the check silently collapsed to just
+        # "~fltk" in spec, adding --without-opengl whenever ~fltk was set
+        # regardless of +opengl or +qt actually being active.
         if "~opengl" in spec and "~fltk" in spec and "~qt" in spec:
             config_args.extend(["--without-opengl", "--without-framework-opengl"])
         # TODO:  opengl dependency and package is missing?
--- a/repos/spack_repo/builtin/packages/octave/package.py
+++ b/repos/spack_repo/builtin/packages/octave/package.py
@@ -122,6 +122,15 @@ class Octave(AutotoolsPackage, GNUMirrorPackage):
     depends_on("llvm", when="+llvm")
     depends_on("gl", when="+opengl")
     depends_on("gl", when="+fltk")
+    # octave's own GUI code (libgui/graphics/gl-select.cc) calls the GL
+    # API directly, but this dependency was missing when +qt is active
+    # without +opengl/+fltk, so the build never had mesa's include/lib
+    # paths and failed with undeclared GL symbols.
+    depends_on("gl", when="+qt")
+    # configure only defines HAVE_OPENGL (and runs the -lGL link check)
+    # if both GL/gl.h AND GL/glu.h are found -- GLU is not part of the
+    # system's opengl package on many distros, so build it via Spack too.
+    depends_on("glu", when="+qt")
     depends_on("qhull", when="+qhull")
     depends_on("qrupdate", when="+qrupdate")
     depends_on("qscintilla", when="+qscintilla")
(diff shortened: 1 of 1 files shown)


cessenat (commented):
LGTM

cessenat (approved):

repos/spack_repo/builtin/packages/octave/package.py
```diff
@@ -327,7 +336,13 @@ def configure_args(self):
...
             config_args.append("--disable-java")
 
-        if "~opengl" and "~fltk" in spec:
+        # The original condition here was the string "~opengl" ANDed with
+        # the membership test "~fltk" in spec, which Python parses as
+        # bool("~opengl") and ("~fltk" in spec) -- the non-empty string
+        # literal is always truthy, so the check silently collapsed to just
+        # "~fltk" in spec, adding --without-opengl whenever ~fltk was set
+        # regardless of +opengl or +qt actually being active.
+        if "~opengl" in spec and "~fltk" in spec and "~qt" in spec:
```
bernhardkaindl: Hola @gonzabotmdp, JFYI, this simpler version should have the same effect:
```suggestion
        if spec.satisifes("~opengl~fltk~qt"):
```
The exensive comment on how Python expression syntax works is not needed, especially not its historical reference from how it was written before. Instead, it only the rationale (if needed at all) why only the combinaation of all three variants disabled shall add these options would be fine, but it's obvious (as in makes sense) to me so it does not need that comment in my humble opinion.

Could you check if this simplification works for you? thanks)

Gracias, Bernhardo
(and review my other comment too)

repos/spack_repo/builtin/packages/octave/package.py
```diff
@@ -122,6 +122,15 @@ class Octave(AutotoolsPackage, GNUMirrorPackage):
...
     depends_on("gl", when="+opengl")
     depends_on("gl", when="+fltk")
+    # octave's own GUI code (libgui/graphics/gl-select.cc) calls the GL
+    # API directly, but this dependency was missing when +qt is active
+    # without +opengl/+fltk, so the build never had mesa's include/lib
+    # paths and failed with undeclared GL symbols.
+    depends_on("gl", when="+qt")
+    # configure only defines HAVE_OPENGL (and runs the -lGL link check)
+    # if both GL/gl.h AND GL/glu.h are found -- GLU is not part of the
+    # system's opengl package on many distros, so build it via Spack too.
```
bernhardkaindl:
```suggestion
    depends_on("gl", when="+qt")
```

bernhardkaindl: Hola @gonzabotmdp, I think we don't need these comments in the code:

- The gl and glu includes are common requirements for opengl, many packages did omit these dependencies likely because when the package was created, those might have been installed on the host. I can across such cases too.
- The PR description is already detailed and as the PR number is always part of the commit message when merged, it can always be looked up.

Can you also check if my simplification using speck.satisifies - see my other comment works for you?

Gracias, Bernhardo :)

gonzabotmdp:
Good call, done -- simplified to `spec.satisfies("~opengl~fltk~qt")` and trimmed the comment down to a short rationale, thanks for the suggestion.

bernhardkaindl (commented):
Hola @gonzabotmdp, just two minor comments to streamline the update, please review and test if working for you.

I hope that's ok, when I've your answer from reviewing my comments,
I'd approve, start CI and enable auto-merge after CI pass.

Bernhard

gonzabotmdp:
Hola Bernhard, gracias por el detalle.

1) La simplificacion a `spec.satisfies("~opengl~fltk~qt")` ya estaba aplicada desde tu primer comentario -- puede que tu vista quedara sobre el diff viejo, la version actual del archivo ya la tiene.

2) Recorte los comentarios de las lineas gl/glu tal como sugeriste, coincido en que son redundantes con la descripcion de la PR.

Ambos ya estan pusheados a la rama.

PR #6204, opened 2026-08-24
scafacos: fix malformed Libs:/Libs.private: lines in generated .pc file

gonzabotmdp:
## Problem

scafacos's own `configure`/`Makefile` generates `scafacos.pc` with a
malformed `Libs:` line. Observed non-deterministically across builds:

- sometimes a dangling trailing colon: `-lmpi:`
- sometimes the following line's label glued on directly with no separator
  or newline: `-lmpiLibs.private:`

Any consumer resolving scafacos via pkg-config (or CMake's
`FindPkgConfig`) picks up the malformed token literally and passes it to
the linker, producing spurious `cannot find -l<garbage>` errors -- observed
downstream with `lammps+scafacos`. This is a bug in scafacos's own build
process, unrelated to compiler/toolchain choice.

## Fix

Post-install hook that:
- splits `Libs.private:` onto its own line, regardless of how many stray
  colons precede it, and
- strips any dangling trailing colon left at the end of the `Libs:` line.

## Testing

Reproduced repeatedly (non-deterministic across rebuilds) on our HPC stack
while building `lammps+scafacos`: link failed with `cannot find -l<garbage>`
traced back to the malformed `.pc` file. With the post-install fix applied,
`scafacos.pc` is well-formed and `lammps+scafacos` links and runs
successfully.

### DIFF
--- a/repos/spack_repo/builtin/packages/scafacos/package.py
+++ b/repos/spack_repo/builtin/packages/scafacos/package.py
@@ -2,6 +2,8 @@
 #
 # SPDX-License-Identifier: (Apache-2.0 OR MIT)
 
+import os
+
 from spack_repo.builtin.build_systems.autotools import AutotoolsPackage
 
 from spack.package import *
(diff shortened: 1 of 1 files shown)


hmenke (changes requested):
Could you please ask Claude to instead create a patch for `configure.ac` (which populates `SCAFACOS_PC_LIBS`) that you then also submit upstream?

gonzabotmdp:
Done -- traced the actual root cause (not just moved the same fix around): `configure`'s own `PC_LIBS = ...` debug notice showed a stray trailing colon coming from `AX_CXX_LIBRARY_LDFLAGS`'s parsing of verbose linker output (in `use_fcs_full_ldflags=yes` mode, the only branch actually reachable today). Filed the fix at the `configure.ac` level upstream: scafacos/scafacos#44.

Since this recipe doesn't run autoreconf (and the release tarball ships a pre-generated `configure`), I've updated this PR's `patch()` to apply the equivalent fix directly to the shipped `configure` script instead of regexing the already-generated `.pc` file post-install. Rebuilding scafacos locally with this now (will report back once it finishes) to confirm the generated `scafacos.pc` comes out correct from a clean build, not just patched after the fact.

hmenke (approved):

PR #6206, opened 2026-08-24
grace: fix Motif detection on Xpm-bundled builds + GCC14 strictness

gonzabotmdp:
Several independent issues found while packaging grace against a modern
Motif + GCC 13/14 toolchain -- all reproduced concretely, not speculative.

## 1. Missing libxpm dependency

Some Motif builds (e.g. 2.3.8) compile Xpm support directly into `libXm.so`
rather than shipping a separate `libXpm`. grace's `configure` explicitly
probes for a standalone `-lXpm` ("checking for `XpmCreatePixmapFromData` in
`-lXpm`... no"), and reports this as **"M\*tif has not been found"** --
misleading, since the real gap has nothing to do with Motif itself.
`libxpm` (standalone, X.Org) provides that symbol regardless of how the
system/vendor Motif was built.

## 2. Missing bare `xpm.h` include path

Even with libxpm linked, `configure` also probes for a bare `"xpm.h"` (not
`"X11/xpm.h"`, which is where libxpm actually installs it -- standard X.Org
convention). Without the bare header on the include path, the final
combined "a Motif >= 1002 compatible API" check still fails despite the
link succeeding.

## 3. motif include not reliably auto-injected

`"Xm/XmAll.h: No such file or directory"` -- motif's include path should
come for free via the normal `depends_on`-driven compiler wrapper, but in
practice that wasn't sufficient. Added explicitly, same pattern as (2).

## 4. Motif-API check runs a test binary before RPATH is applied

The final "a Motif >= 1002 compatible API" check isn't just a compile
check -- it uses `ac_fn_c_try_run`, which **compiles and runs** a test
binary (calling `XmVersion`/`XmRegisterConverters`) during `configure`
itself. At that point Spack's final RPATH hasn't been applied yet (that
only happens at the real link/install step), so the dynamic linker can't
find `libXm.so`/`libXpm.so`, the test binary fails to start, and autoconf
reports "Motif not found" even though compiling and linking both actually
succeeded. Fixed by prepending both libraries' lib dirs to
`LD_LIBRARY_PATH` for the duration of the build.

## 5. GCC 13/14 strictness

Newer GCC makes several old-C patterns hard errors by default: implicit
function declarations (`exit()` used without `<stdlib.h>` -- including in
the autoconf-generated `conftest.c` for the check in (4), which then fails
to even compile), K&R-style function definitions without explicit
parameter types (`Xbae/Draw.c`, legal C89), and implicit int/pointer
conversions. Downgraded back to warnings via `CFLAGS`.

## Testing

All five issues were hit in sequence while bringing grace up on an
AOCC/GCC14 HPC stack (Motif 2.3.8, GCC 14.2.0); each fix was isolated by
reverting it individually and confirming the corresponding failure mode
reappeared. With all five applied, grace configures, builds, links, and
`gracebat -version` runs correctly reporting Motif support.

### DIFF
--- a/repos/spack_repo/builtin/packages/grace/package.py
+++ b/repos/spack_repo/builtin/packages/grace/package.py
@@ -32,6 +32,15 @@ class Grace(AutotoolsPackage):
     depends_on("libice")
     depends_on("libsm")
     depends_on("motif")
+    # Some Motif builds (e.g. 2.3.8) compile Xpm support INTO libXm.so
+    # rather than as a separate libXpm, but grace's configure explicitly
+    # probes for "-lXpm" as a standalone library ("checking for
+    # XpmCreatePixmapFromData in -lXpm... no" -> reported as "M*tif has
+    # not been found", which is misleading since the real problem has
+    # nothing to do with Motif itself). libxpm (standalone, X.Org)
+    # provides that symbol regardless of how the system/vendor Motif was
+    # built.
+    depends_on("libxpm")
     depends_on("jpeg")
     depends_on("libpng")
     depends_on("fftw@2.0:2")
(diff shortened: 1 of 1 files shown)


RemiLacroix-IDRIS (changes requested):
Looks mostly good, I am just wondering whether the flags for GCC 14+ support should be defined conditionally.

repos/spack_repo/builtin/packages/grace/package.py
```diff
@@ -55,6 +64,48 @@ def patch(self):
...
+        # Newer GCC (13/14) treats several patterns that used to be mere
+        # warnings as hard errors by default: implicit function
+        # declarations (e.g. exit() used without <stdlib.h> -- including
+        # in the autoconf-generated conftest.c for the check above, which
+        # otherwise fails to even compile), K&R-style function
+        # definitions without explicit parameter types (Xbae/Draw.c,
+        # legal C89, now -Wimplicit-int), and implicit int/pointer
+        # conversions. Downgrade these back to warnings so the build can
+        # proceed as it always has on older GCC.
+        env.append_flags(
```
RemiLacroix-IDRIS: Maybe this should be in an `if self.spec.satisfaties("%gcc@14:"):` condition?

gonzabotmdp:
Good call -- made it conditional on `%gcc@14:` only.

RemiLacroix-IDRIS:
> Good call -- made it conditional on `%gcc@14:` only.

I just realized that the comment says "Newer GCC (13/14)" but the PR title (and the check) only mentions GCC14. Not sure if something needs to be fixed here.

gonzabotmdp:
Good catch -- the comment was just stale wording, the check itself (`%gcc@14:`) was always correct (GCC 14 is where implicit-function-declaration/implicit-int/int-conversion became hard errors by default; GCC 13 doesn't need this). Fixed the comment to say "GCC 14+" instead of "13/14" so it matches the code.

RemiLacroix-IDRIS (approved):
LGTM!

repos/spack_repo/builtin/packages/grace/package.py
RemiLacroix-IDRIS: Fixed

PR #6207, opened 2026-08-24
lammps: set CUDA_HOST_COMPILER for the GPU package (legacy FindCUDA)

gonzabotmdp:
## Problem

The `GPU` package (`cmake/Modules/Packages/GPU.cmake`) uses the classic
`find_package(CUDA)` module (`FindCUDA`, deprecated upstream in CMake but
still what this specific LAMMPS package uses), which reads the variable
`CUDA_HOST_COMPILER` -- **not** `CMAKE_CUDA_HOST_COMPILER`, the variable set
by the modern `enable_language(CUDA)` path used for the rest of the build.

Since nothing in this recipe ever sets `CUDA_HOST_COMPILER`, `FindCUDA`
falls back to its own internal detection instead of the compiler Spack
actually selected for the spec. In practice this means any attempt to
steer the CUDA host compiler (e.g. via `-DCMAKE_CUDA_HOST_COMPILER=...`,
which works for CMake's modern CUDA language support in general) silently
has **no effect** on the `+cuda` GPU package build -- `nvcc` keeps being
invoked with whatever `FindCUDA` picked on its own.

## Fix

Set `CUDA_HOST_COMPILER` to `self.compiler.cxx` explicitly in the `+cuda`
branch, so the legacy `FindCUDA`-based module stays consistent with the
compiler Spack chose for the rest of the build, instead of silently
diverging from it.

## Testing

Found and confirmed while working around `nvcc` 12.6 not supporting GCC 14
as a host compiler (broken `<type_traits>` builtins) on a system that
otherwise builds everything (including the rest of `lammps` itself) with
GCC 14. Setting `-DCMAKE_CUDA_HOST_COMPILER` alone had zero effect --
confirmed via the actual build log, `nvcc` kept being invoked with
`-ccbin .../gcc-14.2.0/bin/g++` regardless. With `CUDA_HOST_COMPILER` set
explicitly (this patch), the GPU package build honors the intended host
compiler and `lammps+cuda` links and runs correctly on real GPU hardware
(A100).

### DIFF
--- a/repos/spack_repo/builtin/packages/lammps/package.py
+++ b/repos/spack_repo/builtin/packages/lammps/package.py
@@ -627,6 +627,20 @@ class Lammps(CMakePackage, CudaPackage, ROCmPackage, PythonExtension):
             if spec.satisfies("+cuda"):
                 args.append(self.define("PKG_GPU", True))
                 args.append(self.define("GPU_API", "cuda"))
+                # The GPU package (cmake/Modules/Packages/GPU.cmake) uses
+                # the classic find_package(CUDA) module (FindCUDA,
+                # deprecated), which reads CUDA_HOST_COMPILER -- a
+                # different variable from CMAKE_CUDA_HOST_COMPILER, the one
+                # set by the modern enable_language(CUDA) path used for the
+                # rest of the build. Because nothing in this recipe sets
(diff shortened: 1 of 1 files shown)


rbberger (approved):
@gonzabotmdp thanks, LGTM

rbberger:
@spackbot run pipeline

PR #6211, opened 2026-08-24
bufr: use f2py build directory outside /tmp

mathomp4:
## Summary

With `+python` and Python 3.12 or newer, NumPy `f2py` uses Meson to build BUFR's Python extension. By default, f2py creates its build directory with Python's temporary-directory mechanism. On systems where `/tmp` is mounted `noexec`, Meson then cannot execute its compiler sanity-check program.

```text
meson.build:1:0: ERROR: Could not invoke sanity check executable:
[Errno 13] Permission denied:
'/tmp/.../meson-private/sanity_check_for_fortran.exe'
```

Pass f2py's `--build-dir` option from BUFR's CMake custom command, placing the temporary Meson project under CMake's binary directory instead of `/tmp`.

The change is limited to `bufr+python ^python@3.12:`, matching the f2py versions that use Meson.

## Testing

Successfully built on Linux:

```text
bufr@12.3.0+python ^python@3.13 %gcc@15.2.0
```

Before this change, f2py invoked Meson from `/tmp` and failed on its Fortran sanity check. With this change, BUFR builds successfully.

## AI assistance disclosure

GPT-5.6 assisted with identifying the f2py/Meson failure and drafting the package change. The change was reviewed and tested by the submitter.


mathomp4:
Note to @climbfuji: This was found in my test builds of my spack-stack testing. Fun with `noexec`!

mathomp4:
Converting to draft. Something isn't quite right. Still seeing a bufr build failure...though I thought it worked? Weird...

mathomp4:
Okay. Should be good to go. I really hate `f2py` sometimes.

repos/spack_repo/builtin/packages/bufr/package.py
```diff
@@ -89,9 +89,21 @@ def url_for_version(self, version):
...
             filter_file("_lenslmax 120", "_lenslmax 60", "CMakeLists.txt")
 
+        # NumPy f2py uses Meson on Python 3.12 and newer. Direct its
+        # temporary build directory to CMake's already-created binary tree,
+        # rather than the potentially noexec /tmp default.
+        if self.spec.satisfies("+python ^python@3.12:"):
+            filter_file(
+                '-c "${CMAKE_CURRENT_SOURCE_DIR}/_bufrlib.pyf"',
+                '-c "${CMAKE_CURRENT_SOURCE_DIR}/_bufrlib.pyf"\n'
+                '                   --build-dir "${CMAKE_CURRENT_BINARY_DIR}/f2py-build"',
```
alalazo: Wouldn't `--build-dir "${CMAKE_BINARY_DIR}/f2py-build"` be better (it's outside `<build>/python`)?

repos/spack_repo/builtin/packages/bufr/package.py
alalazo: Does this hold for every version listed?

alalazo (commented):
As an alternative, did you consider:
```python
    if self.spec.satisfies("+python"):
        env.set("TMPDIR", self.stage.path)
```
?

repos/spack_repo/builtin/packages/bufr/package.py
mathomp4: Ah. This makes sense. I'll update this.

repos/spack_repo/builtin/packages/bufr/package.py
mathomp4: I will look into this.

mathomp4: Ah yeah. bufr only moved to direct f2py invocation in CMake starting in 12.1.0. Versions 12.0.1 and older used the legacy setup.py / numpy.distutils build system.

mathomp4:
> As an alternative, did you consider:
> 
> ```python
> def setup_build_environment(self, env):
>     if self.spec.satisfies("+python"):
>         env.set("TMPDIR", self.stage.path)
> ```
> 
> ?

I actually started with that approach in an earlier commit, but ran into build failures with it in testing. Passing `--build-dir` directly to f2py proved much more reliable in my testing.

repos/spack_repo/builtin/packages/bufr/package.py
mathomp4: @alalazo I think I addressed this, can you take a look?

repos/spack_repo/builtin/packages/bufr/package.py
mathomp4: I think I addressed this, can you take a look?

PR #6212, opened 2026-08-24
cice: new package

WeirAE:
Add new package [CICE](https://github.com/CICE-Consortium/CICE)
Adding to the authoritative repo for broader testing, feedback and further development, with JCSDA repo to follow. Currently set up to follow [existing build guidance](https://cice-consortium-cice.readthedocs.io/en/main/intro/quickstart.html) creating a directory for users to run manually.

### DIFF
--- a/repos/spack_repo/builtin/packages/cice/package.py
+++ b/repos/spack_repo/builtin/packages/cice/package.py
@@ -36,7 +36,7 @@ class Cice(Package):
         default="src",
         description="Directory name where CICE is installed",
         values=str,
-        )
+    )
 
     depends_on("c", type="build")
     depends_on("fortran", type="build")
--- a/repos/spack_repo/builtin/packages/cice/package.py
+++ b/repos/spack_repo/builtin/packages/cice/package.py
@@ -1,0 +1,64 @@
+# Copyright Spack Project Developers. See COPYRIGHT file for details.
+#
+# SPDX-License-Identifier: (Apache-2.0 OR MIT)
+# ----------------------------------------------------------------------------
+
+from spack_repo.builtin.build_systems.generic import Package
+
+from spack.package import *
+
+
+class Cice(Package):
+    """CICE is a computationally efficient model for simulating the growth, melting,
+    and movement of polar sea ice."""
+
+    homepage = "https://github.com/CICE-Consortium/CICE"
+    url = "https://github.com/CICE-Consortium/CICE/archive/refs/tags/CICE6.6.3.tar.gz"
+
+    maintainers("apcraig")
+
+    license("BSD-3-Clause")
+
+    version("6.6.3", sha256="efc16625486b46f01409c568f532978690bb37cc12d8b8a121601ffe047e89f4")
+    version("6.6.2", sha256="1a265ca44a2d7182cd670ff2023d0ff98bedce7cff7c68fe374de9084ba96b39")
+    version("6.6.1", sha256="2b10591ea0052fe5a183679a4e9f6e083ddbb46537bfa765af0cbe8ed57d0a9e")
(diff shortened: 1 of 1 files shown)


cmelone:
@apcraig confirming you're signing up to be a maintainer for this package

cmelone (changes requested):
I have a few questions since this is a bit of an unconventional package:
- why are there build dependencies if this package is not compiled?
- spack typically manages the installation directory internally. if a user would like to modify the target directory, they can do this via the [`install_tree` config option](https://spack.readthedocs.io/en/latest/config_yaml.html)
- can you explain the architecture of this application? I can provide some guidance on how to do the packaging once I have more info

---

the `BundlePackage` pattern may be what you're looking for https://github.com/search?q=repo%3Aspack%2Fspack-packages+bundlepackage&type=code

WeirAE:
@cmelone 
Thank you - It is definitely a bit unconventional at the moment. Ah interesting, the BundlePackage may be the best way to go, I hadn't submitted a package like this before so I went with the default from what I'd seen before. I have also made a habit of enforcing requirements regardless of build, but if that's not necessary or helpful, I'm happy to take it out.
CICE standalone currently configures the necessary environment variables and components through manual invocation by the user when running the [cice.setup](https://github.com/CICE-Consortium/CICE/blob/main/cice.setup) script. When the UFS weather model calls it, it has an extensive process through [cmake](https://github.com/ufs-community/ufs-weather-model/blob/develop/CICE-interface/CMakeLists.txt). At this phase, I'd like to get the simplest comfiguration in the system and work out any more elaborate processes with the many involved stakeholders at a later date. Unfortunately, that does at least require returning the installed directory or assisting the user with the necessary steps for defining the target directory

apcraig:
Thanks for these efforts, but this is the first I've heard of it.  I can probably help, will need to spin up a bit on what needs to be done.  I see it's really just a script that says to run cice.setup, so it looks like we just need to maintain the release list.  Why can't spack see the CICE releases on github and automatically grab them from within the script?

tgamblin (changes requested):
Can you say a little more about why you want to add this to the main Spack repo?

The package as written just copies source into a directory outside the Spack prefix, which is a model we don't support. The job of a Spack package is to build the software then install *into the prefix spack gives you*.  This package takes a potentially external directory as a variant and just copies the source code there.

Some reasons this doesn't belong in the mainline as-is:
1. I think this is going to be very unexpected behavior for users.  They're going to expect `spack install cice` to install `cice` and to let you `spack load cice` and run it. This doesn't do that.
2. The user could pass any directory as an absolute path, including one outside the install prefix, to `target_dir`, so this package could write essentially anywhere in the filesystem. So Spack isn't really managing the package.
3. If the package isn't built I am not sure why it depends on `c`, `fortran`, `netcdf-fortran`, etc. 

If the point is to build dependencies and stage the source code so that someone can work on it, we have workflows for that -- see https://spack-tutorial.readthedocs.io/en/latest/tutorial_developer_workflows.html.  You can either:

```console
spack install --only dependencies
```

to get just the dependencies, or you could make an environment and `spack develop cice` to have it build the dependencies, stage the source, and iterate on `cice` yourself.

In any case we would want this package to actually run the `cice` build and install into `prefix` before we merge it into the mainline -- otherwise it's going to be very surprising for users.

WeirAE:
OK, thanks for the feedback, those are all very fair points. While the `BundlePackage` is a plausible option, it would be more accurate to ensure the build is actually performed. I think the best thing now is to work with Tony on any necessary changes on the CICE end that would make it feasible and in the meantime continue to test with `spack develop` off of my fork as I have been. Would you prefer to close or convert this to draft pending that work?

cmelone:
changing to draft status is fine

PR #6213, opened 2026-08-24
qt-base: harfbuzz 2.6 is required for all Qt6 versions listed

clintonstimpson:


wdconinc:
Not that I don't believe this but for future archaeologists can you provide a link to where the condition is enforced, maybe in the package, so it can be checked in future newer version bumps?

clintonstimpson:
https://github.com/qt/qtbase/blob/dev/src/gui/configure.cmake#L61
I've confirmed that on RHEL8, the Qt configure will fail if given an older version of harfbuzz.  This failure can happen in spack if one tries to provide an external harfbuzz.

wdconinc:
> https://github.com/qt/qtbase/blob/dev/src/gui/configure.cmake#L61 I've confirmed that on RHEL8, the Qt configure will fail if given an older version of harfbuzz. This failure can happen in spack if one tries to provide an external harfbuzz.

Thanks. Can we add the `freetype@2.2:` requirement as well in the line above? And add a source comment referring to `src/gui/configure.cmake`? I would leave a suggestion, but githut doesn't let me.

wdconinc (approved):

PR #6218, opened 2026-08-25
hipfort: add v7.14.0 

srekolam:


repos/spack_repo/builtin/packages/hipfort/package.py
RikkiButler20, line 58:
```suggestion
    depends_on("cmake@3.18:", type="build", when="@7.0:")
    depends_on("cmake@3.5:", type="build", when="@6.3.0")
    depends_on("cmake@3.0.2:", type="build")
```
Don't we need to update cmake's minimum required version?

RikkiButler20 (changes requested):

RikkiButler20 (approved):
LGTM

PR #6221, opened 2026-08-25
abacus: modernize to CMake build and add v3.9.0.19, v3.10.1

Bitllion:
- Migrate from MakefilePackage (ABACUS 2.2.x, Intel-only) to CMakePackage (ABACUS 3.x) with proper GNU/OpenMPI toolchain support.
- Add versions 3.10.1 and 3.9.0.19, drop 2.2.x line.
- Replace hardcoded Intel MKL dependency with openblas.
- Add lcao/elpa/libxc variants; set GIT_SUBMODULE=OFF for release tarballs.

Verified: 62 dependencies built cleanly; SCF example (pw_Si2) converged to -215.5056984116324 eV identically for both new versions.

### DIFF
--- a/repos/spack_repo/builtin/packages/abacus/package.py
+++ b/repos/spack_repo/builtin/packages/abacus/package.py
@@ -2,22 +2,19 @@
 #
 # SPDX-License-Identifier: (Apache-2.0 OR MIT)
 
-
-import re
-
-from spack_repo.builtin.build_systems.makefile import MakefilePackage
+from spack_repo.builtin.build_systems.cmake import CMakePackage
 
 from spack.package import *
 
 
-class Abacus(MakefilePackage):
+class Abacus(CMakePackage):
     """ABACUS (Atomic-orbital Based Ab-initio Computation at UStc)
     is an open-source computer code package aiming
     for large-scale electronic-structure simulations
     from first principles"""
 
     homepage = "http://abacus.ustc.edu.cn/"
-    url = "https://github.com/abacusmodeling/abacus-develop/archive/refs/tags/v2.2.1.tar.gz"
+    url = "https://github.com/abacusmodeling/abacus-develop/archive/refs/tags/v3.9.0.19.tar.gz"
     git = "https://github.com/abacusmodeling/abacus-develop.git"
 
     maintainers("bitllion")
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/abacus/package.py
```diff
@@ -2,101 +2,52 @@
...
-    url = "https://github.com/abacusmodeling/abacus-develop/archive/refs/tags/v2.2.1.tar.gz"
+    url = "https://github.com/abacusmodeling/abacus-develop/archive/refs/tags/v3.9.0.19.tar.gz"
     git = "https://github.com/abacusmodeling/abacus-develop.git"
 
     maintainers("bitllion")
 
     license("LGPL-3.0-or-later")
 
     version("develop", branch="develop")
-    version("2.2.3", sha256="88dbf6a3bdd907df3e097637ec8e51fde13e2f5e0b44f3667443195481320edf")
```
kshea21, line 28: Could you add deprecation to these versions instead of removing them? That's the policy we follow with packages. https://spack.readthedocs.io/en/latest/packaging_guide_creation.html#deprecate

repos/spack_repo/builtin/packages/abacus/package.py
```diff
@@ -2,101 +2,52 @@
...
-import re
-
-from spack_repo.builtin.build_systems.makefile import MakefilePackage
+from spack_repo.builtin.build_systems.cmake import CMakePackage
 
 from spack.package import *
 
 
-class Abacus(MakefilePackage):
+class Abacus(CMakePackage):
```
kshea21: You can add support for both CMakePackage and MakefilePackage, and we'll have to have at least one round of version releasing that has the Makefile version deprecated before we can remove this support entirely.

kshea21 (changes requested):

Bitllion:
Both review points are now addressed in the latest commit
(63da326f4b -> e2fb8ee188):

1. The 2.2.x versions are kept with `deprecated=True` instead of
   being removed.

2. The package now supports both build systems via a `build_system`
   variant -- CMake for 3.x, and the legacy Intel Makefile path
   (in a `MakefileBuilder`) for the deprecated 2.2.x line.

Verified:
  - `spack spec abacus` -> 3.10.1 + build_system=cmake
  - `spack spec --deprecated abacus@2.2.3` -> makefile path
  - ruff check + format pass

kshea21 (approved):
Great, thanks for making the change and contributing!

PR #6224, opened 2026-08-25
yaml-cpp: add v0.9.0

jcortial-safran:
Release notes: https://github.com/jbeder/yaml-cpp/releases#release-yaml-cpp-0.9.0

An actual release archive was published for 0.9.0, so we use it instead of the archive associated with the tag.

The patch does not need to be applied to 0.9.0.

```
$ spack find yaml-cpp
-- linux-debian13-x86_64_v4 / %c,cxx=gcc@14.2.0 -----------------
yaml-cpp@0.7.0  yaml-cpp@0.8.0  yaml-cpp@0.9.0
==> 3 installed packages
```


jcortial-safran:
@eschnett: Any feedback on the PR ?

cmelone (approved):

PR #6226, opened 2026-08-25
py-tabulate: add v0.10.0

RiMillo:
Missing dependency on `setuptools-scm` (see [`pyproject.toml`](https://github.com/astanin/python-tabulate/blob/v0.9.0/pyproject.toml#L2)) led to unknown version, i.e., directory `tabulate-0.0.0.dist-info` and `Version: 0.0.0` in `METADATA`.

Left a `TODO` comment to notify that in master they switch to `flit` as build back-end (instead of `setuptools`).

### DIFF
--- a/repos/spack_repo/builtin/packages/py_tabulate/package.py
+++ b/repos/spack_repo/builtin/packages/py_tabulate/package.py
@@ -26,5 +26,7 @@ class PyTabulate(PythonPackage):
     version("0.7.7", sha256="83a0b8e17c09f012090a50e1e97ae897300a72b35e0c86c0b53d3bd2ae86d8c6")
 
     # TODO: it moved to flit[-scm] after 0.10.0
+    depends_on("py-setuptools@77.0.3:", type="build", when="@0.10")
+    depends_on("py-setuptools@61.2:", type="build", when="@0.9")
     depends_on("py-setuptools", type="build")
     depends_on("py-setuptools-scm@3.4.3: +toml", type="build", when="@0.9:0.10")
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/py_tabulate/package.py
```diff
@@ -24,4 +25,6 @@ class PyTabulate(PythonPackage):
     version("0.8.3", sha256="8af07a39377cee1103a5c8b3330a421c2d99b9141e9cc5ddd2e3263fea416943")
     version("0.7.7", sha256="83a0b8e17c09f012090a50e1e97ae897300a72b35e0c86c0b53d3bd2ae86d8c6")
 
+    # TODO: it moved to flit[-scm] after 0.10.0
     depends_on("py-setuptools", type="build")
```
RikkiButler20, line 31:
```suggestion
    # TODO: it moved to flit[-scm] after 0.10.0
    depends_on("py-setuptools@77.0.3", type="build", when="@0.10:)
    depends_on("py-setuptools@61.2", type="build", when="@0.9)
    depends_on("py-setuptools", type="build")
```

RikkiButler20 (changes requested):
Update version constraints for `py-setuptools`

repos/spack_repo/builtin/packages/py_tabulate/package.py
RiMillo: Thanks, I changed a bit wrt yours, namely relaxed to have "greater-than" in `setuptool` version, and fixed version `0.10` since after that they moved to another build back-end

RikkiButler20 (approved):
Looks good. Thanks!

PR #6229, opened 2026-08-26
pi: add v0.84.3, remove <= v0.83

alecbcs:
https://github.com/earendil-works/pi/releases/tag/v0.84.3

This PR purposely removes v0.79.6 -> v0.83.0 (which didn't support offline installation) instead of deprecating them since this package hasn't yet made it into a packages release and as far as I know I'm the only user of pi. Removing these versions will reduce the maintenance version and package complexity.


RikkiButler20 (approved):
LGTM

PR #6233, opened 2026-08-26
mdspan: refactoring

tpadioleau:
- favor usage of `--test root` instead of having a dedicated `tests` variant
- remove the variant `examples`, nothing is installed when enabling the variant
- enforce `MDSPAN_INSTALL_STDMODE_HEADERS=OFF`

I also wonder if the variant `benchmarks` is useful. If we keep it I think we could make the dependency only `build`, as nothing is installed, we shouldn't need to preserve the dependency after the installation of the package.


alalazo (approved):
Removing variant may be breaking, but for the time being we have no better way to deprecate them 🙁 PR LGTM.

PR #6234, opened 2026-08-26
libxsmm: fix macOS dylib install names

Sbozzolo:
The GNU Make build on macOS embeds the build-tree path in each shared library's `LC_ID_DYLIB`. Consumers therefore record a path under the Spack stage, which disappears after installation and causes runtime failures such
as:
```
Library not loaded: .../spack-stage-libxsmm-2.1.0-.../spack-src/lib/libxsmm.2.dylib
```

Rewrite the IDs of installed, non-symlink dylibs to their final paths with `install_name_tool`. Linux is unaffected because it uses sonames and rpaths.

The failure was reproduced in: https://github.com/awslabs/palace/actions/runs/32965751532/job/98167779439

(The string-format changes are mechanical updates required by the current Ruff baseline when modifying this recipe.)

### DIFF
--- a/repos/spack_repo/builtin/packages/libxsmm/package.py
+++ b/repos/spack_repo/builtin/packages/libxsmm/package.py
@@ -155,13 +155,13 @@ class MakefileBuilder(makefile.MakefileBuilder):
     def build(self, pkg, spec, prefix):
         # include symbols by default
         make_args = [
-            "CC={0}".format(spack_cc),
-            "CXX={0}".format(spack_cxx),
-            "FC={0}".format(spack_fc),
-            "PREFIX=%s" % prefix,
+            f"CC={spack_cc}",
+            f"CXX={spack_cxx}",
+            f"FC={spack_fc}",
+            f"PREFIX={prefix}",
         ]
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/libxsmm/package.py
```diff
@@ -195,6 +195,17 @@ def install(self, pkg, spec, prefix):
         # always install libraries
         install_tree("lib", prefix.lib)
 
+        # On macOS the Makefile build bakes the build-tree path into each
```
alalazo: Can you remind me why we aren't using the `Makefile` targets here? It seems it deals with this at:

https://github.com/libxsmm/libxsmm/blob/55a8fa6a1e479dec1f5ddbe20684c1cdc0ff7eb1/Makefile#L1107-L1117

Sbozzolo: The `Makefile` does not include this for all the `1.x` versions. I updated to take this into account.

alalazo (approved):

PR #6236, opened 2026-08-26
petsc: add thrust/tuple.h include to curand2.cu for CUDA 13.3

Sbozzolo:
Under CUDA 13.3, thrust headers no longer pull `<thrust/tuple.h>` in transitively. PETSc's `src/sys/classes/random/impls/curand/curand2.cu` uses `thrust::tuple/get/make_tuple`, so nvcc fails with 'namespace "thrust" has no member "tuple"'. Add the missing include via a patch gated on +cuda ^cuda@13.3:. Mirrors the upstream PETSc fix in [MR!9321](https://gitlab.com/petsc/petsc/-/merge_requests/9321), which added the same include to `aijcupm.hpp` but not to `curand2.cu`.

Unfortunately, I cannot upstream this patch (because I cannot open a GitLab account), but I would be thrilled if a PETSc maintainer reading this applied the fix upstream.

Failing build: https://github.com/awslabs/palace/actions/runs/32969345089
Working build (with the patch): https://github.com/awslabs/palace/actions/runs/32975197965/job/98198161666

### DIFF
--- a/repos/spack_repo/builtin/packages/petsc/package.py
+++ b/repos/spack_repo/builtin/packages/petsc/package.py
@@ -231,6 +231,13 @@ class Petsc(Package, CudaPackage, ROCmPackage):
     patch("petsc_modifiable_lvalue.patch", when="@3.21.6:3.22.4+rocm")
     patch("petsc_modifiable_lvalue.patch", when="@3.21.6:3.22.4+cuda")
 
+    # fixes build with: +complex ^cuda@13.3. Upstream fix: petsc!9532.
+    patch(
+        "https://gitlab.com/petsc/petsc/-/commit/c0f7467a2261011568d510ece23f14cad8dcaaa4.diff",
+        sha256="e91c9b9323f22fe8988f5707eb262290e850b81262cf41557dc552848313a6b8",
+        when="@3.16:3.25.5 +cuda +complex ^cuda@13.3:",
+    )
+
     # These require +mpi
     mpi_msg = "Requires +mpi"
     conflicts("+cgns", when="~mpi", msg=mpi_msg)


jczhang07 (approved):

balay:
Hm - PETSc CI runs fine with cuda-13.3/13.4. Perhaps this issue is triggered with `+complex`?  (thats not tested in CI?)

If so - probably best to add the patch to PETSc sources - first, and use it here (and avoid a duplicate patch)

balay:
MR to PETSc at: https://gitlab.com/petsc/petsc/-/merge_requests/9532/

Sbozzolo:
Ah yeah, you are right. There is a `#if defined(PETSC_USE_COMPLEX)` guard

repos/spack_repo/builtin/packages/petsc/package.py
```diff
@@ -231,6 +231,22 @@ class Petsc(Package, CudaPackage, ROCmPackage):
...
+    # curand2.cu uses thrust::tuple/get/make_tuple but only includes
+    # thrust/transform.h etc. CUDA 13.3 (CCCL 3.3) stopped pulling
+    # <thrust/tuple.h> in transitively, so nvcc fails with
+    # 'namespace "thrust" has no member "tuple"'. Upstream fix: petsc!9532.
+    # The tuple uses are guarded by PETSC_USE_COMPLEX, so only +complex builds
+    # trip it (which is why PETSc's own CUDA CI, built without complex, is
+    # unaffected). The lower bound is @3.16: because that is where curand2.cu
+    # started including <petsc/private/randomimpl.h> (the context this patch
+    # matches); no upper bound yet as no PETSc release ships the include.
+    # Remove once a fixed PETSc release is available.
```
balay: suggest removing this comment here

repos/spack_repo/builtin/packages/petsc/package.py
```diff
@@ -231,6 +231,22 @@ class Petsc(Package, CudaPackage, ROCmPackage):
...
+    patch(
+        "https://gitlab.com/petsc/petsc/-/commit/c0f7467a2261011568d510ece23f14cad8dcaaa4.diff",
+        sha256="e91c9b9323f22fe8988f5707eb262290e850b81262cf41557dc552848313a6b8",
+        when="@3.16: +cuda +complex ^cuda@13.3:",
```
balay: [once merged] - the fix will be in @3.25.5 - so perhaps: `when="@3.16:3.25.5"`

BTW: I'm not sure if any of those old versions work with these newer versions of cuda anyway..

repos/spack_repo/builtin/packages/petsc/package.py
```diff
@@ -231,6 +231,19 @@ class Petsc(Package, CudaPackage, ROCmPackage):
...
+    # curand2.cu uses thrust::tuple/get/make_tuple but only includes
+    # thrust/transform.h etc. CUDA 13.3 (CCCL 3.3) stopped pulling
+    # <thrust/tuple.h> in transitively, so nvcc fails with
+    # 'namespace "thrust" has no member "tuple"'. Upstream fix: petsc!9532.
+    # The tuple uses are guarded by PETSC_USE_COMPLEX, so only +complex builds
+    # trip it (which is why PETSc's own CUDA CI, built without complex, is
+    # unaffected).
+    patch(
+        "https://gitlab.com/petsc/petsc/-/commit/c0f7467a2261011568d510ece23f14cad8dcaaa4.diff",
+        sha256="e91c9b9323f22fe8988f5707eb262290e850b81262cf41557dc552848313a6b8",
```
balay: Sorry I mean to say:  we could remove this comment completely. [as its documented in the commit message, and the patch]. 

But we we are to have one - I guess a shorter one like:

```suggestion
    # fixes build with: +complex ^cuda@13.3. Upstream fix: petsc!9532.
```

balay:
for some reason - my suggested fix also removed a couple of additional lines :(. I fixed the branch and pushed now.

balay (approved):

cmelone (approved):

PR #6237, opened 2026-08-26
repo: do not mutate Spec objects

haampie:
Fix a couple issues with packages that mutate `self.spec` or use some internals I would consider not covered by the package API.

`nek5000`, `nektools`, `nekcem` and `stripack` appended to `spec.compiler_flags[...]`

`lci` assigned to `spec.variants["bootstrap"].value`.

`boost` read the private `spec.variants.dict`.

`mfem` called `.copy()` and `berkeleygw` sliced with `[:]` on flags, which works assuming it's a list, but this to me sounds like assuming too much about the implementation and will fail if compiler flags are immutable tuples
(https://github.com/spack/spack/pull/52947).

Use `flag_handler` instead of something ad-hoc in `strippack`, `nektools`, `berkeleygw`

Use the compiler wrapper `spack_fc` instead of the backing compiler in `stripack`.

Drop the custom logic for X11 include paths, cause it coincides with the default `-I` flags added through the compiler wrapper.

Branch on `%fortran=<name>` in `nek5000`, `nektools` and `nekcem` instead of something custom.

The builds of berkeleygw, boost, lci, mfem, nek5000, nekcem, nektools and stripack succeeded using gcc 13.3.0.

### DIFF
--- a/repos/spack_repo/builtin/packages/berkeleygw/package.py
+++ b/repos/spack_repo/builtin/packages/berkeleygw/package.py
@@ -183,14 +183,11 @@ class Berkeleygw(MakefilePackage):
         if spec.satisfies("+mpi"):
             paraflags.append("-DMPI")
 
-        # We need to copy fflags in case we append to it (#34019):
-        cflags = spec.compiler_flags["cflags"][:]
-        fflags = spec.compiler_flags["fflags"][:]
-        cflags.append(self.compiler.cc_pic_flag)
-        fflags.append(self.compiler.fc_pic_flag)
+        cflags = [spec["c"].package.pic_flag]
+        fflags = [spec["fortran"].package.pic_flag]
         if spec.satisfies("+openmp"):
             paraflags.append("-DOMP")
-            fflags.append(self.compiler.openmp_flag)
+            fflags.append(spec["fortran"].package.openmp_flag)
 
         if spec.satisfies("+mpi"):
             buildopts.append("C_PARAFLAG=-DPARA")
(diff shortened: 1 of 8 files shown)


hainest (approved):
Boost changes LGTM.

sethrj (approved):

repos/spack_repo/builtin/packages/berkeleygw/package.py
```diff
@@ -183,14 +183,11 @@ def build(self, spec, prefix):
         if spec.satisfies("+mpi"):
             paraflags.append("-DMPI")
 
-        # We need to copy fflags in case we append to it (#34019):
-        cflags = spec.compiler_flags["cflags"][:]
-        fflags = spec.compiler_flags["fflags"][:]
-        cflags.append(self.compiler.cc_pic_flag)
-        fflags.append(self.compiler.fc_pic_flag)
+        cflags = [spec["c"].package.pic_flag]
+        fflags = [spec["fortran"].package.pic_flag]
```
migueldiascosta, line 187: @haampie doesn't this change the append behaviour?

haampie: The `spec.compiler_flags["cflags"]` value corresponds to `spack install foo cflags=-O3`. Those flags are injected through the compiler wrapper, a package doesn't have to handle them. So if you set CFLAGS=-x as an env variable or in a makefile, and `spack install cflags=-y` on the command line, you get both `-x` and `-y`.

migueldiascosta (approved):
lgtm on berkeleygw

PR #6238, opened 2026-08-26
mille: add v01-01-01

goblirsc:
Add version 01-01-01 of Mille

### DIFF
--- a/repos/spack_repo/builtin/packages/mille/package.py
+++ b/repos/spack_repo/builtin/packages/mille/package.py
@@ -16,6 +16,7 @@ class Mille(CMakePackage):
     git = "https://gitlab.desy.de/millepede/mille.git"
 
     license("LGPL-2.0-only", checked_by="paulgessinger")
+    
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/mille/package.py
```diff
@@ -16,8 +16,8 @@ class Mille(CMakePackage):
     git = "https://gitlab.desy.de/millepede/mille.git"
 
     license("LGPL-2.0-only", checked_by="paulgessinger")
-
```
RikkiButler20, line 19: Can you bring the space back

RikkiButler20 (changes requested):

repos/spack_repo/builtin/packages/mille/package.py
goblirsc: yep, should hopefully again be consistent now

RikkiButler20 (approved):
Looks good. Thanks

PR #6243, opened 2026-08-26
emacs: remove unused dependencies, prevent linking to system deps

alecbcs:
Clean up the Emacs build to remove unused dependencies, prevent silently linking with system dependencies, and limit test execution for develop builds without a stable version.


kwryankrattiger (approved):
LGTM

PR #6245, opened 2026-08-26
Revert "ci: add nightly job to sync source mirror with pkgs in builtin"

tgamblin:
Reverts spack/spack-packages#6069

Job is failing with:

```
==> Installing "clingo-bootstrap@=spack~apps~docs+ipo+optimized+python+static_libstdcpp build_system=cmake build_type=Release commit=2a025667090d71b2c9dce60fe924feb6bde8f667 generator=make patches:=bebb819,ec99431 platform=linux os=centos7 target=x86_64" from a buildcache
==> Compilers have been configured automatically from PATH inspection
==> Starting concretization pool with 4 processes
==> Error: failed to concretize `acfl@=24.04` for the following reasons:
       1. acfl: Only available on Aarch64
  Analyzing the cause of the failure, this may take a moment...
==> Error:      2. acfl: Only available on Aarch64
        required because conflict constraint
          required because acfl@=24.04 requested explicitly
        required because conflict is triggered when target=x86_64:
          required because acfl@=24.04 requested explicitly
==> Error: failed to concretize `acfl@=24.04` for the following reasons:
     1. acfl: Only available on Aarch64
     2. acfl: Only available on Aarch64
        required because conflict constraint
          required because acfl@=24.04 requested explicitly
        required because conflict is triggered when target=x86_64:
          required because acfl@=24.04 requested explicitly
Some sources failed to fetch; uploading the rest
Nothing was mirrored
```

which implies several things I missed in review:
1. The sync job shouldn't need to concretize; we know tarball URLs for versions statically and can add them to the mirror that way. 
2. `acfl` really needs a `redistribute(source=False, binary=False)` directive (probably along with some of the other missing packages, as it's proprietary software that we cannot mirror.


alecbcs (approved):

alecbcs:
I'm happy to proceed with the revert here if that's what folks want to do.

> The sync job shouldn't need to concretize; we know tarball URLs for versions statically and can add them to the mirror that way.

This workflow isn't just adding the source tarballs. It's also picking up the required patches and resources that apply to missing versions. The concretization comes from `spack mirror create` which has always used a concretization to determine which of those additional artifacts need to be pulled into a mirror. (Although I think you're right that we could instead determine which patches and resources to mirror via static analysis instead.)

> acfl really needs a redistribute(source=False, binary=False) directive (probably along with some of the other missing packages, as it's proprietary software that we cannot mirror

Agreed. This workflow is a good canary in the coal mine to finding binary packages and non-redistributable packages that have made it into the repository without being correctly labeled. As such I kinda lean keeping it in and fixing the packages that pop up as problems over time.

tgamblin:
Would rather revert than have this job run needlessly.  GitHub clearly needs the help :).

PR #6246, opened 2026-08-27
openblas: backport OpenMP nested-region deadlock fix

Sbozzolo:
Backport https://github.com/OpenMathLib/OpenBLAS/pull/5949 to 0.3.34, to fix a bug related to OpenMP introduced in 0.3.34

Found updating OpenBlas here: https://github.com/awslabs/palace/actions/runs/32968723827/job/98179740687?pr=865


mathomp4 (approved):

sethrj (approved):
I see that most of the patches in OpenBLAS use "commit" patches, but it's shorter and more self-documenting to use the PR itself:

```suggestion
        "https://github.com/OpenMathLib/OpenBLAS/pull/5949.patch?full_index=1",
```

This includes the author and commit information rather than just the diff so it has a different hash. I'll approve and merge but suggest using the above format in the future. Thanks!

PR #6248, opened 2026-08-27
ci: sync source mirror statically, without concretizing

tgamblin:
Re-adds the nightly source mirror sync from #6069 (reverted in #6245).

Nothing about mirroring sources requires concretization: URLs and sha256 checksums for version tarballs, resources, and patches are all known statically from `package.py` files. The `sync-src-mirror.py` script here:

1. finds every sha256-addressed artifact in the builtin repo that is missing from the mirror, by comparing digests against a listing of the content-addressed `_source-cache/archive/` prefix:
   - version tarballs
   - resources
   - URL patches (in packages and from dependencies, via the patch index)

   while skipping manual-download and non-redistributable packages;

2. fetches each missing artifact with Spack's own fetch strategies, which honor per-package `fetch_options`, fallback URLs, and checksum verification; and

3. uploads each verified artifact to S3 with `aws s3 cp`, one at a time, so disk usage stays bounded.

Mirror paths are computed with Spack's own `default_mirror_layout()`, so uploaded files are found by `spack fetch`. Individual fetch failures are tolerated and only skip that artifact until the next nightly run.

This behaves like `spack mirror create --all`, in that all resources and patches are included regardless of their `when=` conditions -- the mirror should hold artifacts for every possible configuration.

Parts of this should probably be integrated into `spack mirror create` eventually; it lives in spack-packages for now because `spack mirror create --all` still concretizes when run from an environment, and it's hard to write an environment that fetches only artifacts for specific package versions without reconcretizing. Once that's fixed, this script can be replaced.

The mirror is currently missing 587 artifacts, which this job will backfill on its first runs -- modulo dead URLs and stale checksums (e.g. `abduco@0.4`/`@0.5`, whose autogenerated GitHub tarballs no longer match their declared sha256s). #6247 marks the packages whose sources we can't legally mirror so we don't retry them every night.

### DIFF
--- a/.github/workflows/bin/sync-src-mirror.py
+++ b/.github/workflows/bin/sync-src-mirror.py
@@ -1,35 +1,48 @@
+#!/usr/bin/env spack-python
+#
 # Copyright Spack Project Developers. See COPYRIGHT file for details.
 #
 # SPDX-License-Identifier: (Apache-2.0 OR MIT)
 
-"""Print one line per source artifact (version tarball, resource, or patch)
-that is missing from the source mirror -- without concretizing anything.
+"""Sync the source mirror with the package repository -- without concretizing.
 
-Usage: spack python find-missing-mirror-artifacts.py <sha256-file>
+Usage: sync-src-mirror.py <sha256-file> [--upload-to s3://bucket]
 
 ``<sha256-file>`` contains one sha256 digest per line, obtained by listing the
 content-addressed ``_source-cache/archive/`` prefix of the mirror.
 
-Each output line is tab-separated::
+This script:
+ 1. finds every source artifact (version tarball, resource, or patch)
+    that is missing from the mirror;
+ 2. Fetches it with spack's own fetch strategies that honor per-package
+    ``fetch_options``, mirrors, and checksum verification; and
+ 3. Uploads each verified artifact to the mirror with
+   ``aws s3 cp``.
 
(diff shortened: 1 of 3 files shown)


.github/workflows/bin/sync-src-mirror.py
```diff
@@ -0,0 +1,246 @@
...
+) -> None:
+    """Fetch one artifact, verify its checksum, and upload it to the mirror.
+
+    The stage (and the downloaded file with it) is destroyed on exit either
+    way, so disk usage stays bounded to one artifact at a time.
+    """
+    with spack.stage.Stage(fetcher) as stage:
+        stage.fetch()
+        stage.check()
+        aws("s3", "cp", stage.archive_file, f"{s3_url}/{path}", "--no-overwrite", "--no-progress")
```
alecbcs, line 197: Running the aws cp command within the stage's lifecycle is a nice idea to save on disk usage!

alecbcs (approved):
Looks good to me. It took me a while to figure out that not including an `upload_to` flag also disables downloading and staging the packages as more of a dry run and only prints what's missing. 

Overall I like the structure and the ability to upload the files within the stage's lifecycle to save on disk space consumed.

We might need to be careful that this now uses more of the Spack python SDK without being tied into Spack's unit tests so it's possible that we'll modify Spack and accidentally break this script.

PR #6254, opened 2026-08-27
tdls: new package

trsxvz:
Adds a package for TDLS (Tiny Device-callable Linear Solvers), a header-only C++20 library of direct solvers for small dense linear systems: https://github.com/trsxvz/TDLS. Tracks the main branch for now; released versions will follow.

### DIFF
--- a/repos/spack_repo/builtin/packages/tdls/package.py
+++ b/repos/spack_repo/builtin/packages/tdls/package.py
@@ -23,7 +23,6 @@ class Tdls(CMakePackage):
 
     version("main", branch="main")
 
-    depends_on("cxx", type="build")
     depends_on("cmake@3.21:", type="build")
 
     def cmake_args(self):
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/tdls/package.py
```diff
@@ -0,0 +1,33 @@
...
+    homepage = "https://trsxvz.github.io/TDLS/"
+    git = "https://github.com/trsxvz/TDLS.git"
+
+    maintainers("trsxvz")
+
+    license("BSD-3-Clause", checked_by="trsxvz")
+
+    version("main", branch="main")
+
+    depends_on("cxx", type="build")
```
cmelone: if this is header-only, there is no need to depend on a compiler

trsxvz: Right, thanks. project() enabled CXX unconditionally on the TDLS side, hence the dependency. Fixed upstream (the compiler is now only enabled for tests and examples) and dropped here.

cmelone (approved):

PR #6264, opened 2026-08-28
podio: Add version 1.8 and new parquet variant

tmadlener:
[release notes](https://github.com/AIDASoft/podio/releases/tag/v01-08)

I am not entirely sure if the new `parquet` variant could be done better, e.g. we could simply enable parquet support if we have `%arrow+parquet`?

Opinions @wdconinc?

### DIFF
--- a/repos/spack_repo/builtin/packages/podio/package.py
+++ b/repos/spack_repo/builtin/packages/podio/package.py
@@ -130,7 +130,7 @@ class Podio(CMakePackage):
             self.define_from_variant("ENABLE_ARROW", "arrow"),
             self.define_from_variant("ENABLE_RNTUPLE", "rntuple"),
             self.define_from_variant("ENABLE_DATASOURCE", "datasource"),
-            self.define_from_variatn("ENABLE_PARQUET", "parquet"),
+            self.define_from_variant("ENABLE_PARQUET", "parquet"),
             self.define("PODIO_SET_RPATH", True),
             self.define("CMAKE_CXX_STANDARD", self.spec.variants["cxxstd"].value),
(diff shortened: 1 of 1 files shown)


wdconinc (dismissed):
I think this is fine. The support is substantially added in podio and it's not like we automatically pick it up through the arrow headers, so a dedicated variant makes sense.

wdconinc:
FYI, fails build at https://gitlab.spack.io/spack/spack-packages/-/jobs/24120924 
```
AttributeError: 'super' object has no attribute 'define_from_variatn'
```

repos/spack_repo/builtin/packages/podio/package.py
```diff
@@ -122,6 +130,7 @@ def cmake_args(self):
             self.define_from_variant("ENABLE_ARROW", "arrow"),
             self.define_from_variant("ENABLE_RNTUPLE", "rntuple"),
             self.define_from_variant("ENABLE_DATASOURCE", "datasource"),
+            self.define_from_variatn("ENABLE_PARQUET", "parquet"),
```
wdconinc:
```suggestion
            self.define_from_variant("ENABLE_PARQUET", "parquet"),
```

OT @kwryankrattiger Can we catch language issues in GitHub CI static analysis before sending to gitlab?

wdconinc (approved):

cmelone (approved):

repos/spack_repo/builtin/packages/podio/package.py
cmelone: we'll need a type checker for that 🙂

PR #6265, opened 2026-08-28
ecbuild: add 3.15.2

mathomp4:
This adds ecbuild v3.15.2. I mainly care because it [fixed a warning](https://github.com/ecmwf/ecbuild/issues/150) ecbuild would emit when using CMake 4.4+

The warning doesn't break anything, but as I have a fix I need coming in CMake 4.5, well, I want to spare my users extra warnings at CMake time!


climbfuji (approved):
I looked at the diffs between 3.13.1 and 3.15.2, and I am not seeing any new cmake configuration options that would need to be added here. 

Thanks!

cmelone (approved):

PR #6268, opened 2026-08-28
edm4hep: patch for new podio arrow targets

wdconinc:
This PR updates `edm4hep` to use the new `podio +arrow` targets, which ended up landing with a different name than what made it into the 1.1 release. See https://github.com/key4hep/EDM4hep/pull/509.


jmcarcell (approved):
Necessary to configure with the latest tag of podio.

wdconinc:
@copilot resolve the merge conflicts in this pull request

wdconinc:
Now unnecessary after #6357.

PR #6284, opened 2026-08-30
boost: add v1.92.0, v1.91.0

wdconinc:
This PR adds `boost`, v1.91.0 ([release notes](https://www.boost.org/releases/1.91.0/)) and v1.92.0 ([release notes](https://www.boost.org/releases/1.92.0/)). 

Notes for 1.91.0:
- StaticAssert: This component was merged in Config, both header-only. 
- Decimal: This new component is header-only.
- System: "A CMake config file is now installed, even though the library is header-only."
  This will fix a bunch of packages that have `find_package(Boost REQUIRED system)`.

(There was a 1.91.0-1 release tag without artifacts, and it only differs in a CI yml file.)

Notes for 1.92.0:
- "Support requesting header-only libraries as `find_package` components"
  This should now allow `find_package(Boost REQUIRED system)` to work for all header-only components (or library components that become header-only).

Relevant upstream commit for the changed patch (it is literally just a change from 'm' to 't' in the patch): https://github.com/boostorg/build/commit/125be3e86c24257a749c1acb7f0335f6cf5e7475. Technically, this may make the patch unnecessary in some cases, but the patch introduces a more robust approach (e.g. it handles all ABI flags, including 'd', which the upstream commit will still fail to catch).

Test build:
```
ejhvr2l boost@1.92.0~atomic~charconv~chrono~clanglibcpp~cobalt~container~context~contract~conversion~date_time~debug~exception~fiber+filesystem~graph~graph_parallel+icu+iostreams~json+locale+log+math~mpi~mqtt5+multithreaded+nowide~numpy~openmethod~pic+program_options+python~random~regex~serialization+shared~signals2~singlethreaded~stacktrace+system~taggedlayout+test+thread~timer~type_erasure~url~versionedlayout~wave build_system=generic cxxstd=20 patches:=a440f96 visibility=hidden
```

### DIFF
--- a/repos/spack_repo/builtin/packages/boost/package.py
+++ b/repos/spack_repo/builtin/packages/boost/package.py
@@ -31,6 +31,8 @@ class Boost(Package):
     license("BSL-1.0")
 
     version("develop", branch="develop", submodules=True)
+    version("1.92.0", sha256="5c1d40cb8e19adbf740a4ec2da35b3e58f3f5804b1dce44deb53df72193cbc6c")
+    version("1.91.0", sha256="de5e6b0e4913395c6bdfa90537febd9028ea4c0735d2cdb0cd9b45d5f51264f5")
     version("1.90.0", sha256="49551aff3b22cbc5c5a9ed3dbc92f0e23ea50a0f7325b0d198b705e8ee3fc305")
     version("1.89.0", sha256="85a33fa22621b4f314f8e85e1a5e2a9363d22e4f4992925d4bb3bc631b5a0c7a")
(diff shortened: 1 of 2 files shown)


rbberger (commented):
Another relevant change is that since 1.91 the following conflict is too strict, since it is no longer is a compilation error but a runtime error:

https://github.com/spack/spack-packages/blob/6742e591c9cd08e8866c4777cacec96e0aa8663c/repos/spack_repo/builtin/packages/boost/package.py#L339-L344

See https://github.com/boostorg/stacktrace/commit/77647f8b0640ece6135db1c72f4ab5a345a230e2

PR #6288, opened 2026-08-31
`gettext`: disable modula-2

tgamblin:
Gettext has optional support for D and modula2. We already disable D; disable modula2 as well, at least until someone says we need to have it.


michaelkuhn (approved):
LGTM

PR #6289, opened 2026-08-31
parallelio: add pkgconfig build dependency

mathomp4:
ParallelIO 2.6.9 added a pkg-config fallback while locating NetCDF C. When `find_package(NetCDF ...)` does not succeed, CMake calls `find_package(PkgConfig REQUIRED)`.

The Spack package did not declare this build tool, so builds on hosts without a system `pkg-config` fail during configuration with:

```
  Could NOT find PkgConfig (missing: PKG_CONFIG_EXECUTABLE)
```

Add Spack's `pkgconfig` virtual build dependency for versions 2.6.9 and newer. The virtual lets Spack choose the configured provider (normally `pkgconf`).


RikkiButler20 (approved):
LGTM @jedwards4b Any feedback?

PR #6294, opened 2026-08-31
axom: add v0.15.0

bmhan12:
This PR:

Updates Axom recipe for 0.15.0 changes

### DIFF
--- a/repos/spack_repo/builtin/packages/axom/package.py
+++ b/repos/spack_repo/builtin/packages/axom/package.py
@@ -99,6 +99,7 @@ class Axom(CachedCMakePackage, CudaPackage, ROCmPackage):
 
     version("main", branch="main")
     version("develop", branch="develop")
+    version("0.15.0", tag="v0.15.0", commit="da2a50a7ee661896400d49b019e17bbe7ab5bd44")
     version("0.14.0", tag="v0.14.0", commit="146c8c15386a810791b7ab5c7fcb288cadea6151")
     version("0.13.0", tag="v0.13.0", commit="d00f6c66ef390ad746ae840f1074d982513611ac")
     version("0.12.0", tag="v0.12.0", commit="297544010a3dfb98145a1a85f09f9c648c00a18c")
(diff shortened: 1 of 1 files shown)


bmhan12:
Tags: @white238 , @becker33

white238 (approved):

repos/spack_repo/builtin/packages/axom/package.py
```diff
@@ -321,12 +325,12 @@ class Axom(CachedCMakePackage, CudaPackage, ROCmPackage):
...
     # Python
     with when("+python"):
-        depends_on("python")
+        depends_on("python@3.9:")
 
         # extending python allows spack environment views to import axom from python
         extends("python")
 
-        depends_on("py-nanobind@2.7.0:")
+        depends_on("py-nanobind@2.10.0:")
```
alecbcs, line 333: @bmhan12 were the previous python and py-nanobind requirements here incorrect? If they were added for v0.15.0 can we add them in addition to the previous requirements with matching `when=` clauses?

alecbcs (commented):
One quick question regarding +python variant dependency constraints. 

Otherwise looks good to me.

repos/spack_repo/builtin/packages/axom/package.py
bmhan12: Yes, the previous versions were inaccurate, they were arbitrarily decided/undecided by me when the `+python` variant features were under development.
The minimum versions were set to what was default /latest in upstream spack-packages at the time; now, the versions are based on the `nanobind` version we are using and its documented minimum python requirement. 

These versions requirements should apply for axom older than v0.15.0 as well.

Relevant axom issue discussion: https://github.com/llnl/axom/issues/1959

alecbcs: Okay that sounds reasonable. We attempt not to modify existing constraints in place as it'll break reusing old installations so I always make a point to ask about them just to double check.

(If you want to speed up the review in the future, a brief description of the specific changes like this goes a long way haha.)

alecbcs (approved):

PR #6309, opened 2026-09-01
amrex: add v26.09

WeiqunZhang:
Also update CUDA CMake interface for v26.09+.

### DIFF
--- a/repos/spack_repo/builtin/packages/amrex/package.py
+++ b/repos/spack_repo/builtin/packages/amrex/package.py
@@ -240,7 +240,8 @@ class Amrex(CMakePackage, CudaPackage, ROCmPackage):
     depends_on("cmake@3.13:", type="build", when="@18.11:19.03")
     depends_on("cmake@3.14:", type="build", when="@19.04:22.05")
     depends_on("cmake@3.17:", type="build", when="@22.06:23.01")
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/amrex/package.py
```diff
@@ -390,7 +404,10 @@ def cmake_args(self):
             args.append("-DAMReX_CUDA_ERROR_CAPTURE_THIS=ON")
             args.append("-DAMReX_CUDA_ERROR_CROSS_EXECUTION_SPACE_CALL=ON")
             cuda_arch = self.spec.variants["cuda_arch"].value
-            args.append("-DAMReX_CUDA_ARCH=" + self.get_cuda_arch_string(cuda_arch))
+            if self.spec.satisfies("@26.09:"):
+                args.append("-DCMAKE_CUDA_ARCHITECTURES=" + self.get_cuda_architectures(cuda_arch))
```
ax3l, line 408: I would hope the CUDA base package sets those automatically, but does not hurt to set explicitly.

ax3l (approved):
LGTM - checking with @becker33 if our more standard CMake+CUDA base package syntax here is done the right way.

PR #6310, opened 2026-09-01
PAML: Added v4.10.10. Added new ds bin

emwjacobson:
Bumped the version, adding the new "ds" software as well.

Compiled against a handful of gcc versions
| | paml@=4.10.10 |
| -- | -- |
|gcc@=15.3.0 | True |
|gcc@=13.4.0 | True |
|gcc@=12.2.0 | True |
|gcc@=9.5.0 | True |

Maintainer is @snehring

### DIFF
--- a/repos/spack_repo/builtin/packages/paml/package.py
+++ b/repos/spack_repo/builtin/packages/paml/package.py
@@ -18,6 +18,11 @@ class Paml(MakefilePackage):
 
     license("GPL-3.0-or-later")
 
+    version(
+        "4.10.10",
+        sha256="173e8754ad78000371099a96910f2c72b03a7b7b13c717405dc23385ae2f2c0f",
+        url="https://github.com/abacus-gene/paml/archive/refs/tags/v4.10.10.tar.gz",
+    )
     version("4.10.7", sha256="0f29e768b3797b69eadc6332c3d046d8727702052d56c3b729883626c0a5a4e3")
     version(
         "4.10.3",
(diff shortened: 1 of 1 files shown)


snehring (approved):
Checksum is correct. Changes look fine.

repos/spack_repo/builtin/packages/paml/package.py
```diff
@@ -18,6 +18,11 @@ class Paml(MakefilePackage):
 
     license("GPL-3.0-or-later")
 
+    version(
+        "4.10.10",
+        sha256="173e8754ad78000371099a96910f2c72b03a7b7b13c717405dc23385ae2f2c0f",
+        url="https://github.com/abacus-gene/paml/archive/refs/tags/v4.10.10.tar.gz",
```
cmelone, line 24: i'm not sure the url param is necessary given it fits the `url` template

emwjacobson: I thought it didnt match the URL though? The URL is set to "https://github.com/abacus-gene/paml/archive/refs/tags/4.10.7.tar.gz" (missing the "v" before the version number). Removing the `url=` from 4.10.10 fails.

I *could* remove the `url=` from 4.10.10 and update the package url, but then I'd need to modify the `4.10.3` version to manually set the URL.

cmelone: you're right, I missed the `v`. my bad, i'll merge.

cmelone (approved):

PR #6316, opened 2026-09-01
HDF5 patch: h5open_f failing to re-initialize the Fortran interface

climbfuji:
## Description

This PR introduces a patch for hdf5 versions 1.14.5 to the most recent release 2.1.0 for a bug reported upstream as [h5open_f silently fails to re-initialize the Fortran interface after h5close_f (prohibited argument aliasing in h5close_f)](https://github.com/HDFGroup/hdf5/issues/6642) and fixed in https://github.com/HDFGroup/hdf5/pull/6649.

The upstream bug fix is more comprehensive than the patch added here. The patch also only works for versions 1.14.5 and newer (I tested all the way up to and including 2.1.0), even though the bug was introduced in earlier versions (1.13.2).

I tested the patch on the system and with the compiler that triggered the issue, and it fixed the problem.


hyoklee (approved):

climbfuji:
Gentle reminder for the remaining reviewers ... thanks!

PR #6321, opened 2026-09-02
LBANN: support newer protobuf

adamjstewart:
LBANN requires ancient protobuf which requires ancient setuptools which contains known security vulnerabilities. Trying to deprecate ancient setuptools, but need to see if LBANN supports newer protobuf first.


adamjstewart:
Looks like we might be dead in the water until LBANN supports newer protobuf/setuptools versions...

PR #6322, opened 2026-09-02
ZFIT: drop setuptools-scm-git-archive dep

adamjstewart:
setuptools-scm-git-archive is abandoned as setuptools-scm now supports all of its features natively. Also, it requires an ancient version of setuptools with known CVEs. Trying to deprecate this, but need to figure out if we can skip this dep first.


jonas-eschle (approved):
Looks good. I would have not removed the older versions per se, but shouldn't matter. Anyways old. Thanks for that!

RikkiButler20 (approved):

PR #6323, opened 2026-09-02
pajeng new versions

viniciusvgp:


repos/spack_repo/builtin/packages/pajeng/package.py
```diff
@@ -48,6 +52,7 @@ class Pajeng(CMakePackage):
     depends_on("bison")
     depends_on("qt@:4+opengl", when="@:1.3.2+gui")
     depends_on("freeglut", when="@:1.3.2+gui")
+    depends_on("fmt", when="@1.3.7:")
     depends_on("fmt", when="@develop")
```
cmelone: these constraints can be consolidated into `depends_on("fmt", when="@1.3.7:")`

`develop` is an "infinite" version so it will match above 1.3.7

cmelone (approved):

repos/spack_repo/builtin/packages/pajeng/package.py
viniciusvgp: Done at 71dc785, thank you @cmelone!

PR #6324, opened 2026-09-02
llama-cpp: add v0.3.0

yizeyi18:
ggml-org projects released their versioning policy in Aug(https://github.com/ggml-org/ggml/discussions/1579).
This PR appends llama-cpp and ggml's newest versioned releases to spack builtin repo, and adds several missing dependencies for the two packages.

### DIFF
--- a/repos/spack_repo/builtin/packages/ggml/package.py
+++ b/repos/spack_repo/builtin/packages/ggml/package.py
@@ -92,5 +92,5 @@ class Ggml(GGMLPackageBase):
     license("MIT")
 
     version("master", branch="master")
-    version("0.22.0", sha256="34006b8c637dd507c4c9772cf826d8c3fcec00f49c094e74bb6e9e40c26d891d")
+    version("0.22.0", tag="v0.22.0", commit="34dc0e5589504286cb40e13cbdae4bf2b5b4071b", preferred=True)
     version("0.9.4", tag="v0.9.4", commit="72632094336524a9c809e129e8b1c52154543a5a")
--- a/repos/spack_repo/builtin/packages/ggml/package.py
+++ b/repos/spack_repo/builtin/packages/ggml/package.py
@@ -30,7 +30,7 @@ class GGMLPackageBase(CMakePackage, CudaPackage, ROCmPackage):
 
     depends_on("hipblas", when="+rocm")
     depends_on("rocblas", when="+rocm")
-    depends_on("rccl", when="+rccl")
+    depends_on("rccl", when="+rocm")
 
     depends_on("nccl", when="+cuda")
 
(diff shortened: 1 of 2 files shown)


repos/spack_repo/builtin/packages/ggml/package.py
```diff
@@ -78,10 +85,12 @@ class Ggml(GGMLPackageBase):
...
     homepage = "https://github.com/ggml-org/ggml"
     git = "https://github.com/ggml-org/ggml.git"
+    url = "https://github.com/ggml-org/ggml/archive/refs/tags/v0.22.0.tar.gz"
 
     maintainers("rbberger")
 
     license("MIT")
 
     version("master", branch="master")
+    version("0.22.0", sha256="34006b8c637dd507c4c9772cf826d8c3fcec00f49c094e74bb6e9e40c26d891d")
```
rbberger:
```suggestion
    version("0.22.0", tag="v0.22.0", commit="34dc0e5589504286cb40e13cbdae4bf2b5b4071b", preferred=True)
```

repos/spack_repo/builtin/packages/llama_cpp/package.py
```diff
@@ -16,20 +16,26 @@ class LlamaCpp(GGMLPackageBase):
...
     homepage = "https://github.com/ggml-org/llama.cpp"
     git = "https://github.com/ggml-org/llama.cpp.git"
+    url = "https://github.com/ggml-org/llama.cpp/archive/refs/tags/v0.3.0.tar.gz"
 
     maintainers("rbberger")
 
     license("MIT")
 
     version("master", branch="master")
+    version("0.3.0", sha256="d94c02d86db22d68692f6bb5b3854763d5091e52142868dc7251995517c666d1")
```
rbberger:
```suggestion
    version("0.3.0", tag="v0.3.0", commit="c1d0e7a004015f23bc0233470b747b596f29b264", preferred=True)
```

rbberger (changes requested):
This all looks ok. The only change I would request is to maintain the git tag/commit scheme.  The autogenerated tarballs from GitHub aren't guaranteed to be stable in the long run.

Given that the older versions are now always bigger, make the 0.3.0 as preferred and mark the older ones as deprecated. You probably also should add a dependency for `ggml@0.22.0:` for `llama-cpp@0.3.0`

repos/spack_repo/builtin/packages/ggml/package.py
yizeyi18: This version is released via github and works in spack. Is it necessary to change to git based style?

repos/spack_repo/builtin/packages/llama_cpp/package.py
yizeyi18: similar to above, this is also a versioned release in github and could be accessed by spack via version-replacing in url.

repos/spack_repo/builtin/packages/ggml/package.py
rbberger: It works, but the git style is preferred over auto-generated tarballs since if GitHub changes something on their end (which has happened in the past) all SHAS change. If the tarballs would be release artifacts that are stable that would be ok too, but that's not what they have.

yizeyi18:
@spackbot fix style

rbberger (commented):
We don't need the `url`. Also, there's a inconsistency with having a `rccl` variant vs `cuda`. Either also add a `nccl` variant or remove the `rccl` variant.

rbberger (approved):

kshea21 (approved):
Great, thanks everyone.

PR #4119, opened 2026-04-05
py-vllm: add new package

thomas-bouvier:

PR #6337, opened 2026-09-03
PyTorch: add v2.14.0, remove CVEs

adamjstewart:
PyTorch recipe is becoming unmaintainable and they just switched from setuptools to scikit-build, would like to remove deprecated versions to make my life easier.

* https://github.com/pytorch/pytorch/releases/tag/v2.14.0
* https://github.com/pytorch/vision/releases/tag/v0.29.0


repos/spack_repo/builtin/packages/py_sphericart_torch/package.py
```diff
@@ -1,38 +0,0 @@
...
-
-    # pyproject.toml
-    with default_args(type="build"):
-        depends_on("py-wheel@0.36:")
-        depends_on("py-setuptools@77:")
-        depends_on("cmake@3.30:")
-    depends_on("py-numpy", type=("build", "run"))
-
-    # setup.py
-    depends_on("py-torch@2.1:2.9", type=("build", "run"))
```
Luthaf: we don't have to remove the package, sphericart should build fine with the latest PyTorch (we test up to 2.13 on CI, but 2.14 should be fine as well)

adamjstewart: Can you update the package to support newer versions in a separate PR?

repos/spack_repo/builtin/packages/py_sphericart_torch/package.py
RMeli, line 1: This package should not be removed. We need to update it, version `2.0.3` supports `py-torch@:2.13`.

repos/spack_repo/builtin/packages/py_sphericart_torch/package.py
Luthaf: sure, I'll do it ASAP. Should also also to keep py-torch-spex and py-metatrain as-is.

repos/spack_repo/builtin/packages/py_torch_spex/package.py
RMeli, line 1: Likewise, this should not be deleted; `py-sphericart-torch` needs to be updated.

RMeli (changes requested):

repos/spack_repo/builtin/packages/py_metatrain/package.py
RMeli, line 1: These changes should be reverted, once #6339 is merged.

RMeli (changes requested):

repos/spack_repo/builtin/packages/py_sphericart_torch/package.py
adamjstewart: Please submit a separate PR to update, I can wait.

RMeli (dismissed):
Thanks @adamjstewart, the packages I was maintaining are no longer affected by this PR (see #6339) so LGTM.

adamjstewart:
Seeing some ROCm failures in CI:
```
CMake Error at cmake/public/LoadHIP.cmake:79 (find_package):
    By not providing "Findhipfile.cmake" in CMAKE_MODULE_PATH this project has
    asked CMake to find a package configuration file provided by "hipfile", but
    CMake did not find one.
    Could not find a package configuration file provided by "hipfile" with any
    of the following names:
      hipfileConfig.cmake
      hipfile-config.cmake
    Add the installation prefix of "hipfile" to CMAKE_PREFIX_PATH or set
    "hipfile_DIR" to a directory containing one of the above files.  If
    "hipfile" provides a separate development package or SDK, be sure it has
    been installed.
```
@renjithravindrankannath

renjithravindrankannath:
> Seeing some ROCm failures in CI:
> 
> ```
> CMake Error at cmake/public/LoadHIP.cmake:79 (find_package):
>     By not providing "Findhipfile.cmake" in CMAKE_MODULE_PATH this project has
>     asked CMake to find a package configuration file provided by "hipfile", but
>     CMake did not find one.
>     Could not find a package configuration file provided by "hipfile" with any
>     of the following names:
>       hipfileConfig.cmake
>       hipfile-config.cmake
>     Add the installation prefix of "hipfile" to CMAKE_PREFIX_PATH or set
>     "hipfile_DIR" to a directory containing one of the above files.  If
>     "hipfile" provides a separate development package or SDK, be sure it has
>     been installed.
> ```
> 
> @renjithravindrankannath

@adamjstewart , Thanks for pointing. I will create a separate PR to enable 2.14 on ROCm.

thomas-bouvier:
This PR might impact this ongoing work: https://github.com/spack/spack-packages/pull/4119

adamjstewart:
@thomas-bouvier do you see any big concerns with dropping older PyTorch? vLLM would indeed be good to get in, can we only add the latest version of packages?

repos/spack_repo/builtin/packages/py_torch/package.py
bernhardkaindl, line 292: @adamjstewart, @williampiat3:

Because https://github.com/pytorch/pytorch/pull/188263 isn't merged yet, this change would be needed to prevent the build of PyTourch 2.14 from fetching `six` from pypi to continue to support air-gapped build (without internet) for PyTorch 2.14.

Instead of yet another different patch for the new version, use `filter_file()` to universally add passing PYTHON_SIX_SOURCE_DIR to NNPACK which forwards it to PeachPy:

Update: Extended the range to start with 2.12 as it also works for 2.12 and 2.13 with the patch for 2.12 and 2.13 removed:

```suggestion
    def patch(self):
        # Continue to support building PyTorch 2.12+ in air gapped environments
        # This forwards the six source folder to NNPACK which forwards it to PeachPy
        # PyTorch PR: https://github.com/pytorch/pytorch/pull/188263 (pass the path)
        # PeachPy PR: https://github.com/Maratyszcza/PeachPy/pull/136 (replace six)
        if self.spec.satisfies("@2.12: +nnpack"):
            filter_file(
                "_ENV_PASSTHROUGH",
                "_ENV_PASSTHROUGH PYTHON_SIX_SOURCE_DIR",
                "cmake/EnvVarForwarding.cmake"
            )
            filter_file(
                'message(STATUS "Brace yourself, we are building NNPACK")',
                'message(STATUS "Brace yourself, we are building NNPACK")\n'
                "  if(NOT DEFINED PYTHON_SIX_SOURCE_DIR AND DEFINED ENV{PYTHON_SIX_SOURCE_DIR})\n"
                "    set(PYTHON_SIX_SOURCE_DIR $ENV{PYTHON_SIX_SOURCE_DIR} CACHE PATH)\n"
                "  endif()",
                "cmake/External/nnpack.cmake",
                string=True,
            )    
```

PS: This was found using an experimental spack core repo that sandboxes builds:
https://github.com/bernhardkaindl/spack/tree/integrated-sandbox-poc
PPS: Expected to also reproduce using a podman-based airgap test for spack packages:
https://github.com/Hereon-InstituteMS/spack-agent

bernhardkaindl, line 429: Hi @adamjstewart, at least with gcc-16 and clang-22 I need to pass flags to enable AVX512 in order to build PyTorch 2.14 on a AMD Ryzen 9 7945HX which supports AVX512:

```py
tr ' ' '\n' </proc/cpuinfo|grep 512 | sort -u
avx512_bf16
avx512_bitalg
avx512_vbmi2
avx512_vnni
avx512_vpopcntdq
avx512bw
avx512cd
avx512dq
avx512f
avx512ifma
avx512vbmi
avx512vl
```
```suggestion
        enable_or_disable("nnpack")
        if self.spec.satisfies("+nnpack"):
            if self.spec.satisfies("%gcc@16:") or self.spec.satisfies("%clang@22:"):
                avx512_flags = "-mavx512bf16 -mavxneconvert -mavx512vl"
                # FBGEMM’s CMake feature detection overwrites CXX_AVX512_FLAGS:
                env.append_flags("CMAKE_CXX_FLAGS", avx512_flags)
```
Maybe other compilers need those flags too, this was just my initial test.

repos/spack_repo/builtin/packages/py_torch/package.py
```diff
@@ -403,244 +261,31 @@ class PyTorch(PythonPackage, CudaPackage, ROCmPackage):
...
 
     # Make Pytorch build work in air gapped environments (without internet access)
     # This forwards six source folder path to NNPACK which forwards it to PeachPy
     # for versions @2.5:2.11
-    patch("air_gapped_nnpack_cmake_older.patch", when="@2.5:2.11")
+    patch("air_gapped_nnpack_cmake_older.patch", when="@:2.11")
     # for version @2.12: (env forwarding mechanism changed)
     # This error has been raised upstream https://github.com/pytorch/pytorch/pull/188263
-    patch("air_gapped_nnpack_cmake.patch", when="@2.12:")
+    patch("air_gapped_nnpack_cmake.patch", when="@2.12:2.13")
```
bernhardkaindl: This patch to support air-gapped build can be deleted when extending the `filter_file()` o cover `@2.12:`
```py
   patch("air_gapped_nnpack_cmake.patch", when="@2.12:2.13")
```

PS: The same might also apply to the older PyTorch versions that use this patch when extending the coverage of the filter_file() below. I was not able to test 2.10:2.11 (older versoins are removed by this PR) in my environment so far:
```
 patch("air_gapped_nnpack_cmake_older.patch", when="@:2.11")
```

repos/spack_repo/builtin/packages/py_torch/package.py
williampiat3: The air gapped enabling patch is failing only because Pytorch removed one of the variables in the envforwarding.cmake file.
I would rather change the current patch to insert PYTHON_SIX_SOURCE_DIR at the *beginning* of the list rather than the *end* of the list as your patch is actually doing
We will avoid inserting too many lines in the recipe which, for what I have  understood, needs to remain as readable as possible

adamjstewart: This PR deletes 2k lines of code, so I'm not too worried about adding a few more lines to the recipe if it makes life simpler.

adamjstewart: This is pretty microarch-specific, @becker33 @alalazo any reason why this isn't already being applied or a suggestion for how to make this more generic? I was under the impression that Spack already set these flags using this env var.

adamjstewart: Also, can we all take a moment to stop and think about how funny this is and the possible reasons why this was added?
```cmake
message(STATUS "Brace yourself, we are building NNPACK")
```

adamjstewart: Added another patch file and will continue to push on the upstream PR.

repos/spack_repo/builtin/packages/py_torch/air_gapped_nnpack_cmake_2.14.patch
```diff
@@ -0,0 +1,26 @@
+diff --git a/cmake/EnvVarForwarding.cmake b/cmake/EnvVarForwarding.cmake
+index a31ce6db26637..258d458099e27 100644
+--- a/cmake/EnvVarForwarding.cmake
++++ b/cmake/EnvVarForwarding.cmake
+@@ -42,6 +42,7 @@ set(_ENV_PASSTHROUGH
+   TORCH_XPU_ARCH_LIST
+   TRACING_BASED
+   PYTHON_LIB_REL_PATH
++  PYTHON_SIX_SOURCE_DIR
```
bernhardkaindl, line 9: @adamjstewart, please see:
> 1 out of 1 hunk FAILED -- saving rejects to file cmake/EnvVarForwarding.cmake.rej

The reason is that as @williampiat3 said, they removed PYTHON_LIB_REL_PATH from the list with 2.14, and thus your essentially unchanged patch (except for whitespace) from 2.12 does not apply to 2.14:
```diff
diff repos/spack_repo/builtin/packages/py_torch/air_gapped_nnpack_cmake_2.1[24].patch
2c2
< index a31ce6db266..258d458099e 100644
---
> index a31ce6db26637..258d458099e27 100644
11c11
<  
---
> 
14c14
< index 8a4a310d629..2491d4e2cc2 100644
---
> index 8a4a310d6292a..2491d4e2cc2a7 100644
18c18
<  
---
> 
```

The tendency of people to add changes at the end was exactly my reason (also suggested by @williampiat3) to make the change at the top of the list. Besides being safer to actually patch the location you want, it also gives the reader of the a better view what is being patched:

```diff
--- cmake/EnvVarForwarding.cmake
+++ cmake/EnvVarForwarding.cmake
@@ -123,4 +123,5 @@
 # Additional env vars forwarded with the same name.
 set(_ENV_PASSTHROUGH
+  PYTHON_SIX_SOURCE_DIR
   UBSAN_FLAGS
   BLAS
```

I even prefer such small changes directly in the recipe like I suggested, but it's your call.

bernhardkaindl (changes requested):

repos/spack_repo/builtin/packages/py_torch/air_gapped_nnpack_cmake_2.14.patch
bernhardkaindl: > Also, can we all take a moment to stop and think about how funny this is and the possible reasons why this was added?
> ```cmake
> message(STATUS "Brace yourself, we are building NNPACK")
> ```

👍 Yes, I did pick this line as the anchor for the change for two reasons:
- To humor you :)
- And to give context that this change is in the case where the NNPACK build is defined.

So, likewise I picked this to be anchored at a line that from experience is unlikely to be changed (it is too funny!)

Just for some more fun:

If you compiled Eterm, an old X11 terminal emulator and ran its configure script, you saw:
```py
checking for Linux 2.1 or higher... yes
checking for life_signs in -lKenny... no
Oh my god, they killed Kenny! You bastards!
updating cache ./config.cache
creating ./config.status
creating Makefile
```

Source:
https://sources.debian.org/src/eterm/0.8.7-0.4/configure.in#L308
```m4
AC_CHECK_LIB(Kenny, life_signs, , [
  echo "  Oh my god, they killed Kenny!  You bastards!"
])
```

bernhardkaindl: https://news.slashdot.org/comments.pl?sid=278&cid=1825996
> Apparently (unfortunately) he's gotten some flack from folks who apparently think compiles should be serious :P and one thing he is considering is...putting a --with-dead-kenny tag in future versions of Eterm. ;) 

So the final version is even funnier: `HUMOR=none`, `test -z "$HUMOR"` `$HAVE_HUMOR`:
```m4
dnl# Enable/disable humor
AC_ARG_WITH(sense-of-humor, [  --without-sense-of-humor   Specify this if you have no sense of humor], [
            if test "$withval" = "no"; then
              HUMOR=none
            fi
])

if test -z "$HUMOR"; then
AC_CHECK_LIB(Kenny, life_signs, , [
  echo "  Oh my god, they killed Kenny!  You bastards!"
  AC_DEFINE(HAVE_HUMOR, 1, [Defined if user has a sense of humor.])
])
fi
```

williampiat3: Ultimatly since we patched nnpack's spack recipe we should be able to get rid of all these patches if we manage to link pytorch to a preexisting nnpack install but this is out of the scope of this PR

A corrected air_gapped_nnpack_cmake.patch  is the quickest and smallest working fix for a working  airgapped recipe for 2.12:2.14 

 I don't have access to a computer right now otherwise I would have suggested the corrected fix... Sorry for this

adamjstewart: Should be fixed now, I didn't realize GitHub PR patches were separated by commits instead of combined so messed up the formatting.

For now I prefer patch files as it keeps package.py simple, but let's see if we can push on PyTorch to merge some kind of fix sooner rather than later. If anyone is bored, it would be even better if PyTorch built entirely with Spack-installed deps (including nnpack) instead of third_party, but that's a much bigger lift required.

adamjstewart:
@spackbot run pipeline

PR #6338, opened 2026-09-03
voms: allow to compile even when gsoap and zlib-ng are not loaded

dstocco:
The current voms implementation breaks (at least on MacOS), unless gsoap and zlib-ng are loaded.
This happens because the build runtime environment needs to set some flags taken from gsoap and zlib with pkgconfig, but, at build time, those libraries are not in the PKG_CONFIG_PATH.
With this PR, the pkgconfig path of the two libraries is filled from the path, so that the command setting the flag does not crash.

### DIFF
--- a/repos/spack_repo/builtin/packages/voms/package.py
+++ b/repos/spack_repo/builtin/packages/voms/package.py
@@ -45,8 +45,14 @@ class Voms(AutotoolsPackage):
     def setup_build_environment(self, env: EnvironmentModifications) -> None:
         # https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=voms
         pkgconfig = Executable(join_path(self.spec["pkgconfig"].prefix.bin, "pkg-config"))
-        env.set("GSOAP_SSL_PP_CFLAGS", pkgconfig("--cflags", "gsoapssl++", "zlib", output=str))
-        env.set("GSOAP_SSL_PP_LIBS", pkgconfig("--libs", "gsoapssl++", "zlib", output=str))
+        env.set(
+            "GSOAP_SSL_PP_CFLAGS",
+            pkgconfig("--cflags", "gsoapssl++", "zlib", env=env, output=str),
+        )
+        env.set(
+            "GSOAP_SSL_PP_LIBS",
+            pkgconfig("--libs", "gsoapssl++", "zlib", env=env, output=str),
+        )
 
     def autoreconf(self, spec, prefix):
         autogen = Executable("./autogen.sh")


repos/spack_repo/builtin/packages/voms/package.py
```diff
@@ -45,8 +47,15 @@ def patch(self):
     def setup_build_environment(self, env: EnvironmentModifications) -> None:
         # https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=voms
         pkgconfig = Executable(join_path(self.spec["pkgconfig"].prefix.bin, "pkg-config"))
-        env.set("GSOAP_SSL_PP_CFLAGS", pkgconfig("--cflags", "gsoapssl++", "zlib", output=str))
-        env.set("GSOAP_SSL_PP_LIBS", pkgconfig("--libs", "gsoapssl++", "zlib", output=str))
+        pkgconfig_env = os.environ.copy()
```
wdconinc: This is not canonical and leaks the environment from the process that does the install into the build environment.

wdconinc: Already in spack's `lib/spack/spack/build_environment.py` these same directories should already be added to `PKG_CONFIG_PATH`, so I don't understand why this is needed.

dstocco: Not sure, but this is the error I get if this part is not present:

```
 -- lines 1 to 7 --
  Package gsoapssl++ was not found in the pkg-config search path.
  Perhaps you should add the directory containing `gsoapssl++.pc'
  to the PKG_CONFIG_PATH environment variable
  Package 'gsoapssl++' not found
  Package 'zlib' not found
  Command exited with status 1:`
    ~/sw/spack/opt/spack/darwin-m2/pkgconf-2.5.1-aor3wo2pkr62oe4j6facvqehtd2rsply/bin/pkg-config --cflags gsoapssl++ zlib
 ```
 
 Same error if I try to run `spack build-env voms bash`

Comment (dstocco) [reply repos/spack_repo/builtin/packages/voms/package.py]: Notice that in my environment I have: `view: false`. So the installed path are not automatically exposed when activating the environment where  build.

Comment (dstocco) [reply repos/spack_repo/builtin/packages/voms/package.py]: I have cross-checked: without my modifications, if `view: true` the package compiles. If `view: false` it does not.
I have slightly changed the initial commit to avoid using `os.environ`.

Comment (wdconinc) [reply repos/spack_repo/builtin/packages/voms/package.py]: I'm still confused by this issue. When I look in my `/opt/software/linux-skylake/voms-2.1.3-moabmhybrkpsfxoj2udxzre7bgjc26hz/.spack/spack-build-env.txt`, then the build environment already includes the right pkgconfig directories, i.e. `/opt/software/linux-skylake/gsoap-2.8.135-vhz7yb5akdlgvjocbu6h5h36d7ljbodk/lib/pkgconfig` and `/opt/software/linux-skylake/zlib-ng-2.3.3-5vewy42a54dukjx3bam7eyg6y5jr6xyh/lib/pkgconfig`.

@sethrj You're a macos person; is there something that macos `pkgconf` does differently? Do you observe this issue when installing voms?

Comment (dstocco) [reply repos/spack_repo/builtin/packages/voms/package.py]: Looking closer, I do not think it is a problem with MacOS. The `env` that it is passed to the `setup_build_environment` function correctly contains PKG_CONFIG_PATH. However, I suspect that when one launches the `pkgcofig` executable, this is launched on a sub shell, where the environment variables are not correct.
If one is compiling within a spack environment with `view: true`, then this is probably not an issue since the sub shell will see the path to the dependencies.
But when `view: false`, the environment variables in the sub shell where the command is executed are not set.

So, the easy solution is indeed to pass the current build environment `env` to the `pkgconfig` command, so that, whatever the view status, it will be executed with the correct environment variables.

This is exactly what I did in the last commit, with minimal modifications.

Comment (dstocco) [reply repos/spack_repo/builtin/packages/voms/package.py]: You can easily test it on your side with:
```
spack env create --without-view test
spacktivate test
spack add voms
spack install
```
It will fail without my modifications, and it should work with the modifications.

wdconinc: Confirmed.

wdconinc: Thanks for the reproducer. The analysis and fix is correct, but there may be a more systemic way to address this can adding `env=env` (since there are other places where this may pick up the wrong info). See https://spackpm.slack.com/archives/C5VL7V81G/p1788459112116079 where I've opened this up for discussion (no responses yet). If nothing more fundamental by tomorrow, we can go with your fix.

dstocco:
@wdconinc : I have just pushed again since the CI was complaining about two lines being too long (although I run `spack style` locally and got no error).
I have therefore just added some newlines (formatted by black).

repos/spack_repo/builtin/packages/voms/package.py
sethrj: Not sure if I still need input here, but the only things I've noticed on macos with pkgconf is that I've gotten downstream configuration errors when using homebrew-installed pkgconf rather than the spack-installed pkg-config.

wdconinc (approved):
Thanks for this fix. We may find a more systemic way to address this yet, but this fix is good as well.

dstocco:
Thanks for the discussions and for merging. Looking forward for a more systemic fix to the issue.

PR #6340, opened 2026-09-03
graph500: fix build with GCC 14+, add procs_not_power_of_two variant

juntangc:
`graph500` 3.0.0 (2017, unmaintained) fails to build on GCC 14+, which promoted
`-Wimplicit-function-declaration` to an error. This patches the sources, fixes an
unrelated pre-existing `flag_handler` bug, and adds a non-power-of-two rank variant.

**Blast radius: zero.** `graph500` is a leaf — across the 9,044 packages in
`spack_repo/builtin` the string occurs in exactly one file, its own `package.py`.
No dependents, and it declares no `provides(...)`.

---

### 1. GCC 14+ build failure

| Call site | Symbol | Affected targets |
| :--- | :--- | :--- |
| `src/main.c:324` | `isisolated` — defined at `src/csr_reference.c:32`, declared in no header | all four |
| `src/sssp_reference.c:102` | `memset` — via `CLEAN_VISITED()` (`src/bitmap_reference.h:18`), no `<string.h>` in scope | `*_sssp` only |

The `memset` half only bites the SSSP targets: `bfs_reference.c` also expands
`CLEAN_VISITED()` but includes `<string.h>` first, and the `sssp_reference.c`
expansion sits behind `#ifdef SSSP`. New patch adds the prototype and the include
— two lines of actual change.

<details>
<summary>Verified complete — no other implicit declarations remain</summary>

Every TU compiled with `-fsyntax-only -Wimplicit-function-declaration`, pristine vs
patched, both configs (GCC 15.2.0, one major version past the target):

```
PRISTINE   [no -DSSSP]    1:  main.c -> isisolated
           [with -DSSSP]  2:  main.c -> isisolated
                              sssp_reference.c -> memset
PATCHED    [no -DSSSP]    0
           [with -DSSSP]  0
```
</details>

<details>
<summary><code>-fcommon</code> and <code>-fgnu89-inline</code></summary>

**`-fcommon`** (pre-existing, preserved): `column` and `weights` are tentative
definitions in multiple TUs linked together — `csr_reference.h:22/24`,
`csr_reference.c:27/28`, `bfs_reference.c:39`, `sssp_reference.c:32`.

**`-fgnu89-inline`** (new): five functions are bare `inline` with no external
definition — `send_visit` (`bfs_reference.c:61`) and `aml_poll_intra`,
`flush_buffer`, `flush_buffer_intra`, `aml_send_intra` (`aml/aml.c:133/176/193/210`).
Under C99 that links only because the upstream Makefile hardcodes `-O3`; at any
lower optimization level the calls are real and fail to link. GNU89 semantics emit
the external definition. No duplicate-symbol risk — each is defined in exactly one
TU, and `bfs_reference.c`/`bfs_custom.c` never link together.

`static inline` would be the cleaner source fix but touches six sites across two
files including vendored `aml/`; happy to do that instead. Also note the guard is
`%gcc` while clang/oneapi share C99 inline semantics and accept the same flag —
I matched the existing `%gcc@10:` `-fcommon` gate, but can widen both.
</details>

### 2. `flag_handler` crashed on any user-supplied flags

Pre-existing bug. Flags were returned in the `build_system_flags` slot, but
`MakefilePackage` has no `flags_to_build_system_args`, so the base implementation
raises. On `develop` today:

```console
$ spack install graph500 cflags="-g -O0"
==> Error: The Graph500 build system cannot take command line arguments for compiler flags.
```

`return (flags, None, None)` fixes it — and the wrapper is the *only* workable slot,
since `src/Makefile:1` hard-assigns `CFLAGS = ...` (not `?=`), so `env_flags` would
be silently discarded. Package flags are appended after the user's, so package
`-fcommon` still wins over a user `-fno-common`; that ordering is load-bearing.

### 3. New variant: `procs_not_power_of_two`

Defaults to `False`; enabling appends `-DPROCS_PER_NODE_NOT_POWER_OF_TWO`. No source
change needed — `common.h:19-21` gates `SIZE_MUST_BE_A_POWER_OF_TWO` behind an
`#ifndef` on that macro, which switches `MOD_SIZE` (`common.h:32-40`) from bitmask to
modulo arithmetic; `VERTEX_OWNER` and `VERTEX_LOCAL` (`:42-43`) are defined in terms
of it. `aml/aml.c` (42, 301-303, 315) consumes the same macro, so one `-D` covers both
layers. Off by default because modulo is slower and most runs use power-of-two ranks.

---

<details>
<summary>Testing — Rocky 9 / cascadelake, OpenMPI 5.0.10 and Intel MPI</summary>

**`%gcc@14.2.1`** (via `gcc-toolset-14`) — the case this fixes:

| Configuration | Result |
| :--- | :--- |
| `^openmpi`, patch **disabled** | ❌ fails with both errors |
| `^openmpi` | ✅ builds |
| `^openmpi`, `mpirun -n 4` | ✅ runs to completion |

**`%gcc@12.2.1`** — regression coverage:

| Configuration | Result |
| :--- | :--- |
| `^openmpi` / `^intel-oneapi-mpi` | ✅ builds |
| `cflags="-g -O0" ^openmpi` | ✅ builds (errors on `develop`) |
| `^openmpi`, `mpirun -n 3` | ✅ fails as expected |
| `+procs_not_power_of_two ^openmpi`, `mpirun -n 3` | ✅ runs to completion |
| `^intel-oneapi-mpi`, `mpiexec -n 3` | ✅ fails as expected |
| `+procs_not_power_of_two ^intel-oneapi-mpi`, `mpiexec -n 3` | ✅ runs to completion |

```
* **Spack:** 1.3.0.dev0 (https://github.com/spack/spack/commit/307d3b0a68316341401d311c1ef6442ed2f13868)
* **Python:** 3.12.14
* **Platform:** linux-rocky9-cascadelake
```
</details>

Also: changed `mkdir` → `mkdirp`, `.format()` → f-string.


juntangc:
Friendly ping — CI is green and this has been open since Sept 3. graph500 had no maintainers before this PR, so I don't think spackbot pinged anyone.

graph500 currently fails to build on GCC 14+ — two implicit function declarations that GCC 14 promotes to errors. Verified on gcc@14.2.1: fails without the patch, builds and runs with it.

The PR also fixes a separate pre-existing bug: flag_handler returned flags in the build_system_flags slot, which MakefilePackage can't consume, so spack install graph500 cflags=... errors out on develop today.

Tested with OpenMPI and Intel MPI, GCC 12 and 14, including 3-rank runs with and without the new variant. graph500 is a leaf package — nothing depends on it, so blast radius is zero.

@alecbcs would you be able to take a look, or point me at someone better placed?

PR #6346, opened 2026-09-03
conduit: version 0.9.8 requires at minimum CMake 3.26

bmhan12:
This PR:
- Makes explicit that conduit version 0.9.8 requires at minimum CMake 3.26: https://github.com/llnl/conduit/blob/v0.9.8/src/CMakeLists.txt#L5-L6


cyrush (approved):
thanks @bmhan12 !

cmelone (approved):

PR #6347, opened 2026-09-03
copacabana: new package

jfalcou:
Copacabana is the CMake package tools a family of header-only C++ libraries writes its build, tests, documentation and packaging with. It has no CMake config to find: consumers include it by path, and CPM fetches it at configure time. The package therefore installs the tree as it is and hands the prefix to CPM through `CPM_COPACABANA_SOURCE`, which is what the packages for tts, spy, kumi and raberu do.

### DIFF
--- a/repos/spack_repo/builtin/packages/copacabana/package.py
+++ b/repos/spack_repo/builtin/packages/copacabana/package.py
@@ -1,0 +1,28 @@
+# Copyright Spack Project Developers. See COPYRIGHT file for details.
+#
+# SPDX-License-Identifier: (Apache-2.0 OR MIT)
+
+from spack_repo.builtin.build_systems.generic import Package
+
+from spack.package import *
+
+
+class Copacabana(Package):
+    """Copacabana - the CMake package tools a C++ library's build, tests, documentation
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/copacabana/package.py
```diff
@@ -0,0 +1,28 @@
...
+
+
+class Copacabana(Package):
+    """Copacabana - the CMake package tools a C++ library's build, tests, documentation
+    and packaging are written with, shared by a family of header-only libraries."""
+
+    homepage = "https://github.com/jfalcou/copacabana"
+    url = "https://github.com/jfalcou/copacabana/archive/refs/tags/v7.tar.gz"
+    maintainers("jfalcou")
+    git = "https://github.com/jfalcou/copacabana.git"
```
cmelone: nit:

```suggestion
    git = "https://github.com/jfalcou/copacabana.git"
    
    maintainers("jfalcou")
```

jfalcou: Done

cmelone (approved):

PR #6349, opened 2026-09-03
spy: new package

jfalcou:
SPY detects compilers, operating systems, architectures and SIMD extensions at compile time. Same shape as the tts package: the copacabana prefix is handed to CPM, nothing is fetched during the build, and `check()` builds the aggregate test target before ctest.

Depends on the copacabana package, in #6347.


kshea21 (approved):
LGTM and built, thanks!

PR #6352, opened 2026-09-04
Add ectrans versions 1.9.0 and 2.0.0

climbfuji:
## Description

This PR adds ectrans@1.9.0 and ectrans@2.0.0. It doesn't configure any new build options. I noticed that these build options were incomplete for earlier versions as well, and because there are so many new and complicated options, I believe that the package developers (@samhatfield & colleagues) can do a better job here.

I did, however, verify that the find-lapack patch is no longer required to find the correct, spack-provided LAPACK libraries. I also checked FFTW, and those are correct as well. For example, on a Cray with Intel oneAPI MKL instead of cray-libsci as lapack/blas providers, the Intel oneAPI mkl packages were found and used for both LAPACK and FFTW.

### DIFF
--- a/repos/spack_repo/builtin/packages/ectrans/package.py
+++ b/repos/spack_repo/builtin/packages/ectrans/package.py
@@ -28,6 +28,8 @@ class Ectrans(CMakePackage):
 
     version("develop", branch="develop", no_cache=True)
     version("main", branch="main", no_cache=True)
+    version("2.0.0", sha256="dbe7875fca912a431b01defa2c6af51982c1ae1fb9f8eaa3504c40c8762e0638")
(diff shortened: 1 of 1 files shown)


climbfuji:
@samhatfield FYSA

samhatfield:
It looks good to me. Ready for approval IMO.

In terms of build options, it's true, there are a lot of CMake features in ecTrans nowadays, many missing from the recipe. However I think all of the ones missing in the recipe are only relevant when building for GPU. The recipe isn't aware of GPUs at all at the moment, and I don't think there's any real demand for changing that. I suggest we park this until a real need for GPU support arises.

mathomp4 (approved):

cmelone (approved):

PR #6355, opened 2026-09-04
packages: remove duplicate variant definitions

alalazo:
Found these duplications while looking for other variant related info in the `builtin` repo


xylar (approved):
Oops, thanks so much @alalazo.  Agreed, this needs fixing.

wdconinc (approved):
Approving for gfal2 of course.

kshea21 (approved):
Thanks for finding these, approved

PR #6358, opened 2026-09-04
Update package tl-expected: make build possible in air-gapped environments

williampiat3:
tl-expected uses Cmake command FetchContent: https://github.com/TartanLlama/expected/blob/master/CMakeLists.txt#L70
This generates errors when building in an airgapped environment

This patches the recipe to prefetch  the sources using ressource command and FETCHCONTENT_SOURCE_DIR variable


charmoniumQ (approved):

kshea21 (approved):
LGTM, thanks!

PR #6370, opened 2026-09-06
tdls: add v0.1.0 and test support

trsxvz:
Adds the first release of TDLS, v0.1.0, from its tag archive.

Makes `spack install --test=root tdls` build and run the test suites and the examples. The plain install stays compiler-free, the library is header-only. A compiler is only pulled in as a `test` dependency.

The compiler of the spec is handed to CMake explicitly. Spack does not inject the compiler wrapper for a language dependency of type `test`, so `CXX` would otherwise fall back to the compiler found on the PATH. This is spack/spack#45573. The fix is proposed in spack/spack#52982. Once it lands, the explicit setting becomes redundant and stays harmless.

Checked locally: `spack style`, `spack audit packages tdls`, `spack checksum`, and `spack install --test=root tdls@0.1.0` with all 83 tests passing.

### DIFF
--- a/repos/spack_repo/builtin/packages/tdls/package.py
+++ b/repos/spack_repo/builtin/packages/tdls/package.py
@@ -24,9 +24,26 @@ class Tdls(CMakePackage):
     version("main", branch="main")
 
     depends_on("cmake@3.21:", type="build")
+    # A compiler is only needed to build and run the test suites and the
+    # examples under --test. The compiler wrapper is not injected for a
+    # test-typed language dependency (spack/spack#45573), so the compiler
+    # of the spec is handed to CMake explicitly.
+    depends_on("cxx", type="test")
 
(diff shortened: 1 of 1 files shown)


trsxvz:
Hi @alalazo , sorry for the direct ping. Would you mind taking a look at this PR whenever you have some time? Thanks!

trsxvz:
Hi @RikkiButler20, sorry to ping you again. You merged my previous PR quickly so i wondered if you might have time to look at any of these 3 other PRs that i have opened on spack-packages. Theyre all kind of related because they're part of the same nuclear fuel simulation framework that we use in my lab.

https://github.com/spack/spack-packages/pull/6370 (this current one)
https://github.com/spack/spack-packages/pull/6623
https://github.com/spack/spack-packages/pull/6639

PR #6378, opened 2026-09-07
reframe: Switched to py-hatchling since 4.10.0

RiMillo:
As per [`pyproject.toml`](https://github.com/reframe-hpc/reframe/blob/v4.10.0/pyproject.toml)

Maintainers: @victorusu @vkarak


vkarak (approved):
Good catch, thanks!

RikkiButler20 (approved):

PR #6385, opened 2026-09-07
copacabana: add 8

jfalcou:
copacabana v8 is what the current releases of its consumers pin, kumi 5.0 first.


RikkiButler20 (approved):
LGTM

PR #6391, opened 2026-09-08
py-conan: add v1.66.0

LydDeb:

### DIFF
--- a/repos/spack_repo/builtin/packages/py_conan/package.py
+++ b/repos/spack_repo/builtin/packages/py_conan/package.py
@@ -21,7 +21,7 @@ class PyConan(PythonPackage):
     depends_on("python@3.6:", type=("build", "run"))
     depends_on("py-setuptools", type="build")
     depends_on("py-requests@2.25:2", type=("build", "run"))
-    depends_on("py-urllib3@1.26.6:1.26", type=("build", "run"))
+    depends_on("py-urllib3@1.26.6:", type=("build", "run"))
     depends_on("py-colorama@0.3.3:0.4", type=("build", "run"))
     depends_on("py-pyyaml@3.11:6.0", type=("build", "run"))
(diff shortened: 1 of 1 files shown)


repos/spack_repo/builtin/packages/py_conan/package.py
bernhardkaindl: Hi @LydDeb, can you check if you can add a newer urllib dependency for the new version because as it is now, it only builds when enabling --deprecated to build the deprecated urllib 1.x versions (See #4811 for reference)
```py
spack install py-conan
==> Error: failed to concretize `py-conan` for the following reasons:
     1. Cannot satisfy 'py-urllib3@1.26.6:1.26'
     2. Cannot satisfy 'py-urllib3@1.26.6:1.26' (version 2.7.0 does not match)
        required because py-conan depends on py-urllib3@1.26.6:1.26 
          required because py-conan requested explicitly 
```

LydDeb: The constraint on `py-urllib3` originates in the `conans/requirements.txt` file:
```
urllib3>=1.26.6, <1.27
```
Furthermore, I note that the only usage of `urllib3` in Conan 1.66.0 is found in the `conans/client/rest/conan_requester.py` file:
```
return urllib3.Retry(
total=retry,
backoff_factor=0.05,
status_forcelist=retry_status_code_set
)
```
According to the changelogs, this call to `urllib3` does not appear to break when upgrading `urllib3` from version 1.26.6 to 2.7.0.
Therefore, it is reasonable to assume that the version constraint on `urllib3` can be extended.

bernhardkaindl: Thanks for checking, build was successful here with this change.

bernhardkaindl (approved):
Thx, will merge.

PR #6133, opened 2026-08-18
py-protobuf: Add current versions, deprecate versions affected by CVE-2026-0994

bernhardkaindl:
Due to CVE-2026-0994 for py-protobuf, add fixed versions and deprecated the affected versions:
- Add py-protobuf 5.29.6, 6.33.6, 7.34.2 and 7.35.1, deprecate all previous versions.
- Add protofbuf 29.6, 33.6, 34.2 and 35.1

Background: While reviewing
- #6052
  I saw that upstream has bumped it's protobuf version constraint to >= 6.33.5 due to CVE-2026-0994:
  - https://github.com/microsoft/onnxruntime/pull/29606
  - https://github.com/advisories/GHSA-7gcm-g887-7qv7

It appears to be difficult to check which other packages need updates as the py-protofuf versions are pinned to corresponding protobuf versions:

```py
    # https://protobuf.dev/support/version-support/#python
    for ver in range(30, 33):
        depends_on(f"protobuf@{ver}", when=f"@6.{ver}")
    for ver in range(26, 30):
        depends_on(f"protobuf@{ver}", when=f"@5.{ver}")
    for ver in range(21, 26):
        depends_on(f"protobuf@{ver}", when=f"@4.{ver}")
    for ver in range(0, 21):
        depends_on(f"protobuf@3.{ver}", when=f"@3.{ver}")
```
   
The dependents in spack-packages are:
```py
git grep '"py-onnxruntime'
repos/spack_repo/builtin/packages/acts/package.py:    depends_on("py-onnxruntime@1.12:", when="+onnx")
repos/spack_repo/builtin/packages/purify/package.py:    depends_on("py-onnxruntime@1.17.1:", when="+onnxrt")
repos/spack_repo/builtin/packages/py_earth2mip/package.py:            depends_on("py-onnxruntime@1.15.1:")
repos/spack_repo/builtin/packages/py_onnx_opcounter/package.py:        depends_on("py-onnxruntime")
repos/spack_repo/builtin/packages/py_onnxmltools/package.py:    depends_on("py-onnxruntime", type=("build", "run"))
repos/spack_repo/builtin/packages/sopt/package.py:    depends_on("py-onnxruntime@1.17.1:", when="+onnxrt")
```

PR #6401, opened 2026-09-09
py-protobuf: add v6.33.6

LydDeb:
Hi @LydDeb, @wdconinc and other maintainers:

Could you review merging https://github.com/spack/spack-packages/pull/6133 instead?
- #6133


bernhardkaindl (commented):
Hi @LydDeb, @wdconinc and other maintainers: Could you review merging #6133 instead?

PR #6412, opened 2026-09-10
dealii: add dealii v9.8.0, and make sure it compiles with and without CUDA

luca-heltai:
New deal.II version +small fix for mumps.


luca-heltai:
@spackbot fix style

luca-heltai:
Part of https://github.com/dealii/dealii/issues/19728

PR #6421, opened 2026-09-10
fix papi for version 7.0.1

dtaller:
I got a key error  KeyError: 'rocp_sdk' because this variant is only present in 7.2: but the configure logic tried to access it for all versions (including version 7.0.1 which I was testing)


alecbcs (approved):
Looks good to me. Thanks @dtaller!

PR #6423, opened 2026-09-10
xyce: make patch conditional and use fftw dependency

kuberry:
- Patch for issue 454 no longer needed after version 7.10.
- Adds `+verbose` variant
- Modifies CMake options for `+fftw` dependency, when specified

### DIFF
--- a/repos/spack_repo/builtin/packages/xyce/package.py
+++ b/repos/spack_repo/builtin/packages/xyce/package.py
@@ -185,7 +185,7 @@ class Xyce(CMakePackage):
             options.append("-DXyce_VERBOSE_TIME=ON")
         if "+fftw" in spec:
             options.append("-DXyce_USE_FFTW=ON")
-            options.append("-DFFTW_ROOT:PATH={0}".format(spec['fftw'].prefix))
+            options.append(f"-DFFTW_ROOT:PATH={spec['fftw'].prefix}")
 
         return options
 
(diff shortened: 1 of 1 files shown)


TBird2001 (approved):
Looks good to me.

repos/spack_repo/builtin/packages/xyce/package.py
```diff
@@ -174,9 +176,16 @@ def cmake_args(self):
...
+            options.append(f"-Dpybind11_DIR:PATH={pybind11.prefix}")
+            options.append(f"-DPython_ROOT_DIR:FILEPATH={python.prefix}")
             options.append("-DPython_FIND_STRATEGY=LOCATION")
+        if "+verbose" in spec:
+            options.append("-DXyce_VERBOSE_LINEAR=ON")
+            options.append("-DXyce_VERBOSE_NONLINEAR=ON")
+            options.append("-DXyce_VERBOSE_TIME=ON")
+        if "+fftw" in spec:
+            options.append("-DXyce_USE_FFTW=ON")
+            options.append(f"-DFFTW_ROOT:PATH={spec['fftw'].prefix}")
```
cmelone: if possible, would be nice to change these conditionals to the new convention of

```
if self.spec.satisfies("+variant_name"):
```

cmelone (commented):
small nit, but looks good to me

PR #6435, opened 2026-09-11
OpenBLAS: add `no_avx512` variant

pzehner:
Working on the introduction of Spack into the [thermo-hydraulics solver Trust](https://gitlab.com/cea-trust-platform/trust/trust-code), developped at CEA, we systematically used to pass the `NO_AVX512` flag when building OpenBLAS. Our early work adopting Spack showed that some of our own tests are failing when the flag is not passed (note that OpenBLAS' own unit tests pass indifferently with or without the flag). We checked that, in the package, the flag is passed for some micro-architectures (e.g. Skylake), but not for all of them (e.g. Zen5).

We see two ways to circumvent this problem: first, by supporting more architectures in the package file, which would pass the `NO_AVX512` flag automatically when needed. This would certainly be the best way, but it's very possible that not all architectures will be taken into consideration. Second, by adding a variant, disabled by default, to enforce the `NO_AVX512` flag. This seems reasonable, as this would be a variant reflecting a valid, documented flag of OpenBLAS.

This PR proposes the second way, but I am open to discuss about a better way.

@mathomp4, @sethrj

### DIFF
--- a/repos/spack_repo/builtin/packages/openblas/package.py
+++ b/repos/spack_repo/builtin/packages/openblas/package.py
@@ -153,6 +153,8 @@ class Openblas(CMakePackage, MakefilePackage):
             when=_when_condition,
         )
 
+    variant("no_avx512", default=False, description="Disable AVX-512 vectorization")
+
     # virtual dependency
     provides("blas", "lapack")
     provides("lapack@3.9.1:", when="@0.3.15:")
@@ -535,6 +537,10 @@ class MakefileBuilder(makefile.MakefileBuilder):
 
         else:
             args.append("TARGET=" + microarch.name.upper())
+
+        # Disable AVX-512 vectorization of x86_64 architectures if requested
+        if self.spec.satisfies("+no_avx512") and self.spec.target.family == "x86_64":
+            args.append("NO_AVX512=1")
 
         return args
 


sethrj (changes requested):
@pzehner The use of AVX512 is closely related to the "microarchitecture" flags spack uses. What spec (including build target and especially `+dynamic` flag) are you using that causes the failures? I suspect that your use case should be represented without a new variant by `openblas target=x86_64_v4 ~dynamic_dispatch`. **Can you verify whether this meets your needs**? A new variant for a specific  microarch case (which won't apply to arm, riscv, etc.) is not ideal. 

@haampie generated this visualization of @alalazo 's `archspec` database:
uarch_all

The OpenBLAS code tries to set build flags analogously to archspec, and once upon a time I tried (in the openblas recipe) to map the spack targets to the openblas flags. They've mutated quite a lot since then, including adding the dynamic flag.

pzehner:
@sethrj I tried `openblas target=x86_64_v4 ~dynamic_dispatch`, this does not solve my problem.

This is the line I use:

```
openblas@0.3.29 +fortran threads=none
```

I understand that adding a variant like this is not ideal. I'd be happy to find a better way.

About the `archspec` database, I cannot see a high-res version of the image.

sethrj:
@pzehner  Can you post the concretized spec (`spack find -lv openblas`) for the version that's causing you problems?

sethrj:
And can you also see whether the cmake-built version (which omits all the hacky arch stuff) has the same issues?

mathomp4:
> And can you also see whether the cmake-built version (which omits all the hacky arch stuff) has the same issues?

I just looked and I think the CMake builder would need a similar thing a la:
```python
        if self.spec.satisfies("+no_avx512") and self.spec.target.family == "x86_64":
            cmake_defs += [self.define("NO_AVX512", "ON")]
```

I think?

sethrj:
Yes it would,  but I want to know if the current spack develop with `openblas@0.3.29 +fortran ~dynamic_dispatch threads=none build_system=cmake` ends up enabling AVX512.

pzehner:
@sethrj Here you go for the legacy build:

```
fjs2gfk openblas@0.3.29~bignuma~consistent_fpcsr~dynamic_dispatch+fortran~ilp64+locking+pic+shared~static build_system=makefile symbol_suffix=none threads=none
```

And here it is using CMake:

```
5gjb2kj openblas@0.3.29~bignuma~consistent_fpcsr+dynamic_dispatch+fortran~ilp64~ipo+locking+pic+shared~static build_system=cmake build_type=Release generator=make symbol_suffix=none threads=none
```

it still doesn't change the outcome of the test.

sethrj:
I'm also asking about the arch/target which is usually the line above what you pasted?

pzehner:
Oh right, my bad:

```
-- linux-opensuse16-zen5 / %c,cxx,fortran=gcc@15.2.0 ------------
```

pzehner:
Wait, with:

```
-- linux-opensuse16-zen5 / %c,cxx,fortran=gcc@15.2.0 ------------                                              
vparvah openblas@0.3.29~bignuma~consistent_fpcsr~dynamic_dispatch+fortran~ilp64~ipo+locking+pic+shared~static build_system=cmake build_type=Release generator=make symbol_suffix=none threads=none
```

the test passes!

sethrj:
OK, so you're building with `zen5` as the architecture target, which *does* natively have avx512. Can you try building for the generic target `x86_64_v3`, which does *not* have avx512 instructions?
This would look like:
```
spack install openblas target= x86_64_v3
```

pzehner:
I mean, with `~dynamic_dispatch build_system=cmake`, I have the expected behavior on my test, without having to pass `NO_AVX512`.

sethrj:
CMake might be working because it sets "TARGET=GENERIC" whereas I think your zen5-optimized build probably has another target. Please post the first couple of lines of output of your the makefile-based openblas installation that fails.

pzehner:
> ```
> spack install openblas target= x86_64_v3
> ```

Do you mean with the other variants as well:

```
spack install openblas target=x86_64_v3 build_system=cmake ~dynamic_dispatch +fortran threads=none
```

?

sethrj:
Just the makefile one; the cmake build system basically ignores the target.
```
spack install openblas target=x86_64_v3 build_system=make ~dynamic_dispatch +fortran threads=none
```

pzehner:
Ok, so for `+fortran threads=none ~dynamic_dispatch target=x86_64_v3`, the build log contains:

```
==> openblas: Executing phase: 'build'
==> [2026-09-11-17:14:51.023746] /usr/bin/make -s CC=/volatile/<<<REDACTED>>>/spack/opt/linux-zen5/compiler-wrapper-1.1.0-xlfr27z62kzmqik4rnm3ai7gugxmrapo/libexec/spack/gcc/gcc FC=/volatile/<<<REDACTED>>>/spack/opt/linux-zen5/compiler-wrapper-1.1.0-xlfr27z62kzmqik4rnm3ai7gugxmrapo/libexec/spack/gcc/gfortran MAKE_NB_JOBS=0 ARCH=x86_64 DYNAMIC_ARCH=1 DYNAMIC_OLDER=1 TARGET=GENERIC NO_AVX512=1 USE_LOCKING=1 USE_OPENMP=0 USE_THREAD=0 RANLIB=ranlib                                                                                                            
```

I guess what you are looking for is:

```
MAKE_NB_JOBS=0 ARCH=x86_64 DYNAMIC_ARCH=1 DYNAMIC_OLDER=1 TARGET=GENERIC NO_AVX512=1 USE_LOCKING=1 USE_OPENMP=0 USE_THREAD=0 RANLIB=ranlib 
```

and the test passes as well.

But this is a generic build.

sethrj:
Great. Compare that to your current build. I think that since your CFD solver appears sensitive to issues caused by AVX512 instructions, you should build it and your dependsnts (via spack environment or config) with a more "generic" target in all case. Letting spack do its default build of `zen5` is allowing avx512 instructions to be generated by the compiler.

pzehner:
I'm not sure this is a good strategy. The other parts of the solver and the other dependencies just work fine when built with a more specialized target. I think just building OpenBLAS with a more generic target alone would be enough.

sethrj:
I'm guessing the BLAS calls are going to be the most likely to have to deal with edge cases like overflow, nans, infs, … and therefore have differences based on the avx instructions. So yeah likely overkill if it's a numerical difference, but if you're actually trying to run the resulting binary on other systems you might need to avoid the avx instructions if they're resulting in SIGILL.

pzehner:
The code is open source, so we usually recompile it on the place of execution.

PR #6436, opened 2026-09-11
acts: depends on hepmc3 +rootio when necessary

wdconinc:
This PR updates `acts` to depend correctly on the `hepmc3 +rootio` variant when `@45.1: +examples +root` as added in https://github.com/acts-project/acts/commit/5174d7662519e3177221881198d336bfa368f30e. This avoids configuration failures like:
```
> CMake Error at CMakeLists.txt:686 (find_package):
    Found package configuration file:
  
      /home/runner/work/eic-spack/eic-spack/spack/opt/spack/linux-x86_64_v3/hepmc3-3.3.1-nx2b463k6vxppfwoqmrssxbibcusnnjh/share/HepMC3/cmake/HepMC3Config.cmake
  
    but it set HepMC3_FOUND to FALSE so package "HepMC3" is considered to be
    NOT FOUND.
```


alecbcs (approved):

PR #6442, opened 2026-09-11
Fix package builds for xcode tools v27 when on MacOS v26

alecbcs:
XCode 27 launched to the MacOS 26 users yesterday and immediately broke a large swath of Spack builds for me.

I traced this down to the MacOS 27 SDK which exposes `dup3` and `pipe2` implementations. However if you're using the newer version of xcode tools on an older MacOS (which to be clear is very much supported) the SDK incorrectly exposes these implementations which are not available. Thus we got segfaults and linker errors when attempting to make use of them. 

This PR patches the autotools builder, Perl, and Python packages to explicitly fall back to the portable implementation for the two methods that is packaged with autotools itself. This got all of the packages in my default development environment back into a building state.

### DIFF
--- a/repos/spack_repo/builtin/build_systems/autotools.py
+++ b/repos/spack_repo/builtin/build_systems/autotools.py
@@ -335,6 +335,15 @@ To resolve this problem, please try the following:
         """
         os.environ["FORCE_UNSAFE_CONFIGURE"] = "1"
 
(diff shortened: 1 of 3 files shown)


alecbcs:
@spackbot run pipeline

alecbcs:
@spackbot run pipeline 😢

cmelone (approved):

PR #6446, opened 2026-09-11
esmf: add llvm-openmp dependency for Apple Clang

mathomp4:
In ESMF 9.0+, `Darwin.gfortranclang.default` honors `ESMF_OPENMP=ON` rather than forcing `ESMF_OPENMP=OFF` (as previous versions did).

When building with `%apple-clang` and `+openmp`, Apple Clang does not bundle OpenMP runtime headers or libraries, leading to missing `<omp.h>` and `-lomp` errors when compiling ESMF C++ source files.

This PR:
1. Adds `depends_on("llvm-openmp", when="@9: +openmp %apple-clang")`.
2. Injects the include directory and linker flags (`openmp.headers.include_flags`, `openmp.libs.ld_flags`) into `ESMF_CXXCOMPILEOPTS`, `ESMF_CXXLINKOPTS`, `ESMF_F90LINKOPTS`, and `ESMF_SL_LIBOPTS` in `setup_build_environment`. The Fortran compile configuration already supplies `-fopenmp`; the added Fortran link flags explicitly locate `libomp` for ESMF tests and applications.

> **Note on versioning:**
> This change is a complete **no-op for ESMF 8** (which continues to not pull in `llvm-openmp`). While ESMF 9.0.0 has not yet been officially tagged as a final release, this has been verified with `esmf@develop` and recent 9.0.0 beta snapshots. We wanted to get this PR upstream now so the fix is in place ahead of the upcoming ESMF 9.0 release and not lost.

Tested on macOS (Apple Silicon, Apple Clang + GCC 16 gfortran):
- `esmf@develop +openmp %apple-clang` builds and installs cleanly.
- Downstream packages (MAPL) compile successfully against the resulting ESMF installation.
- `spack style -b develop` is completely clean.

### DIFF
--- a/repos/spack_repo/builtin/packages/esmf/package.py
+++ b/repos/spack_repo/builtin/packages/esmf/package.py
@@ -108,6 +108,7 @@ class Esmf(MakefilePackage, PythonExtension):
     depends_on("fortran", type="build")  # generated
 
     # Optional dependencies
+    depends_on("llvm-openmp", when="@9: +openmp %apple-clang", type=("build", "run"))
     depends_on("mpi", when="+mpi")
     depends_on("lapack@3:", when="+external-lapack")
     depends_on("netcdf-c@3.6:", when="+netcdf")
@@ -344,6 +345,11 @@ class MakefileBuilder(makefile.MakefileBuilder):
 
         if spec.satisfies("+openmp"):
             env.set("ESMF_OPENMP", "ON")
+            if "llvm-openmp" in spec:
+                openmp = spec["llvm-openmp"]
+                env.append_flags("ESMF_CXXCOMPILEOPTS", openmp.headers.include_flags)
+                env.append_flags("ESMF_CXXLINKOPTS", openmp.libs.ld_flags)
+                env.append_flags("ESMF_SL_LIBOPTS", openmp.libs.ld_flags)
         else:
             env.set("ESMF_OPENMP", "OFF")
 
--- a/repos/spack_repo/builtin/packages/esmf/package.py
+++ b/repos/spack_repo/builtin/packages/esmf/package.py
@@ -108,7 +108,7 @@ class Esmf(MakefilePackage, PythonExtension):
     depends_on("fortran", type="build")  # generated
 
     # Optional dependencies
-    depends_on("llvm-openmp", when="@9: +openmp %apple-clang", type=("build", "run"))
+    depends_on("llvm-openmp", when="@9: +openmp %apple-clang")
     depends_on("mpi", when="+mpi")
     depends_on("lapack@3:", when="+external-lapack")
     depends_on("netcdf-c@3.6:", when="+netcdf")


repos/spack_repo/builtin/packages/esmf/package.py
```diff
@@ -344,6 +345,11 @@ def setup_build_environment(self, env: EnvironmentModifications) -> None:
 
         if spec.satisfies("+openmp"):
             env.set("ESMF_OPENMP", "ON")
+            if "llvm-openmp" in spec:
+                openmp = spec["llvm-openmp"]
+                env.append_flags("ESMF_CXXCOMPILEOPTS", openmp.headers.include_flags)
+                env.append_flags("ESMF_CXXLINKOPTS", openmp.libs.ld_flags)
```
theurich, line 351: Seems we should also add the equivalent for Fortran, `ESMF_F90COMPILEOPTS` and `ESMF_F90LINKOPTS`? While there is no explicit `use omp_lib` in any of the Fortran library sources, there are some use statements in the bundled tests. Plus strictly speaking I guess the Fortran link options should also know how to satisfy the OpenMP dependency during linking (although in your testing (and probably all practical cases) indirectly satisfied via `libesmf.dylib`).

theurich (commented):
I don't have a system to directly test this myself, but from some research on how LLVM and AppleClang in particular handles OpenMP, this all makes sense to me.

Please see one comment/suggestion.

theurich:
@mathomp4 I am wondering now whether we should make these changes through https://github.com/JCSDA/spack-packages/pull/86? Once ESMF 9.0.0 is officially released, we would then open a PR into upstream spack-packages that brings all the good stuff from JCSDA/spack-packages under ESMF package. What do you think?

repos/spack_repo/builtin/packages/esmf/package.py
mathomp4: Oh yeah. I talked with GPT and it said we should add:
```
                env.append_flags("ESMF_F90LINKOPTS", openmp.libs.ld_flags)
```
for linking purposes.

mathomp4:
> @mathomp4 I am wondering now whether we should make these changes through [JCSDA#86](https://github.com/JCSDA/spack-packages/pull/86)? Once ESMF 9.0.0 is officially released, we would then open a PR into upstream spack-packages that brings all the good stuff from JCSDA/spack-packages under ESMF package. What do you think?

@theurich That might be smart. Can you add these bits to https://github.com/JCSDA/spack-packages/pull/86 (or I can). Then when it's time for 9.0.0 final, we can just use the JCSDA package back to here. I can then close this PR

repos/spack_repo/builtin/packages/esmf/package.py
theurich: I think there should also be
```
env.append_flags("ESMF_F90COMPILEOPTS", openmp.headers.include_flags)
```
I think any code with explicit `use omp_lib` would need that to find the Fortran `omp_lib` module.

theurich:
> > @mathomp4 I am wondering now whether we should make these changes through [JCSDA#86](https://github.com/JCSDA/spack-packages/pull/86)? Once ESMF 9.0.0 is officially released, we would then open a PR into upstream spack-packages that brings all the good stuff from JCSDA/spack-packages under ESMF package. What do you think?
> 
> @theurich That might be smart. Can you add these bits to [JCSDA#86](https://github.com/JCSDA/spack-packages/pull/86) (or I can). Then when it's time for 9.0.0 final, we can just use the JCSDA package back to here. I can then close this PR

@mathomp4 Okay, once we are happy with the changes here, I will cherry-pick from your PR branch here into [JCSDA#86](https://github.com/JCSDA/spack-packages/pull/86).

repos/spack_repo/builtin/packages/esmf/package.py
mathomp4: `llvm-openmp` does not provide the Fortran `omp_lib` module: its installed `include/` directory contains only `omp.h`, `omp-tools.h`, `ompt.h`, and `ompx.h`. For this configuration, ESMF invokes `gfortran` for Fortran sources, and `ESMF_OPENMP_F90COMPILEOPTS` already supplies `-fopenmp`; that flag makes gfortran locate its own intrinsic `omp_lib.mod` (from the GCC installation).

Therefore adding `openmp.headers.include_flags` to `ESMF_F90COMPILEOPTS` would not make `use omp_lib` work. I retained the `ESMF_F90LINKOPTS` addition because ESMF does not otherwise add the separate LLVM `libomp` location to its Fortran link commands.

mathomp4: Hmm. Terra seems to think:

> The requested Fortran compile flag should not be added.
> 
> Evidence from the installed toolchain:
> - llvm-openmp/include contains only C/C++ headers: omp.h, omp-tools.h, ompt.h, and ompx.h.
> - It does not supply omp_lib.mod.
> - ESMF’s Darwin.gfortranclang configuration already provides -fopenmp through ESMF_OPENMP_F90COMPILEOPTS.
> - gfortran -fopenmp finds its own intrinsic omp_lib.mod; its module/library implementation is paired with GCC’s libgomp, not LLVM’s libomp.
> I left the committed ESMF_F90LINKOPTS addition in place because ESMF’s Fortran link commands do need the separate LLVM libomp location in this Apple Clang configuration.
> I replied to the review thread with this explanation: https://github.com/spack/spack-packages/pull/6446#discussion_r4017228552
> Validation: spack spec --fresh 'esmf@develop +openmp %apple-clang' succeeds and the worktree remains clean.

mathomp4: oh weird. Terra commented as me. Creepy :)

theurich: Okay, good argument for `Darwin.gfortranclang.default`, which this PR is for... I lost sight of that. But then now I wonder about `Darwin.llvm.default`.

theurich: > oh weird. Terra commented as me. Creepy :)

Very weird. Scary in fact!

theurich: @mathomp4 what do you think about testing this PR on Alderaan for the `Darwin.llvm.default` config with `+openmp`? Would be good to cover all cases that use Apple Clang.

mathomp4: > @mathomp4 what do you think about testing this PR on Alderaan for the `Darwin.llvm.default` config with `+openmp`? Would be good to cover all cases that use Apple Clang.

I'll give that a try. One moment...

mathomp4: @theurich So I was able to cleanly install as:
```
spack install esmf +openmp ~mpi
```
(I didn't have openmpi built yet and didn't want to wait for it so `mpiuni` won)

Per Terra looking at things:

> Yes. esmf.mk looks correct for this mixed Apple Clang + LLVM Flang OpenMP build.
> 
> - ESMF_COMPILER=llvm, selecting ESMF’s Darwin.llvm.default configuration.
> - C/C++ uses Apple Clang with -Xpreprocessor -fopenmp.
> - C++ compilation has the Spack llvm-openmp include path.
> - C++ linking has -L.../llvm-openmp-21.1.8/lib -lomp.
> - Fortran uses Homebrew LLVM Flang 23.1.1 with -fopenmp.
> - Fortran linking has that same Spack llvm-openmp library path and -lomp.
> - The generated shared-library link option includes the same libomp directory and library.
> 
> The installed libesmf.dylib also confirms the result:
> 
> - It depends on @rpath/libomp.dylib.
> - It embeds an rpath to the Spack llvm-openmp@21.1.8 lib directory.
> - It depends on Flang’s libflang_rt.runtime.dylib.
> - Both libesmf.dylib and libomp.dylib are native arm64 Mach-O binaries.
> - OpenMP symbols used by ESMF, including omp_get_max_threads, are provided by that libomp.
> 
> The only noteworthy detail is the tool-version combination: Homebrew Flang is LLVM 23.1.1, while the Spack llvm-openmp dependency is 21.1.8. The installed OpenMP ABI reports compatibility version 5.0.0, and the successful build plus correct dynamic linkage show no current incompatibility. It is not a malformed flag or missing-runtime situation.

So we might be good with LLVM Flang

theurich: Thanks for running this test. I am fine moving forward, too.

theurich:
Okay, your changes from this PR are now cherry-picked into https://github.com/JCSDA/spack-packages/pull/86.

mathomp4:
> Okay, your changes from this PR are now cherry-picked into [JCSDA#86](https://github.com/JCSDA/spack-packages/pull/86).

@theurich Then I'll close this until 9.0.0 is out.

PR #6454, opened 2026-09-12
paraview: add v6.2.0-RC2

cquammen:


kwryankrattiger (commented):
For the RCs we usually only leave the latest one and replace the older one.

cquammen:
> For the RCs we usually only leave the latest one and replace the older one.

I wondered - old RC removed

kwryankrattiger (approved):

PR #6459, opened 2026-09-13
keynub-licdongle: new package

ab-tools:
Adds `keynub-licdongle` 1.1.1, the SDK for the KeyNub USB-C license dongle (https://www.keynub.com/): the C API, a header-only C++11 wrapper and a flat companion API.

The library ships prebuilt in the upstream repository (`natives/<os>-<arch>/`), so the recipe installs the headers and the shared and static libraries for the target platform and compiles nothing. `requires(policy="one_of")` limits it to Linux, macOS and Windows on x86_64 and aarch64.

Checked: `spack audit packages keynub-licdongle` passes; `spack install keynub-licdongle` on Linux x86_64 installs headers, libraries and licence files, and a C program compiled against the prefix links and prints the version.


haampie:
Hi, is there any particular reason you cannot distribute the source code?

I would not be inclined to merge it: most packages are added because of user demand, this looks rather promotional, and a binary blob does not help the case.

ab-tools:
Hello Harmen,

first thanks a lot for reviewing my PR with a personal comment!

Everything a developer touches is open source under Apache-2.0: the C header, the C++ wrapper, all language bindings, samples and tests. Only the core library that speaks to the dongle's secure element ships prebuilt. This is a copy-protection product, and that core is exactly what an attacker patches to bypass a check, so we keep it as one vetted, signed build for every platform rather than a source tree anyone can modify and redistribute as "the" library.

As we already have several scientific developers using our dongle for software licensing, the Spack package manager felt like the right place for our library, too, and our library is also listed in several other professional and currated package registers for the scientific community like [MATLAB File Exchange](https://www.mathworks.com/matlabcentral/fileexchange/184704-keynub-license-dongle-for-matlab-and-simulink) or [Julia Hub](https://platform.juliahub.com/ui/Packages/General/KeyNubLicenseDongle).

We would therefore really appreciate if you could merge the PR.

Best regards and thanks for your consideration
Andreas

PR #6460, opened 2026-09-13
bali-phy: new package

bredelings:
Add new package BAli-Phy (version 4.3) for Bayesian inference of phylogenies and alignments.  The web page is at https://www.bali-phy.org.  The build system is meson.

Local checks:
- Full source build and installation on Debian x86-64 with GCC 16.2.
- Spack package audits and import/format/lint checks passed.
- `spack test run` passed the installed 20-iteration analysis test.


bredelings:
Hi, can someone approve the CI checks here?

repos/spack_repo/builtin/packages/bali_phy/package.py
```diff
@@ -0,0 +1,73 @@
...
+from spack.package import (
+    Executable,
+    conflicts,
+    depends_on,
+    license,
+    maintainers,
+    version,
+    which,
+    working_dir,
+)
```
cmelone: use `from spack.package import *`

repos/spack_repo/builtin/packages/bali_phy/package.py
```diff
@@ -0,0 +1,73 @@
...
+
+
+class BaliPhy(MesonPackage):
+    """Bayesian co-estimation of phylogenies and multiple sequence alignments."""
+
+    homepage = "https://www.bali-phy.org"
+    url = "https://github.com/bredelings/BAli-Phy/archive/refs/tags/4.3.tar.gz"
+    maintainers("bredelings")
+    license("GPL-2.0-or-later")
```
cmelone:
```suggestion

    maintainers("bredelings")
    
    license("GPL-2.0-or-later")
```

repos/spack_repo/builtin/packages/bali_phy/package.py
```diff
@@ -0,0 +1,73 @@
...
+    depends_on("cli11@2.6.1:", type="build")
+    depends_on("eigen@3.4:", type="build")
+    depends_on("cereal", type="build")
+    depends_on("fmt@12: ~shared")
+    depends_on("xxhash")
+    depends_on("zstd libs=static")
+    depends_on("utf8proc")
+    depends_on("cairo +pdf +png +svg +ft +fc")
+    depends_on("python@3:", type=("build", "run"))
+    depends_on("r", type="run")
```
cmelone: please consider using variants to guard optional dependencies. for instance, I see that there are a few R scripts in the repo, but is this something that is needed by every build?

repos/spack_repo/builtin/packages/bali_phy/package.py
```diff
@@ -0,0 +1,73 @@
...
+        """Run a short seeded analysis using the installed examples and model libraries."""
+        with tempfile.TemporaryDirectory() as directory:
+            with working_dir(directory):
+                bali_phy = Executable(str(Path(self.prefix.bin) / "bali-phy"))
+                bali_phy(
+                    str(Path(self.prefix.share.doc) / "bali-phy/examples/5S-rRNA/5d.fasta"),
+                    "--iterations=20",
+                    "--seed=12345",
+                )
+                assert Path("5d-1/C1.log").is_file()
```
cmelone: spack manages a testing stage dir, a tmpdir isn't necessary

```suggestion
        bali_phy = which("bali-phy", path=self.prefix.bin, required=True)
        fasta = join_path(self.prefix.share.doc, "bali-phy", "examples", "5S-rRNA", "5d.fasta")
        bali_phy(fasta, "--iterations=20", "--seed=12345")
        assert os.path.isfile(join_path("5d-1", "C1.log"))
```

cmelone (changes requested):

repos/spack_repo/builtin/packages/bali_phy/package.py
bredelings: Done.

repos/spack_repo/builtin/packages/bali_phy/package.py
bredelings: Done.

repos/spack_repo/builtin/packages/bali_phy/package.py
bredelings: Done.

repos/spack_repo/builtin/packages/bali_phy/package.py
bredelings: R is used to generate some important plots, so it should normally be installed.  I added a variant where R is installed by default, but its possible to request not installing R.

bredelings:
Thanks for the review -- I implemented these suggestions.  For R, generally you need it for the summary step, but there are situations where you might not want to install it.  So I added a variant where R is installed by default, but you can choose to install without R.

repos/spack_repo/builtin/packages/bali_phy/package.py
```diff
@@ -0,0 +1,65 @@
...
+    # Without R, summary reports omit some convergence diagnostics and plots.
+    variant(
+        "r", default=True, description="Include R for convergence diagnostics and summary plots"
+    )
+
+    depends_on("c", type="build")
+    depends_on("cxx", type="build")
+    depends_on("meson@1.6:", type="build")
+    depends_on("cmake", type="build")  # Discover Cereal and CLI11's CMake metadata.
+    depends_on("pandoc", type="build")
```
cmelone: `doc` variant?

repos/spack_repo/builtin/packages/bali_phy/package.py
```diff
@@ -0,0 +1,65 @@
...
+    def setup_build_environment(self, env):
+        env.set("BOOST_ROOT", self.spec["boost"].prefix)
+
+    # Meson treats Cairo as optional; require it so the package includes drawing tools.
+    def meson_args(self):
+        which("pkg-config", required=True)("--cflags", "--libs", "cairo")
+        return ["-Db_ndebug=true", "-Dwith-mpi=false", "-Dextra-tools=true"]
+
+    # Exercise installed model files and MCMC; --version alone misses incomplete installations.
+    # Keep this small package smoke test alongside the upstream analysis tests.
```
cmelone:
```suggestion
```

repos/spack_repo/builtin/packages/bali_phy/package.py
```diff
@@ -0,0 +1,65 @@
...
+    conflicts("%cxx=llvm@:17", msg="BAli-Phy requires Clang 18 or newer")
+    conflicts("%cxx=apple-clang@:15", msg="BAli-Phy requires Apple Clang 16 or newer")
+
+    def setup_build_environment(self, env):
+        env.set("BOOST_ROOT", self.spec["boost"].prefix)
+
+    # Meson treats Cairo as optional; require it so the package includes drawing tools.
+    def meson_args(self):
+        which("pkg-config", required=True)("--cflags", "--libs", "cairo")
+        return ["-Db_ndebug=true", "-Dwith-mpi=false", "-Dextra-tools=true"]
```
cmelone: re: `-Db_ndebug=true`, the `MesonPackage` class will disable debug by default (`if-release`), so let's remove this. users can control this via the implicit `buildtype` variant

cmelone (changes requested):

repos/spack_repo/builtin/packages/bali_phy/package.py
```diff
@@ -0,0 +1,65 @@
...
+    depends_on("r", type="run", when="+r")
+
+    conflicts("%cxx=gcc@:12", msg="BAli-Phy requires C++23 support (GCC 13 or newer)")
+    conflicts("%cxx=llvm@:17", msg="BAli-Phy requires Clang 18 or newer")
+    conflicts("%cxx=apple-clang@:15", msg="BAli-Phy requires Apple Clang 16 or newer")
+
+    def setup_build_environment(self, env):
+        env.set("BOOST_ROOT", self.spec["boost"].prefix)
+
+    # Meson treats Cairo as optional; require it so the package includes drawing tools.
```
cmelone: if Cairo is indeed required, why not just patch meson.build to require it?

```python
def patch(self):
    filter_file("dependency('cairo', required: false)",
                "dependency('cairo', required: true)", "meson.build", string=True)
```

repos/spack_repo/builtin/packages/bali_phy/package.py
```diff
@@ -0,0 +1,65 @@
...
+    depends_on("meson@1.6:", type="build")
+    depends_on("cmake", type="build")  # Discover Cereal and CLI11's CMake metadata.
+    depends_on("pandoc", type="build")
+    depends_on("boost@1.81: +program_options +random +chrono +json")
+    depends_on("cli11@2.6.1:", type="build")
+    depends_on("eigen@3.4:", type="build")
+    depends_on("cereal", type="build")
+    depends_on("fmt@12: ~shared")
+    depends_on("xxhash")
+    depends_on("zstd libs=static")
```
cmelone: are these static lib constraints requirements or simply preferences? fmt and zstd will build static libs by default

cmelone (commented):
this is looking better, thanks!

please remove the comments on L24 and 28

regarding the patch, we usually prefer they be located outside the spack repo if possible; ie grabbing the patch from a PR, showing that the change will be upstreamed

see: https://github.com/spack/spack-packages/pull/6483/changes#diff-c238a2c41999d44cdd400d39a082ff23c93901c91b793cc79be9d311e104db48R26-R31

otherwise, this looks almost ready to go

cmelone (approved):
LGTM

bredelings:
Thanks for all your help!

PR #6461, opened 2026-09-13
mgis: fix download url of old versions

trsxvz:
MGIS Versions 2.2.0, 2.1.0, 1.2.0 and 1.1.0 cannot be downloaded : Spack builds the archive URL from the version number, but the upstream tags are `MFrontGenericInterfaceSupport-2.2`, `-2.1`, `-1.2`, `-1.1` (which results in an error 404). This fix adds an explicit `url=` to these four versions, as already done for 3.0.0, 2.0.0 and 1.0.0. Checksums are unchanged


kshea21 (approved):
Looks good to me, thanks!

PR #6462, opened 2026-09-13
flex: Fix to not run the host's flex if available

bernhardkaindl:
`flex`: Fix to not run the host's flex if available

This was found when running the build in a Linux LandLock LSM-based sandbox when it denied executing `/usr/bin/flex` during the build of flex.

By disabling bootstrapping and setting `LEX=:`, the `AM_PROG_LEX` probe is disabled and no longer attempts to run the host-installed flex during the build:

```py
configure:14332: checking for flex
configure:14348: found /usr/bin/flex
configure:14359: result: flex
configure:14397: flex conftest.l
..stage-flex-2.6.3-hz7d5zunilpnt6szhs7tcejcoscs5arn/spack-src/configure: line 14399: /usr/bin/flex: Permission denied
configure:14401: $? = 126
configure:14403: checking lex output file root
configure:14414: error: cannot find output from flex; giving up
```

As Landlord returns -EPERM, it is able to uncover such probes which then cause the package to fail because the package(s) do not expect a failure to execute a program they found.


alecbcs (approved):
Looks good to me. Thanks @bernhardkaindl.

Note: I'd really try not to randomly assign or request reviews from folks within merge-to-packages. If you'd like a review more quickly please post in the Spack #pull-requests channel.

PR #6463, opened 2026-09-13
libtiff: When ~zstd is requested, ensure libtiff really disables it

bernhardkaindl:
libtiff: When ~zstd is requested, ensure `libtiff` really disables it

This was found when running the build of `libtiff~zstd` in a Linux LandLock LSM-based sandbox when it denied including the host's `zstd` header and linking with the host's `libzstd` library, failing the build of `libtiff` despite `~zstd`.

With `~std`, the spack recipe passes disabing `zstd` but for some reason, the `cmake`-based build ignores it and fails when linking with the host's libzstd is denied by a Linux LandLock-based LSM sandbox.

When `~std` is satisfied, force-disable `zsted` by passing `zstd` include and library paths that prevent `libtiff` from linking the host's `libzstd`.


adamjstewart:
Has this bug been reported upstream? Would like a way to track progress towards a long-term fix.

PR #6464, opened 2026-09-13
add tfel and mgis snapshot versions

trsxvz:
MFEM-MGIS relies on a private custom spack recipe for its installation. Instead of using this custom local recipe, we want to be able to have a clean, "official" spack-packages support for MFEM-MGIS. However two versions of MFEM-MGIS (versions `1.0.1` and `1.0.4`) depend on development snapshots of TFEL and MGIS rather than on releases. Indeed `1.0.4` was released against dedicated tags of both, and `1.0.1` was released against their master branches at the time.

This PR adds these snapshots as versions of `tfel` and `mgis`, so that an upcoming `mfem-mgis` package can depend on them. The version names sort below every numbered release, so the default versions of `tfel` and `mgis` are unchanged. Also, the PR solves an issue arising with python bindings of some TFEL versions


thelfer:
@trsxvz Seems fine for me, though I don't have the right privileges

thelfer (approved):

trsxvz:
@RikkiButler20 @/ [remove ping because PR was merged] kshea21 Sorry for the direct ping. I posted an RFR on Slack but i haven’t heard back yet, so i thought i would ask here. Would either of you mind checking this PR whenever you have some time ? Sorry again and have a nice day

RikkiButler20 (approved):
Looks good. Thanks for bringing it to my attention.

trsxvz:
> Looks good. Thanks for bringing it to my attention.

Thank you!

PR #6467, opened 2026-09-14
fftx: add v1.3.1 with support for cuda 13

eugeneswalker:
FFTX: Add v1.3.1 with support for CUDA 13.

Question for fftx maintainers: Beyond making the CUDA 13 `conflict` conditional on versions older than v1.3.1-- are there any other changes needed to the configure options to support v1.3.1?


eugeneswalker:
Is Fortran support required by default in the new v1.3.1? I'm seeing these errors when CI tries to build the new version:

* [ fftx@1.3.1 /2hdqrs5wnacrmeo72br5kxphsxcd672t E4S ](https://gitlab.spack.io/spack/spack-packages/-/jobs/24307861)

```
...
>     [spack cc]: Error: SPACK_FC_* variables not set: this usually means a missing `depends_on("fortran", type="build")` in package.py
>     gmake[1]: *** [CMakeFiles/cmTC_f8edc.dir/build.make:81: CMakeFiles/cmTC_f8edc.dir/testFortranCompiler.f.o] Error 1
      gmake[1]: Leaving directory '/tmp/root/spack-stage/spack-stage-fftx-1.3.1-2hdqrs5wnacrmeo72br5kxphsxcd672t/spack-build-2hdqrs5/CMakeFiles/CMakeScratch/TryCompile-TeiuuP'
```

broderickpt:
FWIW, FFTX does nor *require* a Fortran compiler in order to build.  We do have a Fortran example program, that is built if we successfully find a Fortran compiler, but if any pre-requisite to building that example is missing then we skip building it. We have always wanted FFTX to build in the absence of Fortran…

When configuring cmake for the FFTX package in spack it emits messages indicating it couldn’t successfully compile a Fortran.  Standalone, this would mean we skip building the Fortran example and carry on.  Within the spack environment, spack appears to say:

[spack cc]: Error: SPACK_FC_* variables not set: this usually means a missing `depends_on("fortran", type="build")` in package.py

spiralgen (commented):
changes approved

PR #6472, opened 2026-09-15
mp-units: new package

rbberger:
- [x] Requires #6471 

@mpusz FYI


repos/spack_repo/builtin/packages/mp_units/package.py
```diff
@@ -0,0 +1,150 @@
...
+        description="Use C++23 explicit-object syntax instead of CRTP",
+    )
+
+    variant(
+        "contracts",
+        default="gsl-lite",
+        values=("none", conditional("gsl-lite", when="~freestanding")),
+        multi=False,
+        description="Contract checking implementation",
+    )
```
mpusz, line 67: This is only one of the options that should be provided. We also have support for ms-gsl and std.

mpusz (changes requested):
Wow, I must admit that the syntax you use is really nice. I'm really impressed. Thank you for doing this.

repos/spack_repo/builtin/packages/mp_units/package.py
rbberger: I can add `std`, but `ms-gsl` isn't currently available in Spack.

rbberger: @johnwparent said he might have time to add `ms-gsl`

alecbcs (dismissed):
Looks good to me. Feel free to merge when ready @rbberger

repos/spack_repo/builtin/packages/mp_units/package.py
johnwparent: `ms-gsl` coming soon to a Spack near you

johnwparent: See #6651 evidently we had already packaged `ms-gsl` it was just given an unfortunate name. I've renamed and added some newer versions.

rbberger: Updated. Now we can also use ms-gsl.

mpusz (approved):
Thanks for doing it. :-)

johnwparent (approved):

PR #6479, opened 2026-09-15
4c-multiphysics: enable clang and intel support

georghammerl:
- Enable clang and intel oneapi compiler
- Fix versions for `mirco` dependency

Assisted-by: GPT-5.6 Sol


bgoderbauer (approved):
Thanks @georghammerl! I was just wondering, do we link against the same MIRCO for all of this year's releases? Also, I'm not a big fan of the taskflow patch here, but I don't know how to get around it.

georghammerl:
> Thanks @georghammerl! I was just wondering, do we link against the same MIRCO for all of this year's releases? Also, I'm not a big fan of the taskflow patch here, but I don't know how to get around it.

Thanks @bgoderbauer for this hint with mirco. You are right, we need here a different version per release.

bgoderbauer (approved):

georghammerl:
@bernhardkaindl Would you mind having a look at this PR? The changes are intended to address the darwin pipeline failures in my main repository. Many thanks in advance for your review.

PR #6480, opened 2026-09-15
pplacer: Update opam initialization and build process

V-Karch:
`pplacer` was failing to build on aarch64 without these changes


alecbcs (approved):
Looks good to me. Thanks @V-Karch!

PR #6481, opened 2026-09-15
py-equinox: Update python dependency version range

tpadioleau:
If I am not mistaken, the project requires:
- Python ~=3.9 since version [0.11.3](https://github.com/patrick-kidger/equinox/blob/v0.11.3/pyproject.toml)
- Python >=3.9 since version [0.11.7](https://github.com/patrick-kidger/equinox/blob/v0.11.7/pyproject.toml)
- Python >=3.10 since version [0.11.11](https://github.com/patrick-kidger/equinox/blob/v0.11.11/pyproject.toml)

I suggest to merge the two first items with >=3.9 and remove the upper bound.


kshea21:
@abhishek1297, as the maintainer do you agree with these changes?

abhishek1297 (approved):

kshea21 (approved):
Thanks!

PR #6484, opened 2026-09-15
cutensor: Add version 2.8.0.6

RemiLacroix-IDRIS:
After version 2.2.0, Nvidia started distributing separated packages depending on the CUDA version.

Account for that by using the same mechanism as cuDNN recipe, that is adding the CUDA version as a suffix to the cutensor version.


RemiLacroix-IDRIS:
@tehrengruber I see that you contributed to this recipe, maybe you could review this PR?

PR #6488, opened 2026-09-15
seacas: Allow CGNS option for Windows

clintonstimpson:
Also avoids access violation bug in clingo (which was reported to clingo project)


johnwparent (approved):
LGTM, thanks!!

PR #6492, opened 2026-09-16
openmpi: avoid pipe2 on macOS with internal libevent

mathomp4:
## Description

Open MPI 5 vendors Libevent 2.1.12 when built with `+internal-libevent`. Xcode 27 ships the macOS 27 SDK, whose `sys/unistd.h` declares `pipe2()` as available starting with macOS 27. When building on macOS 26 with that SDK, Libevent's Autoconf check successfully compiles and links a `pipe2()` probe. The resulting reference is a weak import, however, because the running macOS 26 system does not provide the symbol.

The resulting Libevent calls the null `_pipe2` weak import at runtime:

```text
MPI_Init
  -> opal_event_init
  -> libevent select_init
  -> evsig_init_
  -> evutil_make_internal_pipe_
  -> pipe2 (null weak import)
```

Set `ac_cv_func_pipe2=no` for Darwin Open MPI 5 builds with the vendored Libevent. This makes its configure step select the existing portable `pipe()` fallback instead.

The condition is limited to `@5 +internal-libevent`, the affected configuration.

This is the same SDK/runtime availability mismatch reported by CPython: https://github.com/python/cpython/issues/153711

I found no existing Open MPI or Libevent issue for this specific failure. Open MPI `main` still vendors Libevent 2.1.12: https://github.com/open-mpi/ompi/blob/main/3rd-party/libevent-2.1.12-stable-ompi.tar.gz

## Verification

Rebuilt Open MPI 5.0.10 with `+internal-libevent` on macOS 26.7 using Xcode 27 / the macOS 27 SDK.

- Confirmed the generated `event-config.h` leaves `EVENT__HAVE_PIPE2` undefined.
- Confirmed `libevent_core` no longer imports `_pipe2`.
- Confirmed ESMF's `ESMF_PrintInfoC`, which previously crashed during `MPI_Init`, completes successfully.
- Confirmed the full MAPL test suite passes using the rebuilt Spack
  environment.

## Diagnosis Assistance

I used GPT-5.6 Terra as an interactive debugging assistant during this investigation. It helped narrow the original MAPL test failures to an independent ESMF/Open MPI initialization crash, inspect the Libevent call path and Mach-O weak import, identify the SDK/runtime availability mismatch for `pipe2()`, and validate the focused Autoconf cache override. The diagnosis and verification above were performed locally against the rebuilt packages and test executables.


alecbcs:
@mathomp4 is this still needed now that https://github.com/spack/spack-packages/pull/6442 has merged?

hppritcha (approved):

mathomp4:
> @mathomp4 is this still needed now that #6442 has merged?

@alecbcs Oh. uh...not sure. So are you saying #6442 does this automatically for us no matter the package? 

Let me try pulling the latest `develop` and trying it out. Maybe I had a stale `develop` from a few days ago?

Also, nowhere in my environment do I set `MACOSX_DEPLOYMENT_TARGET` but in #6442 I see:

```python
        # The macOS 27 SDK exposes these functions as weak imports when
        # targeting older macOS releases. Autoconf link tests mistake that for
        # runtime availability and the resulting binaries call through NULL.
        deployment_target = os.environ.get("MACOSX_DEPLOYMENT_TARGET")
        if self.spec.satisfies("platform=darwin") and deployment_target:
            if int(deployment_target.split(".", 1)[0]) < 27:
                os.environ["ac_cv_func_dup3"] = "no"
                os.environ["ac_cv_func_pipe2"] = "no"
```

Does spack itself set `MACOSX_DEPLOYMENT_TARGET` in the environment for us? Because if not, then I'd never trip this since the `and` would fail.

alecbcs:
Yes #6442 _should_ fix this for all autotools packages since it updates the builder, but sometimes packages call out to autotools directly instead of using the builder methods (like Python).

And yes [Spack sets `MACOSX_DEPLOYMENT_TARGET`](https://github.com/spack/spack/blob/da40c0e8a369a305ea3ca85fb563dc8fb2790dd9/lib/spack/spack/platforms/darwin.py#L65) in a package's build environment.

mathomp4:
> Yes #6442 _should_ fix this for all autotools packages since it updates the builder, but sometimes packages call out to autotools directly instead of using the builder methods (like Python).
> 
> And yes [Spack sets `MACOSX_DEPLOYMENT_TARGET`](https://github.com/spack/spack/blob/da40c0e8a369a305ea3ca85fb563dc8fb2790dd9/lib/spack/spack/platforms/darwin.py#L65) in a package's build environment.

Very cool! I actually have some build instructions for some codes where I expressly say "Do not set this" probably for a reason like this. 

Test build is underway!

mathomp4:
Yep. #6442 seems to fix it. I guess I started my testing of this PR before that merged. I'll close this.

PR #6493, opened 2026-09-16
Initial idea for usages callback within packages to set flags

alecbcs:
This PR adds an initial callback prototype for usages defined on package dependency edges. These changes will be used to test an end-to-end example of enabling a sarif usage on a package that depends on gcc.


becker33 (changes requested):
On one end you use `dependent_spec.name` and on the other you use `dependent_spec`. They should probably both index by name.

alecbcs:
Good catch! I've fixed the asymmetry.

repos/spack_repo/builtin/packages/gcc/package.py
```diff
@@ -783,6 +783,12 @@ def url_for_version(self, version):
             self.gnu_mirror_path = self.gnu_mirror_path.replace("xz", "bz2")
         return super().url_for_version(version)
 
+    def setup_dependent_package(self, module, dependent_spec):
+        if dependent_spec.satisfies(f"%[usages=+sarif]{self.spec}"):
+            dependent_spec["compiler_wrapper"].extra_c_args[dependent_spec.name].append(
```
becker33: I think this needs something to ensure the dict is initialized, and then use `setdefault` to ensure the list is initialized.

becker33 (changes requested):

PR #6417, opened 2026-09-10
kokkos: clean up usage of kokkos_cxx

becker33:
Fixes #3419 

Three separate commits:

1. Make the `kokkos_cxx` variable computation more robust
2. Make `dealii` only rely on it when it's relevant, and otherwise use the spack compiler wrappers
3. Remove the matching trilinos variable in favor of relying directly on kokkos

Issue #3419, opened 2026-02-14
Installation issue: dealii exception in package.py using kokkos_cxx from trilinos


dabele:
### Steps to reproduce the issue

The error occurs when dependencies dealii is installed when all dependencies (in particular trilinos) are already installed. Might happen if building dealii repeatedly for debugging the spack package or because dealii build failed and is retried.

deal.ii tries to use `kokkos_cxx` from the trilinos spec. `kokkos_cxx` falls back to `spack_cxx` if no cuas or other device compiler is used.  But if trilinos is already installed, the `spack_cxx` doesn't exist in trilinos. I guess this is only injected into the spec when building?

<details><summary>spack spec</summary>
<pre>
$ spack spec -I <spec>
 -   dealii@9.6.2+adol-c+arborx+arpack+assimp+cgal~complex~cuda~doc+examples+examples_compile+ginkgo+gmsh+gsl+hdf5~int64~ipo+kokkos+metis+mpi+muparser~nanoflann~netcdf+opencascade~optflags+p4est+petsc+platform-introspection~python+scalapack+simplex+slepc+sundials+symengine+taskflow+threads+trilinos+vtk build_system=cmake build_type=DebugRelease cxxstd=17 generator=make arch=linux-ubuntu22.04-cascadelake %c,cxx,fortran=gcc@11.4.0
[+]      ^adol-c@2.7.2~advanced_branching+atrig_erf~boost+doc+examples~openmp~sparse+stdczero+traceless_refcounting build_system=autotools patches:=715b878 arch=linux-ubuntu22.04-cascadelake %c,cxx=gcc@11.4.0
[+]          ^compiler-wrapper@1.0 build_system=generic arch=linux-ubuntu22.04-cascadelake 
[e]          ^gcc@11.4.0~binutils+bootstrap~graphite~nvptx~piclibs~profiled~strip build_system=autotools build_type=RelWithDebInfo languages:='c,c++,fortran' arch=linux-ubuntu22.04-cascadelake 
[+]          ^gcc-runtime@11.4.0 build_system=generic arch=linux-ubuntu22.04-cascadelake 
[e]          ^gmake@4.3~guile build_system=generic patches:=599f134 arch=linux-ubuntu22.04-cascadelake 
[...]

PR #6494, opened 2026-09-16
kokkos: use compiler wrapper for kokkos_cxx

becker33:
Refactors the `compiler-wrapper` package to expose a public method to compute the wrapper path for a given dependent and language.
Fixes the bug introduced in #6417 that caused `kokkos_cxx` to bypass the Spack compiler wrapper.


tpadioleau:
@spackbot re-run pipeline

tpadioleau:
Hi @becker33, I could do some further testing and I actually still face the issue explained in https://github.com/spack/spack-packages/issues/3419. If I understand correctly the situation, if the compiler wrapper that compiled Kokkos is not present anymore and if a new package depending on this Kokkos uses an other compiler wrapper (because the target is different for example), then the compilation fails.

Based on your branch, making the compiler-wrapper also a link dependency with `depends_on("compiler-wrapper")` in the Kokkos package seems to fix the installation.

PR #6495, opened 2026-09-16
git: make openssh an optional dependency to break cycle

### LINKED ISSUE #49504: Cyclic dependency prevents concretization of openssh with tests
### Steps to reproduce

```console
$ spack env activate --temp
$ spack add openssh
$ spack concretize --fresh --force --test all 
==> Error: failed to concretize `openssh` for the following reasons:
     1. Cannot select a single "version" for package "py-twisted"
```

### Error message

```console
$ spack concretize --fresh --force --test all 
==> Error: failed to concretize `openssh` for the following reasons:
     1. Cannot select a single "version" for package "py-twisted"
```

* **Spack:** 1.0.0.dev0 (b242923a9af5ea2e7fe9214333c60502ef6ba9a2)
* **Python:** 3.13.2
* **Platform:** linux-ubuntu25.04-skylake

### General information

- [x] I have run the failing commands in debug mode and reported the output

ravindra-RKB:
Fixes https://github.com/spack/spack/issues/49504

### Summary
`openssh` (with tests) depends on `py-twisted`. `py-twisted` implicitly pulls in `git` as a build dependency (via `py-hatch-vcs` -> `py-setuptools-scm`). 
Because `git` had an unconditional run dependency on `openssh` for SSH transport support, this created a cyclic dependency (`openssh` -> `py-twisted` -> `git` -> `openssh`), preventing concretization.

This PR makes `openssh` an optional dependency of `git` via a new `+ssh` variant (defaulting to `False`). This breaks the cycle while allowing users to explicitly request Spack's `openssh` if they want a pure Spack-based SSH transport for git.

### General information
- [x] I have run `spack style` and it passes.
- [x] I have run `spack test` and it passes.
- [x] I have searched the issues of this repo and believe this is not a duplicate.


ravindra-RKB:
@jennfshr PTAL

jennfshr (approved):
I've spent a little time considering the virtue/rationale of your solution, and I have decided I agree this is a valid solution. 

Initially, I toyed with the notion of:

```
grep -A 3 openssh ~/.spack/packages.yaml
  openssh:
    externals:
    - spec: openssh@8.0p1
      prefix: /usr
spack concretize --fresh --force --test all
==> Concretized 1 spec:
[e]  lizlhcq  openssh@8.0p1+gssapi build_system=autotools platform=linux os=rhel8 target=x86_64
```

but I now see your point - if this is the only way to successfully concretize, considering the cyclical DAG, then you're not going to be able to successfully depend on a Spack installed openssh package. 

Whereas a way around this is to set a externals spec in `packages.yaml`, the environment then would force deps to use/require a system openssh, which doesn't achieve the desired effect.

Therefore, I agree, this is a viable approach. 

Approved.

haampie (changes requested):
disabling ssh support in git by default is very poor ux.

ravindra-RKB:
@jennfshr if all looks good then merge this

ravindra-RKB:
@haampie You're totally right about the UX. I have updated the PR to keep `+ssh` enabled by default in `git`. 

To break the concretization cycle (`openssh (test) -> py-twisted -> py-hatch-vcs -> py-setuptools-scm -> git -> openssh (run)`), I instead added `depends_on("git~ssh", type="test")` to `openssh`. This ensures that when testing `openssh`, the cycle is avoided, while still giving all regular `git` users SSH support by default. 

@jennfshr @haampie Please take another look!

PR #6499, opened 2026-09-17
py-aiohttp: add v3.13.3

thomas-bouvier:


manuelakuhn (approved):
LGTM and builds fine on rocky8 with gcc@16

thomas-bouvier:
Thanks @manuelakuhn

PR #6500, opened 2026-09-17
acts: add scripts variant to install the python example scripts

madbaron:
This PR adds a `scripts` variant to `acts` (default off, only available with `+python`) that installs the Python example scripts from `Examples/Scripts/Python`.

These scripts (e.g. `material_recording.py`, `material_mapping.py`, `full_chain_odd.py`) are the documented way to use the ACTS Python bindings, and some downstream workflows import functions from them, e.g. `runMaterialRecording`. But ACTS has no install rule for them: `Examples/Scripts/Python` has no `CMakeLists.txt`, so a Spack install of `acts +examples +python` has bindings under `<prefix>/python` but none of the scripts that use them.

With `+scripts`:
- a `run_after("install")` step copies `Examples/Scripts/Python` into `<prefix>/share/acts/Examples/Scripts/Python`
- the run environment sets `ACTS_EXAMPLES_SCRIPTS` to that directory, e.g. `python3 $ACTS_EXAMPLES_SCRIPTS/material_recording.py ...`

The directory exists, with no install rule, in every version the recipe lists (checked v38.0.0, v44.3.0 and `main`), so the variant has no version constraint. Default installs don't change.


wdconinc:
Would you be open to submitting an upstream PR to allow this to be implemented in the Acts build system?

madbaron:
> Would you be open to submitting an upstream PR to allow this to be implemented in the Acts build system?

Yes, I can do that!

alecbcs (approved):
Looks good to me. Waiting on final approval from @stephenswat, @wdconinc.

wdconinc:
If this goes in upstream I'd prefer a patch over a new variant.

PR #6502, opened 2026-09-17
py-jaxlib: fixes for 0.11 +rocm

afzpatel:
I came across a couple errors when trying to build `py-jaxlib@0.11.1+rocm` and `py-jaxlib@0.11.0+rocm`. This includes a missing dependency and an xla bug I had to patch for 0.11.1: https://github.com/openxla/xla/pull/48815


afzpatel:
@spackbot fix style

adamjstewart (dismissed):
I think this looks okay, wish ROCm support was a bit easier. Is the bug only with 0.11.1 and not 0.11.0? Why was this not caught by CI?

We should consider always using Spack-installed XLA instead of constantly having to patch it in several places.

Do you want to add your name to the maintainers list so you get notified when I add a new version and can debug any ROCm-related failures? Could also add your name to the PyTorch and TF recipes for the same reason in a separate PR.

afzpatel:
> I think this looks okay, wish ROCm support was a bit easier. Is the bug only with 0.11.1 and not 0.11.0? Why was this not caught by CI?
> 
> We should consider always using Spack-installed XLA instead of constantly having to patch it in several places.
> 
> Do you want to add your name to the maintainers list so you get notified when I add a new version and can debug any ROCm-related failures? Could also add your name to the PyTorch and TF recipes for the same reason in a separate PR.

Yes, the xla bug is only on 0.11.1.  I actually don't know why CI didn't catch this and I was wondering the same thing. 

Ok I can add my name here for py-jaxlib and I can add it to py-tensorflow to. I haven't worked on pytorch+rocm, @renjithravindrankannath has been working on that.

adamjstewart (dismissed):

afzpatel:
It doesn't seem like py-jaxlib+rocm is being built in ci for some reason. I'm going to try to investigate

afzpatel:
@spackbot rerun pipeline

afzpatel:
ci is passing and is building py-jaxlib+rocm now. I had to remove the additional python dependency for +rocm and I added a `require` to ensure py-jaxlib builds with +rocm. Hopefully this means that future +rocm bugs will be caught by CI.

@adamjstewart I think this is ready to be merged if everything looks good to you.

adamjstewart (approved):

PR #6504, opened 2026-09-17
molds: v2.42.0 and v2.42.1

elfprince13:
Includes various improvements, notably larger (and platform-specific) thunk sizes which prevents linking from frequently failing on aarch64 debug builds


RikkiButler20 (approved):
LGTM

PR #6505, opened 2026-09-17
py-googleapis-common-protos: add v1.52.0 and v1.57.1

thomas-bouvier:


repos/spack_repo/builtin/packages/py_googleapis_common_protos/package.py
RikkiButler20:
```suggestion
        depends_on("py-protobuf@3.6.0:", when="@1.6:")
```
This should include v1.52.0

RikkiButler20 (changes requested):

repos/spack_repo/builtin/packages/py_googleapis_common_protos/package.py
thomas-bouvier: Good catch thanks @RikkiButler20

RikkiButler20 (approved):
LGTM

PR #6509, opened 2026-09-17
gcc, mpt, mvapich2, openglu: fix overlapping version ranges in provides

alalazo:
When several `provides()` conditions hold for the same spec, Spack applies all of them, so the virtual ends up with the tightest version bound. Using `provides("v@:X", when="@Y:")` for increasing Y therefore capped every release at the oldest bound.


repos/spack_repo/builtin/packages/gcc/package.py
```diff
@@ -280,27 +277,29 @@ class Gcc(AutotoolsPackage, GNUMirrorPackage, CompilerPackage):
...
+            ("7.3:7", "1.8.3"),
+            ("8.1", "1.10"),
+            ("8.2:8", "1.10.3"),
+            ("9", "1.12.2"),
+            ("10.1", "1.14.2"),
+            ("10.2", "1.14.4"),
+            ("10.3:10", "1.14.6"),
+            ("11.1", "1.16.3"),
+            ("11.2:11", "1.16.5"),
+            ("12:", "1.18"),
```
haampie: this looks awful, can you restore the declarative form please?

haampie (approved):
thanks

PR #6511, opened 2026-09-17
Fix overlapping depends_on ranges

alalazo:
An upper-bounded dependency declared under an open-ended `when="@P:"` also applies to every newer release. Since overlapping constraints are intersected, a looser bound declared for newer versions was never used. Close the `when` ranges of the older directives.


mathomp4 (approved):

kshea21 (approved):
Thanks!

PR #6528, opened 2026-09-18
flux-accounting: new package

sam-maloney:
@grondo I set myself as the `maintainer` since I was creating the recipe, but wasn't sure whether it makes more sense to just have you listed for uniformity with the other flux packages?


grondo:
Fine for you to be maintainer IMO!

Thanks for doing this?

grondo (approved):

repos/spack_repo/builtin/packages/flux_accounting/package.py
```diff
@@ -0,0 +1,98 @@
...
+    version("0.55.0", sha256="06c57d6a0c21be6c1c330fd934ab8324740e13a8bfb22f64fb701e6b14d11892")
+    version("0.54.0", sha256="46f446581c631b089ecc264b239f54da97c66f98d8a636b24342af659788e474")
+    version("0.53.0", sha256="96a175d090b7348a1caf511bee5d359cd13f7f9118f37e23a5077ec030bf3a76")
+    version("0.52.0", sha256="37b38aa64d25da0ac0301b32a348150f4065cf7d1236f9a247706f7d8c5611e5")
+    version("0.51.0", sha256="695bcdd5b309a868d5a2bcb3edca539a82f23d28c7cb7d1c78a74a68fe20bb81")
+    version("0.50.0", sha256="7d680eac41392914cad91e62c0cb71db4c9cd62eb155b342d64fd2ee2335cff9")
+    version("0.49.0", sha256="384f499215083223f751a363c1bf06d3bafd3f5a21a4360a9ecd8ec7009c3e2b")
+    version("0.48.0", sha256="44e1e7091e10dedf64af5e7dc279e49d9034cdfefff141ac3d014f64a233f71a")
+    version("0.47.0", sha256="2e998b8d60c4779ccad56931e3c606f0c6a5933b9162da4e1539ebf4e429237f")
+    version("0.46.0", sha256="552604291bee427d0e5b126045fcdaf49e8c9aac478c5022971e88b4b1447665")
```
bernhardkaindl, line 42: Hi @sam-maloney, please see this as a general comment, not specifically to flux-accounting.

Of course, you as the future maintainer should be most qualified to answer these questions, these are just suggestion to think about:

Do you need to maintain and support the full list versions, or is it sufficient to carry/support the lastest version(s)?

Spack's policy is to only remove deprecated versions so version accumulate naturally over time, which is why you often see many versions in existing packages, but for for new packages it would normally be most economical to only add the newest version(s).

In general (this may or may not apply to flux-accounting as much) has an impact on e.g. CVE updates/security issues in older versions and needing to take the path to deprecate previously checksummed (so-called "safe" versions) is usually more difficult (needs a PR reviewed and approved) than not adding them in the first place when adding a new package.

Another question: All older versions have exactly the same dependencies and dependency versions?

This is the reason why I'd suggest to only add the latest version(s) or only the versions actively required by users initially, so this suggests to remove the versions which aren't the current latest checksummed version:

```suggestion
```

repos/spack_repo/builtin/packages/flux_accounting/package.py
```diff
@@ -0,0 +1,98 @@
...
+
+    homepage = "https://github.com/flux-framework/flux-accounting"
+    url = "https://github.com/flux-framework/flux-accounting/releases/download/v0.1.0/flux-accounting-0.1.0.tar.gz"
+    git = "https://github.com/flux-framework/flux-accounting.git"
+
+    maintainers("sam-maloney")
+
+    license("LGPL-3.0-only")
+
+    version("main", branch="main")
```
bernhardkaindl, line 23: Do you need the main or master version?
```suggestion
```

For me, it didn't build:
```py
spack install flux-accounting@main
==> Warning: Unable to resolve the git commit for flux-accounting. An installation of this binary won't have complete binary provenance.
[x] esh3ehl flux-accounting@main failed: /tmp/bkaindl/spack-stage/spack-stage-flux-accounting-main-esh3ehlzgaxjwosw7htzee45qbtprjsk-kughgbz6.log (2s)
-- lines 1 to 13 --
    File "/home/bkaindl/spack/pr/1/spack/lib/spack/spack/package_base.py", line 1664, in do_stage
      self.do_fetch(mirror_only)
    File "/home/bkaindl/spack/pr/1/spack/lib/spack/spack/package_base.py", line 1644, in do_fetch
      self.stage.fetch(mirror_only, err_msg=err_msg)
    File "/home/bkaindl/spack/pr/1/spack/lib/spack/spack/stage.py", line 901, in fetch
      stage.fetch(mirror_only, err_msg)
    File "/home/bkaindl/spack/pr/1/spack/lib/spack/spack/stage.py", line 624, in fetch
      raise spack.error.FetchError(
  spack.error.FetchError: All fetchers failed for spack-stage-flux-accounting-main-esh3ehlzgaxjwosw7htzee45qbtprjsk
          https://mirror.spack.io/flux-accounting/flux-accounting-main.tar.gz: DetailedHTTPError: GET https://mirror.spack.io/flux-accounting/flux-accounting-main.tar.gz returned 404: Not Found
      [git] https://github.com/flux-framework/flux-accounting.git on branch main: ProcessError: Command exited with status 128:
      /usr/bin/git -c advice.detachedHead=false clone --quiet --depth 1 --single-branch --branch main --filter=blob:none --no-checkout https://github.com/flux-framework/flux-accounting.git flux-accounting
```

For you as the maintainer of the package it would normally be less work to only support the actual releases in the public version of the recipe, if you add a main or master version, it should be building, at list in principle I think. You can always have a private versoin of the recipe if you want to test the current head of the code.

repos/spack_repo/builtin/packages/flux_accounting/package.py
```diff
@@ -0,0 +1,98 @@
...
+    version("0.50.0", sha256="7d680eac41392914cad91e62c0cb71db4c9cd62eb155b342d64fd2ee2335cff9")
+    version("0.49.0", sha256="384f499215083223f751a363c1bf06d3bafd3f5a21a4360a9ecd8ec7009c3e2b")
+    version("0.48.0", sha256="44e1e7091e10dedf64af5e7dc279e49d9034cdfefff141ac3d014f64a233f71a")
+    version("0.47.0", sha256="2e998b8d60c4779ccad56931e3c606f0c6a5933b9162da4e1539ebf4e429237f")
+    version("0.46.0", sha256="552604291bee427d0e5b126045fcdaf49e8c9aac478c5022971e88b4b1447665")
+
+    variant("docs", default=False, description="Build flux manpages and docs")
+
+    depends_on("c", type="build")  # generated
+    depends_on("cxx", type="build")  # generated
```
bernhardkaindl, line 47: As this is a new package, you should know which c/cxx deps are needed, remove what is not needed (test it) and remove `# generated`.

repos/spack_repo/builtin/packages/flux_accounting/package.py
```diff
@@ -0,0 +1,98 @@
...
+
+    depends_on("pkgconfig", type="build")
+
+    # Need autotools when building on master:
+    depends_on("autoconf", type="build", when="@master")
+    depends_on("automake", type="build", when="@master")
+    depends_on("libtool", type="build", when="@master")
+
```
bernhardkaindl, line 55: Whiele I couldn't build '@main', these say '@master' which is not consistent with '@main', I guess. As I'd propse to remove main/master, I'd propose to remove these too, execept if you are going to support builds from main/master publically and if so I guess the name main/master should match.
```suggestion
```

repos/spack_repo/builtin/packages/flux_accounting/package.py
```diff
@@ -0,0 +1,98 @@
...
+    depends_on("flux-core +security")
+    depends_on("flux-core@0.81: +security", when="@0.54.0:")
```
bernhardkaindl, line 57: Even if you'd decide to not add all versions intiailly, the best practice (as far as I understood) would bet to keep such existing dependency rules.

Just an example question regarding the versions: Older versions can use any versions of `flux-core +security`? (that question would not come up if those versions are not checksummed above)

repos/spack_repo/builtin/packages/flux_accounting/package.py
```diff
@@ -0,0 +1,98 @@
...
+    depends_on("jq", type="test")
+    depends_on("lua", type="test")
+    depends_on("valgrind", type="test")
+    depends_on("which", type="test")
+
+    def setup(self):
+        pass
+
+    @when("@master")
+    def setup(self):
```
bernhardkaindl, line 75: Likewise, '@master' is not listed as version, only 'main' is:
```py
pack install flux-accounting@master
==> Error: No version exists that satisfies these input specs:
    flux-accounting@master
```

repos/spack_repo/builtin/packages/flux_accounting/package.py
```diff
@@ -0,0 +1,98 @@
...
+                git("fetch", "origin")
+            except ProcessError:
+                git("fetch")
+
+    def autoreconf(self, spec, prefix):
+        self.setup()
+        if not os.path.exists("configure"):
+            # Bootstrap with autotools
+            bash = which("bash", required=True)
+            bash("./autogen.sh")
```
bernhardkaindl, line 92: Just an observation:

I didn't test/review this closely, but when running autogen to bootstrap, the autotools need to be dependencies. Those are currently depended on when @master, so they might be available using other dependencies indirectly, but if autoreconf/autogen.sh done using normal package installs (not only for master/main), I think the autoconf/automake/libtool deps should ideally be not just limnited to '@master'.

repos/spack_repo/builtin/packages/flux_accounting/package.py
```diff
@@ -0,0 +1,98 @@
...
+
+    def configure_args(self):
+        args = []
+        if "+docs" not in self.spec:
+            args.append("--enable-docs=no")
+        return args
```
bernhardkaindl, line 98: If the package's configure also supports forcibly enabling docs, this example could be followed to ensure that docs are really enabled with `+docs` (else configure should fail):
```py
grep -e Pack -e enable repos/spack_repo/builtin/packages/viennarna/package.py
from spack_repo.builtin.build_systems.autotools import AutotoolsPackage
class Viennarna(AutotoolsPackage):
    """The ViennaRNA Package consists of a C code library and several
        args = self.enable_or_disable("sse")
```

Fully untested, only apply if tested!
```suggestion
        return [self.enable_or_disable("docs")]
```

bernhardkaindl (commented):
Hi Sam @sam-maloney, can you review my review comment below?

Cc: @grondo, maybe for your input on the versions to support initially as the checksummed (so-called "safe" versions in spack) for flux-accounting

repos/spack_repo/builtin/packages/flux_accounting/package.py
sam-maloney: Thanks for the suggestion! I've applied it (without the surrounding brackets)

repos/spack_repo/builtin/packages/flux_accounting/package.py
sam-maloney: These blocks are copied from the other `flux-*` Spack recipes, so I am somewhat relying on them for correctness; however, I can say that `autogen.sh` only needs to be run when building master (or more generally, when not building from a release tarball) because the release tarballs already include the needed build files pre-generated, so it is not needed for building the checksummed versions.

repos/spack_repo/builtin/packages/flux_accounting/package.py
sam-maloney: The `# generated` comments were present in all the other `flux-*` recipes, so I just copied them that way, but I'm happy to remove them!

repos/spack_repo/builtin/packages/flux_accounting/package.py
sam-maloney: Mea culpa. Most of the Flux repos (including accounting) use a `master` branch, but I copied this line from the `flux-pmix` recipe without noticing the difference (pmix uses a `main` branch). For consistency with the other `flux-*` recipes, I would like to keep this one supporting `master` as well, so this has been fixed, and
```
spack install flux-accounting@master
```
now works in my testing.

repos/spack_repo/builtin/packages/flux_accounting/package.py
sam-maloney: Sounds good to me! I wasn't sure what the best practice was, so I just ran:
```
curl https://api.github.com/repos/flux-framework/flux-accounting/releases | jq '.[] | .tag_name + " " + .assets.[0].digest'
```
and those were the versions that returned a checksum, so they were easy to add as a block, but I don't think there's any real need for old versions (particularly given this recipe didn't exist previously).

repos/spack_repo/builtin/packages/flux_accounting/package.py
sam-maloney: The `v0.81` dependency was the only one explicity set in the config file (as least as far as I could see in the Git history). But as suggested, the older checksummed versions have been removed, so this is mostly just an academic note on compatibility at this point.

repos/spack_repo/builtin/packages/flux_accounting/package.py
sam-maloney: Incorrect naming of `master` version as `main` should now be fixed.

repos/spack_repo/builtin/packages/flux_accounting/package.py
sam-maloney: Incorrect naming of `master` version as `main` should now be fixed, and I do want to keep that support so I am electing to leave these dependencies in place.

alecbcs (approved):
Looks good to me. Thanks @sam-maloney!

@bernhardkaindl I'll wait a few business days before merging in case you want to give a final review. Given the logic you commented about is already in the other flux packages, I don't think we need to modify this package until we decide to refactor the rest.

PR #6535, opened 2026-09-19
add magma variant for superlu_dist

liuyangzhuan:


xiaoyeli (approved):

repos/spack_repo/builtin/packages/superlu_dist/package.py
```diff
@@ -134,6 +138,11 @@ def append_from_variant(*args):
         if "~openmp" in spec:
             append_define("CMAKE_DISABLE_FIND_PACKAGE_OpenMP", True)
 
+        append_from_variant("TPL_ENABLE_MAGMALIB", "magma")
+        if "+magma" in spec:
```
alecbcs:
```suggestion
        if spec.satisfies("+magma"):
```

Let's use `spec.satisfies()` instead of `if x in spec` to keep in line with the newer syntax and avoid potential edge cases around dependency variants.

alecbcs (changes requested):
One quick nitpick. Otherwise looks good to me. Thanks @liuyangzhuan

xiaoyeli (approved):

PR #6536, opened 2026-09-19
mathicgb: add tbb variant

d-torrance:


repos/spack_repo/builtin/packages/mathicgb/package.py
```diff
@@ -27,6 +27,8 @@ class Mathicgb(AutotoolsPackage):
     version("1.1", sha256="c756c2265df23fb7417f073cf09d63f05e093eb8136bf33904cec04eac24d5b3")
     version("1.0.2025.05.13", commit="de139564927563afef383174fd3cf8c93ee18ab3")
 
+    variant("tbb", default=True, description="Enable multithreading with TBB")
```
alecbcs:
```suggestion
    variant("tbb", default=False, description="Enable multithreading with TBB")
```

@d-torrance would it be possible to set the default of this variant to `False` so that we don't modify functionality of folks existing configurations and keep the base package install smaller?

d-torrance: Makes sense -- thanks for the suggestion!

alecbcs (approved):
Looks good to me. Thanks @d-torrance

PR #6550, opened 2026-09-20
yasm: depend on cxx when build system is cmake

Growl1234:
yasm's CMake didn't specify its language (C only); as a result, CMake enables both C and C++ for it by default and cause Spack to report error due to `depends_on("cxx", type="build")` not added (see [our CI report](https://storage.googleapis.com/cp2k-ci/cp2k-spack-pdbg-c65076ad_report.txt)).

This is a workaround, while the real issue should be fixed upstream. The related issue is reported as https://github.com/yasm/yasm/issues/320.


Growl1234:
@RMeli Please check this one. The Spack CI test in https://github.com/cp2k/cp2k/pull/5961 is failing due to this problem.

RMeli (approved):

PR #6555, opened 2026-09-20
py-pyscf: add missing gmake build dependency

gordan12:
`py-pyscf` runs cmake but does not depend on `gmake`, so cmake picks up the system make. A make older than 4.4 cannot parse the fifo-style `--jobserver-auth` string Spack exports in `MAKEFLAGS`, and the build fails at the first build step:

```
gmake: *** internal error: invalid --jobserver-auth string 'fifo:/tmp/tmpvdl5h1ej/jobserver_fifo'.  Stop.
error: command '/usr/bin/cmake' failed with exit code 2
```

RHEL 9 and Ubuntu 22.04 both ship make 4.3. Reproduced and fixed on linux-rhel9-sapphirerapids with `%gcc@11.5.0`.


alecbcs (approved):
Looks good to me. Waiting on final approval from @naromero77.

PR #6556, opened 2026-09-20
glx, opengl: restrict headers/libraries to first found

wdconinc:
When `glx` or `opengl` returns the headers path for the include directories, all instances of GL/gl.h are returned. This can contain false positives. E.g. an external `opengl` located under `/usr` matches both `/usr/include/GL/gl.h` and `/usr/share/emscripten/system/include/GL/gl.h`, where the second one leads to mayhem when included in regular builds.

This PR changes the headers and libraries to only return the first hit, which will find `include` before others.


wdconinc:
Draft since not quite satisfied. I would probably prefer a solution that doesn't pick first, but that sorts by depth and then picks first. @biddisco what are your thoughts on this?

PR #6557, opened 2026-09-20
gzip: add v1.15

michaelkuhn:


kshea21:
@spackbot re-run pipeline

michaelkuhn:
This was failing in CI due to an ARM-specific bug. I've pushed a new version that includes a patch fixing it.

PR #6558, opened 2026-09-20
Add support for cray-libsci in quantum espresso

green-br:
Adds support for Cray Libsci in Quantum Espresso.  This is needed as Cray-Libsci uses multiple libraries.


ye-luo:
Could you create an issue detailing the failure with all the spack output files provided for troubleshoot?
Without much detail, it is hard to understand what this PR actually does.

repos/spack_repo/builtin/packages/quantum_espresso/package.py
```diff
@@ -487,12 +487,16 @@ def cmake_args(self):
             else:
                 plugins.append("pw2qmcpack")
 
-        if "^armpl-gcc" in spec or "^acfl" in spec:
+        if "^armpl-gcc" in spec or "^acfl" in spec or "^cray-libsci" in spec:
```
ye-luo, line 490: cray-libsci is usually implicitly linked under the compiler wrapper and doesn't need to be explicitly set. Is spack doing anything special?

green-br: The Cray Compiler Wrappers are optional in Spack - if you setup Spack on Cray as described in HPE on docs at https://cpe.ext.hpe.com/docs/26.03/craype/spack.html where the Compiler Wrappers are not used then you need to treat the Cray libraries as any other dependency.  I prefer this method since it makes it clearer and behaves as a standard compiler.

ye-luo: I see where you are from. "Cray Compiler Wrappers" are bad in many ways.

ye-luo (approved):

repos/spack_repo/builtin/packages/quantum_espresso/package.py
ye-luo: Do you by chance know if the linked cray-libsci is threaded or not?

green-br: It will depend on the variant to cray-libsci - in quantum espresso it's just asking for lapack so no variant so uses default, cray-libsci package has default openmp off and MPI off.

PR #6561, opened 2026-09-21
Update SWIFTSIM package details and versions

green-br:
Adds FFTW support but required updating in general.


green-br:
@spackbot fix style

PR #6562, opened 2026-09-21
py-llguidance: add new package

thomas-bouvier:


manuelakuhn (approved):
The newest available package is 1.8.0. Do you also want to add this?

Other than that, it looks good to me anb builds fine on rocky8 with gcc@16

thomas-bouvier:
Thanks @manuelakuhn. I am not adding more recent versions for now because I am submitting subsets of this PR: https://github.com/spack/spack-packages/pull/4119

PR #6566, opened 2026-09-21
moab: require eigen@:3

xylar:
Since `eigen@5.0.0` was added, `moab+eigen` (and `+tempest`, which also depends on Eigen) concretizes against Eigen 5 and fails to compile:

```
Eigen/src/Core/util/Macros.h:749:2: error: #error Eigen requires at least c++14 support.
```

MOAB's autotools build does not enable C++14, so this constrains it to `eigen@:3`, which is what MOAB (up to 5.6.0) has been built against. Seen with `moab@5.6.0+eigen+fortran+hdf5+mpi+netcdf+pnetcdf+zoltan+tempest` on Ubuntu 24.04 / GCC 13.3 with spack-packages v2026.06.0.


xylar:
@vijaysm, please take a look.

vijaysm:
Yes this is true. On master, the C++ version should be C++17 and so this should no longer be a problem going forward.

vijaysm (approved):

kshea21 (approved):
Great, thanks!

PR #6571, opened 2026-09-21
e3sm-scorpio: add v1.8.2 to v2.0.3, lowercase Fortran modules with cce

xylar:
Adds the SCORPIO releases since 1.8.1 (1.8.2, 1.9.0–1.9.3, 2.0.0–2.0.3).

Also passes `-em -ef` to the Cray Fortran compiler so that module files are written in lowercase, which SCORPIO's build expects; without it the Fortran interface fails to build with `cce` on Frontier. E3SM has carried both in its Spack fork.


alecbcs (dismissed):
Looks good to me. Waiting on final approval from @altheaden.

xylar:
@altheaden is no longer working in the field.  I don't think she will be in a position to approve.

@altheaden, would you like me to remove you as a maintainer?

repos/spack_repo/builtin/packages/e3sm_scorpio/package.py
```diff
@@ -15,8 +15,17 @@ class E3smScorpio(CMakePackage):
     homepage = "https://e3sm.org/scorpio-parallel-io-library/"
     url = "https://github.com/E3SM-Project/scorpio/archive/refs/tags/scorpio-v1.4.1.tar.gz"
 
-    maintainers("xylar", "altheaden")
+    maintainers("xylar", "andrewdnolan")
```
xylar, line 18: @andrewdnolan, do you agree to be a maintainer here?

I will check with @altheaden, since I don't think she is getting GitHub notifications.

altheaden: Yep, all good, go ahead and remove me!

andrewdnolan: Yes, I agree to be a maintainer here!

PR #6574, opened 2026-09-21
py-pybase64: new package

thomas-bouvier:


repos/spack_repo/builtin/packages/py_pybase64/package.py
```diff
@@ -0,0 +1,23 @@
...
+    pypi = "pybase64/pybase64-1.4.3.tar.gz"
+
+    license("BSD-2-Clause")
+
+    version("1.4.3", sha256="c2ed274c9e0ba9c8f9c4083cfe265e66dd679126cd9c2027965d807352f3f053")
+
+    depends_on("c", type="build")
+
+    depends_on("python@3.8:", type=("build", "run"))
+    depends_on("py-setuptools@80:", type="build")
```
manuelakuhn, line 23: Where does the lower bound come from? In the [`pyproject.toml`](https://github.com/mayeut/pybase64/blob/v1.4.3/pyproject.toml#L2) I don't see any.
Could you either remove it or add a comment where it came from, please?

thomas-bouvier: I probably added this lower bound for a reason, but I can check again if needed. Upstream packages tend to forget about setuptools version bounds, resulting in their own package not building.

PR #6575, opened 2026-09-21
py-pycountry: add new package

thomas-bouvier:


manuelakuhn (approved):
LGTM and builds fine on rocky 8 with gcc@16

thomas-bouvier:
Thank you @manuelakuhn

PR #6591, opened 2026-09-21
abacus: update dependencies and CMake options, add v3.9.0.27

Growl1234:


repos/spack_repo/builtin/packages/abacus/package.py
```diff
@@ -49,35 +55,62 @@ class Abacus(CMakePackage, MakefilePackage):
...
-    depends_on("libxc", when="+libxc")
+    depends_on("mpi", when="+mpi")
+    # Older releases require Cereal for LCAO builds
+    # Current versions use it only with LibRI which is not yet in Spack
+    depends_on("cereal", when="@:2")
+    depends_on("cereal", when="@3:3.10 +lcao")
+    depends_on("fftw precision=float,double +openmp", when="+openmp")
+    depends_on("fftw precision=float,double ~openmp", when="~openmp")
+    depends_on("elpa+openmp", when="+elpa +openmp")
+    depends_on("elpa~openmp", when="+elpa ~openmp")
```
RMeli: Probably you need `elpa+mpi` in both cases, since the ELPA variant is available `when="+mpi"`? It is the default, but it is good to enforce it explicitly if it really required. (I assume `elpa~mpi` does not work, given that ELPA is only enabled `when=+mpi`.)

repos/spack_repo/builtin/packages/abacus/package.py
```diff
@@ -49,35 +55,62 @@ class Abacus(CMakePackage, MakefilePackage):
...
+    depends_on("libxc@5.1.7:", when="+libxc")
```
RMeli, line 78: Is the version range applicable to all (old) versions?

repos/spack_repo/builtin/packages/abacus/package.py
```diff
@@ -49,35 +55,62 @@ class Abacus(CMakePackage, MakefilePackage):
...
+    depends_on("dftd4@4.2:", when="+dftd4")
+    depends_on("pexsi@2.0:", when="+pexsi")
+    depends_on("nlohmann-json@3.12:", when="+json")
     depends_on("openblas", when="build_system=cmake")
+    depends_on("scalapack", when="+mpi build_system=cmake")
     depends_on("mkl", when="build_system=makefile")
-    depends_on("cmake", type="build", when="build_system=cmake")
+    depends_on("cmake@3.16:", type="build", when="build_system=cmake")
```
RMeli, line 92: Is the version range applicable to all the olde versions?

repos/spack_repo/builtin/packages/abacus/package.py
```diff
@@ -49,35 +55,62 @@ class Abacus(CMakePackage, MakefilePackage):
...
     depends_on("fortran", type="build")
-    depends_on("mpi")
-    depends_on("cereal")
-    depends_on("fftw+openmp", when="+openmp")
-    depends_on("fftw~openmp", when="~openmp")
-    depends_on("elpa", when="+elpa")
-    depends_on("libxc", when="+libxc")
+    depends_on("mpi", when="+mpi")
+    # Older releases require Cereal for LCAO builds
+    # Current versions use it only with LibRI which is not yet in Spack
```
RMeli:
```suggestion
    # Current versions use it only with LibRI
```
See #6592.

repos/spack_repo/builtin/packages/abacus/package.py
Growl1234: Yes.

repos/spack_repo/builtin/packages/abacus/package.py
Growl1234: Yes for `@3.9:`. I didn't take a look at `@:2` (they were deprecated), but not sure if they can work with newer versions of libxc.

repos/spack_repo/builtin/packages/abacus/package.py
Growl1234: Maybe I should change this after merging #6592. And then I'll add LibRI/LibComm dependencies here.

repos/spack_repo/builtin/packages/abacus/package.py
Growl1234: ELPA is only used as distributed eigensolver in ABACUS, and in ABACUS CMake `ENABLE_ELPA` is defined as a `cmake_dependent_option` under `ENABLE_MPI`. Without MPI ABACUS won't take use of ELPA. I agree it is better to clarify `elpa+mpi` explicitly.

repos/spack_repo/builtin/packages/abacus/package.py
RMeli: Yep. #6592 is now merged, so this can be updated accordingly.

Growl1234: Updated.

repos/spack_repo/builtin/packages/abacus/package.py
```diff
@@ -17,16 +17,22 @@ class Abacus(CMakePackage, MakefilePackage):
     for large-scale electronic-structure simulations
     from first principles"""
 
-    homepage = "http://abacus.ustc.edu.cn/"
-    url = "https://github.com/abacusmodeling/abacus-develop/archive/refs/tags/v3.9.0.19.tar.gz"
-    git = "https://github.com/abacusmodeling/abacus-develop.git"
+    homepage = "https://abacus.deepmodeling.com/"
+    url = "https://github.com/deepmodeling/abacus-develop/archive/refs/tags/v3.9.0.27.tar.gz"
```
RMeli, line 21: FYI, it is not necessary to update the version here, it is automatically substituted by Spack. I often use `v0.0.0.0` to make that clearer. But it does not hurt to change it either, so nothing to change. Just something to know for next time.

repos/spack_repo/builtin/packages/abacus/package.py
```diff
@@ -17,16 +17,22 @@ class Abacus(CMakePackage, MakefilePackage):
...
 
     license("LGPL-3.0-or-later")
 
     version("develop", branch="develop")
+    # Version 3.10.1 is an "LTS" release that predates 3.9.0.19
     version("3.10.1", sha256="06873eba8a4e0bc085177a6580455b28e4b62ea8a18f8afe71a02105756d91a0")
+    version(
+        "3.9.0.27",
+        sha256="066128d48517373b7b6d4d841f6bf2d8ceff94ae8c175ab089ff482b40789008",
+        preferred=True,
```
RMeli, line 36: Maybe worth adding a comment on why this is the case?

repos/spack_repo/builtin/packages/abacus/package.py
RMeli: Ok, since they are deprecated let's not bother. We can remove them soon.

repos/spack_repo/builtin/packages/abacus/package.py
```diff
@@ -49,35 +55,75 @@ class Abacus(CMakePackage, MakefilePackage):
...
-    depends_on("libxc", when="+libxc")
+    depends_on("mpi", when="+mpi")
+    depends_on("fftw precision=float,double +openmp", when="+openmp")
+    depends_on("fftw precision=float,double ~openmp", when="~openmp")
+    depends_on("elpa +mpi +openmp", when="+elpa +openmp")
+    depends_on("elpa +mpi ~openmp", when="+elpa ~openmp")
+    depends_on("libxc@5.1.7:", when="+libxc")
+    depends_on("dftd4@4.2:", when="+dftd4")
+    depends_on("libri", when="+libri")
+    depends_on("libcomm", when="+libri")
```
RMeli, line 81: `libri` already depends directly on `libcomm`. Just to make sure, `libcomm` is a direct dependency of `abacus` too when using `libri`?

Growl1234: ABACUS CMake find LibComm and other dependencies of LibRI when enabling LibRI, as both libri and libcomm have no build system
```cmake
if(ENABLE_LIBRI)
  set_if_higher(CMAKE_CXX_STANDARD 14)
  find_package(LibRI REQUIRED)
  find_package(LibComm REQUIRED)
  find_package(cereal REQUIRED CONFIG)
endif()
```
https://github.com/deepmodeling/abacus-develop/blob/734b8f4697af92631127da7d7be7c3781afdd17f/CMakeLists.txt#L603-L608

repos/spack_repo/builtin/packages/abacus/package.py
Growl1234: I think it's the similar reason to the comment on `3.10.1`. Unfortunately, their LTS versioning branch is now less robust than developing branch due to lack of backport mechanism.

Growl1234: I've updated above comment.

repos/spack_repo/builtin/packages/abacus/package.py
Growl1234: And from the code design, yes, ABACUS directly includes header files from LibComm and cereal when enabling LibRI.

RMeli (approved):
Thanks @Growl1234

